PR #13604, opened 2023-04-23
JDK-8304423: Refactor FdLibm.java

jddarcy:
Minor refactoring of FdLibm.java; will consider future refactorings with more substantive code changes in the future.

---------
### Progress

### Issue
 * [JDK-8304423](https://bugs.openjdk.org/browse/JDK-8304423): Refactor FdLibm.java

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13604/head:pull/13604` \
`$ git checkout pull/13604`

`$ git checkout pull/13604` \
`$ git pull https://git.openjdk.org/jdk.git pull/13604/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13604`

`$ git pr show -t 13604`

</details>

<a href="https://git.openjdk.org/jdk/pull/13604.diff">https://git.openjdk.org/jdk/pull/13604.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13604#issuecomment-1519185027)

### DIFF
--- a/src/java.base/share/classes/java/lang/FdLibm.java
+++ b/src/java.base/share/classes/java/lang/FdLibm.java
@@ -64,6 +64,16 @@ class FdLibm {
     private static final double TWO54    = 0x1.0p54; // 1.80143985094819840000e+16
     private static final double HUGE     = 1.0e+300;
 
+    /*
+     * Constants for bit-wise manipulation of IEEE 754 double
+     * values. These constants are for the high-order 32-bits of a
+     * 64-bit double value: 1 sign bit as the most significant bit,
+     * followed by 11 exponent bits, and then the remaining bits as
+     * the significand.
+     */
+    private static final int SIGN_BIT        = 0x8000_0000;
+    private static final int EXP_BITS        = 0x7ff0_0000;
+    private static final int EXP_SIGNIF_BITS = 0x7fff_ffff;
 
     private FdLibm() {
         throw new UnsupportedOperationException("No FdLibm instances for you.");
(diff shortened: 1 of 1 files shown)


src/java.base/share/classes/java/lang/FdLibm.java
```diff
@@ -64,6 +64,16 @@ class FdLibm {
...
+    /*
+     * Constants for bit-wise manipulation of IEEE 754 double
+     * values. These constants are for the high-order 32-bits of a
+     * 64-bit double value: 1 sign bit as the most significant bit,
+     * followed by 11 exponent bits, and then the remaining bits as
+     * the significand.
+     */
+    private static final int SIGN_BIT        = 0x8000_0000;
+    private static final int EXP_BITS        = 0x7ff0_0000;
+    private static final int EXP_SIGNIF_BITS = 0x7fff_ffff;
```
rgiulietti, line 76: These could be expressed in terms of the constants in `Double`. But perhaps it's not worth doing so, after all, as the reader is assumed to be quite knowledgeable.

Otherwise LGTM.

jddarcy: Yes, I had considered expressing these FDLIBM-internal constants in terms of constants from Double or the JDK-internal class DoubleConsts, where they could be more directly expressible based on 64-bit versions of the values. However, I thought having the constants be more directly assigned in this context would be fine.

If these assignments were to be amended, something like adding a static initializer block with asserts might be okay along the lines of

assert  SIGN_BIT == (int)(DoubleConsts.SIGN_BIT_MASK >>> 32) &&
    EXP_BITS == (int)(DoubleConsts.EXP_BIT_MASK >>> 32) ... ;

bplb (approved):
LGTM2

PR #13606, opened 2023-04-24
Jdk 8302736

tobiasholenstein:


tobiasholenstein:
> 



> Nice analysis, Toby. This point fix looks good to me.
> 
> As @theRealAph mentioned in the bug comments, and since there are other coarse-grained usages of `ThreadWXEnable` in the code (for example, in the `VM/JTR_ENTRY` macros), please file a follow-up RFE to improve this situation. The `ThreadWXEnable` should be as close as possible to the code that does the actual write access to the code cache.

Thanks! I filed https://bugs.openjdk.org/browse/JDK-8307817

dholmes-ora:
This is day one code for the macOS/Aarch64 port which has been in place for two years. Why is this only now being seen to be a problem?

The high-level placement of these calls was done to stop playing whack-a-mole every time we hit a new failure due to a missing `ThreadWXEnable`. I'm all for placing these where they are actually needed but noone seems to be to able to clearly state/identify exactly where that is in the code. The changes in this PR are pushing it down further, but based on the comments e.g.
```
// we might modify the code cache via BarrierSetNMethod::nmethod_entry_barrier
  MACOS_AARCH64_ONLY(ThreadWXEnable __wx(WXWrite, thread));
  return ConfigT::thaw(thread, (Continuation::thaw_kind)kind);
```
we are not pushing it down to where it is actually needed. The trade-off of course is that if we push this too far down we may have to execute it far more often and so take a performance hit. So figuring out the optimum placement for these in the call stack seems rather difficult.

dholmes-ora:
@tobiasholenstein what testing has been done on this? We may need to run all tiers (1-8) to ensure we have completely covered all the code paths affected. Thanks.

dean-long:
> The trade-off of course is that if we push this too far down we may have to execute it far more often and so take a performance hit. So figuring out the optimum placement for these in the call stack seems rather difficult.

Most code does not care what the WXWrite state is.  We could use an alternative approach where code that needs a particular WXWrite state sets it, but when it is done not change the state back.  So instead of using ThreadWXEnable RAII that resets the state when it goes out of scope, we would use thread->enable_wx(WXWrite) before writing into the code cache and we would use thread->enable_wx(WXExec) when transitioning from _thread_in_vm to _thread_in_Java thread state.  The implementation of enable_wx() already makes redundant state transitions cheap.  This allows us to move the thread->enable_wx(WXWrite) to immediately before the write into the code cache without needing to worry about finding an optimal coarser scope if the code writes into the code cache in multiple places.

theRealAph:
> > The trade-off of course is that if we push this too far down we may have to execute it far more often and so take a performance hit. So figuring out the optimum placement for these in the call stack seems rather difficult.
> 
> Most code does not care what the WXWrite state is. We could use an alternative approach where code that needs a particular WXWrite state sets it, but when it is done not change the state back.

Yes, I agree. Given a very fast way to query the current state, we can do the transition the first time we need to write into the code cache. Given a very fast way to query the current state, we'd be fine.

We could even consider trapping on a fault, enabling WX then returning in some cases. I guess there's no reason that'd not work, but I may be wrong.

tobiasholenstein:
> @tobiasholenstein what testing has been done on this? We may need to run all tiers (1-8) to ensure we have completely covered all the code paths affected. Thanks.

All tiers (1-10) were run and passed

tobiasholenstein:
> This is day one code for the macOS/Aarch64 port which has been in place for two years. Why is this only now being seen to be a problem?
I think the reason is _because_ this code exists since day one of macOS/Aarch64 port. We only detect performance regressions within the same OS/architecture. The regression was reported because someone ran code with `-XX:DisableIntrinsic=_dlog` and saw a large performance boost, whereas a performance decrease would be expected. Only after some investigation did we find out that the WXWrite lock was responsible. 

> The high-level placement of these calls was done to stop playing whack-a-mole every time we hit a new failure due to a missing `ThreadWXEnable`. I'm all for placing these where they are actually needed but noone seems to be to able to clearly state/identify exactly where that is in the code. The changes in this PR are pushing it down further, but based on the comments e.g.
> 
> ```
> // we might modify the code cache via BarrierSetNMethod::nmethod_entry_barrier
>   MACOS_AARCH64_ONLY(ThreadWXEnable __wx(WXWrite, thread));
>   return ConfigT::thaw(thread, (Continuation::thaw_kind)kind);
> ```
> 
> we are not pushing it down to where it is actually needed. The trade-off of course is that if we push this too far down we may have to execute it far more often and so take a performance hit. So figuring out the optimum placement for these in the call stack seems rather difficult.

I filed https://bugs.openjdk.org/browse/JDK-8307817 for this. But yes, care has to be taken - it seems tricky to get WXWrite lock right. Perhaps @dean-long 's suggestion is a better approach.

dholmes-ora:
> I think the reason is because this code exists since day one of macOS/Aarch64 port. 

Thanks for the explanation!

TobiHartmann:
> Why is this only now being seen to be a problem?

@dholmes-ora I think people usually don't compare performance between different architectures (macOS/Aarch64 vs. Linux/Aarch64) or when turning the intrinsic(s) on/off. That's why this was only noticed when comparing Math vs. StrictMath performance. Actually, @tobiasholenstein found some similar issues when comparing Math vs. StrictMath performance for other methods (also on x86_64). This will be investigated separately.

I agree that it's a mess and that we need a better approach for the WXWrite in general. But I would prefer to have this as point fix for the performance issues that we can potentially backport and investigate a general solution with [JDK-8307817](https://bugs.openjdk.org/browse/JDK-8307817).

jddarcy:
As a general comment, in case it is relevant, the remaining FDLIBM algorithms that were not already ported to Java have been ported to Java earlier in JDK 21 (JDK-8171407). This may change the performance of StrictMath.${FOO} methods on a given platform compared to earlier JDK releases.

theRealAph:
> As a general comment, in case it is relevant, the remaining FDLIBM algorithms that were not already ported to Java have been ported to Java earlier in JDK 21 (JDK-8171407). This may change the performance of StrictMath.${FOO} methods on a given platform compared to earlier JDK releases.

Good point. I'd forgotten about that. Maybe we should simply disable the intrinsic.

tobiasholenstein:
> Maybe we should simply disable the intrinsic.

I am not sure I understand what you mean with disabling the intrinsics. Do you mean in general or to fix `JDK-8302736`?

If intrinsics are disabled in C2, Math and StrictMath will have the same performance. As mentions in the PR description there are no intrinsics for `Math.tan`, `Math.exp`, `Math.log`, `Math.pow` and `Math.log10` on `macOS aarch64` but there exist c++ implementations in C2: `__ieee754_exp`, `__ieee754_pow`, `__ieee754_log`,` __ieee754_log10`, `SharedRuntime::dtan`

The math functions in `Math.java` are mostly just delegated to `StrictMath`:
```java
@IntrinsicCandidate
public static double log(double a) {
    return StrictMath.log(a); // default impl. delegates to StrictMath
}
```
The main difference is the `@IntrinsicCandidate` which allows C2 to replace the function with an `intrinsics` or if there is no intrinsics available for an architecture with a call to a c++ implementation of the math function. 

Here are the results of the JMH benchmarks Math vs StrictMath  on `mainline`, `mainline with intrinsics disabled` and with this fix `JDK-8302736` :

| JMH Benchmark    | master (intrinsic on)   | master (intrinsic off)  | JDK-8302736        | Arrch64 C2 impl. | 
| ---------------: | -----------------: | -----------------: | -----------------: | :--------------  | 
| **Math.exp**         |   **13358** ops/ms     |  161142 ops/ms     |  **200088** ops/ms     | c++ | 
| StrictMath.exp   |  161669 ops/ms     |  161474 ops/ms     |  161031 ops/ms     | - | 
| **Math.pow**         |   **21598** ops/ms     |  486085 ops/ms     |  **356691** ops/ms     | c++ | 
| StrictMath.pow   |  490299 ops/ms     |  491422 ops/ms     |  494714 ops/ms     | - | 
| **Math.log**         |   **16170** ops/ms     |  221149 ops/ms     |  **210370** ops/ms     | c++ | 
| StrictMath.log   |  224129 ops/ms     |  222821 ops/ms     |  222042 ops/ms     | - | 
| **Math.log10**       |   **14791** ops/ms     |  150701 ops/ms     |  **158154** ops/ms     | c++  | 
| StrictMath.log10 |  152683 ops/ms     |  151418 ops/ms     |  151211 ops/ms     | - | 
| Math.sin         |  267036 ops/ms     |  159221 ops/ms     |  268296 ops/ms     | intrinsic | 
| StrictMath.sin   |  158828 ops/ms     |  158736 ops/ms     |  159116 ops/ms     | - | 
| Math.cos         |  292302 ops/ms     |  173359 ops/ms     |  291640 ops/ms     | intrinsic | 
| StrictMath.cos   |  172939 ops/ms     |  172890 ops/ms     |  172562 ops/ms     | - | 
| **Math.tan**         |   **12475** ops/ms     |   98477 ops/ms     |   **83818** ops/ms     | c++ |
| StrictMath.tan   |   98716 ops/ms     |   98078 ops/ms     |   98835 ops/ms     | - |
| Math.ceil        | 1758784 ops/ms     | 1192446 ops/ms     | 1728416 ops/ms     | intrinsic |
| StrictMath.ceil  | 1189807 ops/ms     | 1197845 ops/ms     | 1193485 ops/ms   | - |
| Math.floor       | 1734748 ops/ms     | 1311019 ops/ms     | 1762023 ops/ms   | intrinsic |
| StrictMath.floor | 1311644 ops/ms     | 1312581 ops/ms     | 1304094 ops/ms  | - |

theRealAph:
> > Maybe we should simply disable the intrinsic.
> 
> I am not sure I understand what you mean with disabling the intrinsics. Do you mean in general or to fix `JDK-8302736`?

OK, so there's a slight advantage to the intrinsics, somewhere between 5-40%. I guess that's worth having, as long as we can tolerate the additional maintenance.

theRealAph:
> @dean-long and @theRealAph are you ok with this change as a point-fix?

I'm pretty nervous, to be honest. I think it'll work.
Could we add a write-enable to`PcDescCache::add_pc_desc`? I Don't know how often that function is used.

dean-long (approved):
It looks like if this causes a regression anywhere it would only be in exception throwing, so I think it's better than what we had.

theRealAph (approved):
Ok, for now. I think we need to revisit the way W^X is handed at some point.

theRealAph:
Ok, for now. I think we need to revisit the way W^X is handed at some point.

tobiasholenstein:
> > @dean-long and @theRealAph are you ok with this change as a point-fix?
> 
> I'm pretty nervous, to be honest. I think it'll work. Could we add a write-enable to`PcDescCache::add_pc_desc`? I Don't know how often that function is used.

`PcDescCache::add_pc_desc` is only called by `CompiledMethod::pc_desc_at` and `CompiledMethod::pc_desc_near` which each are called 10 and 4 times. 

> Ok, for now. I think we need to revisit the way W^X is handed at some point.
okey. I agree we should refine W^X locking at some point

tobiasholenstein:
Thanks @theRealAph , @dholmes-ora , @dean-long , @jddarcy and @TobiHartmann for the inputs and reviews!

PR #13607, opened 2023-04-24
8305566: Change StringDedup thread to derive from JavaThread

kimbarrett:
Please review this change to the string deduplication thread to make it a kind
of JavaThread rather than a ConcurrentGCThread.  There are several pieces to
this change:

(1) New class StringDedupThread (derived from JavaThread), separate from
StringDedup::Processor (which is now just a CHeapObj instead of deriving from
ConcurrentGCThread).  The thread no longer needs to or supports being stopped,
like other similar threads.  It also needs to be started later, once Java
threads are supported.  Also don't need an explicit visitor, since it will be
in the normal Java threads list.  This separation made the changeover a little
cleaner to develop, and made the servicability support a little cleaner too.

(2) The Processor now uses the ThreadBlockInVM idiom to be safepoint polite,
instead of using the SuspendibleThreadSet facility.

(3) Because we're using ThreadBlockInVM, which has a different usage style
from STS, the tracking of time spent by the processor blocked for safepoints
doesn't really work.  It's not very important anyway, since normal thread
descheduling can also affect the normal processing times being gathered and
reported.  So we just drop the so-called "blocked" time and associated
infrastructure, simplifying Stat tracking a bit.  Also renamed the
"concurrent" stat to be "active", since it's all in a JavaThread now.

(4) To avoid #include problems, moved the definition of
JavaThread::is_active_Java_thread from the .hpp file to the .inline.hpp file,
where one of the functions it calls also is defined.

(5) Added servicability support for the new thread.

Testing:
mach5 tier1-3 with -XX:+UseStringDeduplication.
The test runtime/cds/DeterministicDump.java fails intermittently with that
option, which is not surprising - see JDK-8306712.

I was never able to reproduce the failure; it's likely quite timing sensitive.
The fix of changing the type is based on StefanK's comment that ZResurrection
doesn't expect a non-Java thread to perform load-barriers.

---------
### Progress

### Issue
 * [JDK-8305566](https://bugs.openjdk.org/browse/JDK-8305566): Change StringDedup thread to derive from JavaThread

### Reviewers
 * [Stefan Karlsson](https://openjdk.org/census#stefank) (@stefank - **Reviewer**) ⚠️ Review applies to [d4e94b89](https://git.openjdk.org/jdk/pull/13607/files/d4e94b89061d6505d200d671c8fd6493bd903168)
 * [Chris Plummer](https://openjdk.org/census#cjplummer) (@plummercj - **Reviewer**) ⚠️ Review applies to [f17cc6be](https://git.openjdk.org/jdk/pull/13607/files/f17cc6be42dfda121e855fccda921997425f27c7)
 * [Thomas Schatzl](https://openjdk.org/census#tschatzl) (@tschatzl - **Reviewer**) ⚠️ Review applies to [f17cc6be](https://git.openjdk.org/jdk/pull/13607/files/f17cc6be42dfda121e855fccda921997425f27c7)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13607/head:pull/13607` \
`$ git checkout pull/13607`

`$ git checkout pull/13607` \
`$ git pull https://git.openjdk.org/jdk.git pull/13607/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13607`

`$ git pr show -t 13607`

</details>

<a href="https://git.openjdk.org/jdk/pull/13607.diff">https://git.openjdk.org/jdk/pull/13607.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13607#issuecomment-1519651652)


src/hotspot/share/gc/shared/stringdedup/stringDedupProcessor.hpp
```diff
@@ -25,26 +25,23 @@
 #ifndef SHARE_GC_SHARED_STRINGDEDUP_STRINGDEDUPPROCESSOR_HPP
 #define SHARE_GC_SHARED_STRINGDEDUP_STRINGDEDUPPROCESSOR_HPP
 
-#include "gc/shared/concurrentGCThread.hpp"
+#include "memory/allocation.hpp"
 #include "gc/shared/stringdedup/stringDedup.hpp"
```
stefank, line 29: sort order

stefank (approved):
I think this looks sensible to me, though I'm not very familiar with the current StringDedup code, so consider this a partial review only.

src/hotspot/share/gc/shared/stringdedup/stringDedupProcessor.hpp
kimbarrett: oops.  fixed.

shipilev:
Could we please change the synopsis to something more relevant? This PR does not only fix the ZGC test failure, but does a more fundamental change: switching string dedup thread from being `ConcurrentGCThread` to `JavaThread`, and so it affects more than one GC. The synopsis should reflect that, I think. (This would also be cleaner for potential backports, if any).

shipilev (approved):
I like this simplification a lot, thanks! I am running some Shenandoah string-dedup tests now.

shipilev (changes requested):
(Formally requesting the change of synopsis)

shipilev:
> I like this simplification a lot, thanks! I am running some Shenandoah string-dedup tests now.

Ran `gc/shenandoah/TestStringDedup*` on x86_64 and AArch64 for 100 times without a problem. These tests usually fail when there are bugs in string dedup. So I think we are clear there!

plummercj (approved):
SA changes look good

tschatzl (approved):
Please improve the bug name as suggested by @shipilev before pushing.

Looks good otherwise.

kimbarrett:
Thanks all for reviews.  I've changed the bug and PR summaries as requested.

PR #13611, opened 2023-04-24
8306042: C2: failed: Missed optimization opportunity in PhaseCCP (adding LShift->Cast->Add notification)

eme64:
An other case of `uncast` not being type-propagated through.

We have a case like this:
`Phi -> ShiftL -> CastII -> AndI`

The Phi has an updated type, so we should re-run Value on the AndI.

In PhaseCCP::push_and, we do update a similar pattern:
`X -> ShiftL -> AndI`

I extended it to handle this pattern:
`parent -> LShift (use) -> ConstraintCast* -> And`

For this, I implemented:
https://github.com/openjdk/jdk/blob/26f4adaae901822bea984b926c06d1a78f9c6b48/src/hotspot/share/opto/castnode.hpp#L73-L78

I could refactor code from a previous similar fix, for pattern: `ConstraintCast+ -> Sub/Phi`

**Discussion**

https://github.com/openjdk/jdk/blob/4d350f8f4eaabb18482c7656cb56a734e60187cf/src/hotspot/share/opto/castnode.hpp#L78-L79
I would have liked to place a `ResourceMark` between these two lines, to ensure the `internals` data structure is de-allocated after the traversal. But if I add it there, then one cannot modify any outer data-structure, or else one risks re-allocation of the outer data-structure in the inner ResourceMark, and then this memory gets de-allocated once the ResourceMark is cleared, and the outer data-structure is broken. This would for example mean that I could not push to the IGVN worklist inside the callback.

Not having the ResourceMark means a memory leak, until the compile phase is over. But my code is not the only place, there are lots of places where we create a Resource allocated data-structure, but do not use ResourceMarks.

---------
### Progress

### Issue
 * [JDK-8306042](https://bugs.openjdk.org/browse/JDK-8306042): C2: failed: Missed optimization opportunity in PhaseCCP (adding LShift-&gt;Cast-&gt;Add notification)

### Reviewers
 * [Christian Hagedorn](https://openjdk.org/census#chagedorn) (@chhagedorn - **Reviewer**) ⚠️ Review applies to [4d2a4f9e](https://git.openjdk.org/jdk/pull/13611/files/4d2a4f9eaeab7d7c91b8ceb948a0e99da9429773)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13611/head:pull/13611` \
`$ git checkout pull/13611`

`$ git checkout pull/13611` \
`$ git pull https://git.openjdk.org/jdk.git pull/13611/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13611`

`$ git pr show -t 13611`

</details>

<a href="https://git.openjdk.org/jdk/pull/13611.diff">https://git.openjdk.org/jdk/pull/13611.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13611#issuecomment-1521215828)

### DIFF
--- a/src/hotspot/share/opto/castnode.hpp
+++ b/src/hotspot/share/opto/castnode.hpp
@@ -69,6 +69,28 @@ public:
   static Node* make_cast_for_type(Node* c, Node* in, const Type* type, DependencyType dependency);
 
   Node* optimize_integer_cast(PhaseGVN* phase, BasicType bt);
+
+  // Visit all non-cast uses of the node, bypassing ConstraintCasts.
+  // Pattern: this (-> ConstraintCast)* -> non_cast
+  // In other words: find all non_cast nodes such that
+  // non_cast->uncast() == this.
+  template <typename Callback>
+  static void visit_uncasted_uses(const Node* n, Callback callback) {
+    ResourceMark rm;
+    Unique_Node_List internals;
+    internals.push((Node*)n); // start traversal
+    for (uint j = 0; j < internals.size(); ++j) {
+      Node* internal = internals.at(j); // for every internal
+      for (DUIterator_Fast kmax, k = internal->fast_outs(kmax); k < kmax; k++) {
+        Node* internal_use = internal->fast_out(k);
+        if (internal_use->is_ConstraintCast()) {
+          internals.push(internal_use); // traverse this cast also
+        } else {
+          callback(internal_use);
+        }
+      }
+    }
+  }
 };
 
 //------------------------------CastIINode-------------------------------------
--- a/src/hotspot/share/opto/castnode.hpp
+++ b/src/hotspot/share/opto/castnode.hpp
@@ -76,7 +76,6 @@ public:
   // non_cast->uncast() == this.
   template <typename Callback>
   static void visit_uncasted_uses(const Node* n, Callback callback) {
-    ResourceMark rm;
     Unique_Node_List internals;
     internals.push((Node*)n); // start traversal
     for (uint j = 0; j < internals.size(); ++j) {
--- a/src/hotspot/share/opto/castnode.hpp
+++ b/src/hotspot/share/opto/castnode.hpp
@@ -76,6 +76,7 @@ public:
   // non_cast->uncast() == this.
   template <typename Callback>
   static void visit_uncasted_uses(const Node* n, Callback callback) {
+    ResourceMark rm;
     Unique_Node_List internals;
     internals.push((Node*)n); // start traversal
     for (uint j = 0; j < internals.size(); ++j) {
--- a/src/hotspot/share/opto/phaseX.cpp
+++ b/src/hotspot/share/opto/phaseX.cpp
@@ -1951,7 +1951,7 @@ void PhaseCCP::analyze() {
   }
 
   // Push root onto worklist
-  Unique_Node_List worklist;
+  Unique_Node_List worklist(C->comp_arena());
   worklist.push(C->root());
   DEBUG_ONLY(Unique_Node_List worklist_verify;)
 
--- a/src/hotspot/share/opto/phaseX.cpp
+++ b/src/hotspot/share/opto/phaseX.cpp
@@ -1760,31 +1760,16 @@ void PhaseIterGVN::add_users_to_worklist( Node *n ) {
     }
 
     // If changed Cast input, notify down for Phi and Sub - both do "uncast"
+    // Patterns:
+    // ConstraintCast+ -> Sub
+    // ConstraintCast+ -> Phi
     if (use->is_ConstraintCast()) {
-      for (DUIterator_Fast i2max, i2 = use->fast_outs(i2max); i2 < i2max; i2++) {
-        Node* u = use->fast_out(i2);
-        if (u->is_Phi() || u->is_Sub()) {
-          // Phi (.., CastII, ..) or Sub(Cast(x), x)
-          _worklist.push(u);
-        } else if (u->is_ConstraintCast()) {
-          // Follow cast-chains down to Sub: Sub( CastII(CastII(x)), x)
-          // This case is quite rare. Let's BFS-traverse casts, to find Subs:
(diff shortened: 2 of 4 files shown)


src/hotspot/share/opto/castnode.hpp
```diff
@@ -69,6 +69,27 @@ class ConstraintCastNode: public TypeNode {
...
   Node* optimize_integer_cast(PhaseGVN* phase, BasicType bt);
+
+  // Visit all non-cast uses of the node, bypassing ConstraintCasts.
+  // Pattern: this (-> ConstraintCast)* -> non_cast
+  // In other words: find all non_cast nodes such that
+  // non_cast->uncast() == this.
+  template <typename Callback>
+  static void visit_uncasted_uses(const Node* n, Callback callback) {
+    Unique_Node_List internals;
+    internals.push((Node*)n); // start traversal
```
TobiHartmann, line 80: We should get rid of these casts with [JDK-8252694](https://bugs.openjdk.org/browse/JDK-8252694). I added a comment there.

test/hotspot/jtreg/compiler/ccp/TestShiftCastAndNotification.java
```diff
@@ -0,0 +1,62 @@
...
+ * @library /test/lib
+ * @run main/othervm -Xcomp -XX:-TieredCompilation
+ *                   -XX:CompileOnly=compiler.ccp.TestShiftCastAndNotification::test
+ *                   compiler.ccp.TestShiftCastAndNotification
+ */
+
+package compiler.ccp;
+
+public class TestShiftCastAndNotification {
+  static int N;
```
TobiHartmann, line 37: Should use 4-whitespace indentation.

TobiHartmann (approved):
Looks good to me. I like the new `visit_uncasted_uses`.

chhagedorn (approved):
Looks good!

eme64:
@chhagedorn @TobiHartmann Thanks for the reviews.
I could integrate now.

But I have one more discovery:
The reason why I cannot place the mentioned `ResoureMark` is that the `worklist` has its arena as the `Thread::current()->resource_area()` in some cases. One example is the `Unique_Node_List worklist` in `PhaseCCP::analyze`. So then in `PhaseCCP::push_and`, this `worklist` gets captured by the lambda, and passed to `ConstraintCastNode::visit_uncasted_uses`. If there is a `ResourceMark` inside it, and we re-allocate memory for the `worklist` because it grows beyond its current capacity, then we will de-allocate that memory when we exit the `ResourceMark`, and the `worklist` now points to de-allocated memory.

Without a `ResourceMark`, this can be avoided. But then we never release the memory for the `Unique_Node_List internals`. 

The second use of `ConstraintCastNode::visit_uncasted_uses` in `PhaseIterGVN::add_users_to_worklist` does not run into this issue, even if we have a `ResourceMark`. The reason is that there the `worklist` that is captured allocates from a separate arena, namely the `Compile::current()->comp_arena()`.

We could thus also consider having the `worklist` inside `PhaseCCP::analyze` allocate also from the `Compile::current()->comp_arena()`. One downside is that we never de-allocate memory in that arena until the end of `Compile`. But currently, there is also no `ResourceMark` upstream of `PhaseCCP::analyze`, all the way up to `Compile`. So this `worklist` never has its memory de-allocated anyway.

What is your opinion? Should I make such changes, or just integrate what I have now?

TobiHartmann (approved):
The new changes look good to me. Let's clean up memory management of worklists with [JDK-8302670](https://bugs.openjdk.org/browse/JDK-8302670).

eme64:
I now added the `worklist` to a local arena instead. I may change this again in [JDK-8302670](https://bugs.openjdk.org/browse/JDK-8302670), if that is beneficial for memory management of the CCP data-structures.

TobiHartmann (approved):

chhagedorn (approved):
Update looks good!

src/hotspot/share/opto/phaseX.cpp
```diff
@@ -1965,8 +1950,13 @@ void PhaseCCP::analyze() {
...
+  // CCP worklist is placed on a local arena, so that we can allow ResourceMarks on "Compile::current()->resource_arena()".
+  // We also do not want to put the worklist on "Compile::current()->comp_arena()", as that one only gets de-allocated after
+  // Compile is over. The local arena gets de-allocated at the end of its scope.
+  ResourceArea local_arena(mtCompiler);
+  Unique_Node_List worklist(&local_arena);
+
   // Push root onto worklist
-  Unique_Node_List worklist;
   worklist.push(C->root());
   DEBUG_ONLY(Unique_Node_List worklist_verify;)
```
vnkozlov, line 1961: Should you put `worklist_verify` to `local_arena` too?

eme64: @vnkozlov I could do that, but it is not required. The CCP `worklist` gets passed downward much farther, including the  graph walks in `push_child_nodes_to_worklist`. So there it is nice to be able to have `ResourceMarks`. But `worklist_verify` is only modified directly in `PhaseCCP::analyze` and one layer deeper in `PhaseCCP::verify_analyze`. So I would never expect a `ResourceMark` to mess with re-allocation.

So at this point the only reason to add `worklist_verify` to `local_arena` is to ensure it is de-allocated afterward. We could also use a `ResourceMark` and leave it on `Thread::current()->resource_area()`.

Anyway, I'll just move it to `local_arena`, after all it is `DEBUG_ONLY`.

vnkozlov: ok

vnkozlov (approved):

TobiHartmann (approved):
Still good. Ship it!

eme64:
@vnkozlov @chhagedorn @TobiHartmann Thanks for the reviews and discussion!
/integrate

PR #13615, opened 2023-04-24
8306729: Add nominal descriptors of modules and packages to Constants API

asotona:
Constants API already provides models for all loadable constants to help programs manipulating class files and modelling bytecode instructions. However no models of module and package constants are provided by Constants API. Every program manipulating class files must implement own models and validation of modules and packages constants.

This pul request adds `java.lang.constant.ModuleDesc` and `java.lang.constant.PackageDesc` to the Constants API. 

Classfile API will follow up and remove its internal implementations of `PackageDesc` and `ModuleDesc`. 

Please review this pull request and attached CSR.

Thank you,
Adam

---------
### Progress
- [x] Change requires CSR request [JDK-8306730](https://bugs.openjdk.org/browse/JDK-8306730) to be approved

### Issues
 * [JDK-8306729](https://bugs.openjdk.org/browse/JDK-8306729): Add nominal descriptors of modules and packages to Constants API
 * [JDK-8306730](https://bugs.openjdk.org/browse/JDK-8306730): Add nominal descriptors of modules and packages to Constants API (**CSR**)

### Reviewers
 * [Mandy Chung](https://openjdk.org/census#mchung) (@mlchung - **Reviewer**) ⚠️ Review applies to [efab745b](https://git.openjdk.org/jdk/pull/13615/files/efab745bb9eba8a26f4ffec16dd3d01b6c9e1608)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13615/head:pull/13615` \
`$ git checkout pull/13615`

`$ git checkout pull/13615` \
`$ git pull https://git.openjdk.org/jdk.git pull/13615/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13615`

`$ git pr show -t 13615`

</details>

<a href="https://git.openjdk.org/jdk/pull/13615.diff">https://git.openjdk.org/jdk/pull/13615.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13615#issuecomment-1520034320)

### DIFF
--- a/src/java.base/share/classes/java/lang/constant/ConstantUtils.java
+++ b/src/java.base/share/classes/java/lang/constant/ConstantUtils.java
@@ -76,6 +76,60 @@ class ConstantUtils {
      }
 
     /**
+     * Validates the correctness of a binary package name. In particular checks for the presence of
+     * invalid characters in the name.
+     *
+     * @param name the package name
+     * @return the package name passed if valid
+     * @throws IllegalArgumentException if the package name is invalid
+     */
+    public static String validateBinaryPackageName(String name) {
+        for (int i=0; i<name.length(); i++) {
+            char ch = name.charAt(i);
+            if (ch == ';' || ch == '[' || ch == '/')
+                throw new IllegalArgumentException("Invalid package name: " + name);
+        }
+        return name;
+    }
+
+    /**
+     * Validates the correctness of an internal package name.
+     * In particular checks for the presence of invalid characters in the name.
+     *
+     * @param name the package name
+     * @return the package name passed if valid
+     * @throws IllegalArgumentException if the package name is invalid
+     */
+    public static String validateInternalPackageName(String name) {
+        for (int i=0; i<name.length(); i++) {
+            char ch = name.charAt(i);
+            if (ch == ';' || ch == '[' || ch == '.')
+                throw new IllegalArgumentException("Invalid package name: " + name);
+        }
+        return name;
+    }
+
+    /**
+     * Validates the correctness of a module name. In particular checks for the presence of
+     * invalid characters in the name.
+     *
+     * {@jvms 4.2.3} Module and Package Names
+     *
+     * @param name the module name
+     * @return the module name passed if valid
+     * @throws IllegalArgumentException if the module name is invalid
+     */
+    public static String validateModuleName(String name) {
+        for (int i=name.length() - 1; i >= 0; i--) {
+            char ch = name.charAt(i);
+            if ((ch >= '\u0000' && ch <= '\u001F')
+            || ((ch == '\\' || ch == ':' || ch =='@') && (i == 0 || name.charAt(--i) != '\\')))
+                throw new IllegalArgumentException("Invalid module name: " + name);
+        }
+        return name;
+    }
+
+    /**
      * Validates a member name
      *
      * @param name the name of the member
--- a/src/java.base/share/classes/java/lang/constant/ConstantUtils.java
+++ b/src/java.base/share/classes/java/lang/constant/ConstantUtils.java
@@ -76,12 +76,14 @@ class ConstantUtils {
      }
 
     /**
-     * Validates the correctness of a binary package name. In particular checks for the presence of
-     * invalid characters in the name.
+     * Validates the correctness of a binary package name.
+     * In particular checks for the presence of invalid characters in the name.
+     * Empty package name is allowed.
      *
      * @param name the package name
      * @return the package name passed if valid
      * @throws IllegalArgumentException if the package name is invalid
+     * @throws NullPointerException if the package name is {@code null}
      */
     public static String validateBinaryPackageName(String name) {
         for (int i=0; i<name.length(); i++) {
@@ -95,10 +97,12 @@ class ConstantUtils {
     /**
      * Validates the correctness of an internal package name.
      * In particular checks for the presence of invalid characters in the name.
+     * Empty package name is allowed.
      *
      * @param name the package name
      * @return the package name passed if valid
      * @throws IllegalArgumentException if the package name is invalid
+     * @throws NullPointerException if the package name is {@code null}
      */
     public static String validateInternalPackageName(String name) {
         for (int i=0; i<name.length(); i++) {
@@ -110,14 +114,16 @@ class ConstantUtils {
     }
 
     /**
-     * Validates the correctness of a module name. In particular checks for the presence of
-     * invalid characters in the name.
+     * Validates the correctness of a module name.
+     * In particular checks for the presence of invalid characters in the name.
+     * Empty module name is allowed.
      *
      * {@jvms 4.2.3} Module and Package Names
      *
      * @param name the module name
      * @return the module name passed if valid
      * @throws IllegalArgumentException if the module name is invalid
+     * @throws NullPointerException if the module name is {@code null}
      */
     public static String validateModuleName(String name) {
         for (int i=name.length() - 1; i >= 0; i--) {
--- a/src/java.base/share/classes/java/lang/constant/ModuleDesc.java
+++ b/src/java.base/share/classes/java/lang/constant/ModuleDesc.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package java.lang.constant;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A nominal descriptor for a {@code Module} constant.
+ *
+ * <p>To create a {@linkplain ModuleDesc} for a module, use {@link #of}.
+ *
+ * @since 21
+ */
+public sealed interface ModuleDesc
+        permits ModuleDescImpl {
+
+    /**
+     * Returns a {@linkplain ModuleDesc} for a module,
+     * given the name of the module.
+     *
+     * <p>{@jvms 4.2.3} Module names are not encoded in "internal form" like
+     * class and interface names, that is, the ASCII periods (.) that separate
+     * the identifiers in a module name are not replaced by ASCII forward
+     * slashes (/).
+     *
+     * <p>Module names may be drawn from the entire Unicode codespace, subject
+     * to the following constraints:
+     * <ul>
+     * <li>A module name must not contain any code point in the range
+     * '&#92;u0000' to '&#92;u001F' inclusive.
+     * <li>The ASCII backslash (\) is reserved for use as an escape character in
+     * module names. It must not appear in a module name unless it is followed
+     * by an ASCII backslash, an ASCII colon (:), or an ASCII at-sign (@).
+     * The ASCII character sequence \\ may be used to encode a backslash in a
+     * module name.
+     * <li>The ASCII colon (:) and at-sign (@) are reserved for future use in
+     * module names.
+     * They must not appear in module names unless they are escaped.
+     * The ASCII character sequences \: and \@ may be used to encode a colon and
+     * an at-sign in a module name.
+     * </ul>
+     * @param name module name
+     * @return a {@linkplain ModuleDesc} describing the desired module
+     * @throws NullPointerException if the argument is {@code null}
+     * @throws IllegalArgumentException if the name string is not in the
+     * correct format
+     */
+    static ModuleDesc of(String name) {
+        ModuleDescImpl.validateModuleName(requireNonNull(name));
+        return new ModuleDescImpl(name);
+    }
+
+    /**
+     * Returns the module name of this {@linkplain ModuleDesc}.
+     *
+     * @return the module name
+     */
+    String moduleName();
+
+    /**
+     * Compare the specified object with this descriptor for equality.  Returns
+     * {@code true} if and only if the specified object is also a
+     * {@linkplain ModuleDesc} and both describe the same module.
+     *
+     * @param o the other object
+     * @return whether this descriptor is equal to the other object
+     */
+    @Override
+    boolean equals(Object o);
+}
--- a/src/java.base/share/classes/java/lang/constant/ModuleDesc.java
+++ b/src/java.base/share/classes/java/lang/constant/ModuleDesc.java
@@ -60,7 +60,7 @@ public sealed interface ModuleDesc
      *
      * @return the module name
      */
-    String moduleName();
+    String name();
 
     /**
      * Compare the specified object with this descriptor for equality.
--- a/src/java.base/share/classes/java/lang/constant/ModuleDesc.java
+++ b/src/java.base/share/classes/java/lang/constant/ModuleDesc.java
@@ -56,7 +56,7 @@ public sealed interface ModuleDesc
     }
 
     /**
-     * Returns the module name of this {@linkplain ModuleDesc}.
+     * Returns the module name of this {@link ModuleDesc}.
      *
      * @return the module name
      */
--- a/src/java.base/share/classes/java/lang/constant/ModuleDesc.java
+++ b/src/java.base/share/classes/java/lang/constant/ModuleDesc.java
@@ -29,7 +29,7 @@ import static java.util.Objects.requireNonNull;
 /**
  * A nominal descriptor for a {@code Module} constant.
  *
- * @apiNote
+ * <p>
  * To create a {@link ModuleDesc} for a module, use the {@link #of(String)}
  * method.
  *
--- a/src/java.base/share/classes/java/lang/constant/ModuleDesc.java
+++ b/src/java.base/share/classes/java/lang/constant/ModuleDesc.java
@@ -29,7 +29,9 @@ import static java.util.Objects.requireNonNull;
 /**
  * A nominal descriptor for a {@code Module} constant.
  *
- * <p>To create a {@link ModuleDesc} for a module, use the {@link #of} method.
+ * @apiNote
+ * To create a {@link ModuleDesc} for a module, use the {@link #of(String)}
+ * method.
  *
  * @jvms 4.4.11 The CONSTANT_Module_info Structure
  * @since 21
@@ -61,9 +63,9 @@ public sealed interface ModuleDesc
     String moduleName();
 
     /**
-     * Compare the specified object with this descriptor for equality.  Returns
-     * {@code true} if and only if the specified object is also a
-     * {@link ModuleDesc} and both describe the same module.
+     * Compare the specified object with this descriptor for equality.
+     * Returns {@code true} if and only if the specified object is
+     * also a {@link ModuleDesc} and both describe the same module.
      *
      * @param o the other object
      * @return whether this descriptor is equal to the other object
--- a/src/java.base/share/classes/java/lang/constant/ModuleDesc.java
+++ b/src/java.base/share/classes/java/lang/constant/ModuleDesc.java
@@ -27,9 +27,9 @@ package java.lang.constant;
 import static java.util.Objects.requireNonNull;
 
 /**
- * A nominal descriptor for a {@code Module} constant {@jvms 4.4.11}.
+ * A nominal descriptor for a {@code Module} constant.
  *
- * <p>To create a {@linkplain ModuleDesc} for a module, use the {@link #of} method.
+ * <p>To create a {@link ModuleDesc} for a module, use the {@link #of} method.
  *
  * @jvms 4.4.11 The CONSTANT_Module_info Structure
  * @since 21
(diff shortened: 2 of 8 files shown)


src/java.base/share/classes/java/lang/constant/PackageDescImpl.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Validates the correctness of a binary package name. In particular checks for the presence of
+     * invalid characters in the name.
+     *
+     * @param name the package name
+     * @return the package name passed if valid
+     * @throws IllegalArgumentException if the package name is invalid
+     */
+    public static String validateBinaryPackageName(String name) {
```
liach, line 37: All these utility methods should be moved to ConstantUtils.

src/java.base/share/classes/java/lang/constant/PackageDesc.java
```diff
@@ -0,0 +1,107 @@
...
+package java.lang.constant;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A nominal descriptor for a {@code Package} constant.
+ *
+ * <p>To create a {@linkplain PackageDesc} for a package, use {@link #of} or
+ * {@link #ofInternalName(String)}.
+ *
```
liach, line 34: Needs a link to jvms 4.4.12. Same for ModuleDesc (4.4.11)

asotona: fixed, thanks.

src/java.base/share/classes/java/lang/constant/PackageDescImpl.java
asotona: fixed, thanks.

src/java.base/share/classes/java/lang/constant/ConstantUtils.java
```diff
@@ -75,6 +75,60 @@ static String validateInternalClassName(String name) {
...
+     */
+    public static String validateModuleName(String name) {
+        for (int i=name.length() - 1; i >= 0; i--) {
+            char ch = name.charAt(i);
+            if ((ch >= '\u0000' && ch <= '\u001F')
+            || ((ch == '\\' || ch == ':' || ch =='@') && (i == 0 || name.charAt(--i) != '\\')))
+                throw new IllegalArgumentException("Invalid module name: " + name);
+        }
+        return name;
+    }
```
RogerRiggs, line 130: If these are performance sensitive or get used a lot, should there be an array of flags indexed by the byte/char to indicate the valid cases?

src/java.base/share/classes/java/lang/constant/ConstantUtils.java
```diff
@@ -75,6 +75,60 @@ static String validateInternalClassName(String name) {
          return name;
      }
 
+    /**
+     * Validates the correctness of a binary package name. In particular checks for the presence of
+     * invalid characters in the name.
```
RogerRiggs, line 80: It may be useful to note that the empty string are considered valid. 
Also add \@throws NPE; there's an implicit null check in checking the length.
Also in validateBinaryPackageName and validateModuleName.

src/java.base/share/classes/java/lang/constant/package-info.java
```diff
@@ -89,6 +89,11 @@
  * It is also suitable for describing {@code invokedynamic} call sites in bytecode
  * reading and writing APIs.
  *
+ * <p>Another members of this package are {@link java.lang.constant.ModuleDesc}
```
RogerRiggs, line 92: "Another" -> "Other"

src/java.base/share/classes/java/lang/constant/ConstantUtils.java
asotona: I'll add that notes, thanks.

src/java.base/share/classes/java/lang/constant/ModuleDescImpl.java
```diff
@@ -0,0 +1,28 @@
...
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package java.lang.constant;
+
+record ModuleDescImpl(String moduleName) implements ModuleDesc {
```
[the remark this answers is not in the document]
asotona: I'll add it, thanks.

test/jdk/java/lang/constant/PackageDescTest.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+import java.lang.constant.PackageDesc;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+class PackageDescTest {
+    @ParameterizedTest
```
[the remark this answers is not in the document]
asotona: added, thanks.

src/java.base/share/classes/java/lang/constant/package-info.java
asotona: Fixed, thanks.

ExE-Boss:
Note that other `*Impl` classes in `java.lang.constant` perform validation in their constructors and provide custom `toString()` formatting and they also don’t use records.

asotona:
> Note that other `*Impl` classes in `java.lang.constant` perform validation in their constructors and provide custom `toString()` formatting and they also don’t use records.

Performing validation in constructors I see as a blocker for potential performance optimisations from trusted code within the `java.lang.constant` package.

Custom `toString()` is a good point, returning for example `ModuleDescImpl[moduleName=mymodule]` is not ideal.
I'll fix it, thanks.

Records are very well designed exactly for this purpose and I'm not aware of any reason to don't use them.

asotona:
> Note that other `*Impl` classes in `java.lang.constant` perform validation in their constructors and provide custom `toString()` formatting and they also don’t use records.

BTW: for example `ClassDesc:of(String name)` performs repeated validations and I see it as a performance bug.
`ClassDesc::of` calls `ConstantUtils::validateBinaryClassName`, then it performs conversion `binaryToInternal` and calls `ClassDesc::ofDescriptor`, which checks for `ConstantUtils::arrayDepth` and calls `ReferenceClassDescImpl::new`, which again performns validation by calling `ConstantUtils::skipOverFieldSignature` and checking the result. There is plenty of space for performance improvements here, as all the validation and conversion can be done in one pass only.

src/java.base/share/classes/java/lang/constant/ModuleDesc.java
```diff
@@ -0,0 +1,92 @@
...
+     * by an ASCII backslash, an ASCII colon (:), or an ASCII at-sign (@).
+     * The ASCII character sequence \\ may be used to encode a backslash in a
+     * module name.
+     * <li>The ASCII colon (:) and at-sign (@) are reserved for future use in
+     * module names.
+     * They must not appear in module names unless they are escaped.
+     * The ASCII character sequences \: and \@ may be used to encode a colon and
+     * an at-sign in a module name.
+     * </ul>
+     * @param name module name
```
mlchung, line 64:
```suggestion
     * @param name the module name
```

src/java.base/share/classes/java/lang/constant/PackageDesc.java
```diff
@@ -0,0 +1,107 @@
...
+ */
+package java.lang.constant;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A nominal descriptor for a {@code Package} constant {@jvms 4.4.12}.
+ *
+ * <p>To create a {@linkplain PackageDesc} for a package, use {@link #of} or
+ * {@link #ofInternalName(String)}.
```
mlchung, line 33:
```suggestion
 * <p>To create a {@linkplain PackageDesc} for a package, use the {@link #of} or
 * {@link #ofInternalName(String)} method.
```

src/java.base/share/classes/java/lang/constant/PackageDesc.java
```diff
@@ -0,0 +1,107 @@
...
+ * @since 21
+ */
+public sealed interface PackageDesc
+        permits PackageDescImpl {
+
+    /**
+     * Returns a {@linkplain PackageDesc} for a package,
+     * given the name of the package, such as {@code "java.lang"}.
+     * <p>
+     * {@jls 13.1}
```
mlchung, line 44: Do you mean to reference JLS 6.7?

Suggest to move this reference after `throws` in the see also section.

src/java.base/share/classes/java/lang/constant/PackageDesc.java
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * Returns a {@linkplain PackageDesc} for a package,
+     * given the name of the package in internal form,
+     * such as {@code "java/lang"}.
+     * <p>
+     * {@jvms 4.2.1} In this internal form, the ASCII periods (.) that normally
+     * separate the identifiers
+     * which make up the binary name are replaced by ASCII forward slashes (/).
+     * @param name the fully qualified class name, in internal (slash-separated)
```
mlchung, line 65:
```suggestion
     * @param name the fully qualified package name, in internal (slash-separated) form
```

src/java.base/share/classes/java/lang/constant/PackageDesc.java
```diff
@@ -0,0 +1,107 @@
...
+public sealed interface PackageDesc
+        permits PackageDescImpl {
+
+    /**
+     * Returns a {@linkplain PackageDesc} for a package,
+     * given the name of the package, such as {@code "java.lang"}.
+     * <p>
+     * {@jls 13.1}
+     *
+     * @param name the fully qualified (dot-separated) binary package name
```
mlchung, line 46: JLS 13.1 does not describe a package has a binary name.   Is it correct to say "binary package name"?

```suggestion
     * @param name the fully qualified (dot-separated) package name
```

src/java.base/share/classes/java/lang/constant/PackageDesc.java
```diff
@@ -0,0 +1,107 @@
...
+        ConstantUtils.validateBinaryPackageName(requireNonNull(name));
+        return new PackageDescImpl(ConstantUtils.binaryToInternal(name));
+    }
+
+    /**
+     * Returns a {@linkplain PackageDesc} for a package,
+     * given the name of the package in internal form,
+     * such as {@code "java/lang"}.
+     * <p>
+     * {@jvms 4.2.1} In this internal form, the ASCII periods (.) that normally
```
mlchung, line 62: Suggest not to copy JVMS 4.2.1 here but instead just add it to see also section.

src/java.base/share/classes/java/lang/constant/PackageDesc.java
```diff
@@ -0,0 +1,107 @@
...
+     * default package
+     */
+    String packageInternalName();
+
+    /**
+     * Returns the fully qualified (dot-separated) binary package name
+     * of this {@linkplain PackageDesc}.
+     *
+     * @return the package name, or the empty string for the
+     * default package
```
mlchung, line 91:
```suggestion
     * unnamed package
```

See JLS 7.4.2

src/java.base/share/classes/java/lang/constant/PackageDesc.java
```diff
@@ -0,0 +1,107 @@
...
+        ConstantUtils.validateInternalPackageName(requireNonNull(name));
+        return new PackageDescImpl(name);
+    }
+
+    /**
+     * Returns the fully qualified (slash-separated) internal package name
+     * of this {@linkplain PackageDesc}.
+     *
+     * @return the package name, or the empty string for the
+     * default package
```
mlchung, line 82:
```suggestion
     * unnamed package
```

src/java.base/share/classes/java/lang/constant/package-info.java
```diff
@@ -89,6 +89,11 @@
  * It is also suitable for describing {@code invokedynamic} call sites in bytecode
  * reading and writing APIs.
  *
+ * <p>Other members of this package are {@link java.lang.constant.ModuleDesc}
+ * and  {@link java.lang.constant.PackageDesc}. They represent module and package
```
mlchung, line 93:
```suggestion
 * <p>Other members of this package are {@link ModuleDesc}
 * and  {@link PackageDesc}. They represent module and package
```

same package and so no need to qualify.

src/java.base/share/classes/java/lang/constant/ModuleDesc.java
```diff
@@ -0,0 +1,92 @@
...
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package java.lang.constant;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A nominal descriptor for a {@code Module} constant {@jvms 4.4.11}.
```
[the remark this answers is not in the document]
asotona: Fixed, thanks.

src/java.base/share/classes/java/lang/constant/PackageDesc.java
```diff
@@ -0,0 +1,107 @@
...
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package java.lang.constant;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A nominal descriptor for a {@code Package} constant {@jvms 4.4.12}.
```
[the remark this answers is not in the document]
asotona: Fixed, thanks.

src/java.base/share/classes/java/lang/constant/PackageDesc.java
asotona: Fixed, thanks.

src/java.base/share/classes/java/lang/constant/ModuleDesc.java
```diff
@@ -0,0 +1,92 @@
...
+ * @since 21
+ */
+public sealed interface ModuleDesc
+        permits ModuleDescImpl {
+
+    /**
+     * Returns a {@linkplain ModuleDesc} for a module,
+     * given the name of the module.
+     *
+     * <p>{@jvms 4.2.3} Module names are not encoded in "internal form" like
```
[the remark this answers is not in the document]
asotona: Fixed, thanks.

src/java.base/share/classes/java/lang/constant/PackageDesc.java
asotona: Fixed, thanks.

src/java.base/share/classes/java/lang/constant/PackageDesc.java
```diff
@@ -0,0 +1,110 @@
...
+ * questions.
+ */
+package java.lang.constant;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A nominal descriptor for a {@code Package} constant.
+ *
+ * @apiNote
```
mlchung, line 32: can drop `@apiNote` as it's not needed to be. `<p>` if you want to a new paragraph.

src/java.base/share/classes/java/lang/constant/PackageDesc.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Returns a {@link PackageDesc} for a package,
+     * given the name of the package, such as {@code "java.lang"}.
+     *
+     * @param name the fully qualified (dot-separated) package name
+     * @return a {@link PackageDesc} describing the desired package
+     * @throws NullPointerException if the argument is {@code null}
+     * @throws IllegalArgumentException if the name string is not in the
+     * correct format
+     * @jls 6.7 Fully Qualified Names and Canonical Names
```
mlchung, line 51:
```suggestion
     * @jls 6.5.3 Module Names and Package Names
```

src/java.base/share/classes/java/lang/constant/PackageDesc.java
```diff
@@ -0,0 +1,110 @@
...
+     * given the name of the package in internal form,
+     * such as {@code "java/lang"}.
+     *
+     * @param name the fully qualified package name, in internal
+     * (slash-separated) form
+     * @return a {@link PackageDesc} describing the desired package
+     * @throws NullPointerException if the argument is {@code null}
+     * @throws IllegalArgumentException if the name string is not in the
+     * correct format
+     * @jvms 4.2.1 Binary Class and Interface Names
```
mlchung, line 70: Worth adding `@jvm 4.2.3 Module and Package Names`

src/java.base/share/classes/java/lang/constant/PackageDesc.java
```diff
@@ -0,0 +1,110 @@
...
+     * @see PackageDesc#of(String)
+     */
+    static PackageDesc ofInternalName(String name) {
+        ConstantUtils.validateInternalPackageName(requireNonNull(name));
+        return new PackageDescImpl(name);
+    }
+
+    /**
+     * Returns the fully qualified (slash-separated) internal package name
```
mlchung, line 79:
```suggestion
     * Returns the fully qualified (slash-separated) package name in internal form
```

src/java.base/share/classes/java/lang/constant/PackageDesc.java
```diff
@@ -0,0 +1,110 @@
...
+     * of this {@link PackageDesc}.
+     *
+     * @return the package name, or the empty string for the
```
mlchung, line 82:
```suggestion
     * @return the package name in internal form, or the empty string for the
```

src/java.base/share/classes/java/lang/constant/ModuleDesc.java
```diff
@@ -0,0 +1,75 @@
...
+ * questions.
+ */
+package java.lang.constant;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A nominal descriptor for a {@code Module} constant.
+ *
+ * @apiNote
```
mlchung, line 32: can drop `@apiNote` as it's not needed to be.  `<p>` if you want to a new paragraph.

src/java.base/share/classes/java/lang/constant/PackageDesc.java
```diff
@@ -0,0 +1,110 @@
...
+     * unnamed package
+     * @see PackageDesc#packageName()
+     */
+    String packageInternalName();
+
+    /**
+     * Returns the fully qualified (dot-separated) binary package name
```
mlchung, line 89:
```suggestion
     * Returns the fully qualified (dot-separated) package name
```

mlchung (commented):
I wonder if `packageName()` and `packageInternalName()` methods can simply be `name()` and `internalName()` as the type name  is `PackageDesc` and `package` prefix seems to be unnecessary.   Similarly, `moduleName()` can be `name()`.   Have this be discussed and rejected?

minborg:
Have we considered moving the Impl classes into a non-public area such as `jdk.internal` to improve encapsulation and reduce perceived footprint?

src/java.base/share/classes/java/lang/constant/PackageDesc.java
asotona: Set back to `<p>` as suggested, thanks.

src/java.base/share/classes/java/lang/constant/PackageDesc.java
asotona: Added, thanks.

src/java.base/share/classes/java/lang/constant/ModuleDesc.java
asotona: Set back to `<p>` as suggested, thanks.

asotona:
> I wonder if `packageName()` and `packageInternalName()` methods can simply be `name()` and `internalName()` as the type name is `PackageDesc` and `package` prefix seems to be unnecessary. Similarly, `moduleName()` can be `name()`. Have this be discussed and rejected?

I've searched the history of Classfile API discussions and it actually wasn't discussed.
I agree the prefix is unnecessary. 
Thanks for pointing it out.

mlchung:
> Have we considered moving the Impl classes into a non-public area such as `jdk.internal` to improve encapsulation and reduce perceived footprint?

I don't see much of a problem to keep the non-public classes in `java.lang.constant` since it's not open and no reflective access.   Not sure what perceived footprint you referred to as the JDK footprint won't change.

src/java.base/share/classes/java/lang/constant/PackageDesc.java
```diff
@@ -0,0 +1,111 @@
...
+     * such as {@code "java/lang"}.
+     *
+     * @param name the fully qualified package name, in internal
+     * (slash-separated) form
+     * @return a {@link PackageDesc} describing the desired package
+     * @throws NullPointerException if the argument is {@code null}
+     * @throws IllegalArgumentException if the name string is not in the
+     * correct format
+     * @jvms 4.2.1 Binary Class and Interface Names
+     * @jvm 4.2.3 Module and Package Names
```
mlchung, line 71:
```suggestion
     * @jvms 4.2.3 Module and Package Names
```

src/java.base/share/classes/java/lang/constant/PackageDesc.java
```diff
@@ -0,0 +1,111 @@
...
+     * @see PackageDesc#internalName()
+     */
+    default String name() {
+        return ConstantUtils.internalToBinary(internalName());
+    }
+
+    /**
+     * Compare the specified object with this descriptor for equality.
+     * Returns {@code true} if and only if the specified object is
+     * also a {@linkp PackageDesc} and both describe the same package.
```
mlchung, line 104:
```suggestion
     * also a {@link PackageDesc} and both describe the same package.
```

mlchung (approved):
Thanks for the update.   Two typos.  please do `make docs` to verify the output.

src/java.base/share/classes/java/lang/constant/ConstantUtils.java
```diff
@@ -75,6 +75,66 @@ static String validateInternalClassName(String name) {
...
+     * @return the module name passed if valid
+     * @throws IllegalArgumentException if the module name is invalid
+     * @throws NullPointerException if the module name is {@code null}
+     */
+    public static String validateModuleName(String name) {
+        for (int i=name.length() - 1; i >= 0; i--) {
+            char ch = name.charAt(i);
+            if ((ch >= '\u0000' && ch <= '\u001F')
+            || ((ch == '\\' || ch == ':' || ch =='@') && (i == 0 || name.charAt(--i) != '\\')))
+                throw new IllegalArgumentException("Invalid module name: " + name);
```
AlanBateman, line 133: test/jdk/java/lang/module/ModuleNames.java has tables of legal and illegal module names, including tests that escape backslash, @, and :. It might be useful to run these tests on this method.

mlchung:
I keep pondering if this API should allow creating `ModuleDesc` and `PackageDesc` with an empty name.  

JLS 6.5.3. Meaning of Module Names and Package Names
> The module name M, whether simple or qualified, denotes the module (if any) with that name.

The module system will reject a `CONSTANT_Module_info` of zero-length name while JVMS does not prohibit it.

The module system rejects the unnamed package in a named module.    `CONSTANT_Package_info` represents a package exported or opened by a module.

src/java.base/share/classes/java/lang/constant/ConstantUtils.java
asotona: I've updated `ModuleDescTest` with all the positive and negative cases from `test/jdk/java/lang/module/ModuleNames.java`, except for the empty name.
Thanks for the review.

mlchung:
Discussed with @briangoetz offline.   We can let `PackageDesc` and `ModuleDesc` model unnamed package/module but the ClassFile API should check to make sure their names are not empty before using them to write to a classfile.

For example, if we had a constructor in `ClassDesc(PackageDesc package, String name)`, it could be tolerant of empty packages and treat it as just `new ClassDesc(name)`.

PR #13616, opened 2023-04-24
8306738: Select num workers for safepoint ParallelCleanupTask

xmas92:
Select num workers for safepoint ParallelCleanupTask to be at most the number of parallel cleanup tasks `SAFEPOINT_CLEANUP_NUM_TASKS`

Unsure if `WorkerThreads::max_workers()` is more correct than `WorkerThreads::created_workers()` or `WorkerThreads::active_workers()`. 

Tested tier1-3 Oracle platforms.

### DIFF
--- a/src/hotspot/share/runtime/safepoint.cpp
+++ b/src/hotspot/share/runtime/safepoint.cpp
@@ -603,7 +603,8 @@ void SafepointSynchronize::do_cleanup_tasks() {
   WorkerThreads* cleanup_workers = heap->safepoint_workers();
   if (cleanup_workers != nullptr) {
     // Parallel cleanup using GC provided thread pool.
-    cleanup_workers->run_task(&cleanup);
+    const uint num_workers = MIN2<uint>(SAFEPOINT_CLEANUP_NUM_TASKS, cleanup_workers->max_workers());
+    cleanup_workers->run_task(&cleanup, num_workers);
   } else {
     // Serial cleanup using VMThread.
     cleanup.work(0);


src/hotspot/share/runtime/safepoint.cpp
```diff
@@ -603,7 +603,8 @@ void SafepointSynchronize::do_cleanup_tasks() {
   WorkerThreads* cleanup_workers = heap->safepoint_workers();
   if (cleanup_workers != nullptr) {
     // Parallel cleanup using GC provided thread pool.
-    cleanup_workers->run_task(&cleanup);
+    const uint num_workers = MIN2<uint>(SAFEPOINT_CLEANUP_NUM_TASKS, cleanup_workers->max_workers());
+    cleanup_workers->run_task(&cleanup, num_workers);
```
shipilev, line 607: If you go to the definition of `WorkerThreads::run_task(WorkerTask* task)`, then you'll notice it uses `_active_workers`. So we should be doing `MIN2(..., cleanup_workers->active_workers()` here if the current behavior is to be preserved better. I think active workers is correct in the presence of dynamic GC worker count.

shipilev (commented):
The idea makes sense.

src/hotspot/share/runtime/safepoint.cpp
shipilev: Now that I am looking at it, I wonder if we have a pre-existing performance issue here: what if active GC workers is very low for some GC-specific reason? We then run all safepoint cleanups, even for non-GC safepoints, using that small worker count? This might be the accidental reliance on the number of GC workers to match the static fraction of the number of CPUs allocated to the JVM, which is often true, but this would not hold exactly under `UseDynamicNumberOfGCThreads`.

tschatzl: It is a "known" issue that if the particular task does not select its number of threads properly before executing the task (or not at all like this task) that there may be perf issues.

If this is what you want I would suggest that `active_workers()` as limit preserves the current behavior best. Otherwise the task should calculate and set its optimal number of tasks itself as it does in this change.

Not sure if setting it to `SAFEPOINT_CLEANUP_NUM_TASKS` is a good idea if some/most of this work is trivial/empty but it's probably a good first guess.

The only GC that does not seem to set safepoint_workers to something non-default seems to be Shenandoah (the stw collectors use the "main" workerthreads they also use for evacuation, so this should be maxed out mostly, and ZGC explicitly sets them to max_workers at initialization).

tschatzl: @xmas92: what is the reason for this change, did you see any situation where an inappropriate amount (too few) number of threads were used? Just curious.

shipilev: BTW, as we are micro-optimizing here, then we can be quite a bit smarter about this:

```
class ParallelCleanupTask {

...
  size_t expected_workers() {
    // This worker does the light-weight tasks and guarantees progress
    size_t workers = 1;

    // If there are heavy-weight tasks pending, add more workers for them
    if (SymbolTable::needs_rehashing()) workers++;
    if (StringTable::needs_rehashing()) workers++;
    if (_do_lazy_roots)                 workers++;

    return workers;
  }
```

xmas92: @tschatzl I noticed the behaviour when I added JFR events to worker threads activity, when I was profiling generational ZGC. There we have separate WorkerThreads for each generation. And a pool a "Runtime" WorkerThreads which is only used for `CollectedHeap::safepoint_workers()`. We do not configure this and only give the VM ParallelGCThreads number of workers. 

It seems correct to respect this flag for heap inspection and heap dump, but seemed strange and unnecessary to spin up all workers when we know that there is a limit to the amount of parallel work. (Just looking at the code, the heap inspect always uses one worker, and heap dumper uses the active number of workers)

Overall my thoughts on which number to use is that:
  * **Active**: Is a bit of a strange API, because it depends on some external set state, it is the current behaviour, so maybe it should be kept and changed in a different PR.
  * **Created**: Some sort of best effort, only use as many as the system has created so far, also seems to respect `UseDynamicNumberOfGCThreads` better. 
  * **Max**: We are a task that requires X workers, we have at least X workers in `CollectedHeap::safepoint_workers()` let us use them. Sort of the most pure API, and is easier to reason about.

The idea with this PR was mostly to put an upper limit to the number of workers. So I will change to use the `active_workers`. 

Maybe `expected_workers()` is worthwhile, I have no data on this. This PR was not motivated by any performance numbers, only the missmatch between the number of workers used and the amount of work available.

tschatzl: >It seems correct to respect this flag for heap inspection and heap dump, but seemed strange and unnecessary to spin up all workers when we know that there is a limit to the amount of parallel work. (Just looking at the code, the heap inspect always uses one worker, and heap dumper uses the active number of workers)

Not sure. Imho both should determine the number of threads based on some rough estimate of work too.

>[....]
>Active: Is a bit of a strange API, because it depends on some external set state, it is the current behaviour, so maybe it should be kept and changed in a different PR.

This has been long time ago added purely for convenience to not have pass it along in a sequence of parallel tasks.
I.e. in a lot of cases the "active" number is determined at start of gc (based on e.g. heap size/overall expected work), and then reduced based on the particular task like suggested in this CR.

It is true that this value should probably not be stored in `WorkerThreads`, but then the GCs would need to pass it along elsewhere.

shipilev:
I still believe we can do it significantly better! The key here is not only minimizing the parallelism beyound the number of tasks, but also recognize that most tasks are so light-weight, they do not need a separate worker thread. In fact, if we estimate we execute _only_ the light-weight tasks, we might as well do things serially without any thread handoff.

For example, my crude experiment (patch below) cuts down the safepoint cleanup time from ~40 us to ~1 us on my M1 Mac running G1 with `while (true) sink = new int[10000]` with 1 GB heap. Pretty sure we can polish it even further.

```
diff --git a/src/hotspot/share/runtime/safepoint.cpp b/src/hotspot/share/runtime/safepoint.cpp
index 2ff593a0143..64dcd0202ae 100644
--- a/src/hotspot/share/runtime/safepoint.cpp
+++ b/src/hotspot/share/runtime/safepoint.cpp
@@ -547,6 +547,19 @@ public:
     _do_lazy_roots(!VMThread::vm_operation()->skip_thread_oop_barriers() &&
                    Universe::heap()->uses_stack_watermark_barrier()) {}
 
+  size_t expected_workers() {
+    // This worker does the light-weight tasks and guarantees progress
+    size_t workers = 1;
+
+    // Potential heavy-weights, allocate another worker for them
+    // TODO: InlineCacheBuffer should be here too.
+    if (_do_lazy_roots) {
+      workers++;
+    }
+
+    return workers;
+  }
+
   void work(uint worker_id) {
     // These tasks are ordered by relative length of time to execute so that potentially longer tasks start first.
     if (_subtasks.try_claim_task(SafepointSynchronize::SAFEPOINT_CLEANUP_SYMBOL_TABLE_REHASH)) {
@@ -601,9 +614,11 @@ void SafepointSynchronize::do_cleanup_tasks() {
   assert(heap != nullptr, "heap not initialized yet?");
   ParallelCleanupTask cleanup;
   WorkerThreads* cleanup_workers = heap->safepoint_workers();
-  if (cleanup_workers != nullptr) {
+  size_t num_workers = cleanup.expected_workers();
+  if (num_workers > 1 && cleanup_workers != nullptr) {
     // Parallel cleanup using GC provided thread pool.
-    cleanup_workers->run_task(&cleanup);
+    num_workers = MIN2<uint>(cleanup_workers->active_workers(), num_workers);
+    cleanup_workers->run_task(&cleanup, num_workers);
   } else {
     // Serial cleanup using VMThread.
     cleanup.work(0);
```

tschatzl:
> I still believe we can do it significantly better! The key here is not only minimizing the parallelism beyond the number of tasks, but also recognize that most tasks are so light-weight, they do not need a separate worker thread. In fact, if we estimate we execute _only_ the light-weight tasks, we might as well do things serially without any thread handoff.
> 
> For example, my crude experiment (patch below) cuts down the safepoint cleanup time from ~40 us to ~1 us on my M1 Mac [...]
> 
> Pretty sure we can polish it even further.
> 
[...]
> 
> This could be done in a separate PR, of course, but I don't see why not test some approaches here :)

+1

xmas92:
> I still believe we can do it significantly better! The key here is not only minimizing the parallelism beyond the number of tasks, but also recognize that most tasks are so light-weight, they do not need a separate worker thread. In fact, if we estimate we execute _only_ the light-weight tasks, we might as well do things serially without any thread handoff.
> 
> For example, my crude experiment (patch below) cuts down the safepoint cleanup time from ~40 us to ~1 us on my M1 Mac running G1 with `while (true) sink = new int[10000]` with 1 GB heap. The cleanup time outliers up to 200us are present in current baseline, completely gone with the patch. The whole G1 collection safepoint takes ~700 us in that experiment, so that improvement is quite palpable: before the patch, the cleanup took about 5% of total safepoint time on average and up to 30% in spikes.
> 
> Pretty sure we can polish it even further.
> 
> ```
> diff --git a/src/hotspot/share/runtime/safepoint.cpp b/src/hotspot/share/runtime/safepoint.cpp
> index 2ff593a0143..64dcd0202ae 100644
> --- a/src/hotspot/share/runtime/safepoint.cpp
> +++ b/src/hotspot/share/runtime/safepoint.cpp
> @@ -547,6 +547,19 @@ public:
>      _do_lazy_roots(!VMThread::vm_operation()->skip_thread_oop_barriers() &&
>                     Universe::heap()->uses_stack_watermark_barrier()) {}
>  
> +  size_t expected_workers() {
> +    // This worker does the light-weight tasks and guarantees progress
> +    size_t workers = 1;
> +
> +    // Potential heavy-weights, allocate another worker for them
> +    // TODO: InlineCacheBuffer should be here too.
> +    if (_do_lazy_roots) {
> +      workers++;
> +    }
> +
> +    return workers;
> +  }
> +
>    void work(uint worker_id) {
>      // These tasks are ordered by relative length of time to execute so that potentially longer tasks start first.
>      if (_subtasks.try_claim_task(SafepointSynchronize::SAFEPOINT_CLEANUP_SYMBOL_TABLE_REHASH)) {
> @@ -601,9 +614,11 @@ void SafepointSynchronize::do_cleanup_tasks() {
>    assert(heap != nullptr, "heap not initialized yet?");
>    ParallelCleanupTask cleanup;
>    WorkerThreads* cleanup_workers = heap->safepoint_workers();
> -  if (cleanup_workers != nullptr) {
> +  size_t num_workers = cleanup.expected_workers();
> +  if (num_workers > 1 && cleanup_workers != nullptr) {
>      // Parallel cleanup using GC provided thread pool.
> -    cleanup_workers->run_task(&cleanup);
> +    num_workers = MIN2<uint>(cleanup_workers->active_workers(), num_workers);
> +    cleanup_workers->run_task(&cleanup, num_workers);
>    } else {
>      // Serial cleanup using VMThread.
>      cleanup.work(0);
> ```
> 
> This could be done in a separate PR, of course, but I don't see why not test some approaches here :)

Sounds promising. I can look into this in this PR.

We essentially want to find which tasks are cheaper than spinning up worker threads. And when the parallel work can amortise the spinning up worker threads cost.

Given that this comment exists `// These tasks are ordered by relative length of time to execute so that potentially longer tasks start first.` With the order:
1.  `SAFEPOINT_CLEANUP_SYMBOL_TABLE_REHASH`
2.  `SAFEPOINT_CLEANUP_STRING_TABLE_REHASH`
3.  `SAFEPOINT_CLEANUP_LAZY_ROOT_PROCESSING`
4.  `SAFEPOINT_CLEANUP_UPDATE_INLINE_CACHES`
5.  `SAFEPOINT_CLEANUP_REQUEST_OOPSTORAGE_CLEANUP`

Given that you @shipilev have 3 and 4 as heavier tasks than 1 and 2, and some other original developer had a different estimation, it would probably be better to have some more dynamic estimator for the actual work of each task.

But it is probably worthwhile here to have the `ParallelCleanupTask` query the subsystems at creation and figure out before we launch the task how many different tasks to perform, and then start the tasks based on this. It might be likely that in most cases <2 cleanup task have to be performed and worker threads are unnecessary. It also makes it possible to have the `task needs cleanup` property be stable over the lifetime of the ParallelCleanupTask.

Also if we have cleanup tasks that should always be lumped together into one parallel task should probably add some `enum SafepointParallelCleanupTasks` and claim based on these.

So we need to investigate the cost of each operation, the possibility (and granularity) of estimating the work for each task and when the cost of spinning up worker threads is amortised by the parallel speedup.

xmas92:
> > So we need to investigate the cost of each operation, the possibility (and granularity) of estimating the work for each task and when the cost of spinning up worker threads is amortised by the parallel speedup.
> 
> Yes.
> 
> I think the guiding principle here is that thread hand-off (delivering the wake up signal, wake up lag, waiting for worker thread termination) can easily take tens of microseconds. We established that in Parallel Streams work in JDK 8. So for tasks that only notify other, already threaded subsystems, spinning up the thread just to perform the notification of _another_ thread seems wasteful.
> 
> I think the only subtasks that do actual work are lazy roots (since they walk the unbounded number of threads) and IC buffer cleanups (when they have work, they can walk lots of IC stubs). The compounding factor is that string/symbol/oopstorage cleanup notification paths all take `ServiceLock`, and not only we spin up worker threads unnecessarily to essentially deliver a flag update, we also contend the worker threads over that lock! Oops. I wonder if that's where part of the cleanup time spikes comes from.
> 
> I believe that we don't need to go very deep in this. The crude patch I have above is probably already 80% there, we "only" need to handle the IC buffer updates there, and test this actually works on many scenarios. There is no need, IMO, in trying to micro-optimize the stability of "ParallelCleanupTask::estimated_workers", or change the way we claim the subtasks, etc, since from the ballbark experiments I had, those things consume about a microsecond -- diminishing returns for the complexity the deeper rewrite would entail. Just estimating the number of safepoint workers gets us lots of bang for little buck :)

Alright, my only fear is underestimating the work for some workload, and providing a change where we increase the safepoint times due to selecting to few workers.

coleenp:
I think I put the safepoint cleanup ordering code in assuming that the symbol table and string table rehashing takes a long time when it needs rehashing.  But most/all of the time, it only notifies the ServiceThread that these tables need resizing.  You can put those two at the end.  From what I can tell from this, you only need 3 threads maximum to do the cleanup tasks.  Aleksey's expected_workers() function makes a lot of sense to me.

shipilev:
@xmas92, what is your plan for this? It would be perfect to get it into JDK 21!

xmas92:
> @xmas92, what is your plan for this? It would be perfect to get it into JDK 21!
Same!

I propose an implementation which estimates the expected workers conservatively. That is:
  * Keep the previous behaviour with respect to `active_workers()` (Never more than `active_workers()`)
  * Use at most the expected number of workers, which is:
    * +1 if SymbolTable needs to actually rehash with alt hash
    * +1 if StringTable needs to actually rehash with alt hash
    * +1 If InlineCacheBuffer has any work to do
    * +1 If thread roots requires processing
  * If only one (or zero) workers are expected do not use any worker threads

This seems like a good first step too me, which in the worst case results in four workers and in the common case one or two, depending on GC and VM state.

There are some potential RFEs here to improve the expected workers for inline cache buffers and thread roots where we can find some lower bound  where the work required cannot amortise the cost of using worker threads. 

Also I am not 100% happy with the `uint xxx_expected_workers()` function names and signatures. So if there are any alternative suggestions, I am all ears. 

(Also the xxxTable code duplication is pretty atrocious)

src/hotspot/share/runtime/safepoint.cpp
```diff
@@ -547,6 +547,20 @@ class ParallelCleanupTask : public WorkerTask {
     _do_lazy_roots(!VMThread::vm_operation()->skip_thread_oop_barriers() &&
                    Universe::heap()->uses_stack_watermark_barrier()) {}
 
+  uint expected_num_workers() const {
```
coleenp, line 550: should this be 'static' for internal linkage?

coleenp (approved):
I think this looks fine from the symbol/string table perspective. This rehashing should never be done except in very exceptional circumstances.

> (Also the xxxTable code duplication is pretty atrocious)

Yes.  It is.

src/hotspot/share/runtime/safepoint.cpp
xmas92: It is a method and uses the non static member field `_do_lazy_roots`. Not a free function.

tschatzl (approved):
lgtm

shipilev:
I plan to look at it today, thanks.

src/hotspot/share/classfile/stringTable.cpp
```diff
@@ -529,20 +531,26 @@ bool StringTable::do_rehash() {
   return true;
 }
 
+bool StringTable::should_grow() {
+  return get_load_factor() > PREF_AVG_LIST_LEN && !_local_table->is_max_size_reached();
+}
+
+uint StringTable::rehash_table_expected_workers() {
```
shipilev, line 538: We always reply either `0` or `1` from these. So logically, it seems to answer whether we need heavy work. So, this method is then simpler, and matching the structure of the relevant `rehash_table`:

```
// Quick checks if we would do any heavy-weight safepoint work in rehash_table.
// Add shortcut cases here when changing rehash_table.
bool StringTable::needs_safepoint_rehash_table() {
  // Grow instead of rehash
  if (should_grow()) return false;

  // Already rehashed
  if (_rehashed) return false;

  // Should and can do alt-rehashing
  if (!_needs_rehashing) return false;
  if (!_local_table->is_safepoint_safe()) return false;

  return true;
}
```

shipilev (commented):
This looks very promising to me, but I have a few readability suggestions.

src/hotspot/share/runtime/safepoint.cpp
coleenp: oh ok, good.

src/hotspot/share/classfile/stringTable.cpp
xmas92: I prefer using a boolean return value as well. I just had trouble with coming up with a non confusing name for the method. Maybe `needs_safepoint_rehash_table` is good enough.

I'll rewrite the condition with multiple ifs. 

My knowledge of the concurrent hash table is rather limited, but what does the `Should and can do alt-rehashing` comment mean? AFAIK the `_local_table->is_safepoint_safe()` check is just to see if someone is currently resizing the table when we entered the safepoint.

xmas92: I see now, you are describing the negation. `!(!_needs_rehashing || !_local_table->is_safepoint_safe())`
Confused me as the other comments described the condition.
I think I'll rewrite the comment to be more in the same style as the once above.

shipilev: Yeah, feel free. I think the readability improvement is not in comments themselves, but in repeating the basic structure of `rehash_table()` :)

shipilev (approved):
I like this, thanks!

shipilev:
> I like this, thanks!

Rechecked my previous experiment with this new patch. For `java -Xlog:safepoint -Xms1g -Xmx1g -XX:+AlwaysPreTouch -XX:+UseParallelGC Alloc.java`, we get about 50us back!

plot

xmas92:
@coleenp @tschatzl 
I'll leave this open until tomorrow for if you have any comments on 1ca4aab23863629d544a8ebe8e87fadda90a0031 and then integrate.

coleenp (approved):
The new function rehash_table_expects... looks fine in this change, and thank you for the comments.  I think your code pushes the motivation for a superclass to have these duplicated functions over the top.  I'll file another RFE for that for later.  Thank you @shipilev for doing the performance measurements.

PR #13621, opened 2023-04-24
8306137: Open source several AWT ScrollPane related tests

TejeshR13:
Open source few AWT ScrollPane related tests.

---------
### Progress

### Issue
 * [JDK-8306137](https://bugs.openjdk.org/browse/JDK-8306137): Open source several AWT ScrollPane related tests

### Reviewers
 * [Alexey Ivanov](https://openjdk.org/census#aivanov) (@aivanov-jdk - **Reviewer**) ⚠️ Review applies to [84c0961c](https://git.openjdk.org/jdk/pull/13621/files/84c0961cb44569b951b2e4489eb65627620fbf41)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13621/head:pull/13621` \
`$ git checkout pull/13621`

`$ git checkout pull/13621` \
`$ git pull https://git.openjdk.org/jdk.git pull/13621/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13621`

`$ git pr show -t 13621`

</details>

<a href="https://git.openjdk.org/jdk/pull/13621.diff">https://git.openjdk.org/jdk/pull/13621.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13621#issuecomment-1520584760)


test/jdk/java/awt/ScrollPane/ComponentScrollTest/ComponentScrollTest.java
```diff
@@ -0,0 +1,99 @@
...
+import java.awt.event.AdjustmentEvent;
+import java.awt.event.AdjustmentListener;
+
+import java.lang.reflect.InvocationTargetException;
+
+public class ComponentScrollTest implements AdjustmentListener {
+    public ScrollPane scrollpane;
+    public Frame frame;
+    public int count = 0;
+    public static void main(String[] args) throws InterruptedException, InvocationTargetException {
```
lawrence-andrew, line 48: Can you please leave a blank line.

test/jdk/java/awt/ScrollPane/ComponentScrollTest/ComponentScrollTest.java
```diff
@@ -0,0 +1,99 @@
...
+                    Thread.sleep(5000);
+                } catch (InterruptedException ie) {
+                }
+                System.out.println("Count = " + count);
+                if (count > 50) {
+                    throw new RuntimeException();
+                }
+            });
+        } finally {
+            EventQueue.invokeLater(() -> {
```
mrserb, line 87: Why the invokeLater is used here?

TejeshR13: I preferred asynchronous EDT would be better since its finally block. Is there any particular reason that we should use `invokeAndWait` than `invokeLater`?

aivanov-jdk: > I preferred asynchronous EDT would be better since its finally block. Is there any particular reason that we should use `invokeAndWait` than `invokeLater`?

You want to **wait** until the frame is disposed of before exiting the main thread otherwise it doesn't make sense.

aivanov-jdk (commented):
I thought the preferred layout for tests is to use a flat layout without creating *additional folder* for each test.

So, `test/jdk/java/awt/ScrollPane/ComponentScrollTest/ComponentScrollTest.java` should be under `ScrollPane`: `test/jdk/java/awt/ScrollPane/ComponentScrollTest.java`.

Similarly, `test/jdk/java/awt/ScrollPane/ScrollPaneInsets/ScrollPaneExtraScrollBar.java` goes to `test/jdk/java/awt/ScrollPane/ScrollPaneExtraScrollBar.java`.

The test `test/jdk/java/awt/ScrollPane/RemoveChild` consists of two .java source files. Can the Semaphore class be moved into `ScrollPaneRemoveAdd.java` as a static nested class?

test/jdk/java/awt/ScrollPane/ComponentScrollTest/ComponentScrollTest.java
prrace: yes, I just noticed this in a different proposed change and I agree wait makes more sense.

prrace:
> I thought the preferred layout for tests is to use a flat layout without creating _additional folder_ for each test.
> 
> So, `test/jdk/java/awt/ScrollPane/ComponentScrollTest/ComponentScrollTest.java` should be under `ScrollPane`: `test/jdk/java/awt/ScrollPane/ComponentScrollTest.java`.
> 
> Similarly, `test/jdk/java/awt/ScrollPane/ScrollPaneInsets/ScrollPaneExtraScrollBar.java` goes to `test/jdk/java/awt/ScrollPane/ScrollPaneExtraScrollBar.java`.
> 
> The test `test/jdk/java/awt/ScrollPane/RemoveChild` consists of two .java source files. Can the Semaphore class be moved into `ScrollPaneRemoveAdd.java` as a static nested class?

Yes, remove these redundant components where the test consists of just one file.

jtreg creates a separate WD anyway and the longer the path length the more likely we'll run into problems on windows.

test/jdk/java/awt/ScrollPane/RemoveChild/Semaphore.java
aivanov-jdk, line 1: If it's left as an external file, it requires its own copyright header. I suggest moving it into `ScrollPaneRemoveAdd.java` as static nested class and then drop `RemoveChild` directory.

Please add a blank line between methods.

test/jdk/java/awt/ScrollPane/ScrollPaneInsets/ScrollPaneExtraScrollBar.java
```diff
@@ -0,0 +1,96 @@
...
+import java.awt.Frame;
+import java.awt.Insets;
+import java.awt.Rectangle;
+import java.awt.Robot;
+import java.awt.ScrollPane;
+
+import java.awt.event.InputEvent;
+
+public class ScrollPaneExtraScrollBar
+{
```
aivanov-jdk, line 42:
```suggestion
public class ScrollPaneExtraScrollBar {
```
The opening brace should be on the previous line.

test/jdk/java/awt/ScrollPane/ScrollPaneInsets/ScrollPaneExtraScrollBar.java
```diff
@@ -0,0 +1,96 @@
...
+    Robot robot;
+
+    public static void main(String[] args) throws Exception {
+        ScrollPaneExtraScrollBar scrollTest = new ScrollPaneExtraScrollBar();
+        scrollTest.init();
+        scrollTest.start();
+    }
+    public void init() throws Exception {
+
+       robot = new Robot();
```
aivanov-jdk, line 54:
```suggestion
    }

   public void init() throws Exception {
        robot = new Robot();
```
A blank line between methods is missing.  
The method starts with an unnecessary blank line.  
The first statement is not indented by four spaces (but only by three).

test/jdk/java/awt/ScrollPane/ScrollPaneInsets/ScrollPaneExtraScrollBar.java
```diff
@@ -0,0 +1,96 @@
...
+
+        EventQueue.invokeAndWait(() -> {
+                f = new Frame();
+                sp = new ScrollPane(ScrollPane.SCROLLBARS_AS_NEEDED);
+                sp.add(new Button("TEST"));
+                f.add("Center", sp);
+                f.pack();
+                f.setLocationRelativeTo(null);
+                f.setVisible(true);
+        });
```
aivanov-jdk, line 64: The lambda expression should be indented by four spaces less.

test/jdk/java/awt/ScrollPane/ScrollPaneInsets/ScrollPaneExtraScrollBar.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    public void start() throws Exception {
+        try {
+            if (System.getProperty("os.name").startsWith("Windows")) {
```
aivanov-jdk, line 69: Should we use [`OSInfo` class](https://github.com/openjdk/jdk/blob/master/src/java.desktop/share/classes/sun/awt/OSInfo.java)?

It's used in [`bug4847375.java`](https://github.com/openjdk/jdk/blob/master/test/jdk/javax/swing/JFileChooser/4847375 /bug4847375.java), [`bug6550546.java`](https://github.com/openjdk/jdk/blob/master/test/jdk/javax/swing/JFileChooser/6550546/bug6550546.java) and others.

The `@requires` tag should be added to the test tags so that it's not selected for other platforms.

To avoid extensive indentation, reverse the condition in the `if` statement and return early.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation/ScrollPaneLimitation.java
```diff
@@ -0,0 +1,174 @@
...
+        // Check for Win32 environment. This test is only for Win32.
+        // Even more: not for Windows 98: it doesn't support scroll children that long.
+        try {
+            Properties prop = System.getProperties();
+
+            String os = prop.getProperty("os.name", "");
+            if (os != null &&
+                    (os.indexOf("98") == -1) &&
+                    (os.startsWith("Win") || os.startsWith("win") ||
+                            os.startsWith("WIN"))) {
```
aivanov-jdk, line 83: Should use `OSInfo`?

`@requires` tag should be added to the test so that it's not selected for other platforms.

test/jdk/java/awt/ScrollPane/ScrollPaneInsets/ScrollPaneExtraScrollBar.java
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/*
+  @test
+  @bug 4152524
+  @summary Test that scroll pane doesn't have scroll bars visible when it is shown for the first time with SCROLLBARS_AS_NEEDED style
```
aivanov-jdk, line 27: Wrap long lines so that they fit into 80-column limit or, at the very least, into 100-column limit.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation/ScrollPaneLimitation.java
```diff
@@ -0,0 +1,174 @@
...
+                    frame.pack();
+                    frame.setLocationRelativeTo(null);
+                    frame.setVisible(true);
+                    pane.doLayout();
+
+                    try {
+                        Thread.sleep(1000);
+                    } catch (InterruptedException e) {
+                        throw new RuntimeException("test was interrupted");
+                    }
```
aivanov-jdk, line 113:
```suggestion
                    robot.delay(1000);
```
Much cleaner, isn't it?

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation/ScrollPaneLimitation.java
```diff
@@ -0,0 +1,174 @@
...
+        } finally {
+            EventQueue.invokeLater(() -> {
+                if (frame != null) {
+                    frame.dispose();
+                }
+            });
+        }
+    }
+
+    class MyPanel extends Component {
```
aivanov-jdk, line 169: Can be made `static`.
```suggestion
    private static class MyPanel extends Component {
```

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest/ScrollPaneWindowsTest.java
```diff
@@ -0,0 +1,188 @@
...
+
+    public void start() throws Exception {
+        Properties prop = System.getProperties();
+        String os = prop.getProperty("os.name", "").toUpperCase();
+        System.out.println("OS= " + os);
+        if (!os.equals("WINDOWS 2000") && !os.equals("WINDOWS 2003") &&
+                !os.equals("WINDOWS XP")) {
+            System.out.println("This test is for Windows 2000/2003/XP only.");
+            return;
+        }
```
aivanov-jdk, line 91: Again, use `OSInfo` or better `@requires` tag.

test/jdk/java/awt/ScrollPane/ScrollPositionIntact/ScrollPositionIntact.java
```diff
@@ -0,0 +1,106 @@
...
+                frame = new Frame();
+                frame.setSize(200, 100);
+                frame.add(sp);
+                frame.setVisible(true);
+
+                try {
+                    Thread.sleep(1000);
+                } catch (Exception ex) {
+                    throw new RuntimeException("Test failure: interruption?\n\n");
+                }
```
aivanov-jdk, line 77: It doesn't make sense: you're blocking EDT. It's probably meant to be run on main thread to allow EDT to process events and display the frame.

aivanov-jdk (changes requested):

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest/ScrollPaneWindowsTest.java
prrace: The whole comment needs reworking and thinking about.
Is it trying to say this test is obsolete after XP ?
In which case since even the successors of Visa, WIndows 7, 8 and 8.1 are all dead
the comment is really really unwelcome in 2023. And perhaps the test is too.

test/jdk/java/awt/ScrollPane/ComponentScrollTest/ComponentScrollTest.java
TejeshR13: Ok, didn't think in that sense, I'll better use invokeAndWait.....

test/jdk/java/awt/ScrollPane/RemoveChild/ScrollPaneRemoveAdd.java
```diff
@@ -0,0 +1,115 @@
...
+
+            Point buttonLoc = button.getLocationOnScreen();
+            Dimension buttonSize = button.getSize();
+            robot.mouseMove(buttonLoc.x + buttonSize.width / 2, buttonLoc.y + buttonSize.height / 2);
+            robot.delay(50);
+            robot.mousePress(InputEvent.BUTTON1_DOWN_MASK);
+            robot.delay(50);
+            robot.mouseRelease(InputEvent.BUTTON1_DOWN_MASK);
+
+            Toolkit.getDefaultToolkit().sync();
```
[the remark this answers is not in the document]
TejeshR13: Not sure y both has been used here, will remove sync and test in CI, if its fine there then can be removed.

test/jdk/java/awt/ScrollPane/ScrollPositionIntact/ScrollPositionIntact.java
TejeshR13: Yeah, I've added robot delay in-between EDT.

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest/ScrollPaneWindowsTest.java
TejeshR13: Ok, then I will not move the test, should it be left as applet or it should be removed?

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation/ScrollPaneLimitation.java
TejeshR13: Done.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation/ScrollPaneLimitation.java
TejeshR13: Done.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation/ScrollPaneLimitation.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneInsets/ScrollPaneExtraScrollBar.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneInsets/ScrollPaneExtraScrollBar.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneInsets/ScrollPaneExtraScrollBar.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneInsets/ScrollPaneExtraScrollBar.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/RemoveChild/Semaphore.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/RemoveChild/ScrollPaneRemoveAdd.java
```diff
@@ -0,0 +1,115 @@
...
+                    frame.pack();
+                    frame.setLocationRelativeTo(null);
+                    frame.setVisible(true);
+                    pane.remove(0);
+                    pane.add(button);
+            });
+
+            Point buttonLoc = button.getLocationOnScreen();
+            Dimension buttonSize = button.getSize();
+            robot.mouseMove(buttonLoc.x + buttonSize.width / 2, buttonLoc.y + buttonSize.height / 2);
```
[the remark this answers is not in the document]
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/RemoveChild/ScrollPaneRemoveAdd.java
```diff
@@ -0,0 +1,115 @@
...
+
+    public void start() throws Exception {
+        try {
+            EventQueue.invokeAndWait(() -> {
+                    frame.pack();
+                    frame.setLocationRelativeTo(null);
+                    frame.setVisible(true);
+                    pane.remove(0);
+                    pane.add(button);
+            });
```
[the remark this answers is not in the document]
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPositionIntact/ScrollPositionIntact.java
```diff
@@ -0,0 +1,106 @@
...
+            EventQueue.invokeLater(new Runnable() {
+                public void run() {
+                    int i;
+                    i = (int) (sp.getScrollPosition().getX());
+                    if (i == 0) {
+                        // actual position MAY be not equal to X_POS; still, it must be > 0.
+                        throw new RuntimeException("Test failure: zero scroll position.\n\n");
+                    }
+                }
+            });
```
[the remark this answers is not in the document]
TejeshR13: Updated to `invokeAndWait`() with lambda expressions.

test/jdk/java/awt/ScrollPane/RemoveChild/ScrollPaneRemoveAdd.java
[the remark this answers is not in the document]
TejeshR13: > > I thought the preferred layout for tests is to use a flat layout without creating _additional folder_ for each test.
> > So, `test/jdk/java/awt/ScrollPane/ComponentScrollTest/ComponentScrollTest.java` should be under `ScrollPane`: `test/jdk/java/awt/ScrollPane/ComponentScrollTest.java`.
> > Similarly, `test/jdk/java/awt/ScrollPane/ScrollPaneInsets/ScrollPaneExtraScrollBar.java` goes to `test/jdk/java/awt/ScrollPane/ScrollPaneExtraScrollBar.java`.
> > The test `test/jdk/java/awt/ScrollPane/RemoveChild` consists of two .java source files. Can the Semaphore class be moved into `ScrollPaneRemoveAdd.java` as a static nested class?
> 
> Yes, remove these redundant components where the test consists of just one file.
> 
> jtreg creates a separate WD anyway and the longer the path length the more likely we'll run into problems on windows.

Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest/ScrollPaneWindowsTest.java
mrserb: probably the idea was to exclude "win 98/etc"? please check for what bug this test was created and if is it applicable for win10 or not.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation.java
```diff
@@ -70,18 +72,12 @@ public void init() throws Exception {
...
         try {
             Properties prop = System.getProperties();
-
             String os = prop.getProperty("os.name", "");
-            if (os != null &&
-                    (os.indexOf("98") == -1) &&
-                    (os.startsWith("Win") || os.startsWith("win") ||
-                            os.startsWith("WIN"))) {
-
+            if (os != null && (os.indexOf("98") == -1)) {
```
aivanov-jdk, line 80: Shall we drop the filter for Windows 98? Java cannot be run on Windows 9x for quite a long time, so it doesn't make sense to preserve this condition.

test/jdk/java/awt/ScrollPane/ScrollPaneRemoveAdd.java
```diff
@@ -76,22 +76,23 @@ public void actionPerformed(ActionEvent e) {
...
-            robot.mouseMove(buttonLoc.x + buttonSize.width / 2, buttonLoc.y + buttonSize.height / 2);
+            robot.mouseMove(buttonLoc.x + buttonSize.width / 2,
+                    buttonLoc.y + buttonSize.height / 2);
             robot.delay(50);
             robot.mousePress(InputEvent.BUTTON1_DOWN_MASK);
             robot.delay(50);
             robot.mouseRelease(InputEvent.BUTTON1_DOWN_MASK);
 
-            Toolkit.getDefaultToolkit().sync();
+   //         Toolkit.getDefaultToolkit().sync();
```
aivanov-jdk: Instead of commenting out the line, remove it.

test/jdk/java/awt/ScrollPane/ScrollPaneRemoveAdd.java
```diff
@@ -102,14 +103,66 @@ public void start() throws Exception {
...
                 if (frame != null) {
                     frame.dispose();
                 }
             });
         }
     }
+
+    private static class Semaphore {
+        volatile boolean state = false;
+        Object lock = new Object();
```
aivanov-jdk, line 120: Since it's a lock object, it should be `final`.

test/jdk/java/awt/ScrollPane/ScrollPaneRemoveAdd.java
```diff
@@ -102,14 +103,66 @@ public void start() throws Exception {
...
+        volatile int waiting = 0;
```
aivanov-jdk, line 121: Why are both `state` and `waiting` marked as volatile? Either is accessed from synchronized blocks protected by `lock` except for `getState` which should also use the `lock` object.

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest/ScrollPaneWindowsTest.java
TejeshR13: I tested locally as well as in CI system, all the run's were green.

test/jdk/java/awt/ScrollPane/ScrollPaneExtraScrollBar.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+/*
+  @test
+  @bug 4152524
+  @requires os.family=="windows"
+  @summary Test that scroll pane doesn't have scroll bars visible when it is
+  shown for the first time with SCROLLBARS_AS_NEEDED style
+  @key headful
+*/
```
aivanov-jdk, line 31: In majority of cases, the tags have a leading asterisk.

test/jdk/java/awt/ScrollPane/ComponentScrollTest.java
```diff
@@ -0,0 +1,99 @@
...
+import java.awt.event.AdjustmentEvent;
+import java.awt.event.AdjustmentListener;
+
+import java.lang.reflect.InvocationTargetException;
+
+public class ComponentScrollTest implements AdjustmentListener {
+    public ScrollPane scrollpane;
+    public Frame frame;
+    public int count = 0;
+    public static void main(String[] args) throws InterruptedException, InvocationTargetException {
```
aivanov-jdk, line 48:
```suggestion
    public int count = 0;

    public static void main(String[] args)
            throws InterruptedException, InvocationTargetException {
```
Add a blank line between fields and method, wrap `throws` clause to fit into 80-column limit.

test/jdk/java/awt/ScrollPane/ComponentScrollTest.java
```diff
@@ -0,0 +1,99 @@
...
+                    public Dimension getPreferredSize() {
+                        return new Dimension(500, 500);
+                    }
+                    public void paint(Graphics g) {
+                        g.drawLine(0, 0, 500, 500);
+                    }
+                });
+                frame.add(scrollpane);
+        });
+        scrollpane.getVAdjustable().addAdjustmentListener(this);
```
aivanov-jdk, line 68: The lambda expression has incorrect indentation.

You should add the listener on EDT too.

test/jdk/java/awt/ScrollPane/ComponentScrollTest.java
```diff
@@ -0,0 +1,99 @@
...
+    public void start() throws InterruptedException, InvocationTargetException {
+        try {
+            EventQueue.invokeAndWait(() -> {
+                frame.pack();
+                frame.setVisible(true);
+                scrollpane.getVAdjustable().setValue(20);
+                try {
+                    Thread.sleep(5000);
+                } catch (InterruptedException ie) {
+                }
```
aivanov-jdk, line 80: This `sleep` blocks EDT. Is it intended?

test/jdk/java/awt/ScrollPane/ScrollPaneRemoveAdd.java
```diff
@@ -0,0 +1,168 @@
...
+            EventQueue.invokeAndWait(() -> {
+                frame.pack();
+                frame.setLocationRelativeTo(null);
+                frame.setVisible(true);
+                pane.remove(0);
+                pane.add(button);
+            });
+
+            Point buttonLoc = button.getLocationOnScreen();
+            Dimension buttonSize = button.getSize();
```
aivanov-jdk, line 87: Both `getLocationOnScreen` and `getSize` should also be called on EDT?

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest.java
```diff
@@ -0,0 +1,189 @@
...
+import java.awt.event.InputEvent;
+import java.awt.event.KeyEvent;
+
+public class ScrollPaneWindowsTest implements AdjustmentListener {
+    ScrollPane sp;
+    Panel p;
+    Robot robot;
+    Frame frame;
+    Insets paneInsets;
+    public static Object LOCK = new Object();
```
aivanov-jdk, line 53: The lock object should be `final`.

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest.java
```diff
@@ -0,0 +1,189 @@
...
+    ScrollPaneAdjustable vScroll;
+    ScrollPaneAdjustable hScroll;
+    boolean notifyReceived = false;
+
+    public static void main(String[] args) throws Exception {
+        if (Double.valueOf(System.getProperty("os.version")) < 5.0) {
+            System.out.println("This test is for Windows 2000 and above.");
+            return;
+        }
```
aivanov-jdk, line 62: Shall this be dropped too? Java won't start on Windows versions before Vista (6.0), so does it serve a purpose other than a historical reasons?

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest.java
```diff
@@ -0,0 +1,189 @@
...
+            p = new Panel();
+            p.setLayout(null);
+            p.setSize(new Dimension(800, 800));
+        });
+        robot = new Robot();
+        sp = new ScrollPane(ScrollPane.SCROLLBARS_ALWAYS);
+        vScroll = (ScrollPaneAdjustable) sp.getVAdjustable();
+        hScroll = (ScrollPaneAdjustable) sp.getHAdjustable();
+        vScroll.addAdjustmentListener(this);
+        hScroll.addAdjustmentListener(this);
```
aivanov-jdk, line 81: ScrollPane should be created inside `invokeAndWait` block, listeners should be added on EDT too.

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest.java
```diff
@@ -0,0 +1,189 @@
...
+            return;
+        }
+        ScrollPaneWindowsTest scrollTest = new ScrollPaneWindowsTest();
+        scrollTest.init();
+        scrollTest.start();
+    }
+
+    public void init() throws Exception {
+        EventQueue.invokeAndWait(() -> {
+            frame = new Frame();
```
aivanov-jdk, line 70:
```suggestion
            frame = new Frame("ScrollPaneWindowsTest");
```

test/jdk/java/awt/ScrollPane/ScrollPositionIntact.java
```diff
@@ -0,0 +1,95 @@
...
+ */
+/*
+ @test
+ @bug 6404832
+ @summary Tests that scroll position is not changed by validate() for mode SCROLLBARS_NEVER
+ @key headful
+ @run main ScrollPositionIntact
+*/
+
+import java.awt.*;
```
aivanov-jdk, line 31: Avoid wildcard imports.

test/jdk/java/awt/ScrollPane/ScrollPositionIntact.java
```diff
@@ -0,0 +1,95 @@
...
+                pa.setPreferredSize(new Dimension(600, 50));
+                pa.setBackground(Color.red);
+                sp = new ScrollPane(ScrollPane.SCROLLBARS_NEVER);
+                sp.setSize(200, 50);
+                pa.setLayout(new GridLayout(1, 3));
+                pa.add("West", new Label("west", Label.LEFT));
+                pa.add("West", new Label());
+                pa.add("East", new Label("East", Label.RIGHT));
+                sp.add(pa);
+                frame = new Frame();
```
aivanov-jdk, line 61:
```suggestion
                frame = new Frame("ScrollPositionIntact");
```

test/jdk/java/awt/ScrollPane/ScrollPaneExtraScrollBar.java
```diff
@@ -49,44 +50,41 @@ public static void main(String[] args) throws Exception {
...
+        robot = new Robot();
         EventQueue.invokeAndWait(() -> {
-                f = new Frame();
-                sp = new ScrollPane(ScrollPane.SCROLLBARS_AS_NEEDED);
-                sp.add(new Button("TEST"));
-                f.add("Center", sp);
-                f.pack();
-                f.setLocationRelativeTo(null);
-                f.setVisible(true);
+            f = new Frame();
```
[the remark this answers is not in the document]
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneExtraScrollBar.java
```diff
@@ -49,44 +50,41 @@ public static void main(String[] args) throws Exception {
...
-                robot.delay(100);
-                robot.waitForIdle();
-                Rectangle r = f.getBounds();
-                robot.mouseMove(r.x + r.width - 1, r.y + r.height - 1);
-                robot.mousePress(InputEvent.BUTTON1_DOWN_MASK);
-                robot.mouseMove(r.x + r.width + 50, r.y + r.height + 50);
-                robot.mouseRelease(InputEvent.BUTTON1_DOWN_MASK);
+            robot.delay(100);
+            robot.waitForIdle();
+            Rectangle r = f.getBounds();
```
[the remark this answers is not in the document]
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneExtraScrollBar.java
```diff
@@ -49,44 +50,41 @@ public static void main(String[] args) throws Exception {
...
+            robot.mouseMove(r.x + r.width - 1, r.y + r.height - 1);
+            robot.mousePress(InputEvent.BUTTON1_DOWN_MASK);
+            robot.mouseMove(r.x + r.width + 50, r.y + r.height + 50);
+            robot.mouseRelease(InputEvent.BUTTON1_DOWN_MASK);
 
-                Thread.sleep(1000);
+            Thread.sleep(1000);
```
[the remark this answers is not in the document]
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation.java
TejeshR13: Yeah, we can drop it if its not supported.

test/jdk/java/awt/ScrollPane/ScrollPaneRemoveAdd.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneRemoveAdd.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneRemoveAdd.java
TejeshR13: I don't think its required to be volatile, not sure.

test/jdk/java/awt/ScrollPane/ScrollPaneExtraScrollBar.java
TejeshR13: But I have been doing without leading asterisk for all the open sourcing, let me keep it this way.

test/jdk/java/awt/ScrollPane/ComponentScrollTest.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ComponentScrollTest.java
TejeshR13: Yeah, I guess 5000 might not be required 1000 might be fine, but either delay or sleep is required I feel.

test/jdk/java/awt/ScrollPane/ScrollPaneRemoveAdd.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPositionIntact.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPositionIntact.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ComponentScrollTest.java
prrace: It seems just wrong to me. This sleep should be on the main thread after returning from invokeAndWait

TejeshR13: Ok, moved it outside invokeAndWait, works fine.

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest.java
```diff
@@ -0,0 +1,183 @@
...
+                    sp.getLocationOnScreen().y + sp.getHeight() / 2);
+            testOneScrollbar(vScroll);
+
+            robot.delay(100);
+            robot.waitForIdle();
+
+            robot.mouseMove(sp.getLocationOnScreen().x + sp.getWidth() / 2,
+                    sp.getLocationOnScreen().y + sp.getHeight() - paneInsets.bottom / 2);
+            testOneScrollbar(hScroll);
+        } finally{
```
turbanoff, line 106:
```suggestion
        } finally {
```

test/jdk/java/awt/ScrollPane/ScrollPositionIntact.java
```diff
@@ -0,0 +1,99 @@
...
+                frame.requestFocus();
+
+                sp.setScrollPosition(X_POS, sp.getScrollPosition().y);
+                pa.validate();
+                // Now, before the fix, in Windows XP, Windows XP theme and on Vista,
+                // scrollposition would be reset to zero..
+                sp.validate();
+
+                int i;
+                i = (int) (sp.getScrollPosition().getX());
```
aivanov-jdk, line 85:
```suggestion
                int i = (int) (sp.getScrollPosition().getX());
```

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation.java
```diff
@@ -0,0 +1,163 @@
...
+import java.awt.event.MouseEvent;
+import java.awt.event.MouseAdapter;
+
+import java.util.Properties;
+
+public class ScrollPaneLimitation {
+    final static int SCROLL_POS = 50000;
+    public static boolean mouseWasPressed = false;
+    public static Component child = null;
+    private Object lock = new Object();
```
aivanov-jdk, line 53:
```suggestion
    private final Object lock = new Object();
```

test/jdk/java/awt/ScrollPane/ScrollPaneRemoveAdd.java
aivanov-jdk: They shouldn't be volatile, a proper synchronisation is used. But `getState` should use `lock` lock to return the state because access to `state` is synchronised using `lock`, not `this`.

test/jdk/java/awt/ScrollPane/ScrollPaneRemoveAdd.java
```diff
@@ -0,0 +1,168 @@
...
+                if (state) {
+                    return;
+                }
+                waiting++;
+                synchronized (this) {
+                    wait();
+                }
+                waiting--;
+            }
+        }
```
aivanov-jdk, line 138: This method is never used, it can safely be dropped.

TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPositionIntact.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest.java
```diff
@@ -0,0 +1,183 @@
...
+            p = new Panel();
+            p.setLayout(null);
+            p.setSize(new Dimension(800, 800));
+            sp = new ScrollPane(ScrollPane.SCROLLBARS_ALWAYS);
+        });
+        robot = new Robot();
+        vScroll = (ScrollPaneAdjustable) sp.getVAdjustable();
+        hScroll = (ScrollPaneAdjustable) sp.getHAdjustable();
+        vScroll.addAdjustmentListener(this);
+        hScroll.addAdjustmentListener(this);
```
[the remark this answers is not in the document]
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation.java
```diff
@@ -72,87 +71,81 @@ public void init() throws Exception {
...
-                                    }
+            EventQueue.invokeAndWait(() -> {
+                pane = new ScrollPane();
+                frame.add(pane);
+                child = new MyPanel();
+                child.addMouseListener(new MouseAdapter() {
+                    public void mousePressed(MouseEvent e) {
+                        if (e.getID() == MouseEvent.MOUSE_PRESSED) {
+                            if (e.getSource() == ScrollPaneLimitation.child
+                                    && e.getY() > SCROLL_POS) {
```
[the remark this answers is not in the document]
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation.java
```diff
@@ -72,87 +71,81 @@ public void init() throws Exception {
...
+            robot.mousePress(InputEvent.BUTTON1_DOWN_MASK);
+            robot.mouseRelease(InputEvent.BUTTON1_DOWN_MASK);
+            robot.waitForIdle();
+            synchronized (lock) {
+                try {
+                    lock.wait(3000);
+                } catch (InterruptedException e) {
+                    throw new RuntimeException("test was interrupted");
                 }
             }
```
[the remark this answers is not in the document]
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ComponentScrollTest.java
```diff
@@ -41,48 +41,55 @@
 
 import java.lang.reflect.InvocationTargetException;
 
-public class ComponentScrollTest implements AdjustmentListener {
+public class ComponentScrollTest {
     public ScrollPane scrollpane;
     public Frame frame;
     public int count = 0;
```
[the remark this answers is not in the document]
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ComponentScrollTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+    public void start() throws Exception {
+        try {
+            EventQueue.invokeAndWait(() -> {
+                frame.pack();
+                frame.setVisible(true);
+                scrollpane.getVAdjustable().setValue(20);
+            });
+
+            Thread.sleep(5000);
```
aivanov-jdk, line 87: Sleeping for 5 seconds could be too much.

It would be great to ensure the updated test reproduces the problem that was fixed by [JDK-4342129](https://bugs.openjdk.org/browse/JDK-4342129) in 1.3.0 and 1.4.0. Otherwise, these 5 seconds could be a waste of time.

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest.java
```diff
@@ -0,0 +1,184 @@
...
+                frame.setVisible(true);
+                paneInsets = sp.getInsets();
+                System.out.println("Insets: right = " + paneInsets.right + " bottom =  " + paneInsets.bottom);
+            });
+
+            robot.delay(100);
+            robot.waitForIdle();
+
+            robot.mouseMove(sp.getLocationOnScreen().x + sp.getWidth() - paneInsets.right / 2,
+                    sp.getLocationOnScreen().y + sp.getHeight() / 2);
```
aivanov-jdk, line 98: You should get the location on the scroll pane and its size inside `invokeAndWait` and use these two new fields to move the mouse.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation.java
```diff
@@ -0,0 +1,159 @@
...
+import java.awt.event.InputEvent;
+import java.awt.event.MouseEvent;
+import java.awt.event.MouseAdapter;
+
+import java.util.Properties;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+public class ScrollPaneLimitation {
+    final static int SCROLL_POS = 50000;
```
aivanov-jdk, line 52:
```suggestion
    static final int SCROLL_POS = 50000;
```
Address the missorted modifiers.

test/jdk/java/awt/ScrollPane/ScrollPositionIntact.java
```diff
@@ -0,0 +1,98 @@
...
+                sp.add(pa);
+                frame = new Frame("ScrollPositionIntact");
+                frame.setSize(200, 100);
+                frame.add(sp);
+                frame.setVisible(true);
+            });
+            Robot robot = new Robot();
+            robot.delay(1000);
+            robot.waitForIdle();
+            EventQueue.invokeAndWait(() -> {
```
aivanov-jdk, line 74:
```suggestion
            });

            Robot robot = new Robot();
            robot.delay(1000);
            robot.waitForIdle();

            EventQueue.invokeAndWait(() -> {
```
Let's add blank lines to separate the code portions visually.

aivanov-jdk (changes requested):

test/jdk/java/awt/ScrollPane/ComponentScrollTest.java
TejeshR13: I guess 1 second will be just fine. Updating it to 1 sec.

test/jdk/java/awt/ScrollPane/ComponentScrollTest.java
```diff
@@ -0,0 +1,99 @@
...
+                }
+            });
+        });
+    }
+
+    public void start() throws Exception {
+        try {
+            EventQueue.invokeAndWait(() -> {
+                frame.pack();
+                frame.setVisible(true);
```
alisenchung, line 81: should frame setup code be moved into init()?

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation.java
```diff
@@ -0,0 +1,154 @@
...
+                        go.countDown();
+                    }
+                    }
+                });
+                pane.add(child);
+                frame.setSize(200, 200);
+                frame.pack();
+                frame.setLocationRelativeTo(null);
+                frame.setVisible(true);
+                pane.doLayout();
```
alisenchung, line 94: frame setup moved to init() here too?

test/jdk/java/awt/ScrollPane/ScrollPaneRemoveAdd.java
```diff
@@ -0,0 +1,150 @@
...
+    public void start() throws Exception {
+        try {
+            EventQueue.invokeAndWait(() -> {
+                frame.pack();
+                frame.setLocationRelativeTo(null);
+                frame.setVisible(true);
+                pane.remove(0);
+                pane.add(button);
+                buttonLoc = button.getLocationOnScreen();
+                buttonSize = button.getSize();
```
alisenchung, line 88: here too

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest.java
```diff
@@ -0,0 +1,189 @@
...
+    public void start() throws Exception {
+        try {
+            EventQueue.invokeAndWait(() -> {
+                sp.add(p);
+                frame.add(sp);
+                frame.pack();
+                frame.setSize(400, 400);
+                frame.setLocationRelativeTo(null);
+                frame.setVisible(true);
+                paneInsets = sp.getInsets();
```
alisenchung, line 92: here too

test/jdk/java/awt/ScrollPane/ComponentScrollTest.java
aivanov-jdk: Does it matter?

TejeshR13: Functionally doesn't matter, I guess its for code readability and uniformity.

aivanov-jdk: Yes, I agree it's better to combine them but then you could go further and remove `start()` and `init()` altogether.

It would be good to be consistent across test groups especially those handled by one person. Yet I admit I didn't pay much attention to this thing.

TejeshR13: I have updated all the test for consistency.

TejeshR13: @aivanov-jdk Can you please re-review the PR and see if its fine. I have updated all the review comments.

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest.java
```diff
@@ -0,0 +1,190 @@
...
+
+    public void testOneScrollbar(ScrollPaneAdjustable scroll) throws Exception {
+        //to Bottom  - right
+        robot.mousePress(InputEvent.BUTTON3_DOWN_MASK);
+        robot.mouseRelease(InputEvent.BUTTON3_DOWN_MASK);
+        robot.delay(2000);
+        robot.waitForIdle();
+
+        notifyReceived = false;
+        synchronized (LOCK) {
```
aivanov-jdk, line 135:
```suggestion
        synchronized (LOCK) {
            notifyReceived = false;
```
The `notifyReceived` must always be read and written inside a synchronised block. Other places should also be corrected.

test/jdk/java/awt/ScrollPane/ScrollPositionIntact.java
```diff
@@ -0,0 +1,104 @@
...
+                pa.add("West", new Label());
+                pa.add("East", new Label("East", Label.RIGHT));
+                sp.add(pa);
+                frame = new Frame("ScrollPositionIntact");
+                frame.setSize(200, 100);
+                frame.add(sp);
+                frame.setVisible(true);
+            });
+        }
+    }
```
aivanov-jdk, line 72: It does not even compile.

Add a blank line before `init` and remove `try`-block to make it compilable.

aivanov-jdk (changes requested):

test/jdk/java/awt/ScrollPane/ComponentScrollTest.java
```diff
@@ -71,14 +71,14 @@ public void adjustmentValueChanged(AdjustmentEvent adjustmentEvent) {
...
+            frame.setVisible(true);
         });
     }
 
     public void start() throws Exception {
         try {
             EventQueue.invokeAndWait(() -> {
-                frame.pack();
-                frame.setVisible(true);
                 scrollpane.getVAdjustable().setValue(20);
```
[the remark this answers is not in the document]
TejeshR13: Looks like its old code.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation.java
```diff
@@ -67,33 +67,31 @@ public void init() throws Exception {
...
+            frame.setVisible(true);
+            pane.doLayout();
+        });
+        robot = new Robot();
+    }
 
+    public void start() throws Exception {
+        try {
             robot.delay(1000);
             robot.waitForIdle();
```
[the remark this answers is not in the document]
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation.java
```diff
@@ -0,0 +1,152 @@
...
+import java.awt.Frame;
+import java.awt.Point;
+import java.awt.Robot;
+import java.awt.ScrollPane;
+
+import java.awt.event.InputEvent;
+import java.awt.event.MouseEvent;
+import java.awt.event.MouseAdapter;
+
+import java.util.Properties;
```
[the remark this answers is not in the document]
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation.java
```diff
@@ -0,0 +1,152 @@
...
+                        }
+                    } else {
+                        throw new RuntimeException("wrong child location");
+                    }
+                }
+
+                point = pane.getLocationOnScreen();
+                Dimension d = pane.getSize();
+                point.x += d.width / 2;
+                point.y += d.height / 2;
```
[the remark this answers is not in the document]
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation.java
```diff
@@ -0,0 +1,152 @@
...
+            });
+            robot.mouseMove(point.x, point.y);
+            robot.mousePress(InputEvent.BUTTON1_DOWN_MASK);
+            robot.mouseRelease(InputEvent.BUTTON1_DOWN_MASK);
+            robot.waitForIdle();
```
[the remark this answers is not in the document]
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation.java
```diff
@@ -0,0 +1,152 @@
...
+import java.awt.event.MouseAdapter;
+
+import java.util.Properties;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+public class ScrollPaneLimitation {
+    static final int SCROLL_POS = 50000;
+    public static Component child = null;
+    static CountDownLatch go = new CountDownLatch(1);
```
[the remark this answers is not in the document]
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneRemoveAdd.java
```diff
@@ -0,0 +1,150 @@
...
+            });
+
+            robot.delay(1000);
+            robot.waitForIdle();
+            robot.mouseMove(buttonLoc.x + buttonSize.width / 2,
+                    buttonLoc.y + buttonSize.height / 2);
```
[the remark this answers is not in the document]
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest.java
```diff
@@ -0,0 +1,190 @@
...
+            frame.pack();
+            frame.setSize(400, 400);
+            frame.setLocationRelativeTo(null);
+            frame.setAlwaysOnTop(true);
+            frame.setVisible(true);
+        });
+        robot = new Robot();
+
+        vScroll.addAdjustmentListener(this);
+        hScroll.addAdjustmentListener(this);
```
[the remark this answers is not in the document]
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest.java
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneRemoveAdd.java
```diff
@@ -0,0 +1,150 @@
...
+            robot.delay(50);
+            robot.mousePress(InputEvent.BUTTON1_DOWN_MASK);
+            robot.delay(50);
+            robot.mouseRelease(InputEvent.BUTTON1_DOWN_MASK);
+
+            robot.delay(50);
```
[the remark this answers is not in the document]
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPositionIntact.java
TejeshR13: My bad, I updated the wrong commit for this particular file. Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneRemoveAdd.java
```diff
@@ -0,0 +1,150 @@
...
+        } finally {
+            EventQueue.invokeAndWait(() -> {
+                if (frame != null) {
+                    frame.dispose();
+                }
+            });
+        }
+    }
+
+    private static class Semaphore {
```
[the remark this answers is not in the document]
TejeshR13: Removed.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation.java
```diff
@@ -67,33 +67,31 @@ public void init() throws Exception {
...
             });
+            pane.add(child);
+            frame.setSize(200, 200);
+            frame.pack();
+            frame.setLocationRelativeTo(null);
+            frame.setAlwaysOnTop(true);
+            frame.setVisible(true);
+            pane.doLayout();
+        });
+        robot = new Robot();
```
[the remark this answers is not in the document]
TejeshR13: Updated.

test/jdk/java/awt/ScrollPane/ScrollPaneExtraScrollBar.java
```diff
@@ -0,0 +1,98 @@
...
+            f.pack();
+            f.setLocationRelativeTo(null);
+            f.setVisible(true);
+        });
+    }
+
+    public void start() throws Exception {
+        try {
+            robot.delay(100);
+            robot.waitForIdle();
```
aivanov-jdk, line 71: Optionally,
```suggestion
            robot = new Robot();
            robot.waitForIdle();
            robot.delay(100);
```
To be consistent with other tests in this review.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation.java
```diff
@@ -0,0 +1,148 @@
...
+                p = child.getLocation();
+                System.out.println("Child pos = " + p);
+                if (p.y != -SCROLL_POS) {
+                    if (child.isLightweight()) {
+                        // If it is lightweight it will always have (0, 0) location.
+                        // Check location of its parent - it is Panel and it should
+                        // be at (inset left, inset top + position)
+                        Container cp = child.getParent();
+                        p = cp.getLocation();
+                        if (p.y != -SCROLL_POS) {
```
aivanov-jdk, line 115:
```suggestion
                        p = cp.getLocation();
                        System.out.println("Child's parent pos = " + p);
                        if (p.y != -SCROLL_POS) {
```
Just to make it clear that this branch is taken. Otherwise the test ends with the message:
```text
Child pos = java.awt.Point[x=0,y=0]
```
which is confusing. Yet no exception is thrown.

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+    public void start() throws Exception {
+        try {
+            EventQueue.invokeAndWait(() -> {
+                paneInsets = sp.getInsets();
+                System.out.println("Insets: right = " + paneInsets.right + " bottom =  " + paneInsets.bottom);
+            });
+
+            robot.delay(100);
+            robot.waitForIdle();
```
aivanov-jdk, line 115:
```suggestion
            robot = new Robot();
            robot.waitForIdle();
            robot.delay(100);

```

aivanov-jdk (changes requested):

test/jdk/java/awt/ScrollPane/ScrollPaneRemoveAdd.java
```diff
@@ -74,30 +73,31 @@ public void actionPerformed(ActionEvent e) {
             });
             frame.pack();
             frame.setLocationRelativeTo(null);
+            frame.setAlwaysOnTop(true);
```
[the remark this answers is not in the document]
TejeshR13: When I run the folder, due to previous test/command prompt the frame used to show up behind command prompt, so to be on safer side I used always on top. When we run individual test, it works fine.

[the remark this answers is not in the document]
aivanov-jdk: It may still not work… You have to disable keyboard input timeout by setting `ForegroundLockTimeout` to zero as explained in [requirements for Windows](https://wiki.openjdk.org/display/ClientLibs/Automated+client+GUI+testing+system+set+up+requirements#AutomatedclientGUItestingsystemsetuprequirements-Windows-SpecificSystemsetupnotes).

Most likely, you don't want to do it on the system that you regularly use, otherwise apps could steal focus.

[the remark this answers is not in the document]
TejeshR13: Its working fine after setting to always on top, tested multiple times.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation.java
TejeshR13: Child location is expected to be 0,0 as explained in the test, whereas parents position is -500, hence no exception is thrown.

test/jdk/java/awt/ScrollPane/ScrollPaneWindowsTest.java
```diff
@@ -73,6 +73,26 @@ public void init() throws Exception {
...
+            vScroll.addAdjustmentListener(new AdjustmentListener() {
+                @Override
+                public void adjustmentValueChanged(AdjustmentEvent adjustmentEvent) {
+                    synchronized (ScrollPaneWindowsTest.LOCK) {
+                        notifyReceived = true;
+                        ScrollPaneWindowsTest.LOCK.notify();
+                    }
+                    System.out.println("Adjustment Vertical Scroll Event called ");
+                }
+            });
```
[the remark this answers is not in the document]
TejeshR13: Updated and tested.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation.java
aivanov-jdk: Yes, this is adding the suggested print will resolve all the confusion because you will see:

```text
...
Scroll pos = java.awt.Point[x=0,y=50000]
Child pos = java.awt.Point[x=0,y=0]
Child's parent pos = java.awt.Point[x=0,y=-50000]
```

Isn't it clearer?

aivanov-jdk (changes requested):
I'm for resolving [this comment](https://github.com/openjdk/jdk/pull/13621#discussion_r1200559746) to make the output clearer.

test/jdk/java/awt/ScrollPane/ScrollPaneLimitation.java
TejeshR13: Yeah, can do it.

aivanov-jdk: Thank you!

aivanov-jdk (approved):

aivanov-jdk (approved):

test/jdk/java/awt/ScrollPane/ScrollPositionIntact.java
```diff
@@ -0,0 +1,103 @@
...
+
+                sp.setScrollPosition(X_POS, sp.getScrollPosition().y);
+                pa.validate();
+                // Now, before the fix, in Windows XP, Windows XP theme and on Vista,
+                // scrollposition would be reset to zero..
+                sp.validate();
+
+                int i = (int) (sp.getScrollPosition().getX());
+                if (i == 0) {
+                    // actual position MAY be not equal to X_POS; still, it must be > 0.
```
mrserb, line 91: If the valid value is "> 0" only, then we should throw an exception if the value is "<=0"

TejeshR13: Yes, right. Updated the test.

PR #13629, opened 2023-04-24
8289735: UTIL_LOOKUP_PROGS fails on pathes with space

RealCLanger:
This is an attempt to fix the issue on Windows when no cygwin Git is installed or the Git for Windows installation has precedence in PATH lookup.
The path to the Windows GIT installation usually resides in `C:\Program Files` which contains a space and thus needs some special handling.

There exists code in `UTIL_LOOKUP_PROGS`/`UTIL_FIXUP_EXECUTABLE` that would handle this. However, it relies on initializations made in `PLATFORM_SETUP_OPENJDK_BUILD_AND_TARGET` and `BASIC_SETUP_PATHS`.
Currently, `UTIL_LOOKUP_PROGS(GIT, git)` is called too early in configure and hence the problematic Windows path is not handled correctly.
My fix makes sure that only the barely necessary tool lookups required for `PLATFORM_SETUP_OPENJDK_BUILD_AND_TARGET` and `BASIC_SETUP_PATHS` are made in `BASIC_SETUP_FUNDAMENTAL_TOOLS` and everything else is moved into another macro called `BASIC_SETUP_TOOLS` that is invoked after path handling is set up correctly.

### DIFF
--- a/make/autoconf/basic_tools.m4
+++ b/make/autoconf/basic_tools.m4
@@ -29,8 +29,8 @@
 RECOMMENDED_PANDOC_VERSION=2.19.2
 
 ###############################################################################
-# Setup the most fundamental tools that relies on not much else to set up,
-# but is used by much of the early bootstrap code.
+# Setup the most fundamental tools which rely on not much else to set up,
+# but are used by much of the early bootstrap code.
 AC_DEFUN_ONCE([BASIC_SETUP_FUNDAMENTAL_TOOLS],
 [
   # Bootstrapping: These tools are needed by UTIL_LOOKUP_PROGS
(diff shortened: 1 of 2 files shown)


erikj79:
I think this change makes sense. Given how small this set of tools is now, I think we should add a comment on `PLATFORM_SETUP_OPENJDK_BUILD_AND_TARGET` in platform.m4 saying something about needing to be careful to only use tools setup in `BASIC_SETUP_FUNDAMENTAL_TOOLS`.

RealCLanger:
> BASIC_SETUP_PATHS

Right, and the same comment for BASIC_SETUP_PATHS. Will update the comments...

erikj79 (approved):

RealCLanger:
Thanks for the reviews.

/integrate

PR #13632, opened 2023-04-25
8306623: (bf) CharBuffer::allocate throws unexpected exception type with some CharSequences

bplb:
When appending a `CharBuffer`, perform a specific check for buffer overflow before invoking `put()`.

---------
### Progress

### Issue
 * [JDK-8306623](https://bugs.openjdk.org/browse/JDK-8306623): (bf) CharBuffer::allocate throws unexpected exception type with some CharSequences

### Reviewers
 * [Alan Bateman](https://openjdk.org/census#alanb) (@AlanBateman - **Reviewer**) ⚠️ Review applies to [118076c9](https://git.openjdk.org/jdk/pull/13632/files/118076c94ae74d2a4967b8b6ec12e83365f196c6)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13632/head:pull/13632` \
`$ git checkout pull/13632`

`$ git checkout pull/13632` \
`$ git pull https://git.openjdk.org/jdk.git pull/13632/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13632`

`$ git pr show -t 13632`

</details>

<a href="https://git.openjdk.org/jdk/pull/13632.diff">https://git.openjdk.org/jdk/pull/13632.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13632#issuecomment-1521042262)

### DIFF
--- a/src/java.base/share/classes/java/nio/X-Buffer.java.template
+++ b/src/java.base/share/classes/java/nio/X-Buffer.java.template
@@ -2045,8 +2045,17 @@ public abstract sealed class $Type$Buffer
      */
     public $Type$Buffer append(CharSequence csq, int start, int end) {
         if (csq instanceof CharBuffer cb) {
-            int pos = position();
+            //
+            // pre-emptively check for overflow as put(int,CharBuffer,int,int)
+            // would throw IndexOutOfBoundsException instead
+            //
             int length = end - start;
+            int pos = position();
+            int lim = limit();
+            int rem = (pos <= lim) ? lim - pos : 0;
+            if (length > rem)
+                throw new BufferOverflowException();
+
             put(pos, cb, start, length);
             position(pos + length);
             return this;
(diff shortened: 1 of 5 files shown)


bplb:
I'm not entirely comfortable with where I've put the test in this patch.

AlanBateman:
Can you add a test to test/jdk/java/nio/Buffer/Basic-X.java.template, I think this regression was missed because this tests didn't check this case.

src/java.base/share/classes/java/nio/X-Buffer.java.template
```diff
@@ -2045,8 +2045,17 @@ public abstract sealed class $Type$Buffer
      */
     public $Type$Buffer append(CharSequence csq, int start, int end) {
         if (csq instanceof CharBuffer cb) {
-            int pos = position();
+            //
+            // pre-emptively check for overflow as put(int,CharBuffer,int,int)
+            // would throw IndexOutOfBoundsException instead
```
AlanBateman, line 2050: The comment is confusing, I think you just want to say that the append method throws BufferOverflowException when there is insufficient space in the buffer.

bplb: Comment updated and test code moved in 118076c.

bplb:
> [...] I think this regression was missed because this tests didn't check this case.

It definitely was.

AlanBateman (approved):

LanceAndersen (approved):

PR #13634, opened 2023-04-25
8306774: Make runtime/Monitor/GuaranteedAsyncDeflationIntervalTest.java more reliable

shipilev:
As seen in some CI runs, doing `Object.wait(1)` to inflate 10K monitors can take quite a while. One of the test cases implicitly relies on threshold heuristics kicking in before the guaranteed interval of 5 seconds expires. Even if we wait for 1ms for each `wait` call, we only inflate 5K monitors in those 5 seconds. Which is okay for 5s and 10% threshold test, which would fire after about 1K monitors inflated. But we can also stall for longer -- because `wait(1)` is not guaranteed to return in 1ms, subject to locking granularity -- which means we can inflate less that 1K monitors, which would mean the threshold heuristics would not fire.

This fix strengthens the test in two ways: 
  1. Uses multiple threads to inflate the monitors, so that we are able to eat the cost of `wait(1)`.
  2. Drops the threshold heuristics threshold to minimum value to increase the chances of it firing.

This also makes the test faster.

@dcubed-ojdk, you might want to run this through your CI to see if it fixes the test troubles.

Additional testing:
 - [x] Ad-hoc runs of the test, 100 iterations without trouble.

---------
### Progress

### Issue
 * [JDK-8306774](https://bugs.openjdk.org/browse/JDK-8306774): Make runtime/Monitor/GuaranteedAsyncDeflationIntervalTest.java more reliable

### Reviewers
 * [Thomas Stuefe](https://openjdk.org/census#stuefe) (@tstuefe - **Reviewer**) ⚠️ Review applies to [76fa7114](https://git.openjdk.org/jdk/pull/13634/files/76fa71147bba652ed0bf1122d15e8597b4fe3f93)
 * [Daniel D. Daugherty](https://openjdk.org/census#dcubed) (@dcubed-ojdk - **Reviewer**) ⚠️ Review applies to [76fa7114](https://git.openjdk.org/jdk/pull/13634/files/76fa71147bba652ed0bf1122d15e8597b4fe3f93)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13634/head:pull/13634` \
`$ git checkout pull/13634`

`$ git checkout pull/13634` \
`$ git pull https://git.openjdk.org/jdk.git pull/13634/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13634`

`$ git pr show -t 13634`

</details>

<a href="https://git.openjdk.org/jdk/pull/13634.diff">https://git.openjdk.org/jdk/pull/13634.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13634#issuecomment-1521440819)


tstuefe (approved):
LGTM.  Why did you steer away from the i-hash solution though, I thought that was rather neat.

shipilev:
> LGTM. Why did you steer away from the i-hash solution though, I thought that was rather neat.

Previous version of this PR used the take-IHC-while-locked trick to inflate the monitors, but I figured it would not work after [JDK-8291555: Implement alternative fast-locking scheme](https://bugs.openjdk.org/browse/JDK-8291555) is here, and would quite probably be even less reliable as Lilliput progresses towards off-side IHC table.

test/hotspot/jtreg/runtime/Monitor/GuaranteedAsyncDeflationIntervalTest.java
```diff
@@ -58,21 +58,41 @@ public class GuaranteedAsyncDeflationIntervalTest {
...
 
-        public static Object[] monitors;
+        // Use a handful of threads to inflate the monitors, to eat the cost of
+        // wait(1) calls. This can be larger than available parallelism, since threads
+        // would be time-waiting.
+        private static final int THREADS = 16;
+
+        private static Thread[] threads;
+        private static Object[] monitors;
+        private static long sink;
```
dcubed-ojdk, line 70: Appears to be unused...

dcubed-ojdk (approved):
Looks like a reasonable fix so thumbs up.
I'm testing in Mach5 now.

dcubed-ojdk:
Mach5 Tier1 testing passed. It doesn't look like the test is
executed in Tier[23].

test/hotspot/jtreg/runtime/Monitor/GuaranteedAsyncDeflationIntervalTest.java
shipilev: Good spot, removed.

shipilev:
> > Previous version of this PR used the take-IHC-while-locked trick to inflate the monitors, but I figured it would not work after [JDK-8291555: Implement alternative fast-locking scheme](https://bugs.openjdk.org/browse/JDK-8291555) is here, and would quite probably be even less reliable as Lilliput progresses towards off-side IHC table.
> 
> Hmmm... I don't remember seeing anything in JDK-8291555 that would break the take-IHC-while-locked trick. I use that trick in some of my monitor stress tests... It would be sad to see that go away...

Perhaps I was too cautious about this. I figured that the practical reason to inflate the stack-locked object when IHC is requested is to deal with various races about the IHC installation in the displaced header. And since JDK-8291555 changes the stack-locking to not replace the header fully (only changing a bit, AFAIU), then IHC installation becomes much easier. But after reading the description again, I think JDK-8291555 only takes care of IHC on fastpath, i.e. when it is already installed. So take-IHC-while-locked might still work.

Still, I don't feel quite safe with relying on it, because it seems (famous last words) an easy future change to allow IHC installation into stack-locked object header after JDK-8291555, and I think there is a tentative plan to move IHC out of the header completely with 32-bit headers, which would make inflation even less of the practical behavior. So I believe `wait(1)` remains the only future-proof way to inflate the monitors, before we start to mess with JVM settings.

shipilev:
> Mach5 Tier1 testing passed. It doesn't look like the test is executed in Tier[23].

Thank you! I'll re-run some local testing and then integrate.

shipilev:
Local testing looks good. GHAs look good. I am integrating now.

/integrate

PR #13643, opened 2023-04-25
8306836: Remove pinned tag for G1 heap regions

tschatzl:
Hi all,

  please review this change that removes the pinned tag from `HeapRegion`.

So that "pinned" tag for G1 heap regions indicates that the region should not move during (young) gc. This applies to now removed archive regions and humongous objects/regions.

With "real" g1 region pinning to deal with gclocker in g1 once and for all upcoming we need a refcount, a single bit is not sufficient anymore. Further there will be a naming conflict as this kind of "pinning" is different to g1 region pinning "pinning". The former indicates "contents can not be moved, but can be reclaimed", while the latter means "contents can not be moved and not reclaimed".

The (current) pinned flag is surprisingly little used, only for policy decisions.

The suggestion this change implements is to remove the "pinned" tag as it is, and reserve it for future g1 region pinning (that needs to store the pinning attribute differently as a refcount anyway).

Testing: tier1-3, gha

Thanks,
  Thomas

---------
### Progress

### Issue
 * [JDK-8306836](https://bugs.openjdk.org/browse/JDK-8306836): Remove pinned tag for G1 heap regions

### Reviewers
 * [Chris Plummer](https://openjdk.org/census#cjplummer) (@plummercj - **Reviewer**) ⚠️ Review applies to [eacf54ba](https://git.openjdk.org/jdk/pull/13643/files/eacf54baafe64c8e1b2feedb40379bc75867ab90)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13643/head:pull/13643` \
`$ git checkout pull/13643`

`$ git checkout pull/13643` \
`$ git pull https://git.openjdk.org/jdk.git pull/13643/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13643`

`$ git pr show -t 13643`

</details>

<a href="https://git.openjdk.org/jdk/pull/13643.diff">https://git.openjdk.org/jdk/pull/13643.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13643#issuecomment-1521836779)


src/hotspot/share/gc/g1/g1HeapVerifier.cpp
```diff
@@ -410,8 +410,8 @@ class VerifyRegionListsClosure : public HeapRegionClosure {
       _old_count++;
     } else {
       // There are no other valid region types. Check for one invalid
-      // one we can identify: pinned without old or humongous set.
-      assert(!hr->is_pinned(), "Heap region %u is pinned but not old or humongous.", hr->hrm_index());
+      // one we can identify before crashing: non-movable.
+      assert(hr->is_young_gc_movable(), "Heap region %u is non-movable.", hr->hrm_index());
```
albertnetymk, line 414: Given this branch is unreachable, can one just use `fatal(...)` without a predicate?

albertnetymk (commented):
The name, `is_young_gc_movable`, seems to suggest it is context sensitive. Maybe it's clearer if it's not defined as an API of heap-region, like other region-type-predicates.

For example, it's odd to see them on the same abstraction level, when the two APIs refer to sth quite different.

```
  assert(hr->is_young_gc_movable(), "Should only be movable region in compaction queue");
  assert(!hr->is_humongous(), "Should be no humongous regions in compaction queue");
```

Additionally, it's confusing to see sth named "*young_gc*" in full-gc code (`g1FullGCPrepareTask.inline.hpp`).

For this particular PR, could one just use `!hr->is_humongous()` where `is_young_gc_movable` is used? (Essentially, inlining the new API.)

src/hotspot/share/gc/g1/g1HeapVerifier.cpp
tschatzl: > Given this branch is unreachable, can one just use `fatal(...)` without a predicate?

I also thought about this: the purpose of the additional assert is/was (I think) to fail with a better message than just crashing with some nondescript message, filtering out other possible cases. 
Will remove this after all.

src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/HSDB.java
```diff
@@ -1104,9 +1104,6 @@ public void addAnnotation(Address addr, OopHandle handle) {
                           } else if (region.isHumongous()) {
                             anno = "Humongous ";
                             bad = false;
-                          } else if (region.isPinned()) {
-                            anno = "Pinned ";
-                            bad = false;
```
plummercj, line 1109: Does this mean that the region will now always be one of Free, Young, Humongous, or Old?

plummercj (commented):
A few files need copyright updates

plummercj (approved):
SA changes look good.

albertnetymk:
> I think you are right about using is_humongous() directly here: the reason we skip compacting of humongous regions during the "main" compaction is intentional here

However, I am unable to discern the difference -- why `is_young_gc_movable` is semantically-correct in one place, but not in the other in this concrete example.

```
bool G1CollectionSetChooser::should_add(HeapRegion* hr) {
  return !hr->is_young() &&
         G1CollectedHeap::heap()->policy()->is_young_gc_movable(hr) &&
```

vs

```
void G1FullCollector::before_marking_update_attribute_table(HeapRegion* hr) {
  if (hr->is_free()) {
    _region_attr_table.set_free(hr->hrm_index());
  } else if (hr->is_humongous()) {
```

Looking at where `G1CollectionSetChooser::should_add` is called, can one use `hr->is_old()` instead of `!hr->is_young() && G1CollectedHeap::heap()->policy()->is_young_gc_movable(hr)`? (In fact, I believe that inlining `should_add` to the caller would result in a smoother code flow and prevent some duplicate region-type queries.)

In my opinion, introducing a new `is_young_gc_movable` API in this particular PR may not be entirely justified. It may make more sense to introduce it in later PRs where region-pinning is supported and the API is actually utilized.

tschatzl:
I removed the `young_gc_is_movable()` predicate; it is probably the wrong time to introduce more abstract concepts like this in this change.

Moved off the refactoring of the `G1CollectionSetChooser::should_add()` and its caller to sometime else too - it's not relevant to this change either.

albertnetymk (approved):

kimbarrett (commented):
Sorry to be late to the review.  I noticed a problem in a comment.

PR #13645, opened 2023-04-25
8291550: RISC-V: jdk uses misaligned memory access when AvoidUnalignedAccess enabled

VladimirKempik:
Please review this attempt to remove misaligned loads and stores in risc-v specific part of jdk.

The patch has two main parts:
 - opcodes loads/stores is now using put_native_uX/get_native_uX
 - some code in template interp got changed to prevent misaligned loads
 
perf stat numbers for trp_lam ( misaligned loads) and trp_sam ( misaligned stores) before the patch: 
```
 169598      trp_lam                                          
  13562      trp_sam  
```

after the patch both numbers are zeroes.
I can see template interpreter to be ~40 % faster on hifive unmatched ( 1 repetition of renaissance philosophers in -Xint mode), and the same performance ( before and after the patch) on thead rvb-ice ( which supports misaligned stores/loads in hw)

tier testing on hw is in progress

---------
### Progress

### Issue
 * [JDK-8291550](https://bugs.openjdk.org/browse/JDK-8291550): RISC-V: jdk uses misaligned memory access when AvoidUnalignedAccess enabled

### Reviewers
 * [Feilong Jiang](https://openjdk.org/census#fjiang) (@feilongjiang - Author)
 * [Fei Yang](https://openjdk.org/census#fyang) (@RealFYang - **Reviewer**)

### Contributors
 * Xiaolin Zheng `<xlinzheng@openjdk.org>`
 * Feilong Jiang `<fjiang@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13645/head:pull/13645` \
`$ git checkout pull/13645`

`$ git checkout pull/13645` \
`$ git pull https://git.openjdk.org/jdk.git pull/13645/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13645`

`$ git pr show -t 13645`

</details>

<a href="https://git.openjdk.org/jdk/pull/13645.diff">https://git.openjdk.org/jdk/pull/13645.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13645#issuecomment-1522030387)


VladimirKempik:
failure in lunux-x86_32 unrelated ( TestJNICriticalStressTest )
I have only changed code in src/hotspot/cpu/riscv/ and nothing outside of it

src/hotspot/cpu/riscv/interp_masm_riscv.cpp
```diff
@@ -191,13 +198,38 @@ void InterpreterMacroAssembler::get_dispatch() {
...
 void InterpreterMacroAssembler::get_cache_index_at_bcp(Register index,
+                                                       Register tmp,
                                                        int bcp_offset,
                                                        size_t index_size) {
   assert(bcp_offset > 0, "bcp is still pointing to start of bytecode");
   if (index_size == sizeof(u2)) {
-    load_unsigned_short(index, Address(xbcp, bcp_offset));
+    if (AvoidUnalignedAccesses)
+    {
+      assert(index != tmp, "must use different register");
```
feilongjiang, line 208:
```suggestion
      assert_different_registers(index, tmp);
```

src/hotspot/cpu/riscv/interp_masm_riscv.cpp
```diff
@@ -191,13 +198,38 @@ void InterpreterMacroAssembler::get_dispatch() {
...
 
 void InterpreterMacroAssembler::get_cache_index_at_bcp(Register index,
+                                                       Register tmp,
                                                        int bcp_offset,
                                                        size_t index_size) {
   assert(bcp_offset > 0, "bcp is still pointing to start of bytecode");
   if (index_size == sizeof(u2)) {
-    load_unsigned_short(index, Address(xbcp, bcp_offset));
+    if (AvoidUnalignedAccesses)
+    {
```
feilongjiang, line 207:
```suggestion
    if (AvoidUnalignedAccesses) {
```

src/hotspot/cpu/riscv/interp_masm_riscv.cpp
```diff
@@ -191,13 +198,38 @@ void InterpreterMacroAssembler::get_dispatch() {
...
+      assert(index != tmp, "must use different register");
+      load_unsigned_byte(index, Address(xbcp, bcp_offset));
+      load_unsigned_byte(tmp, Address(xbcp, bcp_offset+1));
```
feilongjiang, line 210:
```suggestion
      load_unsigned_byte(tmp, Address(xbcp, bcp_offset + 1));
```

src/hotspot/cpu/riscv/interp_masm_riscv.cpp
```diff
@@ -191,13 +198,38 @@ void InterpreterMacroAssembler::get_dispatch() {
...
+    } else {
+      load_unsigned_short(index, Address(xbcp, bcp_offset));
+    }
   } else if (index_size == sizeof(u4)) {
-    lwu(index, Address(xbcp, bcp_offset));
+    if (AvoidUnalignedAccesses)
+    {
+      assert(index != tmp, "must use different register");
+      load_unsigned_byte(index, Address(xbcp, bcp_offset));
+      load_unsigned_byte(tmp, Address(xbcp, bcp_offset+1));
```
feilongjiang, line 221:
```suggestion
      load_unsigned_byte(tmp, Address(xbcp, bcp_offset + 1));
```

src/hotspot/cpu/riscv/interp_masm_riscv.cpp
```diff
@@ -191,13 +198,38 @@ void InterpreterMacroAssembler::get_dispatch() {
...
+      slli(tmp, tmp, 8);
+      add(index, index, tmp);
+      load_unsigned_byte(tmp, Address(xbcp, bcp_offset+2));
```
feilongjiang, line 224:
```suggestion
      load_unsigned_byte(tmp, Address(xbcp, bcp_offset + 2));
```

src/hotspot/cpu/riscv/interp_masm_riscv.cpp
```diff
@@ -191,13 +198,38 @@ void InterpreterMacroAssembler::get_dispatch() {
...
+      slli(tmp, tmp, 16);
+      add(index, index, tmp);
+      load_unsigned_byte(tmp, Address(xbcp, bcp_offset+3));
```
feilongjiang, line 227:
```suggestion
      load_unsigned_byte(tmp, Address(xbcp, bcp_offset + 3));
```

src/hotspot/cpu/riscv/interp_masm_riscv.cpp
```diff
@@ -191,13 +198,38 @@ void InterpreterMacroAssembler::get_dispatch() {
...
+      load_unsigned_byte(tmp, Address(xbcp, bcp_offset+1));
+      slli(tmp, tmp, 8);
+      add(index, index, tmp);
+    } else {
+      load_unsigned_short(index, Address(xbcp, bcp_offset));
+    }
   } else if (index_size == sizeof(u4)) {
-    lwu(index, Address(xbcp, bcp_offset));
+    if (AvoidUnalignedAccesses)
+    {
```
feilongjiang, line 218:
```suggestion
    if (AvoidUnalignedAccesses) {
```

src/hotspot/cpu/riscv/templateInterpreterGenerator_riscv.cpp
```diff
@@ -1098,7 +1098,22 @@ address TemplateInterpreterGenerator::generate_native_entry(bool synchronized) {
     __ ld(x28, Address(xmethod, Method::native_function_offset()));
     address unsatisfied = (SharedRuntime::native_method_throw_unsatisfied_link_error_entry());
     __ mv(t1, unsatisfied);
-    __ ld(t1, Address(t1, 0));
+    if (AvoidUnalignedAccesses)
+    {
```
feilongjiang, line 1102:
```suggestion
    if (AvoidUnalignedAccesses) {
```

src/hotspot/cpu/riscv/templateTable_riscv.cpp
```diff
@@ -1608,7 +1616,14 @@ void TemplateTable::branch(bool is_jsr, bool is_wide) {
...
   if (!is_wide) {
-    __ lhu(x12, at_bcp(1));
+    if (AvoidUnalignedAccesses) {
+        __ lbu(t1, at_bcp(1));
+        __ lbu(x12, at_bcp(2));
+        __ slli(x12, x12, 8);
+        __ add(x12, t1, x12);
+      } else {
+        __ lhu(x12, at_bcp(1));
+      }
```
feilongjiang, line 1626:
```suggestion
    if (AvoidUnalignedAccesses) {
      __ lbu(t1, at_bcp(1));
      __ lbu(x12, at_bcp(2));
      __ slli(x12, x12, 8);
      __ add(x12, t1, x12);
    } else {
      __ lhu(x12, at_bcp(1));
    }
```

src/hotspot/cpu/riscv/templateTable_riscv.cpp
```diff
@@ -2008,7 +2023,17 @@ void TemplateTable::fast_binaryswitch() {
     // else [i = h]
     // Convert array[h].match to native byte-ordering before compare
     __ shadd(temp, h, array, temp, 3);
-    __ ld(temp, Address(temp, 0));
+    if (AvoidUnalignedAccesses)
+    {
```
feilongjiang, line 2027:
```suggestion
    if (AvoidUnalignedAccesses) {
```

src/hotspot/cpu/riscv/templateTable_riscv.cpp
```diff
@@ -2008,7 +2023,17 @@ void TemplateTable::fast_binaryswitch() {
...
+      //array is BytesPerInt (aka 4) alligned
+      __ lwu(t1, Address(temp,4));
+      __ slli(t1, t1, 32);
+      __ lwu(temp, Address(temp, 0));
+      __ add(temp, temp, t1);
+    } else
+    {
```
feilongjiang, line 2034:
```suggestion
    } else {
```

src/hotspot/cpu/riscv/templateTable_riscv.cpp
```diff
@@ -2031,7 +2056,16 @@ void TemplateTable::fast_binaryswitch() {
   Label default_case;
   // Convert array[i].match to native byte-ordering before compare
   __ shadd(temp, i, array, temp, 3);
-  __ ld(temp, Address(temp, 0));
+  if (AvoidUnalignedAccesses)
+  {
```
feilongjiang, line 2060:
```suggestion
  if (AvoidUnalignedAccesses) {
```

src/hotspot/cpu/riscv/templateTable_riscv.cpp
```diff
@@ -2031,7 +2056,16 @@ void TemplateTable::fast_binaryswitch() {
...
+    //array is BytesPerInt (aka 4) alligned
```
feilongjiang, line 2061:
```suggestion
    // array is BytesPerInt (aka 4) alligned
```

src/hotspot/cpu/riscv/interp_masm_riscv.cpp
```diff
@@ -224,7 +256,8 @@ void InterpreterMacroAssembler::get_cache_and_index_at_bcp(Register cache,
                                                            size_t index_size) {
   assert_different_registers(cache, index);
   assert_different_registers(cache, xcpool);
-  get_cache_index_at_bcp(index, bcp_offset, index_size);
+  //register "cache" is trashed in next shadd, so lets use it as a temporary register
```
feilongjiang, line 259:
```suggestion
  // register "cache" is trashed in next shadd, so lets use it as a temporary register
```

src/hotspot/cpu/riscv/interp_masm_riscv.cpp
```diff
@@ -261,7 +294,8 @@ void InterpreterMacroAssembler::get_cache_entry_pointer_at_bcp(Register cache,
                                                                int bcp_offset,
                                                                size_t index_size) {
   assert_different_registers(cache, tmp);
-  get_cache_index_at_bcp(tmp, bcp_offset, index_size);
+  //register "cache" is trashed in next ld, so lets use it as a temporary register
```
feilongjiang, line 297:
```suggestion
  // register "cache" is trashed in next ld, so lets use it as a temporary register
```

src/hotspot/cpu/riscv/templateTable_riscv.cpp
```diff
@@ -368,7 +375,8 @@ void TemplateTable::fast_aldc(LdcType type) {
   // We are resolved if the resolved reference cache entry contains a
   // non-null object (String, MethodType, etc.)
   assert_different_registers(result, tmp);
-  __ get_cache_index_at_bcp(tmp, 1, index_size);
+  //register result is trashed by next load, let's use it as temporary register
```
feilongjiang, line 378:
```suggestion
  // register result is trashed by next load, let's use it as temporary register
```

src/hotspot/cpu/riscv/templateTable_riscv.cpp
```diff
@@ -2008,7 +2023,17 @@ void TemplateTable::fast_binaryswitch() {
     // else [i = h]
     // Convert array[h].match to native byte-ordering before compare
     __ shadd(temp, h, array, temp, 3);
-    __ ld(temp, Address(temp, 0));
+    if (AvoidUnalignedAccesses)
+    {
+      //array is BytesPerInt (aka 4) alligned
```
feilongjiang, line 2028:
```suggestion
      // array is BytesPerInt (aka 4) alligned
```

src/hotspot/cpu/riscv/templateInterpreterGenerator_riscv.cpp
```diff
@@ -1098,7 +1098,22 @@ address TemplateInterpreterGenerator::generate_native_entry(bool synchronized) {
...
+      __ slli(t0, t0, 16);
+      __ add(t1, t1, t0);
+      __ lhu(t0, Address(t, 4));
+      __ slli(t0, t0, 32);
+      __ add(t1, t1, t0);
+      __ lhu(t0, Address(t, 6));
+      __ slli(t0, t0, 48);
+      __ add(t1, t1, t0);
+    } else {
+      __ ld(t1, Address(t1, 0)); //2 bytes alligned, but not 4 or 8
```
feilongjiang, line 1115:
```suggestion
      __ ld(t1, Address(t1, 0)); // 2 bytes alligned, but not 4 or 8
```
Should we put this comment into `if` statement?

feilongjiang (changes requested):
Looks good, with some nits:

src/hotspot/cpu/riscv/interp_masm_riscv.cpp
```diff
@@ -191,13 +198,38 @@ void InterpreterMacroAssembler::get_dispatch() {
...
+      slli(tmp, tmp, 8);
+      add(index, index, tmp);
+    } else {
+      load_unsigned_short(index, Address(xbcp, bcp_offset));
+    }
   } else if (index_size == sizeof(u4)) {
-    lwu(index, Address(xbcp, bcp_offset));
+    if (AvoidUnalignedAccesses)
+    {
+      assert(index != tmp, "must use different register");
```
feilongjiang, line 219:
```suggestion
      assert_different_registers(index, tmp");
```

VladimirKempik:
Thanks for taking a look Feilong Jiang

src/hotspot/cpu/riscv/assembler_riscv.hpp
```diff
@@ -1933,10 +1949,10 @@ enum Nf {
     uint16_t mask = (1U << nbits) - 1;
     val <<= lsb;
     mask <<= lsb;
-    uint16_t target = *(uint16_t *)a;
+    uint16_t target = ld_c_instr(a);
     target &= ~mask;
     target |= val;
-    *(uint16_t *)a = target;
+    sd_c_instr(a, target);
```
RealFYang, line 1955: Compressed instructions are supposed to be located at some addresses at least 2-bytes aligned. So I am thinking that there shouldn't be any unaligned access happening here.

src/hotspot/cpu/riscv/macroAssembler_riscv.hpp
```diff
@@ -344,6 +344,10 @@ class MacroAssembler: public Assembler {
 #define verify_method_ptr(reg) _verify_method_ptr(reg, "broken method " #reg, __FILE__, __LINE__)
 #define verify_klass_ptr(reg) _verify_method_ptr(reg, "broken klass " #reg, __FILE__, __LINE__)
 
+#define assert_addr_alignment_cond(cond, addr)       assert(!(cond) || is_aligned(addr, 4), "bad addr alignment: " INTPTR_FORMAT, p2i(addr))
+#define assert_addr_alignment(addr)                  assert_addr_alignment_cond(true, addr)
+#define assert_offset_alignment_cond(cond, offset)   assert(!(cond) || ((offset) % 4) == 0, "bad offset alignment: %d", offset)
```
RealFYang, line 349: Looks like these macros are not used anywhere?

src/hotspot/cpu/riscv/assembler_riscv.hpp
VladimirKempik: on one hand - you are right, on another one - this way it looks more unified with 4-byte opcodes loads/stores

src/hotspot/cpu/riscv/templateTable_riscv.cpp
```diff
@@ -2008,7 +2023,15 @@ void TemplateTable::fast_binaryswitch() {
...
-    __ ld(temp, Address(temp, 0));
+    if (AvoidUnalignedAccesses) {
+      // array is BytesPerInt (aka 4) aligned
+      __ lwu(t1, Address(temp,4));
+      __ slli(t1, t1, 32);
+      __ lwu(temp, Address(temp, 0));
+      __ add(temp, temp, t1);
+    } else {
+      __ ld(temp, Address(temp, 0));
+    }
```
RealFYang, line 2034: This if-else could be further simplified into a single "__ lwu(temp, Address(temp, 0));" since only the low 32 bits of 'temp' register is used by following 'revb_w_w'.

src/hotspot/cpu/riscv/templateTable_riscv.cpp
```diff
@@ -2031,7 +2054,15 @@ void TemplateTable::fast_binaryswitch() {
...
-  __ ld(temp, Address(temp, 0));
+  if (AvoidUnalignedAccesses) {
+    // array is BytesPerInt (aka 4) aligned
+    __ lwu(t1, Address(temp,4));
+    __ slli(t1, t1, 32);
+    __ lwu(temp, Address(temp, 0));
+    __ add(temp, temp, t1);
+  } else {
+    __ ld(temp, Address(temp, 0));
+  }
```
RealFYang, line 2065: Similar here. This if-else could be further simplified into a single "__ lwu(temp, Address(temp, 0));" too.

RealFYang (changes requested):

src/hotspot/cpu/riscv/nativeInst_riscv.hpp
```diff
@@ -208,18 +208,18 @@ class NativeInstruction {
...
 
 
-  void set_int_at(int offset, jint  i)        { *(jint*)addr_at(offset) = i; }
-  void set_uint_at(int offset, jint  i)       { *(juint*)addr_at(offset) = i; }
-  void set_ptr_at (int offset, address  ptr)  { *(address*) addr_at(offset) = ptr; }
-  void set_oop_at (int offset, oop  o)        { *(oop*) addr_at(offset) = o; }
+  void set_int_at(int offset, jint  i)        { Bytes::put_native_u4(addr_at(offset), i); }
+  void set_uint_at(int offset, jint  i)       { Bytes::put_native_u4(addr_at(offset), i); }
+  void set_ptr_at (int offset, address  ptr)  { Bytes::put_native_u8(addr_at(offset), (u8)ptr); }
+  void set_oop_at (int offset, oop  o)        { Bytes::put_native_u8(addr_at(offset), cast_from_oop<u8>(o)); }
```
RealFYang, line 222: I see there are two spaces between type and name for the second parameter. We should remove one.

src/hotspot/cpu/riscv/templateTable_riscv.cpp
```diff
@@ -280,7 +280,14 @@ void TemplateTable::bipush() {
...
+  if (AvoidUnalignedAccesses) {
+    __ lbu(t1, at_bcp(1));
+    __ lbu(x10, at_bcp(2));
+    __ slli(x10, x10, 8);
+    __ add(x10, t1, x10);
+  } else {
+    __ load_unsigned_short(x10, at_bcp(1));
+  }
   __ revb_w_w(x10, x10);
   __ sraiw(x10, x10, 16);
```
RealFYang, line 292: I think we can further simplify this sequence into something like:
```
if (AvoidUnalignedAccesses) {
  __ load_signed_byte(x10, at_bcp(1));
  __ load_unsigned_byte(t1, at_bcp(2));
  __ slli(x10, x10, 8);
  __ add(x10, x10, t1);
} else {
  __ load_unsigned_short(x10, at_bcp(1));
  __ revb_w_w(x10, x10);
  __ sraiw(x10, x10, 16);
}
```

src/hotspot/cpu/riscv/templateTable_riscv.cpp
```diff
@@ -1608,7 +1616,14 @@ void TemplateTable::branch(bool is_jsr, bool is_wide) {
...
-    __ lhu(x12, at_bcp(1));
+    if (AvoidUnalignedAccesses) {
+      __ lbu(t1, at_bcp(1));
+      __ lbu(x12, at_bcp(2));
+      __ slli(x12, x12, 8);
+      __ add(x12, t1, x12);
+    } else {
+      __ lhu(x12, at_bcp(1));
+    }
     __ revb_h_h(x12, x12); // reverse bytes in half-word and sign-extend
```
RealFYang, line 1627: Similar here. Consider further optimizing this sequence into something like:
```
if (AvoidUnalignedAccesses) {
  __ lb(x12, at_bcp(1));
  __ lbu(t1, at_bcp(2));
  __ slli(x12, x12, 8);
  __ add(x12, x12, t1);
} else {
  __ lhu(x12, at_bcp(1));
  __ revb_h_h(x12, x12); // reverse bytes in half-word and sign-extend
}
```

RealFYang (changes requested):
Thanks for the update. Would you mind several more tweaks? Otherwise, LGTM.

src/hotspot/cpu/riscv/templateTable_riscv.cpp
VladimirKempik: why have you replaced revb_w_w with revb_h_h ?

RealFYang: > why have you replaced revb_w_w with revb_h_h ?

Because it's cheaper in respect of number of instructions emitted when Zbb extension is not available and achieves the same functionality.

VladimirKempik: A thanks, took time to understand the  __ revb_w_w(x10, x10); __ sraiw(x10, x10, 16); was kind of too much here

src/hotspot/cpu/riscv/interp_masm_riscv.cpp
```diff
@@ -191,13 +198,36 @@ void InterpreterMacroAssembler::get_dispatch() {
...
+      load_unsigned_byte(tmp, Address(xbcp, bcp_offset + 1));
+      slli(tmp, tmp, 8);
+      add(index, index, tmp);
+      load_unsigned_byte(tmp, Address(xbcp, bcp_offset + 2));
+      slli(tmp, tmp, 16);
+      add(index, index, tmp);
+      load_unsigned_byte(tmp, Address(xbcp, bcp_offset + 3));
+      slli(tmp, tmp, 24);
+      add(index, index, tmp);
+    } else {
```
theRealAph, line 228: Does it really make sense to do this here? Shouldn't there be an unaligned version of `lwu` in MacroAssembler?

VladimirKempik: Updated the PR, load_word_misaligned looks slightly ugly due to **Address -> (Address + offset)** conversion

src/hotspot/cpu/riscv/macroAssembler_riscv.cpp
```diff
@@ -1689,6 +1689,23 @@ void MacroAssembler::store_sized_value(Address dst, Register src, size_t size_in
   }
 }
 
+void MacroAssembler::load_word_misaligned(Register dst, Address src, Register tmp, bool is_signed) {
```
RealFYang, line 1692: I am afraid that the function name is a bit confusing considering that the well-known global 'wordSize' is 8 on linux-riscv64. But we are actually loading 4 bytes here. How about renaming it as something like "load_int_unaligned"? This would align with existing function names like 'load_unsigned_byte' and 'load_unsigned_short'. Also, it's safer to add an assertion to make sure that 'dst' and 'tmp' are different registers.

VladimirKempik:
On long runs I still can see some small amount of ptr_lam event, however they aren't originating from TemplateInterpreter and should be a part of different PR

`
# Samples: 1K of event 'trp_lam'
# Event count (approx.): 5240
#
# Overhead  Command   Shared Object    Symbol                                                                              
# ........  ........  ...............  ....................................................................................
#
    76.95%  java      [JIT] tid 41259  [.] boolean java.lang.String.equals(java.lang.Object)
     9.22%  Thread-4  [JIT] tid 41259  [.] boolean java.lang.String.equals(java.lang.Object)
     6.47%  Thread-2  [JIT] tid 41259  [.] boolean java.lang.String.equals(java.lang.Object)
     5.65%  Thread-3  [JIT] tid 41259  [.] boolean java.lang.String.equals(java.lang.Object)
     0.53%  java      [JIT] tid 41259  [.] int jdk.internal.org.objectweb.asm.SymbolTable.addConstantUtf8(java.lang.String)
     0.44%  Thread-2  [JIT] tid 41259  [.] int jdk.internal.org.objectweb.asm.SymbolTable.addConstantUtf8(java.lang.String)
     0.38%  Thread-4  [JIT] tid 41259  [.] int jdk.internal.org.objectweb.asm.SymbolTable.addConstantUtf8(java.lang.String)
     0.36%  Thread-3  [JIT] tid 41259  [.] int jdk.internal.org.objectweb.asm.SymbolTable.addConstantUtf8(java.lang.String)
`

feilongjiang:
If I remember correctly, there are some misaligned access at string intrinsics. Here are the related PRs at riscv-collab:

- https://github.com/riscv-collab/riscv-openjdk/pull/19
- https://github.com/riscv-collab/riscv-openjdk/pull/17
- https://github.com/riscv-collab/riscv-openjdk/pull/14

feilongjiang:
> @feilongjiang , do you know any reason why first two (string_equals & string_compare) wasn't ever integrated ?

At that time, we were focused on upstreaming the risc-v port. The misaligned access issues for those intrinsics are not a high priority. So we just reverted string_equals changes, and string_compare was closed before being integrated.

VladimirKempik:
@feilongjiang 
I have applied your patch for string_equals, and now amount of trp_lam event is very low ( about a dozen)
mostly coming from
        Thread-2 43606 851688.227592:          1 trp_lam:        3f88c5e4aa JVM_handle_linux_signal+0xee (/home/vkempik/syntaj/lib/server/libjvm.so)


Would you open a new PR with string_equals patch ? or I can just add that patch into this PR

feilongjiang:
> Would you open a new PR with string_equals patch ? or I can just add that patch into this PR
I think it's okay to just add that patch to the current PR, TIA.

feilongjiang (approved):

src/hotspot/cpu/riscv/c2_MacroAssembler_riscv.cpp
```diff
@@ -1143,55 +1142,28 @@ void C2_MacroAssembler::string_equals(Register a1, Register a2,
...
-    }
-  }
+  bind(SHORT);
+  ld(tmp1, Address(a1));
+  ld(tmp2, Address(a2));
+  xorr(tmp1, tmp1, tmp2);
+  neg(cnt1, cnt1);
+  slli(cnt1, cnt1, LogBitsPerByte);
+  sll(tmp1, tmp1, cnt1);
+  bnez(tmp1, DONE);
```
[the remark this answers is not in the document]
VladimirKempik: that is hard to say.

OoO arches such as thead - don't care about the location of xor opcode here

In order uarches, such as u74/hifive might be affected by such change. however, the memory at address a1/a2 very likely would already be in the l1d cache, due to previous accesses in the same function, so it will be pretty cheap.
u74 is dual-issue, so it may execute these two loads (from l1d$) in parallel, and will wait for them to finish anyway, no matter what is next opcode.

To say for sure, need to check with jmh test org.openjdk.bench.java.lang.StringEquals on hifive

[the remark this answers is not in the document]
VladimirKempik: Before the PR

```
Benchmark                      Mode  Cnt     Score   Error  Units
StringEquals.almostEqual       avgt   25  1214.131 ± 4.400  ns/op
StringEquals.almostEqualUTF16  avgt   25  1213.310 ± 7.156  ns/op
StringEquals.different         avgt   25    20.102 ± 2.306  ns/op
StringEquals.differentCoders   avgt   25    14.780 ± 1.147  ns/op
StringEquals.equal             avgt   25  1218.393 ± 5.275  ns/op
StringEquals.equalsUTF16       avgt   25  1216.750 ± 4.383  ns/op
```


With this PR

```
Benchmark                      Mode  Cnt   Score   Error  Units
StringEquals.almostEqual       avgt   25  28.584 ± 1.178  ns/op
StringEquals.almostEqualUTF16  avgt   25  28.375 ± 1.052  ns/op
StringEquals.different         avgt   25  19.572 ± 1.031  ns/op
StringEquals.differentCoders   avgt   25  14.969 ± 2.348  ns/op
StringEquals.equal             avgt   25  28.603 ± 0.148  ns/op
StringEquals.equalsUTF16       avgt   25  29.217 ± 1.969  ns/op
```

Xor moved

```
Benchmark                      Mode  Cnt   Score   Error  Units
StringEquals.almostEqual       avgt   25  28.455 ± 1.068  ns/op
StringEquals.almostEqualUTF16  avgt   25  28.244 ± 0.920  ns/op
StringEquals.different         avgt   25  18.940 ± 0.831  ns/op
StringEquals.differentCoders   avgt   25  14.566 ± 1.298  ns/op
StringEquals.equal             avgt   25  27.891 ± 0.606  ns/op
StringEquals.equalsUTF16       avgt   25  28.294 ± 0.913  ns/op
```

hard to say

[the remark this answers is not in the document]
RealFYang: Thanks for trying this out. Another issue here is that 8-byte memory accesses at address a1/a2 would exceed the range for strings whose size is smaller than wordSize. I am not quite sure whether that is safe to do. Could we just incorporate changes which only resolves the unaligned access problem here?

src/hotspot/cpu/riscv/templateInterpreterGenerator_riscv.cpp
```diff
@@ -1098,7 +1098,21 @@ address TemplateInterpreterGenerator::generate_native_entry(bool synchronized) {
...
+      __ lhu(t1, Address(t1, 0));
+      __ lhu(t0, Address(t, 2));
+      __ slli(t0, t0, 16);
+      __ add(t1, t1, t0);
+      __ lhu(t0, Address(t, 4));
+      __ slli(t0, t0, 32);
+      __ add(t1, t1, t0);
+      __ lhu(t0, Address(t, 6));
+      __ slli(t0, t0, 48);
+      __ add(t1, t1, t0);
```
RealFYang, line 1112: Maybe another `load_long_misaligned` assembler function for this long sequence?

src/hotspot/cpu/riscv/c2_MacroAssembler_riscv.cpp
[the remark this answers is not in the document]
VladimirKempik: On an fpga I can see these numbers with perfnorm profiler:

Just this PR:

Secondary result "org.openjdk.bench.java.lang.StringEquals.equal:IPC":
  1.245 ±(99.9%) 0.037 insns/clk [Average]
  (min, avg, max) = (1.234, 1.245, 1.258), stdev = 0.010
  CI (99.9%): [1.209, 1.282] (assumes normal distribution)

This PR + moving xor:

Secondary result "org.openjdk.bench.java.lang.StringEquals.equal:IPC":
  1.239 ±(99.9%) 0.050 insns/clk [Average]
  (min, avg, max) = (1.224, 1.239, 1.253), stdev = 0.013
  CI (99.9%): [1.190, 1.289] (assumes normal distribution)

src/hotspot/cpu/riscv/templateInterpreterGenerator_riscv.cpp
VladimirKempik: yeah, there is one small issue with that, loading granularity, here it's two, I will have to make universal "load_long_misaligned" and pass load granularity to it. Still could be useful in future tho

src/hotspot/cpu/riscv/c2_MacroAssembler_riscv.cpp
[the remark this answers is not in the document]
VladimirKempik: You are right, if the object is located at the end of an allocated memory region ( and nothing past it) then this might produce sigsegv.

I'll ty to modify old version of string_equals to use TAIL logic for the tail of long strings, lwu+lhu+lbu is slower than ld but still faster than falling to misaligned access emulator.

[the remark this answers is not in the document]
VladimirKempik: new version results in jmh:

```
Benchmark                      Mode  Cnt   Score   Error  Units
StringEquals.almostEqual       avgt   25  29.459 ± 0.173  ns/op
StringEquals.almostEqualUTF16  avgt   25  29.833 ± 0.596  ns/op
StringEquals.different         avgt   25  19.896 ± 1.024  ns/op
StringEquals.differentCoders   avgt   25  14.986 ± 1.748  ns/op
StringEquals.equal             avgt   25  31.174 ± 0.226  ns/op
StringEquals.equalsUTF16       avgt   25  32.718 ± 0.982  ns/op
```

should not read past an object boundaries anymore

src/hotspot/cpu/riscv/c2_MacroAssembler_riscv.cpp
```diff
@@ -1141,19 +1144,24 @@ void C2_MacroAssembler::string_equals(Register a1, Register a2,
...
+    // length == 4.
+    add(tmp1, a1, cnt1);
+    ld(tmp1, Address(tmp1, 0));
+    add(tmp2, a2, cnt1);
+    ld(tmp2, Address(tmp2, 0));
+    bne(tmp1, tmp2, DONE);
+    j(SAME);
+  } else {
+    add(tmp1, cnt1, wordSize);
+    beqz(tmp1, SAME);
```
RealFYang, line 1163: I think this change here resolves my previous concern. But I witnessed some usage of registers `t0` and `t1` in this function. I think we should replace them with their aliases 'tmp1' and 'tmp2' respectively. Could you please help do that cleanup while you are on it?

RealFYang (changes requested):
Thanks for the update. Would you mind a few more tweaks?

src/hotspot/cpu/riscv/macroAssembler_riscv.cpp
```diff
@@ -1690,6 +1690,80 @@ void MacroAssembler::store_sized_value(Address dst, Register src, size_t size_in
   }
 }
 
+void MacroAssembler::load_int_misaligned(Register dst, Address src, Register tmp, bool is_signed) {
```
feilongjiang, line 1693: `load_long_misaligned` provides `granularity`, do we need this in `load_int_misaligned` too?

src/hotspot/cpu/riscv/interp_masm_riscv.cpp
```diff
@@ -177,7 +177,14 @@ void InterpreterMacroAssembler::check_and_handle_earlyret(Register java_thread)
...
-  lhu(reg, Address(xbcp, bcp_offset));
+  if (AvoidUnalignedAccesses && (bcp_offset % 2)) {
+    lbu(t1, Address(xbcp, bcp_offset));
+    lbu(reg, Address(xbcp, bcp_offset + 1));
+    slli(reg, reg, 8);
+    add(reg, t1, reg);
+  } else {
+    lhu(reg, Address(xbcp, bcp_offset));
+  }
   revb_h(reg, reg);
```
feilongjiang, line 188: Similiar to `sipush`, `revb_h` is not needed for misaligned load. And since here we only load a short, looks like `revb_h_h_u` is enough.
```suggestion
  if (AvoidUnalignedAccesses && (bcp_offset % 2)) {
    lbu(t1, Address(xbcp, bcp_offset));
    lbu(reg, Address(xbcp, bcp_offset + 1));
    slli(t1, t1, 8);
    add(reg, reg, t1);
  } else {
    lhu(reg, Address(xbcp, bcp_offset));
    revb_h_h_u(reg, reg);
  }
  
```

feilongjiang, line 188: Similiar to `sipush`, `revb_h` is not needed for misaligned load. And since here we only load a short, looks like `revb_h_h_u` is enough.
```suggestion
  if (AvoidUnalignedAccesses && (bcp_offset % 2)) {
    lbu(t1, Address(xbcp, bcp_offset));
    lbu(reg, Address(xbcp, bcp_offset + 1));
    slli(t1, t1, 8);
    add(reg, reg, t1);
  } else {
    lhu(reg, Address(xbcp, bcp_offset));
    revb_h_h_u(reg, reg);
  }
  
```

src/hotspot/cpu/riscv/macroAssembler_riscv.cpp
feilongjiang, line 1693: `load_long_misaligned` provides `granularity`, maybe we add this to `load_int_misaligned` too? If granularity is 2, we can just use two `lh`s to load an int.

feilongjiang (commented):
Some comments for new changes.

src/hotspot/cpu/riscv/templateInterpreterGenerator_riscv.cpp
```diff
@@ -1097,8 +1097,12 @@ address TemplateInterpreterGenerator::generate_native_entry(bool synchronized) {
     Label L;
     __ ld(x28, Address(xmethod, Method::native_function_offset()));
     address unsatisfied = (SharedRuntime::native_method_throw_unsatisfied_link_error_entry());
-    __ mv(t1, unsatisfied);
-    __ ld(t1, Address(t1, 0));
+    __ mv(t, unsatisfied);
+    if (AvoidUnalignedAccesses) {
```
RealFYang, line 1101: Flag `AvoidUnalignedAccesses` is checked in function `load_long_misaligned`. No need for another check for the same flag here. So seems that the if-else structure here could be simplified into one single line:
```
__ load_long_misaligned(t1, Address(t, 0), t0, 2); // 2 bytes aligned, but not 4 or 8
```
Looks good otherwise. Thanks again.

feilongjiang (approved):

RealFYang (approved):
Updated change LGTM. Thanks for your patience.

VladimirKempik:
tier1/tier2 are good again, thanks for reviews
/integrate

PR #13652, opened 2023-04-25
8306460: Clear JVM_ACC_QUEUED flag on methods when dumping dynamic CDS archive

ashu-mehra:
This patch clears the method's "queued_for_compilation" flag when dumping the method in CDS archive. Also added an assert in `Method::restore_unshareable_info()` that the method being restored should not have that flag set.

---------
### Progress

### Issue
 * [JDK-8306460](https://bugs.openjdk.org/browse/JDK-8306460): Clear JVM_ACC_QUEUED flag on methods when dumping dynamic CDS archive

### Reviewers
 * [Coleen Phillimore](https://openjdk.org/census#coleenp) (@coleenp - **Reviewer**) ⚠️ Review applies to [82b9c715](https://git.openjdk.org/jdk/pull/13652/files/82b9c715eac502f2bdfb0f131a77daeb42403115)
 * [Ioi Lam](https://openjdk.org/census#iklam) (@iklam - **Reviewer**) ⚠️ Review applies to [82b9c715](https://git.openjdk.org/jdk/pull/13652/files/82b9c715eac502f2bdfb0f131a77daeb42403115)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13652/head:pull/13652` \
`$ git checkout pull/13652`

`$ git checkout pull/13652` \
`$ git pull https://git.openjdk.org/jdk.git pull/13652/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13652`

`$ git pr show -t 13652`

</details>

<a href="https://git.openjdk.org/jdk/pull/13652.diff">https://git.openjdk.org/jdk/pull/13652.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13652#issuecomment-1522204787)


coleenp (approved):
Looks good.

ashu-mehra:
@coleenp thanks for the review.

iwanowww:
There are other compiler-related flags (e.g., `JVM_ACC_NOT_C1_COMPILABLE`/`JVM_ACC_NOT_C2_COMPILABLE`/`JVM_ACC_NOT_C2_OSR_COMPILABLE` ) which should not be persisted into CDS archive.

Also, I assume the following flags should be always unset in the archived version:
`JVM_ACC_IS_PREFIXED_NATIVE`, `JVM_ACC_HAS_RESOLVED_METHOD`, `JVM_ACC_IS_OLD`, `JVM_ACC_IS_OBSOLETE`, `JVM_ACC_ON_STACK`, `JVM_ACC_IS_DELETED`, `JVM_ACC_IS_BEING_REDEFINED`.

coleenp (commented):
Yes, you're right, all these flags shouldn't be in the archive.  I have a patch for JDK-8306851 which will make it easier to unset all of these flags (except has_loops/has_loops_init, which we want set in the archive).  Maybe this change should wait.

ashu-mehra:
@coleenp that makes sense. Once your change is in, I can update this to clear all the flags that shouldn't be in the archived version.

ashu-mehra:
@iwanowww I don't see `JVM_ACC_HAS_RESOLVED_METHOD` and `JVM_ACC_IS_BEING_REDEFINED` used/referred anywhere in the code base. Where do you see them?

ashu-mehra:
@coleenp`is_old`, `is_obsolete`, and  `is_deleted` method flags are set only when a method is redefined, and such methods would not be added to the archive.
I am wondering if, instead of clearing these flags, there should be an assert added that these are not already set when dumping the method to the CDS archive.

ashu-mehra:
@coleenp, @iklam can you please review this PR.

iklam (approved):
LGTM.

src/hotspot/share/oops/instanceKlass.cpp
```diff
@@ -2592,6 +2592,16 @@ void InstanceKlass::remove_unshareable_info() {
...
   _dep_context_last_cleaned = 0;
   _init_monitor = nullptr;
+
+  remove_unshareable_flags();
+}
+
+void InstanceKlass::remove_unshareable_flags() {
+  // clear all the flags/stats that shouldn't be in the archived version
+#if INCLUDE_JVMTI
+  set_is_being_redefined(false);
```
coleenp, line 2602: I think this should assert !is_scratch_class() and clear (?) is_redefined() also.  It's unfortunate that we can't just clear all the status flags here and in Method.

coleenp (changes requested):
I think is_being_redefined shouldn't be set at this point, and should just be asserted.

src/hotspot/share/oops/instanceKlass.cpp
ashu-mehra: > clear (?) is_redefined() also.

There is no such flag; did you mean `has_been_redefined`? Instead of clearing it, shouldn't it be an assert `!has_been_redefined()` as well?

ashu-mehra:
> I think is_being_redefined shouldn't be set at this point, and should just be asserted.

@coleenp Isn't it possible for a class being redefined to be added to the CDS archive? I don't see any check preventing that?

iklam:
> > I think is_being_redefined shouldn't be set at this point, and should just be asserted.
> 
> @coleenp Isn't it possible for a class being redefined to be added to the CDS archive? I don't see any check preventing that. Did I miss something?

Classes that have been redefined are excluded from the CDS archive. See:

https://github.com/openjdk/jdk/blob/488330d53bb782657378424421a9ce2f2eed5e88/src/hotspot/share/classfile/systemDictionaryShared.cpp#L264-L275

ashu-mehra:
thank you @coleenp and @iklam for reviewing this patch.

PR #13654, opened 2023-04-25
8306851: Move Method access flags

coleenp:
This change moves the flags from AccessFlags to either ConstMethodFlags or MethodFlags, depending on whether they are set at class file parse time, which makes them essentially const, or at runtime, which makes them needing atomic access.

This leaves AccessFlags int size because Klass still has JVM flags that are more work to move, but this change doesn't increase Method size.  I didn't remove JVM_RECOGNIZED_METHOD_MODIFIERS with this change since there are several of these in other places, and with this change the code is benign.

Tested with tier1-6, and some manual verification of printing.

---------
### Progress

### Issue
 * [JDK-8306851](https://bugs.openjdk.org/browse/JDK-8306851): Move Method access flags

### Reviewers
 * [Chris Plummer](https://openjdk.org/census#cjplummer) (@plummercj - **Reviewer**) ⚠️ Review applies to [4b92aacf](https://git.openjdk.org/jdk/pull/13654/files/4b92aacf22901382ada523d83c539384fee260d8)
 * [David Holmes](https://openjdk.org/census#dholmes) (@dholmes-ora - **Reviewer**) ⚠️ Review applies to [6687cc0e](https://git.openjdk.org/jdk/pull/13654/files/6687cc0eb476e5ec8528f9d5b536b7ce64bc9ff1)
 * [Doug Simon](https://openjdk.org/census#dnsimon) (@dougxc - Committer) ⚠️ Review applies to [6687cc0e](https://git.openjdk.org/jdk/pull/13654/files/6687cc0eb476e5ec8528f9d5b536b7ce64bc9ff1)
 * [Matias Saavedra Silva](https://openjdk.org/census#matsaave) (@matias9927 - Committer) ⚠️ Review applies to [9b05ab84](https://git.openjdk.org/jdk/pull/13654/files/9b05ab84cb61669602047180763f88c234f3b2b4)
 * [Frederic Parain](https://openjdk.org/census#fparain) (@fparain - **Reviewer**) ⚠️ Review applies to [6687cc0e](https://git.openjdk.org/jdk/pull/13654/files/6687cc0eb476e5ec8528f9d5b536b7ce64bc9ff1)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13654/head:pull/13654` \
`$ git checkout pull/13654`

`$ git checkout pull/13654` \
`$ git pull https://git.openjdk.org/jdk.git pull/13654/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13654`

`$ git pr show -t 13654`

</details>

<a href="https://git.openjdk.org/jdk/pull/13654.diff">https://git.openjdk.org/jdk/pull/13654.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13654#issuecomment-1523679239)

### DIFF
--- a/src/hotspot/share/classfile/classFileParser.cpp
+++ b/src/hotspot/share/classfile/classFileParser.cpp
@@ -1992,7 +1992,7 @@ void MethodAnnotationCollector::apply_to(const methodHandle& m) {
   if (has_annotation(_method_LambdaForm_Compiled) && m->intrinsic_id() == vmIntrinsics::_none)
     m->set_intrinsic_id(vmIntrinsics::_compiledLambdaForm);
   if (has_annotation(_method_Hidden))
-    m->set_hidden();
+    m->set_is_hidden();
   if (has_annotation(_method_Scoped))
     m->set_scoped();
   if (has_annotation(_method_IntrinsicCandidate) && !m->is_synthetic())
@@ -2738,8 +2738,8 @@ Method* ClassFileParser::parse_method(const ClassFileStream* const cfs,
   if (parsed_annotations.has_any_annotations())
     parsed_annotations.apply_to(methodHandle(THREAD, m));
 
-  if (is_hidden() && !m->is_hidden()) { // Mark methods in hidden classes as 'hidden'.
-    m->set_hidden();
+  if (is_hidden()) { // Mark methods in hidden classes as 'hidden'.
+    m->set_is_hidden();
   }
 
   // Copy annotations
--- a/src/hotspot/share/classfile/classFileParser.cpp
+++ b/src/hotspot/share/classfile/classFileParser.cpp
@@ -1978,27 +1978,27 @@ ClassFileParser::FieldAnnotationCollector::~FieldAnnotationCollector() {
 
 void MethodAnnotationCollector::apply_to(const methodHandle& m) {
   if (has_annotation(_method_CallerSensitive))
-    m->set_caller_sensitive(true);
+    m->set_caller_sensitive();
   if (has_annotation(_method_ForceInline))
-    m->set_force_inline(true);
+    m->set_force_inline();
   if (has_annotation(_method_DontInline))
-    m->set_dont_inline(true);
+    m->set_dont_inline();
   if (has_annotation(_method_ChangesCurrentThread))
-    m->set_changes_current_thread(true);
+    m->set_changes_current_thread();
   if (has_annotation(_method_JvmtiMountTransition))
-    m->set_jvmti_mount_transition(true);
+    m->set_jvmti_mount_transition();
   if (has_annotation(_method_InjectedProfile))
-    m->set_has_injected_profile(true);
+    m->set_has_injected_profile();
   if (has_annotation(_method_LambdaForm_Compiled) && m->intrinsic_id() == vmIntrinsics::_none)
     m->set_intrinsic_id(vmIntrinsics::_compiledLambdaForm);
   if (has_annotation(_method_Hidden))
-    m->set_hidden(true);
+    m->set_hidden();
   if (has_annotation(_method_Scoped))
-    m->set_scoped(true);
+    m->set_scoped();
   if (has_annotation(_method_IntrinsicCandidate) && !m->is_synthetic())
-    m->set_intrinsic_candidate(true);
+    m->set_intrinsic_candidate();
   if (has_annotation(_jdk_internal_vm_annotation_ReservedStackAccess))
-    m->set_has_reserved_stack_access(true);
+    m->set_has_reserved_stack_access();
 }
 
 void ClassFileParser::ClassAnnotationCollector::apply_to(InstanceKlass* ik) {
@@ -2738,8 +2738,8 @@ Method* ClassFileParser::parse_method(const ClassFileStream* const cfs,
   if (parsed_annotations.has_any_annotations())
     parsed_annotations.apply_to(methodHandle(THREAD, m));
 
-  if (is_hidden()) { // Mark methods in hidden classes as 'hidden'.
-    m->set_hidden(true);
+  if (is_hidden() && !m->is_hidden()) { // Mark methods in hidden classes as 'hidden'.
+    m->set_hidden();
   }
 
   // Copy annotations
--- a/src/hotspot/share/oops/constMethod.cpp
+++ b/src/hotspot/share/oops/constMethod.cpp
@@ -430,14 +430,14 @@ void ConstMethod::print_on(outputStream* st) const {
   ResourceMark rm;
   st->print_cr("%s", internal_name());
   Method* m = method();
-  st->print(" - method:       " PTR_FORMAT " ", p2i(m));
+  st->print(" - method:           " PTR_FORMAT " ", p2i(m));
   if (m != nullptr) {
     m->print_value_on(st);
   }
   st->cr();
-  st->print(" - flags:             "); _flags.print_on(st);               st->cr();
+  st->print(" - flags:            0x%x  ", _flags.as_int()); _flags.print_on(st); st->cr();
   if (has_stackmap_table()) {
-    st->print(" - stackmap data:       ");
+    st->print(" - stackmap data:    ");
     stackmap_data()->print_value_on(st);
     st->cr();
   }
--- a/src/hotspot/share/oops/constMethod.cpp
+++ b/src/hotspot/share/oops/constMethod.cpp
@@ -199,7 +199,7 @@ u2* ConstMethod::checked_exceptions_length_addr() const {
 }
 
 u2* ConstMethod::exception_table_length_addr() const {
-  assert(has_exception_handler(), "called only if table is present");
+  assert(has_exception_table(), "called only if table is present");
   if (has_checked_exceptions()) {
     // If checked_exception present, locate immediately before them.
     return (u2*) checked_exceptions_start() - 1;
@@ -217,7 +217,7 @@ u2* ConstMethod::exception_table_length_addr() const {
 
 u2* ConstMethod::localvariable_table_length_addr() const {
   assert(has_localvariable_table(), "called only if table is present");
-  if (has_exception_handler()) {
+  if (has_exception_table()) {
     // If exception_table present, locate immediately before them.
     return (u2*) exception_table_start() - 1;
   } else {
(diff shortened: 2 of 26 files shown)


plummercj (approved):
The SA changes look good. I think these changes make @iklam's #13663 fix unnecessary, but harmless.

coleenp:
Thanks Chris.  We were wondering what to do with JVM_RECOGNIZED_METHOD_MODIFIERS but we'll clean them up in another pass.

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -2738,8 +2738,8 @@ Method* ClassFileParser::parse_method(const ClassFileStream* const cfs,
   if (parsed_annotations.has_any_annotations())
     parsed_annotations.apply_to(methodHandle(THREAD, m));
 
-  if (is_hidden()) { // Mark methods in hidden classes as 'hidden'.
-    m->set_hidden(true);
+  if (is_hidden() && !m->is_hidden()) { // Mark methods in hidden classes as 'hidden'.
```
dholmes-ora, line 2741: This seems odd - how would m already be marked hidden? And why do we care? The check-and-branch is more expensive than just setting the field.

src/hotspot/share/oops/constMethodFlags.hpp
```diff
@@ -0,0 +1,93 @@
...
+#ifndef SHARE_OOPS_CONSTMETHODFLAGS_HPP
+#define SHARE_OOPS_CONSTMETHODFLAGS_HPP
+
+#include "utilities/globalDefinitions.hpp"
+#include "utilities/macros.hpp"
+
+class outputStream;
+
+// The ConstMethodFlags class contains the parse-time flags associated with
+// an Method, and their associated accessors.
```
dholmes-ora, line 34: s/an/a/
s/their/its/

src/hotspot/share/oops/constMethodFlags.hpp
```diff
@@ -0,0 +1,93 @@
...
+   flag(has_localvariable_table   , 1 << 2) \
+   flag(has_exception_table       , 1 << 3) \
+   flag(has_generic_signature     , 1 << 4) \
+   flag(has_method_parameters     , 1 << 5) \
+   flag(is_overpass               , 1 << 6) \
+   flag(has_method_annotations    , 1 << 7) \
+   flag(has_parameter_annotations , 1 << 8) \
+   flag(has_type_annotations      , 1 << 9) \
+   flag(has_default_annotations   , 1 << 10) \
+   flag(caller_sensitive          , 1 << 11) \
```
dholmes-ora, line 53: Nit: we should consistently use either `x` or `is_x` for `x` in `overpass, caller_sensitive, hidden, scoped, ...`

src/hotspot/share/oops/method.cpp
```diff
@@ -736,24 +731,27 @@ bool Method::compute_has_loops_flag() {
       case Bytecodes::_if_acmpne:
       case Bytecodes::_goto:
       case Bytecodes::_jsr:
-        if (bcs.dest() < bcs.next_bci()) _access_flags.set_has_loops();
+        if (bcs.dest() < bcs.next_bci()) {
+          return set_has_loops();
```
dholmes-ora, line 735: I don't understand the new return logic here. The break gets us out of the switch but we are still in the while loop, but the return takes us all the way out. ???

src/hotspot/share/oops/method.hpp
```diff
@@ -820,78 +795,29 @@ class Method : public Metadata {
...
-    return (_flags & _hidden) != 0;
-  }
+  bool jvmti_mount_transition() const { return constMethod()->jvmti_mount_transition(); }
+  void set_jvmti_mount_transition() { constMethod()->set_jvmti_mount_transition(); }
 
-  void set_hidden(bool x) {
-    _flags = x ? (_flags | _hidden) : (_flags & ~_hidden);
-  }
+  bool is_hidden() const { return constMethod()->is_hidden(); }
+  void set_hidden() { constMethod()->set_is_hidden(); }
```
dholmes-ora, line 808: The naming is inconsistent here regarding `is` - either both should have it or neither IMO.

src/hotspot/share/oops/method.hpp
```diff
@@ -939,24 +865,16 @@ class Method : public Metadata {
...
+  void clear_not_c1_compilable()              {       set_not_c1_compilable(false); }
+  void clear_not_c2_compilable()              {       set_not_c2_compilable(false); }
 
-  bool    is_not_c1_osr_compilable() const    { return is_not_c1_compilable(); }  // don't waste an accessFlags bit
-  void   set_not_c1_osr_compilable()          {       set_not_c1_compilable(); }  // don't waste an accessFlags bit
-  void clear_not_c1_osr_compilable()          {     clear_not_c1_compilable(); }  // don't waste an accessFlags bit
-  bool   is_not_c2_osr_compilable() const     { return access_flags().is_not_c2_osr_compilable();  }
-  void  set_not_c2_osr_compilable()           {       _access_flags.set_not_c2_osr_compilable();   }
-  void clear_not_c2_osr_compilable()          {       _access_flags.clear_not_c2_osr_compilable(); }
+  bool    is_not_c1_osr_compilable() const    { return not_c1_compilable(); }  // don't waste a flags bit
```
dholmes-ora, line 871: Again inconsistent naming with `is` but also `osr` in this case. I would expect being compilable to be quite different to being "osr compilable".

dholmes-ora (commented):
General idea is good but I have some issues with naming inconsistencies. A few other queries.

Thanks.

src/hotspot/share/oops/method.hpp
dholmes-ora: Ah I get this bit now - but the comment didn't make it clear. For C1 compilable and osr-compilable are considered the same.

src/hotspot/share/oops/method.cpp
coleenp: Yes, I had a version of this change where we only let has_loops get set once, and this code set it over and over again, which is unnecessary. Once it's true, it stays true.

src/hotspot/share/classfile/classFileParser.cpp
coleenp: This hidden flag is set by both the attribute during class file parsing and set to true if the whole class is hidden.  My original version of the assert_is_safe() function only allowed a flag to be set once, but I've changed it to only set to true.  So I can undo this.

src/hotspot/share/oops/constMethodFlags.hpp
coleenp: That's my preference too, but I was trying to not change all the callers to is_caller_sensitive, for example, or providing a set of wrappers to change the "x" calls to "is_x" calls to the flags interface.

src/hotspot/share/oops/method.hpp
coleenp: This one is easy to fix because there aren't calls everywhere.

src/hotspot/share/oops/method.cpp
coleenp: The early break and not resetting has_loops once it's true, saves the atomic access also.

src/hotspot/share/oops/constMethodFlags.hpp
coleenp: The 'is' in not_x_compilable can be added without too much fanout of lines and files that I don't want to change, so I added the 'is' to these.

coleenp:
@dougxc can you look at the JVMCI changes?

src/hotspot/share/oops/method.hpp
```diff
@@ -865,13 +865,14 @@ class Method : public Metadata {
...
+  void clear_is_not_c2_osr_compilable()       { set_is_not_c2_osr_compilable(false); }
 
-  bool    is_not_c1_osr_compilable() const    { return not_c1_compilable(); }  // don't waste a flags bit
-  void   set_not_c1_osr_compilable()          {       set_not_c1_compilable(); }  // don't waste a flags bit
-  void clear_not_c1_osr_compilable()          {     clear_not_c1_compilable(); }  // don't waste a flags bit
-  void clear_not_c2_osr_compilable()          {       set_not_c2_osr_compilable(false); }
+  // not_c1_osr_compilable == not_c1_compilable
+  bool    is_not_c1_osr_compilable() const    { return is_not_c1_compilable(); }
+  void   set_is_not_c1_osr_compilable()       {       set_is_not_c1_compilable(); }
+  void clear_is_not_c1_osr_compilable()       {     clear_is_not_c1_compilable(); }
```
dholmes-ora, line 875: Nit: don't need extra spaces after `{`

dholmes-ora (approved):
Thanks for the updates. I understand about the fanout from making `is` naming fully consistent.

dougxc (approved):

src/hotspot/share/oops/method.hpp
```diff
@@ -602,31 +589,30 @@ class Method : public Metadata {
...
 
   bool compute_has_loops_flag();
-
-  bool has_jsrs() {
-    return access_flags().has_jsrs();
-  };
-  void set_has_jsrs() {
-    _access_flags.set_has_jsrs();
-  }
+  bool set_has_loops() { set_has_loops_flag(); set_has_loops_flag_init(); return true; }
```
matias9927, line 606: Since this has multiple statements it should probably be on different lines

src/hotspot/share/oops/method.hpp
```diff
@@ -602,31 +589,30 @@ class Method : public Metadata {
...
+  bool has_monitors() const                      { return is_synchronized() || has_monitor_bytecodes(); }
 
   // monitor matching. This returns a conservative estimate of whether the monitorenter/monitorexit bytecodes
-  // propererly nest in the method. It might return false, even though they actually nest properly, since the info.
+  // properly nest in the method. It might return false, even though they actually nest properly, since the info.
   // has not been computed yet.
-  bool guaranteed_monitor_matching() const       { return access_flags().is_monitor_matching(); }
-  void set_guaranteed_monitor_matching()         { _access_flags.set_monitor_matching(); }
+  bool guaranteed_monitor_matching() const       { return monitor_matching(); }
+  void set_guaranteed_monitor_matching()         { set_monitor_matching(); }
```
matias9927, line 615: Is this method just obsolete now? If so it might be worth replacing the callers with `set_monitor_matching()` unless `set_monitor_matching()` is still meant to be private.

matias9927 (changes requested):
Nice change! Just some small nits but it otherwise looks good.

src/hotspot/share/oops/method.hpp
coleenp: The reason I left that was to anchor the comment.  There is nowhere good to put that in the X macro.  Also, didn't want to fix the callers.  It's a good point about making monitor_matching() private, but also not really doable with the X macro.  So that's why I left it.

src/hotspot/share/oops/constMethod.cpp
```diff
@@ -436,6 +435,7 @@ void ConstMethod::print_on(outputStream* st) const {
     m->print_value_on(st);
   }
   st->cr();
+  st->print(" - flags:             "); _flags.print_on(st);               st->cr();
```
fparain, line 438: Method prints its flags as an int and in decoded form, but ConstMethod only prints the decoded form. Any particular reason for this difference?

coleenp: No reason for this difference.  The only reason I added the int form for MethodFlags was because AccessFlags were printed also with the int form.

matias9927 (approved):
Looks good, thanks!

src/hotspot/share/oops/constMethod.cpp
coleenp: I fixed the constMethod printing with the last commit.

coleenp:
Thanks David, Chris, Doug, Matias and Fred for reviewing.
/integrate

PR #13656, opened 2023-04-25
8194743: Compiler implementation for Statements before super()

archiecobbs:
This is a first draft of a patch for JEP 447.

Summary of changes:

1. Track when we're within a constructor "prologue" via new flag `AttrContext.ctorPrologue`
1. Add checks for illegal early access to `this` in constructor prologues, and update existing checks to distinguish between static context vs. constructor prologue context
1. Verify allowed placement of `super()`/`this()` calls via new method `Check.checkSuperInitCalls()`
1. Remove/refactor assumptions in several places that `super()`/`this()` was always the first statement

The changes in `Flow.java` are an example of <span>#</span>4. `Flow.FlowAnalyzer` checks for uncaught checked exceptions. For initializer blocks, this was previously done by requiring that any checked exceptions thrown be declared as thrown by all constructors containing `super()`. This list of checked exceptions was being pre-calculated before recursing into the initial constructors. This worked because initializer blocks were executed at the beginning of each initial constructor right after `super()` is called.

Now initializer blocks are traversed as each `super()` invocation is encountered, reflecting what actually happens at runtime. Similarly, final fields are marked as DA after encountering `this()`, not automatically at the beginning of those  constructors. These changes produce equivalent checks, but are compatible with the new flexibility of placement of `super()`/`this()` as well as possible future changes that could occur along these same lines.

---------
### Progress
- [x] Change requires CSR request [JDK-8302041](https://bugs.openjdk.org/browse/JDK-8302041) to be approved

### Issues
 * [JDK-8194743](https://bugs.openjdk.org/browse/JDK-8194743): Compiler implementation for Statements before super() (**Sub-task** - P4)
 * [JDK-8302041](https://bugs.openjdk.org/browse/JDK-8302041): Permit additional statements before this/super in constructors (**CSR**)

### Reviewers
 * [Vicente Romero](https://openjdk.org/census#vromero) (@vicente-romero-oracle - **Reviewer**) ⚠️ Review applies to [e2f88137](https://git.openjdk.org/jdk/pull/13656/files/e2f881373e8e01309b8d34802b4d13b4b8bf9efe)
 * [Julian Waters](https://openjdk.org/census#jwaters) (@TheShermanTanker - Committer) ⚠️ Review applies to [e2f88137](https://git.openjdk.org/jdk/pull/13656/files/e2f881373e8e01309b8d34802b4d13b4b8bf9efe)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13656/head:pull/13656` \
`$ git checkout pull/13656`

`$ git checkout pull/13656` \
`$ git pull https://git.openjdk.org/jdk.git pull/13656/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13656`

`$ git pr show -t 13656`

</details>

<a href="https://git.openjdk.org/jdk/pull/13656.diff">https://git.openjdk.org/jdk/pull/13656.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13656#issuecomment-1522395525)


src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -5606,6 +5588,9 @@ private void attribClassBody(Env<AttrContext> env, ClassSymbol c) {
             }
         }
 
+        // Check for proper placement of super()/init() calls.
```
ExE-Boss, line 5591:
```suggestion
        // Check for proper placement of super()/this() calls.
```

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Flow.java
```diff
@@ -2848,6 +2839,24 @@ public void visitThrow(JCThrow tree) {
...
+            }
+
+            // If this(): at this point all final uninitialized fields will get initialized
+            if (isConstructor && name == names._this) {
+                for (int address = firstadr; address < nextadr; address++) {
+                    VarSymbol sym = vardecls[address].sym;
+                    if (isFinalUninitializedField(sym) && !sym.isStatic())
+                        letInit(tree.pos(), sym);
+                }
+            }
```
ExE-Boss, line 2859: The common `isConstructor` check can be moved to an enclosing `if` statement:

```suggestion
            if (isConstructor) {
                // If super(): at this point all initialization blocks will execute
                Name name = TreeInfo.name(tree.meth);
                if (name == names._super) {
                    forEachInitializer(classDef, false, def -> {
                        scan(def);
                        clearPendingExits(false);
                    });
                }

                // If this(): at this point all final uninitialized fields will get initialized
                else if (name == names._this) {
                    for (int address = firstadr; address < nextadr; address++) {
                        VarSymbol sym = vardecls[address].sym;
                        if (isFinalUninitializedField(sym) && !sym.isStatic())
                            letInit(tree.pos(), sym);
                    }
                }
            }
```

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
archiecobbs: Thanks for catching that - fixed in 0e638da2daa.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Flow.java
archiecobbs: Yep, thanks - applied in 0e638da2daa.

liach:
Quick question, since I don't know much about javac, I wonder what happens if you declare a local class or an anonymous class in the ctor prologue? Will that local/anonymous class have a mandated outer instance parameter in its constructor, and thus having access to a newly allocated but non-initialized object via `ClassName.this`?

archiecobbs:
> what happens if you declare a local class or an anonymous class in the ctor prologue?

You can declare such a class in the prologue, but you can't instantiate it there:
```java
    public static class LocalTest {
        public LocalTest() {
            class Foo {
                Foo() {
                    LocalTest.this.hashCode();
                }
            }
            new Foo();    // ILLEGAL
            super();
            new Foo();    // ALLOWED 
        }
    }       
```

Note that:
* The prologue includes the statements prior to the superclass constructor *and* the parameter expressions within the superclass constructor invocation
* The rules that apply to the prologue are the *same rules* that previously applied only to parameter expressions within the superclass constructor invocation

So if we "backport" your question to JDK 20, you get the current behavior:
```java
import java.util.concurrent.atomic.*;
public class LocalTest extends AtomicReference<Object> {
    public class Inner {
    }
    public LocalTest(int x) {
        super(new Inner());     // ILLEGAL
    }
    public LocalTest(char x) {
        super(null);
        new Inner();            // ALLOWED
    }
}
```

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Source.java
```diff
@@ -238,6 +238,7 @@ public enum Feature {
         UNCONDITIONAL_PATTERN_IN_INSTANCEOF(JDK21, Fragments.FeatureUnconditionalPatternsInInstanceof, DiagKind.PLURAL),
         RECORD_PATTERNS(JDK21, Fragments.FeatureDeconstructionPatterns, DiagKind.PLURAL),
         STRING_TEMPLATES(JDK21, Fragments.FeatureStringTemplates, DiagKind.PLURAL),
+        SUPER_INIT(JDK21, Fragments.FeatureSuperInit, DiagKind.NORMAL),
```
liach, line 241: Now JDK mainline is 22, this should update the target version too.

archiecobbs: Thanks - should be fixed in a5f8cc5e8ff.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -1044,6 +1049,9 @@ public void visitMethodDef(JCMethodDecl tree) {
                 chk.validate(tree.recvparam, newEnv);
             }
 
+            // Is this method a constructor?
+            boolean isConstructor = tree.name == names.init;
```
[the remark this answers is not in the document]
archiecobbs: Thanks. I fixed that example and a few others as well in c6cf80fc621.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
```diff
@@ -3955,6 +3958,128 @@ private void checkCyclicConstructor(JCClassDecl tree, Symbol ctor,
...
+
+        @Override
+        public void visitReturn(JCReturn tree) {
+            if (constructor && initCall == null && earlyReturn == null)
+                earlyReturn = tree;             // we have seen a return but not (yet) a super()/this()
+            super.visitReturn(tree);
+        }
+
+        @Override
+        public void visitClassDef(JCClassDecl tree) {
```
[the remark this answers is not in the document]
archiecobbs: Yes, that's the intent. For good measure, I just added an explicit test to verify that `super()` inside a lambda is not allowed in b0b13b2ad22.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
```diff
@@ -3955,6 +3958,128 @@ private void checkCyclicConstructor(JCClassDecl tree, Symbol ctor,
...
+ *    - The super()/this() call, if any, must be a top-level statement in the
+ *      constructor, i.e., not nested inside any other statement or block
+ *    - There must be no return statements prior to the super()/this() call
+ **************************************************************************/
+
+    void checkSuperInitCalls(JCClassDecl tree) {
+        new SuperThisChecker().check(tree);
+    }
+
+    private class SuperThisChecker extends TreeScanner {
```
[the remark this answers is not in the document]
archiecobbs: Thanks - fixed in 599d761b309.

vicente-romero-oracle (approved):
looks good to me

TheShermanTanker:
Is this all clear for sponsorship?

liach:
Just curious, is it worth adding a test case where we call `super` multiple times or in a control flow, like
```java
public Example() {
    if (Utilities.condition()) super();
    super();
}
```

TheShermanTanker (approved):

archiecobbs:
> Is this all clear for sponsorship?

Not sure if you're asking me or not, but from my perspective, yes - at least technically speaking.

> Just curious, is it worth adding a test case where we call super multiple times or in a control flow

Happy to add it if you like. That example fails two of the checks [here](https://github.com/archiecobbs/jdk/blob/e2f881373e8e01309b8d34802b4d13b4b8bf9efe/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java#L4043): (a) all `super()` invocations must be "top level", and (b) there can only be one `super()` statement.

archiecobbs:
Regarding the latest update: In some cases the compiler was treating code before `super()` differently from code inside `super()` parameter expressions, although they should be exactly the same (the constructor "prologue" rules apply to both). This is fixed by removing some leftover/redundant code.

In summary, the `isSelfCall` flag is made obsolete by the newer, more general flag `ctorPrologue`, so it can be removed.

liach:
In the latest update to [JEP 401](https://openjdk.org/jeps/401) [Value Objects](https://bugs.openjdk.org/browse/JDK-8251554), it proposes to reuse the pre-initialization context from here for its "regulated Constructors". However, the regulated context anticipates local classes to be static while our pre-initialization context currently declares local classes non-static (except interfaces, records, and enums, which are implicitly static) and thus cannot be used before super call. Should we make pre-initialization local classes static to ease the compiler implementation for "regulated Constructors" in the future?

archiecobbs:
> Should we make pre-initialization local classes static to ease the compiler implementation for "regulated Constructors" in the future?

I could be missing something but I don't see a potential problem.

All of the actions that would be prohibited in regulated constructors will be prohibited in a pre-construction context, so the compiler should already be doing the right thing.

If I understand your question, the particular action you're worried about is "Construction of an inner class with this as an implicit enclosing instance".

But that can't happen in a pre-construction context - you can _declare_ a local class in a pre-construction context, you just can't _instantiate_ it.

Here's an example:
```java
public class Test {

    public int x;

    public Test() {
        class Local {
            { System.out.println(x); }
        }
        //new Local();     // would be illegal
        super();
        new Local();       // this is ok!
    }
}
```
Note, this is a subtlety that didn't exist before, because there was no way to declare a local class that required an outer instance before `super()` and then instantiate it after `super()`. Therefore, there was no reason to allow even the declaration such a class.

liach:
I understand what you mean, since you've clarified long ago. Instead, I mean this from JEP 401:
> Local and anonymous classes may be declared, but (as in a static context) they have no enclosing instance.

Currently local classes declared in pre-construction context have an enclosing instance, and I wonder if we can make them not require enclosing instances like in a static context. Local classes declared after the super/this delegate constructor call in source code will still have an enclosing instance.

I think there are already similar rules implemented for anonymous classes before this JEP, where the anonymous `Cat$1` does not require an implicit `Cat` for its constructor while `Cat$2` does after compilation:
```java
public class Cat extends Dog {
	public Cat() {
		super(new Runnable() {
			@Override
			public void run() {
				System.out.println("Running");
			}
		});

		new Runnable() {
			public void run() {
				System.out.println("Meow");
			}
		}.run();
	}

	public static void main(String... args) {
		new Cat();
	}
}

class Dog {
	Dog(Runnable action) {
		action.run();
	}
}
```

archiecobbs:
> I understand what you mean, since you've clarified long ago. Instead, I mean this from JEP 401:
> 
> > Local and anonymous classes may be declared, but (as in a static context) they have no enclosing instance.
> 
> Currently local classes declared in pre-construction context have an enclosing instance, and I wonder if we can make them not require enclosing instances like in a static context. Local classes declared after the super/this delegate constructor call in source code will still have an enclosing instance.

The problem is that this is not how the compiler currently behaves.

Consider this class:
```java
public class Test {

    public Test(int x) {
    }

    public Test() {
        this(switch (0) {
          default -> {
            class Local {
            }
            yield 0;
          }
        });
    }
}
```
According to JLS 21 §15.9.2, class `Local` does not have an enclosing instance.

Yet here is what's actually emitted by the compiler:
```
$ javap -classpath classes -c Test\$1Local
Compiled from "Test.java"
class Test$1Local {
  final Test this$0;

  Test$1Local(Test);
    Code:
       0: aload_0
       1: aload_1
       2: putfield      #1                  // Field this$0:LTest;
       5: aload_0
       6: invokespecial #7                  // Method java/lang/Object."<init>":()V
       9: return
}
```
An explicit goal of JEP 447 is **any existing program must compile to the same bytecode**. So if we adhere to that goal, local classes declared in pre-construction contexts must have outer instances.

I think this is the most natural behavior anyway. Anonymous classes, with their "immediate enclosing instance with respect to superclass S", and their declare-and-instantiate-all-at-once property, are the oddballs. It's appropriate for them to have a special rule where their implicit outer instance "disappears" in a pre-construction context because it would never be possible to provide one.

So my suggestion is to alter this wording in JEP 401:

> Local and anonymous classes may be declared, but (as in a static context) they have no enclosing instance. Inner classes may refer to enclosing instances or captured enclosing variables from their own regulated constructors without error.
> 
> These rules coincide with the restrictions imposed in a pre-construction context, as described by [JEP 447](https://openjdk.org/jeps/447), except that they allow for writes to instance fields.

To this:

> Local classes may be declared, but since they require an enclosing instance, they can't be instantiated. Anonymous classes may be declared, but (as in a static context) they have no enclosing instance. Inner classes may refer to enclosing instances or captured enclosing variables from their own regulated constructors without error.
> 
> These rules coincide with the restrictions imposed in a pre-construction context, as described by [JEP 447](https://openjdk.org/jeps/447), except that they allow for writes to instance fields.

Also:

> I think there are already similar rules implemented for anonymous classes before this JEP, where the anonymous `Cat$1` (printing "Running") does not require an implicit `Cat` for its constructor while `Cat$2` (printing "Meow") does after compilation:

As remarked above, anonymous classes are different. It's not possible to declare them in one place and instantiate them in another place, so they indeed will not have enclosing instances when declared in a pre-construction context but will have them otherwise, and this behavior is not changing.

liach:
Sounds reasonable. We can always change pre-initialization context local classes in a later patch if Valhalla development deems it necessary, or create other mechanism if Valhalla finds pre-initialization context not a fit.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java
```diff
@@ -4595,6 +4606,32 @@ JCDiagnostic getDiagnostic(JCDiagnostic.DiagnosticType dkind,
         }
     }
 
+    /**
+     * Specialization of {@link InvalidSymbolError} for illegal
+     * early accesses within a constructor prologue.
+     */
+    class RefBeforeCtorCalledError extends InvalidSymbolError {
```
mcimadamore, line 4613: Why doesn't this subclass StaticError?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Flow.java
```diff
@@ -471,6 +479,24 @@ protected void scanSyntheticBreak(TreeMaker make, JCTree swtch) {
                 scan(brk);
             }
         }
+
+        // Do something with all static or non-static field initializers and initialization blocks.
+        // Note: This method also sends nested class definitions to the handler.
+        protected void forEachInitializer(JCClassDecl classDef, boolean statik, Consumer<? super JCTree> handler) {
```
mcimadamore, line 485:
```suggestion
        protected void forEachInitializer(JCClassDecl classDef, boolean isStatic, Consumer<? super JCTree> handler) {
```

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java
archiecobbs: In the new specification, constructor prologues are no longer considered a static context. Instead they're a pre-construction context (new concept). So emitting an error that says something "cannot be referenced from a static context" would be wrong. The new exception class `RefBeforeCtorCalledError` is intended to be the pre-construction context analogue of `StaticError`.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Flow.java
archiecobbs: Fixed in 150d794c1e6, thanks.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java
mcimadamore: I'm aware of that - but I note that the implementation uses a STATIC_ERR under the hood. Hence my comment on using a custom static error with a custom diagnostic.

archiecobbs: Oh I see. Hmm, I'll admit ignorance on how exactly these `Kind` enum error values like `STATICERR`, `HIDDEN`, etc. are used. Is there a better `Kind` choice for this error? Or does this warrant a new `Kind` value? In the latter case, there is a scary warning about ordering of the enum values so any advice appreciated. Thanks.

mcimadamore: The kind is used in some other parts of Resolve to determine whether some errors should be skipped in some places, or recovered from. I'd advise against creating a new error kind, which might end up invalidating some existing check which depend on STATICERR. E.g.:

```
final boolean shouldStop(Symbol sym, MethodResolutionPhase phase) {
            return phase.ordinal() > maxPhase.ordinal() ||
                 !sym.kind.isResolutionError() || sym.kind == AMBIGUOUS || sym.kind == STATICERR;
        }
```

E.g. you don't want overload resolution to try new candidates because you tried to call a method and you don't have a `this` - e.g. the compiler behavior should be similar as to when an instance method is called and no `this` is available (meaning: the found method is correct, and the error should be reported, no attempt should be made at recovery).

mcimadamore: Yes, I think that is what I was referring to. Thanks for the patience.

archiecobbs: Great - fixed in 3e75e19e230. Thanks for the mini-lesson :)

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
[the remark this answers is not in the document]
mcimadamore: Actually, I disagree a bit with the direction here. Having more mutable state is, I believe, not great. In this case the state has to be discarded after a call to `check`, so IMHO having mutable state lying around is a smell. What do you think @vicente-romero-oracle ?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
```diff
@@ -3955,6 +3951,137 @@ private void checkCyclicConstructor(JCClassDecl tree, Symbol ctor,
...
+ **************************************************************************/
+
+    void checkSuperInitCalls(JCClassDecl tree) {
+        superThisChecker.check(tree);
+    }
+
+    private class SuperThisChecker extends TreeScanner {
+
+        // Match this scan stack: 1=JCMethodDecl, 2=JCExpressionStatement, 3=JCMethodInvocation
+        private static final int MATCH_SCAN_DEPTH = 3;
```
mcimadamore, line 3971: This logic seems a bit fragile in that it relies on the shape of the AST. Also, in principle you can create something that will overflow the int, and will cause the check to spuriously pass for a very very very large number of nested expressions :-)

That said, I understand that, in order to make this code "tighter", you would need to override al visitors that can "nest" code inside other code, which is also not great.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java
```diff
@@ -1503,13 +1503,15 @@ Symbol findVar(Env<AttrContext> env, Name name) {
...
+                if (sym.kind == VAR &&
                         sym.owner.kind == TYP &&
-                        (sym.flags() & STATIC) == 0)
-                    return new StaticError(sym);
-                else
-                    return sym;
+                        (sym.flags() & STATIC) == 0) {
+                    if (staticOnly)
+                        return new StaticError(sym);
+                    if (env1.info.ctorPrologue && (sym.flags_field & SYNTHETIC) == 0)
```
mcimadamore, line 1511: where does the SYNTHETIC check comes from?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java
```diff
@@ -2006,11 +2008,15 @@ Symbol findFun(Env<AttrContext> env, Name name,
...
-                        sym.kind == MTH &&
-                        sym.owner.kind == TYP &&
-                        (sym.flags() & STATIC) == 0) return new StaticError(sym);
-                    else return sym;
+                    if (sym.kind == MTH &&
+                            sym.owner.kind == TYP &&
+                            (sym.flags() & STATIC) == 0) {
+                        if (staticOnly)
+                            return new StaticError(sym);
+                        if (env1.info.ctorPrologue && env1 == env)
```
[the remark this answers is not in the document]
mcimadamore: Should this work?
```
class Test {
    String s = "Hello";

    void m() {
        System.out.println("Hello");
    }

    class Inner {
        Inner() {
            m();
            super();
        }
    }

        new Test().new Inner();
    }
}
```
The proposes changes accept this program. But I can't find supporting text in the JLS changes:
https://cr.openjdk.org/~gbierman/jep447/jep447-20230620/specs/statements-before-super-jls.html

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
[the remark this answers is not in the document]
vicente-romero-oracle: yes @mcimadamore I see your point. And after my comments I think we discussed a similar issue in another bug and we went for the non-mutable state. After revisiting the code I agree that in this case it could be better to avoid having mutable state.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java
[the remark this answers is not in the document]
mcimadamore: After some internal discussion, it is good for javac to accept this, as this case can be supported using the `this$0` parameter that is passed into the constructor of `Inner` (no `this` reference is strictly required). But the JLS text supporting this seems ambiguous:

```
The pre-construction context includes the prologue of the constructor 8.8.7 and the explicit constructor invocation itself. Within a pre-construction context, constructs that refer explicitly or implicitly to the current object are disallowed.
```

It is not clear to me as to whether `m()` refers to the current object or not - as it seems one would have to have intimate knowledge of how the compiler translate the code in order to answer that question.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
archiecobbs: Agree about this appearing fragile. The thinking was that, well, the JLS rule for this is simply:

> _ConstructorBody:_
>     { [_BlockStatements_] } 
>     { [_BlockStatements_] _ExplicitConstructorInvocation_ [_BlockStatements_] }

So this tries to mirror exactly that in the most direct way. The fragility comes from the possibility of some unexpected change to the AST class model... but at least if that ever did happen, you would probably get a clear warning when a bunch of JDK code stopped compiling :)

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java
archiecobbs: This is related to the previous question... some generated classes contain constructors with extra blocks, and these fail the `JCMethodDecl, JCExpressionStatement, JCMethodInvocation` stack check test.

src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties
```diff
@@ -234,8 +234,20 @@ compiler.err.switch.expression.no.result.expressions=\
     switch expression does not have any result expressions
 
 # 0: name
-compiler.err.call.must.be.first.stmt.in.ctor=\
-    call to {0} must be first statement in constructor
+compiler.err.call.must.only.appear.in.ctor=\
+    calls to {0}() may only appear within constructors
```
mcimadamore, line 238: Note that super()/this() in the JLS lingo is called "explicit constructor invocation".

mcimadamore: @GavinBierman do you think the compiler message should mention the official JLS terminology?

src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties
```diff
@@ -234,8 +234,20 @@ compiler.err.switch.expression.no.result.expressions=\
...
+
+# 0: name, 1: name
+compiler.err.redundant.superclass.init=\
+    redundant call to {0}() after {1}() has already been invoked
+
+# 0: name
+compiler.err.calls.not.allowed.here=\
```
mcimadamore, line 245: The compiler message key seems overly general. maybe `constructor.call.not.allowed.here` ?

src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties
mcimadamore: For instance:

```
explicit constructor invocation may only appear within constructors
```

```
redundant explicit constructor invocation
```

```
explicit constructor invocation not allowed here
```

```
'return'' not allowed before explicit constructor invocation
```

mcimadamore (commented):
Overall, this is a very good patch. Not only it adds a new feature (which I found myself reaching for several times), but it also does so by consolidating a bunch of logic in the compiler code. Well done!

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java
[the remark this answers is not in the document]
archiecobbs: > where does env1 == env come from?

To be honest I can't remember. At some point it was needed to make something compile, but I can't come up with an example anymore (if you can, I'll tell you why it was there :)

The thinking must have been: "It can't possibly matter if the AST 'cursor'  is in the pre-construction context of an outer instance class; it only matters if we are in the pre-construction context of the class under construction."

> The proposed changes accept this program. But I can't find supporting text in the JLS changes

The spec is still being finalized ([here's the latest version](https://cr.openjdk.org/~gbierman/jep447/jep447-20230905/specs/statements-before-super-jls.html#jls-8.8)). The newer §15.12.3 should allow this.

src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties
archiecobbs: Agreed - will fix.

src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties
```diff
@@ -234,8 +234,20 @@ compiler.err.switch.expression.no.result.expressions=\
...
+    calls to {0}() not allowed here
+
+# 0: name
+compiler.err.return.before.superclass.initialized=\
+    ''return'' not allowed prior to {0}()
```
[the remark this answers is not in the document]
archiecobbs: Agreed - will fix.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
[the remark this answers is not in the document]
archiecobbs: I agree with you both, and will revert this back to creating a new instance each time. Thanks.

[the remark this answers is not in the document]
mcimadamore: Thanks, and sorry for being late to the party!

src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties
GavinBierman: @mcimadamore Yes, I like the first. Something like "explicit constructor invocation may only appear within a constructor body" is good.

(The JLS actually says that an (explicit) constructor invocation, i.e. this(...) or super(...), is not a statement, but just a thing that can appear in a constructor body.)

archiecobbs: Updated error messages in 738c10a57de.

mcimadamore: Good job! I like the new messages.

PR #13183, opened 2023-03-25
8304918: Remove unused decl field from AnnotatedType implementations

liach:
In `AnnotatedTypeBaseImpl`, a `decl` field is declared, referring to the declaration that the AnnotatedType is from. However, this field is not used anywhere except passing to constructors of other implementations; it's not used in debug messages (unlike in `TypeAnnotation`), not involved in equality checks. Since these implementations are not serializable, we can safely remove the decl field from the implementations. All 1048 java/lang tests pass on my machine.

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8304918](https://bugs.openjdk.org/browse/JDK-8304918): Remove unused decl field from AnnotatedType implementations

### Reviewers
 * [Sergey Tsypanov](https://openjdk.org/census#stsypanov) (@stsypanov - Author)
 * [Joe Darcy](https://openjdk.org/census#darcy) (@jddarcy - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13183/head:pull/13183` \
`$ git checkout pull/13183`

Update a local copy of the PR: \
`$ git checkout pull/13183` \
`$ git pull https://git.openjdk.org/jdk.git pull/13183/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 13183`

View PR using the GUI difftool: \
`$ git pr show -t 13183`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jdk/pull/13183.diff">https://git.openjdk.org/jdk/pull/13183.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13183#issuecomment-1483736379)

PR #9862, opened 2022-08-12
8292275: javac does not emit SYNTHETIC and MANDATED flags for parameters by default

SirYwell:
With this change, javac emits the MethodParameters attribute in cases where the JLS requires the information about synthetic and mandated parameters to be stored (see issue).
Parameter names are *not* emitted unless the `-parameter` flag is set.

The relevant changes are in `ClassWriter`, where we go through the params to see if we need the attribute if the `-parameter` flag is not set (if it is set, both names and flags will be emitted).
For records, the mandated flag wasn't set at all, this is solved by the one line fix in `JavacParser`.

The changes to `CreateSymbols` and `ClassReader` are needed as they weren't able to deal with missing names in the attribute.
I also had to update some tests as they got a new constant pool entry.

Only the mandated flag is covered by tests at the moment, as the occurrences are well-specified in the JLS.
Please let me know if you want tests for specific appearances of synthetic parameters.

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change requires CSR request [JDK-8292467](https://bugs.openjdk.org/browse/JDK-8292467) to be approved

### Issues
 * [JDK-8292275](https://bugs.openjdk.org/browse/JDK-8292275): javac does not emit SYNTHETIC and MANDATED flags for parameters by default
 * [JDK-8292467](https://bugs.openjdk.org/browse/JDK-8292467): javac does not emit SYNTHETIC and MANDATED flags for parameters by default (**CSR**)

### Reviewers
 * [Vicente Romero](https://openjdk.org/census#vromero) (@vicente-romero-oracle - **Reviewer**) ⚠️ Review applies to [5ac3a214](https://git.openjdk.org/jdk/pull/9862/files/5ac3a214be729cd34e37d2ad07ad504328d4c1b8)
 * [Julian Waters](https://openjdk.org/census#jwaters) (@TheShermanTanker - Committer)

### Contributors
 * Chen Liang `<liach@openjdk.org>`

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/9862/head:pull/9862` \
`$ git checkout pull/9862`

Update a local copy of the PR: \
`$ git checkout pull/9862` \
[...]

PR #13664, opened 2023-04-26
8180892: Correct handling of annotations on parameters

liach:
This patch aims to correct handling of annotations on parameters with the help of `MethodParameters` attribute, which will be always available once #9862 is integrated.

It utilizes and expands upon the existing parameter matching logic present in `Executable::getAllGenericParameterTypes`, and delegate parameter parameterized types, parameter annotation, and parameter type annotation accesses through the matched results, if matching is available. If matching failed, it falls back to existing heuristics that works without `MethodParameters` attributes for annotations, in `Executable::handleParameterNumberMismatch` and `TypeAnnotationParser::buildAnnotatedTypes` (renamed `buildAnnotatedTypesWithHeuristics` in this patch)

`ParameterMappingTest` covers these scenarios with class files that have `MethodParameters` or `Signature` attributes stripped or preserved to ensure the new Reflection API implementation works for both class files generated before #9862 and after its integration.

Also special thanks to Joe Darcy for reviewing 8304918 (#13183); this brings much convenience to this patch.

---------
### Progress

### Issue
 * [JDK-8180892](https://bugs.openjdk.org/browse/JDK-8180892): Correct handling of annotations on parameters (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13664/head:pull/13664` \
`$ git checkout pull/13664`

`$ git checkout pull/13664` \
`$ git pull https://git.openjdk.org/jdk.git pull/13664/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13664`

`$ git pr show -t 13664`

</details>

<a href="https://git.openjdk.org/jdk/pull/13664.diff">https://git.openjdk.org/jdk/pull/13664.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13664#issuecomment-1522780288)


jddarcy:
@liach , thank you for working on this fix. I've given the changes an initial skim, but will need to do a more thorough review before approving or offering suggestions for changes. My top priority for the next few weeks will be JDK 21 CSRs so it may take a while for me to circle back to this PR.

liach:
Now that JDK 21 is RDP 1, should we take a look at this now?
This patch makes use of MethodParameters information to map (explicit) parameter annotations to formal parameters. In case that information is not available, the code falls back to the existing heuristic that guesses the offset for inner classes.

Two more potential issues with core reflection and annotations/generics include:
1. `getGenericParameterTypes()[i]` usages should be replaced by `getParameters()[i].getParameterizedType()`, which returns the correct generic type if MethodParameters attribute is present
2. There is some issues with `Constructor.getAnnotatedReceiverType()` due to some effectively-static local and anonymous classes (like ones in static methods and initializers); this is tracked at [JDK-8162500](https://bugs.openjdk.org/browse/JDK-8162500). Those classes don't carry enough information about whether they should have an mandated outer instance parameter, such as through a static flag in InnerClasses attribute; might need Javac tweaks to fix that.

src/java.base/share/classes/java/lang/reflect/Executable.java
```diff
@@ -441,20 +400,92 @@ private ParameterData parameterData() {
...
+
+    /**
+     * Converts information from explicit parameters to information about all parameters,
+     * using information from MethodParameters attribute. Guard calls checks to
+     * {@code matchedParameterCount() == explicit.length}.
+     *
+     * @param <T> the type of information
+     * @param explicit information from explicit parameters
+     * @param implicitMapper generates information for implicit parameters
+     * @param freshCopy if true, will copy explicit information on return to avoid leaks
```
bjkail, line 453: Is the `freshCopy` parameter needed?

liach:
Keep alive.

liach:
@jddarcy Joe, can you take a look at this patch where annotation handling on parameters now uses available `MethodParameters` attribute to match from explicit formal parameter indices to implicit formal parameter indices?

jddarcy:
> @jddarcy Joe, can you take a look at this patch where annotation handling on parameters now uses available `MethodParameters` attribute to match from explicit formal parameter indices to implicit formal parameter indices?

@liach sorry for the delays.

I'll be attending to some other matters in the immediate future; please ping me again in a few weeks if I don't reply first.

liach:
Hello Joe @jddarcy, is there anything that I need to do to move this patch forward? Should I file a CSR for the behavior change to `getAnnotatedParameterTypes()` and `getParameters()[i].getAnnotatedType()` due to this bugfix, as it now annotates correctly?

jddarcy:
> Hello Joe @jddarcy, is there anything that I need to do to move this patch forward? Should I file a CSR for the behavior change to `getAnnotatedParameterTypes()` and `getParameters()[i].getAnnotatedType()` due to this bugfix, as it now annotates correctly?

Hi @liach , sorry for the continued delays. I'm currently working through reviewing a backlog of CSRs for JDK 22. This is (still) on my list to get to.

src/java.base/share/classes/java/lang/reflect/Executable.java
liach: Removed unused parameter.

liach:
This patch has been updated to the latest Class-File API. May I request a review again?

liach:
Keep-alive.

PR #13666, opened 2023-04-26
8306541: Refactor collection set candidate handling to prepare for JDK-8140326

tschatzl:
Hi all,

  please review this refactoring of collection set candidate set handling.

The idea is to improve the interface to collection set candidates and prepare for having collection set candidates available at any time to evacuate them at any young collection.

These preparations to allow for multiple sources for these candidates (from the marking, as now, and from retained regions, i.e. evacuation failed regions as per [JDK-8140326](https://bugs.openjdk.org/browse/JDK-8140326)).

This patch only uses candidates from marking at this time.
 
Also moves gc efficiency out of HeapRegion and associate it to the list element as it's not used otherwise.
    
In detail:
    * the collection set candidates set is not temporarily allocated any more, but the candidate collection set object is available all the time.
    
    * G1CollectionSetCandidates is the main class, representing the current candidates. Contains the "from marking" candidate list only (at this point).
    
    * there are several additional helper sets/lists
      * G1CollectionSetRegionList: list of HeapRegion*, typically sorted by efficiency (but not necessarily). Also does not contain gc efficiences.
      * G1CollectionCandidateList: list of candidates, i.e. HeapRegion* with their gc efficiency. Building block for the actual collection set candidates list.
    
All these sets implement C++ iterators for simpler use in various places.
    
Testing:
 - this patch only: tier1-3, gha
 - with JDK-8140326 tier1-7 (or 8?)

Thanks,
  Thomas

---------
### Progress

### Issue
 * [JDK-8306541](https://bugs.openjdk.org/browse/JDK-8306541): Refactor collection set candidate handling to prepare for JDK-8140326

### Reviewers
 * [Ivan Walulya](https://openjdk.org/census#iwalulya) (@walulyai - **Reviewer**) ⚠️ Review applies to [13b6b3c6](https://git.openjdk.org/jdk/pull/13666/files/13b6b3c68ff2d422c278c8bdf408b5dd0b1d7b61)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13666/head:pull/13666` \
`$ git checkout pull/13666`

`$ git checkout pull/13666` \
`$ git pull https://git.openjdk.org/jdk.git pull/13666/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13666`

`$ git pr show -t 13666`

</details>

<a href="https://git.openjdk.org/jdk/pull/13666.diff">https://git.openjdk.org/jdk/pull/13666.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13666#issuecomment-1523123069)


tschatzl:
The first two commits in the branch https://github.com/openjdk/jdk/compare/master...tschatzl:jdk:remove-pinned-mask show this change in context of what is to come for [JDK-8140326](https://bugs.openjdk.org/browse/JDK-8140326)), which is the fourth patch.

src/hotspot/share/gc/g1/heapRegion.inline.hpp
```diff
@@ -324,10 +333,22 @@ inline void HeapRegion::reset_top_at_mark_start() {
...
 
+inline bool HeapRegion::needs_scrubbing() const {
+  return is_old();
+}
+
 inline bool HeapRegion::in_collection_set() const {
   return G1CollectedHeap::heap()->is_in_cset(this);
 }
 
+inline bool HeapRegion::in_collection_set_candidates() const {
```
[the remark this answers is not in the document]
tschatzl: I inlined a few helpers.

src/hotspot/share/gc/g1/g1CollectionSetCandidates.hpp
```diff
@@ -29,72 +29,210 @@
...
-  ~G1CollectionSetCandidates() {
-    FREE_C_HEAP_ARRAY(HeapRegion*, _regions);
-  }
+  G1CollectionSetRegionList();
+
+  // Append a HeapRegion to the end of this list.
+  void append(HeapRegion* r);
+  // Remove the given list of HeapRegion* from this list. Assumes that the given
+  // list is a prefix of this list.
+  void remove(G1CollectionSetRegionList* list);
```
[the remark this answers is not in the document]
tschatzl: I improved the documentation.

src/hotspot/share/gc/g1/g1CollectionSetChooser.cpp
```diff
@@ -187,9 +195,9 @@ class G1BuildCandidateRegionsTask : public WorkerTask {
       // before we fill them up).
       if (should_add(r) && !G1CollectedHeap::heap()->is_old_gc_alloc_region(r)) {
         add_region(r);
-      } else if (r->is_old()) {
-        // Keep remembered sets for humongous regions, otherwise clean out remembered
-        // sets for old regions.
+      } else if (r->is_old() && !r->is_collection_set_candidate()) {
```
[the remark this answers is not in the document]
tschatzl: That is a change that is necessary later - when pinned/evacuation failure regions are part of the candidates, they show up here. Will remove for now. Apologies.

src/hotspot/share/gc/g1/heapRegion.inline.hpp
```diff
@@ -290,14 +294,19 @@ inline void HeapRegion::reset_parsable_bottom() {
 }
 
 inline void HeapRegion::note_start_of_marking() {
-  set_top_at_mark_start(top());
-  _gc_efficiency = -1.0;
+  assert(top_at_mark_start() == bottom(), "CA region's TAMS must always be at bottom");
+  // Retained regions are not going to be scanned through.
+  if (is_old_or_humongous() && !is_collection_set_candidate()) {
+    set_top_at_mark_start(top());
+  }
```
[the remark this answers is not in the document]
tschatzl: Same as above, some change necessary for later.

src/hotspot/share/gc/g1/g1CollectionSetCandidates.hpp
```diff
@@ -29,72 +29,210 @@
...
-  // set candidates.
-  size_t _remaining_reclaimable_bytes;
-  // The index of the next candidate old region to be considered for
-  // addition to the current collection set.
-  uint _front_idx;
+// A set of HeapRegion*. Elements are typically sorted by decreasing gc efficiency
+// but their efficiencies are not stored.
+class G1CollectionSetRegionList {
+  GrowableArray<HeapRegion*> _regions;
+  size_t _reclaimable_bytes;
```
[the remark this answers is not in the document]
tschatzl: In `G1CollectionSetRegionList::remove` you would need to iterate over all elements that are being removed, which is not the case for now.

The other reason is that `reclaimable_bytes` depends on known live bytes in that region. While currently we exclude regions that may change their contents (e.g. current allocation region) from the collection set, I prefer to be absolutely sure that the values that we are working on do not change and the calculations keep being consistent, i.e. snapshotting the (sum of) reclaimable bytes (one could also snapshot the individual values, but I do not see a gain here).

There does not seem to be any other advantage removing this than not having this additional member (i.e. some simplifications this allows).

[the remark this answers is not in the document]
tschatzl: (After deleting the other message)
There is a use in `G1CollectionSetRegionList::remove` where not having this value would add a loop over the `other` list. If you insist, I can change that.

src/hotspot/share/gc/g1/g1CollectionSetChooser.cpp
```diff
@@ -236,12 +244,16 @@ class G1BuildCandidateRegionsTask : public WorkerTask {
...
+  void sort_and_prune_into(G1CollectionSetCandidates* candidates) {
+    _result.sort_by_efficiency();
+
+    uint num_pruned;
+    size_t pruned_wasted_bytes;
+
+    _result.prune(_num_regions_added, num_pruned, pruned_wasted_bytes);
+    candidates->merge_candidates_from_marking(_result.array(),
+                                              _num_regions_added - num_pruned,
+                                              _reclaimable_bytes_added - pruned_wasted_bytes);
```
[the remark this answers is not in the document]
tschatzl: Okay, changed.

src/hotspot/share/gc/g1/g1CollectionSetCandidates.hpp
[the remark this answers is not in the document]
albertnetymk: > would add a loop over the other list

I don't get it.

```
void G1CollectionSetRegionList::remove(G1CollectionSetRegionList* other) {
#ifdef ASSERT
  // Check that the given list is a prefix of this list.
  int i = 0;
  for (HeapRegion* r : *other) {
    assert(_regions.at(i) == r, "must be in order, but element %d is not", i);
    i++;
  }
#endif

  if (other->length() == 0) {
    return;
  }
  _regions.remove_till(other->length());
  _reclaimable_bytes -= other->reclaimable_bytes();
}
```

If one removes `_reclaimable_bytes`, the last statement will go away. Why do you need an additional loop?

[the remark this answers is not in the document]
tschatzl: I stand corrected :) Removed.

src/hotspot/share/gc/g1/g1CollectionSetCandidates.hpp
```diff
@@ -29,72 +29,208 @@
...
+  void clear();
+
+  HeapRegion* at(uint index);
+
+  uint length() const { return (uint)_regions.length(); }
+
+  G1CollectionSetRegionListIterator begin() const { return _regions.begin(); }
+  G1CollectionSetRegionListIterator end() const { return _regions.end(); }
+
+  void verify() PRODUCT_RETURN;
```
albertnetymk, line 65: Seems unimplemented.

src/hotspot/share/gc/g1/g1CollectionSetCandidates.hpp
```diff
@@ -29,72 +29,208 @@
...
+using G1CollectionSetRegionListIterator = GrowableArrayIterator<HeapRegion*>;
 
-  // The sum of bytes that can be reclaimed in the remaining set of collection
-  // set candidates.
-  size_t _remaining_reclaimable_bytes;
-  // The index of the next candidate old region to be considered for
-  // addition to the current collection set.
-  uint _front_idx;
+// A set of HeapRegion*.
+class G1CollectionSetRegionList {
```
albertnetymk, line 43: Now that this is just a region-list, maybe drop the "CollectionSet" part?

src/hotspot/share/gc/g1/g1CollectionSetCandidates.hpp
```diff
@@ -29,72 +29,208 @@
...
+
+  G1CollectionCandidateList _marking_regions;
+
+  CandidateOrigin* _contains_map;
+  uint _max_regions;
+
+  // The number of regions from the last merge of candidates from the marking.
+  uint _last_marking_candidates_length;
+
+  size_t _reclaimable_bytes;
```
albertnetymk, line 181: Where is this used other than asset?

src/hotspot/share/gc/g1/g1CollectionSetCandidates.hpp
```diff
@@ -29,72 +29,208 @@
...
+  // Remove the given regions from the candidates. All given regions must be part
+  // of the candidates.
+  void remove(G1CollectionSetRegionList* other);
+
+  bool contains(const HeapRegion* r) const;
+
+  const char* get_short_type_str(const HeapRegion* r) const;
+
+  bool is_empty() const;
+  bool has_no_more_marking_candidates() const;
```
albertnetymk, line 214: Maybe the positive variant, sth like `has_marking_candidates`?

src/hotspot/share/gc/g1/g1CollectionSetCandidates.cpp
```diff
@@ -26,78 +26,282 @@
...
+  size_t reclaimable_bytes = 0;
+
+  CandidateOrigin* verify_map = NEW_C_HEAP_ARRAY(CandidateOrigin, _max_regions, mtGC);
+  for (uint i = 0; i < _max_regions; i++) {
+    verify_map[i] = CandidateOrigin::Invalid;
+  }
+
+  verify_helper(&_marking_regions, from_marking, reclaimable_bytes, verify_map);
+
+  assert(length() >= marking_regions_length(), "must be");
```
albertnetymk, line 274: Don't get what the intention is here, given `uint length() const { return marking_regions_length(); }`.

walulyai (approved):
Lgtm!

src/hotspot/share/gc/g1/g1CollectionSetCandidates.hpp
tschatzl: I would like to keep the name as is and avoid generalizations that are unnecessary at this time. If there is additional use for it, we can always factor it out.

src/hotspot/share/gc/g1/g1CollectionSetCandidates.hpp
```diff
@@ -29,72 +29,199 @@
...
+  CandidateOrigin* _contains_map;
+  uint _max_regions;
 
-  // Iterate over all remaining collection set candidate regions.
-  void iterate(HeapRegionClosure* cl);
-  // Iterate over all remaining collection set candidate regions from the end
-  // to the beginning of the set.
-  void iterate_backwards(HeapRegionClosure* cl);
+  // The number of regions from the last merge of candidates from the marking.
+  uint _last_marking_candidates_length;
```
albertnetymk, line 173: Looking at how it is used, I wonder if we can record `min_old_cset_length`, which is what is actually needed.

src/hotspot/share/gc/g1/g1CollectionSet.hpp
```diff
@@ -150,11 +146,13 @@ class G1CollectionSet {
   volatile uint _collection_set_cur_length;
   uint _collection_set_max_length;
 
+  uint _eden_region_length;
+  uint _survivor_region_length;
+
+  G1CollectionSetRegionList _initial_old_regions;
```
albertnetymk, line 152: Why is the whole list saved in the field? I'd expect initial-old-regions is a transient list used to move regions from candidate list to cset (live only inside `G1CollectionSet::finalize_old_part`).

`_initial_old_regions` and `_optional_old_regions` share some similarity on  the name, but semantically, it's closer to eden/survior regions, so sth like `uint _initial_old_region_length;`.

src/hotspot/share/gc/g1/g1ParScanThreadState.cpp
```diff
@@ -59,10 +59,10 @@ G1ParScanThreadState::G1ParScanThreadState(G1CollectedHeap* g1h,
                                            PreservedMarks* preserved_marks,
                                            uint worker_id,
                                            uint num_workers,
-                                           size_t young_cset_length,
-                                           size_t optional_cset_length,
+                                           G1CollectionSet* collection_set,
                                            G1EvacFailureRegions* evac_failure_regions)
   : _g1h(g1h),
+    _collection_set(collection_set),
```
albertnetymk, line 65: Why is this needed?

src/hotspot/share/gc/g1/g1ParScanThreadState.cpp
```diff
@@ -690,24 +691,22 @@ void G1ParScanThreadState::update_numa_stats(uint node_index) {
 
 G1ParScanThreadStateSet::G1ParScanThreadStateSet(G1CollectedHeap* g1h,
                                                  uint num_workers,
-                                                 size_t young_cset_length,
-                                                 size_t optional_cset_length,
+                                                 G1CollectionSet* collection_set,
                                                  G1EvacFailureRegions* evac_failure_regions) :
     _g1h(g1h),
+    _collection_set(collection_set),
```
albertnetymk, line 697: Can't find where this field is used.

src/hotspot/share/gc/g1/g1CollectionSetCandidates.hpp
albertnetymk: It's mostly to avoid confusion. The two even have the same length...

```
G1CollectionCandidateList
G1CollectionSetRegionList
```

src/hotspot/share/gc/g1/g1CollectionSetCandidates.hpp
tschatzl: I would like to defer this suggestion (which is good) as this is an improvement of the existing code which I would like to follow here for this change.

src/hotspot/share/gc/g1/g1CollectionSet.hpp
tschatzl: I do not have a too strong opinion either way, so I'll change it (back).

albertnetymk (approved):

tschatzl:
Thanks @walulyai @albertnetymk for your reviews

/integrate

PR #2261, opened 2021-01-27
8252842: Extend jmap to support parallel heap dump

linzang:
---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed

### Issue
 * [JDK-8252842](https://bugs.openjdk.java.net/browse/JDK-8252842): Extend JMap to support parallel heap dump

### Reviewers
 * [Ralf Schmelter](https://openjdk.java.net/census#rschmelter) (@schmelter-sap - Committer)
 * [Chris Plummer](https://openjdk.java.net/census#cjplummer) (@plummercj - **Reviewer**) ⚠️ Review applies to 6d8e944f3ec7cf1a6870f2f59e4433fcd8252946

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk pull/2261/head:pull/2261` \
`$ git checkout pull/2261`

Update a local copy of the PR: \
`$ git checkout pull/2261` \
`$ git pull https://git.openjdk.java.net/jdk pull/2261/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 2261`

View PR using the GUI difftool: \
`$ git pr show -t 2261`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.java.net/jdk/pull/2261.diff">https://git.openjdk.java.net/jdk/pull/2261.diff</a>

</details>

PR #13667, opened 2023-04-26
JDK-8306441: Two phase segmented heap dump

y1yang0:
### Motivation and proposal
Hi, heap dump brings about pauses for application's execution(STW), this is a well-known pain. JDK-8252842 have added parallel support to heapdump in an attempt to alleviate this issue. However, all concurrent threads competitively write heap data to the same file, and more memory is required to maintain the concurrent buffer queue. In experiments, we did not feel a significant performance improvement from that.

The minor-pause solution, which is presented in this PR, is a two-phase segmented heap dump:

- Phase 1(STW): Concurrent threads directly write data to multiple heap files.
- Phase 2(Non-STW): Merge multiple heap files into one complete heap dump file. This process can happen outside safepoint.

Now concurrent worker threads are not required to maintain a buffer queue, which would result in more memory overhead, nor do they need to compete for locks. The changes in the overall design are as follows:

image
<p align="center">Fig1. Before</p>

image
<p align="center">Fig2. After this patch</p>

### Performance evaluation
| memory | numOfThread | CompressionMode | STW | Total |
| -------| ----------- | --------------- | --- | ---- |
| 8g | 1 T | N | 15.612 | 15.612 |
| 8g | 32 T | N | 2.561725 | 14.498 |
| 8g | 32 T | C1 | 2.3084878 | 14.198 |
| 8g | 32 T | C2 | 10.9355128 | 21.882 |
| 8g | 96 T | N | 2.6790452 | 14.012 |
| 8g | 96 T | C1 | 2.3044796 | 3.589 |
| 8g | 96 T | C2 | 9.7585151 | 20.219 |
| 16g | 1 T | N | 26.278 | 26.278 |
| 16g | 32 T | N | 5.231374 | 26.417 |
| 16g | 32 T | C1 | 5.6946983 | 6.538 |
| 16g | 32 T | C2 | 21.8211105 | 41.133 |
| 16g | 96 T | N | 6.2445556 | 27.141 |
| 16g | 96 T | C1 | 4.6007096 | 6.259 |
| 16g | 96 T | C2 | 19.2965783 | 39.007 |
| 32g | 1 T | N | 48.149 | 48.149 |
| 32g | 32 T | N | 10.7734677 | 61.643 |
| 32g | 32 T | C1 | 10.1642097 | 10.903 |
| 32g | 32 T | C2 | 43.8407607 | 88.152 |
| 32g | 96 T | N | 13.1522042 | 61.432 |
| 32g | 96 T | C1 | 9.0954641 | 9.885 |
| 32g | 96 T | C2 | 38.9900931 | 80.574 |
| 64g | 1 T | N | 100.583 | 100.583 |
| 64g | 32 T | N | 20.9233744 | 134.701 |
| 64g | 32 T | C1 | 18.5023784 | 19.358 |
| 64g | 32 T | C2 | 86.4748377 | 172.707 |
| 64g | 96 T | N | 26.7374116 | 126.08 |
| 64g | 96 T | C1 | 16.8101551 | 17.938 |
| 64g | 96 T | C2 | 80.1626621 | 169.003 |
| 128g | 1 T | N | 233.843 | 233.843 |
| 128g | 32 T | N | 72.9945768 | 207.06 |
| 128g | 32 T | C1 | 36.399436 | 37.005 |
| 128g | 32 T | C2 | 172.8942958 | 394.026 |
| 128g | 96 T | N | 67.6815929 | 336.345 |
| 128g | 96 T | C1 | 35.2457306 | 36.23 |
| 128g | 96 T | C2 | 162.2924705 | 397.392 |
<p align="center">Table1. heap dump generation(full)</p>

| memory | numOfThread | STW | Total |
| ------ | ----------- | --- | ----- |
| 8g | 1 T | 15.612 | 15.612 |
| 8g | 32 T | 2.561725 | 14.498 |
| 8g | 96 T | 2.6790452 | 14.012 |
| 16g | 1 T | 26.278 | 26.278 |
| 16g | 32 T | 5.231374 | 26.417 |
| 16g | 96 T | 6.2445556 | 27.141 |
| 32g | 1 T | 48.149 | 48.149 |
| 32g | 32 T | 10.7734677 | 61.643 |
| 32g | 96 T | 13.1522042 | 61.432 |
| 64g | 1 T | 100.583 | 100.583 |
| 64g | 32 T | 20.9233744 | 134.701 |
| 64g | 96 T | 26.7374116 | 126.08 |
| 128g | 1 T | 233.843 | 233.843 |
| 128g | 32 T | 72.9945768 | 207.06 |
| 128g | 96 T | 67.6815929 | 336.345 |
<p align="center">Table2. heap dump generation(w/o compression)</p>

> N.B. all compression benchmark data are based on -gz=9, i.e. strongest compression
> - **N** means no compression
> - **C1** means all objects are byte[] in heap and they are all empty
> - **C2** means all objects are byte[] in heap and they are full of random data
> - **Total** means the total heap dump including both two phases
> - **STW** means the first phase only.
> - For parallel dump, **Total** = **STW** + **Merge**. For serial dump, **Total** = **STW**

image

<p align="center">Fig3. heap dump generation(full)</p>

image
<p align="center">Fig4. heap dump generation(w/o compression)</p>

In a nutshell, this patch significantly **reduces 71~83% application pause time**.

When compression is enabled, STW/Total time heavily depends on the sparseness of the application heap. If the heap is full of compressible objects(e.g. all objects are empty byte array), Total ≈ STW, merge process is incredibly fast. If the heap data is not suitable for compression(e.g. all objects are full of random data), the STW reduction is not appealing, the total dump time is also increased.

In actual testing, two-phase solution can lead to an increase in the overall time for heapdump(See table above). However, considering the reduction of STW time, I think it is an acceptable trade-off. Furthermore, there is still room for optimization in the second merge phase(e.g. sendfile/splice/copy_file_range instead of read+write combination). Since number of parallel dump thread has a considerable impact on total dump time, I added a parameter that allows users to specify the number of parallel dump thread they wish to run.

Testing: manual tests, MAT verification of heapdump integrity, all jtreg tests

### Future work
- Add parallel heap dump support for HeapDumpBeforeFullGC,HeapDumpOnOutOfMemoryError etc
- Add parallel heap dump during performing JMX HotSpotDiagnosticMXBean.dumpHeap
- Merge segmented heap dump files by spawning new virtual thread
- Optimization of file merging by send_file instead of read-write combination
- Add parallel option for jmap
- Refactor, including remove AbstractDumpWriter etc
- [8311775: [TEST] duplicate verifyHeapDump in several tests](https://bugs.openjdk.org/browse/JDK-8311775)

### Open discussion

- Pauseless heap dump solution?
An alternative pauseless solution is to fork a child process, set the parent process heap to read-only, and dump the heap in child process. Once writing happens in parent process, child process observes them by userfaultfd and corresponding pages are prioritized for dumping. I'm also looking forward to hearing comments and discussions about this solution.

- Client parser support for segmented heap dump
This patch provides a possibility that whether heap dump needs to be complete or not, can the VM directly generate segmented heapdump, and let the client parser complete the merge process? Looking forward to hearing comments from the Eclipse MAT community

---------
### Progress

### Issue
 * [JDK-8306441](https://bugs.openjdk.org/browse/JDK-8306441): Two phase segmented heap dump (**Enhancement** - P4)

### Reviewers
 * [Alex Menkov](https://openjdk.org/census#amenkov) (@alexmenkov - **Reviewer**) ⚠️ Review applies to [7cbb4a10](https://git.openjdk.org/jdk/pull/13667/files/7cbb4a103d8f8995550ad1a4d9d38a7ea02f529b)

### Contributors
 * Kevin Walls `<kevinw@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13667/head:pull/13667` \
`$ git checkout pull/13667`

`$ git checkout pull/13667` \
`$ git pull https://git.openjdk.org/jdk.git pull/13667/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13667`

`$ git pr show -t 13667`

</details>

<a href="https://git.openjdk.org/jdk/pull/13667.diff">https://git.openjdk.org/jdk/pull/13667.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13667#issuecomment-1523143363)


y1yang0:
Hi, can I have a review fot this? It significantly reduces heapdump STW time. Thanks.

y1yang0:
Hi, can I have a review for this patch?

plummercj:
@y1yang0 Sorry no one has been able to review this so far. The serviceability team is very busy for the next few weeks finishing up JDK 21 changes before RDP1. It's unlikely we'll find time for the review before them.

I did take a very quick look at the changes just to understand the scope. One thing I noticed that makes this PR hard to review is the code refactoring and relocating that you did. At first it looks like a lot of old code was deleted and a lot of new code added, but in fact most of the new code is just relocated old code. It makes it very hard to tell if there have been any changes to this code. Is there anything you can do to lessen the amount of apparent new code that is actually just moved code?

y1yang0:
> @y1yang0 Sorry no one has been able to review this so far. The serviceability team is very busy for the next few weeks finishing up JDK 21 changes before RDP1. It's unlikely we'll find time for the review before them.
> 
> I did take a very quick look at the changes just to understand the scope. One thing I noticed that makes this PR hard to review is the code refactoring and relocating that you did. At first it looks like a lot of old code was deleted and a lot of new code added, but in fact most of the new code is just relocated old code. It makes it very hard to tell if there have been any changes to this code. Is there anything you can do to lessen the amount of apparent new code that is actually just moved code?

Hi @plummercj Thanks for your heads-up, I saw RPD1 ends at 06/08, I will ping everyone again in mid June and hope to have reviews when you have time.

The refactoring is necessary because
- All related code about ParDumpWriter is unused after this patch
- In `heapDumperCompression` file, most of the code is used for writing heap dump rather than compression(after applying this patch). Therefore, I renamed it to `heapDumperWriter`. However, considering that this may make the review process difficult, I will revert this change and keep the original file name for now. When the review process reaches a proper time point, I will rename it again.

kevinjwalls:
Just a few notes for now:

I was misled by the title, as we already have the concept of segmented heap dumps (HPROF_HEAP_DUMP_SEGMENT).  This is an enhanced parallel heap dump, could we use that (or similar) for the bug and PR title?

I'm interested in if it is faster in STW time with the parallel write to use compression or not?  (Does not compressing take time, or does compressing mean we write fewer bytes, so ends up being faster.)

The idea of forking a child to do the write seems appealing, although it is a non-Windows solution. Isn't the point of a fork that the parent can keep running without any pause at all?  The child gets a copy of all memory, the fork happened at a safepoint so the Java heap is in a good state, and it can take its time to do the write, then exit without affecting the parent.  The child probably must not make any Java calls or interact with other existing JVM threads.
(fork not vfork.  The child is making no writes to the Java heap so should not cause the entire Java heap to be duplicated in memory.  That overhead might be large though particularly if the app is making changes...)

y1yang0:
> I'm interested in if it is faster in STW time with the parallel write to use compression or not? (Does not compressing take time, or does compressing mean we write fewer bytes, so ends up being faster.)

Hi @kevinjwalls, so far all benchmark data are based on parallel dumps without compression. I will add more benchmark data about parallel dump data with compression later.
 
> The idea of forking a child to do the write seems appealing, although it is a non-Windows solution. Isn't the point of a fork that the parent can keep running without any pause at all? The child gets a copy of all memory, the fork happened at a safepoint so the Java heap is in a good state, and it can take its time to do the write, then exit without affecting the parent. The child probably must not make any Java calls or interact with other existing JVM threads.
(fork not vfork. The child is making no writes to the Java heap so should not cause the entire Java heap to be duplicated in memory. That overhead might be large though particularly if the app is making changes...)

I think the real pauseless is impossible, long time STW is amortized by minor STWs as once writing happens in parent process, child process observes them by userfaultfd and corresponding pages are prioritized for dumping in safepoint region.

y1yang0:
> I'm interested in if it is faster in STW time with the parallel write to use compression or not? (Does not compressing take time, or does compressing mean we write fewer bytes, so ends up being faster.)

I updated the benchmark data when turning compression on. In general, this patch significantly reduces 73~80% application pause time. When compression is enabled, STW/Total time heavily depends on the sparseness of the application heap. If the heap is full of compressible(e.g. all objects are empty byte array), Total ≈ STW, the merge process is incredibly fast. If the heap data is not suitable for compression(e.g. all objects are full of random data), the STW reduction is not appealing, the total dump time is also increased.

alexmenkov:
A lot of code movement is caused by moving AbstractDumpWriter class from heapDumper.cpp to heapDumpCompression.hpp/cpp
I'm not happy with huge heapDumper.cpp file, but this refactoring does not look good to me.
Currently all logic about hprof file structure is in heapDumper.cpp, it's not visible outside.
With your fix the logic is spread in heapDumper.hpp/.cpp and heapDumpCompression.hpp/cpp

y1yang0:
> A lot of code movement is caused by moving AbstractDumpWriter class from heapDumper.cpp to heapDumpCompression.hpp/cpp
> I'm not happy with huge heapDumper.cpp file, but this refactoring does not look good to me.
> Currently all logic about hprof file structure is in heapDumper.cpp, it's not visible outside.
> With your fix the logic is spread in heapDumper.hpp/.cpp and heapDumpCompression.hpp/cpp

I will undo all refactor code. Let us firstly focuse on proposal itself. The refactor can be discussed later.

y1yang0:
Hi @plummercj @kevinjwalls @alexmenkov, sorry to disturb you. RDP1 has ended. Do you have any plans to review this patch recently? Thanks!

alexmenkov (commented):
It looks good in general. I added some change requests.

alexmenkov:
Could you please add info about what testing have been done

src/hotspot/share/services/diagnosticCommand.cpp
```diff
@@ -470,11 +470,14 @@ HeapDumpDCmd::HeapDumpDCmd(outputStream* output, bool heap) :
                "using the given compression level. 1 (recommended) is the fastest, "
                "9 the strongest compression.", "INT", false, "1"),
   _overwrite("-overwrite", "If specified, the dump file will be overwritten if it exists",
-           "BOOLEAN", false, "false") {
+           "BOOLEAN", false, "false"),
+  _parallel("-parallel", "Number of parallel dump thread, it should be less than ParallelGCThread",
```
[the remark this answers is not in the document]
alexmenkov: This change adds new option. most likely this requires CSR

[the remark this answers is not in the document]
y1yang0: > This change adds new option. most likely this requires CSR

Hi @alexmenkov, it seems that adding new parameters to jcmd does not require a CSR, only changes like those to jmap do.

y1yang0:
Hi @alexmenkov,

> Could you please add info about what testing have been done

Including serviceability tests and manual checks for both slowdebug and release version.  I'll do more test and sync test result in the main content of this PR.

I push a new commit to optimize the merge stage.  Now it performs heapdump file merge operation in attach listener thread when possible, which prevents us from occupying the VM Thread, which in turn affects the occurrence of GC and other VM operations.

y1yang0:
Hi @plummercj @kevinjwalls @alexmenkov   do you have any plans to review this patch recently? I have addressed some bugs and it passed manual tests& all jtreg tests.

Thanks.

alexmenkov:
I think commit "use HandshakeClosure instead of VMOperation" is a wrong way to go.
It restricts use of parallel dumping only to attach case.
I have pending changes in heap dumper to support virtual threads and I'm going switch heap dumper to always use DumpMerger.

y1yang0:
Hi @alexmenkov,

> It restricts use of parallel dumping only to attach case.

I'm not sure what you mean. Using handshake won't limit it to only attach cases; it can also be used in other cases like HeapDumpBeforeFullGC. The only difference is that previously, VMThread merged the dump files, and now it's the Attach listener that merges them. For the file merging process, my candidate options are as follows:

1. Execute with VMThread outside the safepoint, which will block VMThread from executing other vmoperations.
2. Execute with Attach listener thread outside the safepoint, which will block Attach listener from processing requests like jcmd/jstack.
3. Create a new temporary thread to execute the file merging outside the safepoint, which will have some resource consumption.

I don't have a strong motivation to use 2, and 3 may be a good solution with the availability of virtual threads. If you have any concerns, we can consider using the most conservative option 1 to simplify the review process, and then optimize the file merging process in a follow-up patch.

alexmenkov:
> Hi @alexmenkov,
> 
> > It restricts use of parallel dumping only to attach case.
> 
> I'm not sure what you mean. Using handshake won't limit it to only attach cases; it can also be used in other cases like HeapDumpBeforeFullGC. The only difference is that previously, VMThread merged the dump files, and now it's the Attach listener that merges them. For the file merging process, my candidate options are as follows:
> 
>     1. Execute with VMThread outside the safepoint, which will block VMThread from executing other vmoperations.
> 
>     2. Execute with Attach listener thread outside the safepoint, which will block Attach listener from processing requests like jcmd/jstack.
> 
>     3. Create a new temporary thread to execute the file merging outside the safepoint, which will have some resource consumption.
> 
> 
> I don't have a strong motivation to use 2, and 3 may be a good solution with the availability of virtual threads. If you have any concerns, we can consider using the most conservative option 1 to simplify the review process, and then optimize the file merging process in a follow-up patch.

I mean that you can't be sure that you can use attach listener thread.
My concerns about attach listener thread are:
- AttachListener can be disabled at all or fail to initialize;
- attach listener thread may be not yet available when we need to perform heap dump;
- need to ensure attach listener thread can't be blocked (for example waiting for next command)

I think it makes sense to go option 1 now and add optimizations as follow-up changes (they will be much smaller and easier to review).

alexmenkov (approved):

y1yang0:
Thanks @alexmenkov for the reviews! I added corresponding jtreg for it, also I found verifyHeapDump is duplicated in several tests, I filed https://bugs.openjdk.org/browse/JDK-8311775 as a follow-up test improvement.

@plummercj @kevinjwalls Can I have a second review when you have time? Thanks.

plummercj:
> @plummercj @kevinjwalls Can I have a second review when you have time? Thanks.

I'll have a look.

plummercj:
BTW, `serviceability/dcmd/gc/HeapDumpAllTest.java` is failing due to build failures when compiling the test.

kevinjwalls:
Thanks for adding the graphic in the description above!


There will be some (many) SA test failures due to the new native thread type.
Run a test e.g. test/hotspot/jtreg/serviceability/sa/TestJhsdbJstackLock.java
and it will fail with "Unable to deduce type of thread from address ..." 

"src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/runtime/Threads.java" 
Probably another "virtualConstructor.addMapping"  in Threads::initialize()
Also we need a new SA class for the new AttachListener thread type.

Mixed feelings about using the attach listener for this.

kevinjwalls:
I think this is a good feature, although I have a few other questions that I didn't yet have time to understand:
if we specify a large number for -parallel=, is it limited to the number of safepoint_workers?
Should we feedback to the user how many threads were really used?
Should the user really have to always say how many threads?  Do they just want to say "use all the worker threads"?
Does UseDynamicNumberOfGCThreads affect this?
Does this work with Serial GC? Maybe it doesn't maybe that's unfortunate (Serial may be appropriate for some large heaps, and you still want a faster heap dump?)

dholmes-ora:
I think this needs further scrutiny - not sure why I haven't noticed this one before now. I'm unclear why we need an explicit AttachListenerThread type to support this.

tstuefe:
Hi, @y1yang0!

Interesting idea! But I have some questions I could not answer by reading through your description:

I don't understand the performance numbers. I assume they correlate with the Y axis of the diagrams? Are these seconds? Of what, the new solution? Or are these percentages? You summarize with "reduces 71~83% application pause time". How do you recon that? You also talk about parallel and serial dump. Does that mean parallel = your patch, serial = stock? 

Bottomline, unless I missed it, it would be nice to see how your patched VM stacks up against the existing heap dump mechanism to see if the effort is worth it.

The design: I understand that you write physically separate files, right? Why do we then, in the second phase, need an AttachListener thread to merge them? Why a VM operation? Could you please explain the merge process in more detail?

>Pauseless heap dump solution?
>An alternative pauseless solution is to fork a child process, set the parent process heap to read-only, and dump the heap in child process. Once writing happens in parent process, child process observes them by userfaultfd and corresponding pages are prioritized for dumping. I'm also looking forward to hearing comments and discussions about this solution.

I assume the whole complexity about setting the heap readonly and `userfaultfd` etc is because you plan to use `vfork(2)`? I would advise against it for many reasons, mainly because getting a bug-free, safe solution will be a pain, and errors will be fiendishly difficult to solve.

That leaves us with a solution based on standard `fork(2)`, which is maybe possible. There are cons, though:

- Since this heapdumper optimization is targeted toward processes with very large heaps, even cloning the page tables may be too much. And an active parent may lead to more COW activity, which you must pay for with CPU and memory. Heapdumps are often done in low-resource situations (e.g. as a last resort for post-crash analysis before killing the JVM due to OOM). Paying the extra footprint for the forked process may be too much.

- I fear unforeseen side effects of process forking. Just one example would be the child JVM inheriting all parent file descriptors. Child would keep them open for as long as it needs to take a heap dump. If there are mechanisms that rely on file descriptors being closed, those will timeout. This is certainly solvable, but I am wary of similar effects like these we don't foresee.

Cheers, Thomas

y1yang0:
Thanks for reviews! 

---------
@kevinjwalls 
> if we specify a large number for -parallel=, is it limited to the number of safepoint_workers?

No, it limits how may active_worker is used?

> Should we feedback to the user how many threads were really used?

Yes, I think it makes sense and I will output it in the upcoming changes.

> Should the user really have to always say how many threads? Do they just want to say "use all the worker threads"?

The main purpose of adding the parallel option is that specifying the number of parallel threads can have a certain impact on the time, and using all worker threads may not always produce the best results. Please see the table above, where 96 threads perform worse than 32 threads in many test groups. Additionally, parallel also controls the use of parallel dump or serial dump.

It is also worth noting that even if the user specifies parallel, the VM will try to use this number of threads to perform parallel dump, but this is not guaranteed. The VM will also choose max active_worker based on the size and utilization of the heap.


> Does this work with Serial GC? Maybe it doesn't maybe that's unfortunate (Serial may be appropriate for some large heaps, and you still want a faster heap dump?)

Unfortunately, it does not currently work with Serial GC. For Serial GC, because there is no worker gang concept, the VM always uses serial dump. We have hardly seen any cases of Serial GC in production environments, so I think support for Serial GC can be added in a follow-up PRs.

-------
@dholmes-ora 
> I'm unclear why we need an explicit AttachListenerThread type to support this.

Because I need the overloaded check is_AttachListener_thread(), which can avoid using the VM thread to execute the dump file merge as much as possible.

-------
@tstuefe 
> I don't understand the performance numbers. I assume they correlate with the Y axis of the diagrams? Are these seconds? Of what, the new solution? Or are these percentages? You summarize with "reduces 71~83% application pause time". How do you recon that? You also talk about parallel and serial dump. Does that mean parallel = your patch, serial = stock?

Yes, these are all in seconds. 1T here means serial, which writes to the file with only one thread, and its result is used as the baseline. Other data, such as 32T, are results obtained by testing using my patch. My calculation method compares the change rate obtained by comparing baseline STW and new STW.

dholmes-ora:
> Because I need the overloaded check is_AttachListener_thread(), which can avoid using the VM thread to execute the dump file merge as much as possible.

I'm unclear what the set of candidate threads is for executing the code that does the `is_AttachListener_thread()` test. Can you not just use `!thread->is_VM_thread()` to keep it out of the VMThread?

y1yang0:
> > Because I need the overloaded check is_AttachListener_thread(), which can avoid using the VM thread to execute the dump file merge as much as possible.
> 
> I'm unclear what the set of candidate threads is for executing the code that does the `is_AttachListener_thread()` test. Can you not just use `!thread->is_VM_thread()` to keep it out of the VMThread?

Many types of threads can execute HeapDumper, such as JMX heapdump, GC HeapDumpOnOutOfMemoryError etc. To ensure safety, I want to make it clear that file merging should only be executed using the current thread as an attach listener thread. For other situations, it should be uniformly completed using the VM Thread. In addition, a potential optimization is that if flags such as HeapDumpOnOutOfMemoryError trigger heapdump, the triggerer does not need to wait for the merge to complete on site(because they are not in interactive mode). It only needs to generate separate files during STW and then specify the AttachListenerThread to complete the file merging operation using handshake. This requires us to have an AttachListenerThread class.

For future extension considerations and for code readability and safety, I added an AttachListenerThread class.

src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/runtime/Threads.java
```diff
@@ -157,21 +157,22 @@ private static synchronized void initialize(TypeDataBase db) {
...
 
     public Threads() {
         _list = VMObjectFactory.newObject(ThreadsList.class, threadListField.getValue());
     }
 
     /** NOTE: this returns objects of type JavaThread, CompilerThread,
-      JvmtiAgentThread, NotificationThread, MonitorDeflationThread and ServiceThread.
-      The latter four are subclasses of the former. Most operations
+      JvmtiAgentThread, NotificationThread, AttachListenerThread,
+      MonitorDeflationThread and ServiceThread. Most operations
```
plummercj, line 169: StringDedupThread was previously missed from this comment. Can you fix this also?

I think the "The latter four are subclasses of the former" comment should stay since all the text the follows relates to this. I just needs to be correct to say "seven" instead of "four".

src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/runtime/Threads.java
```diff
@@ -157,21 +157,22 @@ private static synchronized void initialize(TypeDataBase db) {
...
       (fetching the top frame, etc.) are only allowed to be performed on
       a "pure" JavaThread. For this reason, {@link
       sun.jvm.hotspot.runtime.JavaThread#isJavaThread} has been
       changed from the definition in the VM (which returns true for
       all of these thread types) to return true for JavaThreads and
-      false for the four subclasses. FIXME: should reconsider the
+      false for the five subclasses. FIXME: should reconsider the
```
plummercj, line 175: "five" -> "seven"

src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/runtime/Threads.java
```diff
@@ -195,7 +196,7 @@ public JavaThread createJavaThreadWrapper(Address threadAddr) {
             return thread;
         } catch (Exception e) {
             throw new RuntimeException("Unable to deduce type of thread from address " + threadAddr +
-            " (expected type JavaThread, CompilerThread, MonitorDeflationThread, ServiceThread or JvmtiAgentThread)", e);
+            " (expected type JavaThread, CompilerThread, MonitorDeflationThread, AttachListenerThread, ServiceThread or JvmtiAgentThread)", e);
```
plummercj, line 199: Add StringDedupThread thread.

plummercj (changes requested):

src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/runtime/Thread.java
```diff
@@ -85,6 +85,7 @@ public long allocatedBytes() {
   public boolean   isWatcherThread()             { return false; }
   public boolean   isServiceThread()             { return false; }
   public boolean   isMonitorDeflationThread()    { return false; }
+  public boolean   isAttachListenerThread()      { return false; }
```
plummercj, line 88: The copyright needs updating in this file.

y1yang0:
PING: @plummercj @kevinjwalls any further review comments? Thanks.

kevinjwalls:
AttachListener thread:

I was concerned we might occupy this thread for hundreds of seconds == a few minutes for a large dump.
e.g. if you ping the JVM with a jcmd frequently, this will cause a delay/failure.
The parallel dump has twice as much file io happening (not that it will take twice as long).

But I realise that currently heap dumps happen in the Attach Listener, however slow they are.  It may be that slow operations like heap dump and histogram should generally be worked on in another thread, but that is not a new problem.  So we should not do that as part of this change.

I was looking at ServiceThread, which gets woken up with Service_lock and has a serial list of tasks it does.  Thinking about using ServiceThread here for the joining of the dump, the tool could be detached even while the dump was put together ("parallel"!).
But that might have other impact, delaying another ServiceThread task if one is needed.  It would need testing...

I think we might be best off leaving the long dump as you have it in the AttachListener for now. 8-)

plummercj:
> PING: @plummercj @kevinjwalls any further review comments? Thanks.

I don't plan on doing a full review since Alex and Kevin have it covered, so you just need to get Kevin's final approval now.

y1yang0:
> I think we might be best off leaving the long dump as you have it in the AttachListener for now. 8-)

Consideing a very large heap size, the VM used to pause all threads and spend 1 hour to execute a heap dump, during which time jcmd/jstack and even the entire system were unresponsive. Now, the VM spends only 1 minute pausing all threads to execute the heap dump, and then only pauses the attach listener/vm thread and spends 1 hour merging files. For the attach listener, its total time remains unchanged.

kevinjwalls:
Hi, a few more points today, but I'm getting there. 8-)

1
The previous parallel dump change:
8252842: Extend jmap to support parallel heap dump
..the notes suggest it was going to have a CSR for the new option, but was then decided not to add the option to specify the number of threads.
That CSR was https://bugs.openjdk.org/browse/JDK-8260424 which was withdrawn, but looks like we should do one for this change.
If we had a reasonable expectation of what the best number to use was, we could do the same thing here.  If we might get that wrong, we will need the user to be able to specify.
Actually are we changing the default behaviour: from "parallel if you can", to "serial unless you specify".


2
Running the changes, I got a test failure on macos debug (x64 and aarch64):

test/hotspot/jtreg/serviceability/HeapDump/IntegrityHeapDumpTest.java

On different machines, one x64 and one aarch64, got the same kind of error:
----------stderr:(6/609)----------
com.sun.tools.attach.AttachNotSupportedException: Unable to open socket file /var/folders/zz/zyxvpxvq6csfxvn_n0000000000000/T/.java_pid36008: target process 36008 doesn't respond within 10500ms or HotSpot VM not loaded
	at jdk.attach/sun.tools.attach.VirtualMachineImpl.<init>(VirtualMachineImpl.java:95)
	at jdk.attach/sun.tools.attach.AttachProviderImpl.attachVirtualMachine(AttachProviderImpl.java:58)
	at jdk.attach/com.sun.tools.attach.VirtualMachine.attach(VirtualMachine.java:207)
	at jdk.jcmd/sun.tools.jcmd.JCmd.executeCommandForPid(JCmd.java:113)
	at jdk.jcmd/sun.tools.jcmd.JCmd.main(JCmd.java:97)
	
..so the test fails with: java.lang.RuntimeException: Expected to get exit value of [0], exit value is: [1]

Other tests e.g. test/hotspot/jtreg/serviceability/dcmd/gc/HeapDumpTest.java run OK.

The failing test IntegrityHeapDumpTest.java extends LingeredApp.
test/hotspot/jtreg/serviceability/HeapDump/DuplicateArrayClassesTest.java does this, and runs and passes.

I don't yet understand the failure.  Could things have gone wrong in macosx debug builds in the attach listener.. I didn't get a stack of the LingeredApp under test, but will try again.


3
Docs
https://openjdk.org/guide/#release-notes
This enhancement should probably have a release note, and we could briefly explain what parallel means.  
We probably need to still recommend it is not used on a production server unless you have tested the impact and are happy it does not disturb the application.
Maybe we can offer a recommended usage, a guide number of threads as a starting point.
We also need to explain when this can or cannot be used (SerialGC, any other restrictions? It's not a problem that SerialGC does not use this, just that we should make it clear).  
I can help with the wording if needed, if you'd like to start something (this might make it into other docs also.)

y1yang0:
Hi Kevin, 

1. CSR

> Actually are we changing the default behaviour: from "parallel if you can", to "serial unless you specify".

We are not changing the default behavior. Now the default is still parallel (unless using SerialGC or parallel can not work), the VM will automatically calculate the appropriate parallelism unless the user specifies it.

Currently, this patch only adds the -parallel option to jcmd, AFAIK, this does not require a CSR. Only modifying jmap would require a CSR.

Given the compatibility issues previously mentioned in https://github.com/openjdk/jdk/pull/2261#discussion_r565295921 and subsequent bugs such as [JDK-8219721](https://bugs.openjdk.java.net/browse/JDK-8219721) [JDK-8219896](https://bugs.openjdk.java.net/browse/JDK-8219896), I'm hesitant to add a -parallel option to jmap, as it could lead to similar compatibility issues and bugs. Therefore, the best approach may be to keep it as it is:

- For jmap, VM calculates the number of parallel threads (and fallback to serial when unable to parallelize)
- For jcmd, VM provides fine-grained control, i.e. allowing -parallel=1-X to control whether to use serial or parallel.

Do you think this is acceptable?

2. Test

> Running the changes, I got a test failure on macos debug (x64 and aarch64):
> test/hotspot/jtreg/serviceability/HeapDump/IntegrityHeapDumpTest.java

Fixed.

The reason is that `SATestUtils.createProcessBuilder` in IntegrityHeapDump.java may require [user input for a password](https://github.com/y1yang0/jdk/actions/runs/5599168245/job/15188881990#step:9:978) when starting a process. I have modified test to be consistent with DuplicateClassArrayClassesTest.java to avert this problem.

3. Release Note

>This enhancement should probably have a release note, and we could briefly explain what parallel means.
> I can help with the wording if needed, if you'd like to start something

Thank you in advance! I've created related release-note issue, and I'll draft an initial version for this, please feel free to rephrase it.

kevinjwalls:
Thanks for the update -
Right, on the CSR I may have followed the links too far and been comparing with jmap features, not jcmd features. 8-)



On the default behaviour, running this change, with no -parallel= option, I have been seeing it default to a non-parallel dump, but now I see why:

I see num_dump_thread being set (e.g. 12 on my test system), but I see num_active_workers = 1 (requested is 12).

I just saw a run with no -parallel option which got num_active_workers = 2, so it did a parallel dump, so the default is being set as intended, but the availability of worker threads is not always predictable!

-Xlog:gc shows the GC was using fewer workers than expected.

So that suggested -XX:-UseDynamicNumberOfGCThreads and yes, with that I see e.g. num_active_workers = 23. req_num_dump_thread = 12 for my test.  And all 12 threads work on the heap dump.


-------------------

test/hotspot/jtreg/serviceability/dcmd/gc/HeapDumpTest.java:
we can't see if parallel dumping is happening
Can we add, like in I think another test: run testng/othervm -Xlog:heapdump HeapDumpTest

I would also suggest -XX:-UseDynamicNumberOfGCThreads otherwise as above, the GC can scale back and you don't always get a parallel dump when you expect it.


-------------------

Do we really want -parallel=-1 
..to be a valid input, and to use the workers, when it's clearly an error.

HeapDumpDCmd::execute casts it to a unit so we lose the fact that it was a request for -1.
Can we get the signed value in HeapDumpDCmd::execute, print a message and return like we do when compression level is out of range?
(Passing a non-number when an integer is expected will cause the jcmd to fail, and that seems correct.)


-------------------

So I think I'm done, looks good. 8-)

test/hotspot/jtreg/serviceability/HeapDump/IntegrityHeapDumpTest.java
```diff
@@ -81,7 +81,7 @@ private static void attachDumpAndVerify(File heapDumpFile,
                 .addToolArg("-parallel=" + Runtime.getRuntime().availableProcessors())
                 .addToolArg(heapDumpFile.getAbsolutePath());
 
-        ProcessBuilder processBuilder = SATestUtils.createProcessBuilder(launcher);
+        ProcessBuilder processBuilder = new ProcessBuilder(launcher.getCommand());
```
plummercj, line 84: Can you explain this change? This will result in the test not being run with sudo in cases where sudo is necessary. Before we get here we already checked if we need sudo (not running as root) and verified that passwordless sudo is supported. Otherwise the test is skipped and would never get here. So if sudo is required, the test will fail to attach to the debuggee. Note you might not see this issue if you have DevSecurityMode enabled. See [JDK-8313357](https://bugs.openjdk.org/browse/JDK-8313357)

plummercj: I see now that this is a new test, and not an SA test, so probably should never have been using SATestUtils in the first place (I'm curious as to how that ended up happening). You should also remove the import of SATestUtils.

y1yang0:
> I just saw a run with no -parallel option which got num_active_workers = 2, so it did a parallel dump, so the default is being set as intended, but the availability of worker threads is not always predictable!

Yes, now I've added new log outputs request_dump_thread and num_active_worker to get a better look at the details of dump.

> Can we add, like in I think another test: run testng/othervm -Xlog:heapdump HeapDumpTest

Done. In order not to interfere with existing tests, I added checks to the IntegrityHeapDumpTest.

> Can we get the signed value in HeapDumpDCmd::execute, print a message and return like we do when compression level is out of range?

Good catch! Done.

test/hotspot/jtreg/serviceability/HeapDump/IntegrityHeapDumpTest.java
y1yang0: > Can you explain this change? This will result in the test not being run with sudo in cases where sudo is necessary. Before we get here we already checked if we need sudo (not running as root) and verified that passwordless sudo is supported. Otherwise the test is skipped and would never get here. So if sudo is required, the test will fail to attach to the debuggee. Note you might not see this issue if you have DevSecurityMode enabled. See [JDK-8313357](https://bugs.openjdk.org/browse/JDK-8313357)

I followed what DuplicateArrayClassesTest.java. The reason seems that `SATestUtils.createProcessBuilder` does not touch any verification of whether passwordless sudo is supported:

```java
public class SATestUtils {
    public static ProcessBuilder createProcessBuilder(JDKToolLauncher launcher) {
        List<String> cmdStringList = Arrays.asList(launcher.getCommand());
        if (needsPrivileges()) {
            cmdStringList = addPrivileges(cmdStringList);
        }
        return new ProcessBuilder(cmdStringList);
    }
    public static boolean needsPrivileges() {
        return Platform.isOSX() && !Platform.isRoot();
    }
    private static List<String> addPrivileges(List<String> cmdStringList) {
        if (!Platform.isOSX()) {
            throw new RuntimeException("Can only add privileges on OSX.");
        }

        System.out.println("Adding 'sudo -E -n' to the command.");
        List<String> outStringList = new ArrayList<String>();
        outStringList.add("sudo");
        outStringList.add("-E"); // Preserve existing environment variables.
        outStringList.add("-n"); // non-interactive. Don't prompt for password. Must be cached or not required.
        outStringList.addAll(cmdStringList);
        return outStringList;
    }
```

kevinjwalls:
OK great.  
I think -parallel=0 could be intentional, meaning I don't want parallelism, I don't want parallel dump threads, that's a reasonable thing for a user to expect
(but parallel=-1 is clearly an error).  Can we just relax the constraint on that setting:

    if (parallel < 0) {
      output()->print_cr("Invalid number of parallel dump threads.");
      return;
	  
test/hotspot/jtreg/serviceability/dcmd/gc/HeapDumpTest.java
will need an adjustment where it checks parallelNum and checks the error text

-------------------------------------

Do we really need IntegrityHeapDumpTest.java ?
I don't see that we do.

It runs a LingeredApp, runs "jcmd PID GC.heap_dump -parallel=NUMPROCESSORS FILENAME" and runs verifyHeapDump(heapDumpFile);


The existing test which you have enhanced:
test/hotspot/jtreg/serviceability/dcmd/gc/HeapDumpTest.java:

Uses PidJcmdExecutor() will attach to the current process.  It's different, but we don't need to test that jcmd can attach to a LingeredApp for this change, it's the processing of whatever jcmd command text that's important.

The verification/integrity checking method is the same.

The new pidParallel() method tests a variety of parallel heap dumps, which is good (more variation than IntegrityHeapDumpTest).

If HeapDumpTest.java becomes "@run testng/othervm -Xlog:heapdump HeapDumpTest" then we can check for the parallel heap dump logging, which is maybe the reason for having IntegrityHeapDumpTest:

            // Expect parallel heap dump
            if (Runtime.getRuntime().availableProcessors() > 1) {
                String output = theApp.getProcessStdout();
                Asserts.assertTrue(output.contains("Dump heap objects in parallel"));
                Asserts.assertTrue(output.contains("Merge heap files complete"));
            }


A problem is that it always now passes the -parallel=X option, but we should still have tests that do NOT pass this option, to ensure it's not required.
This probably means overloading the run method?  The run method that doesn't take an "int paralleNum" needs to not pass the -parallel= option.

This did not occur to me earlier, but it would be great to have this Heap Dump testing in one place.

-------------------------------

y1yang0:
> Do we really need IntegrityHeapDumpTest.java ? I don't see that we do.
> A problem is that it always now passes the -parallel=X option, but we should still have tests that do NOT pass this option, to ensure it's not required.

The rationale for adding such a test is that we use HeapDumpCompressedTest to test a relatively big feature, i.e. heap dump compression, parallel heap dump follows this thought, maybe I should revert changes in HeapDumpTest and move all parallel related tests into IntegrityHeapDumpTest(Perhaps rename to HeapDumpParallelTest)? If we did so, HeapDumpTest covers the cases that we dont pass this option, which mentioned before.

kevinjwalls:
> 

Yes, dividing testing by feature so we have HeapDumpTest, HeapDumpCompressedTest, HeapDumpParallelTest... that sounds good.

Currently the HeapDumpTest does more parallel testing than IntegrityHeapDumpTest, so I didn't understand the need, but taking all parallel testing into its own test is good, and yes HeapDump test will not pass a -parallel option.  8-)

test/hotspot/jtreg/serviceability/HeapDump/IntegrityHeapDumpTest.java
plummercj: But DuplicateArrayClassesTest.java does not use SATestUtils.createProcessBuilder(), so I'm still wondering how this use of SATestUtils.createProcessBuilder() got in this new test in the first place.

Yes, a test must call SATestUtils.skipIfCannotAttach() before calling SATestUtils.createProcessBuilder(). But  these calls should only be made by SA tests. Non SA tests have no need for them.

And just to clarify, there are two types of "attaches" going on here. SA attach does a debugger attach, which varies depending on the OS, but requires some sort of root or sudo priveledges on OSX.  jcmds and tools like jstack and jmap use the java Attach API, which is just a socket based connection with the JVM, and just requires that the attaching process be the same user as the JDK process.

y1yang0: >  so I'm still wondering how this use of SATestUtils.createProcessBuilder() got in this new test in the first place

I simply grep existing tests and found this snippet to create java process. There is no special intention, it is an obvious misuse :-)

y1yang0:
> > 
> 
> Yes, dividing testing by feature so we have HeapDumpTest, HeapDumpCompressedTest, HeapDumpParallelTest... that sounds good.
> 
> Currently the HeapDumpTest does more parallel testing than IntegrityHeapDumpTest, so I didn't understand the need, but taking all parallel testing into its own test is good, and yes HeapDump test will not pass a -parallel option. 8-)

Done. Please take a look at the latest version :-)

kevinjwalls:
OK, nearly done!

We always log "parallelism true" which looks odd.  It just means the dump path has space for up to 9,999 log parts.  
In VM_HeapDumper::doit() if we make it log whether it will try a parallel dump, based on all the checks, it simplifies things I think.

Also the variable num_requested_dump_thread is not needed? It's just a copy of _num_dumper_threads which we don't change.

Here is a suggestion for VM_HeapDumper::doit()
...
  WorkerThreads* workers = ch->safepoint_workers();
  uint num_active_workers = workers != nullptr ? workers->active_workers() : 0;
  bool is_parallel = num_active_workers > 0 && can_parallel_dump() && is_parallel_dump();
  log_info(heapdump)("Requested dump threads %u, active workers %u, parallelism %s",
                      _num_dumper_threads, num_active_workers, is_parallel ? "true" : "false");

  if (!is_parallel) {
    _num_dumper_threads = 1;
    work(VMDumperWorkerId);
  } else {
    _num_dumper_threads = clamp(_num_dumper_threads, 2U, num_active_workers);
    uint heap_only_dumper_threads = _num_dumper_threads - 1 /* VMDumper thread */;
    _dumper_controller = new (std::nothrow) DumperController(heap_only_dumper_threads);
    ParallelObjectIterator poi(_num_dumper_threads);
    _poi = &poi;
    workers->run_task(this, _num_dumper_threads);
    _poi = nullptr;
  }
  
  
-----------------

HeapDumpParallelTest.java
We need to change all of the:
  Asserts.assertTrue(out.getStdout().contains(...
to:
  out.shouldContain(...

...then we get the output included in the log if it fails, so we should see what went wrong.

----------------------

y1yang0:
> Also the variable num_requested_dump_thread is not needed? It's just a copy of _num_dumper_threads which we don't change.

This is needed because _num_dumper_threads will change later, the requested dump thread may not equal to actual _num_dumper_threads. `is_parallel_dump()` checks _num_dump_thread, so we cannot use suggested code.

This involves code style favor, to avoid futile effort, do you think the below code is acceptable?
```diff
--- a/src/hotspot/share/services/heapDumper.cpp
+++ b/src/hotspot/share/services/heapDumper.cpp
@@ -1724,15 +1724,8 @@ class VM_HeapDumper : public VM_GC_Operation, public WorkerTask {
   }
   int dump_seq()           { return _dump_seq; }
   bool is_parallel_dump()  { return _num_dumper_threads > 1; }
-  bool can_parallel_dump() {
-    const char* base_path = writer()->get_file_path();
-    assert(base_path != nullptr, "sanity check");
-    if ((strlen(base_path) + 7/*.p\d\d\d\d\0*/) >= JVM_MAXPATHLEN) {
-      // no extra path room for separate heap dump files
-      return false;
-    }
-    return true;
-  }
+  bool can_parallel_dump(uint num_active_workers);
+
   VMOp_Type type() const { return VMOp_HeapDumper; }
   virtual bool doit_prologue();
   void doit();
@@ -1923,6 +1916,33 @@ bool VM_HeapDumper::doit_prologue() {
   return VM_GC_Operation::doit_prologue();
 }
 
+bool VM_HeapDumper::can_parallel_dump(uint num_active_workers) {
+  assert(base_path != nullptr, "sanity check");
+  bool has_extra_path_room = true;
+  const char* base_path = writer()->get_file_path();
+  if ((strlen(base_path) + 7/*.p\d\d\d\d\0*/) >= JVM_MAXPATHLEN) {
+    // no extra path room for separate heap dump files
+    has_extra_path_room = false;
+  }
+
+  bool can_parallel = true;
+  uint num_requested_dump_thread = _num_dumper_threads;
+  if (num_active_workers <= 1 ||         // serial gc?
+      num_requested_dump_thread <= 1 ||  // request serial dump?
+     !has_extra_path_room) {             // has extra path room for segmented dump files?
+    _num_dumper_threads = 1;
+    can_parallel = false;
+  } else {
+    _num_dumper_threads = clamp(num_requested_dump_thread, 2U, num_active_workers);
+    can_parallel = true;
+  }
+
+  log_info(heapdump)("Requested dump threads %u, active dump threads %u, "
+                     "actual dump threads %u, parallelism %s",
+                     num_requested_dump_thread, num_active_workers,
+                     _num_dumper_threads, can_parallel ? "true" : "false");
+  return can_parallel;
+}
 
 // The VM operation that dumps the heap. The dump consists of the following
 // records:
@@ -1970,20 +1990,10 @@ void VM_HeapDumper::doit() {
 
   WorkerThreads* workers = ch->safepoint_workers();
   uint num_active_workers = workers != nullptr ? workers->active_workers() : 0;
-  uint num_requested_dump_thread = _num_dumper_threads;
-  bool can_parallel = can_parallel_dump();
-  log_info(heapdump)("Requested dump threads %u, active dump threads %u, " "parallelism %s",
-                      num_requested_dump_thread, num_active_workers, can_parallel ? "true" : "false");
 
-  if (num_active_workers <= 1 ||         // serial gc?
-      num_requested_dump_thread <= 1 ||  // request serial dump?
-     !can_parallel) {                    // can not dump in parallel?
-    // Use serial dump, set dumper threads and writer threads number to 1.
-    _num_dumper_threads = 1;
+  if (can_parallel_dump(num_active_workers)) {
     work(VMDumperWorkerId);
   } else {
-    // Use parallel dump otherwise
-    _num_dumper_threads = clamp(num_requested_dump_thread, 2U, num_active_workers);
     uint heap_only_dumper_threads = _num_dumper_threads - 1 /* VMDumper thread */;
     _dumper_controller = new (std::nothrow) DumperController(heap_only_dumper_threads);
     ParallelObjectIterator poi(_num_dumper_threads);
```

> out.shouldContain(......then we get the output included in the log if it fails, so we should see what went wrong.

Done.

kevinjwalls:
Yes thanks, that makes VM_HeapDumper::doit() simpler which is completely the idea 8-)

OK in this new VM_HeapDumper::can_parallel_dump there is a use for num_requested_dump_thread variable, as we print the requested and final thread counts.
We were not doing that before, but it is useful.  I didn't see the change of _num_dump_threads affecting whether this was or was not a parallel dump, but not important if we go with your new method.

(we should call num_requested_dump_threads and then we are probably done here 8-) )

test/hotspot/jtreg/serviceability/dcmd/gc/HeapDumpParallelTest.java
```diff
@@ -0,0 +1,141 @@
...
+}
+
+public class HeapDumpParallelTest {
+    static HeapDumpParallel theApp;
+
+    private static void checkAndVerify(OutputAnalyzer out, LingeredApp app, File heapDumpFile, boolean expectSerial) {
+        out.shouldHaveExitValue(0);
+        out.shouldContain("Heap dump file created");
+        if (!expectSerial && Runtime.getRuntime().availableProcessors() > 1) {
+            Asserts.assertTrue(app.getProcessStdout().contains("Dump heap objects in parallel"));
```
kevinjwalls, line 59: I think we just need these other asserts converted to shouldContain, and the assertFalse changed to shouldNotContain.
This might be a annoying, but if you get the test failing and don't get the actual output, that's worse. 8-)
Maybe we can aim to integrate on Monday.

y1yang0: getProcessStdout is actually a string instead of OutputAnalyzer, its name is somewhat deceptive

kevinjwalls:
Yes, the assert is unfortunately unhelpful when it fails, e.g.
  java.lang.RuntimeException: assertTrue: expected true, was false
(and you look up the test source for what the assert means)

We should find a way of using OutputAnalayzer and shouldContain() so we get a readable assert/failure.

Use of getProcessStdout(): I tested and see that it returns a cumulative stdout log from the LingeredApp, so if we assert on that output, we are checking previous iterations.  Change the order of the checkAndVerify calls and it fails.  We should launch a new app for each test and look at the output separately. 

It's not too bad to change the test to launch a LingeredApp and test the output, I made an attempt at that.  I'll share a testcase suggestion here, see what you think.

I may not be seeing how best to drive the test library, and it also may be uncommon to analzye the output of the LingeredApp, not just the jcmd or other tool run with a CommandExecutor.  So I don't see a very elegant way to capture an OutputAnalyzer of the completed LingeredApp, and I'm creating a new OutputAnalyzer(app.getProcessStdout())

kevinjwalls:
import java.io.File;
import java.io.IOException;
import java.nio.file.Files;
import java.util.List;

import jdk.test.lib.Asserts;
import jdk.test.lib.JDKToolLauncher;
import jdk.test.lib.apps.LingeredApp;
import jdk.test.lib.dcmd.PidJcmdExecutor;
import jdk.test.lib.process.OutputAnalyzer;
import jdk.test.lib.process.ProcessTools;

import jdk.test.lib.hprof.HprofParser;

/**
 * @test
 * @bug 8306441
 * @summary Verify the integrity of generated heap dump and capability of parallel dump
 * @library /test/lib
 * @run driver HeapDumpParallelTest
 */

public class HeapDumpParallelTest {

    private static void checkAndVerify(OutputAnalyzer dcmdOut, LingeredApp app, File heapDumpFile, boolean expectSerial) throws IOException {
        dcmdOut.shouldHaveExitValue(0);
        dcmdOut.shouldContain("Heap dump file created");
        OutputAnalyzer appOut = new OutputAnalyzer(app.getProcessStdout());
        appOut.shouldContain("[heapdump]");
        if (!expectSerial && Runtime.getRuntime().availableProcessors() > 1) {
            appOut.shouldContain("Dump heap objects in parallel");
            appOut.shouldContain("Merge heap files complete");
        } else {
            appOut.shouldNotContain("Dump heap objects in parallel");
            appOut.shouldNotContain("Merge heap files complete");
        }
        verifyHeapDump(heapDumpFile);
        if (heapDumpFile.exists()) {
            heapDumpFile.delete();
        }
    }

    private static LingeredApp launchApp() throws IOException {
        LingeredApp theApp = new LingeredApp();
        LingeredApp.startApp(theApp, "-Xlog:heapdump", "-Xmx512m",
                             "-XX:-UseDynamicNumberOfGCThreads",
                             "-XX:ParallelGCThreads=2");
        return theApp;
    }

    public static void main(String[] args) throws Exception {
        String heapDumpFileName = "parallelHeapDump.bin";

        File heapDumpFile = new File(heapDumpFileName);
        if (heapDumpFile.exists()) {
            heapDumpFile.delete();
        }

        LingeredApp theApp = launchApp();
        try {
            // Expect error message
            OutputAnalyzer out = attachJcmdHeapDump(heapDumpFile, theApp.getPid(), "-parallel=" + -1);
            out.shouldContain("Invalid number of parallel dump threads.");

            // Expect serial dump because 0 implies to disable parallel dump
            test(heapDumpFile, "-parallel=" + 0, true);

            // Expect serial dump
            test(heapDumpFile,  "-parallel=" + 1, true);

            // Expect parallel dump
            test(heapDumpFile, "-parallel=" + Integer.MAX_VALUE, false);

            // Expect parallel dump
            test(heapDumpFile, "-gz=9 -overwrite -parallel=" + Runtime.getRuntime().availableProcessors(), false);
        } finally {
            theApp.stopApp();
        }
    }

    private static void test(File heapDumpFile, String arg, boolean expectSerial) throws Exception {
        LingeredApp theApp = launchApp();
        try {
            OutputAnalyzer dcmdOut = attachJcmdHeapDump(heapDumpFile, theApp.getPid(), arg);
            theApp.stopApp();
            checkAndVerify(dcmdOut, theApp, heapDumpFile, expectSerial);
        } finally {
            theApp.stopApp();
        }
    }

    private static OutputAnalyzer attachJcmdHeapDump(File heapDumpFile, long lingeredAppPid, String arg) throws Exception {
        // e.g. jcmd <pid> GC.heap_dump -parallel=cpucount <file_path>
        System.out.println("Testing pid " + lingeredAppPid);
        PidJcmdExecutor executor = new PidJcmdExecutor("" + lingeredAppPid);
        return executor.execute("GC.heap_dump " + arg + " " + heapDumpFile.getAbsolutePath());
    }

    private static void verifyHeapDump(File dump) {
...as before...

y1yang0:
> It's not too bad to change the test to launch a LingeredApp and test the output, I made an attempt at that. I'll share a testcase suggestion here, see what you think.

Thank you for the share! It makes sense to me. This test will start JVMs x4 times, which may take a long time, but until a more elegant solution is available, I think we can reach a consensus at the  moment (and perhaps file another test improvement issue to optimize it in the future.)

/contributor add @kevinjwalls

kevinjwalls:
Thanks for being responsive to all these points.  
(I did worry also about starting so many test apps but I don't see a better way right now.  We can revisit if this is a problem.)

I think we should get this integrated but first, sorry, the test fails with UseSerialGC, which will happen in some testing tier.
We do want it to run with all different options that the test harness may request, and all GCs.
So the test needs to realize it's not parallel, we can make it check the options and force expectSerial to true.  This can do it, I tested with the different collectors:

```
bash-4.2$ git diff
diff --git a/test/hotspot/jtreg/serviceability/dcmd/gc/HeapDumpParallelTest.java b/test/hotspot/jtreg/serviceability/dcmd/gc/HeapDumpParallelTest.java
index 247916c93db..5a9d40fbb30 100644
--- a/test/hotspot/jtreg/serviceability/dcmd/gc/HeapDumpParallelTest.java
+++ b/test/hotspot/jtreg/serviceability/dcmd/gc/HeapDumpParallelTest.java
@@ -24,10 +24,12 @@
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
+import java.util.Arrays;
 import java.util.List;

 import jdk.test.lib.Asserts;
 import jdk.test.lib.JDKToolLauncher;
+import jdk.test.lib.Utils;
 import jdk.test.lib.apps.LingeredApp;
 import jdk.test.lib.dcmd.PidJcmdExecutor;
 import jdk.test.lib.process.OutputAnalyzer;
@@ -40,7 +42,7 @@ import jdk.test.lib.hprof.HprofParser;
  * @bug 8306441
  * @summary Verify the integrity of generated heap dump and capability of parallel dump
  * @library /test/lib
- * @run driver HeapDumpParallelTest
+ * @run main HeapDumpParallelTest
  */

 public class HeapDumpParallelTest {
@@ -50,6 +52,11 @@ public class HeapDumpParallelTest {
         dcmdOut.shouldContain("Heap dump file created");
         OutputAnalyzer appOut = new OutputAnalyzer(app.getProcessStdout());
         appOut.shouldContain("[heapdump]");
+        String opts = Arrays.asList(Utils.getTestJavaOpts()).toString();
+        if (opts.contains("-XX:+UseSerialGC")) {
+            System.out.println("UseSerialGC detected.");
+            expectSerial = true;
+        }
         if (!expectSerial && Runtime.getRuntime().availableProcessors() > 1) {
             appOut.shouldContain("Dump heap objects in parallel");
             appOut.shouldContain("Merge heap files complete");
@@ -137,4 +144,5 @@ public class HeapDumpParallelTest {
             Asserts.fail("Could not parse dump file " + dump.getAbsolutePath());
         }
     }
-}
\ No newline at end of file
+}
+
bash-4.2$		 

```

kevinjwalls:
Great, thanks for your patience with all of this! 8-)

kevinjwalls (approved):

PR #11431, opened 2022-11-30
8250269: Replace ATTRIBUTE_ALIGNED with alignas

TheShermanTanker:
C++11 added the alignas attribute, for the purpose of specifying alignment on types, much like compiler specific syntax such as gcc's \_\_attribute\_\_((aligned(x))) or Visual C++'s \_\_declspec(align(x)).

We can phase out the use of the macro in favor of the standard attribute. In the meantime, we can replace the compiler specific definitions of ATTRIBUTE_ALIGNED with a portable definition. We might deprecate the use of the macro but changing its implementation quickly and cleanly applies the feature where the macro is being used.

Note: With certain parts of HotSpot using ATTRIBUTE_ALIGNED so indiscriminately, this commit will likely take some time to get right

This will require adding the alignas attribute to the list of language features approved for use in HotSpot code. (Completed with [8297912](https://github.com/openjdk/jdk/pull/11446))

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Issue
 * [JDK-8250269](https://bugs.openjdk.org/browse/JDK-8250269): Replace ATTRIBUTE_ALIGNED with alignas (**Enhancement** - P4)

### Reviewers
 * [Roman Kennke](https://openjdk.org/census#rkennke) (@rkennke - **Reviewer**)
 * [Kim Barrett](https://openjdk.org/census#kbarrett) (@kimbarrett - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/11431/head:pull/11431` \
`$ git checkout pull/11431`

Update a local copy of the PR: \
`$ git checkout pull/11431` \
`$ git pull https://git.openjdk.org/jdk.git pull/11431/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 11431`

View PR using the GUI difftool: \
`$ git pr show -t 11431`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jdk/pull/11431.diff">https://git.openjdk.org/jdk/pull/11431.diff</a>

</details>

### Webrev
[...]

PR #13668, opened 2023-04-26
8306901: Macro offset_of confuses Eclipse CDT

reinrich:
With this pr I would like to wrap the body of the macro `offset_of` as defined for gcc in parentheses to fix the described issues with Eclipse CDT.
As an alternative the [`(int)` cast in `byte_offset_of`](https://github.com/openjdk/jdk/blob/44d9f55d0b3c469988be6f1c47f0cfbc433c4490/src/hotspot/share/utilities/sizes.hpp#L59) could be removed. I preferred adding the parentheses to the gcc version of `offset_of` because the impact is smaller.

Of course I'd rather have a local solution for the issue but I couldn't find one that didn't require changes in the source code.

Testing: GHA

---------
### Progress

### Issue
 * [JDK-8306901](https://bugs.openjdk.org/browse/JDK-8306901): Macro offset_of confuses Eclipse CDT

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13668/head:pull/13668` \
`$ git checkout pull/13668`

`$ git checkout pull/13668` \
`$ git pull https://git.openjdk.org/jdk.git pull/13668/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13668`

`$ git pr show -t 13668`

</details>

<a href="https://git.openjdk.org/jdk/pull/13668.diff">https://git.openjdk.org/jdk/pull/13668.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13668#issuecomment-1523693248)

### DIFF
--- a/src/hotspot/share/utilities/globalDefinitions_gcc.hpp
+++ b/src/hotspot/share/utilities/globalDefinitions_gcc.hpp
@@ -140,12 +140,12 @@ inline int g_isfinite(jdouble f)                 { return isfinite(f); }
 // use offsetof() instead, with the invalid-offsetof warning
 // temporarily disabled.
 #define offset_of(klass,field)                          \
-[]() {                                                  \
+([]() {                                                 \
   char space[sizeof (klass)] ATTRIBUTE_ALIGNED(16);     \
   klass* dummyObj = (klass*)space;                      \
   char* c = (char*)(void*)&dummyObj->field;             \
   return (size_t)(c - space);                           \
-}()
+}())
 
 
 #if defined(_LP64) && defined(__APPLE__)


kimbarrett (approved):
Looks good.

reinrich:
Hi Thomas,

> Looks good. Thanks for fixing this, I had noticed this in CDT myself.

thanks for the review!

> I never looked closely at this macro though. I wish we could just use offsetof.

Indeed.

> Also I wonder whether the ATTRIBUTE_ALIGNED(16) would be more correctly `alignas(klass)` but that is pre-existing.

Probably :)
At least it would be cleaner in my eyes than the hard coded 16.
With this pr I'd like to address just the CDT issue though.

Thanks, Richard.

tstuefe:
> Hi Thomas,
> 
> > Looks good. Thanks for fixing this, I had noticed this in CDT myself.
> 
> thanks for the review!
> 
> > I never looked closely at this macro though. I wish we could just use offsetof.
> 
> Indeed.
> 
> > Also I wonder whether the ATTRIBUTE_ALIGNED(16) would be more correctly `alignas(klass)` but that is pre-existing.
> 
> Probably :) At least it would be cleaner in my eyes than the hard coded 16. With this pr I'd like to address just the CDT issue though.
> 

I believe the trick may not work if the type were artificially aligned to something larger than16, e.g. struct XX alignas(64). I guess that depends on what the compiler does when casting a pointer with a smaller alignment to something with a larger alignemnt.

> Thanks, Richard.

tstuefe:
> > > > Also I wonder whether the ATTRIBUTE_ALIGNED(16) would be more correctly `alignas(klass)` but that is pre-existing.
> > > 
> > > 
> > > Probably :) At least it would be cleaner in my eyes than the hard coded 16. With this pr I'd like to address just the CDT issue though.
> > 
> > 
> > I believe the trick may not work if the type were artificially aligned to something larger than16, e.g. struct XX alignas(64). I guess that depends on what the compiler does when casting a pointer with a smaller alignment to something with a larger alignemnt.
> 
> Maybe. I'm not familiar enough with `alignas` right now. Also I would have to 'meditate' a bit about [the alignas-part in the hotspot-style guide](https://github.com/openjdk/jdk/blob/master/doc/hotspot-style.md#alignas). Right now I wouldn't expect extended alignment of a type to affect member offsets. But that might be naive.

Checked with GCC and clang, it seems to work. But I think its still UB (otherwise we could just omit the ALIGN specifier completely).

reinrich:
Thanks for the reviews.
/integrate

TheShermanTanker:
@tstuefe Currently there is a change planned to replace the macro in some place with alignas (see #11431) but I haven't had time to get round to finishing it just yet

PR #13669, opened 2023-04-26
8301686: TLS 1.3 handshake fails if server_name doesn't match resuming session

jaikiran:
Can I please get a review of this change which proposes to fix the issue reported in https://bugs.openjdk.org/browse/JDK-8301686?

The internal implementation of SSLContext caches SSLSession(s). These sessions are for a particular combination or peer host and port. When a TLS handshake completes successfully, the session is then stored in this cache. If/when subsequent handshake attempts against the same host/port combination happens, using this same SSLContext instance, then the internal implementation triggers a session resumption, which is allowed by the TLS RFC. During session resumption, the client then uses the pre-shared key from the previous successful handshake and sends it as part of the `ClientHello` message. 

One other part of the TLS handshake is the `server_name` extension. The client sends a `SNI` in the handshake which the server side can either reject or accept. To facilitate this matching on the server side, the `javax.net.ssl.SNIMatcher` can be configured on the (server side) `SSLParameters`. Setting of `SNIMatcher` is optional. 

If a successful handshake session (that was cached by the client) used a SNI name which the server accepted, then this SNI name is sent as part of the session resumption `ClientHello` along with the pre-shared key.  The current issue is that, during session resumption, on the server side, if the `SNIMatcher` is no longer present, then the server rightly aborts the session resumption, but still ends up sending the pre-shared key extension as part of the `ServerHello` message. The client, upon seeing this pre-shared key extension in the `ServerHello` considers that the session resumption succeeded and ends up using that pre-shared key to derive the early secret. On the server side though, the server has already rejected this resumption request and thus when the client next sends data, the server will no longer be able to decode the data and will fail with `javax.crypto.AEADBadTagException: Tag mismatch` as noted in the JBS issue.

The change in this PR, removes the pre-shared key extension data from the `ServerHello` message, when the server side notices that the session resumption is being aborted.

A new jtreg test has been added which reproduces the issue and verifies the fix. Existing tests in tier1, tier2 and tier3 continue to pass with this change.

---------
### Progress

### Issue
 * [JDK-8301686](https://bugs.openjdk.org/browse/JDK-8301686): TLS 1.3 handshake fails if server_name doesn't match resuming session (**Bug** - P4)

### Reviewers
 * [Daniel Jeliński](https://openjdk.org/census#djelinski) (@djelinski - **Reviewer**) ⚠️ Review applies to [57df739d](https://git.openjdk.org/jdk/pull/13669/files/57df739dcb1a2d0bd07d92f62c799aa25553e4ef)
 * [Bradford Wetmore](https://openjdk.org/census#wetmore) (@bradfordwetmore - **Reviewer**) ⚠️ Review applies to [c6ec3eb9](https://git.openjdk.org/jdk/pull/13669/files/c6ec3eb91b56084ba095b92631ac79fbe940452e)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13669/head:pull/13669` \
`$ git checkout pull/13669`

`$ git checkout pull/13669` \
`$ git pull https://git.openjdk.org/jdk.git pull/13669/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13669`

`$ git pr show -t 13669`

</details>

<a href="https://git.openjdk.org/jdk/pull/13669.diff">https://git.openjdk.org/jdk/pull/13669.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13669#issuecomment-1523185583)


test/jdk/javax/net/ssl/SSLSession/ServerNameRejectedTLSSessionResumption.java
```diff
@@ -0,0 +1,209 @@
...
+import javax.net.ssl.SSLParameters;
+import javax.net.ssl.StandardConstants;
+
+/*
+ * @test
+ * @bug 8301686
+ * @summary verifies that if the server rejects session resumption due to SNI mismatch,
+ *          during TLS handshake, then the subsequent communication between the server and the
+ *          client happens correctly without any errors
+ * @run main/othervm -Djavax.net.ssl.keyStore=${test.src}/../etc/keystore
```
mpdonova, line 44: We've been trying to move away from using the binary keystore and truststore files. The alternative is to extend `SSLContextTemplate` and use `createServerSSLContext()` and `createClientSSLContext()` methods.

jaikiran: Thank you Matthew for that review. I've now updated the PR to extend the `SSLContextTemplate` and use its APIs for `SSLContext` creation in the test.

With this updated test, I've verified that the test continues to fail (as expected) without the source fix and passes with the fix.

jaikiran:
Please keep open. Waiting for a review from security-libs team.

jaikiran:
Please keep open

jaikiran:
Please keep open, expecting a security-libs review.

src/java.base/share/classes/sun/security/ssl/ServerNameExtension.java
```diff
@@ -342,6 +343,9 @@ public void consume(ConnectionContext context,
                         sni, shc.resumingSession.serverNameIndication)) {
                     shc.isResumption = false;
                     shc.resumingSession = null;
+                    // the server is disallowing this session resumption,
```
bradfordwetmore, line 346: Would you please change "the server" to "this server."

test/jdk/javax/net/ssl/SSLSession/ServerNameRejectedTLSSessionResumption.java
```diff
@@ -0,0 +1,210 @@
+/*
```
bradfordwetmore, line 1: Please reformat to <= 80 chars.  This helps reviewers who prefer side/side reviews from having to horizontally scroll, even with large screens.  Thanks.

bradfordwetmore (approved):
Other than the two small comments, LGTM.

jaikiran:
Thank you Brad for the review. I've now updated the PR to address your review suggestions. I've also triggered a CI run with these latest changes.

jaikiran:
The CI job completed without issues - tier1,tier2 and tier3. I'll go ahead and integrate this PR later today/tomorrow.

src/java.base/share/classes/sun/security/ssl/ServerNameExtension.java
```diff
@@ -343,7 +343,7 @@ public void consume(ConnectionContext context,
                         sni, shc.resumingSession.serverNameIndication)) {
                     shc.isResumption = false;
                     shc.resumingSession = null;
-                    // the server is disallowing this session resumption,
+                    // this server is disallowing this session resumption,
                     // so don't include the pre-shared key in the ServerHello handshake message
```
bradfordwetmore, line 347: If you wouldn't mind <= 80 here also.  The other changes look great.  

Thanks for considering.

bradfordwetmore (approved):

src/java.base/share/classes/sun/security/ssl/ServerNameExtension.java
jaikiran: Sorry Brad, I didn't notice that one in my previous update. I've now updated the PR to fix this line.

PR #13598, opened 2023-04-22
8306697: Add method to obtain String for CONSTANT_Class_info in ClassDesc

liach:
Add a method `internalName` to `ClassDesc`, and unifies handling of string representation of a class constant in CONSTANT_Class_info via `ofInternalName` and `internalName` APIs, documented in `ClassDesc` itself. In particular, `ofInternalName` now accepts arrays.

The motivation of this API is that avoiding frequent String creations via caching (enabled by this new API, will be in a separate patch) would speed up Classfile API's [writing of simple class files](https://github.com/openjdk/jdk/blob/master/test/micro/org/openjdk/bench/jdk/classfile/Write.java) by 1/3. See https://mail.openjdk.org/pipermail/classfile-api-dev/2023-April/000296.html for more context.

This API is futureproof: for Valhalla's Q-types, it will return their string representation in CONSTANT_Class_info, which is most likely their full descriptor string.

Javadoc: https://cr.openjdk.org/~liach/8306697/java.base/java/lang/constant/ClassDesc.html

---------
### Progress
- [ ] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [ ] Change requires CSR request [JDK-8306700](https://bugs.openjdk.org/browse/JDK-8306700) to be approved

### Issues
 * [JDK-8306697](https://bugs.openjdk.org/browse/JDK-8306697): Add method to obtain String for CONSTANT_Class_info in ClassDesc
 * [JDK-8306700](https://bugs.openjdk.org/browse/JDK-8306700): Add method to obtain String for CONSTANT_Class_info in ClassDesc (**CSR**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13598/head:pull/13598` \
`$ git checkout pull/13598`

Update a local copy of the PR: \
`$ git checkout pull/13598` \
`$ git pull https://git.openjdk.org/jdk.git pull/13598/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 13598`

View PR using the GUI difftool: \
`$ git pr show -t 13598`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
[...]

PR #13550, opened 2023-04-20
8305669: RuntimeException when running benchmarks through make on Windows/WSL

liach:
This patch replaces `FIXPATH` with `FixPath` on individual path argumenets. The root cause might be that JMH requires passing VM args to benchmarks in quotes, which might have triggered incorrect detections in `FIXPATH` (as in comparison, `$1_MICRO_JAVA_OPTIONS += --add-opens=java.base/java.io=ALL-UNNAMED` right above this patch seems to work fine)
Also removed a useless and wrong Java flag to the java running javac, originally intended to enable running `make test TEST="loom.obsolete"` benchmarks.

Since I only have a windows cygwin environment, I am not quite sure if this works elsewhere.

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8305669](https://bugs.openjdk.org/browse/JDK-8305669): RuntimeException when running benchmarks through make on Windows/WSL

### Reviewers
 * [Erik Joelsson](https://openjdk.org/census#erikj) (@erikj79 - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13550/head:pull/13550` \
`$ git checkout pull/13550`

Update a local copy of the PR: \
`$ git checkout pull/13550` \
`$ git pull https://git.openjdk.org/jdk.git pull/13550/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 13550`

View PR using the GUI difftool: \
`$ git pr show -t 13550`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jdk/pull/13550.diff">https://git.openjdk.org/jdk/pull/13550.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13550#issuecomment-1515629753)

PR #13186, opened 2023-03-26
8304932: MethodTypeDescImpl can be mutated by argument passed to MethodTypeDesc.of

liach:
Fixed the bug where if a caller keeps a reference to the array passed into `MethodTypeDesc.of`, the caller may mutate the Desc via the array and can create invalid MethodTypeDesc.

Unfortunately, since the input array now needs to be copied, the `of` factory suffers from a performance drop. But otherwise, this patch has minor performance gains on `ofDescriptor` factory, even compared to Adam's patch that optimized `ofDescriptor` in #12945.

This patch moves the parameters to an immutable list, to avoid allocations on `parameterList` as well.

Benchmark of Oracle JDK 20: https://gist.github.com/683c6219e183cbc2b336224fc2c0d50a
Benchmark of this patch: https://gist.github.com/22be9371a2370fb4a7b44f1684750ec4
Benchmark of [asotona's patch](https://github.com/openjdk/jdk/pull/12945/files#diff-ac8e413d3e13532a2b0d34a90253c6ddd7a4f04082f792b9d076e9b5a33f2078): https://gist.github.com/eb98579c3b51cafae481049a95a78f80

[sotona vs this](https://jmh.morethan.io/?gists=eb98579c3b51cafae481049a95a78f80,22be9371a2370fb4a7b44f1684750ec4); [20 vs this](https://jmh.morethan.io/?gists=683c6219e183cbc2b336224fc2c0d50a,22be9371a2370fb4a7b44f1684750ec4); [20 vs sotona](https://jmh.morethan.io/?gists=683c6219e183cbc2b336224fc2c0d50a,eb98579c3b51cafae481049a95a78f80), for reference

PR #13671, opened 2023-04-26
8306842: Classfile API performance improvements

asotona:
Following improvements implemented:
- Switch over `String` replaced with switch single char
- Binary search for frames in `StackMapGenerator`
- `StackMapGenerator.rawHandlers` with pre-calculated offsets
- `ClassEntry` is caching `ClassDesc` symbol
- Caching of type symbols in `NameAndTypeEntry` and `MethodTypeEntry`
- Caching `MethodTypeDesc` in `MethodInfo` implementations
- `StackMapGenerator` and `Utils` delegating to cached `MethodTypeDesc` instead of custom parsing

No API change.

Benchmarks show stack map generation improved by 21% and code generation from symbols improved by 30%.

```
Benchmark                     Mode  Cnt       Score       Error  Units
GenerateStackMaps.benchmark  thrpt   10  407931.202 ± 13101.023  ops/s
RebuildMethodBodies.shared   thrpt    4   10258.597 ±   383.699  ops/s
RebuildMethodBodies.unshared thrpt    4    7224.543 ±   256.800  ops/s
```

```
Benchmark                     Mode  Cnt       Score      Error  Units
GenerateStackMaps.benchmark  thrpt   10  495101.110 ± 2389.628  ops/s
RebuildMethodBodies.shared    thrpt   4   13380.272 ±  810.113  ops/s
RebuildMethodBodies.unshared  thrpt   4    9399.863 ±  557.060  ops/s
```

---------
### Progress

### Issue
 * [JDK-8306842](https://bugs.openjdk.org/browse/JDK-8306842): Classfile API performance improvements

### Reviewers
 * [Claes Redestad](https://openjdk.org/census#redestad) (@cl4es - **Reviewer**) ⚠️ Review applies to [d353bb33](https://git.openjdk.org/jdk/pull/13671/files/d353bb339ac2d0281869e6afafaeac2e58e5d367)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13671/head:pull/13671` \
`$ git checkout pull/13671`

`$ git checkout pull/13671` \
`$ git pull https://git.openjdk.org/jdk.git pull/13671/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13671`

`$ git pr show -t 13671`

</details>

<a href="https://git.openjdk.org/jdk/pull/13671.diff">https://git.openjdk.org/jdk/pull/13671.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13671#issuecomment-1540215542)


src/java.base/share/classes/jdk/internal/classfile/impl/AbstractPoolEntry.java
```diff
@@ -556,18 +561,29 @@ public boolean equals(Object o) {
...
+            }
         }
 
         @Override
         public ClassDesc asSymbol() {
-            return Util.toClassDesc(asInternalName());
+            if (sym == null) {
+                sym = Util.toClassDesc(asInternalName());
+            }
+            return sym;
```
ExE-Boss, line 586:
```suggestion
            var sym = this.sym;
            if (sym != null) {
                return sym;
            }
            return this.sym = Util.toClassDesc(asInternalName());
```

src/java.base/share/classes/jdk/internal/classfile/impl/AbstractPoolEntry.java
```diff
@@ -929,7 +971,21 @@ public Utf8Entry descriptor() {
...
+                return ret;
+            }
+        }
+
+        @Override
+        public MethodTypeDesc asSymbol() {
+            if (sym == null) {
+                sym = MethodTypeDesc.ofDescriptor(descriptor().stringValue());
+            }
+            return sym;
```
ExE-Boss, line 988:
```suggestion
            var sym = this.sym;
            if (sym != null) {
                return sym;
            }
            return this.sym = MethodTypeDesc.ofDescriptor(descriptor().stringValue());
```

asotona:
> 1. CodeBuilder.invokeInstruction and CodeBuilder.fieldInstruction can pass their symbols to the NameAndTypeEntry as well, since it's almost always going to be used by stack map generation later.

Yes, they actually do. CodeBuilder conventional methods delegate down to the NaTEntry construction from symbols. I hope I didn't miss any important path.  

> 2. Since the casts to access the field and method type symbols in `NameAndTypeEntryImpl` is quite complex, can we move it to a utility method in `Util` for cleaner call sites?
> 3. `parameterSlots` `parseParameterSlots` `maxLocals` in `Util` can probably operate on a `MethodTypeDesc` than a bitset, especially that the method type symbols are available in most scenarios already.
> 4. `StackMapGenerator.processInvokeInstructions` can probably scan through the `MethodTypeDesc` instead of using the hand-rolled `while (++pos < descLen && (ch = mDesc.charAt(pos)) != ')')` loop. In fact, the whole loop can be replaced by `Util.parameterSlots()`
> 5. `StackMapGenerator.isDoubleSlot(ClassDesc)` should be moved to `Util` and used by `parameterSlots()` etc (assuming they've migrated to `MethodTypeDesc`), and implementation be updated to:
> 
> ```java
>     public static boolean isDoubleSlot(ClassDesc desc) {
>         return desc.isPrimitive() && (desc.charAt(0) == 'D' || desc.charAt(0) == 'J');
>     }
> ```
> 
> to avoid potentially slow string comparisons.

These are great suggestions, I'll work on them.

Thanks!

liach:
Though the proposal to add method to obtain internal names on ClassDesc is rejected, I think we can still obtain a performance boost if we can avoid frequent conversion between internal names and descriptors.

Looking at the usages of `Util.toClassDesc` and `Util.toInternalName`, we can already cache result of `toClassDesc` in `ClassEntry`. We can think similarly for `toInternalName`: the hash for `ClassEntry` may be derived from the class descriptor than the class internal name, so we don't need to call `toInternalName` every time we want to look up a ClassEntry with a ClassDesc in a constant pool.

asotona:
Caching of `ClassDesc` in `ClassEntry` is already a part of this PR and the internal name is key for `ClassEntry` from the very beginning, so obtaining internal name from `ClassEntry` was never an issue.
However `ClassEntry` is a short-live object, newly created for each generated or transformed class and it exists within a constant pool context only.
In order to reach the same performance as with cached internal names in `ClassDesc` instances, we would have to change the user coding schema and not the Classfile API internals.
Every code generation or transformation would have to start with static declaration of `ClassEntry` constants (instead of `ClassDesc`), constructed from to `TemporaryConstantPool` (which is for internal purpose yet).

asotona:
I've added new benchmarks measuring repeated transformations rebuilding method bodies (down to the symbols expansion) from already expanded models.
I’ve added shared and unshared CP alternatives (for curiosity), however the unshared is closer to simulation of building from symbols.
 
Here are results from the actual code base:
 
Benchmark                      Mode  Cnt      Score     Error  Units
RebuildMethodBodies.shared    thrpt    4  10258.597 ± 383.699  ops/s
RebuildMethodBodies.unshared  thrpt    4   7224.543 ± 256.800  ops/s
 
And here is already visible approx. 22% improvement in both (without the rest of proposed improvements yet):
 
Benchmark                      Mode  Cnt      Score      Error  Units
RebuildMethodBodies.shared    thrpt    4  12498.597 ± 309.585  ops/s
RebuildMethodBodies.unshared  thrpt    4   8807.229 ±  167.247  ops/s

It would be also interesting to see a difference with cached internal names in `ClassDesc`.

asotona:
Now we have following benchamrk numbers:
Benchmark                      Mode  Cnt      Score      Error  Units
RebuildMethodBodies.shared    thrpt    4  13585.081 ± 1232.900  ops/s
RebuildMethodBodies.unshared  thrpt    4   9535.584 ± 1201.842  ops/s

Which is approx. 32% performance improvement :)

asotona:
I benchmarked merged this PR with #13598 and benefits of caching of internal names in ClassDesc are insignificant (~2% performance boost).

liach:
If we can hash internal names for constant pool without calling Util.toInternalName on ClassDesc instances, can that offset the performance penalty of toInternalName?

asotona:
> If we can hash internal names for constant pool without calling Util.toInternalName on ClassDesc instances, can that offset the performance penalty of toInternalName?

Yes, that may help. There is already manual hash calculation from the raw bytes, so it should work the same way.
Good idea, thanks.

liach:
I have a proof-of-concept patch here https://github.com/liachmodded/jdk/commit/3e6aa523521085e99b52fb49633676166910100f
which hashes class entries by the symbol's full descriptors instead of by the index of the linked utf8; we can access the hash of a full descriptor given the hash of an internal name easily, should be faster than iterating over or creating the internal name for a hash.

ExE-Boss:
In order to avoid eager conversion of a `ClassDesc` to internal name, I suggest extending `AbstractPoolEntry.Utf8EntryImpl` with `State.CLASS_DESC`:
```java
class AbstractPoolEntry.Utf8EntryImpl extends AbstractPoolEntry implements Utf8Entry {
	enum State { CLASS_DESC, ... }

	// Set in any state other than RAW or CLASS_DESC
	private int hash;
	// Set in any state other than RAW
	private int charLen;
	// Only set in CLASS_DESC state, must be an L-type Class descriptor
	private ClassDesc classDesc;

	Utf8EntryImpl(ConstantPool cpm, int index, ClassDesc cd) {
		super(cpm, Classfile.TAG_UTF8, index, 0);
		this.rawBytes = null;
		this.offset = 0;
		this.rawLen = 0;
		String descriptor = cd.descriptorString();
		if (cd.isArray()) {
			this.state = State.STRING;
			this.stringValue = descriptor;
			this.charLen = descriptor.length();
			this.hash = hashString(descriptor.hashCode());
		} else {
			assert cd.isClassOrInterface() : cd;
			this.state = State.CLASS_DESC;
			this.classDesc = cd;
			this.charLen = descriptor.length() - 2;
		}
	}

	private void inflateClassDesc() {
		String internalName = Utils.toInternalName(this.classDesc);
		assert this.charLen = internalName.length();
		this.stringValue = internalName;
		this.hash = hashString(internalName.hashCode());
		this.state = State.STRING;
	}

	@Override
	public Utf8EntryImpl clone(ConstantPoolBuilder cp) {
		if (cp.canWriteDirect(constantPool))
			return this;
		if (state = State.CLASS_DESC) {
			return (Utf8EntryImpl) cp.utf8Entry(classDesc);
		}
		return (state == State.STRING && rawBytes == null)
			? (Utf8EntryImpl) cp.utf8Entry(stringValue)
			: ((SplitConstantPool) cp).maybeCloneUtf8Entry(this);
	}

	@Override
	public int hashCode() {
		if (state == State.CLASS_DESC) {
			inflateClassDesc();
		}
		...
	}

	@Override
	public String toString() {
		if (state == State.CLASS_DESC) {
			inflateClassDesc();
		}
		...
	}

	@Override
	public char charAt(int index) {
		if (state == State.CLASS_DESC) {
			return classDesc.descriptorString()
				.charAt(Objects.checkIndex(index, charLen) + 1);
		}
		...
	}

	@Override
	public boolean equalsString(String s) {
		if (state == State.CLASS_DESC) {
			inflateClassDesc();
		}
		...
	}

	@Override
	public void writeTo(BufWriter pool) {
		if (rawBytes != null) {
			...
		}
		else {
			if (state == State.CLASS_DESC) {
				inflateClassDesc();
			}
			...
		}
	}
}
```

asotona:
Problem with ClassDesc "embedding" into the Utf8Entry is with its ambiguity.
Utf8Entry mainly contains some name or descriptor (MethodTypeDesc, ClassDesc, PackageDesc, ModuleDesc, etc...) or signature (Signature, MethodSignature, ClassSignature, etc...).
And even for ClassDesc there is ambiguity of the serialized form. When the Utf8Entry is related to ClassEntry it contains descriptor for arrays or internal name for classes. However when the Utf8Entry is related for example to annotation it contains always class descriptor (even for classes).
From this perspective the Utf8Entry should not be responsible for conversions to and from symbols, because it does not know the context and so the right serialised form.
ClassEntry knows the  exact form of its Utf8Entry, Annotation, MethodInfo or (with just a minor ambiguity) also NameAndTypeEntry know which symbols to use and how to serialize/deserialize them.
I think that having Utf8Entry as a big central dispatcher for so many types of symbols and serialisation forms is not the best design and will include a lot of complexity and confusion. 
Current architecture lets interaction with symbols responsibility on the right layer.

 It is unlikely that one Utf8Entry will hold two different symbols, however still possible (for example when Signature and ClassDesc are the same).
And it is still very likely that one symbol describing the same class will be stored in two different Utf8Entries (for example the same class stored as ClassDesc vs. as annotation class).

src/java.base/share/classes/jdk/internal/classfile/impl/StackMapGenerator.java
```diff
@@ -732,13 +743,13 @@ private void processSwitch(RawBytecodeHelper bcs) {
     }
 
     private void processFieldInstructions(RawBytecodeHelper bcs) {
-        var desc = ClassDesc.ofDescriptor(((MemberRefEntry)cp.entryByIndex(bcs.getIndexU2())).nameAndType().type().stringValue());
+        var desc = ((AbstractPoolEntry.NameAndTypeEntryImpl)((MemberRefEntry)cp.entryByIndex(bcs.getIndexU2())).nameAndType()).fieldTypeSymbol();
```
liach, line 746:
```suggestion
        var desc = Util.fieldTypeSymbol((MemberRefEntry)cp.entryByIndex(bcs.getIndexU2())).nameAndType());
```

src/java.base/share/classes/jdk/internal/classfile/instruction/InvokeInstruction.java
```diff
@@ -89,7 +90,7 @@ default Utf8Entry type() {
      * {@return a symbolic descriptor for the method type}
      */
     default MethodTypeDesc typeSymbol() {
-        return MethodTypeDesc.ofDescriptor(type().stringValue());
+        return ((AbstractPoolEntry.NameAndTypeEntryImpl)method().nameAndType()).methodTypeSymbol();
```
liach, line 93:
```suggestion
        return Util.methodTypeSymbol(method().nameAndType());
```

cl4es (commented):
I'm having a look, but hit an early hurdle in that the microbenchmarks under org/openjdk/bench/jdk/classfile all seem to be missing necessary `--add-opens` flags. Micros should be runnable without any extra flags on the command line.

test/micro/org/openjdk/bench/jdk/classfile/RebuildMethodBodies.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.List;
+import jdk.internal.classfile.ClassModel;
+import jdk.internal.classfile.Classfile;
+import jdk.internal.classfile.ClassTransform;
+import jdk.internal.classfile.instruction.*;
+import org.openjdk.jmh.annotations.*;
+
+@BenchmarkMode(Mode.Throughput)
+@State(Scope.Benchmark)
+@Fork(value = 1, jvmArgsAppend = {"--enable-preview"})
```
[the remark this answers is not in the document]
liach: They are added in the `make/RunTests.gmk`: https://github.com/openjdk/jdk/pull/13550/files#diff-041bf69ea79b333b9ce99c1f879e398d698538530a35c361500b72631f059233R599-R608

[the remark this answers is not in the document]
cl4es: That seems misplaced. Please file an RFE to have this cleaned up. 

Each microbenchmark that has to add opens needs to take responsibility for that themselves and not change the environment for everything else. And all micros the benchmarks.jar should be runnable standalone, not rely on quirks in the make runner.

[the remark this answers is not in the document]
asotona: All benchmarks fixed and --add-exports removed from make file.
Thank you for pointing this out.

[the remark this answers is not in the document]
cl4es: Thanks for fixing this!

test/micro/org/openjdk/bench/jdk/classfile/RebuildMethodBodies.java
```diff
@@ -0,0 +1,115 @@
...
+@Measurement(iterations = 4)
+public class RebuildMethodBodies {
+
+    List<ClassModel> shared, unshared;
+    Iterator<ClassModel> it1, it2;
+
+    @Setup(Level.Trial)
+    public void setup() throws IOException {
+        shared = new LinkedList<>();
+        unshared = new LinkedList<>();
```
liach, line 57: LinkedList should be replaced by ArrayList, as:
1. LinkedList's node wrapper around each object is too heavy, ArrayList has smaller wrapper
2. LinkedList iteration needs to follow links while ArrayList access can be machine optimized
3. ArrayList addition is amortized O(1), not really worse than LinkedList additions.

cl4es: I have to side with @liach here: `LinkedList` iteration is significantly slower than `ArrayList` even though the computational complexity is identical. Often by an integer faction. This is due to the sparseness of the linked list data structure on heap and the pointer chasing that entails.  

For minimum overhead of iteration over an immutable list then turning the list into an immutable via `List.copyOf` might be preferable due the JVM's ability to optimize over `@Stable` arrays.

Adhoc [JMH benchmark](https://gist.github.com/cl4es/1f21812241c47f32a9deaffcc996e8b3):

```
Benchmark                            Mode  Cnt    Score    Error   Units
ListIteration.iterateArrayList      thrpt   15  188,724 ± 10,903  ops/ms
ListIteration.iterateImmutableList  thrpt   15  230,901 ±  6,513  ops/ms
ListIteration.iterateLinkedList     thrpt   15   58,289 ±  0,497  ops/ms
```

Is it important to fix this? Perhaps not, but in a microbenchmark like this the fewer cycles we spend on "stuff" that isn't really part of the thing we're testing, the better. Increasingly so as the thing we're testing is getting more and more optimized.

test/micro/org/openjdk/bench/jdk/classfile/RepeatedModelTraversal.java
```diff
@@ -0,0 +1,71 @@
...
+@Warmup(iterations = 3)
+@Measurement(iterations = 4)
+public class RepeatedModelTraversal {
+
+    List<ClassModel> models;
+    Iterator<ClassModel> it;
+
+    @Setup(Level.Trial)
+    public void setup() throws IOException {
+        models = new LinkedList<>();
```
Glavo, line 54: I think `ArrayList` can be used instead of `LinkedList` here too. 

While this may not be an actual improvement, I don't think `LinkedList` should be used unless necessary.

test/micro/org/openjdk/bench/jdk/classfile/RepeatedModelTraversal.java
```diff
@@ -0,0 +1,71 @@
...
+        Files.walk(FileSystems.getFileSystem(URI.create("jrt:/")).getPath("modules/java.base/java/util")).forEach(p ->  {
```
turbanoff, line 55:
```suggestion
        Files.walk(FileSystems.getFileSystem(URI.create("jrt:/")).getPath("modules/java.base/java/util")).forEach(p -> {
```

src/java.base/share/classes/jdk/internal/classfile/impl/StackMapDecoder.java
```diff
@@ -63,31 +63,30 @@ public class StackMapDecoder {
...
-        for(var arg : mdesc.parameterList()) {
-            vtis[i++] = switch (arg.descriptorString()) {
-                case "I", "S", "C" ,"B", "Z" ->  SimpleVerificationTypeInfo.ITEM_INTEGER;
-                case "J" -> SimpleVerificationTypeInfo.ITEM_LONG;
-                case "F" -> SimpleVerificationTypeInfo.ITEM_FLOAT;
-                case "D" -> SimpleVerificationTypeInfo.ITEM_DOUBLE;
-                case "V" -> throw new IllegalArgumentException("Illegal method argument type: " + arg);
+        for(var arg : methodType.parameterList()) {
+            vtis[i++] = switch (arg.descriptorString().charAt(0)) {
+                case 'I', 'S', 'C' ,'B', 'Z' ->  SimpleVerificationTypeInfo.ITEM_INTEGER;
```
turbanoff, line 85:
```suggestion
                case 'I', 'S', 'C' ,'B', 'Z' -> SimpleVerificationTypeInfo.ITEM_INTEGER;
```

test/micro/org/openjdk/bench/jdk/classfile/RebuildMethodBodies.java
```diff
@@ -0,0 +1,115 @@
...
+public class RebuildMethodBodies {
+
+    List<ClassModel> shared, unshared;
+    Iterator<ClassModel> it1, it2;
+
+    @Setup(Level.Trial)
+    public void setup() throws IOException {
+        shared = new ArrayList<>();
+        unshared = new ArrayList<>();
+        Files.walk(FileSystems.getFileSystem(URI.create("jrt:/")).getPath("modules/java.base/java")).forEach(p ->  {
```
turbanoff, line 58:
```suggestion
        Files.walk(FileSystems.getFileSystem(URI.create("jrt:/")).getPath("modules/java.base/java")).forEach(p -> {
```

test/micro/org/openjdk/bench/jdk/classfile/RebuildMethodBodies.java
asotona: Right, fixed, thanks.

test/micro/org/openjdk/bench/jdk/classfile/RepeatedModelTraversal.java
asotona: I've replaced all `LinkedList` with `ArrayList` in the benchmarks, thanks.

src/java.base/share/classes/jdk/internal/classfile/impl/StackMapDecoder.java
```diff
@@ -63,31 +63,30 @@ public class StackMapDecoder {
...
+            vtis = new VerificationTypeInfo[methodType.parameterCount()];
         }
-        for(var arg : mdesc.parameterList()) {
-            vtis[i++] = switch (arg.descriptorString()) {
-                case "I", "S", "C" ,"B", "Z" ->  SimpleVerificationTypeInfo.ITEM_INTEGER;
-                case "J" -> SimpleVerificationTypeInfo.ITEM_LONG;
-                case "F" -> SimpleVerificationTypeInfo.ITEM_FLOAT;
-                case "D" -> SimpleVerificationTypeInfo.ITEM_DOUBLE;
-                case "V" -> throw new IllegalArgumentException("Illegal method argument type: " + arg);
+        for(var arg : methodType.parameterList()) {
```
cl4es, line 83: `MethodTypeDesc::parameterList` copies the internal `ClassDesc[]`, so desugaring loop from `0` to `mdesc.parameterCount()` might be a small improvement for this and a few other cases.

cl4es (approved):
Looks like a good set of improvements for throughput cases. 

For the future I think we should probably add something that measures startup/warmup overheads of the classfile API, since it's likely that usage pattern in many applications will be heavy during bootstrap and then subside (similar to how ASM usage can be startup sensitive but will then disappear from sight once an application is up and running)

src/java.base/share/classes/jdk/internal/classfile/impl/StackMapDecoder.java
liach: I plan to share a backing list in #13186, with API additions so users can avoid copying their input immutable parameter list as well. I see no reason to use a direct array, for MTD is not passed to and used by the VM, unlike MethodType.

cl4es: Good, I think it makes perfect sense for the MTD parameter list to be immutable and shareable without copying.

asotona: Thank you for the review.
Yes, I'm looking forward to see it together with Constants API improvements in one benchamark ;)

liach: Since you are available, can you help review my CSR request at https://bugs.openjdk.org/browse/JDK-8306699 as requested by Joe Darcy so we can get this improvement into release 21?

test/micro/org/openjdk/bench/jdk/classfile/Write.java
```diff
@@ -65,7 +65,17 @@
...
-@Fork(1)
+@Fork(value = 1, jvmArgsAppend = {
+        "--enable-preview",
+        "--add-exports", "java.base/jdk.internal.org.objectweb.asm=ALL-UNNAMED",
+        "--add-exports", "java.base/jdk.internal.org.objectweb.asm.tree=ALL-UNNAMED",
+        "--add-exports", "java.base/jdk.internal.classfile=ALL-UNNAMED",
+        "--add-exports", "java.base/jdk.internal.classfile.attribute=ALL-UNNAMED",
+        "--add-exports", "java.base/jdk.internal.classfile.constantpool=ALL-UNNAMED",
+        "--add-exports", "java.base/jdk.internal.classfile.instruction=ALL-UNNAMED",
+        "--add-exports", "java.base/jdk.internal.classfile.java.lang.constant=ALL-UNNAMED",
```
liach, line 76: Since this package is gone, this export arg can be removed. Same for other benchmarks.

test/micro/org/openjdk/bench/jdk/classfile/GenerateStackMaps.java
```diff
@@ -55,7 +55,15 @@
...
 @BenchmarkMode(Mode.Throughput)
 @State(Scope.Benchmark)
-@Fork(1)
+@Fork(value = 1, jvmArgsAppend = {
+        "--enable-preview",
+        "--add-exports", "java.base/jdk.internal.classfile=ALL-UNNAMED",
+        "--add-exports", "java.base/jdk.internal.classfile.attribute=ALL-UNNAMED",
+        "--add-exports", "java.base/jdk.internal.classfile.constantpool=ALL-UNNAMED",
+        "--add-exports", "java.base/jdk.internal.classfile.instruction=ALL-UNNAMED",
+        "--add-exports", "java.base/jdk.internal.classfile.java.lang.constant=ALL-UNNAMED",
```
liach, line 64: Same here

test/micro/org/openjdk/bench/jdk/classfile/AbstractCorpusBenchmark.java
```diff
@@ -45,7 +45,17 @@
...
-@Fork(1)
+@Fork(value = 1, jvmArgsAppend = {
+        "--enable-preview",
+        "--add-exports", "java.base/jdk.internal.org.objectweb.asm=ALL-UNNAMED",
+        "--add-exports", "java.base/jdk.internal.org.objectweb.asm.tree=ALL-UNNAMED",
+        "--add-exports", "java.base/jdk.internal.classfile=ALL-UNNAMED",
+        "--add-exports", "java.base/jdk.internal.classfile.attribute=ALL-UNNAMED",
+        "--add-exports", "java.base/jdk.internal.classfile.constantpool=ALL-UNNAMED",
+        "--add-exports", "java.base/jdk.internal.classfile.instruction=ALL-UNNAMED",
+        "--add-exports", "java.base/jdk.internal.classfile.java.lang.constant=ALL-UNNAMED",
```
liach, line 56: Here

test/micro/org/openjdk/bench/jdk/classfile/Write.java
asotona: Good catch, thanks!

PR #13672, opened 2023-04-26
8306933: C2: "assert(false) failed: infinite loop" failure

rwestrel:
The assert fires because an infinite loop appears in the graph after
loop opts are over.

After loop opts, the `for(;;)` loop contains a null check and a range
check for `array[i]`. So it's not considered an infinite loop (it has
exits to uncommon traps). The null check and range check are redundant
with the one right before the loop: `int v = array2[k];` IGVN can
optimize it but it doesn't happen until after loop opts when a
`ConvI2L` for the `array[i]` access is processed as part of post loop
opts IGVN. The `for(;;)` loop is then emptied and only contains a
`Loop` and a `Safepoint` nodes.

I propose removing the assert (at least for now) as I don't see a way
to guarantee no infinite loop can appear after loop opts.

---------
### Progress

### Issue
 * [JDK-8306933](https://bugs.openjdk.org/browse/JDK-8306933): C2: "assert(false) failed: infinite loop" failure

### Reviewers
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - **Reviewer**) ⚠️ Review applies to [f943db9d](https://git.openjdk.org/jdk/pull/13672/files/f943db9d2269cb2cf7c95e4e04f650a08b4768de)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13672/head:pull/13672` \
`$ git checkout pull/13672`

`$ git checkout pull/13672` \
`$ git pull https://git.openjdk.org/jdk.git pull/13672/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13672`

`$ git pr show -t 13672`

</details>

<a href="https://git.openjdk.org/jdk/pull/13672.diff">https://git.openjdk.org/jdk/pull/13672.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13672#issuecomment-1523657745)

### DIFF
--- a/test/hotspot/jtreg/compiler/c2/TestInfiniteLoopCompilationFailure.java
+++ b/test/hotspot/jtreg/compiler/c2/TestInfiniteLoopCompilationFailure.java
@@ -26,9 +26,9 @@
  * @bug 8306933
  * @summary C2: "assert(false) failed: infinite loop" failure
  * @run main/othervm -Xcomp -XX:CompileOnly=TestInfiniteLoopCompilationFailure::test -XX:-UseLoopPredicate -XX:-UseProfiledLoopPredicate
- *                   -XX:+StressIGVN -XX:StressSeed=675320863 TestInfiniteLoopCompilationFailure
+ *                   -XX:+UnlockDiagnosticVMOptions -XX:+StressIGVN -XX:StressSeed=675320863 TestInfiniteLoopCompilationFailure
  * @run main/othervm -Xcomp -XX:CompileOnly=TestInfiniteLoopCompilationFailure::test -XX:-UseLoopPredicate -XX:-UseProfiledLoopPredicate
- *                   -XX:+StressIGVN TestInfiniteLoopCompilationFailure
+ *                   -XX:+UnlockDiagnosticVMOptions -XX:+StressIGVN TestInfiniteLoopCompilationFailure
  */
 
 public class TestInfiniteLoopCompilationFailure {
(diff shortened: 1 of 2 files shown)


test/hotspot/jtreg/compiler/c2/TestInfiniteLoopCompilationFailure.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+/*
+ * @test
+ * @bug 8306933
+ * @summary C2: "assert(false) failed: infinite loop" failure
+ * @run main/othervm -Xcomp -XX:CompileOnly=TestInfiniteLoopCompilationFailure::test -XX:-UseLoopPredicate -XX:-UseProfiledLoopPredicate
+ *                   -XX:+StressIGVN -XX:StressSeed=675320863 TestInfiniteLoopCompilationFailure
+ * @run main/othervm -Xcomp -XX:CompileOnly=TestInfiniteLoopCompilationFailure::test -XX:-UseLoopPredicate -XX:-UseProfiledLoopPredicate
+ *                   -XX:+StressIGVN TestInfiniteLoopCompilationFailure
```
chhagedorn, line 31: Since `StressIGVN` is diagnostic, you need to add `-XX:+UnlockDiagnosticVMOptions` here.

chhagedorn (changes requested):
That looks reasonable!

TobiHartmann (approved):
Looks good to me too (besides the missing `-XX:+UnlockDiagnosticVMOptions`).

chhagedorn (approved):
Update looks good!

rwestrel:
@TobiHartmann @chhagedorn thanks for the reviews

PR #13679, opened 2023-04-26
8159337: Introduce a method in Locale class to return the language tags as per RFC 5646 convention

justin-curtis-lu:
Please review this PR which adds the method `caseFoldLanguageTag(String languageTag)` to java.util.Locale.

This method case folds a language tag to adhere to _[section 2.1.1. Formatting of Language Tags of RFC5646](https://www.rfc-editor.org/rfc/rfc5646.html#section-2.1)_. This format is defined as _"All subtags, including extension and private use subtags, use lowercase letters with two exceptions: two-letter and four-letter subtags that neither appear at the start of the tag nor occur after singletons.  Such two-letter subtags are all uppercase ... and four-letter subtags are titlecase."_.

In order to match the behavior of existing language tag related Locale methods, this method matches the 2.1.1 RFC5646 specification with the following **exceptions**:
- Will not case fold variant subtags
- Will not case fold private use subtags prefixed by "lvariant"

As an example, `caseFoldLanguageTag("ja-kana-jp-x-lvariant-Oracle-JDK-Standard-Edition")` returns _"ja-Kana-JP-x-lvariant-Oracle-JDK-Standard-Edition"_. Further examples can be seen in the test file.

---------
### Progress
- [x] Change requires CSR request [JDK-8305623](https://bugs.openjdk.org/browse/JDK-8305623) to be approved

### Issues
 * [JDK-8159337](https://bugs.openjdk.org/browse/JDK-8159337): Introduce a method in Locale class to return the language tags as per RFC 5646 convention
 * [JDK-8305623](https://bugs.openjdk.org/browse/JDK-8305623): Introduce a method in Locale class to return the language tags as per RFC 5646 convention (**CSR**)

### Reviewers
 * [Naoto Sato](https://openjdk.org/census#naoto) (@naotoj - **Reviewer**) ⚠️ Review applies to [96781281](https://git.openjdk.org/jdk/pull/13679/files/967812819d722e46397d2b04d04147f933cbe591)
 * [Roger Riggs](https://openjdk.org/census#rriggs) (@RogerRiggs - **Reviewer**) ⚠️ Review applies to [96781281](https://git.openjdk.org/jdk/pull/13679/files/967812819d722e46397d2b04d04147f933cbe591)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13679/head:pull/13679` \
`$ git checkout pull/13679`

`$ git checkout pull/13679` \
`$ git pull https://git.openjdk.org/jdk.git pull/13679/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13679`

`$ git pr show -t 13679`

</details>

<a href="https://git.openjdk.org/jdk/pull/13679.diff">https://git.openjdk.org/jdk/pull/13679.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13679#issuecomment-1524122583)


test/jdk/java/util/Locale/CaseFoldLanguageTagTest.java
```diff
@@ -0,0 +1,168 @@
...
+ * <a href="https://www.rfc-editor.org/rfc/rfc5646.html#section-2.1">Tag Syntax</a>).
+ * In addition, the method is tested to ensure that IllegalArgumentException and
+ * NullPointerException are thrown given the right circumstances.
+ */
+public class caseFoldLanguageTagTest {
+
+    @ParameterizedTest
+    @MethodSource("wellFormedTags")
+    public void TestWellFormedTags(String tag, String foldedTag) {
+        assertEquals(foldedTag, Locale.caseFoldLanguageTag(tag), String.format("Folded %s", tag));
```
naotoj, line 56: Would it be helpful if both the expected and the result are recorded?

test/jdk/java/util/Locale/CaseFoldLanguageTagTest.java
```diff
@@ -0,0 +1,168 @@
...
+ * grandfathered, and irregular. For more info, see the following,
+ * <a href="https://www.rfc-editor.org/rfc/rfc5646.html#section-2.1">Tag Syntax</a>).
+ * In addition, the method is tested to ensure that IllegalArgumentException and
+ * NullPointerException are thrown given the right circumstances.
+ */
+public class caseFoldLanguageTagTest {
+
+    @ParameterizedTest
+    @MethodSource("wellFormedTags")
+    public void TestWellFormedTags(String tag, String foldedTag) {
```
naotoj, line 55: Nit: method name should start with a lowercase char

test/jdk/java/util/Locale/CaseFoldLanguageTagTest.java
```diff
@@ -0,0 +1,168 @@
...
+        assertEquals(foldedTag, Locale.caseFoldLanguageTag(tag), String.format("Folded %s", tag));
+    }
+
+    @ParameterizedTest
+    @MethodSource("illFormedTags")
+    public void TestIllFormedTags(String tag) {
+        try {
+            Locale.caseFoldLanguageTag(tag);
+            throw new RuntimeException("$$$ Failure: ill formed tags should throw IAE");
+        } catch (IllegalArgumentException ignored) {}
```
naotoj, line 65: `IllformedLocaleException`? Or if the test passes with this version of impl, should the test throw an RuntimeException? Also instead of try-catch, `assertThrows()` can be used.

test/jdk/java/util/Locale/CaseFoldLanguageTagTest.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    @Test
+    public void TestNPE() {
+        try {
+            Locale.caseFoldLanguageTag(null);
+            throw new RuntimeException("$$$ Failure: NPE should be thrown when tag is null");
```
naotoj, line 72: `assertThrows()` can be used here as well

naotoj (commented):
Looks good. Some minor comments follow:

test/jdk/java/util/Locale/CaseFoldLanguageTagTest.java
justin-curtis-lu: The test does not pass like this, I had forgotten to re run the test after originally switching IAE to IFE in the method itself. (and hence why I did not catch that it is not updated in the test). 

I have switched all the IAE to IFE in the test, and also used assertThrows() as recommended

test/jdk/java/util/Locale/CaseFoldLanguageTagTest.java
justin-curtis-lu: Junit assertEquals provides the actual result in the failure, so no need to record it.

Example failure:
`org.opentest4j.AssertionFailedError: Folded ABC ==> expected: <ab> but was: <abc>`

src/java.base/share/classes/sun/util/locale/LanguageTag.java
```diff
@@ -414,6 +409,48 @@ private boolean parsePrivateuse(StringTokenIterator itr, ParseStatus sts) {
         return found;
     }
 
+    public static String caseFoldTag(String tag) {
+        // Illegal tags
+        if (!parse(tag, null).wellFormed) {
+            throw new IllformedLocaleException("Ill formed tag");
```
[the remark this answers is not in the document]
justin-curtis-lu: Good point, I have updated it so that it provides the sts error message with the exception.

For example,
`Ill formed tag: Invalid subtag: xabadadoo`

src/java.base/share/classes/java/util/Locale.java
```diff
@@ -1754,7 +1806,7 @@ public String toLanguageTag() {
      *
      * <p>Legacy tags with canonical replacements are as follows:
      *
-     * <table class="striped">
+     * <table class="striped" id="legacy_tags_replacement">
```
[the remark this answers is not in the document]
justin-curtis-lu: Thanks for the review and helpful comments, I have addressed this comment and the other ones you left.

src/java.base/share/classes/sun/util/locale/LanguageTag.java
```diff
@@ -414,6 +407,50 @@ private boolean parsePrivateuse(StringTokenIterator itr, ParseStatus sts) {
         return found;
     }
 
+    public static String caseFoldTag(String tag) {
+        LanguageTag parsedTag = parse(tag, null);
```
naotoj, line 411: I think we don't need to create a field `parseSts` but provide a local one here and use it in `parse(tag, sts), as that parse status is only referred to within this method.

test/jdk/java/util/Locale/CaseFoldLanguageTagTest.java
naotoj: Ah, ok then

src/java.base/share/classes/sun/util/locale/LanguageTag.java
```diff
@@ -414,6 +407,50 @@ private boolean parsePrivateuse(StringTokenIterator itr, ParseStatus sts) {
...
+            return LEGACY.get(potentialLegacy)[0];
+        }
+        // Non-legacy tags
+        StringBuilder bldr = new StringBuilder();
+        String[] subtags = tag.split("-");
+        boolean privateFound = false;
+        boolean singletonFound = false;
+        boolean privUseVarFound = false;
+        for (int i = 0; i < subtags.length; i++) {
+            if (privUseVarFound) {
```
RogerRiggs, line 429: The following block of code might be easier to read if subtags[i] was put in a local.

src/java.base/share/classes/sun/util/locale/LanguageTag.java
```diff
@@ -414,6 +407,50 @@ private boolean parsePrivateuse(StringTokenIterator itr, ParseStatus sts) {
...
+            // If subtag is not 2 letter, 4 letter, or variant
+            // under the right conditions, then it should be lower-case
+            } else {
+                if (isPrivateusePrefix(subtags[i])) {
+                    privateFound = true;
+                } else if (isExtensionSingleton(subtags[i])) {
+                    singletonFound = true;
+                } else if (subtags[i].equals(PRIVUSE_VARIANT_PREFIX)) {
+                    privUseVarFound = true;
+                }
```
RogerRiggs, line 446: These flags can become true but inside the loop cannot become false again.
Is that correct?  For example, I think there can be multiple extension singletons.

test/jdk/java/util/Locale/CaseFoldLanguageTagTest.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    @Test
+    public void throwNPE() {
+        assertThrows(NullPointerException.class, () ->
+                Locale.caseFoldLanguageTag(null));
+    }
+
+    private static Stream<Arguments> wellFormedTags() {
+        return Stream.of(
```
RogerRiggs, line 75: There are no test cases with multiple singleton extensions either as valid or invalid tests.

src/java.base/share/classes/sun/util/locale/LanguageTag.java
justin-curtis-lu: That was intentional as the booleans represent if a private prefix, singleton prefix, or "lvariant" has been found in general (not that the current position is the suffix of one of those prefixes).

Consider if there is a multiple extension singleton (eg: `a-BBB-CCC-DDD`)

Once singletonFound is set to `true`, it is assumed all further subtags should be lowercased unless it is a private use prefixed by "lvariant". Thus it would be folded to `a-bbb-ccc-ddd` as expected.

Did that address your point?

justin-curtis-lu:
Hi @RogerRiggs, thanks for the review. 

I have dropped the wildcard import (and disabled it on IntellJ for the future), made subtags[i] a local var, added tests with multiple extension singletons, and pre allocated the string builder.

naotoj (approved):

RogerRiggs (approved):
Looks good.

PR #13681, opened 2023-04-26
8306950: [REDO] JDK-8305252 make_method_handle_intrinsic may call java code under a lock

coleenp:
This change redoes the intrinsic method table change to not hold a lock while calling make_method_handle_intrinsic.  The first commit is the original patch and the second is the fix to the bug it caused.
The Atomic operations might not be needed.
Tested with tier1-7, where tests in multiple tiers were regularly failing.

---------
### Progress

### Issue
 * [JDK-8306950](https://bugs.openjdk.org/browse/JDK-8306950): [REDO] JDK-8305252 make_method_handle_intrinsic may call java code under a lock

### Reviewers
 * [David Holmes](https://openjdk.org/census#dholmes) (@dholmes-ora - **Reviewer**) ⚠️ Review applies to [515d3ef7](https://git.openjdk.org/jdk/pull/13681/files/515d3ef7f68db9b287667940f7b7531531dede5d)
 * [Matias Saavedra Silva](https://openjdk.org/census#matsaave) (@matias9927 - Committer)
 * [Ioi Lam](https://openjdk.org/census#iklam) (@iklam - **Reviewer**) ⚠️ Review applies to [65a9503d](https://git.openjdk.org/jdk/pull/13681/files/65a9503db086c2038fb3d9f8df28b98fe3bbc23f)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13681/head:pull/13681` \
`$ git checkout pull/13681`

`$ git checkout pull/13681` \
`$ git pull https://git.openjdk.org/jdk.git pull/13681/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13681`

`$ git pr show -t 13681`

</details>

<a href="https://git.openjdk.org/jdk/pull/13681.diff">https://git.openjdk.org/jdk/pull/13681.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13681#issuecomment-1524216606)

### DIFF
--- a/src/hotspot/share/classfile/systemDictionary.cpp
+++ b/src/hotspot/share/classfile/systemDictionary.cpp
@@ -1583,11 +1583,13 @@ void SystemDictionary::methods_do(void f(Method*)) {
   }
 
   auto doit = [&] (InvokeMethodKey key, Method* method) {
-    f(method);
+    if (method != nullptr) {
+      f(method);
+    }
   };
 
   {
-    MutexLocker ml(InvokeMethodTable_lock);
+    MutexLocker ml(InvokeMethodIntrinsicTable_lock);
     _invoke_method_intrinsic_table.iterate_all(doit);
   }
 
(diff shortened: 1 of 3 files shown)


dholmes-ora (approved):
Re-do looks good. Sorry we missed the issue in the initial review.

Thanks.

src/hotspot/share/classfile/systemDictionary.cpp
```diff
@@ -1950,14 +1950,16 @@ Method* SystemDictionary::find_method_handle_intrinsic(vmIntrinsicID iid,
   // Method has been added as the value.
   {
     MonitorLocker ml(THREAD, InvokeMethodIntrinsicTable_lock);
-    while (met == nullptr) {
+    while (true) {
```
matias9927, line 1953: Given that you break out of this loop based on the state of `created`, it might be better to change the loop to `while(!created)`.

matias9927 (changes requested):

src/hotspot/share/classfile/systemDictionary.cpp
coleenp: I agree, that does look better.  Thanks!

matias9927 (approved):
Changes look good! Thanks!

iklam (changes requested):
Looks good, but I think we should remove the unnecessary null check. Since this is tricky code, we should make the logic as simple as possible.

coleenp:
Thanks David for the review, and Matias and Ioi for the comments.  In an abundance of caution, I'm rerunning tiers 4-7 on the changes.

iklam (approved):

src/hotspot/share/classfile/systemDictionary.cpp
```diff
@@ -1939,40 +1941,66 @@ Method* SystemDictionary::find_method_handle_intrinsic(vmIntrinsicID iid,
...
   {
-    MutexLocker ml(THREAD, InvokeMethodTable_lock);
-    InvokeMethodKey key(signature, iid_as_int);
-    Method** met = _invoke_method_intrinsic_table.get(key);
-    if (met != nullptr) {
-      return *met;
+    MonitorLocker ml(THREAD, InvokeMethodIntrinsicTable_lock);
+    bool created = false;
+    // Loop until the current thread won the race and will try to create the full entry.
+    while (!created) {
```
ExE-Boss, line 1955: Maybe use:
```cpp
do { ... } while (!created);
```

src/hotspot/share/classfile/systemDictionary.cpp
dholmes-ora: No it doesn't! Sorry but there are two reasons for getting out of the loop - one involves a return and one involves the break - only putting one of those in the loop condition doesn't clarify anything IMO.

src/hotspot/share/classfile/systemDictionary.cpp
```diff
@@ -1939,40 +1941,65 @@ Method* SystemDictionary::find_method_handle_intrinsic(vmIntrinsicID iid,
...
+      met = _invoke_method_intrinsic_table.put_if_absent(key, &created);
+      if (met != nullptr && *met != nullptr) {
```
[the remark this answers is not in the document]
dholmes-ora: Good catch

src/hotspot/share/classfile/systemDictionary.cpp
coleenp: So I have 4 reviewers and 3 different opinions, on something that people's eyes differ on but are the same thing.

src/hotspot/share/classfile/systemDictionary.cpp
coleenp: There's another variation of the same code.

src/hotspot/share/classfile/systemDictionary.cpp
coleenp: By that logic, all loops should be while (true) with a break in the middle for the condition?

dholmes-ora: As you know we spent some time trying to express this loop correctly and using the "obvious" loop conditions did not work because we had to re-check the condition after the loop to figure out whether we needed to return or not. The `while(true)` captured the fact there there was no simple/obvious loop termination condition here and that you need to check the loop body to see exactly what conditions exit the loop and in what way. Picking one of those conditions gives a false impression about the operation of the loop IMO and doesn't aid in understandability. YMMV.

coleenp: It was a lot of work to reason about this loop.  I changed it back.

matias9927 (approved):
Updates look good!

coleenp:
The n-1th version passes tier4-7 again, and the nth version already passed 1-7.  Thanks for reviewing David, Matias and Ioi.
/integrate

PR #13686, opened 2023-04-27
8306980: Generated docs should contain correct GPLv2 Legal Documents

tkiriyama:
I modified the location from which javadoc copies some legal files to the generated documentation.  If --legal-notices　option is set to default or nothing is specified,, GPLv2 Legal Documents are copied from legal/java.base/ directory, such as LICENSE, ADDITIONAL_LICENSE_INFO and ASSEMBLY_EXCEPTION.

Would you please review this fix?

---------
### Progress

### Integration blocker
&nbsp;⚠️ Title mismatch between PR and JBS for issue [JDK-8306980](https://bugs.openjdk.org/browse/JDK-8306980)

### Issue
 * [JDK-8306980](https://bugs.openjdk.org/browse/JDK-8306980): Generated docs should contain correct Legal Documents (**Bug** - P3) ⚠️ Title mismatch between PR and JBS. ⚠️ Issue is already resolved. Consider making this a "backport pull request" by setting the PR title to `Backport <hash>` with the hash of the original commit. See [Backports](https://wiki.openjdk.org/display/SKARA/Backports).

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13686/head:pull/13686` \
`$ git checkout pull/13686`

`$ git checkout pull/13686` \
`$ git pull https://git.openjdk.org/jdk.git pull/13686/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13686`

`$ git pr show -t 13686`

</details>

<a href="https://git.openjdk.org/jdk/pull/13686.diff">https://git.openjdk.org/jdk/pull/13686.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13686#issuecomment-1525106357)


test/langtools/jdk/javadoc/doclet/testLegalNotices/TestLegalNotices.java
```diff
@@ -103,27 +105,47 @@ void test(Path base, Path src, Path legal, OptionKind optionKind, IndexKind inde
         }
         javadoc(args.toArray(new String[0]));
 
-        Set<Path> expectFiles = getExpectFiles(optionKind, indexKind, legal);
-        Set<Path> foundFiles = listFiles(out.resolve("legal"));
+        Set<File> expectFiles = getExpectFiles(optionKind, indexKind, legal);
+        Set<File> foundFiles = listFiles(out.resolve("legal"));
```
jonathan-gibbons, line 109: Do not change return type from `Path` to `File`

jonathan-gibbons (changes requested):
The change to use a mixture of `Path` and `File` is bizarre and mostly gratuitous and unnecessary.
Please stay using `nio.Path` objects.

tkiriyama:
@jonathan-gibbons 
Thank you for your review.
As your advice, I fixed the test to not use java.io.File.

jonathan-gibbons (changes requested):
This seems like the wrong solution.

It does not seem right for javadoc to rummage around finding legal files that may or may not be the right ones to use.   By design, it is supposed to just copy the files from a directory that has been provided. If the files in that directory are wrong or missing, we should fix that, not fix javadoc to go looking elsewhere.

tkiriyama:
Bundling duplicate files is inefficient.
Since the existing design references files under java.base, isn't it reasonable to use those files?

jonathan-gibbons:
> Bundling duplicate files is inefficient.
> Since the existing design references files under java.base, isn't it reasonable to use those files?

While the files being copied are represented in the JDK image with symbolic links, these links are (or should be) implicitly followed when the files are copied into the generated documentation.

jonathan-gibbons:
If you have evidence that symbolic links are being copied to the generated documentation, then the correct fix would be to ensure that any symbolic links **are** followed, and not to simulate following the links as suggested in this PR.

However, I would want to see evidence that symbolic links are ending up in generated documentation before pursuing this any further.

tkiriyama:
As you said, there were no problems with platforms other than Windows.
However, as mentioned in the bug issue, on Windows, symbolic links are not supported and the contents of some files like "ADDITIONAL_LICENSE_INFO" are not included in the generated documentation. This fix will ensure that the contents of the files are reliably included  in the generated documentation.

jonathan-gibbons (changes requested):
I still (strongly) dislike baking such specific knowledge of the legal files into the javadoc tool. That's just conceptually wrong.

It turns out these files are generated by the `jlink` tool, which somewhat weirdly generates these "descriptive links" to use instead of symbolic links on platforms that do not support symbolic links.

See [jlink DefaultImageBuilder writeSymLinkEntry](https://github.com/openjdk/jdk/blob/aff9cea05468daf60fa80c7d9993b3aa8497b0c5/src/jdk.jlink/share/classes/jdk/tools/jlink/builder/DefaultImageBuilder.java#L469)

In the absence of being able to change this `jlink` behavior, maybe we can work around it.
Instead of baking knowledge of the names and locations of specific files into the Standard Doclet, one possibility would be to read the file, and if it exactly matches `Please see .*\R` then we could substitute the contents of the named file.

In other words, the logic becomes roughly

```
var s = Files.readString(path);
if (s _matches_ "Please see (.*)\\R") {. // follow jlink descriptive links
    path = Path.of(_specified-file_);
    s = Files.readString(path);
}
Files.writeString(_target_, s);
```

jonathan-gibbons (changes requested):
I still (strongly) dislike baking such specific knowledge of the legal files into the javadoc tool. That's just conceptually wrong.

It turns out these files are generated by the `jlink` tool, which somewhat weirdly generates these "descriptive links" to use instead of symbolic links on platforms that do not support symbolic links.

See [jlink DefaultImageBuilder writeSymLinkEntry](https://github.com/openjdk/jdk/blob/aff9cea05468daf60fa80c7d9993b3aa8497b0c5/src/jdk.jlink/share/classes/jdk/tools/jlink/builder/DefaultImageBuilder.java#L469)

In the absence of being able to change this `jlink` behavior, maybe we can work around it.
Instead of baking knowledge of the names and locations of specific files into the Standard Doclet, one possibility would be to read the file, and if it exactly matches `Please see .*\R` then we could substitute the contents of the named file.

In other words, the logic becomes roughly

```
var s = Files.readString(path);
if (s _matches_ "Please see (.*)\\R") {. // follow jlink descriptive links
    path = Path.of(_specified-file_);
    s = Files.readString(path);
}
Files.writeString(_target_, s);
```

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/HtmlDoclet.java
```diff
@@ -367,6 +371,18 @@ private void copyLegalFiles(boolean includeJQuery) throws DocletException {
...
+                    if (isDefault) {
+                        Pattern pattern = Pattern.compile("Please see (.*)\\R"); // follow jlink descriptive links
+                        Matcher matcher = pattern.matcher(Files.readString(entry));
+                        if (matcher.find()) {
+                            try {
+                                Path descriptive_link = entry.getParent().resolve(matcher.group(1));
+                                if (Files.isRegularFile(descriptive_link)) {
+                                    entry = descriptive_link;
+                                }
+                            } catch(InvalidPathException e) {}
```
turbanoff, line 383: nit
```suggestion
                            } catch (InvalidPathException e) {}
```

tkiriyama: Thank you for your suggestion. I fixed it.

tkiriyama:
@jonathan-gibbons 
I would like to request another review for my fix. I have incorporated your comments into the fix.

tkiriyama:
@jonathan-gibbons 
Do you think it will take much longer for your review?

tkiriyama:
@turbanoff 

Would you review and sponsor this fix, please?

tkiriyama:
@jonathan-gibbons @turbanoff 

I'm sorry to trouble you, but could you review this fix again?

turbanoff:
@tkiriyama sorry, I'm not an OpenJDK reviewer.

tkiriyama:
@turbanoff 
OK. Thank you for your reply.

@jonathan-gibbons 
Would you mind  sponsoring this pull request?

tkiriyama:
I can't find anyone to sponsor this pull request.
Could someone please review this fix?

jaikiran:
Hello @tkiriyama, I agree with Jon, it feels wrong to be doing such a change in the javadoc code. I understand why it's being done - on Windows, the per module license files (in this case for the `jdk.javadoc` module) do not contain the actual license text. That's due to how jlink generates them on systems (like Windows) which do not support symbolic links.

I think we should ask for input from people who are more familiar with jlink on what is expected of other parts of JDK code when such code wants to copy over the license files. Perhaps @AlanBateman or @mlchung would have some suggestions.

mlchung:
> On Windows, the per module legal files (in this case for the jdk.javadoc module) do not contain the actual legal text. That's due to how jlink generates them on systems (like Windows) which do not support symbolic links.

This is by design to avoid duplicating `COPYRIGHT` and `LICENSE` from java.base in each module. 

One idea to fix this is to extend jlink `--dedup-legal-notices` to accept an option that specify the module names that always wants to keep a copy of the legal files.

jaikiran:
Hello Mandy,

> One idea to fix this is to extend jlink --dedup-legal-notices to accept an option that specify the module names that always wants to keep a copy of the legal files.

That sounds reasonable. So `jdk.javadoc` through of build system could then be marked as requiring a copy of these legal notice files when the JDK image is being built. 
Do you think that the option should be applicable only on systems where symbolic link isn't applicable or should the `LegalNoticeFilePlugin` use that option to just create copies of the legal files for a chosen module, irrespective of which platform it is for?

mlchung:
It requests not to deduplicate and hence I would say it applies to all platforms.   The option could look like:
```
--dedup-legal-notices [excludes=<comma-separated-module-names>]
```

e.g. `--dedup-legal-notices excludes=jdk.javadoc`

jaikiran:
Thank you Mandy. If it's OK then I'll go ahead and file an enhancement for jlink tool and propose a PR with these suggested changes. This current issue can then be used to do the build side changes to mark `jdk.javadoc` module as not needing deduplication.

jaikiran:
I've now created https://bugs.openjdk.org/browse/JDK-8317634  and opened a PR for it https://github.com/openjdk/jdk/pull/16066.

mlchung:
#16066 does not fully solve the problem as it only works on JDK image that ensures javadoc's legal files are copied. 

These legal files are part of the javadoc output like other resource files such as `*.js`.     Like what Jon suggests, javadoc can/should copy these files from its resources directory like all other files such as `*.js` copied to the output directory and javadoc no longer  relies on the jlink output.

jaikiran:
https://github.com/openjdk/jdk/pull/16370 supersedes this PR. So I think this current PR can now be closed.

tkiriyama:
Thanks! I will close this pull request.

PR #13687, opened 2023-04-27
JDK-8306983: Do not invoke external programs when switch terminal to raw mode on selected platforms

lahodaj:
To support JShell and other usecases, the JDK uses JLine, which provides line-editing functionality inside a terminal.

JLine has several ways to work with the terminal, based on various additional libraries and tools. Most of them are not directly usable inside the JDK, so currently, on non-Windows platforms, the JLine inside the JDK will use external programs, like `stty`, to inspect the terminal's properties and to switch the terminal to raw mode.

This is slightly problematic, as the external programs might be missing, and while this is not that big problem for JShell, it might be a problem for other potential uses of JLine, like using it inside `System.console()`.

So, the proposal here is to call the corresponding native methods directly, on selected platforms for now (Linux and Mac OS/X), instead of invoking the external programs. On Windows, this has always been the case, we are using a custom implementation of the interface that maps native and Java function for JNA there, and the proposal is to do the same here. We take the appropriate mapping interface for JNA, and provide hand-written implementation for it, using JNI.

The Windows implementation is mostly unchanged, the changes are mostly non-Windows only. The overview of the changes is:
 - `LastErrorException` is moved from the Windows-specific code to the platform-neutral code, as it is used by all the native implementations. This is the only change that directly affects the Windows-specific code
 -  `unix/classes/jdk/internal/org/jline/terminal/impl/jna/JnaNativePty.java` and `unix/classes/jdk/internal/org/jline/terminal/impl/jna/JnaTerminalProvider.java` are created based on the corresponding files from JLine. In JLine, these files directly call platform-specific implementations, but in this patch, the platform specific code is commented out, and replaced with calls to `JDKNativePty`, which is a new platform-specific class, that delegates to the actual platform-specific implementations. Note the `JnaNativePty.java` and `JnaTerminalProvider.java` already exist in the Windows part, but have different pieces of code commented out, which makes them substantially different.
 - for Linux and Mac OS/X, there are platform-specific implementations based on the corresponding code from JLine, with the hand-written JNI-based implementation of the JNA-based existing interfaces. They also have an implementation of `JDKNativePty`, which just delegates to the given platform's `"NativePty"` implementation.
 - the `JdkConsoleProviderImpl.java` has been tweaked to not allow the implementation based on the external programs, and gracefully falling back to the standard implementation. JShell is kept unchanged.

The reasons for this organization are: limiting duplication, mostly adhering to the JDK's platform specific build (which is different from the normal JLine's platform-neutral build) and limiting the difference between standard JLine and JLine inside JDK, to help future upgrades.

---------
### Progress

### Issue
 * [JDK-8306983](https://bugs.openjdk.org/browse/JDK-8306983): Do not invoke external programs when switch terminal to raw mode on selected platforms

### Reviewers
 * [Erik Joelsson](https://openjdk.org/census#erikj) (@erikj79 - **Reviewer**) ⚠️ Review applies to [4cf8f67e](https://git.openjdk.org/jdk/pull/13687/files/4cf8f67e43f442a5c48cd30349740ac2cb638d6e)

### Contributors
 * Adam Sotona `<asotona@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13687/head:pull/13687` \
`$ git checkout pull/13687`

`$ git checkout pull/13687` \
`$ git pull https://git.openjdk.org/jdk.git pull/13687/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13687`

`$ git pr show -t 13687`

</details>

<a href="https://git.openjdk.org/jdk/pull/13687.diff">https://git.openjdk.org/jdk/pull/13687.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13687#issuecomment-1532724133)


make/modules/jdk.internal.le/Lib.gmk
```diff
@@ -27,14 +27,15 @@ include LibCommon.gmk
...
 
   $(eval $(call SetupJdkLibrary, BUILD_LIBLE, \
       NAME := le, \
       OPTIMIZATION := LOW, \
-      CFLAGS := $(CFLAGS_JDKLIB), \
+      CFLAGS := $(CXXFLAGS_JDKLIB), \
       LDFLAGS := $(LDFLAGS_JDKLIB), \
-      LIBS := $(JDKLIB_LIBS) user32.lib, \
+      LIBS_windows := $(JDKLIB_LIBS) user32.lib, \
+      LIBS_unix := $(JDKLIB_LIBS) -lstdc++, \
```
erikj79, line 38: If this is a C++ library, I think it's better to specify `TOOLCHAIN := TOOLCHAIN_LINK_CXX` than adding `-lstdc++`.

erikj79: We may also need `$(LIBCXX)` to force static linking when that is enabled for the build.

naotoj (commented):
Changes in `JdkConsoleProviderImpl.java` look good to me. Didn't look at other diffs closely, but it seems some other files are missing Oracle copyright text.

erikj79 (approved):
Build changes look good.

/reviewers 2

src/jdk.internal.le/linux/native/lible/CLibrary.cpp
```diff
@@ -0,0 +1,206 @@
+/*
+ * Copyright (c) 2015, 2018, Oracle and/or its affiliates. All rights reserved.
```
bplb, line 2: Copyright year 2023 only if this is new.

src/jdk.internal.le/linux/native/lible/CLibrary.cpp
```diff
@@ -0,0 +1,206 @@
...
+    ws_ypixel= env->GetFieldID(winsize_j, "ws_ypixel", "S");
+    CHECK_NULL(ws_ypixel);
+}
+
+JNIEXPORT void JNICALL Java_jdk_internal_org_jline_terminal_impl_jna_linux_CLibraryImpl_tcgetattr
+  (JNIEnv *env, jobject, jint fd, jobject result) {
+    termios data;
+
+    if (tcgetattr(fd, &data) != 0) {
+        jobject exc = env->NewObject(lastErrorExceptionClass,
```
bplb, line 103: Could lines 103-106 be converted to a utility function which would then be invoked at current lines 103, 141, 163, and 197?

src/jdk.internal.le/linux/native/lible/CLibrary.cpp
```diff
@@ -0,0 +1,206 @@
...
+        return ;
+    }
+
+    env->SetIntField(result, c_iflag, data.c_iflag);
+    env->SetIntField(result, c_oflag, data.c_oflag);
+    env->SetIntField(result, c_cflag, data.c_cflag);
+    env->SetIntField(result, c_lflag, data.c_lflag);
+    env->SetIntField(result, c_line, data.c_line);
+    jbyteArray c_ccValue = (jbyteArray) env->GetObjectField(result, c_cc);
+    env->SetByteArrayRegion(c_ccValue, 0, NCCS, (signed char *) data.c_cc);//TODO: cast?
```
bplb, line 116: Cast instead to `(jbyte*)` here and at lines 136 and 204?

src/jdk.internal.le/linux/native/lible/CLibrary.cpp
```diff
@@ -0,0 +1,206 @@
...
+}
+
+/*
+ * Class:     jdk_internal_org_jline_terminal_impl_jna_linux_CLibraryImpl
+ * Method:    isatty
+ * Signature: (I)I
+ */
+JNIEXPORT jint JNICALL Java_jdk_internal_org_jline_terminal_impl_jna_linux_CLibraryImpl_isatty
+  (JNIEnv *, jobject, jint fd) {
+    return isatty(fd);
```
bplb, line 183: Do we care if the native `isatty()` returns zero? Or is this dealt with somewhere else?

src/jdk.internal.le/macosx/native/lible/CLibrary.cpp
```diff
@@ -0,0 +1,210 @@
...
+    nativelong_value = env->GetFieldID(nativelong_j, "value", "J");
+    CHECK_NULL(nativelong_value);
+}
+
+JNIEXPORT void JNICALL Java_jdk_internal_org_jline_terminal_impl_jna_osx_CLibraryImpl_tcgetattr
+  (JNIEnv *env, jobject, jint fd, jobject result) {
+    termios data;
+
+    if (tcgetattr(fd, &data) != 0) {
+        jobject exc = env->NewObject(lastErrorExceptionClass,
```
bplb, line 109: Could lines 109-112 be converted to a utility function which would then be invoked at current lines 109, 145, 167, and 197?

src/jdk.internal.le/macosx/native/lible/CLibrary.cpp
```diff
@@ -0,0 +1,210 @@
...
+        env->Throw((jthrowable) exc);
+        return ;
+    }
+
+    env->SetLongField(env->GetObjectField(result, c_iflag), nativelong_value, data.c_iflag);
+    env->SetLongField(env->GetObjectField(result, c_oflag), nativelong_value, data.c_oflag);
+    env->SetLongField(env->GetObjectField(result, c_cflag), nativelong_value, data.c_cflag);
+    env->SetLongField(env->GetObjectField(result, c_lflag), nativelong_value, data.c_lflag);
+    jbyteArray c_ccValue = (jbyteArray) env->GetObjectField(result, c_cc);
+    env->SetByteArrayRegion(c_ccValue, 0, NCCS, (signed char *) data.c_cc);
```
bplb, line 121: Cast instead to `(jbyte*)` here and at lines 140 and 208?

src/jdk.internal.le/macosx/native/lible/CLibrary.cpp
```diff
@@ -0,0 +1,210 @@
...
+}
+
+/*
+ * Class:     jdk_internal_org_jline_terminal_impl_jna_osx_CLibraryImpl
+ * Method:    isatty
+ * Signature: (I)I
+ */
+JNIEXPORT jint JNICALL Java_jdk_internal_org_jline_terminal_impl_jna_osx_CLibraryImpl_isatty
+  (JNIEnv *, jobject, jint fd) {
+    return isatty(fd);
```
bplb, line 187: Do we care if the native `isatty()` returns zero? Or is this dealt with somewhere else?

src/jdk.internal.le/macosx/native/lible/CLibrary.cpp
```diff
@@ -0,0 +1,210 @@
+/*
+ * Copyright (c) 2015, 2018, Oracle and/or its affiliates. All rights reserved.
```
bplb, line 2: Copyright year 2023 only?

src/jdk.internal.le/linux/native/lible/CLibrary.cpp
lahodaj: This method is only used to determine with the fd is attached to a terminal (returns 1) or not (return 0). The reasons why it is not attached to a terminal are not really important. The value is used here:
https://github.com/lahodaj/jdk/blob/4cf8f67e43f442a5c48cd30349740ac2cb638d6e/src/jdk.internal.le/unix/classes/jdk/internal/org/jline/terminal/impl/jna/JnaNativePty.java#L186

src/jdk.internal.le/macosx/native/lible/CLibrary.cpp
lahodaj: This method is only used to determine with the fd is attached to a terminal (returns 1) or not (return 0). The reasons why it is not attached to a terminal are not really important. The value is used here:
https://github.com/lahodaj/jdk/blob/4cf8f67e43f442a5c48cd30349740ac2cb638d6e/src/jdk.internal.le/unix/classes/jdk/internal/org/jline/terminal/impl/jna/JnaNativePty.java#L186

vicente-romero-oracle (approved):

bplb (approved):
C++ looks fine.

PR #13689, opened 2023-04-27
JDK-8306112 Implementation of JEP 445: Flexible Main Methods and Anonymous Main Classes (Preview)

JimLaskey:
Add flexible main methods and anonymous main classes to the Java language.


src/java.base/share/classes/sun/launcher/resources/launcher.properties
```diff
@@ -236,6 +236,8 @@ java.launcher.cls.error6=\
 java.launcher.cls.error7=\
      Error: Unable to initialize main class {0}\n\
      Caused by: {1}: {2}
+java.launcher.cls.error8=\
+    Error: no non-private no argument constructor found in class {0}
```
AlanBateman, line 240: Since this feature is for new developers then the error messages will need to be understand, maybe it should be "zero argument" and give an example to help.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Enter.java
```diff
@@ -419,6 +428,54 @@ public void visitPackageDef(JCPackageDecl tree) {
             }
         };
 
+    // Restructure top level to be an top level anonymous class.
+    public static void constructAnonymousMainClass(JCCompilationUnit tree,
```
mcimadamore, line 432: Question: any reason as to why this is done here and not in the parser? Typically we don't want to do tree transformation at parse time, as that messes up clients that want to access the "non-desugared" tree (such as IDE) and expect some mapping between source to AST. But if you do the rewriting in Enter, not much changes, that is, clients such as IDEs would still see something that doesn't resemble the source.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java
```diff
@@ -2432,7 +2430,9 @@ Symbol findIdentInternal(Env<AttrContext> env, Name name, KindSelector kind) {
 
         if (kind.contains(KindSelector.TYP)) {
             sym = findType(env, name);
-
+            if ((sym.flags() & ANONYMOUS_MAIN_CLASS) != 0) {
```
mcimadamore, line 2433: If `sym` is also `SYNTHTIC` (which it is) - do you need this? In what case? E.g. if there's a legitimate case for this I'm wondering if we should just test for `SYNTHETIC` rather than `ANONYMOUS_MAIN_CLASS`.

jddarcy:
I assume future iterations of the PR will include handling and tests for the main class attribute of jar files.

src/java.base/share/native/libjli/java.c
```diff
@@ -539,12 +541,55 @@ JavaMain(void* _args)
...
+    case MAIN_NONSTATIC | MAIN_WITHOUT_ARGS: {
+        constructor = (*env)->GetMethodID(env, mainClass, "<init>", "()V");
+        CHECK_EXCEPTION_NULL_LEAVE(constructor);
+        mainObject = (*env)->NewObject(env, mainClass, constructor);
+        CHECK_EXCEPTION_NULL_LEAVE(mainObject);
+        mainID = (*env)->GetMethodID(env, mainClass, "main",
+                                     "()V");
+        CHECK_EXCEPTION_NULL_LEAVE(mainID);
+        (*env)->CallVoidMethod(env, mainObject, mainID);
+        break;
```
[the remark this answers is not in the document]
JimLaskey: Good idea. If @hidden doesn't work then we can eat the trace entries.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -38,6 +38,11 @@
 import java.io.IOException;
 import java.io.PrintStream;
 import java.io.UnsupportedEncodingException;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandleInfo;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodHandles.Lookup;
+import java.lang.invoke.MethodType;
```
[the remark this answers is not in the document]
JimLaskey: Yep.

src/java.base/share/classes/sun/launcher/resources/launcher.properties
JimLaskey:
```
java.launcher.cls.error8=\
    Error: no non-private zero argument constructor found in class {0}\n\
    remove private from existing constructor or define as:\n\
\   public {0}() {}
```

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java
JimLaskey: Yes, this is a remnant from earlier work.

src/jdk.compiler/share/classes/com/sun/tools/javac/parser/VirtualParser.java
```diff
@@ -0,0 +1,209 @@
...
+     *
+     * @param parser           parent parser
+     * @param parserAction     function that takes a parser and invokes a method on that parser
+     * @param commitOnSuccess  if true updates the active parser scanner to reflect the effects
+     *                         of the action
+     * @return {@link Optional} result of the action
+     *
+     * @param <E> return type of parserAction
+     */
+    public static <E> Optional<E> speculativeParse(JavacParser parser,
```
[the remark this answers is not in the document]
JimLaskey: dropping

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Enter.java
JimLaskey: The source file name is not available until after parsing.

mcimadamore: I see that logic in JavaCompiler - I wonder if that's just the way it is, or if there's a deeper reason as to why the sourcefile is set on the toplevel unit *after* parsing (I don't think I can see any, in which case that might be changed if that makes the rest of the code simpler). @lahodaj what do you think?

test/jdk/tools/launcher/OnrampMainTest.java
```diff
@@ -0,0 +1,135 @@
...
+ */
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * @test
+ * @summary test execution priority of main methods
+ * @run main OnrampMainTest
+ */
+public class OnrampMainTest extends TestHelper {
```
mcimadamore, line 31: Should more tests be added for inherited methods?

dholmes-ora:
I don't see anything hotspot related here (no idea why the auto-mapping added it) so can we remove hotspot-runtime please.

src/jdk.compiler/share/classes/com/sun/tools/javac/launcher/Main.java
```diff
@@ -423,21 +434,35 @@ private void execute(String mainClassName, String[] appArgs, Context context)
         ClassLoader cl = context.getClassLoader(ClassLoader.getSystemClassLoader());
         try {
             Class<?> appClass = Class.forName(mainClassName, true, cl);
-            Method main = appClass.getDeclaredMethod("main", String[].class);
-            int PUBLIC_STATIC = Modifier.PUBLIC | Modifier.STATIC;
-            if ((main.getModifiers() & PUBLIC_STATIC) != PUBLIC_STATIC) {
+            Method main = MainMethodFinder.findMainMethod(appClass);
+            if (!PreviewFeatures.isEnabled() && (!isStatic(main) || !isPublic(main))) {
```
lahodaj, line 438: It seems one can run a main method without parameters without `--enable-preview` using this codepath. That is presumably not intended.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Enter.java
lahodaj: I believe we were discussing this some time ago, and there were some problems. I don't recall the exact details, but I'll try to look into this later.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Enter.java
```diff
@@ -419,6 +428,54 @@ public void visitPackageDef(JCPackageDecl tree) {
...
+            log.error(null, Errors.BadFileName(simplename));
+        }
+        Name name = names.fromString(simplename);
+
+        ListBuffer<JCTree> topDefs = new ListBuffer<>();
+        ListBuffer<JCTree> defs = new ListBuffer<>();
+
+        for (JCTree def : tree.defs) {
+            if (def.hasTag(Tag.PACKAGEDEF)) {
+                log.error(null, Errors.AnonymousMainClassShouldNotHavePackageDeclaration);
```
lahodaj, line 462:
```suggestion
                log.error(def.pos(), Errors.AnonymousMainClassShouldNotHavePackageDeclaration);
```

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Enter.java
```diff
@@ -419,6 +428,54 @@ public void visitPackageDef(JCPackageDecl tree) {
...
+            preview.warnPreview(tree.pos, feature);
+        }
+
+        make.at(tree.pos);
+        String simplename = PathFileObject.getSimpleName(tree.sourcefile);
+        if (simplename.endsWith(".java")) {
+            simplename = simplename.substring(0, simplename.length() - ".java".length());
+        }
+        if (!SourceVersion.isIdentifier(simplename) || SourceVersion.isKeyword(simplename)) {
+            log.error(null, Errors.BadFileName(simplename));
```
lahodaj, line 453:
```suggestion
            log.error(tree.pos(), Errors.BadFileName(simplename));
```

src/jdk.compiler/share/classes/com/sun/tools/javac/tree/JCTree.java
```diff
@@ -623,6 +623,30 @@ public <R,D> R accept(TreeVisitor<R,D> v, D d) {
         public Tag getTag() {
             return TOPLEVEL;
         }
+
+        public boolean isAnonymousMainClass() {
```
lahodaj, line 627: This method seems a bit confusing to me. I believe the fields and methods will be stripped from `defs` if/when the wrapping class is created, which will mean this method will start to return `false`, no? It overall does not seem like a generally useful predicate.

(If I understand this correctly, if we created the wrapping class in parser neither of these two methods would be needed.)

src/jdk.compiler/share/classes/com/sun/tools/javac/tree/JCTree.java
```diff
@@ -623,6 +623,30 @@ public <R,D> R accept(TreeVisitor<R,D> v, D d) {
...
+                    return true;
+                }
+            }
+            return false;
+        }
+
+        public JCClassDecl getAnonymousMainClass() {
+            // Find anonymous main class.
+            for (JCTree def : defs) {
+                if (def.hasTag(CLASSDEF)) {
```
lahodaj, line 640: Nit, in cases like this, we lately tend to write `def instanceof JCClassDecl cls`, although we understand this way to check the AST is only safe inside javac.

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
```diff
@@ -0,0 +1,166 @@
...
+     */
+    public static Method findMainMethod(Class<?> mainClass) throws NoSuchMethodException {
+        try {
+            Method mainMethod = mainClass.getMethod("main", String[].class);
+
+            if (mainMethod.getDeclaringClass() != mainClass) {
+                System.err.println("WARNING: static main in super class will be deprecated.");
+            }
+
+            return mainClass.getMethod("main", String[].class);
```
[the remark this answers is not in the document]
JimLaskey: Yes

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
```diff
@@ -0,0 +1,166 @@
...
+     * @param declc  the top level declaring class
+     * @param refc   the declaring class or super class
+     * @param mains  accumulated main methods
+     */
+    private static void gatherMains(Class<?> declc, Class<?> refc, List<Method> mains) {
+        if (refc != null && refc != Object.class) {
+            gatherMains(declc, refc.getSuperclass(), mains);
+
+            for (Method method : refc.getDeclaredMethods()) {
+                int argc = method.getParameterCount();
```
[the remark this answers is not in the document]
JimLaskey: Remnant.

src/jdk.compiler/share/classes/com/sun/tools/javac/launcher/Main.java
JimLaskey: MainMethodFinder won't return a no arg method is without --enable-preview. The local testing here is to specialize the error massages.

src/jdk.compiler/share/classes/com/sun/tools/javac/parser/JavacParser.java
```diff
@@ -3898,11 +3918,34 @@ public JCTree.JCCompilationUnit parseCompilationUnit() {
...
+
+                // Do to a significant number of existing negative tests
+                // this code speculatively tests to see if a top level method
+                // or field can parse. If the method or field can parse then
+                // it is parsed. Otherwise, parsing continues as though
+                // anonymous main classes did not exist and error reporting
+                // is the same as in the past.
+                if (Feature.ANONYMOUS_MAIN_CLASSES.allowedInSource(source) &&
+                        !isDeclaration() &&
+                        (token.kind == VOID || token.kind == IDENTIFIER)) {
```
lahodaj, line 3933: These checks (`token.kind == VOID || token.kind == IDENTIFIER`) for token kind will fail to parse fields of primitive types (or methods with primitive types as a return type)? Either the tests must include the primitive types, or (maybe) just this test can be avoided completely?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Enter.java
JimLaskey: Changing

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Enter.java
JimLaskey: Changing

src/jdk.compiler/share/classes/com/sun/tools/javac/tree/JCTree.java
JimLaskey: Changing

src/jdk.compiler/share/classes/com/sun/tools/javac/tree/JCTree.java
JimLaskey: Inlining

src/jdk.compiler/share/classes/com/sun/tools/javac/parser/JavacParser.java
JimLaskey: Changing

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Enter.java
lahodaj: I've sketched this:
https://github.com/lahodaj/jdk/commit/efe55f7d354ed7bbf91077d058823d682db501b9

I don't have too strong opinion here, there might be more cleanup possible after a change like this, and might be a bit cleaner outside of the parser; but forces the parser to work with files which is somewhat less clean.

JimLaskey: This is basically what I did originally until I decided that I needed a distinct class name. `log.currentSourceFile()` is the magic in this code. I'm willing to move this way if we all agree.

src/java.base/share/native/libjli/java.c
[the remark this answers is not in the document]
JimLaskey: Many dependencies (ex. JDI) on main being the first frame. Will set up a separate RFE witch should include using the LauncherHelper in the source code launcher.

mcimadamore (approved):
The compiler changes look good - I like howthe code got simpler by adding the main class wrapper at parser time. As I suggested elsewhere, do we want to change `Pretty` so that the AST of an unnamed class is rendered like if the unnamed class wasn't there (e.g. so that the output is like the source where the AST comes from) ?

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
```diff
@@ -0,0 +1,164 @@
...
+        return 0;
+    }
+
+    /**
+     * {@return priority main method or null if none found}
+     * @param mainClass main class
+     */
+    public static Method findMainMethod(Class<?> mainClass) throws NoSuchMethodException {
+        try {
+            Method mainMethod = mainClass.getMethod("main", String[].class);
```
jaikiran, line 139: Hello Jim, I think this specific line is trying to find a `public static void main(String[])` method from the launched class. In the current form of this implementation, this has the potential of returning a non-static `public void main(String[])` from here. I think a `isStatic(mainMethod)` would be needed here before returning this method as the found method.

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
```diff
@@ -0,0 +1,164 @@
...
+
+            if (mainMethod.getDeclaringClass() != mainClass) {
+                System.err.println("WARNING: static main in super class will be deprecated.");
```
jaikiran, line 142: Similarly, this warning would have to be logged only if the method is `static`. Furthermore, do you think we should include the declaring class in the log message to provide some context on what's causing this warning? Something like:
> WARNING: static main(String[]) in super class foo.bar.Parent will be deprecated.

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
```diff
@@ -0,0 +1,164 @@
...
+            }
+
+            return mainMethod;
+        } catch (NoSuchMethodException nsme) {
+            if (!PreviewFeatures.isEnabled()) {
+                throw nsme;
+            }
+
+            List<Method> mains = new ArrayList<>();
+            gatherMains(mainClass, mainClass, mains);
```
jaikiran, line 152: The `try` block above is there to find `public static void main(String[])` in the launched class. When it's not found, then we gather the potential main methods (as listed in the JEP). The implementation of `gatherMains(...)`, in its current form starts gathering the main methods from `refc.getSuperclass()`, where `refc` in this case is the `mainClass`, which is the launched class.

So if I'm reading this correctly, then I think it's going to completely skip the launched class for looking any potential main methods and instead starting looking for them in the launched's class super hierarchy.

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
```diff
@@ -0,0 +1,164 @@
...
+
+            if (mains.isEmpty()) {
+                throw new NoSuchMethodException("No main method found");
+            }
+
+            mains.sort(MainMethodFinder::compareMethods);
```
jaikiran, line 158: Perhaps skip the call to `sort` and just return the found main method if `mains.size() == 1`?

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
```diff
@@ -0,0 +1,164 @@
...
+            return -1;
+        } else if (bClass.isAssignableFrom(aClass)) {
+            return 1;
+        }
+
+        return 0;
+    }
+
+    /**
+     * {@return priority main method or null if none found}
```
jaikiran, line 134: I think this javadoc is perhaps outdated? The current implementation of this method throws a `NoSuchMethodException` when it can't find any eligible main method.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -840,36 +844,79 @@ private static boolean doesExtendFXApplication(Class<?> mainClass) {
...
+        mainType = (isStatic ? 0 : MAIN_NONSTATIC) | (noArgs ? MAIN_WITHOUT_ARGS : 0);
+    }
+
     // Check the existence and signature of main and abort if incorrect
-    static void validateMainClass(Class<?> mainClass) {
+    public static void validateMainClass(Class<?> mainClass) {
         Method mainMethod = null;
         try {
-            mainMethod = mainClass.getMethod("main", String[].class);
+            mainMethod = MainMethodFinder.findMainMethod(mainClass);
```
jaikiran, line 872: The `MainMethodFinder.findMainMethod(...)` throws a `NoSuchMethodException` when it can't find the regular `public static void main(String[])` or, when preview is enabled, any other eligible main methods. That will now mean that the next line here which catches a `NoSuchMethodException` can potentially end up aborting with a (very confusing) error message about JavaFX application. We should perhaps change that catch block to something like:
```
} catch (NoSuchMethodException nsme) {
   if (!PreviewFeatures.isEnabled()) {
                  // invalid main or not FX application, abort with an error
            abort(null, "java.launcher.cls.error4", mainClass.getName(),
                  JAVAFX_APPLICATION_CLASS_NAME);
   } else {
       // abort with something more clear error?
   }
```

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -840,36 +844,79 @@ private static boolean doesExtendFXApplication(Class<?> mainClass) {
...
         }
 
+        setMainType(mainMethod);
+
         /*
-         * getMethod (above) will choose the correct method, based
+         * findMainMethod (above) will choose the correct method, based
          * on its name and parameter type, however, we still have to
-         * ensure that the method is static and returns a void.
+         * ensure that the method is static (non-preview) and returns a void.
```
jaikiran, line 893: I think it's probably going to be easier to read and maintain if we moved these checks for `static` and `void` return type into the `MainMethodFinder.findMainMethod(...)` itself.

src/jdk.compiler/share/classes/com/sun/tools/javac/launcher/Main.java
```diff
@@ -407,37 +410,70 @@ private String compile(Path file, List<String> javacOpts, Context context) throw
...
         ClassLoader cl = context.getClassLoader(ClassLoader.getSystemClassLoader());
         try {
             Class<?> appClass = Class.forName(mainClassName, true, cl);
-            Method main = appClass.getDeclaredMethod("main", String[].class);
-            int PUBLIC_STATIC = Modifier.PUBLIC | Modifier.STATIC;
-            if ((main.getModifiers() & PUBLIC_STATIC) != PUBLIC_STATIC) {
+            Method mainMethod = MainMethodFinder.findMainMethod(appClass);
+            int mods = mainMethod.getModifiers();
+            boolean isStatic = Modifier.isStatic(mods);
+            boolean isPublic = Modifier.isStatic(mods);
```
jaikiran, line 440: This looks like a typo. This should have been `Modifier.isPublic(mods);`

src/jdk.compiler/share/classes/com/sun/tools/javac/launcher/Main.java
```diff
@@ -407,37 +410,70 @@ private String compile(Path file, List<String> javacOpts, Context context) throw
         return mainClassName;
     }
 
+    private static boolean isPublic(Method method) {
+        return method != null && Modifier.isPublic(method.getModifiers());
+    }
+
+    private static boolean isStatic(Method method) {
+        return method != null && Modifier.isStatic(method.getModifiers());
+    }
```
jaikiran, line 419: It looks like these new methods aren't being used.

src/jdk.compiler/share/classes/com/sun/tools/javac/launcher/Main.java
```diff
@@ -407,37 +410,70 @@ private String compile(Path file, List<String> javacOpts, Context context) throw
...
-            main.setAccessible(true);
-            main.invoke(0, (Object) appArgs);
+
+            // Similar to sun.launcher.LauncherHelper#executeMainClass
+            // but duplicated here to prevent additional launcher frames
+            mainMethod.setAccessible(true);
+
+            if (isStatic) {
+                if (noArgs) {
+                    mainMethod.invoke(appClass);
```
jaikiran, line 457: Given that `Method.invoke(...)` on a static method ignores the `obj` param that's passed to it, perhaps pass `null`?

test/jdk/tools/launcher/InstanceMainTest.java
```diff
@@ -0,0 +1,151 @@
...
+import java.nio.file.Path;
+
+/**
+ * @test
+ * @summary test execution priority of main methods
+ * @run main InstanceMainTest
+ */
+public class InstanceMainTest extends TestHelper {
+
+    @Test
```
jaikiran, line 33: Does this compile? I don't see imports for this annotation and this is launched as a `@run main ....`

test/jdk/tools/launcher/OnrampMainTest.java
JimLaskey: Test added

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
JimLaskey: Yes this is an oversight on my part.

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
```diff
@@ -0,0 +1,164 @@
...
+            gatherMains(declc, refc.getSuperclass(), mains);
+
+            for (Method method : refc.getDeclaredMethods()) {
+                // Must be named "main", public|protected|package-private and either
+                // no arguments or one string array argument.
+                if ("main".equals(method.getName()) &&
+                        !isPrivate(method) &&
+                        correctArgs(method) &&
+                        // Only statics in the declaring class
+                        (!isStatic(method) || declc == refc)
```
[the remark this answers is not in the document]
JimLaskey: abstract would be overshadowed by its implementation and native is a fair game.

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
JimLaskey: Changing

src/jdk.compiler/share/classes/com/sun/tools/javac/launcher/Main.java
JimLaskey: Changing

src/jdk.compiler/share/classes/com/sun/tools/javac/launcher/Main.java
JimLaskey: Remnants - removing

src/jdk.compiler/share/classes/com/sun/tools/javac/launcher/Main.java
JimLaskey: Nostalgia and a useful reminder.

test/jdk/tools/launcher/InstanceMainTest.java
JimLaskey: Package private

src/jdk.compiler/share/classes/com/sun/tools/javac/parser/JavacParser.java
```diff
@@ -3921,6 +3967,43 @@ public JCTree.JCCompilationUnit parseCompilationUnit() {
...
+        if (simplename.endsWith(".java")) {
+            simplename = simplename.substring(0, simplename.length() - ".java".length());
+        }
+        if (!SourceVersion.isIdentifier(simplename) || SourceVersion.isKeyword(simplename)) {
+            log.error(primaryPos, Errors.BadFileName(simplename));
+        }
+
+        Name name = names.fromString(simplename);
+        JCModifiers anonMods = F.at(primaryPos)
+                .Modifiers(Flags.FINAL|Flags.MANDATED|Flags.SYNTHETIC|Flags.UNNAMED_CLASS, List.nil());
```
vicente-romero-oracle, line 3999: I wonder if testing if a class has flags: `Flags.FINAL|Flags.MANDATED|Flags.SYNTHETIC` should be enough to know if it is unnamed or not and we don't need to use the new `UNNAMED_CLASS` flag

forax: `Flags.MANDATED` on a class is currently enough to know if it's a unnamed class or not, but it's not enough for classes not produced by javac.

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
```diff
@@ -0,0 +1,171 @@
...
+
+        return 0;
+    }
+
+    /**
+     * {@return priority main method or null if none found}
+     *
+     * @param mainClass main class
+     *
+     * @throws NoSuchMethodException when not and preview and no method found
```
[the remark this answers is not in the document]
JimLaskey: oops - never on a Friday

src/jdk.compiler/share/classes/com/sun/tools/javac/parser/JavacParser.java
vicente-romero-oracle: good point

forax: Just to be clear, here, Flags.UNNAMED_CLASS is needed because an an unnamed class must have a main(), which is tested later.

src/jdk.compiler/share/classes/com/sun/tools/javac/launcher/Main.java
```diff
@@ -412,32 +415,57 @@ private String compile(Path file, List<String> javacOpts, Context context) throw
...
+
+                if (noArgs) {
+                    mainMethod.invoke(instance);
+                } else {
+                    mainMethod.invoke(instance, (Object)mainArgs);
+                }
+            }
         } catch (ClassNotFoundException e) {
             throw new Fault(Errors.CantFindClass(mainClassName));
         } catch (NoSuchMethodException e) {
```
vicente-romero-oracle, line 466: `getDeclaredConstructor` can also throw a NSME but this error message would still say that the main method couldn't be found unless at this point we are sure that the class must have a no-args constructor

JimLaskey: I was avoiding moving things so not to change behaviour when not in preview, but I think I really need to break it down into separate try blocks; get method, get constructor, execute method. Let me attempt.

vicente-romero-oracle (approved):
looks good

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
```diff
@@ -0,0 +1,171 @@
...
+            }
+
+            List<Method> mains = new ArrayList<>();
+            gatherMains(mainClass, mainClass, mains);
+
+            if (mains.isEmpty()) {
+                throw new NoSuchMethodException("No main method found");
+            }
+
+            if (1 < mains.size()) {
```
AlanBateman, line 162: Checking if mains.size() > 1 might be easier on the eyes.

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
```diff
@@ -0,0 +1,171 @@
...
+            }
+
+            return mainMethod;
+        } catch (NoSuchMethodException nsme) {
+            if (!PreviewFeatures.isEnabled()) {
+                throw nsme;
+            }
+
+            List<Method> mains = new ArrayList<>();
+            gatherMains(mainClass, mainClass, mains);
```
AlanBateman, line 156: Instead of gatherMains, did you consider first looking for static main(String[], then static main()? Asking because I expected to only see the walk up the hierarchy when looking for an instance main.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -843,36 +847,79 @@ private static boolean doesExtendFXApplication(Class<?> mainClass) {
...
-        if (!Modifier.isStatic(mod)) {
-            abort(null, "java.launcher.cls.error2", "static",
-                  mainMethod.getDeclaringClass().getName());
+        int mods = mainMethod.getModifiers();
+        boolean isStatic = Modifier.isStatic(mods);
+        boolean isPublic = Modifier.isPublic(mods);
+        boolean noArgs = mainMethod.getParameterCount() == 0;
+
+        if (!PreviewFeatures.isEnabled()) {
+            if (!isStatic || !isPublic || noArgs) {
```
AlanBateman, line 904: You can use && here and avoid the nested if.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -843,36 +847,79 @@ private static boolean doesExtendFXApplication(Class<?> mainClass) {
...
+    private static void setMainType(Method mainMethod) {
+        int mods = mainMethod.getModifiers();
+        boolean isStatic = Modifier.isStatic(mods);
+        boolean noArgs = mainMethod.getParameterCount() == 0;
+        mainType = (isStatic ? 0 : MAIN_NONSTATIC) | (noArgs ? MAIN_WITHOUT_ARGS : 0);
+    }
+
     // Check the existence and signature of main and abort if incorrect
-    static void validateMainClass(Class<?> mainClass) {
+    public static void validateMainClass(Class<?> mainClass) {
```
AlanBateman, line 872: Is there a reason that this is changed to public, maybe left over from a previous iteration?

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -843,36 +847,79 @@ private static boolean doesExtendFXApplication(Class<?> mainClass) {
...
         }
 
+        setMainType(mainMethod);
+
         /*
-         * getMethod (above) will choose the correct method, based
+         * findMainMethod (above) will choose the correct method, based
          * on its name and parameter type, however, we still have to
-         * ensure that the method is static and returns a void.
+         * ensure that the method is static (non-preview) and returns a void.
```
AlanBateman, line 896: Have you looked into findMainMethod checking the return type? Right now, we have findMainMethod returning a Method that needs further checking.

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
```diff
@@ -0,0 +1,171 @@
...
+import java.lang.reflect.Modifier;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MainMethodFinder {
+    private static boolean isPrivate(Method method) {
+        return method != null && Modifier.isPrivate(method.getModifiers());
+    }
+
+    private static boolean isPublic(Method method) {
```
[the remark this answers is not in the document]
JimLaskey: Changed

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
```diff
@@ -0,0 +1,171 @@
...
+package jdk.internal.misc;
+
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MainMethodFinder {
+    private static boolean isPrivate(Method method) {
+        return method != null && Modifier.isPrivate(method.getModifiers());
```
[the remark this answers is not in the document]
JimLaskey: Remnant of previous usage. Changed.

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    private static boolean correctArgs(Method method) {
+        int argc = method.getParameterCount();
+
+        return argc == 0 || argc == 1 && method.getParameterTypes()[0] == String[].class;
+    }
+
+    /**
+     * Gather all the "main" methods in the class heirarchy.
```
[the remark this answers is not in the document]
JimLaskey: Changed

test/jdk/tools/launcher/InstanceMainTest.java
```diff
@@ -0,0 +1,151 @@
...
+ */
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * @test
+ * @summary test execution priority of main methods
+ * @run main InstanceMainTest
+ */
+public class InstanceMainTest extends TestHelper {
```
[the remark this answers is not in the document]
JimLaskey: InstanceMainTest does check precedence. Are you expecting a test for all combinations?

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
```diff
@@ -0,0 +1,171 @@
...
+     *
+     * @throws NoSuchMethodException when not preview and no method found
+     */
+    public static Method findMainMethod(Class<?> mainClass) throws NoSuchMethodException {
+        try {
+            Method mainMethod = mainClass.getMethod("main", String[].class);
+            int mods = mainMethod.getModifiers();
+
+            if (Modifier.isStatic(mods) && mainMethod.getDeclaringClass() != mainClass) {
+                System.err.println("WARNING: static main in super class will be deprecated.");
```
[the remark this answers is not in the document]
JimLaskey: JEP 445 is not deprecating this. No advanced notice has been given.

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
JimLaskey: I never ever use > symbol. least to greatest left to right (maths).

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
JimLaskey: 99.99% of the time it will be a single method in a shallow hierarchy, so cost its low. Only reason I broke out public static main was to ensure performance for existing code was the same.

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
```diff
@@ -0,0 +1,171 @@
...
+            return -1;
+        } else if (bClass.isAssignableFrom(aClass)) {
+            return 1;
+        }
+
+        return 0;
+    }
+
+    /**
+     * {@return priority main method or null if none found}
```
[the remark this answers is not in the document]
JimLaskey: Changed

src/java.base/share/classes/sun/launcher/LauncherHelper.java
JimLaskey: Easier to see when removing preview code.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
JimLaskey: Remnant. Changed.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
JimLaskey: At some point I would like to fold both launchers into a single launcher. To do so would require having the source code launcher use the same error reporting and adapting downstream accordingly (tests et al). There is also a separate project to support multi-file source code launcher and I didn't want to disrupt the flow of that project at this point in time.

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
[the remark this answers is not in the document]
AlanBateman: Ah, I thought it was being deprecated. The JEP currently has "We deprecate the existing behavior that the launcher searches for a public static void main(String[] args) method in a superclass if one is not found in the launched class, issuing a runtime warning if such a method is found".

test/jdk/tools/launcher/InstanceMainTest.java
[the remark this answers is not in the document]
AlanBateman: > InstanceMainTest does check precedence. Are you expecting a test for all combinations?

Sorry, we discussed this in previous comments, ignore my latest comment on this.

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    /**
+     * {@return priority main method if none found}
+     *
+     * @param mainClass main class
+     *
+     * @throws NoSuchMethodException when not preview and no method found
+     */
+    public static Method findMainMethod(Class<?> mainClass) throws NoSuchMethodException {
```
AlanBateman, line 129: The latest iteration of changes removes the warning but the latest iteration of the JEP proposes a warning when an instance main or static no-args main is preferred over an inherited static main. Is it that the JEP is just running ahead a bit and the changes in the PR will catch up soon?

JimLaskey: Yes. There is some juggling going on to get the code to align with the requirement and the interpretation thereof.

kishida:
Can we use enum?

```java
public class Main{
  enum Foo {A, B}
  void main() {
    System.out.println(Foo.A);
  }
}
```
↑ OK

```java
enum Foo {A, B}
void main() {
  System.out.println(Foo.A);
}
```
↑error(enumtest.java is the filename)
```
enumtest.java:1: Error : package enumtest does not exist
  enum Foo {A, B}
  ^
```

JimLaskey:
Good catch - will look into it.

test/jdk/tools/launcher/InstanceMainTest.java
```diff
@@ -0,0 +1,188 @@
...
```
jddarcy, line 31: By my reading of the spec, "main" methods can be defined in record classes and enum classes as well as normal classes.
Are there tests for record and enum main method operation?

forax: You can also have a `main` method inside an interface. For enum classes and interfaces, the main has to be static.

JimLaskey: updating tests

jddarcy:
In case I overlooked them, it would be good to have some core reflection tests of what a unnamed class looks like reflectively, basically a core reflection analogue to the draft javac/javax.lang.model test:

https://github.com/openjdk/jdk/pull/14140/files#diff-106fc5b2c4d391546c7fa8a75e68fbd963733964f8f7ddf577606b5afda9b889

src/java.base/share/classes/jdk/internal/misc/MainMethodFinder.java
AlanBateman: > Yes. There is some juggling going on to get the code to align with the requirement and the interpretation thereof.

Is this stable now, specifically the "Selecting a main method" section and the behavior change where an instance main takes precedence over an "inherited" static main? Just checking before doing another pass over the method finder.

JimLaskey: I think so - additional changes will be coming from Joe via https://github.com/openjdk/jdk/pull/14165

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1832,12 +1869,29 @@ private String getCanonicalName0() {
...
+     *
+     * @apiNote
+     * An unnamed class is not an {@linkplain #isAnonymousClass anonymous class}.
+     *
+     * @since 21
+     */
+    @PreviewFeature(feature=PreviewFeature.Feature.UNNAMED_CLASSES,
+                    reflective=true)
+    public boolean isUnnamedClass() {
+        return PreviewFeatures.isEnabled() && isSynthetic() && isTopLevelClass();
```
jddarcy, line 1884: Please add a check for final-ness as @mlchung suggested in the other PR of the reflection changes stand-alone.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java
```diff
@@ -1367,10 +1368,17 @@ public String className() {
...
-        }
+         @Override @DefinedBy(Api.LANGUAGE_MODEL)
+         public Name getQualifiedName() {
+             return isUnnamed() ? fullname.subName(0, 0) /* empty name */ : fullname;
+             // return fullname;
+         }
+
+         @Override @DefinedBy(Api.LANGUAGE_MODEL)
+         public Name getSimpleName() {
+             // return isUnnamed() ? name.subName(0, 0) /* empty name */ : name;
```
jddarcy, line 1379: This line could be deleted.

test/langtools/tools/javac/processing/model/element/TestUnnamedClass.java
```diff
@@ -0,0 +1,200 @@
...
+    void checkUnnamedClassProperties(TypeElement unnamedClass) {
+        if (unnamedClass.getNestingKind() != NestingKind.TOP_LEVEL) {
+            messager.printError("Unnamed class is not top-level.", unnamedClass);
+        }
+
+        if (!unnamedClass.isUnnamed()) {
+            messager.printError("Unnamed class is _not_ indicated as such.", unnamedClass);
+        }
+
+        if (unnamedClass.getSimpleName().isEmpty()) {
```
jddarcy, line 140: Presumably the sense of the predicates testing names needs to now be reserved due to subsequent updates to the javax.lang.model support.

jddarcy (approved):
Core reflection and javax.lang.model APIs look fine.

src/java.base/share/classes/java/lang/Class.java
JimLaskey: Changing

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java
JimLaskey: Changing

test/langtools/tools/javac/processing/model/element/TestUnnamedClass.java
JimLaskey: True, but we should ensure that what we are shipping now does what we claim.

PR #13701, opened 2023-04-27
JDK-8294427 - Check boxes and radio buttons have rendering issues on Windows in High DPI env

rajamah:
Problem:

Check boxes and radio buttons in Windows Look-and-Feel have rendering issues when window is moved from display with one scale to display with a different scale on a multi-monitor setup:

- Scrawly ticks in checkboxes;
- Wrong circle relations in selected radio buttons.

Root-cause:
We open theme on AWT Toolkit Window which always has Primary Monitor DPI. 
Due to this when the app window goes to Secondary Screen with different DPI UI buttons
appear incorrectly rendered. 
Following is a list proposed changes to fix this issue.

Proposed Fix with Summary of changes:

1. Open a new Theme Handle per the DPI of the Screen where the App window is.
--> This makes sure we get the correct size for UI buttons based on the DPI of the screen where the app.
window is.

2. GetPartSize() of icons returns size based on standard size = 96 DPI.
--> This change makes sure that the default size of UI buttons is 96 since we use this on Java side to layout UI.

3. Rect size for icons in native paintBackground() function is fetched from Windows that specific DPI.
-->This makes sure that the UI buttons aren't stretched because the size calculated on Java side is different from what Windows      returns. Thus UI buttons are scaled correctly once we get their size back from Windows.
 
4. Adjust width and the height of the resolution variant image so that for scaling values of 1.25 , 2.25 , and such we  always  floor, while for 1.5, 1.75, 2.5, 2.75 , and such we always ceil.  	 
--> This helps make sure that for .25s scaling we get better rendering. 
This is because when we go from Double to Int for pixel rendering we sometimes need to floor or ceil to get crisp rendering.

As of now with these changes the rendering is crisp and good for all scaling factors with the exception .25s wherein some small artifact is seen sometimes in rendering of buttons but is still much better compared to what we have now.

Testing:

Tested locally on my Windows machine with a 2 monitor setup  125%, 150%, 175%, 225% scaling values and automated testing.
Also tested with SwingSet2 .

___________________________________
 Monitor 1                |    Monitor 2
(Primary) 		         |
				 |
        125%		 |    175%
	150%		 |    175%
	150%		 |    225%
	175%		 |    175%
	175%	         |    150%
	175%	         |    225%
_____________________ |_____________	

Also tested on setup with scaling values of  up-to 350%.

---------
### Progress

### Issue
 * [JDK-8294427](https://bugs.openjdk.org/browse/JDK-8294427): Check boxes and radio buttons have rendering issues on Windows in High DPI env (**Bug** - P3)

### Reviewers
 * [Alisen Chung](https://openjdk.org/census#achung) (@alisenchung - Committer) ⚠️ Review applies to [0719ec3b](https://git.openjdk.org/jdk/pull/13701/files/0719ec3b022fce1e4d290a5b593c39c5bd4bbb10)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13701/head:pull/13701` \
`$ git checkout pull/13701`

`$ git checkout pull/13701` \
`$ git pull https://git.openjdk.org/jdk.git pull/13701/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13701`

`$ git pr show -t 13701`

</details>

<a href="https://git.openjdk.org/jdk/pull/13701.diff">https://git.openjdk.org/jdk/pull/13701.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13701#issuecomment-1526282059)


src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
```diff
@@ -81,27 +83,45 @@ public static boolean isXPStyleEnabled() {
     }
 
     // this should be called only with writeLock held
-    private static Long getThemeImpl(String widget) {
-        Long theme = widgetToTheme.get(widget);
+    private static Long getThemeImpl(String widget, int dpi) {
+        Long theme = null;
+
+        if(dpiAwareWidgetToTheme.containsKey(dpi))
+        {
```
mrserb, line 90: It seems some changes use different types of formatting, please unify.

src/java.desktop/windows/native/libawt/windows/ThemeReader.cpp
```diff
@@ -116,8 +115,8 @@ BOOL InitThemes() {
     DTRACE_PRINTLN1("InitThemes hModThemes = %x\n", hModThemes);
     if(hModThemes) {
         DTRACE_PRINTLN("Loaded UxTheme.dll\n");
-        OpenThemeDataFunc = (PFNOPENTHEMEDATA)GetProcAddress(hModThemes,
-                                                        "OpenThemeData");
+        OpenThemeDataFuncForDpi = (PFNOPENTHEMEDATAFORDPI)GetProcAddress(
+                                   hModThemes, "OpenThemeDataForDpi");
```
mrserb, line 119: Do we plan to backport this change to the old versions of JDK? If yes will we plan to support old WIndows? If yes probably we should fallback to the "OpenThemeData" if "OpenThemeDataForDpi" is not available?

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
```diff
@@ -241,7 +277,7 @@ private static native Point getPoint(long theme, int part, int state,
     public static Point getPoint(String widget, int part, int state, int property) {
         readLock.lock();
         try {
-            return getPoint(getTheme(widget), part, state, property);
+            return getPoint(getTheme(widget, defaultDPI), part, state, property);
```
mrserb, line 280: Why does everything else use defaultDPI, is it possible that some values depend on the DPI?

src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/XPStyle.java
```diff
@@ -675,14 +679,20 @@ protected void paintToImage(Component c, Image image, Graphics g,
             w = bi.getWidth();
             h = bi.getHeight();
 
+            // Get DPI to pass further to ThemeReader.paintBackground()
+            Graphics2D g2d = (Graphics2D) g;
+            AffineTransform  at = g2d.getTransform();
+            int dpi = (int)(at.getScaleX() * 96);
```
mrserb, line 685: I think this will not work if the image is not VolatileImage, the scale will be = 1 and as a result, the image could be twice smaller than expected on the HiDPI monitor since rescale operation in another place is deleted. Please double-check.

src/java.desktop/windows/native/libawt/windows/ThemeReader.cpp
prrace: I guess this could (should?) be backported but Windows 10 is the oldest supported OS no matter what the JDK version.
And only the more recent versions of Windows 10 at that ...

aivanov-jdk: > I guess this could (should?) be backported but Windows 10 is the oldest supported OS no matter what the JDK version. And only the more recent versions of Windows 10 at that ...

I think it should be backported. The `OpenThemeDataForDpi` function was introduced in Windows 10 version 1703 which is already 5 years old and is unsupported, that is the function is available in the supported versions of Windows 10 and 11.

I may be useful to be able to run Java on previous versions though, yet it's not a priority.

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
aivanov-jdk: > Why does everything else use defaultDPI, is it possible that some values depend on the DPI?

Because Swing's user space coordinates use the default 96dpi. Some values do depend on the DPI, yet layout code still uses the default DPI.

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
```diff
@@ -111,11 +131,15 @@ private static Long getTheme(String widget) {
             try {
                 if (!valid) {
                     // Close old themes.
-                    for (Long value : widgetToTheme.values()) {
-                        closeTheme(value);
+                    if (!dpiAwareWidgetToTheme.isEmpty()) {
+                        for (Long value :
+                             dpiAwareWidgetToTheme.get(dpi).values()) {
```
aivanov-jdk, line 136: I think it's still possible that `get(dpi)` returns `null`, you should handle `null` gracefully.

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
```diff
@@ -111,11 +131,15 @@ private static Long getTheme(String widget) {
...
+                            closeTheme(value);
+                        }
+                        dpiAwareWidgetToTheme.get(dpi).clear();
+                        dpiAwareWidgetToTheme.clear();
```
aivanov-jdk, line 140: You clear the entire map, yet it don't call `closeTheme` for dpi values other than the passed one.

You should iterate over `dpiAwareWidgetToTheme.values().values()` (pseudo-code).

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
```diff
@@ -124,12 +148,17 @@ private static Long getTheme(String widget) {
         }
 
         // mostly copied from the javadoc for ReentrantReadWriteLock
-        Long theme = widgetToTheme.get(widget);
+        Long theme = null;
+
+        if(dpiAwareWidgetToTheme.containsKey(dpi))
+        {
+            theme = dpiAwareWidgetToTheme.get(dpi).get(widget);
```
aivanov-jdk, line 155: Avoid using `contains` followed by `get`, use `get` instead and handle `null` value.

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
```diff
@@ -139,14 +168,21 @@ private static Long getTheme(String widget) {
...
+                                               int rectBottom, int w, int h, int stride);
 
     public static void paintBackground(int[] buffer, String widget,
-           int part, int state, int x, int y, int w, int h, int stride) {
+           int part, int state, int x, int y, int w, int h, int stride, int dpi) {
         readLock.lock();
         try {
-            paintBackground(buffer, getTheme(widget), part, state, x, y, w, h, stride);
+
+            /* We get the part size vased on the theme for current screen DPI and pass it to paintBackground */
```
aivanov-jdk, line 179:
```suggestion
            /* We get the part size based on the theme for current screen DPI and pass it to paintBackground */
```

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
```diff
@@ -139,14 +168,21 @@ private static Long getTheme(String widget) {
...
+            Dimension d = getPartSize(getTheme(widget, dpi), part, state);
+            paintBackground(buffer,
+                            getTheme(widget, dpi),
+                            part, state,
+                            (int)d.getWidth(),
+                            (int)d.getHeight() , w, h, stride);
```
aivanov-jdk, line 185:
```suggestion
                            d.width,
                            d.height, w, h, stride);
```
You can access the fields directly and avoid casting.

src/java.desktop/windows/native/libawt/windows/ThemeReader.cpp
```diff
@@ -236,18 +238,20 @@ static void assert_result(HRESULT hres, JNIEnv *env) {
  * Signature: (Ljava/lang/String;)J
  */
 JNIEXPORT jlong JNICALL Java_sun_awt_windows_ThemeReader_openTheme
-(JNIEnv *env, jclass klass, jstring widget) {
+(JNIEnv* env, jclass klass, jstring widget, jint dpi) {
 
-    LPCTSTR str = (LPCTSTR) JNU_GetStringPlatformChars(env, widget, NULL);
+    LPCTSTR str = (LPCTSTR)JNU_GetStringPlatformChars(env, widget, NULL);
```
aivanov-jdk, line 243: I'd rather keep the original formatting for casts as well as for `JNIEnv *env` in the declaration.

src/java.desktop/share/classes/sun/swing/CachedPainter.java
```diff
@@ -314,8 +314,9 @@ public int getHeight(ImageObserver observer) {
 
         @Override
         public Image getResolutionVariant(double destWidth, double destHeight) {
-            int w = (int) Math.ceil(destWidth);
-            int h = (int) Math.ceil(destHeight);
+            int w = (int) Math.floor(destWidth + 0.5);
+            int h = (int) Math.floor(destHeight + 0.5);
```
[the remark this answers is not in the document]
rajamah: No, Region.clipRound does something little different than what I intend to do in
this code . 
It works same as my code for 0.25 and 0.75 case. But not for 0.5 case.
For example in case of destwidth and destHeight  = 19.5, 
for my current code I see them rounded to 20, while with clipRound it is 
rounded to 19.

src/java.desktop/windows/native/libawt/windows/ThemeReader.cpp
mrserb: > It may be useful to be able to run Java on previous versions though, yet it's not a priority.

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
rajamah: What is GC ?

aivanov-jdk: > What is GC ?

I guess Graphics Configuration.

> Then that layout code will be broken if layout will be calculated on already visible component, I guess it should take into account the GC where the component is located.

It works fine to the best of Rajat's and my testing because the layout doesn't change. The “icons” don't scale progressively, Windows returns the same size for many DPI values, Swing scales the pixels up, this fact is taken into account when rendering the theme image to avoid additional transformations.

Changing the preferred size dynamically based on the current DPI of the monitor could be a better way in the long term. Dropping `CachedPainter` from the DPI-aware rendering path of the themes is also on the table.

src/java.desktop/windows/native/libawt/windows/ThemeReader.cpp
aivanov-jdk: > > It may be useful to be able to run Java on previous versions though, yet it's not a priority.
> 
> Then probably it is better to add a fallback since that method was added to the win 10 after GA, and it does not exist in the early version.

But the earlier versions of Windows 10 are already unsupported, including the one where the `OpenThemeDataForDpi` function was added.

If required, the fallback could be implemented under a new bug. Anyway, I don't have a strong opinion on this matter.

src/java.desktop/share/classes/sun/swing/CachedPainter.java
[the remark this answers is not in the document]
mrserb: I guess it should do the same rounding the component do. If the size of the component is 19.5 and we round it to 20, then the image should use the same round logic. BTW why 19 is better than 20 here?

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
mrserb: We already changed the preferred size dynamically for most of the components, since the size depends on the fonts and the fonts depends on the GC, see:
https://hg.openjdk.org/jdk/jdk11/rev/55a43beaa529
https://hg.openjdk.org/jdk/jdk/rev/ea50ef230511
https://github.com/openjdk/jdk/commit/11aadc9d98d364b91114c028c7e2eff8de2f2bf0

src/java.desktop/share/classes/sun/swing/CachedPainter.java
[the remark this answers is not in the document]
rajamah: So, I added this code specifically to have improved rendering for the 25% cases, like 125%, 225%, 325% as rendering was a little off for these scale values and empirically based on what we saw it looked better with this calculation.
It is also important to note that windows is not consistent in what size it provides for a part(button) . For example for 150% it provides part size as 20 and we calculate it as 20 in above code.  For 175% windows still provides 20 again (while we would expect higher value) and we calculate it as 23 and it works out good. But as I said earlier for 125% or 225% it works best with above code.
The 20 vs 19 thing that I mentioned above was to just show that clipRound doesn't calculate it same was as my code and can have rendering repercussions as it is very sensitive to pixel variation , even 1 pixel variation causes  rendering artifacts.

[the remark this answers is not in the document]
mrserb: >So, I added this code specifically to have improved rendering for the 25% cases, like 125%, 225%, 325% as rendering was a little off f

But why it looks off if the clipRound is used? Does that size scales worse or does it cut the image at some direction?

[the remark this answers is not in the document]
mrserb: From the picture above it does not seem that the difference is just 1 pixel, no?

src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/XPStyle.java
```diff
@@ -675,14 +705,20 @@ protected void paintToImage(Component c, Image image, Graphics g,
             w = bi.getWidth();
             h = bi.getHeight();
 
+            // Get DPI to pass further to ThemeReader.paintBackground()
+            Graphics2D g2d = (Graphics2D) g;
+            AffineTransform  at = g2d.getTransform();
```
turbanoff, line 710: nit
```suggestion
            AffineTransform at = g2d.getTransform();
```

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
```diff
@@ -81,27 +82,38 @@ public static boolean isXPStyleEnabled() {
     }
 
     // this should be called only with writeLock held
-    private static Long getThemeImpl(String widget) {
-        Long theme = widgetToTheme.get(widget);
+    private static Long getThemeImpl(String widget, int dpi) {
+        Long theme = null;
+
+        if(dpiAwareWidgetToTheme.containsKey(dpi)) {
```
turbanoff, line 88: Can we just call `dpiAwareWidgetToTheme.get(dpi)` and compare result with `null` ? It's faster then 2 containsKey+get calls.

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
```diff
@@ -81,27 +82,38 @@ public static boolean isXPStyleEnabled() {
...
                 setWindowTheme(widget.substring(0, i));
-                theme = openTheme(widget.substring(i+2));
+                theme = openTheme(widget.substring(i + 2), dpi);
                 setWindowTheme(null);
             } else {
-                theme = openTheme(widget);
+                theme = openTheme(widget, dpi);
+            }
+
+            if (dpiAwareWidgetToTheme.get(dpi) != null) {
```
turbanoff, line 104: use computeIfAbsent

src/java.desktop/share/classes/sun/swing/CachedPainter.java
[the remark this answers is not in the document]
aivanov-jdk: Perhaps, this needs additional details. The size of checkbox and radio button is 13 as reported by Theme API for the standard DPI of 96. Being a prime number, 13 doesn't scale well, there's always fractional part unless the scale is whole. Thus, the scaled size never matches.

As the result, `CachedPainter` seems to apply a scale to compensate for that — whether we choose ceil or floot.

The entire thing should have been simpler: a `MultiResolutionImage` which paints the background according to the scale requested. Yet it proved to be not as simple. We're planning to look further into simplifying it if possible and, alternatively or in addition to, considering [dynamic layout based on graphics configuration](https://github.com/openjdk/jdk/pull/13701#discussion_r1187985627) as you suggested above.

The current solution may be not perfect but it provides better rendering compared to what we have now: *the controls look crisp only on the main display*.

src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/XPStyle.java
aivanov-jdk: Could you elaborate, please? How can this be not a `VolatileImage`?

If the components are painted into a `BufferedImage`, the scale needs to be specified explicitly if it's the intent.

src/java.desktop/share/classes/sun/swing/CachedPainter.java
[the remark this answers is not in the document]
mrserb: But still why the size on the components are so different? is it really caused by one "pixel/unit" we requested from the native?

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
```diff
@@ -103,18 +103,8 @@ private static Long getThemeImpl(String widget, int dpi) {
...
-                {
-                    dpiAwareWidgetToTheme.get(dpi).put(widget, theme);
-                }
-            }
-            else
-            {
-                dpiAwareWidgetToTheme.put(dpi, new HashMap<>());
-                dpiAwareWidgetToTheme.get(dpi).put(widget, theme);
-            }
+            dpiAwareWidgetToTheme.computeIfAbsent(dpi, key->new HashMap<>()).put(widget, theme);
```
aivanov-jdk, line 106:
```suggestion
            dpiAwareWidgetToTheme.computeIfAbsent(dpi, key -> new HashMap<>()).put(widget, theme);
```
There are usually spaces around `->` in lambda expressions.

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
```diff
@@ -132,9 +122,10 @@ private static Long getTheme(String widget, int dpi) {
                 if (!valid) {
                     // Close old themes.
                     if (!dpiAwareWidgetToTheme.isEmpty()) {
-                        for (Long value :
-                             dpiAwareWidgetToTheme.get(dpi).values()) {
-                            closeTheme(value);
+                        for (Map <String, Long> dpiVal: dpiAwareWidgetToTheme.values()) {
```
aivanov-jdk, line 125:
```suggestion
                        for (Map<String, Long> dpiVal : dpiAwareWidgetToTheme.values()) {
```
No space between type and its type arguments; spaces around `:` in for-each loop.

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
```diff
@@ -150,7 +141,7 @@ private static Long getTheme(String widget, int dpi) {
         // mostly copied from the javadoc for ReentrantReadWriteLock
         Long theme = null;
 
-        if(dpiAwareWidgetToTheme.containsKey(dpi))
+        if (dpiAwareWidgetToTheme.get(dpi) != null)
         {
```
aivanov-jdk:
```suggestion
        if (dpiAwareWidgetToTheme.get(dpi) != null) {
```
For consistent Java style.

src/java.desktop/share/classes/sun/swing/CachedPainter.java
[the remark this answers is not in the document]
rajamah: So, basically for w,h =19 we get a DPI of 140 and Dimension d = getPartSize(getTheme(widget, dpi), part, state); at line 172 of ThemeReader.java call returns a part size of 13 x 13 provided by windows as it is closer to DPI=120 (for 125% scaling).
While for  w,h =20 we get a DPI of 147 and Dimension d = getPartSize(getTheme(widget, dpi), part, state); at line 172 of ThemeReader.java call returns a part size of 20 x 20 provided by windows as it is closer to DPI=144 (for 150% scaling).
Thus part size calculated by us and provided by Windows matches and hence we see bigger properly rendered buttons while for the w,h=19 case we have 19x19 , while windows gives 13x13 
and hence you see scaled down buttons with bad rendering.

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
```diff
@@ -81,27 +82,36 @@ public static boolean isXPStyleEnabled() {
...
-    private static Long getThemeImpl(String widget) {
-        Long theme = widgetToTheme.get(widget);
+    private static Long getThemeImpl(String widget, int dpi) {
+        Long theme = null;
+
+        Map<String, Long> dpiWidgetToTheme = dpiAwareWidgetToTheme.get(dpi);
+
+        if (dpiWidgetToTheme != null) {
+            theme = dpiWidgetToTheme.get(widget);
+        }
```
aivanov-jdk, line 92: It looks the the entire method could be implemented as a chained called to `computeIfAbsent`:

```java
return dpiAwareWidgetToTheme.computeIfAbsent(dpi, key -> new HashMap<>())
                            .computeIfAbsent(widget, w -> this::reallyOpenTheme);
```

where `reallyOpenTheme` is a new helper method the body of which is that of the `if (theme == null)` block below.

If implementable, it'll make the code shorter and, hopefully, more readable.

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
```diff
@@ -124,12 +139,17 @@ private static Long getTheme(String widget) {
         }
 
         // mostly copied from the javadoc for ReentrantReadWriteLock
-        Long theme = widgetToTheme.get(widget);
+        Long theme = null;
+
+        if (dpiAwareWidgetToTheme.get(dpi) != null)
+        {
+            theme = dpiAwareWidgetToTheme.get(dpi).get(widget);
```
turbanoff, line 146: avoid calling `dpiAwareWidgetToTheme.get(dpi)` twice. Assign to a variable and reuse

src/java.desktop/windows/native/libawt/windows/ThemeReader.cpp
```diff
@@ -424,8 +428,9 @@ JNIEXPORT void JNICALL Java_sun_awt_windows_ThemeReader_paintBackground
 
     rect.left = 0;
     rect.top = 0;
-    rect.bottom = h;
-    rect.right = w;
+    rect.bottom = rectBottom ;
+    rect.right  = rectRight;
```
aivanov-jdk, line 432:
```suggestion
    rect.bottom = rectBottom;
    rect.right = rectRight;
```

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
```diff
@@ -111,11 +121,16 @@ private static Long getTheme(String widget) {
             try {
                 if (!valid) {
                     // Close old themes.
-                    for (Long value : widgetToTheme.values()) {
-                        closeTheme(value);
+                    if (!dpiAwareWidgetToTheme.isEmpty()) {
```
aivanov-jdk, line 124: The `!isEmpty()` seems redundant. If the map is empty, `values()` returns empty set; iterating over empty set does nothing.

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
```diff
@@ -111,11 +121,16 @@ private static Long getTheme(String widget) {
...
+                        for (Map<String, Long> dpiVal : dpiAwareWidgetToTheme.values()) {
+                            for (Long value : dpiVal.values()) {
+                                closeTheme(value);
+                            }
+                        }
+                        dpiAwareWidgetToTheme.get(dpi).clear();
+                        dpiAwareWidgetToTheme.clear();
+                        valid = true;
```
aivanov-jdk, line 132: The `valid` flag should be set to `true` even if the map was empty. Otherwise, it may never be set to `true`. One more point to dropping `if (!isEmpty())`.

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
```diff
@@ -111,11 +121,16 @@ private static Long getTheme(String widget) {
...
-                    for (Long value : widgetToTheme.values()) {
-                        closeTheme(value);
+                    if (!dpiAwareWidgetToTheme.isEmpty()) {
+                        for (Map<String, Long> dpiVal : dpiAwareWidgetToTheme.values()) {
+                            for (Long value : dpiVal.values()) {
+                                closeTheme(value);
+                            }
+                        }
+                        dpiAwareWidgetToTheme.get(dpi).clear();
+                        dpiAwareWidgetToTheme.clear();
```
aivanov-jdk, line 131:
```suggestion
                        for (Map<String, Long> dpiVal : dpiAwareWidgetToTheme.values()) {
                            for (Long value : dpiVal.values()) {
                                closeTheme(value);
                            }
                            dpiVal.clear();
                        }
                        dpiAwareWidgetToTheme.clear();
```
Avoid additional call to `get`.

aivanov-jdk (changes requested):

src/java.desktop/windows/classes/sun/awt/windows/ThemeReader.java
aivanov-jdk: This is even more serious: all the theme handles are closed but the maps for DPIs other than the passed `dpi` aren't cleared. With `clear` inside the loop, maps for all DPIs are cleared.

aivanov-jdk: I've thought about it more, and I was **wrong**.

Clearing the `dpiVal` map isn't needed at all because as the next step the outer map gets cleared, meaning that maps with theme handles are thrown away. Thus, no need to waste time to run `dpiVal.clear()` — `dpiAwareWidgetToTheme.clear()` takes care of those too.

That is remove the line `dpiVal.clear()`, please.

src/java.desktop/share/classes/sun/swing/CachedPainter.java
alisenchung, line 318: Does this problem only affect windows? Will this change in shared code affect the painting of other platforms?

rajamah: I think this function was added by the fix for [JDK-8143064](https://bugs.openjdk.org/browse/JDK-8143064) : Icons are not properly rendered with Windows L&F on HiDPI display

which was for Windows.
I tested my change on other platforms as well , didn't see any regression.

src/java.desktop/windows/native/libawt/windows/ThemeReader.cpp
```diff
@@ -424,8 +428,9 @@ JNIEXPORT void JNICALL Java_sun_awt_windows_ThemeReader_paintBackground
 
     rect.left = 0;
     rect.top = 0;
-    rect.bottom = h;
-    rect.right = w;
+    rect.bottom = rectBottom;
+    rect.right = rectRight;
+
```
aivanov-jdk, line 433: Perhaps, the newly added blank is redundant here.

aivanov-jdk (changes requested):
Overall, it looks good to me except for minor (stylistic) comments.

Please update the copyright year in the modified files.

src/java.desktop/share/classes/sun/swing/CachedPainter.java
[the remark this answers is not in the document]
mrserb: Probably I missed something but I would like to clarify. As you mention above we ask Windows to return the image for the component for some specific size, the Windows may return image of exactly requested size if available or any other size. What we will do if the size is different? In the image above it seems we use it as is, but should we? probably we should rescale it?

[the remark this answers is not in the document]
rajamah: So, we ask Windows for the component image based on a particular DPI , not some specific size.
The basic premise here is that we expect Windows returning image that would match the DPI , and then we perform our calculations accordingly. In the case above (if Region.clipRound is used) that basic premise is not met (if w, h=19) due to us calculating a DPI value of 140 which is closer to 120 or 125% scaling while our scaling is acytually 150% and Windows returns us a smaller size of the component than expected.
So we don't need to add extra code to rescale , I believe.
If we provide windows with correctly calculated DPI value which my code change does it should work fine as Windows will give us image of the correct size.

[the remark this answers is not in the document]
rajamah: The above code calls getImage() and passes width and height to it.

This is the flow :
getImage(){CachedPainter.java} -> paintToImage() {XPStyle.java} -> ThemeReader.paintBackground ()

We calculate and set scale in getImage() with this:

Line 172, CachedPainter.java

 if (volatileImage == null) {
     if ((w != baseWidth || h != baseHeight)) {
         g2.scale((double) w / baseWidth, 
                 (double) h / baseHeight); 
     }
     paintToImage(c, image, g2, baseWidth, baseHeight, args);
 }
 
 So, scale changes based on height and width of the image which we get from the
 code you mentioned above.
 
 This scale is then used in paintToImage() to calculate DPI at Line 711 of XPStyle.java:
 
   // Get DPI to pass further to ThemeReader.paintBackground()
  Graphics2D g2d = (Graphics2D) g;
  AffineTransform at = g2d.getTransform();
  int dpi = (int)(at.getScaleX() * 96);
 
 
 This DPI is then used to get the part size from windows and it changes as I explained in the previous comment.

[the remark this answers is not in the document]
mrserb: II am not sure that is correct way to calculate the dpi based on the "adhoc" change of the image size. That will add the mismatch between the code where we create the image, and the code where we request the image. We probably should change them both, or calculate the dpi in some other way.

Also the fact that on the image above one component is rendered twice smaller that another make me think that we did not rescale the image if the Windows return the image of the different size(You probably can prove that by always returning the wrong size from the native and check how it will look on HiDPI screen)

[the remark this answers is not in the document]
aivanov-jdk: > II am not sure that is correct way to calculate the dpi based on the "adhoc" change of the image size. That will add the mismatch between the code where we create the image, and the code where we request the image.

There's always a mismatch (at this time). We already discussed it [above](https://github.com/openjdk/jdk/pull/13701#discussion_r1195604571).

For example, for 125% scale, Windows returns 13×13 like for 100%; yet Swing expects 16.25. This is handled by rendering the 13×13-sized theme background into 16×16-sized bitmap — *to avoid any scale transformations being applied* to keep the image *crisp*.

> We probably should change them both, or calculate the dpi in some other way.

The DPI is calculated based on the scale set on the Graphics. Rajat provided the relevant excerpt:

https://github.com/openjdk/jdk/blob/727836ea85394b8baacd19a834f2993f06084c44/src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/XPStyle.java#L708-L711

This scale is calculated based on the size of the image.

So, we probably should… change. As [I already said](https://github.com/openjdk/jdk/pull/13701#discussion_r1195604571), this fix should've been as simple as a `MultiResolutionImage` where [`getResolutionVariant`](https://github.com/openjdk/jdk/blob/a780bf0dda4c2c3dce4337ccea37bcdce52045de/src/java.desktop/share/classes/sun/swing/CachedPainter.java#L316) returns the image of the correct size if it's already stored or retrieves it from the system. But it's not the case, and `CachedPainter` seems to be a hindrance rather than a helper.

That being said, we're planning to look further into simplifying it. Perhaps, `XPStyle.SkinPainter` shouldn't extend `CachedPainter`…

> Also the fact that on the image above one component is rendered twice smaller that another make me think that we did not rescale the image if the Windows return the image of the different size(You probably can prove that by always returning the wrong size from the native and check how it will look on HiDPI screen)

The point of the this fix is to *avoid rescaling* what Windows returns. And it *succeeds*: check boxes and radio buttons look much better — sharper and crisper — when you drag Swing UI frame from a display with one DPI to that with another; they even look better on the main display if its DPI is 120 or 168, that is 125% and 175% correspondingly.

src/java.desktop/windows/native/libawt/windows/ThemeReader.cpp
aivanov-jdk: I think it's *unfeasible* to provide the fallback to keep the code simpler.

If `OpenThemeDataForDpi` isn't available, we cannot get per-DPI images. Then we have to fallback to using the DPI of the main display and scale up / down all the returned values and images.

src/java.desktop/share/classes/sun/swing/CachedPainter.java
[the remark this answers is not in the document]
mrserb: > > II am not sure that is correct way to calculate the dpi based on the "adhoc" change of the image size. That will add the mismatch between the code where we create the image, and the code where we request the image.
> 
> There's always a mismatch (at this time). We already discussed it [above](https://github.com/openjdk/jdk/pull/13701#discussion_r1195604571).
> 
> For example, for 125% scale, Windows returns 13×13 like for 100%; yet Swing expects 16.25. This is handled by rendering the 13×13-sized theme background into 16×16-sized bitmap — _to avoid any scale transformations being applied_ to keep the image _crisp_.

Are you sure? Currently the code for image [creation](https://github.com/openjdk/jdk/blob/79a4ac791c826656b3e984fe54dc472c62efd028/src/java.desktop/share/classes/sun/swing/CachedPainter.java#L155) and the image [requests](https://github.com/openjdk/jdk/blob/79a4ac791c826656b3e984fe54dc472c62efd028/src/java.desktop/share/classes/sun/swing/CachedPainter.java#L317) use the same "ceil" logic.

> The DPI is calculated based on the scale set on the Graphics. Rajat provided the relevant excerpt:
> 
> https://github.com/openjdk/jdk/blob/727836ea85394b8baacd19a834f2993f06084c44/src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/XPStyle.java#L708-L711
> 
> This scale is calculated based on the size of the image.

I am still would like to get clarification about the size we changed by the patch, is it a size of the destination area we should render the component to, or something different. If it is a size of the destination area how we can change that to calculate the DPI? 

> > Also the fact that on the image above one component is rendered twice smaller that another make me think that we did not rescale the image if the Windows return the image of the different size(You probably can prove that by always returning the wrong size from the native and check how it will look on HiDPI screen)
> 
> The point of the this fix is to _avoid rescaling_ what Windows returns. And it _succeeds_: check boxes and radio buttons look much better — sharper and crisper — when you drag Swing UI frame from a display with one DPI to that with another; they even look better on the main display if its DPI is 120 or 168, that is 125% and 175% correspondingly.

Since we cannot guarantee that the Windows always return exact size we should have a backup code to rescale the image.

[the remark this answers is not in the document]
aivanov-jdk: > > There's always a mismatch (at this time). We already discussed it [above](https://github.com/openjdk/jdk/pull/13701#discussion_r1195604571).
> >
> > For example, for 125% scale, Windows returns 13×13 like for 100%; yet Swing expects 16.25. This is handled by rendering the 13×13-sized theme background into 16×16-sized bitmap — _to avoid any scale transformations being applied_ to keep the image _crisp_.
> 
> Are you sure? Currently the code for image [creation](https://github.com/openjdk/jdk/blob/79a4ac791c826656b3e984fe54dc472c62efd028/src/java.desktop/share/classes/sun/swing/CachedPainter.java#L155) and the image [requests](https://github.com/openjdk/jdk/blob/79a4ac791c826656b3e984fe54dc472c62efd028/src/java.desktop/share/classes/sun/swing/CachedPainter.java#L317) use the same "ceil" logic.

Yes, I am. The base width and height are 13×13, in 150% the parameters to `getResolutionVariant` are 16.25 and 16.25 — so the image created was 17×17.

Yet, as you can see, it doesn't match the destination width and height, so a scale to compensate for it is applied in `CachedPainter`.

With the updated formula, the image create is 16×16 which is closer to the requested size. The scale to compensate the different is still applied but it results in less distortion, which eventually looks better.

> > > Also the fact that on the image above one component is rendered twice smaller that another make me think that we did not rescale the image if the Windows return the image of the different size(You probably can prove that by always returning the wrong size from the native and check how it will look on HiDPI screen)
> > 
> > The point of the this fix is to _avoid rescaling_ what Windows returns. And it _succeeds_: check boxes and radio buttons look much better — sharper and crisper — when you drag Swing UI frame from a display with one DPI to that with another; they even look better on the main display if its DPI is 120 or 168, that is 125% and 175% correspondingly.
> 
> Since we cannot guarantee that the Windows always return exact size we should have a backup code to rescale the image.

No, we should not have a backup code to rescale the image: we request Windows to paint the image in the resolution appropriate for the DPI set for a display we're rendering for. If, for whatever reason, Windows returns an image that is too small for this DPI, you will notice it in all Windows native controls and apps, Java apps will not be different.

aivanov-jdk (approved):
Looks good to me.

Check boxes, radio buttons as well as checks in menus look crisp on all displays.

The following image displays the difference between the current and proposed versions. The current JDK code is on the left; build with code from this PR is on the right. SwingSet2 was moved from main 200%-scale display to a secondary 150%-scale display.

Radio check in menu, Check in menu and check boxes and radio buttons without the fix and with the fix

alisenchung (approved):

src/java.desktop/windows/native/libawt/windows/ThemeReader.cpp
```diff
@@ -90,7 +90,7 @@ typedef HRESULT (__stdcall *PFNGETTHEMETRANSITIONDURATION)
                 (HTHEME hTheme, int iPartId, int iStateIdFrom, int iStateIdTo,
                  int iPropId, DWORD *pdwDuration);
 
-static PFNOPENTHEMEDATA OpenThemeDataFunc = NULL;
+static PFNOPENTHEMEDATAFORDPI OpenThemeDataFuncForDpi = NULL;
```
aivanov-jdk, line 93: All other function have the `-Func` postfix, we should follow the naming convention.

aivanov-jdk (changes requested):

aivanov-jdk (approved):

PR #13706, opened 2023-04-27
8302496: Runtime.exit incorrectly says it never throws an exception

stuart-marks:
Textual specification clarifications.

---------
### Progress
- [x] Change requires CSR request [JDK-8307081](https://bugs.openjdk.org/browse/JDK-8307081) to be approved

### Issues
 * [JDK-8302496](https://bugs.openjdk.org/browse/JDK-8302496): Runtime.exit incorrectly says it never throws an exception
 * [JDK-8307081](https://bugs.openjdk.org/browse/JDK-8307081): Runtime.exit incorrectly says it never throws an exception (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13706/head:pull/13706` \
`$ git checkout pull/13706`

`$ git checkout pull/13706` \
`$ git pull https://git.openjdk.org/jdk.git pull/13706/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13706`

`$ git pr show -t 13706`

</details>

<a href="https://git.openjdk.org/jdk/pull/13706.diff">https://git.openjdk.org/jdk/pull/13706.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13706#issuecomment-1526753185)

### DIFF
--- a/src/java.base/share/classes/java/lang/Runtime.java
+++ b/src/java.base/share/classes/java/lang/Runtime.java
@@ -283,8 +283,7 @@ public class Runtime {
 
     /**
      * Immediately {@linkplain ##termination terminates} the Java Virtual Machine.
-     * If a security manager is present and its {@link SecurityManager#checkExit checkExit}
-     * method disallows exiting with the specified status, throws {@link SecurityException}.
+     * If the security manager denies exiting, throws {@link SecurityException}.
      * Otherwise, termination of the Java Virtual Machine is unconditional and immediate.
      * This method does not initiate the {@linkplain ##shutdown shutdown sequence}, nor does
      * it wait for the shutdown sequence to finish if it is already in progress. An
--- a/src/java.base/share/classes/java/lang/Runtime.java
+++ b/src/java.base/share/classes/java/lang/Runtime.java
@@ -141,18 +141,19 @@ public class Runtime {
 
     /**
      * Initiates the <a href="#shutdown">shutdown sequence</a> of the Java Virtual Machine.
-     * This method blocks indefinitely; it never returns or throws an exception (that is, it
-     * does not complete either normally or abruptly). The argument serves as a status code;
-     * by convention, a nonzero status code indicates abnormal termination.
+     * If a security manager is present and its {@link SecurityManager#checkExit checkExit}
+     * method disallows exiting with the specified status, throws {@link SecurityException}.
+     * Otherwise, this method blocks indefinitely; it neither returns nor throws an exception
+     * (that is, it does not complete either normally or abruptly). The argument serves as a
+     * status code. By convention, a nonzero status code indicates abnormal termination.
      *
-     * <p> Invocations of this method are serialized such that only one
-     * invocation will actually proceed with the shutdown sequence and
-     * terminate the VM with the given status code. All other invocations
-     * simply block indefinitely.
+     * <p> Successful invocations of this method are serialized such that only one invocation
+     * will proceed with the shutdown sequence and terminate the VM with the given status code.
+     * All other invocations will perform no action and block indefinitely.
      *
-     * <p> Because this method always blocks indefinitely, if it is invoked from
-     * a shutdown hook, it will prevent that shutdown hook from terminating.
-     * Consequently, this will prevent the shutdown sequence from finishing.
+     * <p> Because a successful invocation of this method blocks indefinitely, if it is invoked
+     * from a shutdown hook, it will prevent that shutdown hook from terminating. Consequently,
+     * this will prevent the shutdown sequence from finishing.
      *
      * <p> The {@link System#exit(int) System.exit} method is the
      * conventional and convenient means of invoking this method.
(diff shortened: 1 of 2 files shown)


stuart-marks:
Please review CSR [JDK-8307081](https://bugs.openjdk.org/browse/JDK-8307081).

src/java.base/share/classes/java/lang/Runtime.java
```diff
@@ -141,18 +141,19 @@ private Runtime() {}
 
     /**
      * Initiates the <a href="#shutdown">shutdown sequence</a> of the Java Virtual Machine.
```
ExE-Boss, line 143:
```suggestion
     * Initiates the {@linkplain ##shutdown shutdown sequence} of the Java Virtual Machine.
```

src/java.base/share/classes/java/lang/Runtime.java
```diff
@@ -280,10 +281,13 @@ public boolean removeShutdownHook(Thread hook) {
     }
 
     /**
-     * Immediately <a href="#termination">terminates</a> the Java Virtual Machine. Termination
-     * is unconditional and immediate. This method does not initiate the
-     * <a href="#shutdown">shutdown sequence</a>, nor does it wait for the shutdown sequence
-     * to finish if it is already in progress. This method never returns normally.
+     * Immediately <a href="#termination">terminates</a> the Java Virtual Machine.
```
ExE-Boss, line 284:
```suggestion
     * Immediately {@linkplain ##termination terminates} the Java Virtual Machine.
```

src/java.base/share/classes/java/lang/System.java
```diff
@@ -1886,18 +1886,17 @@ public static Logger getLogger(String name, ResourceBundle bundle) {
...
 
     /**
-     * Initiates the <a href="Runtime.html#shutdown">shutdown sequence</a> of the
-     * Java Virtual Machine. This method always blocks indefinitely. The argument
-     * serves as a status code; by convention, a nonzero status code indicates
-     * abnormal termination.
-     * <p>
-     * This method calls the {@code exit} method in class {@code Runtime}. This
-     * method never returns normally.
+     * Initiates the <a href="#shutdown">shutdown sequence</a> of the Java Virtual Machine.
```
ExE-Boss, line 1889:
```suggestion
     * Initiates the {@linkplain Runtime##shutdown shutdown sequence} of the Java Virtual Machine.
```

ExE-Boss (commented):
These can use the new generalised `@linkplain` tags available since **JDK 20**:

src/java.base/share/classes/java/lang/Runtime.java
```diff
@@ -141,18 +141,19 @@ private Runtime() {}
 
     /**
      * Initiates the <a href="#shutdown">shutdown sequence</a> of the Java Virtual Machine.
-     * This method blocks indefinitely; it never returns or throws an exception (that is, it
-     * does not complete either normally or abruptly). The argument serves as a status code;
-     * by convention, a nonzero status code indicates abnormal termination.
+     * If a security manager is present and its {@link SecurityManager#checkExit checkExit}
+     * method disallows exiting with the specified status, throws {@link SecurityException}.
+     * Otherwise, this method blocks indefinitely; it neither returns nor throws an exception
```
AlanBateman, line 146: The SM is not long for this world.  So I'm in two minds about it making a farewell appearance in the first paragraph of the method description when we know we will have to remove this text very soon.  If we really need to change this now then maybe it could be reduced to "When a security manager is not set, this method never returns ..." so that the distraction of checkExit is not in the first paragraph.

src/java.base/share/classes/java/lang/Runtime.java
stuart-marks: Good idea. I've updated these and a couple more relating to the shutdown sequence and termination. (There are lots more but in unrelated areas, so I left them for now.)

src/java.base/share/classes/java/lang/Runtime.java
stuart-marks: Yeah I had some misgivings mentioning the SM first, but if it's not mentioned, then the assertions in the rest of the text become hedged and uncertain. The SM, checkExit, and whether the exit value is allowed all have to be mentioned somewhere anyway. It would be awkward to try to wedge all of this into a subordinate or parenthetical clause of some other sentence.

Even though we know the SM is on the way out, it's still part of Java SE 21, and it ought to be specified correctly. When the SM is actually removed, we'll have to revisit these specs anyway. Choosing different wording now won't make the future change any easier or harder.

AlanBateman: One thing you could do is reduce the words down to something like "Unless denied exiting by the security manager, this method blocks indefinitely ..." to avoid complicating the first paragraph with something that only arises with a niche execution mode. If needed, a later paragraph could have this detail, or just leave it to the existing @throws text where the permission check is specified.

stuart-marks: Updated.

AlanBateman (approved):
Updated version, and the wording of halt's description, looks okay.

stuart-marks:
@AlanBateman Great. Can you take a look at the CSR [JDK-8307081](https://bugs.openjdk.org/browse/JDK-8307081) as well? Thanks.

PR #13711, opened 2023-04-28
8293117: Add atomic bitset functions

kimbarrett:
Please review this change that adds and/or/xor bitops to Atomic.

Some offline discussion reached consensus on the following API (with T being a
template parameter that must be an integral type):
  T fetch_then_and(volatile T* dest, T bits, atomic_memory_order order)
  T fetch_then_or(volatile T* dest, T bits, atomic_memory_order order)
  T fetch_then_xor(volatile T* dest, T bits, atomic_memory_order order)
  T and_then_fetch(volatile T* dest, T bits, atomic_memory_order order)
  T or_then_fetch(volatile T* dest, T bits, atomic_memory_order order)
  T xor_then_fetch(volatile T* dest, T bits, atomic_memory_order order)
with order defaulting to memory_order_conservative.

I'm hoping there won't be more bike-shedding on the names.

This naming convention differs from what exists for add/sub, where we have
fetch_and_add, add_and_fetch, and the like.  fetch_and_and and and_and_fetch
just looked too weird.  For consistency we should probably rename the add/sub
operations to use "then".

A default implementation is provided, using a CAS loop for all operations.
That implementation can be overridden on a per-platform basis.  Currently
there aren't any platform-specific implementations; such will be added by
followup RFEs.  We'll want to override for most (if not all) platforms, since
the CAS loop implementation is generally less than optimal (in some cases
rather comically so).

For some platforms a CAS loop is necessary to obtain either the old or the new
value, but have a better implementation available if neither value is needed.
However, we can't easily make use of such.  We would need to either add
functions returning void to the API or change the API to return expression
templates so we can detect how the result is used.  However, we may be able to
use compiler intrinsics for some platforms, which may be able to specialize
for the usage context.  If that turns out to be a problem then we can expand
the API accordingly later.

Testing:
mach5 tier1, including new gtests for the new functionality.

---------
### Progress

### Issue
 * [JDK-8293117](https://bugs.openjdk.org/browse/JDK-8293117): Add atomic bitset functions

### Reviewers
 * [Coleen Phillimore](https://openjdk.org/census#coleenp) (@coleenp - **Reviewer**) ⚠️ Review applies to [dbb0f3ca](https://git.openjdk.org/jdk/pull/13711/files/dbb0f3ca099c7be0bac48f0350d44903edd52204)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13711/head:pull/13711` \
`$ git checkout pull/13711`

`$ git checkout pull/13711` \
`$ git pull https://git.openjdk.org/jdk.git pull/13711/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13711`

`$ git pr show -t 13711`

</details>

<a href="https://git.openjdk.org/jdk/pull/13711.diff">https://git.openjdk.org/jdk/pull/13711.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13711#issuecomment-1527121028)


src/hotspot/share/runtime/atomic.hpp
```diff
@@ -576,6 +690,99 @@ struct Atomic::PlatformXchg {
...
+class Atomic::PrefetchBitopsUsingCmpxchg {
+  template<typename T, typename Op>
+  T bitop(T volatile* dest, atomic_memory_order order, Op operation) const {
+    T old_value;
+    T new_value;
+    T fetched_value = Atomic::load(dest);
+    do {
+      old_value = fetched_value;
+      new_value = operation(old_value);
+      fetched_value = Atomic::cmpxchg(dest, old_value, new_value, order);
```
shipilev, line 703: At some point in future, it would be nice to replace the retry-loops like these with _weak_ CASes. It could be done in platform-specific code, though, but it would seem awkward to specialize for all LL/SC-retry-loop platforms just to get access to weak CAS.

src/hotspot/share/runtime/atomic.hpp
```diff
@@ -576,6 +690,99 @@ struct Atomic::PlatformXchg {
...
+  template<typename T>
+  T or_then_fetch(T volatile* dest, T bits, atomic_memory_order order) const {
+    return bits | Atomic::fetch_then_or(dest, bits, order);
+  }
+
+  template<typename T>
+  T xor_then_fetch(T volatile* dest, T bits, atomic_memory_order order) const {
+    return bits ^ Atomic::fetch_then_xor(dest, bits, order);
+  }
+};
```
shipilev, line 775: Is this used somewhere? Looks like all current helpers use CASes instead of this "prefetch" code.

test/hotspot/gtest/runtime/test_atomic.cpp
```diff
@@ -199,3 +199,97 @@ TEST(AtomicEnumTest, scoped_enum) {
...
+template<typename T>
+const T AtomicBitopsTestSupport<T>::_change_value;
+
+TEST(AtomicBitopsTest, int32) {
+  AtomicBitopsTestSupport<int32_t>()();
+}
+
+TEST(AtomicBitopsTest, int64) {
+  AtomicBitopsTestSupport<int64_t>()();
+}
```
shipilev, line 287: These 64-bit tests would fail on 32-bit systems, are they not? They are outside the requirements for the contract to only support `sizeof(int)` and `sizeof(void*`). Probably just `#ifdef _LP64` these?

shipilev (commented):
This look okay, a couple of comments.

coleenp (approved):
Thank you for doing this.  I'll file a follow up to convert some duplicated code to use these after this is pushed.

src/hotspot/share/runtime/atomic.hpp
kimbarrett: Yes.  I haven't checked, but I would guess the majority of our CAS uses are loops that would benefit from
using a weak CAS on platforms where that's available.

src/hotspot/share/runtime/atomic.hpp
kimbarrett: Not used yet.  It's used by some of the platform-specific implementations I have queued up.

test/hotspot/gtest/runtime/test_atomic.cpp
kimbarrett: Well-spotted.  I forgot to conditionalize for `_LP64` as is done for other tests in this file.

shipilev (approved):

dholmes-ora (approved):
Seems quite reasonable.

I think for the naming it wouldn't be a terrible idea to consider using uppercase for the operation name: AND/OR/XOR/ADD/SUB. That would make the operation name stand out more clearly. Just a thought.

Thanks.

kimbarrett:
> @kimbarrett do you have GHA configured? This says 3 checks have passed but this sort of change might have problems compiling for other platforms.

Manually triggered GHA testing - completed successfully.

kimbarrett:
Thanks for reviews @coleenp , @shipilev , and @dholmes-ora .

PR #13716, opened 2023-04-28
8306929: Avoid CleanClassLoaderDataMetaspaces safepoints when previous versions are shared

kstefanj:
Hi all, 

Please review this change to avoid CleanClassLoaderDataMetaspaces safepoint when there is nothing that can be cleaned up.

**Summary**
When transforming/redefining classes a previous version list is linked together in the InstanceKlass. The original class is added to this list if it is still used or shared. The difference between shared and used is not currently noted. This leads to a problem when doing concurrent class unloading, because during that we postpone some potential work to a safepoint (since we are not in one). This is the CleanClassLoaderDataMetaspaces and it is triggered by the ServiceThread if there is work to be done, for example if InstanceKlass::_has_previous_versions is true.

Since we currently does not differentiate between shared and "in use" we always set _has_previous_versions if anything is on this list. This together with the fact that shared previous versions should never be cleaned out leads to this safepoint being triggered after every concurrent class unloading even though there is nothing that can be cleaned out.

This can be avoided by making sure the _previous_versions list is only cleaned when there are non-shared classes on it. This change renames `_has_previous_versions` to `_clean_previous_versions` and only updates it if we have non-shared classes on the list.

**Testing**
* A lot of manual testing verifying that we do get the safepoint when we should. 
* Added new test to verify expected behavior by parsing the logs. The test uses JFR to trigger redefinition of some shared classes (when -Xshare:on).
* Mach5 run of new test and tier 1-3

---------
### Progress

### Issue
 * [JDK-8306929](https://bugs.openjdk.org/browse/JDK-8306929): Avoid CleanClassLoaderDataMetaspaces safepoints when previous versions are shared

### Reviewers
 * [Coleen Phillimore](https://openjdk.org/census#coleenp) (@coleenp - **Reviewer**) ⚠️ Review applies to [39c3a1c1](https://git.openjdk.org/jdk/pull/13716/files/39c3a1c1e4c4a00ea04770a03656ae0a0611de4d)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13716/head:pull/13716` \
`$ git checkout pull/13716`

`$ git checkout pull/13716` \
`$ git pull https://git.openjdk.org/jdk.git pull/13716/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13716`

`$ git pr show -t 13716`

</details>

<a href="https://git.openjdk.org/jdk/pull/13716.diff">https://git.openjdk.org/jdk/pull/13716.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13716#issuecomment-1527543176)

### DIFF
--- a/src/hotspot/share/oops/instanceKlass.hpp
+++ b/src/hotspot/share/oops/instanceKlass.hpp
@@ -715,7 +715,7 @@ public:
   }
 
  private:
-  static bool  _clean_previous_versions;
+  static bool  _should_clean_previous_versions;
  public:
   static void purge_previous_versions(InstanceKlass* ik) {
     if (ik->has_been_redefined()) {
@@ -724,7 +724,7 @@ public:
   }
 
   static bool should_clean_previous_versions_and_reset();
-  static bool clean_previous_versions() { return _clean_previous_versions; }
+  static bool should_clean_previous_versions() { return _should_clean_previous_versions; }
 
   // JVMTI: Support for caching a class file before it is modified by an agent that can do retransformation
   void set_cached_class_file(JvmtiCachedClassFileData *data) {
--- a/src/hotspot/share/oops/instanceKlass.hpp
+++ b/src/hotspot/share/oops/instanceKlass.hpp
@@ -715,7 +715,7 @@ public:
   }
 
  private:
-  static bool  _has_previous_versions;
+  static bool  _clean_previous_versions;
  public:
   static void purge_previous_versions(InstanceKlass* ik) {
     if (ik->has_been_redefined()) {
@@ -723,8 +723,8 @@ public:
     }
   }
 
-  static bool has_previous_versions_and_reset();
-  static bool has_previous_versions() { return _has_previous_versions; }
+  static bool should_clean_previous_versions_and_reset();
+  static bool clean_previous_versions() { return _clean_previous_versions; }
 
   // JVMTI: Support for caching a class file before it is modified by an agent that can do retransformation
   void set_cached_class_file(JvmtiCachedClassFileData *data) {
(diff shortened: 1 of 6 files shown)


coleenp (approved):
This looks good. Thanks for all the testing and adding the new test.

src/hotspot/share/oops/instanceKlass.hpp
```diff
@@ -715,16 +715,16 @@ class InstanceKlass: public Klass {
   }
 
  private:
-  static bool  _has_previous_versions;
+  static bool  _clean_previous_versions;
```
sspitsyn, line 718: Nit: I'd suggest to name it as `_should_clean_previous_versions`.
       Then the corresponding function needs to be named as `should_clean_previous_versions()`.

test/hotspot/jtreg/serviceability/jvmti/RedefineClasses/RedefineSharedClassJFR.java
```diff
@@ -0,0 +1,102 @@
...
+                    .shouldContain("Class unloading: clean_previous_versions = false")
+                    .shouldNotContain("Class unloading: clean_previous_versions = true")
+                    .shouldContain("scratch class added; class is shared")
+                    // If the below line occurs, then clean_previous_versions will be
+                    // true and the above shouldContain will trigger. This check is to
+                    // show the intention that we don't expect any non-shared transformed
+                    // classes to be in use.
+                    .shouldNotContain("scratch class added; one of its methods is on_stack.")
+                    .shouldHaveExitValue(0);
+                return;
```
sspitsyn, line 94: The fragments 61-74 and 79-93 have a big common part which can be a good base for a refactoring.
But it can be not worth it. So, I leave it up to you.

sspitsyn (commented):
Thank you for taking care about it.
I've posted a couple of comments but it it looks good anyway.
Thanks,
Serguei

sspitsyn (approved):
Thank you for the update.
Looks good.
Thanks,
Serguei

kstefanj:
Thanks again @coleenp and @sspitsyn for the reviews.

/integrate

PR #13718, opened 2023-04-28
8307106: Allow concurrent GCs to walk CLDG without ClassLoaderDataGraph_lock

fisk:
A concurrent GC with concurrent class unloading can't currently walk the CLDG without the CLDG_lock today. We should add some synchronization code so it can do that safely. This patch adds the missing bits.

---------
### Progress

### Issue
 * [JDK-8307106](https://bugs.openjdk.org/browse/JDK-8307106): Allow concurrent GCs to walk CLDG without ClassLoaderDataGraph_lock

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13718/head:pull/13718` \
`$ git checkout pull/13718`

`$ git checkout pull/13718` \
`$ git pull https://git.openjdk.org/jdk.git pull/13718/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13718`

`$ git pr show -t 13718`

</details>

<a href="https://git.openjdk.org/jdk/pull/13718.diff">https://git.openjdk.org/jdk/pull/13718.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13718#issuecomment-1527572346)

### DIFF
--- a/src/hotspot/share/classfile/classLoaderDataGraph.cpp
+++ b/src/hotspot/share/classfile/classLoaderDataGraph.cpp
@@ -261,8 +261,15 @@ ClassLoaderData* ClassLoaderDataGraph::add(Handle loader, bool has_class_mirror_
   return loader_data;
 }
 
+inline void assert_is_safepoint_or_gc() {
+  assert(SafepointSynchronize::is_at_safepoint() ||
+         Thread::current()->is_ConcurrentGC_thread() ||
+         Thread::current()->is_Worker_thread(),
+         "Must be called by safepoint or GC");
+}
+
 void ClassLoaderDataGraph::cld_unloading_do(CLDClosure* cl) {
-  assert_locked_or_safepoint_weak(ClassLoaderDataGraph_lock);
+  assert_is_safepoint_or_gc();
   for (ClassLoaderData* cld = _unloading; cld != nullptr; cld = cld->next()) {
     assert(cld->is_unloading(), "invariant");
     cl->do_cld(cld);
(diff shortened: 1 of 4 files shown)


stefank (approved):

xmas92 (approved):

coleenp (approved):
Looks good to me too. Thanks for the preview, Erik.

src/hotspot/share/classfile/classLoaderDataGraph.cpp
```diff
@@ -520,7 +527,8 @@ bool ClassLoaderDataGraph::do_unloading() {
       prev->set_next(data);
     } else {
       assert(dead == _head, "sanity check");
-      _head = data;
+      // The GC might be walking this concurrently
+      Atomic::store(&_head, data);
```
dholmes-ora, line 531: If you are using `load_acquire` then surely this must be a `release-store` as they need to be paired.

dholmes-ora (commented):
I don't see any change here where code stops acquiring the CLDG lock. Is this just a preparatory change?

Using acquire/release etc is certainly necessary to be able to walk the list safely, but it is not obvious it is sufficient. I can't tell when nodes in the CLDG actually get deleted such that the GC thread doing the walking can't access a no longer existing node.

Thanks.

src/hotspot/share/classfile/classLoaderDataGraph.cpp
fisk: The load_acquire synchronizes with release_store on insertions. This relaxed unlinking store, doesn't publish anything new that the load_acquire is interested in. It rolls back the head to a CLD that was installed at some earlier point with release_store, meaning it has already been published safely before.

dholmes-ora: Thanks for clarifying

dholmes-ora (approved):
Thanks for the additional explanations @fisk.

fisk:
> Thanks for the additional explanations @fisk.

Thanks @dholmes-ora!

PR #13719, opened 2023-04-28
8305714 : Add an extra test for JDK-8292755

weibxiao:
Add a testing case for undefined classes in Java JShell to make sure JShell does not crash when an undefined class is used.

This test was passed for Linux, MacOS, and Windows by mach5 build and test in Oracle.

Reviewer: @coffeys @lahodaj

---------
### Progress

### Issue
 * [JDK-8305714](https://bugs.openjdk.org/browse/JDK-8305714): Add an extra test for JDK-8292755

### Reviewers
 * [Sean Coffey](https://openjdk.org/census#coffeys) (@coffeys - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13719/head:pull/13719` \
`$ git checkout pull/13719`

`$ git checkout pull/13719` \
`$ git pull https://git.openjdk.org/jdk.git pull/13719/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13719`

`$ git pr show -t 13719`

</details>

<a href="https://git.openjdk.org/jdk/pull/13719.diff">https://git.openjdk.org/jdk/pull/13719.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13719#issuecomment-1527644599)

### DIFF
--- a/test/langtools/jdk/jshell/UndefinedClassTest.java
+++ b/test/langtools/jdk/jshell/UndefinedClassTest.java
@@ -23,7 +23,7 @@
 
 /**
  * @test
- * @bug 8305714
+ * @bug 8292755
  * @summary make sure JShell not crashing while throwing undefined exception
  * @modules
  *     jdk.compiler/com.sun.tools.javac.api
--- a/test/langtools/jdk/jshell/UndefinedClassTest.java
+++ b/test/langtools/jdk/jshell/UndefinedClassTest.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/**
+ * @test
+ * @bug 8305714
+ * @summary make sure JShell not crashing while throwing undefined exception
+ * @modules
+ *     jdk.compiler/com.sun.tools.javac.api
+ *     jdk.compiler/com.sun.tools.javac.main
+ *     jdk.jshell/jdk.internal.jshell.tool:open
+ *     jdk.jshell/jdk.internal.jshell.tool.resources:open
+ *     jdk.jshell/jdk.jshell:open
+ * @library /tools/lib
+ * @build toolbox.ToolBox toolbox.JarTask toolbox.JavacTask
+ * @build Compiler UITesting
+ * @build UndefinedClassTest
+ * @run testng UndefinedClassTest
+ */
+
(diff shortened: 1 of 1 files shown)


test/langtools/jdk/jshell/UndefinedClassTest.java
```diff
@@ -0,0 +1,67 @@
...
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/**
+ * @test
+ * @bug 8305714
```
coffeys, line 26: 8292755 would be the bug ID here since it refers to the product bug. 

Is there a reason why you couldn't fold this test into the test/langtools/jdk/jshell/ClassesTest.java test ?

weibxiao: Updated the bug ID.

Assuming ClassesTest is to verify the definition of Class.

With this patch, I can catch the following error message described in the attached file in JShell Console. 
image


Without the patch of 8292755 for JDK17,  JShell will crash with this test case. After applying this patch in JDK17U, JShell console will print something like below, 

Exception in thread "main" java.lang.InternalError: Exception during analyze - java.lang.AssertionError: isSubtype UNKNOWN
at jdk.jshell/jdk.jshell.TaskFactory$AnalyzeTask.analyze(TaskFactory.java:415)
at jdk.jshell/jdk.jshell.TaskFactory$AnalyzeTask.<init>(TaskFactory.java:406)
at jdk.jshell/jdk.jshell.TaskFactory.lambda$analyze$1(TaskFactory.java:178)
at jdk.jshell/jdk.jshell.TaskFactory.lambda$runTask$4(TaskFactory.java:213)
at jdk.compiler/com.sun.tools.javac.api.JavacTaskPool.getTask(JavacTaskPool.java:193)
at jdk.jshell/jdk.jshell.TaskFactory.runTask(TaskFactory.java:206)
at jdk.jshell/jdk.jshell.TaskFactory.analyze(TaskFactory.java:175)

test/langtools/jdk/jshell/UndefinedClassTest.java
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/**
+ * @test
+ * @bug 8292755
+ * @summary make sure JShell not crashing while throwing undefined exception
```
coffeys, line 27: I'd suggest "InternalError seen while throwing undefined exception" as summary.

test/langtools/jdk/jshell/UndefinedClassTest.java
```diff
@@ -0,0 +1,67 @@
...
+    public UndefinedClassTest() {
+        super(true);
+    }
+
+    public void testUndefinedClass() throws Exception{
+        String code = "@FunctionalInterface\n" +
+                "interface RunnableWithThrowable {\n" +
+                " void run() throws Throwable;\n" +
+                "\n" +
+                " // You can also replace `static` with `default` and test again\n" +
```
coffeys, line 55: why the comment in test code ? maybe you can create a 2nd test instead :

```
    public void testUndefinedClass1() throws Exception{
        String code = "@FunctionalInterface\n" +
                "interface RunnableWithThrowable {\n" +
                " void run() throws Throwable;\n" +
                "\n" +
                " default RunnableWithThrowable getInstance() {\n" +
                " return () -> { throw new NotExist(); };\n" +
                " }\n" +
                "}";
        // set terminal height so that help output won't hit page breaks
        System.setProperty("test.terminal.height", "1000000");
        doRunTest((inputSink, out) -> {
            inputSink.write(code + "\n");
            waitOutput(out, "NotExist");
        });
    }

```

coffeys (commented):
Looks good - minor suggestions

test/langtools/jdk/jshell/UndefinedClassTest.java
weibxiao: Updated the summary. And I remove the code to set the terminal height. I don't think it is necessary since there are only two lines for the print-out.

weibxiao:
@lahodaj , let me know if you have any concerns and comments. Thanks.

coffeys (approved):

PR #13721, opened 2023-04-28
8305903: Deflate monitors of dead objects before they become unreachable

rkennke:
With compact object headers ([JDK-8305895](https://bugs.openjdk.org/browse/JDK-8305895)), I've seen occasional failures in G1, where the refinement thread tries to parse a heap region that has dead objects, and would sometimes see an object with a monitor that has already been deflated. And because deflation does not bother to restore the header of dead objects, when heap iteration tries to load the Klass* of the dead object, it would reach to unknown memory and crash.

In OM::deflate_monitor() we check object_peek(), and if that returns null, then the object header is not updated (and can't be, because the object cannot be reached anymore). Concurrent GCs that process weak handles concurrently ensure that the object doesn't leak out by returning null there (via a barrier). However, for runtime code, at this point, there is no safe way to grab the object and update the header, because the GC might already have reclaimed it. The last safe point in time where we can do that is in WeakProcessor::Task::work() and OopStorage::weak_oops_do() itself, as soon as we detect that the object is properly unreachable.

The fix is to restore the header of dead objects just before they become unreachable. This can be done in the closures used by WeakProcessor::weak_oops_do(), right before the weak root will be cleared.

Notice that this is only a bug with compact object headers. It doesn't hurt to fix this in general, though.

Testing:
 - [x] tier1
 - [x] tier2

---------
### Progress

### Issue
 * [JDK-8305903](https://bugs.openjdk.org/browse/JDK-8305903): Deflate monitors of dead objects before they become unreachable

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13721/head:pull/13721` \
`$ git checkout pull/13721`

`$ git checkout pull/13721` \
`$ git pull https://git.openjdk.org/jdk.git pull/13721/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13721`

`$ git pr show -t 13721`

</details>

<a href="https://git.openjdk.org/jdk/pull/13721.diff">https://git.openjdk.org/jdk/pull/13721.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13721#issuecomment-1527998038)

### DIFF
--- a/src/hotspot/share/gc/shared/oopStorage.inline.hpp
+++ b/src/hotspot/share/gc/shared/oopStorage.inline.hpp
@@ -29,6 +29,7 @@
 
 #include "memory/allocation.hpp"
 #include "oops/oop.hpp"
+#include "runtime/objectMonitor.hpp"
 #include "runtime/safepoint.hpp"
 #include "utilities/align.hpp"
 #include "utilities/count_trailing_zeros.hpp"
@@ -262,6 +263,7 @@ public:
       if (_is_alive->do_object_b(v)) {
         result = _f(ptr);
       } else {
+        ObjectMonitor::maybe_deflate_dead(ptr);
         *ptr = nullptr;            // Clear dead value.
       }
     }
--- a/src/hotspot/share/gc/shared/weakProcessor.inline.hpp
+++ b/src/hotspot/share/gc/shared/weakProcessor.inline.hpp
@@ -34,6 +34,7 @@
 #include "gc/shared/weakProcessorTimes.hpp"
 #include "gc/shared/workerThread.hpp"
 #include "prims/resolvedMethodTable.hpp"
+#include "runtime/objectMonitor.hpp"
 #include "utilities/debug.hpp"
 #include "utilities/enumIterator.hpp"
 
@@ -65,6 +66,7 @@ public:
       _keep_alive->do_oop(p);
       ++_live;
     } else {
+      ObjectMonitor::maybe_deflate_dead(p);
       *p = nullptr;
       ++_new_dead;
     }
--- a/src/hotspot/share/runtime/objectMonitor.cpp
+++ b/src/hotspot/share/runtime/objectMonitor.cpp
@@ -600,6 +600,22 @@ bool ObjectMonitor::deflate_monitor() {
   return true;  // Success, ObjectMonitor has been deflated.
 }
 
+// We might access the dead object headers for parsable heap walk, make sure
+// headers are in correct shape, e.g. monitors deflated.
+void ObjectMonitor::maybe_deflate_dead(oop* p) {
+  oop obj = *p;
+  assert(obj != NULL, "must not yet been cleared");
+  markWord mark = obj->mark();
+  if (mark.has_monitor()) {
+    ObjectMonitor* monitor = mark.monitor();
+    if (p == monitor->_object.ptr_raw()) {
+      assert(monitor->object_peek() == obj, "lock object must match");
+      markWord dmw = monitor->header();
+      obj->set_mark(dmw);
+    }
+  }
+}
+
 // Install the displaced mark word (dmw) of a deflating ObjectMonitor
 // into the header of the object associated with the monitor. This
 // idempotent method is called by a thread that is deflating a
--- a/src/hotspot/share/runtime/objectMonitor.hpp
+++ b/src/hotspot/share/runtime/objectMonitor.hpp
@@ -322,7 +322,10 @@ class ObjectMonitor : public CHeapObj<mtObjectMonitor> {
   // Use the following at your own risk
   intx      complete_exit(JavaThread* current);
 
- private:
+  // Deflate monitor in dead object if necessary.
+  static void maybe_deflate_dead(oop* p);
+
+private:
   void      AddWaiter(ObjectWaiter* waiter);
   void      INotify(JavaThread* current);
   ObjectWaiter* DequeueWaiter();


coleenp (approved):
This looks good to me, and I tested tier1-4 and ran some performance tests on this change also.

dholmes-ora:
Sorry I don't understand: how can we deflate and delete the monitor, yet not update the object header?

rkennke:
> Sorry I don't understand: how can we deflate and delete the monitor, yet not update the object header?

In OM::deflate_monitor() we check object_peek(), and if that returns null, then the object header is not updated (and can't be, because the object cannot be reached anymore). Concurrent GCs that process weak handles concurrently ensure that the object doesn't leak out by returning null there. However, for runtime code, at this point, there is no safe way to grab the object and update the header, because the GC might already have reclaimed it. The last safe point in time where we can do that is in WeakProcessor::Task::work() and OopStorage::weak_oops_do() itself, as soon as we detect that the object is properly unreachable.

dholmes-ora:
So how/when is the monitor actually deleted in such a case?

rkennke:
> So how/when is the monitor actually deleted in such a case?

ObjectSynchronizer::deflate_monitor_list() calls OM::deflate_monitor(). When the OM's object is unreachable, the update of the header is skipped, and the monitor owner set to DEFLATER_MARKER (otherwise the object header will *also* be updated to the original mark). When the concurrent deflation gets to actually deflating phase (after a handshake) it will pick up the OM from the OM list and see that it's marked for deflation and deallocate it.

dcubed-ojdk:
> how can we deflate and delete the monitor, yet not update the object header?

This is the code in `ObjectMonitor::deflate_monitor()` that detects the object has died:

```
  if (obj == nullptr) {
    // If the object died, we can recycle the monitor without racing with
    // Java threads. The GC already broke the association with the object.
    set_owner_from(nullptr, DEFLATER_MARKER);
    assert(contentions() >= 0, "must be non-negative: contentions=%d", contentions());
    _contentions = INT_MIN; // minimum negative int
  } else {
```

It used to be that ObjectMonitors were considered strong roots so the object
could not die while an ObjectMonitor was still holding its oop reference. We
changed that a release or two back and ObjectMonitor now holds a weak ref
to the object. That required addition of the above block to handle the case
when we wanted to deflate the ObjectMonitor and discovered that the object
had "left the building".

Of course, because the object is now gone, the `deflate_monitor()` code
can't fix the header and that didn't used to be a problem. Well actually it's
not really a problem in the current mainline, but will be with Lilliput.

dcubed-ojdk (approved):
Thumbs up.

stefank (changes requested):
Is there another way to solve this issue? It looks really wrong to put object monitor cleaning code into the generic OopStorage / WeakProcessor code. Could this be pushed out to the calling GC code instead?

kimbarrett:
I agree with Stefan that it seems wrong to be putting object monitor cleaning code into the generic OopStorage and
WeakProcessor code.

> Of course, because the object is now gone, the `deflate_monitor()` code can't fix the header and that didn't used to be a problem. Well actually it's not really a problem in the current mainline, but will be with Lilliput.

Why do we need to fix the header of a dead object?  It's dead.  Who cares what's in the header?  Nobody should
be touching dead objects.  Yes, I know there is heap walking stuff that does that, which is arguably a bug.  If its a
choice between fixing that or doing something like this, well, I'd really like to not do this.

kimbarrett:
In particular, during concurrent refinement, we're looking at either parsable
or unparsable parts of a region when processig a card. In the unparsable part,
we don't look at the dead objects for size information to find object
boundaries. Instead we use the mark bits to find live objects, ignoring dead
objects completely. In the parsable part, the dead objects have been
overwritten with filler objects that it is safe to examine. The dead objects
are replaced by fillers concurrently, moving the parsable boundry along the
way. (The replacement by filler objects is an optimization to make card
scanning faster, since no bitmap searching is required to step over one.) At
least, that's all how I think it's supposed to work; it's been a while since I
delved deeply into that code and it's changed somewhat since I last did so.

So I want to better understand the failures being reported, because what's
being described doesn't seem like it should happen and may indicate a bug
elsewhere.

rkennke:
> I agree with Stefan that it seems wrong to be putting object monitor cleaning code into the generic OopStorage and
> 
> WeakProcessor code.
> 
> 
> 
> > Of course, because the object is now gone, the `deflate_monitor()` code can't fix the header and that didn't used to be a problem. Well actually it's not really a problem in the current mainline, but will be with Lilliput.
> 
> 
> 
> Why do we need to fix the header of a dead object?  It's dead.  Who cares what's in the header?  Nobody should
> 
> be touching dead objects.  Yes, I know there is heap walking stuff that does that, which is arguably a bug.  If its a
> 
> choice between fixing that or doing something like this, well, I'd really like to not do this.
> 
> 

Hmm ok, what choice do we have? In Lilliput we synchronize GC threads with the monitor deflation handshake (which I am also going to upstream ASAP). Given that it is the G1 refinement thread that heap-walks dead objects with monitors, I guess it may be possible to let the refinement thread join the STS around heap walking and thus block deletion of monitors until it's done. I'm not totally sure if that'd work or even if that is feasible (how much time does G1 refinement spend in heap walking? Would it block monitor deflation for too long? What if monitors got deflated right before we start walking? Etc)

tschatzl:
G1 should never scan headers of dead objects, and if so, this is a bug somewhere else. As @kimbarrett mentioned, it is better to find and fix the bug and not paper over it.

https://tschatzl.github.io/2022/08/04/concurrent-marking.html describes the interaction between refinement and class unloading somewhat.

  1) Initially all objects are live and can be walked
  2) G1 decides to unload some classes. From this point on, for all old regions, below TAMS (copied to something we call "parsable bottom", i.e. PB), refinement code exclusively uses the bitmap from the marking for identifying live objects in that area. Objects above PB are all live, and so are parsable always.
  3) Marking threads now scrub the area between a region's bottom and PB, i.e. they put filler objects spanning all dead objects. This makes the heap parsable again.
  4) After finishing that scrubbing for a region, PB is set to bottom. Refinement can walk the heap region completely again - dead objects have been replaced with filler objects.

So in theory refinement threads should never see any dead object. The only issue I can think of may be problems with synchronizing reset of PB with refinement threads that might cause them reading dead objects (i.e. writing of dead objects not completed while PB has been reset).

The matter is different if somebody else keeps references to dead objects and tries to modify headers/contents afterwards; one cause could be that monitor handling modifies headers of dead objects, i.e. what has been replaced by filler objects. Then if that overwrite replaces a filler object header, refinement would obviously happily try to parse that object.

Do you have more information what GC is doing when these errors happen? If between Remark and Cleanup pause, then there might be a problem with that synchronization. If outside, then it's much more likely some change modifying dead areas of the heap (which seems to be an obvious no-can-do to me). In any case please provide more information about these crashes.

Also I do not understand why this would be "only a bug with compact object headers" - can you elaborate? All of the causes for refinement seeing a dead object seem to be quite independent of compact objects to my understanding.

rkennke:
> G1 should never scan headers of dead objects, and if so, this is a bug somewhere else. As @kimbarrett mentioned, it is better to find and fix the bug and not paper over it.
> 
> https://tschatzl.github.io/2022/08/04/concurrent-marking.html describes the interaction between refinement and class unloading somewhat.
> 
> 1. Initially all objects are live and can be walked
> 2. G1 decides to unload some classes. From this point on, for all old regions, below TAMS (copied to something we call "parsable bottom", i.e. PB), refinement code exclusively uses the bitmap from the marking for identifying live objects in that area. Objects above PB are all live, and so are parsable always.
> 3. Marking threads now scrub the area between a region's bottom and PB, i.e. they put filler objects spanning all dead objects. This makes the heap parsable again.
> 4. After finishing that scrubbing for a region, PB is set to bottom. Refinement can walk the heap region completely again - dead objects have been replaced with filler objects.
> 
> So in theory refinement threads should never see any dead object. The only issue I can think of may be problems with synchronizing reset of PB with refinement threads that might cause them reading dead objects (i.e. writing of dead objects not completed while PB has been reset).
> 
> The matter is different if somebody else keeps references to dead objects and tries to modify headers/contents afterwards; one cause could be that monitor handling modifies headers of dead objects, i.e. what has been replaced by filler objects. Then if that overwrite replaces a filler object header, refinement would obviously happily try to parse that object.
> 
> Do you have more information what GC is doing when these errors happen? If between Remark and Cleanup pause, then there might be a problem with that synchronization. If outside, then it's much more likely some change modifying dead areas of the heap (which seems to be an obvious no-can-do to me). In any case please provide more information about these crashes.
> 
> Also I do not understand why this would be "only a bug with compact object headers" - can you elaborate? All of the causes for refinement seeing a dead object seem to be quite independent of compact objects to my understanding.

Thanks for the explanations, Thomas!
I am now trying to reproduce the original issue that we've encountered back when we did the original fix (https://github.com/openjdk/lilliput/pull/28). That has been a while (Nov 2021) and so far my testing hasn't shown the bug. It is well possible that some upstream JDK/G1 changes or changes in Lilliput since then have fixed the real underlying issue. I'll put this PR on hold until I am able to reproduce the issue, and would withdraw it if I can't.

Again, thanks!
Roman

kimbarrett:
> > Why do we need to fix the header of a dead object?  It's dead.  Who cares what's in the header?  Nobody should
> > be touching dead objects.  Yes, I know there is heap walking stuff that does that, which is arguably a bug.  If its a
> > choice between fixing that or doing something like this, well, I'd really like to not do this.
> 
> Hmm ok, what choice do we have? In Lilliput we synchronize GC threads with the monitor deflation handshake (which I am also going to upstream ASAP). Given that it is the G1 refinement thread that heap-walks dead objects with monitors, I guess it may be possible to let the refinement thread join the STS around heap walking and thus block deletion of monitors until it's done. I'm not totally sure if that'd work or even if that is feasible (how much time does G1 refinement spend in heap walking? Would it block monitor deflation for too long? What if monitors got deflated right before we start walking? Etc)

The "heap walk" I was referring to was things like the heap dumper and such.
As already explained, concurrent refinement shouldn't be looking at dead objects.

The concurrent refinement threads already do their work within STS, leaving it
to yield when requested, or to wait for work. But relying on STS to block
concurrent refinement doesn't work, since JavaThreads may also do refinement
work, and they of course don't use STS.

rkennke:
> > > Why do we need to fix the header of a dead object?  It's dead.  Who cares what's in the header?  Nobody should
> > > be touching dead objects.  Yes, I know there is heap walking stuff that does that, which is arguably a bug.  If its a
> > > choice between fixing that or doing something like this, well, I'd really like to not do this.
> > 
> > 
> > Hmm ok, what choice do we have? In Lilliput we synchronize GC threads with the monitor deflation handshake (which I am also going to upstream ASAP). Given that it is the G1 refinement thread that heap-walks dead objects with monitors, I guess it may be possible to let the refinement thread join the STS around heap walking and thus block deletion of monitors until it's done. I'm not totally sure if that'd work or even if that is feasible (how much time does G1 refinement spend in heap walking? Would it block monitor deflation for too long? What if monitors got deflated right before we start walking? Etc)
> 
> The "heap walk" I was referring to was things like the heap dumper and such. As already explained, concurrent refinement shouldn't be looking at dead objects.
> 
> The concurrent refinement threads already do their work within STS, leaving it to yield when requested, or to wait for work. But relying on STS to block concurrent refinement doesn't work, since JavaThreads may also do refinement work, and they of course don't use STS.

Ok. Heap dump and such is not affected by this 'bug' AFAIK. The idea is not to use STS to block refinement, but to ensure that GC threads can safely reach through monitors in headers, by synchronizing with the monitor deflation protocol (and basically blocking deflation until all threads can handshake). See https://github.com/openjdk/lilliput/pull/27 . Java threads already sync with the deflation thread in this way, and GC threads need to do the same with Lilliput. I'll post that change for review soon.

rkennke:
The change does not seem to be necessary anymore.

PR #13732, opened 2023-04-28
8307132: Cleanup the code of sun.java2d.cmm.lcms package

mrserb:
The description from big to small:

- Our code to handle the image types we support is too generic, the LCMSImageLayout class can handle, byte, short, int, and double types, and many various properties of the image layout. As a result, we pass a couple of good parameters to the LCMSImageLayout and have to use safeXX methods in it to calculate the final layout, and then validate it.
  This patch moves the layout properties calculation to one place - the constructor of LCMSImageLayout, and from the outside of the class, we now pass only the data array and the number of components per pixel:
 - The usage of Double type is removed, we do not use that type currently, and do not plan to support it in the future. Note that we support the float type, and I tried to implement it, but unfortunately, it is [intentionally ](https://github.com/mm2/Little-CMS/issues/356)slow. So will continue to use short type instead of float.
 - Discussed a few times the  `do { } while (false);` block is removed.

---------
### Progress

### Issue
 * [JDK-8307132](https://bugs.openjdk.org/browse/JDK-8307132): Cleanup the code of sun.java2d.cmm.lcms package

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13732/head:pull/13732` \
`$ git checkout pull/13732`

`$ git checkout pull/13732` \
`$ git pull https://git.openjdk.org/jdk.git pull/13732/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13732`

`$ git pr show -t 13732`

</details>

<a href="https://git.openjdk.org/jdk/pull/13732.diff">https://git.openjdk.org/jdk/pull/13732.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13732#issuecomment-1530773365)


src/java.desktop/share/native/liblcms/LCMS.c
```diff
@@ -45,10 +46,9 @@
 
 #define SigHead TagIdConst('h','e','a','d')
 
-#define DT_BYTE     0
-#define DT_SHORT    1
-#define DT_INT      2
-#define DT_DOUBLE   3
+#define DT_BYTE     sun_java2d_cmm_lcms_LCMSImageLayout_DT_BYTE
+#define DT_SHORT    sun_java2d_cmm_lcms_LCMSImageLayout_DT_SHORT
+#define DT_INT      sun_java2d_cmm_lcms_LCMSImageLayout_DT_INT
```
prrace, line 51: I don't understand why this isn't a step backwards.
It is overloading the "size" to mean the "type"
How would you tell the difference between SHORT vs UNSIGNED SHORT - both 2 bytes,
or INT vs FLOAT - both 4 bytes ?
or LONG vs DOUBLE ..
If we ever need to, then this all needs to be reverted.
% cd java/lang 
%  grep BYTES *java 
Byte.java:    public static final int BYTES = SIZE / Byte.SIZE;
Character.java:    public static final int BYTES = SIZE / Byte.SIZE;
Double.java:    public static final int BYTES = SIZE / Byte.SIZE;
Float.java:    public static final int BYTES = SIZE / Byte.SIZE;
Integer.java:    public static final int BYTES = SIZE / Byte.SIZE;
Long.java:    public static final int BYTES = SIZE / Byte.SIZE;
Short.java:    public static final int BYTES = SIZE / Byte.SIZE;

mrserb:
@prrace The PR is updated, and the tests are green. Please take a look.

prrace (approved):

PR #13734, opened 2023-04-29
8307145: windowsaccessbridge.dll erroneously includes private methods in its C API

TheShermanTanker:
In windowsaccessbridge(-64).dll the following methods are exported in a def file to C API which in turn call their real implementation, which are instance methods of the class WinAccessBridge:
```
addJavaEventNotification
removeJavaEventNotification
addAccessibilityEventNotification
removeAccessibilityEventNotification
```

However, they are nowhere to be seen in the actual C interface, in AccessBridgeWindowsEntryPoints.cpp. Your guess is as good as mine as to how on earth MSVC is still capable of compiling and linking this without any errors whatsoever, but in any case, this is a severe oversight and should be properly defined in the C API lest this happy accident within MSVC is fixed by Microsoft in the future

---------
### Progress

### Issue
 * [JDK-8307145](https://bugs.openjdk.org/browse/JDK-8307145): windowsaccessbridge.dll erroneously includes private methods in its C API (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13734/head:pull/13734` \
`$ git checkout pull/13734`

`$ git checkout pull/13734` \
`$ git pull https://git.openjdk.org/jdk.git pull/13734/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13734`

`$ git pr show -t 13734`

</details>

<a href="https://git.openjdk.org/jdk/pull/13734.diff">https://git.openjdk.org/jdk/pull/13734.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13734#issuecomment-1528859450)

### DIFF
--- a/src/jdk.accessibility/windows/native/libwindowsaccessbridge/AccessBridgeWindowsEntryPoints.cpp
+++ b/src/jdk.accessibility/windows/native/libwindowsaccessbridge/AccessBridgeWindowsEntryPoints.cpp
@@ -43,6 +43,30 @@ extern HWND theDialogWindow;
 extern "C" {
 #endif
 
+    void addJavaEventNotification(jlong type) {
+        if (theWindowsAccessBridge != nullptr) {
+            theWindowsAccessBridge->addJavaEventNotification(type);
+        }
+    }
+
+    void removeJavaEventNotification(jlong type) {
+        if (theWindowsAccessBridge != nullptr) {
+            theWindowsAccessBridge->removeJavaEventNotification(type);
+        }
+    }
+
+    void addAccessibilityEventNotification(jlong type) {
+        if (theWindowsAccessBridge != nullptr) {
+            theWindowsAccessBridge->addAccessibilityEventNotification(type);
+        }
+    }
+
+    void removeAccessibilityEventNotification(jlong type) {
+        if (theWindowsAccessBridge != nullptr) {
+            theWindowsAccessBridge->removeAccessibilityEventNotification(type);
+        }
+    }
+
     /**
      * Windows_run - where Windows executables will load/unload us
      *
--- a/src/jdk.accessibility/windows/native/libwindowsaccessbridge/AccessBridgeWindowsEntryPoints.cpp
+++ b/src/jdk.accessibility/windows/native/libwindowsaccessbridge/AccessBridgeWindowsEntryPoints.cpp
@@ -43,30 +43,6 @@ extern HWND theDialogWindow;
 extern "C" {
 #endif
 
-    void addJavaEventNotification(jlong type) {
-        if (theWindowsAccessBridge != nullptr) {
-            theWindowsAccessBridge->addJavaEventNotification(type);
-        }
-    }
-
-    void removeJavaEventNotification(jlong type) {
-        if (theWindowsAccessBridge != nullptr) {
-            theWindowsAccessBridge->removeJavaEventNotification(type);
-        }
-    }
-
-    void addAccessibilityEventNotification(jlong type) {
-        if (theWindowsAccessBridge != nullptr) {
-            theWindowsAccessBridge->addAccessibilityEventNotification(type);
-        }
-    }
-
-    void removeAccessibilityEventNotification(jlong type) {
-        if (theWindowsAccessBridge != nullptr) {
-            theWindowsAccessBridge->removeAccessibilityEventNotification(type);
-        }
-    }
-
     /**
      * Windows_run - where Windows executables will load/unload us
      *
--- a/src/jdk.accessibility/windows/native/libwindowsaccessbridge/AccessBridgeWindowsEntryPoints.h
+++ b/src/jdk.accessibility/windows/native/libwindowsaccessbridge/AccessBridgeWindowsEntryPoints.h
@@ -40,6 +40,11 @@
 extern "C" {
 #endif
 
+    void addJavaEventNotification(jlong type);
+    void removeJavaEventNotification(jlong type);
+    void addAccessibilityEventNotification(jlong type);
+    void removeAccessibilityEventNotification(jlong type);
+
     void Windows_run();
 
     void releaseJavaObject(long vmID, JOBJECT64 object);
--- a/src/jdk.accessibility/windows/native/libwindowsaccessbridge/AccessBridgeWindowsEntryPoints.h
+++ b/src/jdk.accessibility/windows/native/libwindowsaccessbridge/AccessBridgeWindowsEntryPoints.h
@@ -40,11 +40,6 @@
 extern "C" {
 #endif
 
-    void addJavaEventNotification(jlong type);
-    void removeJavaEventNotification(jlong type);
-    void addAccessibilityEventNotification(jlong type);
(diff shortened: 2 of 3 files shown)


NikitkoCent:
> These don't seem to be part of the public API, so this is probably deliberate and not a bug.
> https://docs.oracle.com/javase/accessbridge/2.0.2/api.htm#jab-api-specification

@prrace should non-public API methods be removed from .DEF then?

TheShermanTanker:
I should think so, it doesn't make sense to export a function that doesn't even exist at all. And to make matters worse Microsoft Visual C still does export functions with those names (you can easily verify this with dumpbin) which is even more confusing. It may be a cursed Microsoft extension that has simply gone under the radar, and is somehow exporting the C++ class instance methods as static, or even exporting dummy functions, neither of which is what we want it to do!

prrace:
So perhaps something internal in the various accessbridge dll's needs it. How will you test this ?
Who here has JAWS ? If you don't have JAWS and therefore can't use it to verify the A11Y test suite then
I don't think you can test this change in a way that makes me comfortable in doing anything here.

savoptik:
@prrace Maybe try using free NVDA for testing?

TheShermanTanker:
The problem as I see it is that the .def is exporting something that simply does not exist at all, even if it is a non-public API and the symbols are simply exported from the DLL for internal use by other accessbridge code, then these methods should still be placed in the extern "C" block in the .cpp (but maybe not in the corresponding header since they aren't public). Conversely if it is never meant to be exported but the accessbridge DLLs somehow need the .def files to export phantom, nonexistent symbols to work properly, then something is seriously wrong with them and we should probably take a look at what on earth is going on there

djelinski:
FWIW, this is what dumpbin displays for these entries:
```
          1    2 0000351C addAccessibilityEventNotification = ?addAccessibilityEventNotification@WinAccessBridge@@QEAAX_J@Z (public: void __cdecl WinAccessBridge::addAccessibilityEventNotification(__int64))
          2    4 00003628 addJavaEventNotification = ?addJavaEventNotification@WinAccessBridge@@QEAAX_J@Z (public: void __cdecl WinAccessBridge::addJavaEventNotification(__int64))
          3   42 00006D94 removeAccessibilityEventNotification = ?removeAccessibilityEventNotification@WinAccessBridge@@QEAAX_J@Z (public: void __cdecl WinAccessBridge::removeAccessibilityEventNotification(__int64))
          4   44 00006EA0 removeJavaEventNotification = ?removeJavaEventNotification@WinAccessBridge@@QEAAX_J@Z (public: void __cdecl WinAccessBridge::removeJavaEventNotification(__int64))
```
There's no way to get a pointer to an instance of `WinAccessBridge` using the supported interface, and calling these methods with garbage value for `this` would likely crash.

prrace:
> > FWIW, this is what dumpbin displays for these entries:
> > ```
> >           1    2 0000351C addAccessibilityEventNotification = ?addAccessibilityEventNotification@WinAccessBridge@@QEAAX_J@Z (public: void __cdecl WinAccessBridge::addAccessibilityEventNotification(__int64))
> >           2    4 00003628 addJavaEventNotification = ?addJavaEventNotification@WinAccessBridge@@QEAAX_J@Z (public: void __cdecl WinAccessBridge::addJavaEventNotification(__int64))
> >           3   42 00006D94 removeAccessibilityEventNotification = ?removeAccessibilityEventNotification@WinAccessBridge@@QEAAX_J@Z (public: void __cdecl WinAccessBridge::removeAccessibilityEventNotification(__int64))
> >           4   44 00006EA0 removeJavaEventNotification = ?removeJavaEventNotification@WinAccessBridge@@QEAAX_J@Z (public: void __cdecl WinAccessBridge::removeJavaEventNotification(__int64))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > There's no way to get a pointer to an instance of `WinAccessBridge` using the supported interface, and calling these methods with garbage value for `this` would likely crash.
> 
> Dear lord, the Microsoft linker is truly cursed, through and through
> 
> @prrace What sort of testing would need to be done for such a change?

@azuev-java and @kumarabhi006 can help with review and testing of this PR.

TheShermanTanker:
Bumping

TheShermanTanker:
Anyone?

aivanov-jdk:
I'll take a look… hopefully next week.

TheShermanTanker:
Bumping

aivanov-jdk:
> > These don't seem to be part of the public API, so this is probably deliberate and not a bug.
> > https://docs.oracle.com/javase/accessbridge/2.0.2/api.htm#jab-api-specification
> 
> @prrace should non-public API methods be removed from .DEF then?

I think it's the best solution. These four functions are not part of the public API to AccessBridge, they're missing from [`AccessBridgeCalls.c`](https://github.com/openjdk/jdk/blob/master/src/jdk.accessibility/windows/native/bridge/AccessBridgeCalls.c) and [`AccessBridgeCalls.h`](https://github.com/openjdk/jdk/blob/master/src/jdk.accessibility/windows/native/include/bridge/AccessBridgeCalls.h).

They should not be exported. If the functions aren't loaded by its name, which is implemented in `AccessBridgeCalls.c`, or otherwise linked to from other modules (`.exe` or `.dll`), they don't need to be exported. These functions are part of the `WinAccessBridge` class, none of them cannot be called directly.

Did you try removing them from the `.def` file? Does anything break?

TheShermanTanker:
I don't know about the more extensive test suite Phil alludes to, though

TheShermanTanker:
Bumping

aivanov-jdk:
> I don't know about the more extensive test suite Phil alludes to, though

Neither do I know about more extensive testing. I presume these four entry points were not used externally. Yet _“compiles”_ may not be enough. What does your _“works”_ mean? Did you run it with an AccessBridge client?

TheShermanTanker:
@aivanov-jdk The client does indeed function on my end without issue, but I don't know whether it will do so on other devices as well

TheShermanTanker:
JTReg sure does have a sense of humour, immediately crashing and leaving that ugly cross on the tests the moment I say that

djelinski (approved):
LGTM. These methods are not usable outside of the windowsaccessbridge dll. Caller would need a working instance of WinAccessBridge class to pass in `this` parameter, and there's no way to get one outside of the DLL.

TheShermanTanker:
Right, thanks Daniel

TheShermanTanker:
I will change the title of the PR first

TheShermanTanker:
Thanks all for reviewing this change!

aivanov-jdk:
> @aivanov-jdk The client does indeed function on my end without issue, but I don't know whether it will do so on other devices as well

That's good to know.

I would've approved it, if I'd had a slightest chance to read your reply.

TheShermanTanker:
Ah, sorry about that. I've had several changes starved of reviews recently and lying around for almost a month, and was feeling fairly frustrated and impatient :(

PR #13735, opened 2023-04-30
8156916: intrinsify MethodHandles.arrayConstructor()

liach:
Try to intrinsify MethodHandles.arrayConstructor for java.lang, java.lang.invoke, and java.util types, which InvokerBytecodeGenerator can refer to directly in a class constant.

In fact, the performance difference of Array.newInstance and anewarray is quite negligible per benchmark:
```
Benchmark                                                            Mode  Cnt     Score     Error  Units
MethodHandlesArrayConstructorCall.baselineConstruct                  avgt    5     3.467 ±   0.163  ns/op
MethodHandlesArrayConstructorCall.constantFoldConstruct              avgt    5     3.403 ±   0.070  ns/op
MethodHandlesArrayConstructorCall.constantFoldNonInvocableConstruct  avgt    5     3.445 ±   0.231  ns/op
MethodHandlesArrayConstructorCall.instanceConstruct                  avgt    5     4.034 ±   0.168  ns/op
MethodHandlesArrayConstructorCall.instanceNonInvocableConstruct      avgt    5     4.038 ±   0.104  ns/op
MethodHandlesArrayConstructorColdStart.testCreateInvocable             ss   10  1425.991 ± 308.216  us/op
MethodHandlesArrayConstructorColdStart.testCreateNonInvocable          ss   10  1401.480 ± 258.402  us/op
MethodHandlesArrayConstructorColdStart.testCreateObject                ss   10  1213.018 ± 166.522  us/op
MethodHandlesArrayConstructorColdStart.testCreatePrimitive             ss   10  1163.461 ± 178.549  us/op
```

---------
### Progress

### Issue
 * [JDK-8156916](https://bugs.openjdk.org/browse/JDK-8156916): intrinsify MethodHandles.arrayConstructor()

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13735/head:pull/13735` \
`$ git checkout pull/13735`

`$ git checkout pull/13735` \
`$ git pull https://git.openjdk.org/jdk.git pull/13735/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13735`

`$ git pr show -t 13735`

</details>

<a href="https://git.openjdk.org/jdk/pull/13735.diff">https://git.openjdk.org/jdk/pull/13735.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13735#issuecomment-1528951733)

### DIFF
--- a/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
+++ b/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
@@ -4326,12 +4326,7 @@ return mh1;
      * @since 9
      */
     public static MethodHandle arrayConstructor(Class<?> arrayClass) throws IllegalArgumentException {
-        if (!arrayClass.isArray()) {
-            throw newIllegalArgumentException("not an array class: " + arrayClass.getName());
-        }
-        MethodHandle ani = MethodHandleImpl.getConstantHandle(MethodHandleImpl.MH_Array_newInstance).
-                bindTo(arrayClass.getComponentType());
-        return ani.asType(ani.type().changeReturnType(arrayClass));
+        return MethodHandleImpl.makeArrayElementAccessor(arrayClass, MethodHandleImpl.ArrayAccess.CONSTRUCT);
     }
 
     /**
--- a/src/java.base/share/classes/java/lang/invoke/InvokerBytecodeGenerator.java
+++ b/src/java.base/share/classes/java/lang/invoke/InvokerBytecodeGenerator.java
@@ -841,6 +841,9 @@ class InvokerBytecodeGenerator {
                 case ARRAY_LENGTH:
                     emitArrayLength(name);
                     continue;
+                case ARRAY_CONSTRUCTOR:
+                    emitArrayConstructor(name);
+                    continue;
                 case IDENTITY:
                     assert(name.arguments.length == 1);
                     emitPushArguments(name, 0);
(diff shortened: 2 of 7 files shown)


JornVernee:
`Array::newInstance` is already intrinsified. It behaves pretty much the same as `anewarray` if the array type is constant ([1], [2]). So, I'm not sure I see how intrinsifying this could improve performance.

What kind of performance improvement are you seeing? Could you add a comparison with a method handle that calls `Array::newInstance` bound to the array type? (that should have the same performance as the current implementation).

[1]: https://github.com/openjdk/jdk/blob/b3dbf28bc0614bee2f7137af95389134155c9511/src/hotspot/share/opto/library_call.cpp#L4051
[2]: https://github.com/openjdk/jdk/blob/b3dbf28bc0614bee2f7137af95389134155c9511/src/hotspot/share/opto/parse3.cpp#L282

src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -4326,12 +4326,7 @@ private MethodHandle getDirectMethodForConstant(byte refKind, Class<?> defc, Mem
      * @since 9
      */
     public static MethodHandle arrayConstructor(Class<?> arrayClass) throws IllegalArgumentException {
-        if (!arrayClass.isArray()) {
-            throw newIllegalArgumentException("not an array class: " + arrayClass.getName());
-        }
-        MethodHandle ani = MethodHandleImpl.getConstantHandle(MethodHandleImpl.MH_Array_newInstance).
-                bindTo(arrayClass.getComponentType());
-        return ani.asType(ani.type().changeReturnType(arrayClass));
```
liach, line 4334: Since `Array.newInstance` is already intrinsified, this patch should probably dropped and the issue be no longer valid. However, one nit I have is that we can convert this `asType` to `viewAsType(..., false)` since `Array.newInstance`'s return type is trusted.

liach:
> What kind of performance improvement are you seeing? Could you add a comparison with a method handle that calls `Array::newInstance` bound to the array type? (that should have the same performance as the current implementation).

That explains why `NonInvocable` (delegates to Array.newInstance) isn't slower than `anewarray` opcode (`Invocable`) in the benchmark results above. I will drop this patch and close the related JBS issue as no longer applicable.

PR #13738, opened 2023-04-30
8305919: java/lang/Thread/virtual/HoldsLock.java#id0 failed, ThreadInfo.getLockInfo() return null

AlanBateman:
HoldsLock.java#id0 has been failing intermittently recently due to threads left over from previous tests terminating. HoldsLock.java#id1 doesn't fail as it runs in /othervm mode. The test uses ThreadMXBean::getAllThreadIds to get the ID of all threads and calls ThreadMXBean::getThreadInfo on each thread. If a thread from a previous test terminates then getThreadInfo returns null and the test fails.

The test can be trivially fixed to check for null but looking at it afresh, the test can be simplified to use existing test infrastructure and to just call ThreadMXBean::getThreadInfo on the carrier. The test can also be renamed to make it clearer that it is testing a carrier threads wait for a virtual thread. The old test includes a disabled test for Thread.holdsLock but there is further VM work required before that is useful and it would be better to develop new tests at part of that work.

---------
### Progress

### Issue
 * [JDK-8305919](https://bugs.openjdk.org/browse/JDK-8305919): java/lang/Thread/virtual/HoldsLock.java#id0 failed, ThreadInfo.getLockInfo() return null

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13738/head:pull/13738` \
`$ git checkout pull/13738`

`$ git checkout pull/13738` \
`$ git pull https://git.openjdk.org/jdk.git pull/13738/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13738`

`$ git pr show -t 13738`

</details>

<a href="https://git.openjdk.org/jdk/pull/13738.diff">https://git.openjdk.org/jdk/pull/13738.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13738#issuecomment-1529111551)


jaikiran (approved):
Hello Alan, these change look good to me.

> The test uses ThreadMXBean::getAllThreadIds to get the ID of all threads and calls ThreadMXBean::getThreadInfo on each thread. If a thread from a previous test terminates then getThreadInfo returns null and the test fails.

The test was problem listed only in `-Xcomp` but it wasn't clear why this issue would affect only `-Xcomp`, but then reading through the JBS issue, I can see that this failure was reproduced even without `-Xcomp` after the problem listing was done.

AlanBateman:
> The test was problem listed only in `-Xcomp` but it wasn't clear why this issue would affect only `-Xcomp`, but then reading through the JBS issue, I can see that this failure was reproduced even without `-Xcomp` after the problem listing was done.

Right, not specific to -Xcomp, it's more about the test runs in agentvm mode so it's possible to get interference from tests that ran previously in the same VM.

PR #13740, opened 2023-05-01
8307163: JLONG_FORMAT_SPECIFIER should be updated on Windows

TheShermanTanker:
Windows no longer uses I64d anywhere in their newer compilers, instead using the conforming lld specifiers. Minor cleanup here in JLI code to reflect that

---------
### Progress

### Issue
 * [JDK-8307163](https://bugs.openjdk.org/browse/JDK-8307163): JLONG_FORMAT_SPECIFIER should be updated on Windows

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13740/head:pull/13740` \
`$ git checkout pull/13740`

`$ git checkout pull/13740` \
`$ git pull https://git.openjdk.org/jdk.git pull/13740/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13740`

`$ git pr show -t 13740`

</details>

<a href="https://git.openjdk.org/jdk/pull/13740.diff">https://git.openjdk.org/jdk/pull/13740.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13740#issuecomment-1529938108)

### DIFF
--- a/src/hotspot/share/utilities/globalDefinitions_visCPP.hpp
+++ b/src/hotspot/share/utilities/globalDefinitions_visCPP.hpp
@@ -102,7 +102,7 @@ inline int g_isfinite(jfloat  f)                 { return _finite(f); }
 inline int g_isfinite(jdouble f)                 { return _finite(f); }
 
 // Formatting.
-#define FORMAT64_MODIFIER "I64"
+#define FORMAT64_MODIFIER "ll"
 
 #define offset_of(klass,field) offsetof(klass,field)
 
--- a/src/java.base/windows/native/libjli/java_md.h
+++ b/src/java.base/windows/native/libjli/java_md.h
@@ -39,7 +39,7 @@
 #define MAXPATHLEN      MAX_PATH
 #define MAXNAMELEN      MAX_PATH
 
-#define JLONG_FORMAT_SPECIFIER "%I64d"
+#define JLONG_FORMAT_SPECIFIER "%lld"
 
 /*
  * Function prototypes.


dholmes-ora (commented):
The VM still uses PRId64.

TheShermanTanker:
Is the globalDefinitions declaration what you're referring to? In any case, the Microsoft Visual C compiler from version 2019 typedef's PRId64 as lld, and recommends using it over the older I64d specifier, so HotSpot should also have its declarations updated as well

src/hotspot/share/utilities/globalDefinitions_visCPP.hpp
```diff
@@ -102,7 +102,7 @@ inline int g_isfinite(jfloat  f)                 { return _finite(f); }
 inline int g_isfinite(jdouble f)                 { return _finite(f); }
 
 // Formatting.
-#define FORMAT64_MODIFIER "I64"
+#define FORMAT64_MODIFIER "ll"
```
dholmes-ora, line 105: Interesting - this seems un-needed and should be replaced in its one use by `INT64_FORMAT_X_0` - but separate PR I think.

dholmes-ora:
> Is the globalDefinitions declaration what you're referring to? 

Yes. I'm not clear on the background to all these PRI* format modifiers - all seems rather convoluted.

TheShermanTanker:
> > Is the globalDefinitions declaration what you're referring to?
> 
> Yes. I'm not clear on the background to all these PRI* format modifiers - all seems rather convoluted.

Ah, I see. PRId64 (the 64 bit signed format specifier) used to be %I64d on Windows for a while, but Microsoft has long since replaced them with the proper %lld format specifier a while ago, and strongly encourages C and C++ code on Windows to do the same: https://learn.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=msvc-170

It's not a critical issue, but it still is better to replace the outdated formatting in native and HotSpot code which we define ourselves (as opposed to directly using PRId64) to use what Microsoft themselves have changed the specifier to

I've also just noticed that we should probably also change jlong from __int64 to long long in Windows specific JNI as well to go along with this change

TheShermanTanker:
Bumping

tstuefe (approved):
Good

TheShermanTanker:
Alright, thanks thomas :)

/integrate

dholmes-ora (commented):
Belated okay.

PR #13742, opened 2023-05-01
8290005: com/sun/jndi/ldap/LdapCBPropertiesTest.java failling with NullPointerException

mpdonova:
In this PR, I added methods to the TransportContext class to synchronize access to the handshakeContext field. I also updated locations in the code that rely on the handshakeContext field to not be null to use the synchronized methods.

Thanks

---------
### Progress

### Issue
 * [JDK-8290005](https://bugs.openjdk.org/browse/JDK-8290005): com/sun/jndi/ldap/LdapCBPropertiesTest.java failling with NullPointerException (**Bug** - P4)

### Reviewers
 * [Jamil Nimeh](https://openjdk.org/census#jnimeh) (@jnimeh - **Reviewer**) ⚠️ Review applies to [77e4d873](https://git.openjdk.org/jdk/pull/13742/files/77e4d873b866951cd9ef80018bc0d1414b3a4f02)
 * [Xue-Lei Andrew Fan](https://openjdk.org/census#xuelei) (@XueleiFan - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13742/head:pull/13742` \
`$ git checkout pull/13742`

`$ git checkout pull/13742` \
`$ git pull https://git.openjdk.org/jdk.git pull/13742/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13742`

`$ git pr show -t 13742`

</details>

<a href="https://git.openjdk.org/jdk/pull/13742.diff">https://git.openjdk.org/jdk/pull/13742.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13742#issuecomment-1530011818)


src/java.base/share/classes/sun/security/ssl/SSLEngineImpl.java
```diff
@@ -760,11 +760,12 @@ private Plaintext decode(
     @Override
     public Runnable getDelegatedTask() {
         engineLock.lock();
+        HandshakeContext context = conContext.handshakeContext;
         try {
-            if (conContext.handshakeContext != null && // PRE or POST handshake
+            if (context != null && // PRE or POST handshake
                     !conContext.handshakeContext.taskDelegated &&
                     !conContext.handshakeContext.delegatedActions.isEmpty()) {
```
dfuch, line 767: Shouldn't replace `conContext.handshakeContext` with `context` in those two lines as well?

src/java.base/share/classes/sun/security/ssl/TransportContext.java
```diff
@@ -458,6 +455,12 @@ SSLException fatal(Alert alert, String diagnostic,
         }
     }
 
+    private void terminateHandshakeContext() {
+        handshakeCtxLock.lock();
+        handshakeContext = null;
+        handshakeCtxLock.unlock();
```
dfuch, line 461: The usual pattern is to use try { } finally { } with locks (even though in this particular case I don't see how the assignment would throw) - but if more things need to be done in the future here then at least the try { } finally { } will be in place.

```suggestion
        handshakeCtxLock.lock();
        try {
            handshakeContext = null;
        } finally {
            handshakeCtxLock.unlock();
        }    
```

src/java.base/share/classes/sun/security/ssl/SSLEngineImpl.java
mpdonova: Yes, I definitely should have done that. Thanks.

src/java.base/share/classes/sun/security/ssl/TransportContext.java
mpdonova: I added the try/finally.

src/java.base/share/classes/sun/security/ssl/TransportContext.java
```diff
@@ -89,6 +88,7 @@ final class TransportContext implements ConnectionContext {
 
     // handshake context
     HandshakeContext                handshakeContext = null;
+    Lock                            handshakeCtxLock = new ReentrantLock();
```
dfuch, line 91: `handshakeCtxLock` should be declared final.

dfuch (commented):
LGTM. You will need approval from a security-dev Reviewer before pushing though.
In particular I am not sure about the implication of removing the lock for around SSLEngine::getHandshakeSession() and SSLSocket::getApplicationProtocols()

mpdonova:
Can a security-dev Reviewer take a look at this? Thanks!

mpdonova:
Hi, I'm still looking for a Reviewer for this PR. Thanks!

jnimeh (approved):
This looks good to me.

src/java.base/share/classes/sun/security/ssl/SSLEngineImpl.java
```diff
@@ -911,13 +912,7 @@ public SSLSession getSession() {
 
     @Override
     public SSLSession getHandshakeSession() {
-        engineLock.lock();
```
XueleiFan, line 914: I'm not very sure of this update. By removing the enginLock on socket/engine level here,  is it possible there are two threads that one read the handshake session and another on write?

mpdonova: `engineLock` was used to here to make sure that `conContext.handshakeContext` is not null before accessing it. The problem is that the `handshakeContext` is modified (set to null) by the `TransportContext` class. So using a lock in SSLEngineImpl or SSLSocketImpl doesn't protect `handshakeContext` from becoming null between the check and use.

The actual SSLSession object was never protected from concurrent modification (unless it has its own lock.)

When reviewing the code to answer this, I noticed that I needed to update `SSLEngineImpl.getHandshakeApplicationProtocol()` so there is a new commit.

XueleiFan: I had a search of the value assignment of handshakeSession.  Here is one example of the calling stack:
SSLEngineImpl.DelegatedTask.run()->TransportContext.dispatch()->SSLHandshake().consume()->HandshakeConsumer.consume()->SSLExtensions.product()->ServerHello.produce()->shc.handshakeSession = session

The engineLock was placed on SSLEngineImpl.DelegatedTask.run() and released after complete the job, which means the value assignment of handshakeSession is synchronized.

The locks in SSLEngineImpl and SSLSocketImpl are used for operations synchronization so that other classes may not need additional locks in most cases.

mpdonova: I see what you're saying with respect to SSLEngineImpl. It looks like all of the public methods are synchronized with `engineLock` so I shouldn't have made any changes there. I will revert them.

Are you ok with the changes in SSLSocketImpl?

XueleiFan: I may update both SSLSocketImpl and SSLEngineImpl, as SSLSocketImpl uses the locks similar to SSLEngineImpl.

XueleiFan: Yes, socket close is a headache problem for me.

jnimeh: There's a bit of a history with SSLSocket closures since the new handshaker was brought into JDK11.  Some of it dealt with synchronization, others with properly handling full vs. half-duplex closes.  You may want to look up some of the bug history on changes surrounding the SSLSocket closures just to make sure we don't undo any fixes that were previously integrated.  As Xuelei's comment indicates, this is a delicate part of the handshaker.

mpdonova: I reverted the changes to SSLEngineImpl and looked at the history of SSLSocketImpl and TransportContext but didn't see anything that I would be undoing with this change.

mpdonova: Sorry for the delay on any updates here. 

I updated this branch and verified the tests still pass. I ran jdk_security3 tests from test/jdk/TEST.groups. Is there anything else I should do to test this change?

mpdonova: The only lock I added was in `TransportContext` to synchronize access to the `handshakeContext` field, but I understand your reluctance to make any changes with locks. The problem is that SSLSocketImpl tries to access `conContext.handshakeContext.negotiatedProtocol` after TransportContext has set handshakeContext to null.

TransportContext also has a `protocolVersion` field. Is it possible to just use that instead?

XueleiFan: > TransportContext also has a `protocolVersion` field. Is it possible to just use that instead?

Did you mean a change in duplexCloseOutput() like the following?

```
-           // The protocol version may have been negotiated.
-           ProtocolVersion pv = conContext.handshakeContext.negotiatedProtocol;
+           // The protocol version may have been negotiated.  The 
+           // conContext.handshakeContext.negotiatedProtocol is not used as there
+           // may be a race to set it to null.
+          ProtocolVersion pv = conContext.protocolVersion;
            if (pv == null || (!pv.useTLS13PlusSpec())) {
                hasCloseReceipt = true;
            }       
```

I like the idea as it looks like a safe update in the current implementation.

mpdonova: Yes, that is what I was thinking.I will make the change and test it out. Thanks.

XueleiFan (approved):
Looks good to me.  Thank you!

PR #13750, opened 2023-05-02
8303942: FileMapInfo::write_bytes aborts on a short os::write

afshin-zafari:
`os::write` is implemented using loops until the whole bytes are written. All uses of `os::write` in a loop are changed to single call. 
Platform dependent versions of the `os::write` are also renamed and moved to private sections accordingly.

### DIFF
--- a/src/hotspot/os/posix/perfMemory_posix.cpp
+++ b/src/hotspot/os/posix/perfMemory_posix.cpp
@@ -104,20 +104,14 @@ static void save_memory_to_file(char* addr, size_t size) {
   } else {
     ssize_t result;
 
-    for (size_t remaining = size; remaining > 0;) {
-
-      result = os::write(fd, addr, remaining);
-      if (result == OS_ERR) {
-        if (PrintMiscellaneous && Verbose) {
-          warning("Could not write Perfdata save file: %s: %s\n",
-                  destfile, os::strerror(errno));
-        }
-        break;
+    result = os::write(fd, addr, size);
+    if (result == OS_ERR) {
+      if (PrintMiscellaneous && Verbose) {
+        warning("Could not write Perfdata save file: %s: %s\n",
+                destfile, os::strerror(errno));
       }
-
-      remaining -= (size_t)result;
-      addr += result;
     }
+
 
     result = ::close(fd);
     if (PrintMiscellaneous && Verbose) {
--- a/src/hotspot/share/cds/filemap.cpp
+++ b/src/hotspot/share/cds/filemap.cpp
@@ -361,7 +361,6 @@ void SharedClassPathEntry::set_name(const char* name, TRAPS) {
 }
 
 void SharedClassPathEntry::copy_from(SharedClassPathEntry* ent, ClassLoaderData* loader_data, TRAPS) {
-  assert(ent != NULL, "sanity");
   _type = ent->_type;
   _is_module_path = ent->_is_module_path;
   _timestamp = ent->_timestamp;
@@ -1686,7 +1685,7 @@ size_t FileMapInfo::write_heap_region(ArchiveHeapInfo* heap_info) {
 void FileMapInfo::write_bytes(const void* buffer, size_t nbytes) {
   assert(_file_open, "must be");
   ssize_t n = os::write(_fd, buffer, (unsigned int)nbytes);
-  if (n < 0 || (size_t)n != nbytes) {
+  if (n < 0) {
     // If the shared archive is corrupted, close it and remove it.
     close();
     remove(_full_path);
--- a/src/hotspot/share/jfr/recorder/repository/jfrEmergencyDump.cpp
+++ b/src/hotspot/share/jfr/recorder/repository/jfrEmergencyDump.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -372,20 +372,26 @@ static void write_repository_files(const RepositoryIterator& iterator, char* con
     assert(fqn != NULL, "invariant");
     current_fd = open_exclusivly(fqn);
     if (current_fd != invalid_fd) {
-      const int64_t size = file_size(current_fd);
+      const size_t size = (size_t)file_size(current_fd);
       assert(size > 0, "invariant");
-      int64_t bytes_read = 0;
-      int64_t bytes_written = 0;
+      unsigned int bytes_read = 0;
+      unsigned int bytes_written = 0;
       while (bytes_read < size) {
         const ssize_t read_result = os::read_at(current_fd, copy_block, (int)block_size, bytes_read);
         if (-1 == read_result) {
           log_info(jfr)( // For user, should not be "jfr, system"
-              "Unable to recover JFR data");
+              "Unable to recover JFR data, read failed.");
           break;
         }
-        bytes_read += (int64_t)read_result;
-        assert(bytes_read - bytes_written <= (int64_t)block_size, "invariant");
-        bytes_written += (int64_t)os::write(emergency_fd, copy_block, bytes_read - bytes_written);
+        bytes_read += read_result;
+        assert(bytes_read - bytes_written <= block_size, "invariant");
+        const ssize_t write_result = os::write(emergency_fd, copy_block, bytes_read - bytes_written);
+        if (-1 == write_result) {
+          log_info(jfr)( // For user, should not be "jfr, system"
+              "Unable to recover JFR data, write failed.");
+          break;
+        }
+        bytes_written += write_result;
         assert(bytes_read == bytes_written, "invariant");
       }
       ::close(current_fd);
--- a/src/hotspot/share/jfr/writers/jfrStreamWriterHost.inline.hpp
+++ b/src/hotspot/share/jfr/writers/jfrStreamWriterHost.inline.hpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -74,17 +74,13 @@ inline void StreamWriterHost<Adapter, AP>::write_bytes(void* dest, const void* b
 template <typename Adapter, typename AP>
 inline void StreamWriterHost<Adapter, AP>::write_bytes(const u1* buf, intptr_t len) {
   assert(len >= 0, "invariant");
-  while (len > 0) {
-    const unsigned int nBytes = len > INT_MAX ? INT_MAX : (unsigned int)len;
-    const ssize_t num_written = os::write(_fd, buf, nBytes);
-    if (errno == ENOSPC) {
-      JfrJavaSupport::abort("Failed to write to jfr stream because no space left on device", false);
-    }
-    guarantee(num_written > 0, "Nothing got written, or os::write() failed");
-    _stream_pos += num_written;
-    len -= num_written;
-    buf += num_written;
+  const unsigned int nBytes = len > INT_MAX ? INT_MAX : (unsigned int)len;
+  const ssize_t num_written = os::write(_fd, buf, nBytes);
+  if (errno == ENOSPC) {
+    JfrJavaSupport::abort("Failed to write to jfr stream because no space left on device", false);
   }
+  guarantee(num_written > 0, "Nothing got written, or os::write() failed");
+  _stream_pos += num_written;
 }
 
 template <typename Adapter, typename AP>
--- a/src/hotspot/share/runtime/os.hpp
+++ b/src/hotspot/share/runtime/os.hpp
@@ -229,6 +229,7 @@ class os: AllStatic {
   // Get summary strings for system information in buffer provided
   static void  get_summary_cpu_info(char* buf, size_t buflen);
   static void  get_summary_os_info(char* buf, size_t buflen);
+  static ssize_t pd_write(int fd, const void *buf, unsigned int nBytes);
 
   static void initialize_initial_active_processor_count();
 
--- a/src/hotspot/share/services/heapDumperCompression.cpp
+++ b/src/hotspot/share/services/heapDumperCompression.cpp
@@ -55,14 +55,9 @@ char const* FileWriter::write_buf(char* buf, ssize_t size) {
   assert(_fd >= 0, "Must be open");
   assert(size > 0, "Must write at least one byte");
 
-  while (size > 0) {
-    ssize_t n = os::write(_fd, buf, (uint) size);
-    if (n <= 0) {
-      return os::strerror(errno);
-    }
-
-    buf += n;
-    size -= n;
+  ssize_t n = os::write(_fd, buf, (uint)size);
+  if (n <= 0) {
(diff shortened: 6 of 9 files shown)


src/hotspot/os/posix/perfMemory_posix.cpp
```diff
@@ -104,21 +104,15 @@ static void save_memory_to_file(char* addr, size_t size) {
...
-      result = os::write(fd, addr, remaining);
-      if (result == OS_ERR) {
-        if (PrintMiscellaneous && Verbose) {
-          warning("Could not write Perfdata save file: %s: %s\n",
-                  destfile, os::strerror(errno));
-        }
-        break;
+    result = os::write(fd, addr, size);
+    if (result == OS_ERR) {
+      if (PrintMiscellaneous && Verbose) {
```
coleenp, line 109: It's not really part of this issue but since the line is changed, can you change it to unconditionally
   log_info(os)("Could not write...);
And remove PrintMiscellaneous & Verbose.

coleenp (approved):
This change looks good.

iklam:
Can you update os.hpp to indicate that the buffer will be fully written? 

I would also request that the input size to be changed to size_t, to be consistent with the C library. There are too many dubious casting of size_t to int in the code.

```
#include <unistd.h>
ssize_t write(int fd, const void *buf, size_t count);
```

Also, when an error happens, what is the returned value? Is it always negative. or will you return the number of partially written bytes?

For failures, I think returning the number of partially written bytes is not useful. The failure would be caused by an unrecoverable error, so you can't try to write the remaining bytes again (or else we are back to the original loop!). For simplicity, this function can simply return -1 to indicate failure, and 0 to indicate success.

afshin-zafari:
> Can you update os.hpp to indicate that the buffer will be fully written?
> 
> I would also request that the input size to be changed to size_t, to be consistent with the C library. There are too many dubious casting of size_t to int in the code.

Comment is added including the possible return values of 0 and -1. Parameter type is changed to `size_t` and all invocations changed.

> For failures, I think returning the number of partially written bytes is not useful. The failure would be caused by an unrecoverable error, so you can't try to write the remaining bytes again (or else we are back to the original loop!). For simplicity, this function can simply return -1 to indicate failure, and 0 to indicate success.

The return values are changed as suggested. All the calls to `os::write` are changed accordingly.

src/hotspot/os/posix/perfMemory_posix.cpp
afshin-zafari: Done.

src/hotspot/share/jfr/recorder/repository/jfrEmergencyDump.cpp
```diff
@@ -372,21 +372,26 @@ static void write_repository_files(const RepositoryIterator& iterator, char* con
     assert(fqn != NULL, "invariant");
     current_fd = open_exclusivly(fqn);
     if (current_fd != invalid_fd) {
-      const int64_t size = file_size(current_fd);
+      const size_t size = (size_t)file_size(current_fd);
```
iklam, line 375: There's an existing bug here: error code of -1 from `file_size` is not handled.

src/hotspot/share/cds/filemap.cpp
```diff
@@ -361,7 +361,6 @@ void SharedClassPathEntry::set_name(const char* name, TRAPS) {
 }
 
 void SharedClassPathEntry::copy_from(SharedClassPathEntry* ent, ClassLoaderData* loader_data, TRAPS) {
-  assert(ent != NULL, "sanity");
```
iklam, line 364: This removal seems to be unrelated to this PR.

src/hotspot/share/jfr/writers/jfrStreamWriterHost.inline.hpp
```diff
@@ -74,17 +74,13 @@ inline void StreamWriterHost<Adapter, AP>::write_bytes(void* dest, const void* b
...
-    guarantee(num_written > 0, "Nothing got written, or os::write() failed");
-    _stream_pos += num_written;
-    len -= num_written;
-    buf += num_written;
+  const unsigned int nBytes = len > INT_MAX ? INT_MAX : (unsigned int)len;
+  const ssize_t num_written = os::write(_fd, buf, nBytes);
+  if (num_written < 0 && errno == ENOSPC) {
+    JfrJavaSupport::abort("Failed to write to jfr stream because no space left on device", false);
   }
+  guarantee(num_written == 0, "Not all the bytes got written, or os::write() failed");
```
iklam, line 82: guarantee() seems to be a bad way of handling this. I would suggest filing an RFE for more robust error handling.

iklam (approved):
Looks good Just some minor nits.

src/hotspot/share/cds/filemap.cpp
afshin-zafari: Good catch, thanks.
The change is not part of mines. I think, it came up after merging with master. 
The line is back to the code now.

src/hotspot/share/jfr/recorder/repository/jfrEmergencyDump.cpp
afshin-zafari: The type is returned to signed and the assert checks the -1.

src/hotspot/share/jfr/writers/jfrStreamWriterHost.inline.hpp
afshin-zafari: This RFE is created:
https://bugs.openjdk.org/browse/JDK-8307579
The return values of the os::write in jfrStreamWriterHost.inline.hpp:StreamWriterHost<Adapter, AP>::write_bytes(), should be handled more robustly.

src/hotspot/share/jfr/recorder/repository/jfrEmergencyDump.cpp
```diff
@@ -374,19 +374,24 @@ static void write_repository_files(const RepositoryIterator& iterator, char* con
     if (current_fd != invalid_fd) {
       const int64_t size = file_size(current_fd);
       assert(size > 0, "invariant");
-      int64_t bytes_read = 0;
-      int64_t bytes_written = 0;
+      unsigned int bytes_read = 0;
+      unsigned int bytes_written = 0;
```
dholmes-ora, line 378: Why have you changed this to a 32-bit type?

src/hotspot/share/runtime/os.hpp
```diff
@@ -229,6 +229,7 @@ class os: AllStatic {
   // Get summary strings for system information in buffer provided
   static void  get_summary_cpu_info(char* buf, size_t buflen);
   static void  get_summary_os_info(char* buf, size_t buflen);
+  static ssize_t pd_write(int fd, const void *buf, size_t nBytes);
```
dholmes-ora, line 232: What is the required meaning of the return value here? I'm assuming < 0 -> error; while >= 0 -> bytes written?

src/hotspot/share/runtime/os.hpp
```diff
@@ -644,7 +645,8 @@ class os: AllStatic {
   //File i/o operations
 
   static ssize_t read_at(int fd, void *buf, unsigned int nBytes, jlong offset);
-  static ssize_t write(int fd, const void *buf, unsigned int nBytes);
+  // Writes the bytes completely. Returns 0 on success, -1 otherwise.
+  static ssize_t write(int fd, const void *buf, size_t nBytes);
```
dholmes-ora, line 649: We don't need a ssize_t return type if we only return -1 or 0. Also this should be specified to return OS_ERR on error and OS_OK on success. The callers should also check for OS_ERR and OS_OK rather than -1, < 0, > 0 etc.

Or we could simply make this a boolean function.

dholmes-ora (changes requested):

src/hotspot/share/jfr/recorder/repository/jfrEmergencyDump.cpp
afshin-zafari: Changed back to `int64_t`.

src/hotspot/share/runtime/os.hpp
afshin-zafari: Return value changed to `bool`. All calls changed accordingly.

src/hotspot/share/runtime/os.hpp
afshin-zafari: Return value changed to `bool`. All calls changed accordingly.

src/hotspot/share/jfr/writers/jfrStreamWriterHost.inline.hpp
```diff
@@ -74,17 +74,13 @@ inline void StreamWriterHost<Adapter, AP>::write_bytes(void* dest, const void* b
...
-    const unsigned int nBytes = len > INT_MAX ? INT_MAX : (unsigned int)len;
-    const ssize_t num_written = os::write(_fd, buf, nBytes);
-    if (errno == ENOSPC) {
-      JfrJavaSupport::abort("Failed to write to jfr stream because no space left on device", false);
-    }
-    guarantee(num_written > 0, "Nothing got written, or os::write() failed");
-    _stream_pos += num_written;
-    len -= num_written;
-    buf += num_written;
+  const unsigned int nBytes = len > INT_MAX ? INT_MAX : (unsigned int)len;
```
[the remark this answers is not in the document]
afshin-zafari: The `os::write` itself writes in a loop.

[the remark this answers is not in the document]
mgronlun: Yes, but only loops INT_MAX now?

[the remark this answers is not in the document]
afshin-zafari: Oh, yes. I got your point now. `len` can be larger than `INT_MAX`. I put the loop back.

src/hotspot/os/posix/perfMemory_posix.cpp
```diff
@@ -97,33 +97,21 @@ static void save_memory_to_file(char* addr, size_t size) {
...
-          warning("Could not write Perfdata save file: %s: %s\n",
-                  destfile, os::strerror(errno));
-        }
-        break;
-      }
-
-      remaining -= (size_t)result;
-      addr += result;
+    bool successful_write = os::write(fd, addr, size);
+    if (!successful_write) {
```
dholmes-ora, line 106: You don't need to introduce the local variable here.

src/hotspot/os/posix/perfMemory_posix.cpp
```diff
@@ -961,11 +949,12 @@ static int create_sharedmem_file(const char* dirname, const char* filename, size
     int zero_int = 0;
     result = (int)os::seek_to_file_offset(fd, (jlong)(seekpos));
     if (result == -1 ) break;
-    result = os::write(fd, &zero_int, 1);
-    if (result != 1) {
+    bool successful_write = os::write(fd, &zero_int, 1);
+    if (!successful_write) {
```
dholmes-ora, line 953: Ditto no need for a local

src/hotspot/share/cds/filemap.cpp
```diff
@@ -1685,8 +1685,8 @@ size_t FileMapInfo::write_heap_region(ArchiveHeapInfo* heap_info) {
 
 void FileMapInfo::write_bytes(const void* buffer, size_t nbytes) {
   assert(_file_open, "must be");
-  ssize_t n = os::write(_fd, buffer, (unsigned int)nbytes);
-  if (n < 0 || (size_t)n != nbytes) {
+  bool successful_write = os::write(_fd, buffer, nbytes);
+  if (!successful_write) {
```
dholmes-ora, line 1689: No local please

src/hotspot/share/jfr/recorder/repository/jfrEmergencyDump.cpp
```diff
@@ -380,13 +380,18 @@ static void write_repository_files(const RepositoryIterator& iterator, char* con
...
-              "Unable to recover JFR data");
+              "Unable to recover JFR data, read failed.");
           break;
         }
         bytes_read += (int64_t)read_result;
         assert(bytes_read - bytes_written <= (int64_t)block_size, "invariant");
-        bytes_written += (int64_t)os::write(emergency_fd, copy_block, bytes_read - bytes_written);
-        assert(bytes_read == bytes_written, "invariant");
+        const bool successful_write = os::write(emergency_fd, copy_block, bytes_read - bytes_written);
+        if (!successful_write) {
```
dholmes-ora, line 389: No local please.

src/hotspot/share/jfr/writers/jfrStreamWriterHost.inline.hpp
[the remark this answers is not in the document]
dholmes-ora: @mgronlun why does this code break the write up into INT_MAX chunks? Is the incoming `len` parameter really potentially not containable in a `size_t`? Using `intptr_t` for a length seems suspect.

src/hotspot/share/services/heapDumperCompression.cpp
```diff
@@ -55,14 +55,9 @@ char const* FileWriter::write_buf(char* buf, ssize_t size) {
...
-  while (size > 0) {
-    ssize_t n = os::write(_fd, buf, (uint) size);
-    if (n <= 0) {
-      return os::strerror(errno);
-    }
-
-    buf += n;
-    size -= n;
+  bool successful_write = os::write(_fd, buf, (size_t)size);
+  if (!successful_write) {
```
dholmes-ora, line 59: No local please

dholmes-ora (changes requested):
Thanks for the updates @afshin-zafari ! One last nit with the usage of the boolean function.

src/hotspot/os/posix/perfMemory_posix.cpp
afshin-zafari: Done.

src/hotspot/share/cds/filemap.cpp
afshin-zafari: Done

src/hotspot/share/services/heapDumperCompression.cpp
afshin-zafari: Done

src/hotspot/share/jfr/recorder/repository/jfrEmergencyDump.cpp
afshin-zafari: Done

mgronlun:
"@mgronlun why does this code break the write up into INT_MAX chunks? Is the incoming len parameter really potentially not containable in a size_t? Using intptr_t for a length seems suspect."

I think it has mostly to do with legacy os::write() implementations and being able to write completely on all platforms.

The len was size_t up until this bug: https://bugs.openjdk.org/browse/JDK-8252090

dholmes-ora (approved):
Looks good. Nothing further from me.

Thanks for your patience.

afshin-zafari:
> "@mgronlun why does this code break the write up into INT_MAX chunks? Is the incoming len parameter really potentially not containable in a size_t? Using intptr_t for a length seems suspect."
> 
> I think it has mostly to do with legacy os::write() implementations and being able to write completely on all platforms.
> 
> The len was size_t up until this bug: https://bugs.openjdk.org/browse/JDK-8252090

Dear @mgronlun, do you want any changes in this PR? Or the change is acceptable.

mgronlun (approved):

coleenp (approved):
This looks good, thanks for adding the comment and Windows handling for >unsigned int bytes.

afshin-zafari:
Thank you @coleenp, @dholmes-ora, @mgronlun and @iklam for all your comments!
/integrate

dholmes-ora:
@afshin-zafari if you make non-trivial changes after reviews have been completed, please ensure at least 2 reviewers see the final set of changes. Thanks.

afshin-zafari:
> @afshin-zafari if you make non-trivial changes after reviews have been completed, please ensure at least 2 reviewers see the final set of changes. Thanks.

Yes, for sure. @coleenp and @mgronlun (in slack channel) approved it before I integrated. Thanks.

PR #13758, opened 2023-05-02
8307244: Remove redundant class RMIIIOPServerImpl

kevinjwalls:
Removal of class, looks like it was missed in the JDK9 removal of RMIIIOP.
This class is not referenced by other classes or tests.

---------
### Progress
- [x] Change requires CSR request [JDK-8307329](https://bugs.openjdk.org/browse/JDK-8307329) to be approved

### Issues
 * [JDK-8307244](https://bugs.openjdk.org/browse/JDK-8307244): Remove redundant class RMIIIOPServerImpl
 * [JDK-8307329](https://bugs.openjdk.org/browse/JDK-8307329): Remove redundant class RMIIIOPServerImpl (**CSR**)

### Reviewers
 * [Serguei Spitsyn](https://openjdk.org/census#sspitsyn) (@sspitsyn - **Reviewer**) ⚠️ Review applies to [8dd59435](https://git.openjdk.org/jdk/pull/13758/files/8dd59435b33132d8eac4fa280c0e954392abe0e8)
 * [Daniel Fuchs](https://openjdk.org/census#dfuchs) (@dfuch - **Reviewer**) ⚠️ Review applies to [8dd59435](https://git.openjdk.org/jdk/pull/13758/files/8dd59435b33132d8eac4fa280c0e954392abe0e8)
 * [Alan Bateman](https://openjdk.org/census#alanb) (@AlanBateman - **Reviewer**) ⚠️ Review applies to [8dd59435](https://git.openjdk.org/jdk/pull/13758/files/8dd59435b33132d8eac4fa280c0e954392abe0e8)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13758/head:pull/13758` \
`$ git checkout pull/13758`

`$ git checkout pull/13758` \
`$ git pull https://git.openjdk.org/jdk.git pull/13758/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13758`

`$ git pr show -t 13758`

</details>

<a href="https://git.openjdk.org/jdk/pull/13758.diff">https://git.openjdk.org/jdk/pull/13758.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13758#issuecomment-1531916406)


sspitsyn (approved):
As I see, this transport was deprecated for some time.
But it is not clear in what release.
The fix looks good to me.
Thanks,
Serguei

AlanBateman:
JSR-160 MR3 changed RMIConnector to make the IIOP transport optional. For Java 8 this was JDK-8001048. For Java 9, the IIOP transport was removed via JDK-8043937. It should be okay to remove RMIIIOPServerImpl now. There's an example in the RMIConnector javadoc that uses a service URL and the "iiop" protocol, which I think is okay to leave.

dfuch:
Looks good to me. I probably wouldn't have bothered with removing the example of IIOP JMXServiceURL as arguably functional implementations of that might still exist (as long as they don't extend RMIIIOPServerImpl, which is not a requirement).
The constructor of  RMIIIOPServerImpl throws UnsupportedOperationException unconditionally, so there can't exist any functional subclasses of that class that could be instantiated on Java versions posterior to Java 9.
As such removing that class from the public API sounds reasonable.

dfuch (approved):

AlanBateman (approved):
Removing RMIIIOPServerImpl is fine, I think I agree with Daniel on leaving out the change to RMIConnector. It may use "iiop" but it just an example.

kevinjwalls:
Thanks for all the comments and reviews.
On the RMIConnector example, I had thought it was misleading to use as an example something which used to work, but which no longer works.  But this change doesn't affect whether other address syntaxes are recognised.

kevinjwalls:
Putting the example back in RMIConnector, but I was keen to remove "iiop" from search results here, so with a generic placeholder for the protocol.

src/java.management.rmi/share/classes/javax/management/remote/rmi/RMIConnector.java
```diff
@@ -185,7 +185,7 @@ private RMIConnector(RMIServer rmiServer, JMXServiceURL address,
      * syntaxes, for example:</p>
      *
      * <pre>
-     * service:jmx:iiop://<em>[host[:port]]</em>/stub/<em>encoded-stub</em>
+     * service:jmx:protocolname://<em>[host[:port]]</em>/stub/<em>encoded-stub</em>
```
AlanBateman, line 188: "protocolname" could be read as literal here. I think change it back to "iiop" or replace it with something that like "myprotocol" that is more obviously an example protocol name.

kevinjwalls: ok

kevinjwalls: Could it really?  But ok, done. 8-)

PR #13760, opened 2023-05-02
8307298: JFR: Ensure jdk.jfr.internal.TypeLibrary is initialized only once

roberttoyonaga:
Hello, may I have a review of a small change that enforces that `jdk.jfr.internal.TypeLibrary` only gets initialized once?
Previously, it was the case that `TypeLibrary` was only initialized once, but since [this change](https://github.com/openjdk/jdk/pull/12884/files) it can now be initialized multiple times. This can be a problem because re-initialization [results in attempting to add elements to `jdk.jfr.internal.Type#fields`](https://github.com/openjdk/jdk/blob/master/src/jdk.jfr/share/classes/jdk/jfr/internal/MetadataLoader.java#L264) which may have already been changed to an unmodifiableList after the initial initialization of `TypeLibrary`.

This is of concern for JFR support in GraalVM Native Image. Specifically, multiple initializations of TypeLibrary breaks the ability to use JFR at native image built time. This is because, if JFR is used at image build time, the data of a few of the JFR singletons such as MetadataRepository must be re-initialized for runtime recordings to start with a "clean-slate". However, re-initializing MetadataRepository, also has the effect of calling `TypeLibrary#initialize()` which now unconditionally attempts to initialize TypeRepository, resulting in errors.

Example of exception upon second initialization of `TypeRepository`: 
```
Caused by: java.lang.InternalError: java.lang.UnsupportedOperationException
	at jdk.jfr/jdk.jfr.internal.MetadataLoader.createTypes(MetadataLoader.java:195)
	at jdk.jfr/jdk.jfr.internal.TypeLibrary.initialize(TypeLibrary.java:103)
	at jdk.jfr/jdk.jfr.internal.MetadataRepository.initializeJVMEventTypes(MetadataRepository.java:72)
	at jdk.jfr/jdk.jfr.internal.MetadataRepository.<init>(MetadataRepository.java:68)
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.jfr.Target_jdk_jfr_internal_MetadataRepository.<clinit>(Target_jdk_jfr_internal_MetadataRepository.java:42)
	at java.base/jdk.internal.misc.Unsafe.ensureClassInitialized0(Native Method)
	at java.base/jdk.internal.misc.Unsafe.ensureClassInitialized(Unsafe.java:1160)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.classinitialization.ClassInitializationSupport.ensureClassInitialized(ClassInitializationSupport.java:168)
	... 11 more
Caused by: java.lang.UnsupportedOperationException
	at java.base/java.util.Collections$UnmodifiableCollection.add(Collections.java:1085)
	at jdk.jfr/jdk.jfr.internal.Type.add(Type.java:227)
	at jdk.jfr/jdk.jfr.internal.MetadataLoader.addFields(MetadataLoader.java:264)
	at jdk.jfr/jdk.jfr.internal.MetadataLoader.buildTypes(MetadataLoader.java:202)
	at jdk.jfr/jdk.jfr.internal.MetadataLoader.createTypes(MetadataLoader.java:193)
```

---------
### Progress

### Issue
 * [JDK-8307298](https://bugs.openjdk.org/browse/JDK-8307298): JFR: Ensure jdk.jfr.internal.TypeLibrary is initialized only once

### Reviewers
 * [Severin Gehwolf](https://openjdk.org/census#sgehwolf) (@jerboaa - **Reviewer**)
 * [Erik Gahlin](https://openjdk.org/census#egahlin) (@egahlin - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13760/head:pull/13760` \
`$ git checkout pull/13760`

`$ git checkout pull/13760` \
`$ git pull https://git.openjdk.org/jdk.git pull/13760/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13760`

`$ git pr show -t 13760`

</details>

<a href="https://git.openjdk.org/jdk/pull/13760.diff">https://git.openjdk.org/jdk/pull/13760.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13760#issuecomment-1540449648)


src/jdk.jfr/share/classes/jdk/jfr/internal/TypeLibrary.java
```diff
@@ -98,6 +98,9 @@ private static ValueDescriptor createDurationField() {
     }
 
     public static synchronized void initialize() {
+        if (initialized) {
+            return;
+        }
```
jerboaa, line 103: Perhaps it would make sense to add a comment as to why that check needs to be there? Something like this perhaps?

```
// TypeLibrary is initialized once by MetadataRepository (which is a singleton). So shouldn't happen practice.
// However, some tools (GraalVM native image) may call the initialization routine twice.
```

jerboaa: Thanks!

roberttoyonaga:
Hi @egahlin when you have time, can you please have a look at this change?

egahlin:
> Hi @egahlin when you have time, can you please have a look at this change?

Looks reasonable.

Is there a (GraalVM) reason the boolean is volatile? Should be sufficient with synchronized.

(I will probably refactor the code, so implicitFieldTypes field can be removed, but it probably makes sense to keep that in a separate fix)

roberttoyonaga:
> Is there a (GraalVM) reason the boolean is volatile? Should be sufficient with synchronized.

Yes, the problem isn't that there's a race to initialize `TypeLibrary`. Instead, the reason this is needed for Native Image is that JFR is used in two stages, both at build time and runtime. Because of this, JFR singletons such as `MetadataRepository` must be cleared and reinitialized between stages in order to keep JFR data from one stage leaking into the next (ie. we reinitialize after build time before running). This means that `TypeLibrary` will attempt to be initialized twice.

jerboaa (approved):
Looks good. I've set reviewers to 2, so that this won't be the only review.

egahlin (approved):

jerboaa:
Windows builds are failing due to a network glitch. Seems unrelated.

PR #13761, opened 2023-05-02
8307165: java/awt/dnd/NoFormatsDropTest/NoFormatsDropTest.java timed out

DamonGuy:
When converting this test to be applicable to be open sourced, changes were made. I reverted some of these changes to get the test stable again. Now it passes when run on various OS's 100x.


prrace:
The delay -> sleep changes really are no-ops .. except one you made shorter ..
And the calls like validate() seem OK .. 
Do you have any explanation as to why these changes fix the time out ?

test/jdk/java/awt/dnd/NoFormatsDropTest/NoFormatsDropTest.java
```diff
@@ -94,17 +94,14 @@ public void init() {
...
-        frame.validate();
     }
 
     public void start() throws InterruptedException, AWTException,
             InvocationTargetException {
         try {
             Robot robot = new Robot();
-            robot.setAutoWaitForIdle(true);
-            robot.delay(FRAME_ACTIVATION_TIMEOUT);
+            Thread.sleep(FRAME_ACTIVATION_TIMEOUT);
```
honkar-jdk, line 104: Curious to know the difference of using `Thread.sleep()` instead of `robot.delay()` here ?

test/jdk/java/awt/dnd/NoFormatsDropTest/NoFormatsDropTest.java
```diff
@@ -94,17 +94,14 @@ public void init() {
 
         frame.getToolkit().addAWTEventListener(this, AWTEvent.MOUSE_EVENT_MASK);
         frame.pack();
-        frame.setLocationRelativeTo(null);
```
honkar-jdk, line 97: It would be better to retain `frame.setLocationRelativeTo(null)` since you are calling `getLocationOnScreen()` later on.

DamonGuy:
> The delay -> sleep changes really are no-ops .. except one you made shorter .. And the calls like validate() seem OK .. Do you have any explanation as to why these changes fix the time out ?

The original test had this shorter delay, so I experimented with reverting the change here. After your suggestion to make ```passed``` volatile, the test passes on windows. The original intermittent failure due to timeout seems to have been caused by the setAutoWaitForIdle() call. 

On my local machine, this line causes the test to take much longer to start for some reason. After removing this, the test consistently passes on 100 repeat runs of the test on all OS's except Windows. Windows fails intermittently as well, but changing the ```passed``` var to volatile fixes this. Now all OS's are passing 100 repeat runs.

test/jdk/java/awt/dnd/NoFormatsDropTest/NoFormatsDropTest.java
DamonGuy: I have restored the line

test/jdk/java/awt/dnd/NoFormatsDropTest/NoFormatsDropTest.java
DamonGuy: Initially, I thought ```Thread.sleep``` may have been the cause since there are some synchronization lines. But, I think robot.delay does the same behavior here. So, to keep the modern test format, I reverted this to robot.delay

prrace (approved):
OK.  These changes should improve things. Let's push it and see.

mrserb:
Can you please clarify how long the affected test was executed before the fix, it was slow or hung?

PR #13762, opened 2023-05-02
8179502: Enhance OCSP, CRL and Certificate Fetch Timeouts

jnimeh:
This set of enhancements extends the allowed syntax for the `com.sun.security.ocsp.timeout`, `com.sun.security.crl.timeout` and `com.sun.security.crl.readtimeout` System properties.  These properties retain their current behavior where a purely numeric value is interpreted in seconds, but now the numeric value may also be appended with "ms" (case-insensitive) to be interpreted as milliseconds.

This enhancement also adds two new System properties: `com.sun.security.cert.timeout` and `com.sun.security.cert.readtimeout` which follow the same new allowed syntax.  These timeouts only come into play when an AIA extension on a certificate is followed for pulling the issuing authority certificate and only when the `com.sun.security.enableAIAcaIssuers` property is true (default false).

JBS: https://bugs.openjdk.org/browse/JDK-8179502
CSR: https://bugs.openjdk.org/browse/JDK-8300722

---------
### Progress
- [x] Change requires CSR request [JDK-8300722](https://bugs.openjdk.org/browse/JDK-8300722) to be approved

### Issues
 * [JDK-8179502](https://bugs.openjdk.org/browse/JDK-8179502): Enhance OCSP, CRL and Certificate Fetch Timeouts
 * [JDK-8300722](https://bugs.openjdk.org/browse/JDK-8300722): Enhance OCSP, CRL and Certificate Fetch Timeouts (**CSR**)

### Reviewers
 * [Sean Mullan](https://openjdk.org/census#mullan) (@seanjmullan - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13762/head:pull/13762` \
`$ git checkout pull/13762`

`$ git checkout pull/13762` \
`$ git pull https://git.openjdk.org/jdk.git pull/13762/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13762`

`$ git pr show -t 13762`

</details>

<a href="https://git.openjdk.org/jdk/pull/13762.diff">https://git.openjdk.org/jdk/pull/13762.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13762#issuecomment-1532204616)


src/java.base/share/classes/sun/security/action/GetPropertyAction.java
```diff
@@ -160,4 +161,45 @@ public Properties run() {
...
+            def = 0;
+        }
+
+        String propVal = System.getProperty(prop, "").trim();
+        if (propVal.length() == 0) {
+            return def;
+        }
+
+        // Determine if "ms" is on the end of the string
+        boolean isMillis = propVal.toLowerCase().endsWith("ms");
```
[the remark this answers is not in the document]
jnimeh: Well, all the existing documentation already states that they are in seconds.  That was why I didn't add any additional suffixes.  The goal was to make it so folks don't need to make any changes if the existing seconds-level granularity is sufficient for them.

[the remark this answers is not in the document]
jnimeh: Okay, we can make that change.

src/java.base/share/classes/sun/security/provider/certpath/OCSP.java
seanjmullan, line 1: I see there is no way to individually control the OCSP read and connect timeouts like there is for certs and CRLs. Perhaps this isn't as big an issue, but when you set the OCSP timeout, it really means 2x what you set.

jnimeh: Yes, I noticed that too.  I wasn't sure if we needed to make a change there.  I opted to leave well-enough alone since nobody was asking for it and it's one less property to keep track of.  All of these property sets end up with a max latency of connect-timeout + read-timeout, and by default they are set to the same values.  So in practice much of the time they are all 2x.

It's easy enough I think to make a separate property for `com.sun.security.ocsp.readtimeout` and then the existing `.timeout` property would be for connect timeouts (keeping in line with the other props).  I don't think it will introduce significant risk but I will highlight that in the CSR.

jnimeh: > I think you should also apply the cert and CRL timeouts to the `LDAPCertStore` implementation, using the JNDI properties: `com.sun.jndi.ldap.connect.timeout` and `com.sun.jndi.ldap.read.timeout`.

I will look into this.

jnimeh: I've added the OCSP readtimeout property, seems to be working well.  As discussed offline we'll hold off on the LDAP changes for now.

src/java.base/share/classes/sun/security/action/GetPropertyAction.java
```diff
@@ -160,4 +161,48 @@ public Properties run() {
...
+     *      then the default value will be returned.  If a negative value for
+     *      the "def" parameter is supplied, zero will be returned if the
+     *      property's value does not conform to the allowed syntax.
+     */
+    public static int privilegedGetTimeoutProp(String prop, int def) {
+        if (def < 0) {
+            def = 0;
+        }
+
+        String propVal = System.getProperty(prop, "").trim();
```
seanjmullan, line 186: You should call `privilegedGetProperty` here instead of `System.getProperty` so the call is wrapped in `doPrivileged` when an SM is active.

src/java.base/share/classes/sun/security/action/GetPropertyAction.java
```diff
@@ -160,4 +161,48 @@ public Properties run() {
...
+        if (propVal.toLowerCase().endsWith("ms")) {
+            propVal = propVal.substring(0, propVal.length() - 2);
+            isMillis = true;
+        } else if (propVal.toLowerCase().endsWith("s")) {
+            propVal = propVal.substring(0, propVal.length() - 1);
+        }
+
+        // Next check to make sure the string is built only from digits
+        if (propVal.matches("^\\d+$")) {
+            int timeout = Integer.parseInt(propVal);
```
seanjmullan, line 202: Is this guaranteed never to throw `NumberFormatException`? It might be safer to catch it just in case.

jnimeh: I'll change this to catch NFE, but I'm pretty sure the pattern will only ever return on true if the string is comprised solely of digits from start to end - I could never get a string that would pass when it shouldn't.  But point taken, better safe than sorry.

src/java.base/share/classes/sun/security/action/GetPropertyAction.java
jnimeh: Good catch.  Will fix.

src/java.base/share/classes/sun/security/action/GetPropertyAction.java
```diff
@@ -160,4 +161,58 @@ public Properties run() {
...
+        if (propVal.matches("^\\d+$")) {
+            try {
+                int timeout = Integer.parseInt(propVal);
+                return isMillis ? timeout : timeout * 1000;
+            } catch (NumberFormatException nfe) {
+                if (dbg != null) {
+                    dbg.println("Warning: Unexpected " + nfe +
+                            " for timeout value " + propVal +
+                            ". Using default value of " + def + " msec.");
+                }
```
seanjmullan, line 211: This would also be useful debug for the else case on line 214-216 if the value is not an integer.

src/java.base/share/classes/sun/security/provider/certpath/URICertStore.java
```diff
@@ -127,8 +128,12 @@ class URICertStore extends CertStoreSpi {
     // allowed when downloading CRLs
     private static final int DEFAULT_CRL_READ_TIMEOUT = 15000;
 
+    // Default connect and read timeouts for CA certificate fetching (15 sec)
```
seanjmullan, line 131: Does 15 seconds make sense as the default timeout, especially for certs? CRLs are generally larger than certs, so a longer read timeout makes sense.

I'm ok with keeping these default values the same for consistency, but I think we should re-evaluate each of these default timeouts and compare them to other products/technologies to see if some adjustments may be needed - can you file a follow-on RFE for that?

jnimeh:
@seanjmullan if you have a chance would you mind taking a quick look at the release note for this change? (https://bugs.openjdk.org/browse/JDK-8308582)
It is pretty close to the CSR solution text, but I tried to trim it down a little.  It's still pretty long, but there's a lot to convey.

src/java.base/share/classes/sun/security/action/GetPropertyAction.java
jnimeh: Sounds good, I can add that.

src/java.base/share/classes/sun/security/provider/certpath/URICertStore.java
jnimeh: Yes, I can make a follow on for that.

jnimeh: Filed https://bugs.openjdk.org/browse/JDK-8308601

seanjmullan (approved):
Looks good. I think there may be value in moving some of the test code into the testlibrary, like the AIA and CRL https servers so other tests can use it, but we can explore that more later if the opportunity arises.

PR #13767, opened 2023-05-03
8307139: Fix signed integer overflow in compiler code, part 1

dean-long:
These changes attempt to fix signed overflow caught by running tier1 with -ftrapv.  I expect more changes will be needed.
Most of the fixes are straight-forward and involve using unsigned or java_* functions that wrap.  However, I did try to improve the usefulness of _debug_idx because as it was the high digits of the value were monotonic but unpredictable.  Now the high digits use the compile_id, which seems like an improvement.

### DIFF
--- a/src/hotspot/share/c1/c1_Canonicalizer.cpp
+++ b/src/hotspot/share/c1/c1_Canonicalizer.cpp
@@ -89,9 +89,9 @@ void Canonicalizer::do_Op2(Op2* x) {
         { jint a = x->x()->type()->as_IntConstant()->value();
           jint b = x->y()->type()->as_IntConstant()->value();
           switch (x->op()) {
-            case Bytecodes::_iadd: set_constant(a + b); return;
-            case Bytecodes::_isub: set_constant(a - b); return;
-            case Bytecodes::_imul: set_constant(a * b); return;
+            case Bytecodes::_iadd: set_constant(java_add(a, b)); return;
+            case Bytecodes::_isub: set_constant(java_subtract(a, b)); return;
+            case Bytecodes::_imul: set_constant(java_multiply(a, b)); return;
             case Bytecodes::_idiv:
               if (b != 0) {
                 if (a == min_jint && b == -1) {
@@ -335,8 +335,8 @@ void Canonicalizer::do_NegateOp(NegateOp* x) {
   ValueType* t = x->x()->type();
   if (t->is_constant()) {
     switch (t->tag()) {
-      case intTag   : set_constant(-t->as_IntConstant   ()->value()); return;
-      case longTag  : set_constant(-t->as_LongConstant  ()->value()); return;
+      case intTag   : set_constant(java_negate(t->as_IntConstant   ()->value())); return;
+      case longTag  : set_constant(java_negate(t->as_LongConstant  ()->value())); return;
       case floatTag : set_constant(-t->as_FloatConstant ()->value()); return;
       case doubleTag: set_constant(-t->as_DoubleConstant()->value()); return;
       default       : ShouldNotReachHere();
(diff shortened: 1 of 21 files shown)


src/hotspot/share/opto/node.cpp
```diff
@@ -68,30 +68,17 @@ extern int nodes_created;
...
-    }
-    assert(bump >= 0 && bump < mod, "");
-    new_debug_idx += bump;
-  }
-  Compile::set_debug_idx(new_debug_idx);
-  set_debug_idx(new_debug_idx);
+  // The decimal digits of _debug_idx are <compile_id> followed by 5 digits of <_idx>
   Compile* C = Compile::current();
   assert(C->unique() < (INT_MAX - 1), "Node limit exceeded INT_MAX");
+  uintx new_debug_idx = (uintx)C->compile_id() * 100000 + _idx;
```
vnkozlov, line 74: Should we assert that _idx < 100000? We can use bigger multiplier since debug_idx is 64 bit value now.

dean-long: I can make the multiplier 10000000000 so we don't have to assert or artificially restrict _idx further than INT_MAX.  I also need to change the type from uintx to uint64_t so I don't break 32-bit ports.

src/hotspot/share/opto/idealGraphPrinter.cpp
```diff
@@ -377,9 +377,6 @@ void IdealGraphPrinter::visit_node(Node *n, bool edges, VectorSet* temp_set) {
     const Type *t = node->bottom_type();
     print_prop("type", t->msg());
     print_prop("idx", node->_idx);
-#ifdef ASSERT
-    print_prop("debug_idx", node->_debug_idx);
-#endif
```
vnkozlov, line 382: Why you removed this?

dean-long: print_prop() only works for int.  I could add an overload that works for uint64_t, but then I realized debug_idx is redundant for IGV, as we already have the compile_id and node _idx.

TobiHartmann (approved):
Looks reasonable to me. Another review would be good.

I'm actually wondering if anyone is using `BreakAtNode` or if we should simply deprecate/remove it.

dean-long:
> I'm actually wondering if anyone is using `BreakAtNode` or if we should simply deprecate/remove it.

I was wondering the same thing.

src/hotspot/share/c1/c1_Canonicalizer.cpp
```diff
@@ -335,8 +335,8 @@ void Canonicalizer::do_NegateOp(NegateOp* x) {
   ValueType* t = x->x()->type();
   if (t->is_constant()) {
     switch (t->tag()) {
-      case intTag   : set_constant(-t->as_IntConstant   ()->value()); return;
-      case longTag  : set_constant(-t->as_LongConstant  ()->value()); return;
+      case intTag   : set_constant(java_negate(t->as_IntConstant   ()->value())); return;
+      case longTag  : set_constant(java_negate(t->as_LongConstant  ()->value())); return;
```
robcasloz, line 339:
```suggestion
      case intTag   : set_constant(java_negate(t->as_IntConstant()->value())); return;
      case longTag  : set_constant(java_negate(t->as_LongConstant()->value())); return;
```

src/hotspot/share/opto/intrinsicnode.cpp
```diff
@@ -244,12 +244,12 @@ static const Type* bitshuffle_value(const TypeInteger* src_type, const TypeInteg
     int bitcount = population_count(static_cast<julong>(bt == T_INT ? maskcon & 0xFFFFFFFFL : maskcon));
     if (opc == Op_CompressBits) {
       // Bit compression selects the source bits corresponding to true mask bits
-      // and lays them out contiguously at desitination bit poistions starting from
+      // and lays them out contiguously at desitination bit positions starting from
```
robcasloz, line 247:
```suggestion
      // and lays them out contiguously at destination bit positions starting from
```

src/hotspot/share/opto/idealGraphPrinter.cpp
robcasloz: Removing `debug_idx` from IGV graph dumps makes sense to me as well.

robcasloz (approved):
Looks good!

dean-long:
Thanks Vladimir, Tobias, and Robert.
@vnkozlov, are you OK with these changes?

PR #13770, opened 2023-05-03
8307181: MemoryLayout.structLayout uses undocumented strict alignment constraints

mcimadamore:
This patch adds documentation for the behavior of the `MemoryLayout::structLayout` factory.

This factory throws an `IllegalArgumentException` if one of the member layouts passed to it occurs at an offset within the struct that is not aligned.

As part of this patch, I've also added some api notes on how to use this factory when unaligned elements are present.

---------
### Progress
- [x] Change requires CSR request [JDK-8307335](https://bugs.openjdk.org/browse/JDK-8307335) to be approved

### Issues
 * [JDK-8307181](https://bugs.openjdk.org/browse/JDK-8307181): MemoryLayout.structLayout uses undocumented strict alignment constraints
 * [JDK-8307335](https://bugs.openjdk.org/browse/JDK-8307335): MemoryLayout.structLayout uses undocumented strict alignment constraints (**CSR**)

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - **Reviewer**) ⚠️ Review applies to [592a3db6](https://git.openjdk.org/jdk/pull/13770/files/592a3db63b21cece89421dc673796b9cff84c472)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13770/head:pull/13770` \
`$ git checkout pull/13770`

`$ git checkout pull/13770` \
`$ git pull https://git.openjdk.org/jdk.git pull/13770/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13770`

`$ git pr show -t 13770`

</details>

<a href="https://git.openjdk.org/jdk/pull/13770.diff">https://git.openjdk.org/jdk/pull/13770.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13770#issuecomment-1532651691)

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
@@ -739,6 +739,28 @@ public sealed interface MemoryLayout permits SequenceLayout, GroupLayout, Paddin
      * @return a struct layout with the given member layouts.
      * @throws IllegalArgumentException if the sum of the {@linkplain #bitSize() bit sizes} of the member layouts
      * overflows.
+     * @throws IllegalArgumentException if a member layouts in {@code elements} occurs at an offset (relative to the start
+     * of the struct layout) which is not compatible with its alignment constraint.
+     *
+     * @apiNote This factory does not automatically align element layouts, by inserting additional {@linkplain PaddingLayout
+     * padding layout} elements. As such, the following struct layout creation will fail with an exception:
+     *
+     * {@snippet lang = java:
+     * structLayout(JAVA_SHORT, JAVA_INT)
+     * }
+     *
+     * To avoid the exception, clients can either insert additional padding layout elements:
+     *
+     * {@snippet lang = java:
+     * structLayout(JAVA_SHORT, MemoryLayout.ofPadding(16), JAVA_INT)
+     * }
+     *
+     * Or, alternatively, they can use a member layout which features a smaller alignment constraint. This will result
+     * in a <em>packed</em> struct layout:
+     *
+     * {@snippet lang = java:
+     * structLayout(JAVA_SHORT, JAVA_INT.withBitAlignment(16))
+     * }
      */
     static StructLayout structLayout(MemoryLayout... elements) {
         Objects.requireNonNull(elements);
--- a/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
@@ -739,7 +739,7 @@ public sealed interface MemoryLayout permits SequenceLayout, GroupLayout, Paddin
      * @return a struct layout with the given member layouts.
      * @throws IllegalArgumentException if the sum of the {@linkplain #bitSize() bit sizes} of the member layouts
      * overflows.
-     * @throws IllegalArgumentException if a member layouts in {@code elements} occurs at an offset (relative to the start
+     * @throws IllegalArgumentException if a member layout in {@code elements} occurs at an offset (relative to the start
      * of the struct layout) which is not compatible with its alignment constraint.
      *
      * @apiNote This factory does not automatically align element layouts, by inserting additional {@linkplain PaddingLayout


minborg:
Isn't this true also for sequence layouts where, for example, the second element might be misaligned for `sequenceLayout(2, structLayout(JAVA_INT, JAVA_SHORT))`? Or, is this already covered?

mcimadamore:
> Isn't this true also for sequence layouts where, for example, the second element might be misaligned for `sequenceLayout(2, structLayout(JAVA_INT, JAVA_SHORT))`? Or, is this already covered?

That case already has some javadoc text:

https://cr.openjdk.org/~pminborg/panama/21/v2/javadoc/api/java.base/java/lang/foreign/MemoryLayout.html#sequenceLayout(long,java.lang.foreign.MemoryLayout)

(both factories throw if the alignment of the element layout is bigger than its size). It seems like I only forgot to add extra doc for structs.

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -739,6 +739,28 @@ static SequenceLayout sequenceLayout(MemoryLayout elementLayout) {
      * @return a struct layout with the given member layouts.
      * @throws IllegalArgumentException if the sum of the {@linkplain #bitSize() bit sizes} of the member layouts
      * overflows.
+     * @throws IllegalArgumentException if a member layouts in {@code elements} occurs at an offset (relative to the start
+     * of the struct layout) which is not compatible with its alignment constraint.
```
JornVernee, line 743:
```suggestion
     * @throws IllegalArgumentException if a member layout in {@code elements} occurs at an offset (relative to the start
     * of the struct layout) which is not compatible with its alignment constraint.
```

mcimadamore:
> > Isn't this true also for sequence layouts where, for example, the second element might be misaligned for `sequenceLayout(2, structLayout(JAVA_INT, JAVA_SHORT))`? Or, is this already covered?
> 
> That case already has some javadoc text:
> 
> https://cr.openjdk.org/~pminborg/panama/21/v2/javadoc/api/java.base/java/lang/foreign/MemoryLayout.html#sequenceLayout(long,java.lang.foreign.MemoryLayout)
> 
> (both factories throw if the alignment of the element layout is bigger than its size). It seems like I only forgot to add extra doc for structs.

Actually, now that I look at it, it seems that the exception we raise is not correct. I think we need to throw if the element size is not a multiple of the alignment. That is, the size determines the stride, so that has to be a multiple of the alignment, otherwise it won't work.

JornVernee (approved):
Nice api note!

JornVernee:
> Actually, now that I look at it, it seems that the exception we raise is not correct. I think we need to throw if the element size is not a multiple of the alignment. That is, the size determines the stride, so that has to be a multiple of the alignment, otherwise it won't work.

Hmm, I guess this doesn't work out if we have an element layout which has a size that is not a multiple of its alignment constraint (i.e. it lacks trailing padding). For instance a struct of size 12, with alignment 8. The second element of a sequence with that struct layout would not be aligned.

I think we went back-and-forth on the need to enforce trailing padding in the `structLayout` factory a few times.

mcimadamore:
> > Actually, now that I look at it, it seems that the exception we raise is not correct. I think we need to throw if the element size is not a multiple of the alignment. That is, the size determines the stride, so that has to be a multiple of the alignment, otherwise it won't work.
> 
> Hmm, I guess the current check/assertion doesn't work out if we have an element layout which has a size that is not a multiple of its alignment constraint (i.e. it lacks trailing padding). For instance a struct of size 12, with alignment 8. The second element of a sequence with that struct layout would not be aligned.
> 
> I think we went back-and-forth on the need to enforce trailing padding in the `structLayout` factory a few times.

I think the rationale is: we can't enforce, in general, that a layout size is a multiple of its alignment. Because, for instance, we might want to construct "hyper-aligned" layouts (e.g. a struct layout aligned at page boundary. That's fine.

But I think when you use a layout as a sequence element, then we should check that the thing makes sense - e.g. that the element layout has a size that is indeed a multiple of the alignment. Otherwise it just seems like a "bad" sequence.

mcimadamore:
> Actually, now that I look at it, it seems that the exception we raise is not correct. I think we need to throw if the element size is not a multiple of the alignment. That is, the size determines the stride, so that has to be a multiple of the alignment, otherwise it won't work.

If that's ok, I'd like to address this other issue in a separate PR/CSR, for clarity.

PR #13771, opened 2023-05-03
8307058: Implementation of Generational ZGC

stefank:
Hi all,

Please review the implementation of Generational ZGC, which can be turned on by adding -XX:+ZGenerational in addition to using -XX:+UseZGC. Generational ZGC is a major rewrite of the non-generational ZGC version that exists in the openjdk/jdk repository. It splits the heap into two generations; the young generation where newly allocated objects are born, and the old generation where long-lived objects get promoted to. The motivation for introducing generations is to allow ZGC to reclaim memory faster by not having to walk the entire object graph every time a garbage collection is run. This should make Generational ZGC suitable for more workloads. In particular workloads that previously hit allocation stalls because of high allocation rates, large live sets, or limited spare machine resources, have the potential to work better with Generational ZGC. For an in-depth description of Generational ZGC, see https://openjdk.org/jeps/439.

The development of Generational ZGC started around the same time as the development of JDK 17. At that point we forked off the Generational ZGC development into its own branch and let non-generational live unaffected in openjdk/jdk. This safe-guarded non-generational ZGC and allowed Generational ZGC to move unhindered, without the shackles of having to fit into another GC implementation's design and quirks. Since then, almost all of the ZGC files have been changed. Moving forward to today, when it's ready for us to upstream Generational ZGC, we now need to deliver Generational ZGC without disrupting our current user-base. We have therefore opted to initially include both versions of ZGC in the code base, but with the intention to deprecate non-generational ZGC in a future release. Existing users running with only -XX:+UseZGC will get the non-generational ZGC, and users that want the new Generational ZGC need to run with -XX:+ZGenerational in addition to -XX:+UseZGC. The intention is to give the users time to validate and deploy their workloads with the new GC implementation.

Including both the new evolution of a GC and its legacy predecessor poses a few challenges for us GC developers. The first reaction could be to try to mash the two implementations together and sprinkle the GC code with conditional statements or dynamic dispatches. We have done similar experiments before. When ZGC was first born, we started an experiment where we converted G1 into getting the same features as the evolving ZGC. It was quite clear to us how time consuming and complex things end up being when we tried to keep both the original G1 working, and at the same time implemented the ZGC-alike G1. Given this experience, we don't see that as a viable solution to deliver a maintainable and evolving Generational ZGC. Our pragmatic suggestion to these challenges is to let Generational ZGC live under the current gc/z directories and let the legacy, non-generational ZGC be completely separated in its own directories. This way we can continue to move quickly with the continued development of Generational ZGC and let the non-generational ZGC be mostly untouched until it gets deprecated, and eventually removed. The non-generational ZGC directory will be gc/x and all the classes of non-generational have been prefixed with X instead of Z. An alternative to this rename could be to namespace out non-generational ZGC. We experimented with that, but it was too easy to accidentally cross-compile Generational ZGC code into non-generational ZGC, so we didn't like that approach.

Most of the stand-alone cleanups and enhancements outside of the ZGC code have already been upstreamed to openjdk/jdk. There are still a few patches that could/should be pushed separately, but they will be easier to understand by also looking at the Generational ZGC code, so they will be sent out after this PR has been published. The patches that could be published separately are:

* 59d1e96af6a UPSTREAM: Introduce check_oop infrastructure to check oops in the oop class
* ca9edf8aa79 UPSTREAM: RISCV tmp reg cleanup resolve_jobject
* 4bec9c69b67 CLEANUP: barrierSetNMethod_aarch64.cpp
* b67d03a3f04 UPSTREAM: Add relaxed add&fetch for aarch64 atomics
* a2824734d23 UPSTREAM: lir_xchg
* 36cd39c0126 UPSTREAM: assembler_ppc CMPLI
* 447259cea42 UPSTREAM: assembler_ppc ANDI
* 9417323499a UPSTREAM: Add VMErrorCallback infrastructure 

Regarding all the changesets you see in this PR, they form the history of the development of Generational ZGC. It might look a bit unconventional to what you are used to see in openjdk development. What we have done is to use merges with the 'ours' strategy to ignore the previous Generational ZGC patches, and then rebased and flattened the changes on top of the merge. This effectively gives us the upsides of having a rebased repository and the upsides of retaining the history in the repository. The downside could be that GitHub now lists all those changesets in the PR. Given that this patch is so big, and that you likely only want to see a part of it, I suggest that you pull down the PR branch and then compare it to the openjdk/jdk changeset this PR is based against:

git fetch https://github.com/openjdk/zgc zgc_master
git diff zgc_master... <sub-directory of interest>

There have been many contributors of this patch over the years. I'll do my best to poke Skara into listing you all, but if you see that I've missed your name please reach out to me and I'll fix it.

Testing: we have been continuously running Generational ZGC through Oracle's tier1-8 testing.


fisk (approved):
I have obviously stared at this code since its inception. To me it doesn't just look good, it looks fantastic.

stefank:
> Thanks for fixing PPC64! With this, the VM compiles and the `test/hotspot/jtreg/gc` tests are passing on linux PPC64le.
> 
> I'm glad to see this PR for JDK 21 LTS. It's a big step forward for ZGC. Congratulations!

Thanks for porting Generational ZGC to PPC!

TheRealMDoerr:
"test/hotspot/jtreg/gc" and "test/hotspot/jtreg/compiler/gcbarriers" are also passing with JTREG="VM_OPTIONS=-XX:+UseZGC -XX:+ZGenerational" on linux PPC64 le.
I've quickly checked Spec JBB 2005 with ZGC performance. Generational mode was about 7% faster on Power10.

test/hotspot/jtreg/ProblemList-generational-zgc.txt
```diff
@@ -0,0 +1,117 @@
...
+
+#############################################################################
+#
+# List of quarantined tests for testing with Generational ZGC.
+#
+#############################################################################
+
+# Quiet all SA tests
+
+resourcehogs/serviceability/sa/TestHeapDumpForLargeArray.java 8000000   generic-all
```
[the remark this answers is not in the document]
stefank: Sounds like a good idea. I've created JDK-8307393 and will update the problem list.

test/jdk/com/sun/jdi/ThreadMemoryLeakTest.java
```diff
@@ -27,7 +27,7 @@
  * @summary JDI memory leak when creating and destroying many threads
  *
  * @comment Don't allow -Xcomp or -Xint as they impact memory useage and number of iterations
- * @requires (vm.compMode == "Xmixed")
+ * @requires (vm.compMode == "Xmixed") & !(vm.gc.Z & vm.opt.final.ZGenerational)
```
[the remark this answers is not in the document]
plummercj: When I first wrote this test, it ended up failing with ZGC because I hadn't tested it. I considered excluding it for the same reason you've given, but then considered that the test might expose a leak with one GC, but not others, so I decided to fix it.

test/jdk/ProblemList-generational-zgc.txt
```diff
@@ -0,0 +1,30 @@
...
+#
+# Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+# or visit www.oracle.com if you need additional information or have any
+# questions.
+#
+
+#############################################################################
+#
+# List of quarantined tests for testing with Generational ZGC.
+#
```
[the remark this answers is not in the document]
stefank: It seems like these tests are only run with all GCs at the end of the development cycle. I've run them manually and verified that these tests fail as well. I'm going to problem list them.

That run also revealed that jstat doesn't like when we report the initial capacity of the old generation as zero. See the calculation in:
src/jdk.jcmd/share/classes/sun/tools/jstat/resources/jstat_options
```
  column {
    header "^O^"        /* Old Space - Percent Used */
    data (1-((sun.gc.generation.1.space.0.capacity - sun.gc.generation.1.space.0.used)/sun.gc.generation.1.space.0.capacity)) * 100
    align right
    scale raw
    width 6
    format "0.00"
  }
```

I can work around the test problem by faking the capacity to be non-zero, but that's not a pretty solution IMO.

test/jdk/com/sun/jdi/ThreadMemoryLeakTest.java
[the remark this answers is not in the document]
stefank: I've created JDK-8307402. I'll push a problem list entry.

TheRealMDoerr:
I'm getting build warnings on all linux platforms with gcc-11.3.0:
``` 
src/hotspot/share/gc/z/zDriver.cpp:84:13: error: In the GNU C Library, "minor" is defined
 by <sys/sysmacros.h>. For historical compatibility, it is
 currently defined by <sys/types.h> as well, but we plan to
 remove this soon. To use "minor", include <sys/sysmacros.h>
 directly. If you did not intend to use a system-defined macro
 "minor", you should undefine it after including <sys/types.h>. [-Werror]
   84 | ZDriverMinor* ZDriver::minor() {
```

stefank:
> I'm getting build warnings on all linux platforms with gcc-11.3.0:
> 
> ```
> src/hotspot/share/gc/z/zDriver.cpp:84:13: error: In the GNU C Library, "minor" is defined
>  by <sys/sysmacros.h>. For historical compatibility, it is
>  currently defined by <sys/types.h> as well, but we plan to
>  remove this soon. To use "minor", include <sys/sysmacros.h>
>  directly. If you did not intend to use a system-defined macro
>  "minor", you should undefine it after including <sys/types.h>. [-Werror]
>    84 | ZDriverMinor* ZDriver::minor() {
> ```

That's unfortunate as minor and major are quite central to Generational ZGC and having to rename those functions will make the code look worse. I wonder if we should undef minor and major where needed.

test/jdk/ProblemList-generational-zgc.txt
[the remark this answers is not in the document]
stefank: The jhsdb tests have been ProblemListed. The jstat test is going to be fixed with #13796.

xmas92:
> I'm getting build warnings on all linux platforms with gcc-11.3.0:
> 
> ```
> src/hotspot/share/gc/z/zDriver.cpp:84:13: error: In the GNU C Library, "minor" is defined
>  by <sys/sysmacros.h>. For historical compatibility, it is
>  currently defined by <sys/types.h> as well, but we plan to
>  remove this soon. To use "minor", include <sys/sysmacros.h>
>  directly. If you did not intend to use a system-defined macro
>  "minor", you should undefine it after including <sys/types.h>. [-Werror]
>    84 | ZDriverMinor* ZDriver::minor() {
> ```

@TheRealMDoerr I cannot reproduce this with gcc but can see the issue with clangd.
Can you check if this patch solves the issue you are seeing?
```diff
diff --git a/src/hotspot/share/gc/z/zDriver.hpp b/src/hotspot/share/gc/z/zDriver.hpp
index 640ea6575ef..7fa650b1fa1 100644
--- a/src/hotspot/share/gc/z/zDriver.hpp
+++ b/src/hotspot/share/gc/z/zDriver.hpp
@@ -29,6 +29,14 @@
 #include "gc/z/zThread.hpp"
 #include "gc/z/zTracer.hpp"
 
+#ifdef minor
+#undef minor
+#endif
+
+#ifdef major
+#undef major
+#endif
+
 class VM_ZOperation;
 class ZDriverMinor;
 class ZDriverMajor;
```

TheRealMDoerr:
@xmas92: Thanks for your quick solution. Your patch solves the problem. If you want to integrate it, please also add a comment why this is needed.

xmas92:
> @xmas92: Thanks for your quick solution. Your patch solves the problem. If you want to integrate it, please also add a comment why this is needed.

Thanks for testing it. Will do.

test/hotspot/jtreg/runtime/stringtable/StringTableCleaningTest.java
```diff
@@ -108,7 +114,7 @@ private static String getGcEndString() {
         } else if (GC.G1.isSelected()) {
             return gcEndPrefix + g1Suffix;
         } else if (GC.Z.isSelected()) {
-            return gcEndPrefix + zEndSuffix;
+            return gcEndPrefix + "(" + zEndSuffix +  ")|(" + xEndSuffix + ")";
```
turbanoff, line 117: nit
```suggestion
            return gcEndPrefix + "(" + zEndSuffix + ")|(" + xEndSuffix + ")";
```

test/hotspot/gtest/gc/z/test_zForwarding.cpp
```diff
@@ -39,6 +42,42 @@ using namespace testing;
...
+    ZGeneration::_young = &ZHeap::_heap->_young;
+
+    *const_cast<ZGenerationId*>(&ZGeneration::_old->_id) = ZGenerationId::old;
+    *const_cast<ZGenerationId*>(&ZGeneration::_young->_id) = ZGenerationId::young;
+
+    ZGeneration::_old->_seqnum = 1;
+    ZGeneration::_young->_seqnum = 2;
+
+    bool reserved = os::attempt_reserve_memory_at((char*)ZAddressHeapBase, ZGranuleSize, false /* executable */);
+    ASSERT_TRUE(reserved);
```
RealFYang, line 68: Hi, Thanks for the great work!
I have performed some tests on linux-riscv64 Hifive Unmatched board. So far, I only witnessed one gtest failure:

```
$ make test TEST=gtest:ZForwardingTest
Building target 'test' in configuration 'linux-riscv64-server-release'
Test selection 'gtest:ZForwardingTest', will run:
* gtest:ZForwardingTest/server

Running test 'gtest:ZForwardingTest/server'
Note: Google Test filter = ZForwardingTest*
[==========] Running 4 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 4 tests from ZForwardingTest
[ RUN      ] ZForwardingTest.setup_vm
test/hotspot/gtest/gc/z/test_zForwarding.cpp:68: Failure
Value of: reserved
  Actual: false
Expected: true
[  FAILED  ] ZForwardingTest.setup_vm (0 ms)
[ RUN      ] ZForwardingTest.find_empty_vm
[       OK ] ZForwardingTest.find_empty_vm (1 ms)
[ RUN      ] ZForwardingTest.find_full_vm
[       OK ] ZForwardingTest.find_full_vm (8 ms)
[ RUN      ] ZForwardingTest.find_every_other_vm
[       OK ] ZForwardingTest.find_every_other_vm (0 ms)
[----------] 4 tests from ZForwardingTest (761 ms total)

[----------] Global test environment tear-down
ERROR: RUN_ALL_TESTS() failed. Error 1
[==========] 4 tests from 1 test suite ran. (762 ms total)
[  PASSED  ] 3 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] ZForwardingTest.setup_vm

 1 FAILED TEST
Finished running test 'gtest:ZForwardingTest/server'
Test report is stored in build/linux-riscv64-server-release/test-results/gtest_ZForwardingTest_server

==============================
Test summary
==============================
   TEST                                              TOTAL  PASS  FAIL ERROR
>> gtest:ZForwardingTest/server                          4     3     1     0 <<
==============================
TEST FAILURE
```

The gtest failed this assertion where 'reserved' return by function os::attempt_reserve_memory_at is false.
I find the reason is that the mmap call at the bottom returns a different address instead of the requested one (ZAddressHeapBase). I think that is possible since we are not sure if the requested address is available before the mmap call, right? So I guess we might need some changes here for this gtest.

test/hotspot/jtreg/runtime/stringtable/StringTableCleaningTest.java
stefank: Thanks!

test/hotspot/gtest/gc/z/test_zForwarding.cpp
stefank: Thanks for reporting. It would be interesting to see what address you get and compare it to the range [ZAddressHeapBase, ZAddressHeapBase+ZAddressOffsetMax).

yadongw: We emailed to erik to discuss this issue two months ago, and maybe he missed it.
ZForwardingTest does not guarantee a successful invoke of os::commit_memory for ZAddressHeapBase, and we saw some conflicts between ZAddressHeapBase and the metadata address space on the RISC-V hardware of 39-bits virtual address. There is no failure in the normal initialization phase of JVM, because the commit order of them is guaranteed.

stefank: Could you provide the values for `reserved`, `ZAddressHeapBase`, and `ZAddressOffsetMax, when this test is failing. I'd like to know if we can make a workaround for you, or if we have to turn off the test for riscv.

src/hotspot/cpu/x86/gc/z/zBarrierSetAssembler_x86.cpp
```diff
@@ -186,7 +301,208 @@ void ZBarrierSetAssembler::load_at(MacroAssembler* masm,
...
+    __ push(rax);
+    __ push(rbx);
+    __ push(rcx);
+
+    __ lea(rcx, ref_addr);
+    __ xorq(rax, rax);
+    __ movptr(rbx, Address(r15, ZThreadLocalData::store_good_mask_offset()));
+
+    __ lock();
+    __ cmpxchgq(rbx, Address(rcx, 0));
```
merykitty, line 483: `ref_addr` is not necessarily materialised here?

src/hotspot/cpu/x86/gc/z/zBarrierSetAssembler_x86.cpp
```diff
@@ -186,7 +301,208 @@ void ZBarrierSetAssembler::load_at(MacroAssembler* masm,
...
 }
 
-#ifdef ASSERT
+static void emit_store_fast_path_check(MacroAssembler* masm, Address ref_addr, bool is_atomic, Label& medium_path) {
+  if (is_atomic) {
+    // Atomic operations must ensure that the contents of memory are store-good before
+    // an atomic operation can execute.
+    // A not relocatable object could have spurious raw null pointers in its fields after
+    // getting promoted to the old generation.
+    __ cmpw(ref_addr, barrier_Relocation::unpatched);
```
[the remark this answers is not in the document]
merykitty: I think we use the flag `UseStoreImmI16` for these kinds of situations.

src/hotspot/cpu/x86/gc/z/zBarrierSetAssembler_x86.cpp
fisk: I think it is, yes. But we want to ensure it's in a register that isn't rbx or rax. So I figured I'd just force materialize it in rcx and call it a day. It might be possible to micro optimize this further and even use the live information we have gathered to eliminate some of the spilling, but I'd like to hold off on that until we integrate. It's again only for atomics, and also happens at most once per field.

test/hotspot/gtest/gc/z/test_zForwarding.cpp
RealFYang: @stefank : I ran this gtest for 5 times and here is what I got.
```
ZAddressHeapBase :     0x800000000
ZAddressOffsetMax:     0x800000000
ZGranuleSize          :     0x200000
```
In os::pd_attempt_reserve_memory_at() which is called by os::attempt_reserve_memory_at(), return value by anon_mmap() [1] is one of: 0x3f8d5ff000, 0x3f649fe000, 0x3f5d3ff000, 0x3f68077000 and 0x3f555ff000

So seems that those values are not in the range [ZAddressHeapBase, ZAddressHeapBase+ZAddressOffsetMax).

[1] https://github.com/openjdk/jdk/blob/master/src/hotspot/os/linux/os_linux.cpp#L3334

src/hotspot/cpu/x86/gc/z/zBarrierSetAssembler_x86.cpp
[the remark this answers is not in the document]
merykitty: @fisk Thanks a lot for your explanations.

test/hotspot/gtest/gc/z/test_zForwarding.cpp
stefank: That's unfortunate. Could you try this patch, which probes the address range to see if it can reserve the memory somewhere else within `[ZAddressHeapBase, ZAddressHeapBase+ZAddressOffsetMax)`:
https://github.com/stefank/jdk/tree/zgc_generational_review_test_zforwarding

RealFYang: > That's unfortunate. Could you try this patch, which probes the address range to see if it can reserve the memory somewhere else within `[ZAddressHeapBase, ZAddressHeapBase+ZAddressOffsetMax)`: https://github.com/stefank/jdk/tree/zgc_generational_review_test_zforwarding

@stefank : Good news is that this gtest case can now pass on linux-riscv64 with this patch. I tried several times and it seems we could always reserve a space of ZGranuleSize bytes at address 0x852000000. Thanks for fixing this!

stefank: Thanks for helping out with this. I've now pushed the proposed patch.

src/hotspot/share/code/relocInfo.hpp
```diff
@@ -1078,6 +1081,31 @@ class metadata_Relocation : public DataRelocation {
...
+
+ private:
+  friend class RelocIterator;
+  friend class RelocationHolder;
+  barrier_Relocation() : DataRelocation(relocInfo::barrier_type) { }
+
+ public:
+  int offset() override                  { ShouldNotReachHere(); return 0; }
+  address value() override               { ShouldNotReachHere(); return nullptr; }
+  void set_value(address value) override { ShouldNotReachHere(); }
```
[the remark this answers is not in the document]
fisk: That is a good question. I think we used to use Relocation:: pd_address_in_code, which on x86 asserts that it has to be a DataRelocation. But it seems like we are not using that any more and it just looks weird. I will remove this and inherit from Relocation instead.

stefank:
FYI: GitHub doesn't seem to handle our new merges correctly and lists changes that we have recently merged in from openjdk/jdk. Make sure to look at the patches locally or use the webrevs above.

xmas92 (approved):
Having been able to contribute to generational ZGC over the last year has been a privilege. 

The code is well structured and the preparatory upstreaming work has made the integration with other subcomponents of hotspot rather minimal. 

The X/Z split feels like a pragmatic solution for having the two ZGC codebases coexisting. 

Just like Erik, I am a little biased. I approve of this PR!

robcasloz (approved):
Compiler-related changes look good!

stefank:
Thanks all who have helped building Generational ZGC! I think it is time to integrate.
/integrate

PR #13773, opened 2023-05-03
8307236: Rendezvous GC threads under STS for monitor deflation

rkennke:
Object monitors are deflated concurrently by the MonitorDeflationThread. It first unlinks monitors from objects (i.e. restore the original object header), then handshakes (with a no-op) all Java threads, and only then destroys the monitors. This way, Java threads can safely (and racily) access monitors before the handshake, because the monitors are guaranteed to still exist when a Java thread racily reads a mark-word that is being unlinked, and the monitor can safely be destroyed after the handshake, because all Java threads would then read the correct unlinked mark-word.

However, GC threads are not rendezvous'ed like that, and can read potentially dead monitors. At least with the upcoming Compact Object Headers this is going to be a problem, because the compressed Klass* is then part of the mark-word.

In order to safely access monitors via object headers concurrently from GC threads, we need to rendezvous them after unlinking and before destroying the monitors, just like Java threads do, via handshake. This is important so that concurrent GCs (ZGC, Shenandoah, G1) can safely access object's Klass* (and thus object size, layout, etc) during concurrent GC phases.

This only implements the parts that do the rendezvous, it still requires that affected concurrent GC threads are under SustainableThreadSet (most of them already are!). This will be implemented in later PR.

Testing:
 - [x] tier1
 - [x] tier2

---------
### Progress

### Issue
 * [JDK-8307236](https://bugs.openjdk.org/browse/JDK-8307236): Rendezvous GC threads under STS for monitor deflation

### Reviewers
 * [Erik Österlund](https://openjdk.org/census#eosterlund) (@fisk - **Reviewer**) ⚠️ Review applies to [68357ae2](https://git.openjdk.org/jdk/pull/13773/files/68357ae2c0af357f82cc982486682eebf1a823ce)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13773/head:pull/13773` \
`$ git checkout pull/13773`

`$ git checkout pull/13773` \
`$ git pull https://git.openjdk.org/jdk.git pull/13773/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13773`

`$ git pr show -t 13773`

</details>

<a href="https://git.openjdk.org/jdk/pull/13773.diff">https://git.openjdk.org/jdk/pull/13773.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13773#issuecomment-1533006215)

### DIFF
--- a/src/hotspot/share/runtime/monitorDeflationThread.hpp
+++ b/src/hotspot/share/runtime/monitorDeflationThread.hpp
@@ -41,6 +41,8 @@ class MonitorDeflationThread : public JavaThread {
 
   // Hide this thread from external view.
   bool is_hidden_from_external_view() const { return true; }
+
+  bool is_monitor_deflation_thread() const override  { return true; }
 };
 
 #endif // SHARE_RUNTIME_MONITORDEFLATIONTHREAD_HPP
--- a/src/hotspot/share/runtime/monitorDeflationThread.hpp
+++ b/src/hotspot/share/runtime/monitorDeflationThread.hpp
@@ -42,7 +42,7 @@ class MonitorDeflationThread : public JavaThread {
   // Hide this thread from external view.
   bool is_hidden_from_external_view() const { return true; }
 
-  bool is_monitor_deflation_thread() const override  { return true; }
+  bool is_monitor_deflation_thread() const { return true; }
 };
 
 #endif // SHARE_RUNTIME_MONITORDEFLATIONTHREAD_HPP
--- a/src/hotspot/share/runtime/synchronizer.cpp
+++ b/src/hotspot/share/runtime/synchronizer.cpp
@@ -1552,7 +1552,9 @@ size_t ObjectSynchronizer::deflate_idle_monitors(ObjectMonitorsHashtable* table)
         timer.start();
       }
     } else {
-      assert(current->is_VM_thread(), "only VM thread may get here");
+      // This is not a monitor deflation thread.
+      // No handshake or rendezvous is needed when we are already at safepoint.
+      assert_at_safepoint();
     }
 
     // After the handshake, safely free the ObjectMonitors that were
--- a/src/hotspot/share/runtime/synchronizer.cpp
+++ b/src/hotspot/share/runtime/synchronizer.cpp
@@ -24,6 +24,7 @@
 
 #include "precompiled.hpp"
 #include "classfile/vmSymbols.hpp"
+#include "gc/shared/suspendibleThreadSet.hpp"
 #include "jfr/jfrEvents.hpp"
 #include "logging/log.hpp"
 #include "logging/logStream.hpp"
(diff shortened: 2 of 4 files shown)


fisk (approved):
Looks good.

src/hotspot/share/runtime/synchronizer.cpp
```diff
@@ -1528,6 +1539,11 @@ size_t ObjectSynchronizer::deflate_idle_monitors(ObjectMonitorsHashtable* table)
       // ObjectMonitors that were deflated in this cycle.
       HandshakeForDeflation hfd_hc;
       Handshake::execute(&hfd_hc);
+      // Also, we sync and desync GC threads around the handshake, so that they can
+      // safely read the mark-word and look-through to the object-monitor, without
+      // being afraid that the object-monitor is going away.
+      VM_RendezvousGCThreads sync_gc;
+      VMThread::execute(&sync_gc);
```
shipilev, line 1546: Not sure why this block is under `current->is_Java_thread()` block. Shouldn't we rendezvous with GC threads regardless?

fisk: > Not sure why this block is under `current->is_Java_thread()` block. Shouldn't we rendezvous with GC threads regardless?

Deflation is done either concurrently by the service thread, or in a safepoint by the VM thread. The current->is_Java_thread() means it's the concurrent case from the service thread. That's when it's needed. Inside of safepoints, the GC threads have already been synchronized.

shipilev: Okay, I see. Can we then move this block into separate `current->is_service_thread()`-checked block, and `assert(current->is_VM_thread())` on `else` branch? This would make this distinction very clear.

src/hotspot/share/runtime/synchronizer.cpp
```diff
@@ -1528,13 +1539,20 @@ size_t ObjectSynchronizer::deflate_idle_monitors(ObjectMonitorsHashtable* table)
...
 
       if (ls != nullptr) {
         ls->print_cr("after handshaking: in_use_list stats: ceiling="
                      SIZE_FORMAT ", count=" SIZE_FORMAT ", max=" SIZE_FORMAT,
                      in_use_list_ceiling(), _in_use_list.count(), _in_use_list.max());
         timer.start();
       }
+    } else {
+      assert(current->is_VM_thread(), "only VM thread may get here");
```
shipilev, line 1555: Actually, maybe we should assert we are at safepoint? To protect the accidental call from the VMOp that is executed by VMThread, but outside of safepoint.

```
} else {
  // This is not a monitor deflation thread.
  // No handshake or rendezvous is needed when we are already at safepoint.
  assert_at_safepoint();
}
```

shipilev (approved):
This looks fine to me. @fisk might want to re-review :)

fisk (approved):

src/hotspot/share/runtime/monitorDeflationThread.hpp
```diff
@@ -41,6 +41,8 @@ class MonitorDeflationThread : public JavaThread {
 
   // Hide this thread from external view.
   bool is_hidden_from_external_view() const { return true; }
+
+  bool is_monitor_deflation_thread() const override  { return true; }
```
shipilev, line 45: Don't say "override" here, because clang would then trigger lots of warnings for methods that were not marked as such. Yes, it is unfortunate, but...

shipilev (approved):

PR #13774, opened 2023-05-03
8307348 - Parallelize heap walk for ObjectCount(AfterGC) JFR event collection

olivergillespie:
ObjectCount(AfterGC) event does a full single-threaded heap scan at a safepoint. After https://bugs.openjdk.org/browse/JDK-8215624, it is trivial to use the parallel version of the heap scan, reducing the time spent at the safepoint, and thus reducing the overhead of this event.

The performance improvement is obvious, but just for confirmation, on my 16-core host, at around 1GB occupancy:

```
Before: 770ms ( [3.059s][debug][gc,phases         ] GC(13) Report Object Count 770.317ms )
 After:  92ms ( [2.335s][debug][gc,phases         ] GC(13) Report Object Count  91.742ms )
```

Question 1: Should this be the default behaviour for populate_table (use the number active workers as the parallelism, if nothing else specified)?

Question 2: Is active_workers the correct value to use here? Or is max_workers more appropriate?

### DIFF
--- a/src/hotspot/share/gc/shared/gcTrace.cpp
+++ b/src/hotspot/share/gc/shared/gcTrace.cpp
@@ -109,7 +109,12 @@ void GCTracer::report_object_count_after_gc(BoolObjectClosure* is_alive_cl) {
     KlassInfoTable cit(false);
     if (!cit.allocation_failed()) {
       HeapInspection hi;
-      hi.populate_table(&cit, is_alive_cl);
+      uint parallel_threads = 1;
+      WorkerThreads* workers = Universe::heap()->safepoint_workers();
+      if (workers != nullptr) {
+        parallel_threads = workers->active_workers();
+      }
+      hi.populate_table(&cit, is_alive_cl, parallel_threads);
       ObjectCountEventSenderClosure event_sender(cit.size_of_instances_in_words(), Ticks::now());
       cit.iterate(&event_sender);
     }


shipilev:
Why not just `hi.populate_table(&cit, is_alive_cl, ParallelGCThreads);`, and let the `populate_table` deal with the rest? I think we have a convention that `ParallelGCThreads` is roughly the proxy for the number of GC threads at paused operation.

(It is weird that `HeapInspection::populate_table` uses `safepoint_workers` -- maybe that's for additional isolation from the GC threads -- let's not proliferate it here. `populate_table` also caps the worker count at `max_workers`, which answers one of your questions.

olivergillespie:
> Why not just `hi.populate_table(&cit, is_alive_cl, ParallelGCThreads);`, and let the `populate_table` deal with the rest? I think we have a convention that `ParallelGCThreads` is roughly the proxy for the number of GC threads at paused operation.
> 
> (It is weird that `HeapInspection::populate_table` uses `safepoint_workers` -- maybe that's for additional isolation from the GC threads -- let's not proliferate it here. `populate_table` also caps the worker count at `max_workers`, which answers one of your questions)

Thanks, that's fine by me, whatever is most idiomatic. Updated.

shipilev (approved):
I looked if there might be a better option, like passing the `WorkerThreads*` from callers to actually figure out the number of active workers from the GC itself, but all of this cuts rather deep. I would say that should be done in a separate PR. `ParallelGCThreads` should work well meanwhile.

@albertnetymk, @tschatzl might have an opinion here.

olivergillespie:
Passed tier2 and tier3 tests for me locally on Linux x86_64 and aarch64, with G1, Serial and Parallel, with just a few (14) tier2 failures in `java/net/httpclient/` which I believe are due to a bad configuration in my local setup.

tschatzl:
> I looked if there might be a better option, like passing the `WorkerThreads*` from callers to actually figure out the number of active workers from the GC itself, but all of this cuts rather deep. I would say that should be done in a separate PR. `ParallelGCThreads` should work well meanwhile.
> 
> @albertnetymk, @tschatzl might have an opinion here.

As you suggested, the typical way is passing the `WorkerThreads*` along instead of passing a thread number and the code selecting the `WorkerThread*` by itself. Actually I'm not sure why safepoint_workers are used at all.

The changes to do that does not seem to be that significant to me actually, so I would prefer that. The current change is an improvement already though.

albertnetymk:
> Actually I'm not sure why safepoint_workers are used at all.

I believe it's semantically incorrect to use `safepoint_workers` here. Maybe `HeapInspection` should live in `gc` folder.

> The changes to do that does not seem to be that significant to me actually, so I would prefer that.

How does that affect another caller, `VM_GC_HeapInspection::doit`? Unclear to me how one can get gc-workers in that context.

Could one introduce another API in `class CollectedHeap`, sth like `virtual WorkerThreads* gc_workers() { return nullptr; }`, next to `safepoint_workers`?

shipilev:
> How does that affect another caller, `VM_GC_HeapInspection::doit`? Unclear to me how one can get gc-workers in that context.

I think `safepoint_workers` is "fine" to use from `VM_GC_HeapInspection`. That VMOp does not know the status of GC threads when executing the VM op. The GC threads might be currently parked at STS and not available. `safepoint_workers` are supposed to be separate from normal (concurrent) GC workers then. 

So, the contract for `safepoint_workers` is relatively safe for doing this the object count heap walk too. I agree it would be cleaner to use the GC workers, though. The problem that I see there is that `report_object_count_after_gc` is called from rather generic places like `GenMarkSweep::mark_sweep_phase1`, where we don't know how many workers do we have (or even which exact GC we are running). That would probably require lifting `workers()` straight to `CollectedHeap`.

tschatzl:
> > How does that affect another caller, `VM_GC_HeapInspection::doit`? Unclear to me how one can get gc-workers in that context.
> 
> I think `safepoint_workers` is "fine" to use from `VM_GC_HeapInspection`. That VMOp does not know the status of GC threads when executing the VM op. The GC threads might be currently parked at STS and not available. `safepoint_workers` are supposed to be separate from normal (concurrent) GC workers then.

I agree.

> 
> So, the contract for `safepoint_workers` is relatively safe for doing this the object count heap walk too. I agree it would be cleaner to use the GC workers, though. The problem that I see there is that `report_object_count_after_gc` is called from rather generic places like `GenMarkSweep::mark_sweep_phase1`, where we don't know how many workers do we have (or even which exact GC we are running). That would probably require lifting `workers()` straight to `CollectedHeap`.

Can you elaborate about this? From what I can see, the callers of `report_object_count_after_gc` other than heap inspection always have GC workers available. As for `GenMarkSweep::mark_sweep_phase1` specifically - this is serial gc. There are no workers for it, and it's not supposed to use multiple threads. One would just pass `nullptr` here I assume.

  Thomas

shipilev (changes requested):
[changing my review back to "Request Changes"]

albertnetymk:
> I think safepoint_workers is "fine" to use from VM_GC_HeapInspection.

I see; the name contains "GC", but it doesn't really correspond to a gc-safepoint (young/full gc pause). (To make matters worse, that vm-op can potentially perform a full-gc...)

Then, `populate_table` will get diff workers, depending whether it's called inside a gc-safepoint or not.

olivergillespie:
Thanks for the comments. Please check the updated version, it updates to have each caller pass the appropriate workers. The VMOp now grabs (and activates, as necessary) the safepoint workers instead of delegating that work, and populate can simply use the workers->active_workers as passed in (or run serially, as appropriate).

This is now a slightly more substantial change - not only are we parallelizing, but we're moving the GC-based ObjectCountAfterGC away from safepoint workers.

src/hotspot/share/gc/shared/gcVMOperations.cpp
```diff
@@ -166,8 +166,15 @@ void VM_GC_HeapInspection::doit() {
       log_warning(gc)("GC locker is held; pre-dump GC was skipped");
     }
   }
+  WorkerThreads* workers = Universe::heap()->safepoint_workers();
+  if (workers != nullptr) {
+      // The GC provided a WorkerThreads to be used during a safepoint.
+      // Can't run with more threads than provided by the WorkerThreads.
+      const uint capped_parallel_thread_num = MIN2(_parallel_thread_num, workers->max_workers());
+      WithActiveWorkers with_active_workers(workers, capped_parallel_thread_num);
```
albertnetymk, line 174: `WithActiveWorkers` is an stack-obj, so it must be in the same scope as using workers.

olivergillespie: Oh, good catch. So can I do something like:

```
HeapInspection inspect;
if (workers != nullptr) {
    // The GC provided a WorkerThreads to be used during a safepoint.
    // Can't run with more threads than provided by the WorkerThreads.
    const uint capped_parallel_thread_num = MIN2(_parallel_thread_num, workers->max_workers());
    WithActiveWorkers with_active_workers(workers, capped_parallel_thread_num);
    inspect.heap_inspection(_out, capped_parallel_thread_num);
    inspect.heap_inspection(_out, workers);
} else {
    inspect.heap_inspection(_out, _parallel_thread_num);
    inspect.heap_inspection(_out, nullptr);
}
```

?

(I'm not familiar with the exact workings of StackObj, sorry)

src/hotspot/share/memory/heapInspection.cpp
```diff
@@ -564,44 +564,33 @@ void ParHeapInspectTask::work(uint worker_id) {
...
-      WithActiveWorkers with_active_workers(workers, capped_parallel_thread_num);
-
-      ParallelObjectIterator poi(workers->active_workers());
-      ParHeapInspectTask task(&poi, cit, filter);
-      // Run task with the active workers.
-      workers->run_task(&task);
-      if (task.success()) {
-        return task.missed_count();
-      }
+  ResourceMark rm;
```
shipilev, line 569: I think this `ResourceMark` should remain at both blocks. So that on failed exit from the parallel block, we rollback resource allocations and start over with serial walk.

shipilev (commented):
This looks better to me. Some comments:

albertnetymk (approved):

src/hotspot/share/memory/heapInspection.cpp
olivergillespie: Ah okay, I didn't understand it before, my bad. Updated, thanks.

shipilev (approved):
This looks okay to me.

src/hotspot/share/gc/shared/gcVMOperations.cpp
```diff
@@ -167,7 +167,16 @@ void VM_GC_HeapInspection::doit() {
     }
   }
   HeapInspection inspect;
-  inspect.heap_inspection(_out, _parallel_thread_num);
+  WorkerThreads* workers = Universe::heap()->safepoint_workers();
+  if (workers != nullptr) {
+    // The GC provided a WorkerThreads to be used during a safepoint.
+    // Can't run with more threads than provided by the WorkerThreads.
+    const uint capped_parallel_thread_num = MIN2(_parallel_thread_num, workers->max_workers());
+    WithActiveWorkers with_active_workers(workers, capped_parallel_thread_num);
```
tschatzl, line 175: I would *almost* rely on the workgang's 'active_workers()' here because all collectors set this proportional to heap size (people tend to start 128M VMs on hundreds-of-thread machines... ). The problem is that one heap inspection VM operation. Maybe put the `WithActiveWorkers` there and use `workers->active_workers()` here?

This is just a suggestion.

tschatzl (approved):
See the suggestion.

src/hotspot/share/gc/shared/gcVMOperations.cpp
olivergillespie: Sorry, I don't think I understand the suggestion. Are you suggesting to use WithActiveWorkers only for the heap inspection VM op, and workers->active_workers() everywhere else (for the collector use-case)?

> The problem is that one heap inspection VM operation. Maybe put the WithActiveWorkers there

This area *is* that one heap inspection VM operation, isn't it?

tschatzl: You are right, I misread the code. Ignore my comment :) Ship it.

shipilev (approved):

src/hotspot/share/gc/shared/gcVMOperations.cpp
olivergillespie: Cool, thanks for reviewing :)

olivergillespie:
Re-ran tiers 2/3/4 on the refactor, Linux x86_64 and aarch64, all good. Marking for integration.

PR #13776, opened 2023-05-03
JDK-8307356: Metaspace: simplify BinList handling

tstuefe:
In preparation for Lilliput, I'd like to simplify BinList handling a bit.

BinList are a data structure that stores small blocks that had been prematurely deallocated for re-use. Due to the way BinList is implemented, it cannot store blocks smaller than two words. That causes complexities in metaspace and makes estimating memory usage awkward (important for testing). We can eliminate this complexity by shrinking a Block to just one word. 

----

Details:

A BinList block before: 
``` Block { Block* next; size_t size; }```
A BinList block now: 
``` Block { Block* next; }```

We don't need to store the block size, since all blocks in a single sub-list have the same size. In fact, we used the block size only for verification.

With this patch, gross- and net-allocation sizes in Metaspace are the same for 64-bit (see `get_raw_word_size_for_requested_word_size`), which will make writing tests that try to predict metaspace usage based on allocation pattern a lot less onerous to write. For 32-bit, unfortunately, we still need to align allocation sizes to 2 words since allocation addresses need to be aligned to 64-bit.

---------
### Progress

### Issue
 * [JDK-8307356](https://bugs.openjdk.org/browse/JDK-8307356): Metaspace: simplify BinList handling (**Enhancement** - P4)

### Reviewers
 * [Roman Kennke](https://openjdk.org/census#rkennke) (@rkennke - **Reviewer**) ⚠️ Review applies to [e24e177a](https://git.openjdk.org/jdk/pull/13776/files/e24e177ac75e9d31139b9ccba2943b95a2e2e14b)
 * [Coleen Phillimore](https://openjdk.org/census#coleenp) (@coleenp - **Reviewer**) ⚠️ Review applies to [e24e177a](https://git.openjdk.org/jdk/pull/13776/files/e24e177ac75e9d31139b9ccba2943b95a2e2e14b)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13776/head:pull/13776` \
`$ git checkout pull/13776`

`$ git checkout pull/13776` \
`$ git pull https://git.openjdk.org/jdk.git pull/13776/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13776`

`$ git pr show -t 13776`

</details>

<a href="https://git.openjdk.org/jdk/pull/13776.diff">https://git.openjdk.org/jdk/pull/13776.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13776#issuecomment-1534574692)

### CODE REVIEW

File: src/hotspot/share/memory/metaspace/binList.hpp [hunk 1]
```diff
@@ -73,32 +74,25 @@ namespace metaspace {
...
-  // Smallest block size must be large enough to hold a Block structure.
-  STATIC_ASSERT(smallest_word_size * sizeof(MetaWord) >= sizeof(Block));
+  // Block size must be exactly one word size.
+  STATIC_ASSERT(sizeof(Block) == BytesPerWord);
   STATIC_ASSERT(num_lists > 0);
 
 public:
 
-  // Minimal word size a block must have to be manageable by this structure.
-  const static size_t MinWordSize = smallest_word_size;
```
rkennke, line 98: Maybe keep MinWordSize = 1, then we can also keep using it, e.g. in code below:

int index = (int)(word_size - MinWordSize);

and many other places. Otherwise I wouldn't know what that '1' means.

src/hotspot/share/memory/metaspace/freeBlocks.cpp
```diff
@@ -53,7 +50,7 @@ MetaWord* FreeBlocks::remove_block(size_t requested_word_size) {
     // Blocks which are larger than a certain threshold are split and
     //  the remainder is handed back to the manager.
     const size_t waste = real_size - requested_word_size;
-    if (waste > MinWordSize) {
```
rkennke, line 56: Has the calculation been wrong before or should that be > 1 or am I missing something?

src/hotspot/share/memory/metaspace/metaspaceArena.cpp
```diff
@@ -60,7 +60,7 @@ chunklevel_t MetaspaceArena::next_chunk_level() const {
 void MetaspaceArena::salvage_chunk(Metachunk* c) {
   assert_lock_strong(lock());
   size_t remaining_words = c->free_below_committed_words();
-  if (remaining_words > FreeBlocks::MinWordSize) {
+  if (remaining_words > 0) {
```
rkennke, line 63: Same here.

rkennke (changes requested):
Looks mostly ok. I'd keep MinWordSize constants, though.

src/hotspot/share/memory/metaspace/freeBlocks.cpp
tstuefe: The calculation before had been wrong; MinWordSize is inclusive

test/hotspot/jtreg/TEST.groups
```diff
@@ -377,8 +377,6 @@ tier1_runtime = \
  -runtime/memory/ReserveMemory.java \
  -runtime/Metaspace/FragmentMetaspace.java \
  -runtime/Metaspace/FragmentMetaspaceSimple.java \
- -runtime/Metaspace/elastic/TestMetaspaceAllocationMT1.java \
- -runtime/Metaspace/elastic/TestMetaspaceAllocationMT2.java \
```
tstuefe, line 381: Reviewer hint: I need these tests back in GHA tier 1 for the moment (also because more Metaspace changes are coming for Lilliput)

src/hotspot/share/memory/metaspace/binList.hpp
```diff
@@ -114,7 +111,7 @@ class BinListImpl {
 
   static size_t word_size_for_index(int index) {
     assert(index >= 0 && index < num_lists, "Invalid index %d", index);
-    return MinWordSize + index;
+    return index + MinWordSize;
```
dholmes-ora, line 114: Pre-existing but I don't understand what `word_size_for_index` actually means??

src/hotspot/share/memory/metaspace/binList.hpp
```diff
@@ -156,13 +154,13 @@ class BinListImpl {
...
       assert(b != nullptr, "Sanity");
-      assert(b->_word_size >= word_size &&
-             b->_word_size == real_word_size,
-             "bad block size in list[%d] (" BLOCK_FORMAT ")", index, BLOCK_FORMAT_ARGS(b));
       _blocks[index] = b->_next;
       _counter.sub(real_word_size);
       *p_real_word_size = real_word_size;
-      return (MetaWord*)b;
+      MetaWord* p = (MetaWord*)b;
+      assert(real_word_size == 1 || p[real_word_size - 1] == 0,
```
dholmes-ora, line 161: Isn't the first case also handled by the second??

src/hotspot/share/memory/metaspace/binList.hpp
```diff
@@ -181,12 +179,12 @@ class BinListImpl {
...
     MemRangeCounter local_counter;
     for (int i = 0; i < num_lists; i++) {
-      const size_t s = MinWordSize + i;
+      const size_t s = word_size_for_index(i);
       int pos = 0;
       for (Block* b = _blocks[i]; b != nullptr; b = b->_next, pos++) {
-        assert(b->_word_size == s,
-               "bad block size in list[%d] at pos %d (" BLOCK_FORMAT ")",
-               i, pos, BLOCK_FORMAT_ARGS(b));
+        assert(s == 1 || ((MetaWord*)b)[s - 1] == 0,
```
dholmes-ora, line 185: Again isn't the first case handled by the second?

src/hotspot/share/memory/metaspace/metaspaceCommon.hpp
```diff
@@ -37,24 +37,27 @@ namespace metaspace {
 
 // Metaspace allocation alignment:
 
-// 1) Metaspace allocations have to be aligned such that 64bit values are aligned
-//  correctly.
+// Metaspace allocations have to be aligned such that 64bit values are aligned
```
dholmes-ora, line 40: nit existing: s/64bit/64-bit/

src/hotspot/share/memory/metaspace/metaspaceCommon.hpp
```diff
@@ -37,24 +37,27 @@ namespace metaspace {
...
+//  correctly. We currently don't hold members with a larger alignment requirement
+//  than 64-bit inside MetaData, so 8-byte alignment is enough.
 //
-// 2) Klass* structures allocated from Metaspace have to be aligned to KlassAlignmentInBytes.
+// Klass* structures need to be aligned to KlassAlignmentInBytes, but since that is
+// 64 bit, we don't need special handling for allocating Klass*.
```
dholmes-ora, line 45: nit 64-bit

dholmes-ora (commented):
A few drive-by comments as this isn't an area I'm familiar with.

I'm unclear if these kind of helper changes should be held off until we know what release Lilliput will be targeted to? Is this change valuable enough without Lilliput?

Thanks.

tstuefe:
Hi @dholmes-ora,

> I'm unclear if these kind of helper changes should be held off until we know what release Lilliput will be targeted to? Is this change valuable enough without Lilliput?

I think so, yes. It will allow me to simplify tests even independently from Lilliput.

src/hotspot/share/memory/metaspace/binList.hpp
tstuefe: This probably would deserve a comment, but I recently got obsessed with patch sizes :/

This line checks the trailing canary of this block that is written in BinList::add_block(). But if the blocks are of size 1, there is no space for the canary (in BinList::add_block(), the canary is overwritten by the 1-word-sized Block Header). Therefore, if the block is size 1, we cannot check the canary.

src/hotspot/share/memory/metaspace/binList.hpp
tstuefe: See answer above. I realize I probably should factor this out into a reasonably named method.

dholmes-ora (commented):
Thanks those updates make the canary handling much clearer.

src/hotspot/share/memory/metaspace/binList.hpp
```diff
@@ -138,9 +148,10 @@ class BinListImpl {
   void add_block(MetaWord* p, size_t word_size) {
     assert(word_size >= MinWordSize &&
            word_size <= MaxWordSize, "bad block size");
+    DEBUG_ONLY(write_canary(p, word_size);)
     const int index = index_for_word_size(word_size);
     Block* old_head = _blocks[index];
-    Block* new_head = new(p)Block(old_head, word_size);
+    Block* new_head = new(p)Block(old_head);
```
coleenp, line 154: nit: add spaces around (p)

src/hotspot/share/memory/metaspace/metaspaceArena.cpp
```diff
@@ -241,7 +246,7 @@ MetaWord* MetaspaceArena::allocate(size_t requested_word_size) {
   }
 
   // Primary allocation
-  p = allocate_inner(requested_word_size);
+  p = allocate_inner(raw_word_size);
```
coleenp, line 249: Why wasn't this always raw_word_size which seems to be the aligned size based on the requested word size? ('aligned' might be a better word than 'raw' here)  If I'm reading it correctly.

src/hotspot/share/memory/metaspace/metaspaceArena.cpp
```diff
@@ -365,7 +373,13 @@ void MetaspaceArena::deallocate_locked(MetaWord* p, size_t word_size) {
   UL2(trace, "deallocating " PTR_FORMAT ", word size: " SIZE_FORMAT ".",
       p2i(p), word_size);
 
+  // Only blocks that had been allocated via MetaspaceArena::allocate(size) must be handed in
+  // to MetaspaceArena::deallocate(), and only with the same size that had been original used for allocation.
+  // Therefore the pointer must be aligned correctly, and size can be alignment-adjusted (the latter
+  // only matters on 32-bit):
+  assert_is_aligned_metaspace_pointer(p);
   size_t raw_word_size = get_raw_word_size_for_requested_word_size(word_size);
+
```
coleenp, line 382: If this is deallocating a pointer returned from metaspace, why isn't raw_word_size the same as word_size?

test/hotspot/gtest/metaspace/test_binlist.cpp
```diff
@@ -46,8 +46,7 @@ using metaspace::MemRangeCounter;
 template <class BINLISTTYPE>
 struct BinListBasicTest {
 
-  static const size_t minws;
-  static const size_t maxws;
+    static const size_t maxws;
```
coleenp, line 49: indentation seems off?

test/hotspot/jtreg/TEST.groups
tstuefe: Yes, I have planned to make them faster, or cut them down a bit in automatic mode: https://bugs.openjdk.org/browse/JDK-8303815

tstuefe: (Patch is up for review; I cut down the time they take significantly)

src/hotspot/share/memory/metaspace/metaspaceArena.cpp
tstuefe: I think it was a plain oversight.

tstuefe:
@coleenp @rkennke very belated thanks for your reviews. This PR got lost under a pile of other works, but I brushed it up (mostly cosmetic changes) and would like to push it in preparation of follow-up work I plan.

coleenp (approved):
@tstuefe Still looks good.

rkennke (approved):

tstuefe:
Thanks @rkennke and @coleenp !

/integrate

PR #13582, opened 2023-04-21
8305896: Alternative full GC forwarding

rkennke:
Currently, the full-GC modes of Serial, Shenandoah and G1 GCs are forwarding objects by over-writing the object header with the new object location. Unfortunately, for compact object headers ([JDK-8294992](https://bugs.openjdk.org/browse/JDK-8294992)) this would not work, because the crucial class information is also stored in the header, and we could no longer iterate over objects until the headers would be restored. Also, the preserved-headers tables would grow quite large.

I propose to use an alternative algorithm for full-GC (sliding-GC) forwarding that uses a special encoding so that the forwarding information fits into the lowest 32 bits of the header.

It exploits the insight that, with sliding GCs, objects from one region will only ever be forwarded to one of two possible target regions. For this to work, we need to divide the heap into equal-sized regions. This is already the case for Shenandoah and G1, and can easily be overlaid for Serial GC, by assuming either the whole heap as a single region (if it fits) or by using SpaceAlignment-sized virtual regions.

We also build and maintain a table that has N elements, where N is the number of regions. Each entry is two addresses, which are the start-address of the possible target regions for each source region.

With this, forwarding information would be encoded like this:
 - Bits 0 and 1: same as before, we put in '11' to indicate that the object is forwarded.
 - Bit 2: Used for 'fallback'-forwarding (see below)
 - Bit 3: Selects the target region 0 or 1. Look up the base address in the table (see above)
 - Bits 4..31 The number of heap words from the target base address

This works well for all sliding GCs in Serial, G1 and Shenandoah. The exception is in G1, there is a special mode called 'serial compaction' which acts as a last-last-ditch effort to squeeze more space out of the heap by re-forwarding the tails of the compaction chains. Unfortunately, this breaks the assumption of the sliding-forwarding-table. When that happens, we initialize a fallback table, which is a simple open hash-table, and set the Bit 2 in the forwarding to indicate that we shall look up the forwardee in the fallback-table.

All the table accesses can be done unsynchronized because:
- Serial GC is single-threaded anyway
- In G1 and Shenandoah, GC worker threads divide up the work such that each worker does disjoint sets of regions.
- G1 serial compaction is single-threaded

[...]

PR #5955, opened 2021-10-14
8275527: Refactor forward pointer access

rkennke:
Accessing the forward pointer is currently a little inconsistent. Some code paths call oopDesc::forwardee() / oopDesc::is_forwarded(), some code paths call forwardee() and check it for ==/!= NULL, some code paths even call markWord::decode_pointer() and markWord::is_marked() instead.

This change attempts to make the situation more consistent. For simple cases it preserves oopDesc::forwardee() / is_forwarded(), some cases need to use the markWord for consistency in concurrent GC, they now use markWord::forwardee() and markWord::is_forwarded(). Also, checking whether or not an object is forwarded is now consistently done using is_forwarded() and not by checking forwardee ==/!= NULL. This also resolves the mess in G1 full GC that changes not-forwarded objects to have a NULL (fake-) pointer. This is not necessary, because we can just as well use the lock bits to determine whether or not the object is forwarded.

Testing:
 - [x] tier
 - [x] tier2
 - [x] hotspot_gc

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed

### Issue
 * [JDK-8275527](https://bugs.openjdk.java.net/browse/JDK-8275527): Refactor forward pointer access

### Reviewers
 * [Thomas Schatzl](https://openjdk.java.net/census#tschatzl) (@tschatzl - **Reviewer**) ⚠️ Review applies to a0e3a6a5f9673f574c485a6bdd812c0f7ae090c2
 * [Stefan Karlsson](https://openjdk.java.net/census#stefank) (@stefank - **Reviewer**) ⚠️ Review applies to 9c2ed2ff02ce5f7197ea28c41cdb7b119b7332f1

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk pull/5955/head:pull/5955` \
`$ git checkout pull/5955`

Update a local copy of the PR: \
`$ git checkout pull/5955` \
`$ git pull https://git.openjdk.java.net/jdk pull/5955/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 5955`

View PR using the GUI difftool: \
`$ git pr show -t 5955`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
[...]

PR #13779, opened 2023-05-03
8305898: Alternative self-forwarding mechanism

rkennke:
Currently, the Serial, Parallel and G1 GCs store a pointer to self into object headers, when compaction fails, to indicate that the object has been looked at, but failed compaction into to-space. This is problematic for compact object headers ([JDK-8294992](https://bugs.openjdk.org/browse/JDK-8294992)) because it would (temporarily) over-write the crucial class information, which we need for heap parsing. I would like to propose an alternative: use the bit #3 (previously biased-locking bit) to indicate that an object is 'self-forwarded'. That preserves the crucial class information in the upper bits of the header until the full header gets restored.

---------
### Progress

### Integration blocker
&nbsp;⚠️ Dependency #13582 must be integrated first

### Issue
 * [JDK-8305898](https://bugs.openjdk.org/browse/JDK-8305898): Alternative self-forwarding mechanism (**Enhancement** - P4)

### Reviewers
 * [Aleksey Shipilev](https://openjdk.org/census#shade) (@shipilev - **Reviewer**) 🔄 Re-review required (review was made when pull request targeted the [master](https://git.openjdk.org/jdk/tree/master) branch)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13779/head:pull/13779` \
`$ git checkout pull/13779`

`$ git checkout pull/13779` \
`$ git pull https://git.openjdk.org/jdk.git pull/13779/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13779`

`$ git pr show -t 13779`

</details>

<a href="https://git.openjdk.org/jdk/pull/13779.diff">https://git.openjdk.org/jdk/pull/13779.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13779#issuecomment-1535118983)


src/hotspot/share/oops/oop.inline.hpp
```diff
@@ -267,27 +268,76 @@ bool oopDesc::is_forwarded() const {
...
   } else {
-    return cast_to_oop(old_mark.decode_pointer());
+    return forwardee(old_mark);
+  }
+}
+
+oop oopDesc::forward_to_self_atomic(markWord compare, atomic_memory_order order) {
+  if (UseAltGCForwarding) {
+    markWord m = compare;
+    // If mark is displaced, we need to preserve real header during GC.
```
shipilev, line 306:
```suggestion
    // If mark is displaced, we need to preserve the real header during GC.
```

src/hotspot/share/oops/oop.inline.hpp
```diff
@@ -267,27 +268,76 @@ bool oopDesc::is_forwarded() const {
...
+  }
+}
+
+oop oopDesc::forwardee(markWord header) const {
+  assert(header.is_marked(), "only decode when actually forwarded");
+  if (header.self_forwarded()) {
+    assert(UseAltGCForwarding, "Only use self-fwd bits when using alt GC forwarding");
+    return cast_to_oop(this);
+  } else {
+    return cast_to_oop(header.decode_pointer());
```
shipilev, line 332: I think this path misses the original assert:

```
assert(is_forwarded(), "only decode when actually forwarded");
```

src/hotspot/share/oops/oop.inline.hpp
```diff
@@ -267,27 +268,76 @@ bool oopDesc::is_forwarded() const {
 // Used by scavengers
 void oopDesc::forward_to(oop p) {
   markWord m = markWord::encode_pointer_as_mark(p);
-  assert(m.decode_pointer() == p, "encoding must be reversible");
+  assert(forwardee(m) == p, "encoding must be reversable");
```
shipilev, line 271:
```suggestion
  assert(forwardee(m) == p, "encoding must be reversible");
```

src/hotspot/share/oops/oop.inline.hpp
```diff
@@ -267,27 +268,76 @@ bool oopDesc::is_forwarded() const {
...
+    assert(forwardee(m) == cast_to_oop(this), "encoding must be reversable");
+    markWord old_mark = cas_set_mark(m, compare, order);
+    if (old_mark == compare) {
+      return nullptr;
+    } else {
+      assert(old_mark.is_marked(), "must be marked here");
+      return forwardee(old_mark);
+    }
+  } else {
+    return forward_to_atomic(oop(this), compare, order);
```
shipilev, line 322:
```suggestion
    return forward_to_atomic(cast_to_oop(this), compare, order);
```

src/hotspot/share/oops/oop.inline.hpp
```diff
@@ -267,27 +268,76 @@ bool oopDesc::is_forwarded() const {
...
   if (old_mark == compare) {
     return nullptr;
   } else {
-    return cast_to_oop(old_mark.decode_pointer());
+    return forwardee(old_mark);
+  }
+}
+
+oop oopDesc::forward_to_self_atomic(markWord compare, atomic_memory_order order) {
+  if (UseAltGCForwarding) {
```
shipilev, line 304: Do you want to assert in `oopDesc::forward_to` and `oopDesc::forward_to_atomic` that they are not called with self-forwarding arguments?

src/hotspot/share/oops/oop.inline.hpp
```diff
@@ -267,27 +268,76 @@ bool oopDesc::is_forwarded() const {
...
+    }
+  } else {
+    return forward_to_atomic(oop(this), compare, order);
+  }
+}
+
+oop oopDesc::forwardee(markWord header) const {
+  assert(header.is_marked(), "only decode when actually forwarded");
+  if (header.self_forwarded()) {
+    assert(UseAltGCForwarding, "Only use self-fwd bits when using alt GC forwarding");
```
shipilev, line 329: This assert looks excessive, as `self_forwarded` asserts the same?

shipilev (commented):
Looks okay at the first glance, comments:

src/hotspot/share/oops/oop.inline.hpp
shipilev: Ah yes, nevermind then.

src/hotspot/share/oops/oop.inline.hpp
```diff
@@ -266,28 +267,78 @@ bool oopDesc::is_forwarded() const {
...
+  if (UseAltGCForwarding) {
+    markWord m = mark();
+    // If mark is displaced, we need to preserve the real header during GC.
+    // It will be restored to the displaced header after GC.
+    assert(SafepointSynchronize::is_at_safepoint(), "we can only safely fetch the displaced header at safepoint");
+    if (m.has_displaced_mark_helper()) {
+      m = m.displaced_mark_helper();
+    }
+    m = m.set_self_forwarded();
+    assert(forwardee(m) == cast_to_oop(this), "encoding must be reversable");
```
shipilev, line 286:
```suggestion
    assert(forwardee(m) == cast_to_oop(this), "encoding must be reversible");
```

src/hotspot/share/oops/oop.inline.hpp
```diff
@@ -266,28 +267,78 @@ bool oopDesc::is_forwarded() const {
...
+  if (UseAltGCForwarding) {
+    markWord m = compare;
+    // If mark is displaced, we need to preserve the real header during GC.
+    // It will be restored to the displaced header after GC.
+    assert(SafepointSynchronize::is_at_safepoint(), "we can only safely fetch the displaced header at safepoint");
+    if (m.has_displaced_mark_helper()) {
+      m = m.displaced_mark_helper();
+    }
+    m = m.set_self_forwarded();
+    assert(forwardee(m) == cast_to_oop(this), "encoding must be reversable");
```
shipilev, line 315:
```suggestion
    assert(forwardee(m) == cast_to_oop(this), "encoding must be reversible");
```

src/hotspot/share/oops/oop.inline.hpp
```diff
@@ -266,28 +267,78 @@ bool oopDesc::is_forwarded() const {
 
 // Used by scavengers
 void oopDesc::forward_to(oop p) {
+  assert(p != cast_to_oop(this) || !UseAltGCForwarding, "Must not be called with self-forwarding");
```
shipilev, line 270: Now that I had my morning coffee, I do have a question about the contract here. Can we accidentally call `oop->forward_to(compaction_point)` when `oop == compaction_point` from the compaction code? I guess that would be innocuous for the thing we want to protect against: recording the _promotion failure_, rather than the self-forwarding itself. In other words, the fact that object is self-forwarded might not exactly mean it failed the promotion, might just be a lucky coincidence?

If so, maybe this whole thing should be `oopDesc::forward_failed()` or some such, and then let the code decide how to record it, either with self-forwarding address (legacy) or with this new bit.

shipilev (commented):
More comments

src/hotspot/share/oops/oop.inline.hpp
rkennke: > Now that I had my morning coffee, I do have a question about the contract here. Can we accidentally call `oop->forward_to(compaction_point)` when `oop == compaction_point` from the compaction code?

No, that doesn't seem to happen. In this case, the object doesn't get forwarded at all. If it would happen, it could and should be ignored, because it would result in extra stuff to be executed.

> I guess that would be innocuous for the thing we want to protect against: recording the _promotion failure_, rather than the self-forwarding itself. In other words, the fact that object is self-forwarded might not exactly mean it failed the promotion, might just be a lucky coincidence?

No, we want to protect against self-forwarding, because that would irrecoverably destroy the Klass* with compact headers.

> If so, maybe this whole thing should be `oopDesc::forward_failed()` or some such, and then let the code decide how to record it, either with self-forwarding address (legacy) or with this new bit.

Yes, I guess I could do that.

fisk (changes requested):

src/hotspot/share/oops/oop.inline.hpp
```diff
@@ -266,28 +267,78 @@ bool oopDesc::is_forwarded() const {
...
   markWord m = markWord::encode_pointer_as_mark(p);
-  assert(m.decode_pointer() == p, "encoding must be reversible");
+  assert(forwardee(m) == p, "encoding must be reversible");
   set_mark(m);
 }
 
+void oopDesc::forward_failed() {
```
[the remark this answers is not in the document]
rkennke: Yeah.
It's even more confusing that we now have the notion of forward-failed, which aims to hide the implementation detail of self-forwarding, but forwardee() still exposes it. And probably has to, because that is how the forwarding logic of GCs currently work, and I'm not sure it is useful to change that. I need to mull over this a bit.

src/hotspot/share/oops/oop.inline.hpp
shipilev: Yeah, perhaps due to the self-forwarding contract with `forwardee`, this is not significantly cleaner. The encapsulation does not achieve much if we have the gaping hole from the other side of this abstraction. So the original `forward_to_self` is already good. Sorry for pushing in the wrong direction :)

shipilev: My only left-over concern is that the assert might still fail when self-forwarding for non-promotion-failure reasons, but that might as well indicate a performance problem in GC code that should avoid self-forwardings on the common path to begin with.

src/hotspot/share/oops/markWord.hpp
```diff
@@ -103,17 +104,20 @@ class markWord {
   // Constants
   static const int age_bits                       = 4;
   static const int lock_bits                      = 2;
-  static const int first_unused_gap_bits          = 1;
-  static const int max_hash_bits                  = BitsPerWord - age_bits - lock_bits - first_unused_gap_bits;
+  static const int self_forwarded_bits            = 1;
```
albertnetymk, line 107: This warrants some update to the doc above, right?

src/hotspot/share/gc/g1/g1ParScanThreadState.cpp
```diff
@@ -207,7 +207,7 @@ void G1ParScanThreadState::do_oop_evac(T* p) {
 
   markWord m = obj->mark();
   if (m.is_marked()) {
-    obj = cast_to_oop(m.decode_pointer());
+    obj = obj->forwardee(m);
```
mmyxym, line 210: Shall we have a method "oop::forwardee_not_self" which guarantee to be not self fowarded? So we can remove the self-forward if-check in GC critical path.

rkennke: Are there any paths that don't need to handle self-forwarded state?

Also, it seems to me that the path would be dominated by the load of the mark-word. Testing-and-branching for the self-forwarded bit seems like the minor problem there. It would be nice if we could tell the C++ compiler that the branch is expected to be uncommon, so it could shape the emitted code accordingly, but, afaik, we can't.

If we were to micro-optimize the forwarding-decoding, then it would be more useful to optimize the common pattern:

```
if (o->is_forwarded()) { // Loads and tests the mark-word
  oop fwd = o->forwardee(); // Loads mark-word again, and decode forwardee.
 ...
}
```

To something like:

```
oop fwd = o->forwardee(); // Return nullptr when not forwarded
if (fwd != nullptr) {
  ...
}
```

There's a way to improve further on this, as I proposed back in the early versions of #5955, which also avoids the decoding altogether if not forwarded, but it's a little more clunky:

```
OopForwarding fwd(obj);
if (fwd.is_forwarded()) {
  oop forwardee = fwd.forwardee();
  ...
}
```

where the scoped OopForwarding object would encapsulate the markWord and the testing and decoding of the fwd-ptr, but it has been rejected back then. But it would certainly help more than an oopDesc::forwardee_not_self() approach (if that were even possible, which I think it is not).

mmyxym: Sorry. It's my misunderstanding. G1ParScanThreadState::do_oop_evac still needs to handle self forward.

src/hotspot/share/oops/oop.inline.hpp
```diff
@@ -270,28 +271,87 @@ bool oopDesc::is_forwarded() const {
...
+    markWord m = mark();
+    // If mark is displaced, we need to preserve the real header during GC.
+    // It will be restored to the displaced header after GC.
+    assert(SafepointSynchronize::is_at_safepoint(), "we can only safely fetch the displaced header at safepoint");
+    if (m.has_displaced_mark_helper()) {
+      m = m.displaced_mark_helper();
+    }
+    m = m.set_self_forwarded();
+    assert(forwardee(m) == cast_to_oop(this), "encoding must be reversible");
+    set_mark(m);
```
stefank, line 292: Could you explain why we need to restore the displaced mark word here? I wonder why it matters what the old mark word is and why this code can't be changed to something like this:
```
    markWord m = markWord::prototype().set_self_forwarded();
    set_mark(m);
```

rkennke: The main reason why we need to implement the self-forwarding is to preserve the upper bits of the mark-word (specifically, the Klass* that we want to place there with JEP 450). If the header is displaced, we would install the self-forwarded bit in the tagged native-pointer that is in the header, meaning we loose the ability to get to the actual header to fetch the Klass*. We do restore the header later, but GC needs access to the Klass* meanwhile.

stefank: Thanks. While reviewing this patch I was thinking about the current JDK code, which as far as I understand doesn't need to restore the displaced header here.

rkennke: Right. It might be cleaner to move that part out of here and into #13961, WDYT?

stefank: I think that would be good.

PR #13460, opened 2023-04-13
8305959: x86: Improve itable_stub

bulasevich:
Async profiler shows that applications spend up to 10% in itable_stubs.

The current inefficiency of itable stubs is as follows. The generated itable_stub scans itable twice: first it checks if the object class is a subtype of the resolved_class, and then it finds the holder_class that implements the method. I suggest doing this in one pass: with a first loop over itable, check pointer equality to both holder_class and resolved_class. Once we have finished searching for resolved_class, continue searching for holder_class in a separate loop if it has not yet been found.

This approach gives 1-10% improvement on the synthetic benchmarks and 3% improvement on Naive Bayes benchmark from the Renaissance Benchmark Suite (Intel Xeon X5675).

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8305959](https://bugs.openjdk.org/browse/JDK-8305959): x86: Improve itable_stub

### Reviewers
 * [Paul Hohensee](https://openjdk.org/census#phh) (@phohensee - **Reviewer**) ⚠️ Review applies to [7a259831](https://git.openjdk.org/jdk/pull/13460/files/7a2598314fa02baad68f823d93fb14a9bfd3bf22)
 * [Aleksey Shipilev](https://openjdk.org/census#shade) (@shipilev - **Reviewer**)
 * [Andrew Haley](https://openjdk.org/census#aph) (@theRealAph - **Reviewer**) ⚠️ Review applies to [7a259831](https://git.openjdk.org/jdk/pull/13460/files/7a2598314fa02baad68f823d93fb14a9bfd3bf22)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13460/head:pull/13460` \
`$ git checkout pull/13460`

Update a local copy of the PR: \
`$ git checkout pull/13460` \
`$ git pull https://git.openjdk.org/jdk.git pull/13460/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 13460`

View PR using the GUI difftool: \
`$ git pr show -t 13460`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
[...]

PR #13792, opened 2023-05-04
8307352: AARCH64: Improve itable_stub

bulasevich:
This is a change for AARCH similar to https://github.com/openjdk/jdk/pull/13460 

The change replaces two separate iterations over the itable with a new algorithm consisting of two loops. First, we look for a match with resolved_klass, checking for a match with holder_klass along the way. Then we continue iterating (not starting over) the itable using the second loop, checking only for a match with holder_klass.

InterfaceCalls openjdk benchmark performance results on A53, A72, Neoverse N1 and V1 micro-architectures:

```
Cortex-A53 (Pi 3 Model B Rev 1.2)

test1stInt2Types    37.5      37.358     0.38
test1stInt3Types   160.166   148.04      8.19
test1stInt5Types   158.131   147.955     6.88
test2ndInt2Types    52.634    53.291    -1.23
test2ndInt3Types   201.39    181.603    10.90
test2ndInt5Types   195.722   176.707    10.76
testIfaceCall      157.453   140.498    12.07
testIfaceExtCall   175.46    154.351    13.68
testMonomorphic     32.052    32.039     0.04
                                 AVG:    6.85

Cortex-A72 (Pi 4 Model B Rev 1.2)

test1stInt2Types    27.4796   27.4738    0.02
test1stInt3Types    66.0085   64.9374    1.65
test1stInt5Types    67.9812   66.2316    2.64
test2ndInt2Types    32.0581   32.062    -0.01
test2ndInt3Types    68.2715   65.6643    3.97
test2ndInt5Types    68.1012   65.8024    3.49
testIfaceCall       64.0684   64.1811   -0.18
testIfaceExtCall    91.6226   81.5867   12.30
testMonomorphic     26.7161   26.7142    0.01
                                 AVG:    2.66

Neoverse N1 (m6g.metal)

test1stInt2Types     2.9104    2.9086    0.06
test1stInt3Types    10.9642   10.2909    6.54
test1stInt5Types    10.9607   10.2856    6.56
test2ndInt2Types     3.3410    3.3478   -0.20
test2ndInt3Types    12.3291   11.3089    9.02
test2ndInt5Types    12.328    11.2704    9.38
testIfaceCall       11.0598   10.3657    6.70
testIfaceExtCall    13.0692   11.2826   15.84
testMonomorphic      2.2354    2.2341    0.06
                                 AVG:    6.00

Neoverse V1 (c7g.2xlarge)

test1stInt2Types    2.2317     2.2320   -0.01
test1stInt3Types    6.6884     6.1911    8.03
test1stInt5Types    6.7334     6.2193    8.27
test2ndInt2Types    2.4002     2.4013   -0.04
test2ndInt3Types    7.9603     7.0372   13.12
test2ndInt5Types    7.9532     7.0474   12.85
testIfaceCall       6.7028     6.3272    5.94
testIfaceExtCall    8.3253     6.9416   19.93
testMonomorphic     1.2446     1.2544   -0.79
                                 AVG:    7.48 
```

Testing: jtreg tier1-3, release & fastdebug

---------
### Progress

### Issue
 * [JDK-8307352](https://bugs.openjdk.org/browse/JDK-8307352): AARCH64: Improve itable_stub (**Enhancement** - P4)

### Reviewers
 * [Volker Simonis](https://openjdk.org/census#simonis) (@simonis - **Reviewer**) ⚠️ Review applies to [1687c4bd](https://git.openjdk.org/jdk/pull/13792/files/1687c4bd80e44fc6631205fefcdf64e2ac2c4034)
 * [Evgeny Astigeevich](https://openjdk.org/census#eastigeevich) (@eastig - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13792/head:pull/13792` \
`$ git checkout pull/13792`

`$ git checkout pull/13792` \
`$ git pull https://git.openjdk.org/jdk.git pull/13792/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13792`

`$ git pr show -t 13792`

</details>

<a href="https://git.openjdk.org/jdk/pull/13792.diff">https://git.openjdk.org/jdk/pull/13792.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13792#issuecomment-1589442109)

### DIFF
--- a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
@@ -1228,10 +1228,12 @@ void MacroAssembler::lookup_interface_method_stub(Register recv_klass,
   add(recv_klass, recv_klass, vtable_start_offset + ioffset);
   // itableOffsetEntry[] itable = recv_klass + Klass::vtable_start_offset() + sizeof(vtableEntry) * recv_klass->_vtable_len;
   // temp_itbl_klass = itable[0]._interface;
-  ldr(temp_itbl_klass, Address(recv_klass, scan_temp, Address::lsl(exact_log2(vtableEntry::size_in_bytes()))));
+  int vtblEntrySize = vtableEntry::size_in_bytes();
+  guarantee(vtblEntrySize == wordSize, "ldr shift amount must be lsl#3");
+  ldr(temp_itbl_klass, Address(recv_klass, scan_temp, Address::lsl(exact_log2(vtblEntrySize))));
   mov(holder_offset, zr);
   // scan_temp = &(itable[0]._interface)
-  lea(scan_temp, Address(recv_klass, scan_temp, Address::lsl(exact_log2(vtableEntry::size_in_bytes()))));
+  lea(scan_temp, Address(recv_klass, scan_temp, Address::lsl(exact_log2(vtblEntrySize))));
 
   // Initial checks:
   //   - if (holder_klass != resolved_klass), go to "scan for resolved"
@@ -1294,7 +1296,8 @@ void MacroAssembler::lookup_interface_method_stub(Register recv_klass,
   // Finally, scan_temp contains holder_klass vtable offset
   bind(L_holder_found);
   ldrw(method_result, Address(scan_temp, ooffset - ioffset));
-  add(recv_klass, recv_klass, (itable_index << LogBytesPerWord) + in_bytes(itableMethodEntry::method_offset()) - vtable_start_offset - ioffset);
+  add(recv_klass, recv_klass, itable_index * wordSize + in_bytes(itableMethodEntry::method_offset())
+    - vtable_start_offset - ioffset); // substract offsets to restore the original value of recv_klass
   ldr(method_result, Address(recv_klass, method_result, Address::uxtw(0)));
 }
 
--- a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
@@ -1229,7 +1229,7 @@ void MacroAssembler::lookup_interface_method_stub(Register recv_klass,
   // itableOffsetEntry[] itable = recv_klass + Klass::vtable_start_offset() + sizeof(vtableEntry) * recv_klass->_vtable_len;
   // temp_itbl_klass = itable[0]._interface;
   int vtblEntrySize = vtableEntry::size_in_bytes();
-  guarantee(vtblEntrySize == wordSize, "ldr shift amount must be lsl#3");
+  assert(vtblEntrySize == wordSize, "ldr shift amount must be lsl#3");
   ldr(temp_itbl_klass, Address(recv_klass, scan_temp, Address::lsl(exact_log2(vtblEntrySize))));
   mov(holder_offset, zr);
   // scan_temp = &(itable[0]._interface)
--- a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
@@ -1215,22 +1215,21 @@ void MacroAssembler::lookup_interface_method_stub(Register recv_klass,
   assert_different_registers(resolved_klass, recv_klass, holder_klass, temp_itbl_klass, scan_temp, holder_offset);
   assert_different_registers(method_result, recv_klass, holder_klass, temp_itbl_klass, scan_temp, holder_offset);
 
-  int vtable_base = in_bytes(Klass::vtable_start_offset());
-  int itentry_off = in_bytes(itableMethodEntry::method_offset());
-  int scan_step = itableOffsetEntry::size() * wordSize;
-  int vte_size = vtableEntry::size_in_bytes();
+  int vtable_start_offset = in_bytes(Klass::vtable_start_offset());
+  int itable_offset_entry_size = itableOffsetEntry::size() * wordSize;
   int ioffset = in_bytes(itableOffsetEntry::interface_offset());
   int ooffset = in_bytes(itableOffsetEntry::offset_offset());
-  assert(vte_size == wordSize, "else adjust times_vte_scale");
+  assert(vtableEntry::size_in_bytes() == wordSize, "else adjust times_vte_scale");
 
-  Label L_loop_scan_resolved_entry, L_resolved_found, L_holder_found;
+  Label L_loop_search_resolved_entry, L_resolved_found, L_holder_found;
 
-  // temp_itbl_klass = recv_klass.itable[0]
-  // scan_temp = &recv_klass.itable[0]
   ldrw(scan_temp, Address(recv_klass, Klass::vtable_length_offset()));
-  add(recv_klass, recv_klass, vtable_base + ioffset);
+  add(recv_klass, recv_klass, vtable_start_offset + ioffset);
+  // itableOffsetEntry[] itable = recv_klass + Klass::vtable_start_offset() + recv_klass->_vtable_len;
+  // temp_itbl_klass = itable[0]._interface;
   ldr(temp_itbl_klass, Address(recv_klass, scan_temp, Address::lsl(3)));
   mov(holder_offset, zr);
+  // scan_temp = &(itable[0]._interface)
   lea(scan_temp, Address(recv_klass, scan_temp, Address::lsl(3)));
 
   // Initial checks:
@@ -1238,67 +1237,63 @@ void MacroAssembler::lookup_interface_method_stub(Register recv_klass,
   //   - if (itable[0] == holder_klass), shortcut to "holder found"
   //   - if (itable[0] == 0), no such interface
   cmp(resolved_klass, holder_klass);
-  br(Assembler::NE, L_loop_scan_resolved_entry);
+  br(Assembler::NE, L_loop_search_resolved_entry);
   cmp(holder_klass, temp_itbl_klass);
   br(Assembler::EQ, L_holder_found);
   cbz(temp_itbl_klass, L_no_such_interface);
 
   // Loop: Look for holder_klass record in itable
   //   do {
-  //     temp_itbl_klass = *(scan_temp += scan_step);
+  //     temp_itbl_klass = *(scan_temp += itable_offset_entry_size);
   //     if (temp_itbl_klass == holder_klass) {
   //       goto L_holder_found; // Found!
   //     }
   //   } while (temp_itbl_klass != 0);
   //   goto L_no_such_interface // Not found.
-  Label L_scan_holder;
-  bind(L_scan_holder);
-    ldr(temp_itbl_klass, Address(pre(scan_temp, scan_step)));
+  Label L_search_holder;
+  bind(L_search_holder);
+    ldr(temp_itbl_klass, Address(pre(scan_temp, itable_offset_entry_size)));
     cmp(holder_klass, temp_itbl_klass);
     br(Assembler::EQ, L_holder_found);
-    cbnz(temp_itbl_klass, L_scan_holder);
+    cbnz(temp_itbl_klass, L_search_holder);
 
   b(L_no_such_interface);
 
   // Loop: Look for resolved_class record in itable
-  //   L_save_holder_offset:
-  //     holder_offset = index;
-  //   do {
-  //     temp_itbl_klass = *(scan_temp += scan_step);
-  //     if (temp_itbl_klass == holder_klass) {
-  //        goto L_save_holder_offset;
+  //   while (true) {
+  //     temp_itbl_klass = *(scan_temp += itable_offset_entry_size);
+  //     if (temp_itbl_klass == 0) {
+  //       goto L_no_such_interface;
   //     }
   //     if (temp_itbl_klass == resolved_klass) {
   //        goto L_resolved_found;  // Found!
   //     }
-  //   } while (temp_itbl_klass != 0);
-  //   goto L_no_such_interface // Not found.
+  //     if (temp_itbl_klass == holder_klass) {
+  //        holder_offset = scan_temp;
+  //     }
+  //   }
   //
-  Label L_save_holder_offset;
-  bind(L_save_holder_offset);
-    mov(holder_offset, scan_temp);
-    // continue searching for resolved_found (checked above: resolved found != resolved class)
-  Label L_loop_scan_resolved;
-  bind(L_loop_scan_resolved);
-    ldr(temp_itbl_klass, Address(pre(scan_temp, scan_step)));
-    bind(L_loop_scan_resolved_entry);
-    cmp(holder_klass, temp_itbl_klass);
-    br(Assembler::EQ, L_save_holder_offset);
+  Label L_loop_search_resolved;
+  bind(L_loop_search_resolved);
+    ldr(temp_itbl_klass, Address(pre(scan_temp, itable_offset_entry_size)));
+  bind(L_loop_search_resolved_entry);
+    cbz(temp_itbl_klass, L_no_such_interface);
     cmp(resolved_klass, temp_itbl_klass);
     br(Assembler::EQ, L_resolved_found);
-    cbnz(temp_itbl_klass, L_loop_scan_resolved);
-
-  b(L_no_such_interface);
+    cmp(holder_klass, temp_itbl_klass);
+    br(Assembler::NE, L_loop_search_resolved);
+    mov(holder_offset, scan_temp);
+    b(L_loop_search_resolved);
 
   // See if we already have a holder klass. If not, go and scan for it.
   bind(L_resolved_found);
-  cbz(holder_offset, L_scan_holder);
+  cbz(holder_offset, L_search_holder);
   mov(scan_temp, holder_offset);
 
   // Finally, scan_temp contains holder_klass vtable offset
   bind(L_holder_found);
   ldrw(method_result, Address(scan_temp, ooffset - ioffset));
-  add(recv_klass, recv_klass, (itable_index << 3) + itentry_off - (vtable_base + ioffset));
+  add(recv_klass, recv_klass, (itable_index << 3) + in_bytes(itableMethodEntry::method_offset()) - vtable_start_offset - ioffset);
   ldr(method_result, Address(recv_klass, method_result, Address::uxtw(0)));
 }
 
--- a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
@@ -1209,11 +1209,12 @@ void MacroAssembler::lookup_interface_method_stub(Register recv_klass,
                                                   Register method_result,
                                                   Register temp_itbl_klass,
                                                   Register scan_temp,
-                                                  Register holder_offset,
                                                   int itable_index,
                                                   Label& L_no_such_interface) {
+  // 'method_result' is only used as output register at the very end of this method.
+  // Until then we can reuse it as 'holder_offset'.
+  Register holder_offset = method_result;
   assert_different_registers(resolved_klass, recv_klass, holder_klass, temp_itbl_klass, scan_temp, holder_offset);
-  assert_different_registers(method_result, recv_klass, holder_klass, temp_itbl_klass, scan_temp, holder_offset);
 
   int vtable_start_offset = in_bytes(Klass::vtable_start_offset());
   int itable_offset_entry_size = itableOffsetEntry::size() * wordSize;
@@ -1225,12 +1226,12 @@ void MacroAssembler::lookup_interface_method_stub(Register recv_klass,
 
   ldrw(scan_temp, Address(recv_klass, Klass::vtable_length_offset()));
   add(recv_klass, recv_klass, vtable_start_offset + ioffset);
-  // itableOffsetEntry[] itable = recv_klass + Klass::vtable_start_offset() + recv_klass->_vtable_len;
+  // itableOffsetEntry[] itable = recv_klass + Klass::vtable_start_offset() + sizeof(vtableEntry) * recv_klass->_vtable_len;
   // temp_itbl_klass = itable[0]._interface;
-  ldr(temp_itbl_klass, Address(recv_klass, scan_temp, Address::lsl(3)));
+  ldr(temp_itbl_klass, Address(recv_klass, scan_temp, Address::lsl(exact_log2(vtableEntry::size_in_bytes()))));
   mov(holder_offset, zr);
   // scan_temp = &(itable[0]._interface)
-  lea(scan_temp, Address(recv_klass, scan_temp, Address::lsl(3)));
+  lea(scan_temp, Address(recv_klass, scan_temp, Address::lsl(exact_log2(vtableEntry::size_in_bytes()))));
 
   // Initial checks:
   //   - if (holder_klass != resolved_klass), go to "scan for resolved"
@@ -1293,7 +1294,7 @@ void MacroAssembler::lookup_interface_method_stub(Register recv_klass,
   // Finally, scan_temp contains holder_klass vtable offset
   bind(L_holder_found);
   ldrw(method_result, Address(scan_temp, ooffset - ioffset));
-  add(recv_klass, recv_klass, (itable_index << 3) + in_bytes(itableMethodEntry::method_offset()) - vtable_start_offset - ioffset);
+  add(recv_klass, recv_klass, (itable_index << LogBytesPerWord) + in_bytes(itableMethodEntry::method_offset()) - vtable_start_offset - ioffset);
   ldr(method_result, Address(recv_klass, method_result, Address::uxtw(0)));
 }
 
--- a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
@@ -1196,6 +1196,112 @@ void MacroAssembler::lookup_interface_method(Register recv_klass,
   }
 }
 
+// Look up the method for a megamorphic invokeinterface call in a single pass over itable:
+// - check recv_klass (actual object class) is a subtype of resolved_klass from CompiledICHolder
+// - find a holder_klass (class that implements the method) vtable offset and get the method from vtable by index
+// The target method is determined by <holder_klass, itable_index>.
+// The receiver klass is in recv_klass.
+// On success, the result will be in method_result, and execution falls through.
+// On failure, execution transfers to the given label.
+void MacroAssembler::lookup_interface_method_stub(Register recv_klass,
+                                                  Register holder_klass,
+                                                  Register resolved_klass,
+                                                  Register method_result,
+                                                  Register temp_itbl_klass,
+                                                  Register scan_temp,
+                                                  Register holder_offset,
+                                                  int itable_index,
+                                                  Label& L_no_such_interface) {
+  assert_different_registers(resolved_klass, recv_klass, holder_klass, temp_itbl_klass, scan_temp, holder_offset);
+  assert_different_registers(method_result, recv_klass, holder_klass, temp_itbl_klass, scan_temp, holder_offset);
+
+  int vtable_base = in_bytes(Klass::vtable_start_offset());
+  int itentry_off = in_bytes(itableMethodEntry::method_offset());
+  int scan_step = itableOffsetEntry::size() * wordSize;
+  int vte_size = vtableEntry::size_in_bytes();
+  int ioffset = in_bytes(itableOffsetEntry::interface_offset());
+  int ooffset = in_bytes(itableOffsetEntry::offset_offset());
+  assert(vte_size == wordSize, "else adjust times_vte_scale");
+
+  Label L_loop_scan_resolved_entry, L_resolved_found, L_holder_found;
+
+  // temp_itbl_klass = recv_klass.itable[0]
+  // scan_temp = &recv_klass.itable[0]
+  ldrw(scan_temp, Address(recv_klass, Klass::vtable_length_offset()));
+  add(recv_klass, recv_klass, vtable_base + ioffset);
+  ldr(temp_itbl_klass, Address(recv_klass, scan_temp, Address::lsl(3)));
+  mov(holder_offset, zr);
+  lea(scan_temp, Address(recv_klass, scan_temp, Address::lsl(3)));
+
+  // Initial checks:
+  //   - if (holder_klass != resolved_klass), go to "scan for resolved"
+  //   - if (itable[0] == holder_klass), shortcut to "holder found"
+  //   - if (itable[0] == 0), no such interface
+  cmp(resolved_klass, holder_klass);
+  br(Assembler::NE, L_loop_scan_resolved_entry);
+  cmp(holder_klass, temp_itbl_klass);
+  br(Assembler::EQ, L_holder_found);
+  cbz(temp_itbl_klass, L_no_such_interface);
+
+  // Loop: Look for holder_klass record in itable
+  //   do {
+  //     temp_itbl_klass = *(scan_temp += scan_step);
+  //     if (temp_itbl_klass == holder_klass) {
+  //       goto L_holder_found; // Found!
+  //     }
+  //   } while (temp_itbl_klass != 0);
+  //   goto L_no_such_interface // Not found.
+  Label L_scan_holder;
+  bind(L_scan_holder);
+    ldr(temp_itbl_klass, Address(pre(scan_temp, scan_step)));
+    cmp(holder_klass, temp_itbl_klass);
+    br(Assembler::EQ, L_holder_found);
+    cbnz(temp_itbl_klass, L_scan_holder);
+
+  b(L_no_such_interface);
+
+  // Loop: Look for resolved_class record in itable
+  //   L_save_holder_offset:
+  //     holder_offset = index;
+  //   do {
+  //     temp_itbl_klass = *(scan_temp += scan_step);
+  //     if (temp_itbl_klass == holder_klass) {
+  //        goto L_save_holder_offset;
+  //     }
+  //     if (temp_itbl_klass == resolved_klass) {
+  //        goto L_resolved_found;  // Found!
+  //     }
+  //   } while (temp_itbl_klass != 0);
+  //   goto L_no_such_interface // Not found.
+  //
+  Label L_save_holder_offset;
+  bind(L_save_holder_offset);
+    mov(holder_offset, scan_temp);
+    // continue searching for resolved_found (checked above: resolved found != resolved class)
+  Label L_loop_scan_resolved;
+  bind(L_loop_scan_resolved);
+    ldr(temp_itbl_klass, Address(pre(scan_temp, scan_step)));
+    bind(L_loop_scan_resolved_entry);
+    cmp(holder_klass, temp_itbl_klass);
+    br(Assembler::EQ, L_save_holder_offset);
+    cmp(resolved_klass, temp_itbl_klass);
+    br(Assembler::EQ, L_resolved_found);
+    cbnz(temp_itbl_klass, L_loop_scan_resolved);
+
+  b(L_no_such_interface);
+
+  // See if we already have a holder klass. If not, go and scan for it.
+  bind(L_resolved_found);
+  cbz(holder_offset, L_scan_holder);
+  mov(scan_temp, holder_offset);
+
+  // Finally, scan_temp contains holder_klass vtable offset
+  bind(L_holder_found);
+  ldrw(method_result, Address(scan_temp, ooffset - ioffset));
+  add(recv_klass, recv_klass, (itable_index << 3) + itentry_off - (vtable_base + ioffset));
+  ldr(method_result, Address(recv_klass, method_result, Address::uxtw(0)));
+}
+
 // virtual method calling
 void MacroAssembler::lookup_virtual_method(Register recv_klass,
                                            RegisterOrConstant vtable_index,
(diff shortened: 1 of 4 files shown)


theRealAph:
Thanks.

For this to be reviewable, we'll need:

A benchmark, and some data.
An explanation of why it's better than the existing implementation.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1197,6 +1197,108 @@ void MacroAssembler::lookup_interface_method(Register recv_klass,
...
+  int ooffset = in_bytes(itableOffsetEntry::offset_offset());
+  assert(vte_size == wordSize, "else adjust times_vte_scale");
+
+  Label L_loop_scan_resolved_entry, L_resolved_found, L_holder_found;
+
+  // temp_itbl_klass = recv_klass.itable[0]
+  // scan_temp = &recv_klass.itable[0]
+  ldrw(scan_temp, Address(recv_klass, Klass::vtable_length_offset()));
+  add(recv_klass, recv_klass, vtable_base + ioffset);
+  ldr(temp_itbl_klass, Address(recv_klass, scan_temp, Address::lsl(3)));
```
theRealAph, line 1233:
```suggestion
  ldr(temp_itbl_klass, Address(recv_klass, scan_temp, Address::lsl(LogBytesPerInt)));
```

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1197,6 +1197,108 @@ void MacroAssembler::lookup_interface_method(Register recv_klass,
...
+  // Loop: Look for holder_klass record in itable
+  //   do {
+  //     index += step;
+  //     temp_itbl_klass = itable[index];
+  //     if (temp_itbl_klass == holder_klass) {
+  //       goto L_holder_found; // Found!
+  //     }
+  //   } while (temp_itbl_klass != 0);
+  //   goto L_no_such_interface // Not found.
+  Label L_scan_holder;
```
theRealAph, line 1256: I don't think the pseudocode matches the assembly. This is more like

```
  do {
    temp_itbl_klass = *(scan_temp += scan_step);
    if (holder_klass, temp_itbl_klass ) ...

```

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1197,6 +1197,108 @@ void MacroAssembler::lookup_interface_method(Register recv_klass,
...
+  //     }
+  //   } while (temp_itbl_klass != 0);
+  //   goto L_no_such_interface // Not found.
+  //
+  Label L_loop_scan_resolved;
+  bind(L_loop_scan_resolved);
+    ldr(temp_itbl_klass, Address(pre(scan_temp, scan_step)));
+    bind(L_loop_scan_resolved_entry);
+    cmp(holder_klass, temp_itbl_klass);
+    csel(holder_offset, scan_temp, holder_offset, Assembler::EQ);
```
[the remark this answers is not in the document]
bulasevich: Right. Thank you.
Besides micro-benchmarks I have seen good results on real applications.
I changed `csel` to branch. And it works even better. High noise makes it is hard to see the improvement, but it looks like this change makes us 0.25% faster on InterfaceCalls micro-benchmarks:
```diff
   // Loop: Look for resolved_class record in itable
+  //   L_save_holder_offset:
+  //     holder_offset = index;
   //   do {
-  //     index += step;
-  //     temp_itbl_klass = itable[index];
+  //     temp_itbl_klass = *(scan_temp += scan_step);
   //     if (temp_itbl_klass == holder_klass) {
-  //        holder_offset = index;
+  //        goto L_save_holder_offset;
   //     }
   //     if (temp_itbl_klass == resolved_klass) {
   //        goto L_resolved_found;  // Found!
   //     }
   //   } while (temp_itbl_klass != 0);
   //   goto L_no_such_interface // Not found.
   //
+  Label L_save_holder_offset;
+  bind(L_save_holder_offset);
+    mov(holder_offset, scan_temp);
+    // continue searching for resolved_found (checked above: resolved found != resolved class)
   Label L_loop_scan_resolved;
   bind(L_loop_scan_resolved);
     ldr(temp_itbl_klass, Address(pre(scan_temp, scan_step)));
     bind(L_loop_scan_resolved_entry);
     cmp(holder_klass, temp_itbl_klass);
-    csel(holder_offset, scan_temp, holder_offset, Assembler::EQ);
+    br(Assembler::EQ, L_save_holder_offset);
     cmp(resolved_klass, temp_itbl_klass);
     br(Assembler::EQ, L_resolved_found);
     cbnz(temp_itbl_klass, L_loop_scan_resolved);
```

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
bulasevich: Yes. Thanks!

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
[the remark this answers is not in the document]
theRealAph: > Right. Thank you. Besides micro-benchmarks I have seen good results on real applications. I changed `csel` to branch. And it works even better. High noise makes it is hard to see the improvement, but it looks like this change makes us 0.25% faster on InterfaceCalls micro-benchmarks:

On what hardware?

src/hotspot/cpu/aarch64/vtableStubs_aarch64.cpp
```diff
@@ -192,28 +193,15 @@ VtableStub* VtableStubs::create_itable_stub(int itable_index) {
...
-  const ptrdiff_t  typecheckSize = __ pc() - start_pc;
-  start_pc = __ pc();
-
   // Get selected method from declaring class and itable index
-  __ lookup_interface_method(// inputs: rec. class, interface, itable index
-                             recv_klass_reg, holder_klass_reg, itable_index,
-                             // outputs: method, scan temp. reg
-                             rmethod, temp_reg,
-                             L_no_such_interface);
+  __ lookup_interface_method_stub(recv_klass_reg, holder_klass_reg, resolved_klass_reg, rmethod,
```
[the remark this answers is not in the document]
bulasevich: Yes, there are similar calls to lookup_interface_method stubs from TemplateTable::invokeinterface and from VtableStubs::create_itable_stub. Unfortunately, the instructions in between the lookup_interface_method calls are different, making it difficult to create a common stub for them. Performance is not an issue for the interpreter. I think the dilemma is: (1) create a new but complicated common stub vs. (2) add a faster implementation for VtableStubs, and leave the TemplateTable implementation intact. At the moment, I prefer the second way. Let me create a JBS issue with the idea of switching TemplateTable::invokeinterface (both aarch and x86 impl) to a new implementation if anyone encounters the problem of TemplateTable::invokeinterface performance.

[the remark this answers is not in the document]
theRealAph: > Yes, there are similar calls to lookup_interface_method stubs from TemplateTable::invokeinterface and from VtableStubs::create_itable_stub. Unfortunately, the instructions in between the lookup_interface_method calls are different, making it difficult to create a common stub for them. Performance is not an issue for the interpreter.

Really? if performance were not an issue for the interpreter we'd all use the C++ interpreter.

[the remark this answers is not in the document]
bulasevich: What I am trying to say is that I have not found any benchmark or application where a performance impact of TemplateTable::invokeinterface is visible. I would be happy to rework it as well once we see a benchmark which clearly demonstrates the gain, and the gain is substantial to justify the code complexity.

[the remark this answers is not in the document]
theRealAph: Efficient systems are made from thousands of tiny optimizations, and few of them on their own are large enough to make a measurable difference to performance. So it is for vehicles, so it is for computers.

Having said that, I'd probably be opposed to anything in this area that made for more code or added complexity.

src/hotspot/cpu/aarch64/vtableStubs_aarch64.cpp
```diff
@@ -192,28 +193,15 @@ VtableStub* VtableStubs::create_itable_stub(int itable_index) {
...
+                                  temp_reg, temp_reg2, temp_reg3, itable_index, L_no_such_interface);
 
   const ptrdiff_t lookupSize = __ pc() - start_pc;
```
eastig, line 200: I think you can rename `lookupSize` in `codesize`.

src/hotspot/cpu/aarch64/vtableStubs_aarch64.cpp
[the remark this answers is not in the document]
eastig: @bulasevich 
> What I am trying to say is that I have not found any benchmark

IMHO you can run the benchmark mentioned in the description with `-Xint`.

> Unfortunately, the instructions in between the lookup_interface_method calls are different, making it difficult to create a common stub for them.

If it needs a lot of work, we can create a JBS issue to implement the same approach for the interpreter. When it is being implemented some common parts might be identified and refactored.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1195,6 +1195,112 @@ void MacroAssembler::lookup_interface_method(Register recv_klass,
...
+
+  Label L_loop_scan_resolved_entry, L_resolved_found, L_holder_found;
+
+  // temp_itbl_klass = recv_klass.itable[0]
+  // scan_temp = &recv_klass.itable[0]
+  ldrw(scan_temp, Address(recv_klass, Klass::vtable_length_offset()));
+  add(recv_klass, recv_klass, vtable_base + ioffset);
+  ldr(temp_itbl_klass, Address(recv_klass, scan_temp, Address::lsl(3)));
+  mov(holder_offset, zr);
+  lea(scan_temp, Address(recv_klass, scan_temp, Address::lsl(3)));
```
eastig, line 1233: The comments to the code help little to understand the origins of the generated code.
I understand the generated code corresponds to:
```
temp_itbl_klass = (itableOffsetEntry*)(recv_klass->start_of_itable())[0].interface_klass()
```
which is
```
temp_itbl_klass = *(recv_klass + Klass::vtable_start_offset() + sizeof(intptr_t)*recv_klass->_vtable_length + itableOffsetEntry::offset_offset())
```

You store `recv_klass + Klass::vtable_start_offset() +itableOffsetEntry::offset_offset()` into `recv_klass`.
BTW, you assume `sizeof(intptr_t) == 8` to use `Address::lsl(3)`. Should this be mentioned in comments? 
You store `recv_klass + sizeof(intptr_t)*recv_klass->_vtable_length` into `scan_temp` to access `itable[i].interface_klass()`.

I think the code is worth adding more comments.

src/hotspot/cpu/aarch64/vtableStubs_aarch64.cpp
simonis: Agree.

simonis (commented):
I wonder if you can get rid of `temp_reg3` (and the `holder_offset` argument of `MacroAssembler::lookup_interface_method_stub()`) by defining `resolved_klass_reg` as `r17` in `VtableStubs::create_itable_stub()` and defining `Register holder_offset = method_result` as local variable in `MacroAssembler::lookup_interface_method_stub()` (the usages of `holder_offset` and `method_result` don't seem to overlap).

Otherwise looks good.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1195,6 +1195,112 @@ void MacroAssembler::lookup_interface_method(Register recv_klass,
...
+  //   goto L_no_such_interface // Not found.
+  //
+  Label L_save_holder_offset;
+  bind(L_save_holder_offset);
+    mov(holder_offset, scan_temp);
+    // continue searching for resolved_found (checked above: resolved found != resolved class)
+  Label L_loop_scan_resolved;
+  bind(L_loop_scan_resolved);
+    ldr(temp_itbl_klass, Address(pre(scan_temp, scan_step)));
+    bind(L_loop_scan_resolved_entry);
```
eastig, line 1283: Is the indentation of `bind` correct?

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1195,6 +1195,112 @@ void MacroAssembler::lookup_interface_method(Register recv_klass,
...
+                                                  Register method_result,
+                                                  Register temp_itbl_klass,
+                                                  Register scan_temp,
+                                                  Register holder_offset,
+                                                  int itable_index,
+                                                  Label& L_no_such_interface) {
+  assert_different_registers(resolved_klass, recv_klass, holder_klass, temp_itbl_klass, scan_temp, holder_offset);
+  assert_different_registers(method_result, recv_klass, holder_klass, temp_itbl_klass, scan_temp, holder_offset);
+
+  int vtable_base = in_bytes(Klass::vtable_start_offset());
```
eastig, line 1217: Why not to use `vtable_start_offset` to be explicit?

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1195,6 +1195,112 @@ void MacroAssembler::lookup_interface_method(Register recv_klass,
...
+  int itentry_off = in_bytes(itableMethodEntry::method_offset());
```
eastig, line 1218: `itmentry_method_offset`?

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1195,6 +1195,112 @@ void MacroAssembler::lookup_interface_method(Register recv_klass,
...
+  int scan_step = itableOffsetEntry::size() * wordSize;
```
eastig, line 1219: `itoentry_size_in_bytes`?

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1195,6 +1195,112 @@ void MacroAssembler::lookup_interface_method(Register recv_klass,
...
+  int vte_size = vtableEntry::size_in_bytes();
+  int ioffset = in_bytes(itableOffsetEntry::interface_offset());
```
eastig, line 1221: `itoentry_interface_offset`?

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1195,6 +1195,112 @@ void MacroAssembler::lookup_interface_method(Register recv_klass,
...
+  int ooffset = in_bytes(itableOffsetEntry::offset_offset());
```
eastig, line 1222: `itoentry_offset_offset`?

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1195,6 +1195,112 @@ void MacroAssembler::lookup_interface_method(Register recv_klass,
...
+
+  // Loop: Look for holder_klass record in itable
+  //   do {
+  //     temp_itbl_klass = *(scan_temp += scan_step);
+  //     if (temp_itbl_klass == holder_klass) {
+  //       goto L_holder_found; // Found!
+  //     }
+  //   } while (temp_itbl_klass != 0);
+  //   goto L_no_such_interface // Not found.
+  Label L_scan_holder;
```
eastig, line 1253: `L_search_holder_class`?

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1195,6 +1195,112 @@ void MacroAssembler::lookup_interface_method(Register recv_klass,
...
+// The receiver klass is in recv_klass.
+// On success, the result will be in method_result, and execution falls through.
+// On failure, execution transfers to the given label.
+void MacroAssembler::lookup_interface_method_stub(Register recv_klass,
+                                                  Register holder_klass,
+                                                  Register resolved_klass,
+                                                  Register method_result,
+                                                  Register temp_itbl_klass,
+                                                  Register scan_temp,
+                                                  Register holder_offset,
```
eastig, line 1211: `holder_klass_itoentry`?

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1195,6 +1195,112 @@ void MacroAssembler::lookup_interface_method(Register recv_klass,
...
+  //   do {
+  //     temp_itbl_klass = *(scan_temp += scan_step);
+  //     if (temp_itbl_klass == holder_klass) {
+  //        goto L_save_holder_offset;
+  //     }
+  //     if (temp_itbl_klass == resolved_klass) {
+  //        goto L_resolved_found;  // Found!
+  //     }
+  //   } while (temp_itbl_klass != 0);
+  //   goto L_no_such_interface // Not found.
```
eastig, line 1274: How about the following variant, it will have back jumps to one point instead of two:
```
bind(L_loop_scan_resolved);
   ldr(temp_itbl_klass, Address(pre(scan_temp, scan_step)));
bind(L_loop_scan_resolved_entry);
   cbz(temp_itbl_klass, L_no_such_interface);
   cmp(resolved_klass, temp_itbl_klass);
   br(Assembler::EQ, L_resolved_found);
   cmp(holder_klass, temp_itbl_klass);
   br(Assembler::NE, L_loop_scan_resolved);
   mov(holder_offset, scan_temp);
   b(L_loop_scan_resolved);
```

src/hotspot/cpu/aarch64/vtableStubs_aarch64.cpp
bulasevich: Done. Thanks

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
bulasevich: OK

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
bulasevich: The value is applied once. I removed the variable and use `in_bytes(itableMethodEntry::method_offset())` in place.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
bulasevich: Let it be itable_offset_entry_size

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
bulasevich: The correct explicit name would be itable_offset_entry_interface_offset, which I think is hard to read. if you don't mind, let us keep the short name for this case:
```
  int vtable_start_offset = in_bytes(Klass::vtable_start_offset());
  int itable_offset_entry_size = itableOffsetEntry::size() * wordSize;
  int ioffset = in_bytes(itableOffsetEntry::interface_offset());
  int ooffset = in_bytes(itableOffsetEntry::offset_offset());
```

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
bulasevich: Right. This variant is more natural and has the same performance. Thanks!

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
bulasevich: OK

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
bulasevich: fixed. thanks!

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
bulasevich: A simple cpp code explains things better than words. Let me put it in a way that explains data access, rather than strictly following a sequence of register manipulation:
```cpp
  int vtable_start_offset = in_bytes(Klass::vtable_start_offset());
  int ioffset = in_bytes(itableOffsetEntry::interface_offset());

  ldrw(scan_temp, Address(recv_klass, Klass::vtable_length_offset()));
  add(recv_klass, recv_klass, vtable_start_offset + ioffset);
  // itableOffsetEntry[] itable = recv_klass + Klass::vtable_start_offset() + recv_klass->_vtable_len;
  // temp_itbl_klass = itable[0]._interface;
  ldr(temp_itbl_klass, Address(recv_klass, scan_temp, Address::lsl(3)));
  // scan_temp = &(itable[0]._interface)
  lea(scan_temp, Address(recv_klass, scan_temp, Address::lsl(3)));
```

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1195,6 +1195,112 @@ void MacroAssembler::lookup_interface_method(Register recv_klass,
...
+  //   - if (holder_klass != resolved_klass), go to "scan for resolved"
+  //   - if (itable[0] == holder_klass), shortcut to "holder found"
+  //   - if (itable[0] == 0), no such interface
+  cmp(resolved_klass, holder_klass);
+  br(Assembler::NE, L_loop_scan_resolved_entry);
+  cmp(holder_klass, temp_itbl_klass);
+  br(Assembler::EQ, L_holder_found);
+  cbz(temp_itbl_klass, L_no_such_interface);
+
+  // Loop: Look for holder_klass record in itable
```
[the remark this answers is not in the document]
bulasevich: The structure of itable is now explained in comment above
`  // itableOffsetEntry[] itable = recv_klass + Klass::vtable_start_offset() + recv_klass->_vtable_len;`

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
bulasevich: In fact it holds an offset to a holder_class pointer in the itable, not the whole entry offset. Let us leave is as is.

src/hotspot/cpu/aarch64/vtableStubs_aarch64.cpp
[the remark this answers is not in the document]
bulasevich: Right. I created a separate issue for this task: https://bugs.openjdk.org/browse/JDK-8313870

bulasevich:
> I wonder if you can get rid of `temp_reg3` (and the `holder_offset` argument of `MacroAssembler::lookup_interface_method_stub()`) by defining `resolved_klass_reg` as `r17` in `VtableStubs::create_itable_stub()` and defining `Register holder_offset = method_result` as local variable in `MacroAssembler::lookup_interface_method_stub()` (the usages of `holder_offset` and `method_result` don't seem to overlap).

I think it does not work. Non-overlapping registers are already optimized: resolved_klass and method_result are actually reusing the same rmethod(r12) register.

> Otherwise looks good.

Thanks!

eastig (approved):
LGTM. Thanks for the great job.

simonis:
> > I wonder if you can get rid of `temp_reg3` (and the `holder_offset` argument of `MacroAssembler::lookup_interface_method_stub()`) by defining `resolved_klass_reg` as `r17` in `VtableStubs::create_itable_stub()` and defining `Register holder_offset = method_result` as local variable in `MacroAssembler::lookup_interface_method_stub()` (the usages of `holder_offset` and `method_result` don't seem to overlap).
> 
> I think it does not work. Non-overlapping registers are already optimized: resolved_klass and method_result are actually reusing the same rmethod(r12) register.
> 
> > Otherwise looks good.
> 
> Thanks!

Sorry if my comment was unclear. What I meant was this:
```patch
diff --git a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
index 8bf3b37f471..6e91df7156f 100644
--- a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
@@ -1209,11 +1209,13 @@ void MacroAssembler::lookup_interface_method_stub(Register recv_klass,
                                                   Register method_result,
                                                   Register temp_itbl_klass,
                                                   Register scan_temp,
-                                                  Register holder_offset,
                                                   int itable_index,
                                                   Label& L_no_such_interface) {
+  // 'method_result' is only used as output register at the very end of this method.
+  // Until then we can reuse it as 'holder_offset'.
+  Register holder_offset = method_result;
   assert_different_registers(resolved_klass, recv_klass, holder_klass, temp_itbl_klass, scan_temp, holder_offset);
-  assert_different_registers(method_result, recv_klass, holder_klass, temp_itbl_klass, scan_temp, holder_offset);
+  assert_different_registers(method_result, recv_klass, holder_klass, temp_itbl_klass, scan_temp);
 
   int vtable_start_offset = in_bytes(Klass::vtable_start_offset());
   int itable_offset_entry_size = itableOffsetEntry::size() * wordSize;
diff --git a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.hpp b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.hpp
index 72dd98ab617..004bed97b7f 100644
--- a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.hpp
+++ b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.hpp
@@ -949,7 +949,6 @@ class MacroAssembler: public Assembler {
                                     Register method_result,
                                     Register temp_reg,
                                     Register temp_reg2,
-                                    Register temp_reg3,
                                     int itable_index,
                                     Label& L_no_such_interface);
 
diff --git a/src/hotspot/cpu/aarch64/vtableStubs_aarch64.cpp b/src/hotspot/cpu/aarch64/vtableStubs_aarch64.cpp
index 73736566c90..6e345c96ce2 100644
--- a/src/hotspot/cpu/aarch64/vtableStubs_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/vtableStubs_aarch64.cpp
@@ -175,10 +175,9 @@ VtableStub* VtableStubs::create_itable_stub(int itable_index) {
   // so all registers except arguments are free at this point.
   const Register recv_klass_reg     = r10;
   const Register holder_klass_reg   = r16; // declaring interface klass (DECC)
-  const Register resolved_klass_reg = rmethod; // resolved interface klass (REFC)
+  const Register resolved_klass_reg = r17; // resolved interface klass (REFC)
   const Register temp_reg           = r11;
   const Register temp_reg2          = r15;
-  const Register temp_reg3          = r17;
   const Register icholder_reg       = rscratch2;
 
   Label L_no_such_interface;
@@ -195,7 +194,7 @@ VtableStub* VtableStubs::create_itable_stub(int itable_index) {
   // Receiver subtype check against REFC.
   // Get selected method from declaring class and itable index
   __ lookup_interface_method_stub(recv_klass_reg, holder_klass_reg, resolved_klass_reg, rmethod,
-                                  temp_reg, temp_reg2, temp_reg3, itable_index, L_no_such_interface);
+                                  temp_reg, temp_reg2, itable_index, L_no_such_interface);
 
   // Reduce "estimate" such that "padding" does not drop below 8.
   const ptrdiff_t estimate = 124;
```
I.e. make `resolved_klass_reg` different from `method_result` such that `method_result` can be reused as `holder_offset` because the two don't overlap. I leave it up to you if you want these changes. A quick check showed that they compile and successfully execute tier1 tests.

Otherwise still looks good.

simonis (approved):

shipilev (commented):
I find this version oddly deviating from [x86_64 improvement](https://github.com/openjdk/jdk/commit/8cdd95e8a2a7814ab7983fb3f41e6fa5793d410f) done before it, which does not simplify the review. Anyway, nits:

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1196,6 +1196,107 @@ void MacroAssembler::lookup_interface_method(Register recv_klass,
...
+  int ooffset = in_bytes(itableOffsetEntry::offset_offset());
+  assert(vtableEntry::size_in_bytes() == wordSize, "else adjust times_vte_scale");
+
+  Label L_loop_search_resolved_entry, L_resolved_found, L_holder_found;
+
+  ldrw(scan_temp, Address(recv_klass, Klass::vtable_length_offset()));
+  add(recv_klass, recv_klass, vtable_start_offset + ioffset);
+  // itableOffsetEntry[] itable = recv_klass + Klass::vtable_start_offset() + recv_klass->_vtable_len;
+  // temp_itbl_klass = itable[0]._interface;
+  ldr(temp_itbl_klass, Address(recv_klass, scan_temp, Address::lsl(3)));
```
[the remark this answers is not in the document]
bulasevich: This magic actually comes from the original MacroAssembler::lookup_interface_method code.
Let me use exact_log2(vtableEntry::size_in_bytes()) value here to show that it is related to vtable entry size:
```
  ldrw(scan_temp, Address(recv_klass, Klass::vtable_length_offset()));
  add(recv_klass, recv_klass, vtable_start_offset + ioffset);
  // itableOffsetEntry[] itable = recv_klass + Klass::vtable_start_offset() + sizeof(vtableEntry) * recv_klass->_vtable_len;
  // temp_itbl_klass = itable[0]._interface;
  ldr(temp_itbl_klass, Address(recv_klass, scan_temp, Address::lsl(exact_log2(vtableEntry::size_in_bytes()))));
```

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1196,6 +1196,107 @@ void MacroAssembler::lookup_interface_method(Register recv_klass,
...
+  mov(holder_offset, zr);
+  // scan_temp = &(itable[0]._interface)
+  lea(scan_temp, Address(recv_klass, scan_temp, Address::lsl(3)));
```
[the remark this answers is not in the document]
bulasevich: OK

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1196,6 +1196,107 @@ void MacroAssembler::lookup_interface_method(Register recv_klass,
...
+
+  // See if we already have a holder klass. If not, go and scan for it.
+  bind(L_resolved_found);
+  cbz(holder_offset, L_search_holder);
+  mov(scan_temp, holder_offset);
+
+  // Finally, scan_temp contains holder_klass vtable offset
+  bind(L_holder_found);
+  ldrw(method_result, Address(scan_temp, ooffset - ioffset));
+  add(recv_klass, recv_klass, (itable_index << 3) + in_bytes(itableMethodEntry::method_offset()) - vtable_start_offset - ioffset);
```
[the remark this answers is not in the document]
bulasevich: Let it be (itable_index << LogBytesPerWord)

bulasevich:
> I.e. make resolved_klass_reg different from method_result such that method_result can be reused as holder_offset because the two don't overlap

I see now. This is right, thank you!

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1225,12 +1226,12 @@ void MacroAssembler::lookup_interface_method_stub(Register recv_klass,
 
   ldrw(scan_temp, Address(recv_klass, Klass::vtable_length_offset()));
   add(recv_klass, recv_klass, vtable_start_offset + ioffset);
-  // itableOffsetEntry[] itable = recv_klass + Klass::vtable_start_offset() + recv_klass->_vtable_len;
+  // itableOffsetEntry[] itable = recv_klass + Klass::vtable_start_offset() + sizeof(vtableEntry) * recv_klass->_vtable_len;
   // temp_itbl_klass = itable[0]._interface;
-  ldr(temp_itbl_klass, Address(recv_klass, scan_temp, Address::lsl(3)));
+  ldr(temp_itbl_klass, Address(recv_klass, scan_temp, Address::lsl(exact_log2(vtableEntry::size_in_bytes()))));
```
eastig, line 1231: The shift in `ldr` must be 3.
Can we have a static assert checking that? There is an assert in `Address::encode` which will catch the issue in debug builds.

theRealAph: I'm very surprised that isn't a guarantee. I had to look to be sure. I'd simply change `assert(ext().shift() <= 0 || ext().shift() == (int)size, "bad shift");` to a `guarantee`.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1293,7 +1294,7 @@ void MacroAssembler::lookup_interface_method_stub(Register recv_klass,
   // Finally, scan_temp contains holder_klass vtable offset
   bind(L_holder_found);
   ldrw(method_result, Address(scan_temp, ooffset - ioffset));
-  add(recv_klass, recv_klass, (itable_index << 3) + in_bytes(itableMethodEntry::method_offset()) - vtable_start_offset - ioffset);
+  add(recv_klass, recv_klass, (itable_index << LogBytesPerWord) + in_bytes(itableMethodEntry::method_offset()) - vtable_start_offset - ioffset);
```
eastig, line 1297: Maybe use `itable_index * wordSize` instead of `itable_index << LogBytesPerWord` to look the same as in the x86 variant of `lookup_interface_method_stub`. 
Also a comment that `- vtable_start_offset - ioffset` is used to restore the original value of `recv_klass` would be useful.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
bulasevich: I want to avoid modifying the assembler_aarch64.hpp file within this change, so I add guarantee in lookup_interface_method.
By the way, [assembler_aarch64.hpp](https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/aarch64/assembler_aarch64.hpp) contains 100+ assert calls and 34 guarantee calls. I think more cases can be changed, although I am not sure what the rule is for using this or that call.

eastig: > I want to avoid modifying the assembler_aarch64.hpp file within this change, so I add guarantee in lookup_interface_method. By the way, [assembler_aarch64.hpp](https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/aarch64/assembler_aarch64.hpp) contains 100+ assert calls and 34 guarantee calls. I think more cases can be changed, although I am not sure what the rule is for using this or that call.

A JBS issue?

theRealAph: No, I don't think so.

The balance between `assert()` and `guarantee()` is a matter of judgement, not hard rules. 

We don't want to do consistency checking at runtime if we can avoid it, because it adds to startup time.

The `guarantee()` that an operand fits into its field in an instruction has proved its worth, many times. This is because AArch64 has a variety of offset fields, and these are  of varying lengths. It's very easy to be caught out, and sometimes this has happened even after release.

In most other places, an `assert()` is all we need. This is because the thing being asserted doesn't vary much, and we only need to check it once, when we run a debug build.

In the case of checking that the shift in a memory-access instruction matches the operand size, the check will be inlined. I haven't looked, but I expect it will be eliminated by the C++ optimizer at compile time.

In the case of the ldr instruction here, the operand size is determined already by the `ldr` instruction: it's `wordSize`. It cannot be anything else. The AArch64 back end asserts that elsewhere: 

```
  int vte_size    = vtableEntry::size_in_bytes();
  assert(vte_size == wordSize, "else adjust times_vte_scale");
```

bulasevich: I agree. I keep assert in MacroAssembler::lookup_interface_method_stub and change assert to guarantee in Address::encode.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -1228,10 +1228,12 @@ void MacroAssembler::lookup_interface_method_stub(Register recv_klass,
   add(recv_klass, recv_klass, vtable_start_offset + ioffset);
   // itableOffsetEntry[] itable = recv_klass + Klass::vtable_start_offset() + sizeof(vtableEntry) * recv_klass->_vtable_len;
   // temp_itbl_klass = itable[0]._interface;
-  ldr(temp_itbl_klass, Address(recv_klass, scan_temp, Address::lsl(exact_log2(vtableEntry::size_in_bytes()))));
+  int vtblEntrySize = vtableEntry::size_in_bytes();
+  assert(vtblEntrySize == wordSize, "ldr shift amount must be lsl#3");
```
eastig, line 1232: This assert relies on that `vtblEntrySize` is the same in debug and release builds.
In theory, the assert can be true  in a debug build and false in a release build.
As the `guarantee` in `encode` will work for both debug and release builds, having the assert here looks redundant.
If we want to have an assert here, I'd use the `static_assert`:
```
static_assert(sizeof(vtableEntry) == wordSize, "ldr lsl shift amount must be 3");
```

BTW, "lsl#3" in the error message "ldr shift amount must be lsl#3" looks confusing to me.

simonis (approved):
Still good.

eastig (approved):
Overall lgtm

bulasevich:
@eastig, @simonis, @shipilev, @theRealAph Thanks for reviewing! If there are no more comments, I will integrate it.

theRealAph (approved):

eastig (approved):
lgtm

PR #13795, opened 2023-05-04
JDK-8307331: Correctly update line maps when class redefine rewrites bytecodes

adinn:
This small change ensures that repeated bytecode rewrites necessitated by class pool index updates are applied cumulatively when updating the method line number table. The current code applies each change to the original table which means only the last one is applied (and even then with the wrong adjustment).

---------
### Progress

### Issue
 * [JDK-8307331](https://bugs.openjdk.org/browse/JDK-8307331): Correctly update line maps when class redefine rewrites bytecodes

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13795/head:pull/13795` \
`$ git checkout pull/13795`

`$ git checkout pull/13795` \
`$ git pull https://git.openjdk.org/jdk.git pull/13795/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13795`

`$ git pr show -t 13795`

</details>

<a href="https://git.openjdk.org/jdk/pull/13795.diff">https://git.openjdk.org/jdk/pull/13795.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13795#issuecomment-1534403851)

### DIFF
--- a/src/hotspot/share/runtime/relocator.cpp
+++ b/src/hotspot/share/runtime/relocator.cpp
@@ -407,11 +407,24 @@ void Relocator::adjust_exception_table(int bci, int delta) {
   }
 }
 
+static void print_linenumber_table(unsigned char* table) {
+  CompressedLineNumberReadStream stream(table);
+  tty->print_cr("-------------------------------------------------");
+  while (stream.read_pair()) {
+    tty->print_cr("   - line %d: %d", stream.line(), stream.bci());
+  }
+  tty->print_cr("-------------------------------------------------");
+}
 
 // The width of instruction at "bci" is changing by "delta".  Adjust the line number table.
 void Relocator::adjust_line_no_table(int bci, int delta) {
   if (method()->has_linenumber_table()) {
-    CompressedLineNumberReadStream  reader(method()->compressed_linenumber_table());
+    // if we already made adjustments then use the updated table
+    unsigned char *table = compressed_line_number_table();
+    if (table == nullptr) {
+      table = method()->compressed_linenumber_table();
+    }
+    CompressedLineNumberReadStream  reader(table);
     CompressedLineNumberWriteStream writer(64);  // plenty big for most line number tables
     while (reader.read_pair()) {
       int adjustment = (reader.bci() > bci) ? delta : 0;
@@ -420,6 +433,10 @@ void Relocator::adjust_line_no_table(int bci, int delta) {
     writer.write_terminator();
     set_compressed_line_number_table(writer.buffer());
     set_compressed_line_number_table_size(writer.position());
+    if (TraceRelocator) {
+      tty->print_cr("Adjusted line number table");
+      print_linenumber_table(compressed_line_number_table());
+    }
   }
 }
 


adinn:
This fix has been tested manually using the test case linked from the JIRA. An automated test might be possible but is difficult to perform as it would involve hot swapping code and checking line number adjustments via JDWP. Advice from the serviceability team on how to implement such a test would be welcome.

adinn:
@coleenp @plummercj Any chance of feedback or a review for this patch?

DanHeidinga:
@adinn Looking at the closely related code, is the same problem present for `adjust_exception_table` & `adjust_local_var_table`?  Both appear to always reach for the original value from the method, though unlike the line number table, there's no member variable cached in the Relocator for either of them.

plummercj:
@adinn This is not code I'm at all familiar with. Perhaps @sspitsyn can help.

sspitsyn (approved):
Looks good.
Thank you for taking care about it!
Thanks,
Serguei

adinn:
> @adinn Looking at the closely related code, is the same problem present for `adjust_exception_table` & `adjust_local_var_table`? Both appear to always reach for the original value from the method, though unlike the line number table, there's no member variable cached in the Relocator for either of them.

@DanHeidinga  I also thought that at first -- but it turns out the answer is no. Those other cases are different because the relocator directly updates the relevant offsets in the data located at the end of the ConstMethod. The line number table needs to be uncompressed and recompressed at each update, possibly ending up with a different size compressed array. So, the changes need to accumulate in a succession of compressed arrays held in the relocator the last of which gets recombined with the new version of the method  when the method clone eventually happens.

adinn:
@sspitsyn Thanks for the review.

dholmes-ora:
@adinn Please wait for two reviews for hotspot changes unless designated as trivial. Thanks.

adinn:
@dholmes-ora Apologies for jumping the gun. Do I need to obtain a second, retrospective review?

sspitsyn:
> Do I need to obtain a second, retrospective review?

Yes, Serviceability changes need two reviews unless it is not as trivial.

adinn:
@sspitsyn @dholmes-ora Could you suggest a suitable reviewer?

dholmes-ora:
I can't specifically suggest anyone but will try to find someone. I looked but don't know the code.

adinn:
@dholmes-ora Thanks. Perhaps @coleenp might be able to review it?

tstuefe (commented):
This looks fine to me. Thank you for the well written description text in the JBS issue.

I hesitated a bit at the use of TraceRelocator instead of our usual LogStream/LogTarget combination with UL, but I see that this is pre-existing usage, so its fine.

Another preexisting issue I noted is that CompressedWriteStream manages its internal buffer via RA, and grows inside RA too. That is suboptimal since we usually cannot reclaim the old buffer on resize since its rarely the top arena allocation. Oh, we also just allocate, we never even try resizing... (CompressedWriteStream::grow()).

coleenp (commented):
Sorry I didn't see this in my mail until just now.  Wow, good analysis and makes sense.  Late second review.

PR #13796, opened 2023-05-04
8307428: jstat tests doesn't tolerate dash in the O column

stefank:
When running jstat tests like the following:
test/jdk/sun/tools/jstatd/TestJstatdServer.java

with Generational ZGC we get a failure because the O (old generation percentage) is reported as `-` and not a number. The reason why it is reported as `-` is that the current capacity of the old generation is zero and that leads to a divide-by-zero in this line:
https://github.com/openjdk/jdk/blob/82a8e91ef7c3b397f9cce3854722cfe4bace6f2e/src/jdk.jcmd/share/classes/sun/tools/jstat/resources/jstat_options#L1029

G1 has some workarounds for this situation where the reported capacity is slightly above 0. I'm a bit reluctant to add such a hack into Generational ZGC. I've talked to the jstat maintainers and they propose that we simply relax the test.

Tested locally by running the jstat/jstad tests in the Generational ZGC branch.

---------
### Progress

### Issue
 * [JDK-8307428](https://bugs.openjdk.org/browse/JDK-8307428): jstat tests doesn't tolerate dash in the O column

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13796/head:pull/13796` \
`$ git checkout pull/13796`

`$ git checkout pull/13796` \
`$ git pull https://git.openjdk.org/jdk.git pull/13796/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13796`

`$ git pr show -t 13796`

</details>

<a href="https://git.openjdk.org/jdk/pull/13796.diff">https://git.openjdk.org/jdk/pull/13796.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13796#issuecomment-1534425606)

### DIFF
--- a/test/jdk/sun/tools/jstatd/JstatGCUtilParser.java
+++ b/test/jdk/sun/tools/jstatd/JstatGCUtilParser.java
@@ -46,7 +46,7 @@ public class JstatGCUtilParser {
         S0(GcStatisticsType.PERCENTAGE_OR_DASH),
         S1(GcStatisticsType.PERCENTAGE_OR_DASH),
         E(GcStatisticsType.PERCENTAGE_OR_DASH),
-        O(GcStatisticsType.PERCENTAGE),
+        O(GcStatisticsType.PERCENTAGE_OR_DASH),
         M(GcStatisticsType.PERCENTAGE_OR_DASH),
         CCS(GcStatisticsType.PERCENTAGE_OR_DASH),
         YGC(GcStatisticsType.INTEGER_OR_DASH),


tschatzl:
Is it possible to remove the G1 hack in this change too?

stefank:
> Is it possible to remove the G1 hack in this change too? Because since now a zero value is supported in the output, there does not seem to be a reason to keep it for G1.

I prefer if that is done and tested as a separate PR.

plummercj (approved):

stefank:
Thanks for reviewing!
/integrate

PR #13797, opened 2023-05-04
8323108: BufferedImage.setData(Raster) should not cast float and double values to integers

desruisseaux:
The `BufferedImage` Javadoc does not mention any constraint about the data type. In practice, `BufferedImage` with floating point values can be rendered by Java2D as well as integers, provided that a compatible `ColorModel` was supplied at construction time. However calls to `setData(Raster)` unexpectedly cast floating point values to integers. For example sample value 0.8 become 0. This is demonstrated by the `SetData` test case in this pull request.

An easy fix, which is proposed in this pull request, is to replace the whole `BufferedImage.setData(Raster)` method body by a simple call to `WritableRaster.setRect(Raster)`, which handles all `DataBuffer` types correctly. The method contracts documented in their respective Javadoc are compatible. Furthermore an examination of their source code shows that they are equivalent, except for the differences noted in the _Behavioural changes_ section below.

# Source code comparison
For demonstrating that delegating to `WritableRaster.setRect(Raster)` would be equivalent to the current code, one can copy the method body and apply the following changes:

* Remove `dx` parameter, replaced by 0.
* Remove `dy` parameter, replaced by 0.
* Rename `srcRaster` parameter as `r` (like in `BufferedImage`).
* Rename `startY` variable as `i` (like in `BufferedImage`).
* Rename `srcOffX` variable as `startX` (like in `BufferedImage`).
* Rename `srcOffY` variable as `startY` (like in `BufferedImage`).
* Replace `this.minX` by 0 and simplify.
* Replace `this.minY` by 0 and simplify.
* Remove the `switch` statement, keeping only the `TYPE_INT` case.

Then compare with `BufferedImage.setData(Raster)`. The modified block of code from `WritableRaster` is:

```
int width  = r.getWidth();
int height = r.getHeight();
int startX = r.getMinX();
int startY = r.getMinY();
```
We can see that above code is identical to `BufferedImage`. The next modified block of code from `WritableRaster`:

```
int dstOffX = startX;
int dstOffY = startY;

// Clip to this raster
if (dstOffX < 0) {
    width += dstOffX;
    startX -= dstOffX;    // = 0 because dstOffX == startX
    dstOffX = 0;
}
if (dstOffY < 0) {
    height += dstOffY;
    startY -= dstOffY;    // = 0 because dstOffY == startY
    dstOffY = 0;
}
if (dstOffX+width > this.width) {
    width = this.width - dstOffX;
}
if (dstOffY+height > this.height) {
    height = this.height - dstOffY;
}
if (width <= 0 || height <= 0) {
    return;
}
```

Above is equivalent to the following code from `BufferedImage`, except when there is integer underflow (see _Behavioural changes_ below):

```
// Clip to the current Raster
Rectangle rclip = new Rectangle(startX, startY, width, height);
Rectangle bclip = new Rectangle(0, 0, raster.width, raster.height);
Rectangle intersect = rclip.intersection(bclip);
if (intersect.isEmpty()) {
    return;
}
width = intersect.width;
height = intersect.height;
startX = intersect.x;
startY = intersect.y;
```

Next modified block of code from `WritableRaster`:

```
int[] iData = null;
for (int i=0; i < height; i++) {
    // Grab one scanline at a time
    iData = r.getPixels(startX, startY+i, width, 1, iData);
    setPixels(dstOffX, dstOffY+i, width, 1, iData);
}
```

Above is equivalent to `BufferedImage` code (keeping in mind that `startX == dstOffX` and same for Y, because of the `BufferedImage` constraint on raster starting at 0,0):

```
int[] tdata = null;
// remind use get/setDataElements for speed if Rasters are
// compatible
for (int i = startY; i < startY+height; i++)  {
    tdata = r.getPixels(startX,i,width,1,tdata);
    raster.setPixels(startX,i,width,1, tdata);
}
```

# Behavioural changes
This pull request would change the `BufferedImage.setData(Raster)` behaviour in the following ways:

* When using the standard `WritableRaster` implementations provided by OpenJDK, performance improvements may be observed because the JDK overrides `WritableRaster.setRect(Raster)` in various internal subclasses such as `ByteInterleavedRaster`. The overridden methods try to replace the loop by calls to `System.arraycopy(…)`.
* If the user constructed a `BufferedImage` with his own `WritableRaster` implementation which overrides `setRect(Raster)`, his overridden method would now be invoked. It would be compatible with `BufferedImage` Javadoc however, which said _"This class relies on the data fetching and setting methods of `Raster`, and on the color characterization methods of `ColorModel`."_
* The current `BufferedImage` implementation is more robust to integer underflows, because it uses `Rectangle.intersection(Rectangle)` which is underflow-safe. This pull request would lost this safety, unless we modify `WritableRaster.setRect(Raster)` for using `Rectangle.intersection(…)` as well.

# Open questions

* Do we also modify the `WritableRaster.setRect(Raster)` implementation for preserving the integer underflow safety of current `BufferedImage` implementation? The change would be to replace current intersection calculation by call to `Rectangle.intersection(…)`.
* The issue described in this pull request also applies to `BufferedImage.getData()` method and its derivative. Should we fix them as well?

---------
### Progress

### Issue
 * [JDK-8323108](https://bugs.openjdk.org/browse/JDK-8323108): BufferedImage.setData(Raster) should not cast float and double values to integers (**Bug** - P4)

### Reviewers
 * [Phil Race](https://openjdk.org/census#prr) (@prrace - **Reviewer**) ⚠️ Review applies to [65eb0072](https://git.openjdk.org/jdk/pull/13797/files/65eb0072f1b98225748930cf03eda23adfadf73b)
 * [Brian Burkhalter](https://openjdk.org/census#bpb) (@bplb - **Reviewer**) ⚠️ Review applies to [65eb0072](https://git.openjdk.org/jdk/pull/13797/files/65eb0072f1b98225748930cf03eda23adfadf73b)
 * [Sergey Bylokhov](https://openjdk.org/census#serb) (@mrserb - **Reviewer**) ⚠️ Review applies to [65eb0072](https://git.openjdk.org/jdk/pull/13797/files/65eb0072f1b98225748930cf03eda23adfadf73b)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13797/head:pull/13797` \
`$ git checkout pull/13797`

`$ git checkout pull/13797` \
`$ git pull https://git.openjdk.org/jdk.git pull/13797/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13797`

`$ git pr show -t 13797`

</details>

<a href="https://git.openjdk.org/jdk/pull/13797.diff">https://git.openjdk.org/jdk/pull/13797.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13797#issuecomment-1880217019)


src/java.desktop/share/classes/java/awt/image/BufferedImage.java
```diff
@@ -1509,31 +1509,7 @@ public WritableRaster copyData(WritableRaster outRaster) {
...
-        int startX = r.getMinX();
-        int startY = r.getMinY();
-
-        int[] tdata = null;
-
-        // Clip to the current Raster
-        Rectangle rclip = new Rectangle(startX, startY, width, height);
-        Rectangle bclip = new Rectangle(0, 0, raster.width, raster.height);
-        Rectangle intersect = rclip.intersection(bclip);
-        if (intersect.isEmpty()) {
```
mrserb, line 1523: Does the "raster.setRect" has equivalent optimizations? Does it seem that method implements handcrafter "intersection"?

desruisseaux: I do not know why `WritableRaster.setRect(Raster)` computes intersection with its own code in the method body instead of delegating to `Rectangle.intersection(Rectangle)` in the same way as `BufferedImage.setData(Raster)` does. For now I see no technical reason. I'm tempted to add a commit to this pull request for replacing the intersection calculation of `WritableRaster.setRect(Raster)`, but abstained for now because I though that reviewers or sponsors may want to minimize the changes. However I believe that this replacement would be desirable at least for the extra robustness of `Rectangle` regarding underflow and overflow.

mrserb: Can we prove the overflow issue there by some test? If yes then probably it would be useful to fix it as well, probably even as separate bug which can be later backported.

desruisseaux: In theory the following code would demonstrate the overflow:

```java
var sm = new BandedSampleModel(DataBuffer.TYPE_BYTE, 100, 100, 1);
var r1 = Raster.createWritableRaster(sm, new Point(Integer.MAX_VALUE - 150, 0));
var r2 = Raster.createWritableRaster(sm, new Point(Integer.MAX_VALUE -  50, 0));    // Overflow on X.
r1.setRect(r2);
```

It practice it does not happen so easily because the `Raster` constructor detects in advance the overflow and throws a `RasterFormatException` saying _"overflow condition for X coordinates of Raster"_. However because the `Raster.minX` field is protected and non-final, the constructor check can be bypassed as below:

```java
var r2 = new WritableRaster(sm, new Point()) {
    {   // Constructor
        minX = Integer.MAX_VALUE - 50;
        sampleModelTranslateX = minX;
    }
};
```

In which case the overflow happens and manifest itself with an `ArrayIndexOutOfBoundsException`. I do not know if this is considered a too convolved case for being worth a fix. However using the `BufferedImage` code, in addition of being robust to this case, would also be simpler.

desruisseaux:
A comment for keeping this merge request alive. Reminder: it is about resolving data lost in calls to `BufferedImage.setData(Raster)` caused by unnecessary casts of sample values.

desruisseaux:
Ping for keeping this pull request alive. Can we have a reviewer for it, or open a discussion about whether to expand the issue scope as mentioned in the description?

Issue reminder: `BufferedImage.setData(Raster)` lost data when using floating point values, without anything in the Javadoc suggesting that it is the expected behaviour. The proposed fix would resolve the data lost, *simplify* the code, possibly improve performances (but it is not the main goal) and be closer to what javadoc suggests.

Open question: current pull request does the smallest amount of changes, but more simplifications are possible. Should we add them to this pull request?

mrserb:
This change looks fine to me as is, but I think it would be useful to eliminate the next issue with the same patch(it depends on the change size):
@prrace do you have a preference should we fix that in one or separate patches?

>Do we also modify the WritableRaster.setRect(Raster) implementation for preserving the integer underflow safety of current BufferedImage implementation? The change would be to replace current intersection calculation by call to Rectangle.intersection(…).

desruisseaux:
Thanks. Regarding the second issue (`getData()` and its derivatives), a closer look shows that I was wrong to said that it suffers the same problem than `setData(Raster)`. Only the latter needs to be fixed, so the current request should be sufficient.

Regarding the second open question about replacing current handcrafter intersection code by a call to `Rectangle.intersection(Rectangle)`, the following code:

```java
        if (dstOffX < this.minX) {
            int skipX = this.minX - dstOffX;
            width -= skipX;
            srcOffX += skipX;
            dstOffX = this.minX;
        }
        if (dstOffY < this.minY) {
            int skipY = this.minY - dstOffY;
            height -= skipY;
            srcOffY += skipY;
            dstOffY = this.minY;
        }
        if (dstOffX+width > this.minX+this.width) {
            width = this.minX + this.width - dstOffX;
        }
        if (dstOffY+height > this.minY+this.height) {
            height = this.minY + this.height - dstOffY;
        }
        if (width <= 0 || height <= 0) {
            return;
        }
```

would be replaced by the following (this code contains a call to `getBounds()` overrideable method):

```java
        Rectangle dstRect = new Rectangle(dstOffX, dstOffY, width, height);
        dstRect = dstRect.intersection(getBounds());
        if (dstRect.isEmpty()) {
            return;
        }
        srcOffX += dstRect.x - dstOffX;
        srcOffY += dstRect.y - dstOffY;
        dstOffX  = dstRect.x;
        dstOffY  = dstRect.y;
        width    = dstRect.width;
        height   = dstRect.height;
```

Would it be a worthly change?

mrserb:
>would be replaced by the following (this code contains a call to getBounds() overrideable method):

That call could be eliminated if you create the new rect using "minX, minY, width, height". But the new code adds allocation of two new objects which immediately become garbage, probably we can change the code w/o using Rectangle?

desruisseaux:
I do not see variant of `Rectangle.intersection(…)` with `int` arguments [in the Javadoc](https://docs.oracle.com/en/java/javase/21/docs/api/java.desktop/java/awt/Rectangle.html). Maybe HotSpot can avoid the allocation with escape analysis. But anyway, the cost of `Rectangle` allocation is probably small comparatively to the cost of copying some thousands of pixel values.

mrserb:
> But anyway, the cost of Rectangle allocation is probably small comparatively to the cost of copying some thousands of pixel values.

But can we solve both issues, add optimization w/o creating the new objects?

desruisseaux:
> But can we solve both issues, add optimization w/o creating the new objects?

Not with the current `Rectangle` API as far as I can see. If the goal is maximal performances, the current code is likely better. The use of `Rectangle.intersection(Rectangle)` was proposed for safety rather than performance, because `Rectangle` is robust to integer underflow and overflow. However, I think that the current `WritableRaster.setRect(Raster)` code can underflow/overflow only if the checks performed at `Raster` construction time were bypassed, for example by modifying the protected fields after construction.

mrserb:
ok, let's use a solution based on the `Rectangle.intersection`

desruisseaux:
Added a commit doing the replacement of handcrafter intersection in `WritableRaster.setRect(…)` by a call to `Rectangle.intersection(…)`. I couldn't run the jtreg tests however. I tried for a few hours, but I presume that I didn't configured correctly.

Note that `setRect(…)` is overridden in `sun.awt.image.ByteInterleavedRaster` and `sun.awt.image.BytePackedRaster` for optimization purposes. Those methods use the same handcrafter intersection code than the one replaced by this commit. However I didn't updated the code in those subclasses. They should not be subject to underflow / overflow because users should not extend these classes and thus access to the protected fields. We could update those methods anyway for consistency, but I do not know what is the OpenJDK policy in that case.

Regarding `BufferedImage.getData()`, I remember the issue. That method does not have the lost of precision issue documented in this pull request. But it would nevertheless benefit from the same change (delegate to `WritableRaster.setRect(…)`) for performance reasons, because of optimizations in above-cited `ByteInterleavedRaster` and `BytePackedRaster` subclasses.

mrserb:
keep open

prrace (approved):
This all seems reasonable to me, and passes my testing.
I would however like Brian Burkhalter to take a look before it is integrated

mrserb:
@desruisseaux please enable the github actions in your fork, it seems the "Pre-submit test status" was skipped.
https://github.com/openjdk/jdk/pull/13797/checks?check_run_id=18627782594

desruisseaux:
Thanks, done. Waiting for the "okay" from Brian before to launch "/integrate".

bplb (approved):
This looks all right to me except I think the copyright years might need to be updated.

desruisseaux:
Updated the copyright years to 2024. The pre-submit tests are failing, but I believe that it is because the branch is 8 months old. Should I commit a merge from master? (a test on my local machine suggests that it fixes the issues).

mrserb:
> Updated the copyright years to 2024. The pre-submit tests are failing, but I believe that it is because the branch is 8 months old. Should I commit a merge from master? (a test on my local machine suggests that it fixes the issues).

Yes, you can merge the master.

PR #13803, opened 2023-05-04
8280982:  [Wayland] [XWayland] java.awt.Robot taking screenshots

azvegint:
Modern Linux systems often come with [Wayland](https://wayland.freedesktop.org/) by default.
This comes with some difficulties, and one of them is the inability to get screenshots from the system.
This is because we now use the [X Window System API](https://en.wikipedia.org/wiki/X_Window_System) to capture screenshots and it cannot access data outside the [XWayland server](https://wayland.freedesktop.org/xserver.html) 

But this functionality is a very important part of automated testing.

At the moment there are two obvious solutions to this problem, and both use [xdg-desktop-portal](https://github.com/flatpak/xdg-desktop-portal):

1. [org.freedesktop.portal.Screenshot DBUS API](https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.Screenshot)
It has several drawbacks though:
+ It saves a screenshot to disk, which must be read and deleted(may add some delays depending on the type of a disk drive).
+ There is no way to disable the visual "screen flash" after screenshot
+ It asks a user confirmation to save a screenshot. This confirmation can be saved on Gnome 43+. 
Since we would like Ubuntu 22.04 LTS which comes with Gnome 42 this option is not acceptable for us because it would require user confirmation for each screenshot.
But we still can consider this option as a fallback.

2. [org.freedesktop.portal.ScreenCast](https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.ScreenCast)
It typically used by applications that need to capture the contents of the user's screen or a specific window for the purpose of sharing, recording, or streaming.
This might be a bit of overkill, but it avoids several of the problems mentioned in the Screenshot API.

+ implementation is more complicated comparing to Screenshot API
+ no intermediate file, screenshot data can be obtained from memory
+ Permission to make screenshots can be stored with [`restore_token`](https://flatpak.github.io/xdg-desktop-portal/#gdbus-method-org-freedesktop-portal-ScreenCast.SelectSources)

So this PR adds the ability to take screenshots using the ScreenCast API. This functionality is currently disabled by default.

This change also introduces some new behavior for the robot:
A system window now appears asking for confirmation from the user to capture the screen.
+ The user can refuse the screen capture completely. In this case a security exception will be thrown.
+ The user can allow a screen capture for all screens or some of them. For screens without permission there will be a black image.
+ If the user wishes to change their mind about the screens allowed to be captured, the user should use the new `Robot#resetScreenCapturePermission` method

Also added several system properties:
`awt.robot.screencastEnabled` false by default, uses the ScreenCast API if true
`awt.robot.screencastDebug` false by default, prints some debug information if true

For convenience, this change is divided into two commits:
1. added [pipewire headers](https://gitlab.freedesktop.org/pipewire/pipewire/) that are needed for the build
2. main changes 

Changes in the documentation are in a separate PR also for convenience.

At the moment, the planned supported systems are Ubuntu 22.04+, the latest Oracle Linux 9.1 and RHEL 9.1 has some issues with restore_token support.


make/modules/java.desktop/lib/Awt2dLibraries.gmk
```diff
@@ -191,6 +191,9 @@ ifeq ($(call isTargetOs, windows macosx), false)
 
     LIBAWT_XAWT_EXCLUDES := medialib debug
 
+    BUILD_LIBPIPEWIRE_HEADER_DIRS := \
```
erikj79, line 194: This variable name indicates that LIBPIPEWIRE is a library that we would build, but it seems to just be headers. I would suggest dropping the `BUILD_` prefix.

erikj79 (commented):
I would like to understand why we need to add third party headers to our source tree. Are these not generally available on Linux systems?

You say that the functionality isn't enabled by default, is that a runtime thing? It loos like it will be built unconditionally.

mrserb:
> If the user wishes to change their mind about the screens allowed to be captured, the user should use the new Robot#resetScreenCapturePermission method

Do we really need to add this new API? probably we can implement the feature w/o this new method?

src/java.desktop/unix/native/libawt_xawt/awt/screencast_pipewire.c
```diff
@@ -0,0 +1,806 @@
...
+        || !initXdgDesktopPortal()
+        || (pw.pwFd = getPipewireFd(token)) < 0) {
+        doCleanup();
+        return FALSE;
+    }
+
+    return TRUE;
+}
+
+static inline void convertRGBxToBGRx(int* in) {
```
mkartashev, line 138: Will this work correctly on little-endian ARM systems?

src/java.desktop/unix/native/libawt_xawt/awt/screencast_pipewire.c
```diff
@@ -0,0 +1,806 @@
...
+    }                                                               \
+} while(0);
+
+static gboolean loadSymbols() {
+    if (!glib_version_2_68) {
+        DEBUG_SCREENCAST("glib version 2.68+ required\n", NULL);
+        return FALSE;
+    }
+
+    pipewire_libhandle = dlopen("libpipewire-0.3.so.0",
```
mkartashev, line 602: There's a useful macro for wrapping library names: `VERSIONED_JNI_LIB_NAME("pipewire-0.3", "0")`

src/java.desktop/unix/native/libawt_xawt/awt/screencast_pipewire.c
```diff
@@ -0,0 +1,806 @@
...
+        jint jwidth,
+        jint jheight,
+        jintArray pixelArray,
+        jstring jtoken
+) {
+
+    GdkRectangle requestedArea = { jx, jy, jwidth, jheight};
+
+    const gchar *token = jtoken
+                         ? token = (*env)->GetStringUTFChars(env, jtoken, NULL)
```
mkartashev, line 722: The variable `token` is unnecessarily used in its own initializer here.

src/java.desktop/unix/native/libawt_xawt/awt/screencast_pipewire.c
```diff
@@ -0,0 +1,806 @@
...
+                            ? 0
+                            : screenProps->bounds.x - requestedArea.x;
+                    jsize start = jwidth * (preY + y) + preX;
+
+                    jsize len = captureArea.width;
+
+                    (*env)->SetIntArrayRegion(
+                            env, pixelArray,
+                            start, len,
+                            ((jint *) screenProps->captureData)
```
mkartashev, line 789: I'm not sure if endianness of `captureData` matches the expected endianness of `pixelArray`.

azvegint: For now, it is going to be supported on x86-64 only, but it is a good candidate for investigation.

src/java.desktop/unix/native/libawt_xawt/awt/screencast_pipewire.c
azvegint: Fixed.

src/java.desktop/unix/native/libawt_xawt/awt/screencast_pipewire.c
azvegint: Updated.

src/java.desktop/unix/native/libawt_xawt/awt/screencast_pipewire.c
azvegint: I haven't tested it, but I'll try to find a machine to test it on.

src/java.desktop/unix/native/libawt_xawt/awt/screencast_pipewire.c
```diff
@@ -0,0 +1,806 @@
...
+int DEBUG_SCREENCAST_ENABLED = FALSE;
+
+struct ScreenSpace screenSpace = {0};
+static struct PwLoopData pw = {0};
+
+extern JavaVM *jvm;
+jclass screencastHelperClass = NULL;
+jmethodID storeTokenMethodID = NULL;
+
+inline void debug_screencast(
```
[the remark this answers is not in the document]
azvegint: It is used by `screencast_portal.c` and `screencast_pipewire.c`.

src/java.desktop/unix/classes/sun/awt/screencast/ScreencastHelper.java
```diff
@@ -0,0 +1,74 @@
...
+        TOKEN_STORE.put(TOKEN_KEY, token);
+    }
+
+    private static String getToken() {
+        return TOKEN_STORE.get(TOKEN_KEY, null);
+    }
+
+    private static final int DENIED = -11; //defined in native code
+
+    private static synchronized native int getRGBPixelsImpl(
```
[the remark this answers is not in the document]
azvegint: This code was written with the understanding that it will not run in parallel, so the `synchronized` is used.

Moreover, we will get a significant benefit from removing `synchronized` only if `getRGBPixelsImpl` is called frequently from different threads. 
If it is called from a single thread(which seems to me to be the main use case), the results of my quick tests with JMH show that the difference between synchronized and not synchronized is within the margin of error.

Removing the `synchronized` keyword would require additional testing, which at this point I want to avoid.
And since this is not a public API we have no problem revisiting it later.

P.S. I removed `synchronized` modifier from `getRGBPixels`

src/java.desktop/unix/classes/sun/awt/screencast/ScreencastHelper.java
```diff
@@ -0,0 +1,74 @@
...
+    // called from native
+    private static void storeToken(String token) {
+        TOKEN_STORE.put(TOKEN_KEY, token);
+    }
+
+    private static String getToken() {
+        return TOKEN_STORE.get(TOKEN_KEY, null);
+    }
+
+    private static final int DENIED = -11; //defined in native code
```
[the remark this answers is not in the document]
azvegint: removed.

make/modules/java.desktop/lib/Awt2dLibraries.gmk
azvegint: Updated.

src/java.desktop/unix/classes/sun/awt/screencast/ScreencastHelper.java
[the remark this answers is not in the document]
mkartashev: > P.S. I removed synchronized modifier from getRGBPixels
My main point was that you don't seem to need to have both, and while I picked one you picked the other ;-) So it's OK.

src/java.desktop/unix/native/libawt_xawt/awt/screencast_pipewire.c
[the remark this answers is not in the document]
mkartashev: OK, then it actually does need to be externally visible. Thanks!

src/java.desktop/unix/native/libawt_xawt/awt/screencast_pipewire.c
mkartashev: Is this enforced somehow? I mean what happens if this code is built on risc-v or arm?

azvegint:
> > If the user wishes to change their mind about the screens allowed to be captured, the user should use the new Robot#resetScreenCapturePermission method
> 
> Do we really need to add this new API? probably we can implement the feature w/o this new method?

I couldn't think of anything better. 
We need to somehow let the robot know that the saved `restore_token` no longer meets the user's expectations.
It seems that the current robot API is not suitable for this.

mrserb:
> It seems that the current robot API is not suitable for this.

azvegint:
> > It seems that the current robot API is not suitable for this.
> 
> Can we just reset the token when we create a new instance of Robot?

This will be a big problem for automated testing, as each test with the robot will require user confirmation(or even more than one, as some tests instantiate it several times).
This is exactly what we are trying to avoid.

prrace:
I'm thinking about the re-set API

"If the user wishes to change their mind about the screens allowed to be captured, the user should use the new Robot#resetScreenCapturePermission method"

Well the "user" is hosed unless the app calls this every time ... or provides UI for it ..

I want to work through the scenarios and how much of it is specific to the behaviours of the API you are using and so forth.
Since you use the Preferences API for saving the token, if you keyed it off the Robot class rather than the internal API class
then anyone using the Preferences API could delete the token, couldn't they ?

And we presumably can store alongside the token, info about the screens attached at the time we obtained the token ?

Can you point (here in the implementation review) to the exact native functions that popup up the dialog ?
I'm still trying to find my way through that code. Is it opaque to the caller (us) what the user actually approved ?

I can't imagine why we would need to reset "we have all permissions" to "we have no permissions" so I'm supposing
the issue is when they didn't grant permissions .. but if they "deny" permissions that's surely different than "this
token is no longer valid" ?

The first time we need a permission for the screencast API there is a dialog.
The user makes some choices and we then choose to save a token so that the user isn't prompted again.
We must at least know though that they approved SOMETHING but we don't know what ?
Or do we ? 
If they DENY, would we save that ? Interesting question because if we don't save something we'll keep spamming them until they approve :-)

Next time, what happens if for some reaon the token is no longer valid ?
And what might make it invalid ?

If the user denied screen 2, or added screen 2 later,  would that not result in an invalid token and we'd know
to ask again ? No reset needed ? Also I can't figure out if "validating" the token can cause the dialog as a side effect
or if there's some API we explicitly call to request a new token.

azvegint:
> I want to work through the scenarios and how much of it is specific to the behaviours of the API you are using and so forth. Since you use the Preferences API for saving the token, if you keyed it off the Robot class rather than the internal API class then anyone using the Preferences API could delete the token, couldn't they ?

Right now it can be read with a simple call:
`System.out.println(Preferences.userRoot().node("/sun/awt/screencast").get("restoreToken", ""));` (modification is also simple).

Moving to Robot doesn't change anything in this regard. So anyone can delete it.


> And we presumably can store alongside the token, info about the screens attached at the time we obtained the token ?

We can try to do that and switch it depending on the area requested, but I tried to keep it simple.

> Can you point (here in the implementation review) to the exact native functions that popup up the dialog ? I'm still trying to find my way through that code. Is it opaque to the caller (us) what the user actually approved ?

(It is collapsed by the github in the review, so links are to the branch)

basically if you [didn't provide a correct token or no token at all](https://github.com/azvegint/jdk/blob/482471df61490120590bc49886f76dc99c3d7dd8/src/java.desktop/unix/native/libawt_xawt/awt/screencast_portal.c#L530-L537) to [SelectSources call](https://github.com/azvegint/jdk/blob/482471df61490120590bc49886f76dc99c3d7dd8/src/java.desktop/unix/native/libawt_xawt/awt/screencast_portal.c#L539-L546) it will prompt with dialog right after [the Start call](https://github.com/azvegint/jdk/blob/482471df61490120590bc49886f76dc99c3d7dd8/src/java.desktop/unix/native/libawt_xawt/awt/screencast_portal.c#L677-L685)

> I can't imagine why we would need to reset "we have all permissions" to "we have no permissions" so I'm supposing the issue is when they didn't grant permissions .. but if they "deny" permissions that's surely different than "this token is no longer valid" ?

It is not for the  "we have all permissions" case, it is for the case when we only have a permission to some screens. So it won't hurt.

> The first time we need a permission for the screencast API there is a dialog. The user makes some choices and we then choose to save a token so that the user isn't prompted again. We must at least know though that they approved SOMETHING but we don't know what ?  Or do we ?

No we don't.

We receive the response [callbackScreenCastStart](https://github.com/azvegint/jdk/blob/482471df61490120590bc49886f76dc99c3d7dd8/src/java.desktop/unix/native/libawt_xawt/awt/screencast_portal.c#L586-L591) as a separate streams for each allowed display. 

Based on this response we [building a list of **available** displays](https://github.com/azvegint/jdk/blob/482471df61490120590bc49886f76dc99c3d7dd8/src/java.desktop/unix/native/libawt_xawt/awt/screencast_portal.c#L73).
So denied screens will not be there.

At this point we can compare this result with the screens got from X11 API(won't be compatible with Wayland, where we also want to reuse it later) or with Wayland(+1 dependency at this point).

> If they DENY, would we save that ? Interesting question because if we don't save something we'll keep spamming them until they approve :-)

The [DENY case](https://github.com/azvegint/jdk/blob/482471df61490120590bc49886f76dc99c3d7dd8/src/java.desktop/unix/native/libawt_xawt/awt/screencast_portal.c#L586-L591) propagated to java and we are throwing the security exception.

> Next time, what happens if for some reaon the token is no longer valid ?

It just prompts a permission request again.

> And what might make it invalid ?

Adding or removing a display for example.


> If the user denied screen 2, or added screen 2 later, would that not result in an invalid token and we'd know to ask again ? No reset needed ? Also I can't figure out if "validating" the token can cause the dialog as a side effect or if there's some API we explicitly call to request a new token.

Let me elaborate how this `restore_token` works:

We may or may not provide the `restore_token` while calling `SelectSources`.

If the token provided is not valid (we have no control over this, it's checked on the system side), the permission prompt will appear.

Later, after granting permissions, the `restore_token` is returned by [Start](https://flatpak.github.io/xdg-desktop-portal/#gdbus-method-org-freedesktop-portal-ScreenCast.Start) callback. We should [update](https://github.com/azvegint/jdk/blob/482471df61490120590bc49886f76dc99c3d7dd8/src/java.desktop/unix/native/libawt_xawt/awt/screencast_portal.c#L593-L611) it, as it may not be the same as we provided before.

When we are calling the `resetScreenCapturePermission` we are not actually revoking the token from the system, we are just doesn't add it to the `SelectSources` call.  (but, if needed, we can provide it later and it should work).

So whenever we don't provide a valid token, we get a confirmation prompt.

azvegint:
> > It is not for the "we have all permissions" case, it is for the case when we only have a permission to some screens. So it won't hurt.
> 
> But how does the app know when to call it ?

Right now it doesn't know. If the user is not satisfied with the result, he can "reset" it. But it seems to be a weak position.

> OK, so to make sure I understand, you use the token to see if you can obtain a stream for all connected displays. If any "fail" you know the user didn't approve that.
> 
> So we can do this every time right ? And in conjunction with knowing which screen[s] were connected when we captured the token we know the situation changed and so we know it is appropriate to ignore the stored token - causing the user to be re-prompted.
> 
> And even if the screens have changed if we already have the token we need for THIS screen capture we know whether we already have the permission we need. And if we do then we just use it and don't cause a user re-prompt.
> 
> At this point I think we've done everything that "revoke" was needed for.

The problem is how to relate the tokens to the screens. We don't have some universal id that says that the display we saved earlier is this current display.

There is an opaque id of stream:
https://flatpak.github.io/xdg-desktop-portal/#gdbus-method-org-freedesktop-portal-ScreenCast.Start
```
 Stream properties include:

id s

    Opaque identifier. **Will be unique for this stream and local to this session**. 
Will persist with future sessions, if they are restored using a restore token. 
This property was added in version 4 of this interface. Optional. 
```
But it didn't help much, as it local to the session(each token - different session), token looks like `0`, `1`, etc

We can match the stream layout to the screen layout in the system, but it has its flaws:

Let's say we have following layout of screens in system, displays have the same resolution(e.g. 1920*1080):
image

Screen capture permission is granted for `A` and denied for `B`.

1. On a first run we are capturing screen area at `(100, 100, 100, 100)`.
2. We match an acquired `restore_token` to the screen at `(0,0,1920,1080)`
3. User decided to change screens to following layout:
image
4. we are capturing screen area at `(100, 100, 100, 100)` again
5. Oh, we have a `restore_token` for the `(0,0,1920,1080)` screen, let's reuse it.
6. Ouch, we got black pixels because the token was issued for display `A` and it is now `(1920,0,1920,1080)`

> Then [if I have everything right] the only open question is whether when permissions to the required screen or screens was previously denied, do we re-prompt ? My take is that we do not "store" a token that denies access. So next time the app is run they will be re-prompted. But we _probably_ don't want to keep re-prompting within this running instance of the JDK.
> 
> BTW "the app" means "any app that can find this token, doesn't it" ? 

Yes, it is currently shared between Java applications for the same user.


> Also if we follow the suggested policy of not storing tokens which don't grant permission
then we'd also not over-write it in that case.

Currently we only overwrite tokens when we get a new one from the system. In the full DENY case, we do not receive any token.

azvegint:
Please review the changes to remove resetScreenCapturePermission

The token is now stored together with the bounds of the screens for which the screen capture was allowed.

Now when we want to take a screenshot, we create a list of the bounds of the system screens that have the data we're interested in.

Based on this list, we pick up the tokens from the previously saved ones:
a. exact match of the bounds of the screens
b. the same size(screens can be moved, while the old token for them works)

We try to get an image with these tokens, checking that the previously associated bounds have not changed and we get the image from the capture area available to us.
If, for some reason, the old token is withdrawn, a new permission will be requested to capture screens.

If none of the tokens worked, we try to get a new permission.

If none of above worked, we throw a SecurityException.

azvegint:
The system properties have been renamed:

`awt.robot.screenshotMethod` -> `awt.robot.screenshotMethod`
`awt.robot.screencastDebug` -> `awt.robot.screenshotDebug`

`awt.robot.screenshotMethod` - `dbusScreencast` defaults to Wayland session, `x11` defaults to X11 session, and to all other values except `dbusScreencast`.

Note 
<ol>
<li>setting this to `x11` in a Wayland session means screen capture will always fail, as it does in earlier JDKs</li>
<li>setting this to `dbusScreencast` in an X.org session should work but is only recommended for JDK debugging purposes</li>
<li>The already existing `awt.robot.gtk` option is only valid for `x11` and has no effect for `dbusScreencast`.</li>
</ol>


`awt.robot.screenshotDebug` - `false` by default, if `true` prints debug information related if any available.


This may be needed for a smoother experience when we later want to get a screenshot via 
[org.freedesktop.portal.Screenshot DBUS API](https://flatpak.github.io/xdg-desktop-portal/#gdbus-method-org-freedesktop-portal-Screenshot.Screenshot). (it requires Gnome 43+, Ubuntu 22.04 LTS has 42.5).

So we can add `dbusScreenshot` option for the `awt.robot.screenshotMethod` later.

src/java.desktop/unix/classes/sun/awt/screencast/ScreencastHelper.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    public static synchronized void getRGBPixels(
+            int x, int y, int width, int height, int[] pixelArray
+    ) {
+        if (!IS_NATIVE_LOADED) return;
+
+        Rectangle captureArea = new Rectangle(x, y, width, height);
+
+        List<Rectangle> affectedScreenBounds =  getSystemScreensBounds()
```
turbanoff, line 108: nit
```suggestion
        List<Rectangle> affectedScreenBounds = getSystemScreensBounds()
```

mkartashev:
Tried running this version on my machine (Ubuntu 22.04, two displays with 100% and 200% scaling). A few observations:
1. I couldn't get any of the screenshot tests working through `jtreg` (screenshots are all black, no permission is ever asked). Does anybody know how to do that properly? Can this be made to work out of the box?

2. Running manually is OK for the most part except for some tests in certain configurations:

* Test `test/jdk/java/awt/Robot/HiDPIScreenCapture/ScreenCaptureGtkTest.java` consistently fails with `sun.java2d.uiScale` other than `1`. For example,
```
$ java -Dsun.java2d.uiScale=3 ScreenCaptureGtkTest.java
Creating screen capture of java.awt.Rectangle[x=89,y=99,width=100,height=100]
Exception in thread "main" java.lang.SecurityException: Screen Capture in the selected area was not allowed
	at java.desktop/sun.awt.screencast.ScreencastHelper.getRGBPixels(ScreencastHelper.java:161)
	at java.desktop/sun.awt.X11.XRobotPeer.getRGBPixels(XRobotPeer.java:139)
	at java.desktop/java.awt.Robot.createCompatibleImage(Robot.java:606)
	at java.desktop/java.awt.Robot.createScreenCapture(Robot.java:477)
	at ScreenCaptureGtkTest.captureImageOf(ScreenCaptureGtkTest.java:130)
	at ScreenCaptureGtkTest.main(ScreenCaptureGtkTest.java:96)
```
and this is when I allow both screens to be captured in the system dialog that appears after the start of the test.
The same failure can be observed with `java -Dsun.java2d.uiScale=2 HiDPIRobotScreenCaptureTest.java`. Both tests hang after throwing those exceptions, by the way.

* The same test fails in a different fashion with `-Djdk.gtk.version=2`. For example:
```
$ java -Djdk.gtk.version=2 -Dsun.java2d.uiScale=1 ScreenCaptureGtkTest.java
WARNING: the GTK 2 library is deprecated and its support will be removed in a future release
Gtk-Message: 19:17:15.122: Failed to load module "canberra-gtk-module"
Creating screen capture of java.awt.Rectangle[x=89,y=99,width=100,height=100]
Could not load native libraries for ScreencastHelper
Checking color at 139, 149 to be equal to java.awt.Color[r=0,g=255,b=0]... Mismatch: found java.awt.Color[r=0,g=0,b=0] instead
Check image.png
Exception in thread "main" java.lang.RuntimeException: Wrong screen pixel color
	at ScreenCaptureGtkTest.checkPixelColors(ScreenCaptureGtkTest.java:115)
	at ScreenCaptureGtkTest.main(ScreenCaptureGtkTest.java:100)
```

* Test `test/jdk/java/awt/Robot/HiDPIScreenCapture/ScreenCaptureResolutionTest.java` fails and hangs like so:
```
$ java ScreenCaptureResolutionTest.java
Exception in thread "AWT-EventQueue-0" java.lang.RuntimeException: Exception while validating ScreenCapture
	at ScreenCaptureResolutionTest$2.run(ScreenCaptureResolutionTest.java:75)
	at java.desktop/java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:318)
	at java.desktop/java.awt.EventQueue.dispatchEventImpl(EventQueue.java:773)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:720)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:714)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:400)
	at java.base/java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:87)
	at java.desktop/java.awt.EventQueue.dispatchEvent(EventQueue.java:742)
	at java.desktop/java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:203)
	at java.desktop/java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:124)
	at java.desktop/java.awt.EventDispatchThread.pumpEventsForHierarchy(EventDispatchThread.java:113)
	at java.desktop/java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:109)
	at java.desktop/java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:101)
	at java.desktop/java.awt.EventDispatchThread.run(EventDispatchThread.java:90)
```

azvegint:
> Tried running this version on my machine (Ubuntu 22.04, two displays with 100% and 200% scaling). A few observations:
> 
>     1. I couldn't get any of the screenshot tests working through `jtreg` (screenshots are all black, no permission is ever asked). Does anybody know how to do that properly? Can this be made to work out of the box?

jtreg strips env variables down to a very few.
so providing following options should solve the issue:
`jtreg -ea -e:DBUS_SESSION_BUS_ADDRESS=$DBUS_SESSION_BUS_ADDRESS -e:WAYLAND_DISPLAY=$WAYLAND_DISPLAY ...`

This is already addressed by [openjdk/jtreg/pull/152](https://github.com/openjdk/jtreg/pull/152)

>     2. Running manually is OK for the most part except for some tests in certain configurations:
> 
> 
>     * Test `test/jdk/java/awt/Robot/HiDPIScreenCapture/ScreenCaptureGtkTest.java` consistently fails with `sun.java2d.uiScale` other than `1`. For example,
> 
> 
> ```
> $ java -Dsun.java2d.uiScale=3 ScreenCaptureGtkTest.java
> Creating screen capture of java.awt.Rectangle[x=89,y=99,width=100,height=100]
> Exception in thread "main" java.lang.SecurityException: Screen Capture in the selected area was not allowed
> 	at java.desktop/sun.awt.screencast.ScreencastHelper.getRGBPixels(ScreencastHelper.java:161)
> 	at java.desktop/sun.awt.X11.XRobotPeer.getRGBPixels(XRobotPeer.java:139)
> 	at java.desktop/java.awt.Robot.createCompatibleImage(Robot.java:606)
> 	at java.desktop/java.awt.Robot.createScreenCapture(Robot.java:477)
> 	at ScreenCaptureGtkTest.captureImageOf(ScreenCaptureGtkTest.java:130)
> 	at ScreenCaptureGtkTest.main(ScreenCaptureGtkTest.java:96)
> ```
> 
> and this is when I allow both screens to be captured in the system dialog that appears after the start of the test. The same failure can be observed with `java -Dsun.java2d.uiScale=2 HiDPIRobotScreenCaptureTest.java`. Both tests hang after throwing those exceptions, by the way.
> 

I'll take a look.

>     * The same test fails in a different fashion with `-Djdk.gtk.version=2`. For example:

This is intended, GTK2 is not supported and deprecated for removal. 
[JDK-8280031](https://bugs.openjdk.org/browse/JDK-8280031)

mkartashev:
> This is already addressed in jtreg 7.3 by https://github.com/openjdk/jtreg/pull/152

Thanks for the info!

> This is intended, GTK2 is not supported and deprecated for removal. [JDK-8280031](https://bugs.openjdk.org/browse/JDK-8280031)

That's fine, but the test still fails:
```
Exception in thread "main" java.lang.RuntimeException: Wrong screen pixel color
```
Deprecated shouldn't mean dysfunctional, so perhaps remove the line
```
 * @run main/othervm -Djdk.gtk.version=2 -Dsun.java2d.uiScale=1 ScreenCaptureGtkTest
```
from the test if there's no intention to make it work even (and especially) in the future. It will always fail on XWayland.
Or, if removing is hard for some reason, make it conditional and execute this test only in the pure X environment.

prrace:
macos builds are broken with this change.
The problem is unix/classes/sun/awt/screencast, causing unix/classes/sun/awt/X11 to be built but it fails because we don't (and shouldn't) generate the XAWT source on macOS.

You need to add screencast to the exclude list for macOS.

 git diff make/modules/java.desktop/Java.gmk
diff --git a/make/modules/java.desktop/Java.gmk b/make/modules/java.desktop/Java.gmk
index 19cd5b7da83..c9057c09c7f 100644
--- a/make/modules/java.desktop/Java.gmk
+++ b/make/modules/java.desktop/Java.gmk
@@ -68,6 +68,7 @@ EXCLUDE_FILES += \
 ifeq ($(call isTargetOs, macosx), true)
   # exclude all X11 on Mac.
   EXCLUDES += \
+      sun/awt/screencast \
       sun/awt/X11 \
       sun/java2d/x11 \
       sun/java2d/jules \

azvegint:
> macos builds are broken with this change. The problem is unix/classes/sun/awt/screencast, causing unix/classes/sun/awt/X11 to be built but it fails because we don't (and shouldn't) generate the XAWT source on macOS.
> 
> You need to add screencast to the exclude list for macOS.

Nice catch, fixed.

src/java.desktop/unix/classes/sun/awt/screencast/TokenStorage.java
```diff
@@ -0,0 +1,411 @@
...
+            store("remove malformed records");
+        }
+    }
+
+    private static void store(String failMsg) {
+        if (!isWritable()) {
+            return;
+        }
+
+        try (OutputStream output = Files.newOutputStream(PROPS_PATH)) {
```
prrace, line 387: Note that this stream is un-buffered (slow)
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/Files.html#newOutputStream(java.nio.file.Path,java.nio.file.OpenOption...)
Maybe doesn't matter given how small the file is.

src/java.desktop/unix/native/libawt_xawt/awt/screencast_pipewire.c
```diff
@@ -0,0 +1,915 @@
...
+#include "fp_pipewire.h"
+#include <stdio.h>
+
+#include "gtk_interface.h"
+#include "gtk3_interface.h"
+
+int DEBUG_SCREENCAST_ENABLED = FALSE;
+
+#define EXCEPTION_CHECK_DESCRIBE_CLEAR() if ((*env)->ExceptionCheck(env)) { \
+                                            (*env)->ExceptionDescribe(env); \
```
prrace, line 43: Describe already clears the exception
https://docs.oracle.com/en/java/javase/17/docs/specs/jni/functions.html#exceptiondescribe

src/java.desktop/unix/native/libawt_xawt/awt/screencast_pipewire.c
```diff
@@ -0,0 +1,915 @@
...
+}
+
+/**
+ * @param index of a screen
+ * @return TRUE on success
+ */
+static gboolean connectStream(int index) {
+    DEBUG_SCREENCAST("@@@ using screen %i\n", index);
+    if (index >= screenSpace.screenCount) {
+        DEBUG_SCREENCAST("⚠ Wrong index for screen\n", NULL);
```
prrace, line 375: I've seen this "! in a triangle" symbol in a quite a few places in the DEBUG messages.
Looks like you got some non-ascii char in there.

src/java.desktop/unix/native/libawt_xawt/awt/screencast_pipewire.c
azvegint: Updated.

src/java.desktop/unix/classes/sun/awt/screencast/TokenStorage.java
azvegint: Moved to newBufferedWriter/Readed

src/java.desktop/unix/native/libawt_xawt/awt/screencast_pipewire.c
azvegint: Fixed.

prrace (approved):
I'm getting pretty consistent testing results now - meaning the ones that have known
issues are the ones that fail, and I've looked at the code sufficiently.
The restore token seems to be behaving properly now.

I did have one case I've not understood where a test system got very confused about the display resolution of the monitor and needed a full power cycle.
I don't know if a specific test caused that or it was something else, but it isn't repeatable (at least not yet) so shouldn't block integration.

I am quite sure that SOME of the test failures are unstable tests that we've been lucky with all these years  and xwayland is now showing up the instability. Things un-related to robot and more about timing for the most part. Although there's one test that seriously over-stresses robot with > 10,000 calls in a loop to getPixelColor() I'll see if I can push test update fixes to at least some of these to reduce noise.

But I'm OK to approve now.

make/modules/java.desktop/lib/Awt2dLibraries.gmk
```diff
@@ -191,6 +191,9 @@ ifeq ($(call isTargetOs, windows macosx), false)
 
     LIBAWT_XAWT_EXCLUDES := medialib debug
 
+    LIBPIPEWIRE_HEADER_DIRS := \
+    	$(TOPDIR)/src/$(MODULE)/unix/native/libpipewire/include
```
turbanoff, line 195: Are you sure Tab should be used here?
in all other places of this file spaces are used for indentation.

src/java.desktop/unix/classes/sun/awt/screencast/TokenStorage.java
```diff
@@ -0,0 +1,410 @@
...
+        return true;
+    }
+
+    static Set<TokenItem> getTokens(List<Rectangle> affectedScreenBounds) {
+        // We need an ordered set to store tokens
+        // with exact matches at the beginning.
+        LinkedHashSet<TokenItem> result = new LinkedHashSet<>();
+
+        Set<Map.Entry<Object, Object>> entries;
+        synchronized (PROPS) {
```
turbanoff, line 303: I'm not sure I understand purpose of this `synchronized` block.
`entrySet()` returns _view_ of Properties entries, not a copy. It means concurrent thread could modify content of `PROPS` when we do `stream()` below.

From my opinion either we need to perform all work inside `synchronized` to be sure that no concurrent modification are possible. Or just remove `synchronized` (`Properties` is a thread safe class) and _be ready_ of possible concurrent updates

src/java.desktop/unix/classes/sun/awt/screencast/TokenItem.java
```diff
@@ -0,0 +1,143 @@
...
+        StringBuilder sb = new StringBuilder();
+        for (Rectangle bounds : allowedScreensBounds) {
+            sb.append("_%d_%d_%d_%d"
+                    .formatted(bounds.x, bounds.y, bounds.width, bounds.height));
+        }
+        return sb.toString();
+    }
+
+    public static TokenItem parse(String token, Object input) {
+        if (token == null || input == null) return null;
```
turbanoff, line 111: Do we need this `null` checks?
Currently only non-null values are passed to this method

src/java.desktop/unix/classes/sun/awt/screencast/TokenStorage.java
```diff
@@ -0,0 +1,410 @@
...
+                .map(o -> {
+                    String token = String.valueOf(o.getKey());
+                    TokenItem tokenItem =
+                            TokenItem.parse(token, o.getValue());
+                    if (tokenItem == null) {
+                        malformed.add(token);
+                    }
+                    return tokenItem;
+                })
+                .filter(obj -> !Objects.isNull(obj))
```
turbanoff, line 320: `Objects.isNull` wasn't supposed to be used like this. Let's either inline it and use `!= null` or use method reference `.filter(Objects::nonNull)`

src/java.desktop/unix/classes/sun/awt/screencast/TokenStorage.java
```diff
@@ -0,0 +1,410 @@
...
+        // We need an ordered set to store tokens
+        // with exact matches at the beginning.
+        LinkedHashSet<TokenItem> result = new LinkedHashSet<>();
+
+        Set<Map.Entry<Object, Object>> entries;
+        synchronized (PROPS) {
+            entries = PROPS.entrySet();
+        }
+
+        Set<String> malformed = new LinkedHashSet<>();
```
turbanoff, line 307: Can we use plain HashSet here? Seems there is not much reason to save original order of tokens.

src/java.desktop/unix/classes/sun/awt/screencast/TokenStorage.java
```diff
@@ -0,0 +1,410 @@
...
+
+        // 2. Try screens of the same size but in different locations,
+        // screens may have been moved while the token is still valid
+        List<Dimension> dimensions = affectedScreenBounds
+                .stream()
+                .map(rectangle -> new Dimension(
+                        rectangle.width,
+                        rectangle.height
+                ))
+                .collect(Collectors.toCollection(ArrayList::new));
```
turbanoff, line 346: why not plain toList() collector?

prrace:
I spoke too soon :-(
I remembered about an odd result from last week and it seems that whenever I have a full screen window capture fails - at least on my system and I think one lab system.
See test and output below

import java.awt.Frame;
import java.awt.GraphicsDevice;
import java.awt.Rectangle;
import java.awt.Robot;

public class FSW {

    public static void main(String args[]) throws Exception {
        Robot robot = new Robot();
        Frame frame = new Frame("FSW");
        GraphicsDevice gd = frame.getGraphicsConfiguration().getDevice();
        try {
            gd.setFullScreenWindow(frame);
            robot.delay(5000);
            robot.createScreenCapture(new Rectangle(50, 50, 10, 10));
        } finally {
            gd.setFullScreenWindow(null);
            frame.dispose();
        }
    }
}

% java FSW
cropTo:163 Unexpected stride / 4: 0 srcW: 1920

and the capture fails every time.


And again with debugging turned on 

ScreencastWatcher: started
// getRGBPixels affectedScreenBounds [java.awt.Rectangle[x=0,y=0,width=1920,height=1200]]
// getTokens exact matches 1. [Token: 41dc61fb-279b-4069-b408-30873451fb03
	java.awt.Rectangle[x=0,y=0,width=1920,height=1200]
]
// getTokens same sizes 2. [Token: 41dc61fb-279b-4069-b408-30873451fb03
	java.awt.Rectangle[x=0,y=0,width=1920,height=1200]
]
// storeToken old: |41dc61fb-279b-4069-b408-30873451fb03| new |41dc61fb-279b-4069-b408-30873451fb03| allowed bounds [0, 0, 1920, 1200]
// Storing TokenItem:
Token: 41dc61fb-279b-4069-b408-30873451fb03
	java.awt.Rectangle[x=0,y=0,width=1920,height=1200]

initXdgDesktopPortal:236 connection/sender name :1.525 / 1_525
checkVersion:190 ScreenCast protocol version 4
Java_sun_awt_screencast_ScreencastHelper_getRGBPixelsImpl:833 taking screenshot at 
	x:    50 y    50 w    10 h    10 with token |41dc61fb-279b-4069-b408-30873451fb03|
initXdgDesktopPortal:236 connection/sender name :1.525 / 1_525
checkVersion:190 ScreenCast protocol version 4
callbackScreenCastStart:611 available screen count 1
rebuildScreenData:87 
==== screenId#35
rebuildScreenData:132 -----------------------
rebuildScreenData:133 screenId#35
||	bounds         x     0 y     0 w  1920 h  1200
||	capture area   x     0 y     0 w     0 h     0 shouldCapture 0

rebuildScreenData:134 #---------------------#

callbackScreenCastStart:617 rebuildScreenData result |0|
callbackScreenCastStart:630 restore_token |41dc61fb-279b-4069-b408-30873451fb03|
storeRestoreToken:653 saving token, old: |41dc61fb-279b-4069-b408-30873451fb03| > new: |41dc61fb-279b-4069-b408-30873451fb03|
portalScreenCastStart:709 ScreenCastStartResult |0|
getPipewireFd:886 portalScreenCastStart result |0|
checkCanCaptureAllRequiredScreens:847 Found allowed screen bounds in affected screen bounds 0 0 1920 1200
getPipewireFd:900 --- portalScreenCastStart
getPipewireFd:907 pwFd 17
doLoop:557 screenId#35[loc(0,0) size(1920x1200)] @@@ adding screen 0
checkScreen:472 screenId#35
||	bounds         x     0 y     0 w  1920 h  1200
||	capture area   x    50 y    50 w    10 h    10 shouldCapture 1

connectStream:373 @@@ using screen 0
connectStream:411 screenId#35
||	bounds         x     0 y     0 w  1920 h  1200
||	capture area   x    50 y    50 w    10 h    10 shouldCapture 1

startStream:355 screenId#35: stream connecting 0x7f04548e3600
onStreamStateChanged:303 screenId#35[loc(0,0) size(1920x1200)] state 0 (unconnected) -> 1 (connecting) err (null)
onStreamStateChanged:303 screenId#35[loc(0,0) size(1920x1200)] state 1 (connecting) -> 2 (paused) err (null)
onStreamParamChanged:197 screenId#35[loc(0,0) size(1920x1200)] param event id 4
onStreamParamChanged:218 screenId#35[loc(0,0) size(1920x1200)] stream format: Spa:Enum:VideoFormat:BGRx (8)	1920x1200
connectStream:424 screenId#35[loc(0,0) size(1920x1200)] frame size: 1920x1200
doLoop:563 screenId#35[loc(0,0) size(1920x1200)] @@@ screen processed 0
onStreamStateChanged:303 screenId#35[loc(0,0) size(1920x1200)] state 2 (paused) -> 3 (streaming) err (null)
onStreamProcess:236 screenId#35[loc(0,0) size(1920x1200)] hasFormat 1 captureDataReady 0 shouldCapture 1
onStreamProcess:271 screenId#35
||	bounds         x     0 y     0 w  1920 h  1200
||	capture area   x    50 y    50 w    10 h    10 shouldCapture 1

onStreamProcess:272 screenId#35[loc(0,0) size(1920x1200)] got a frame of size 0 offset 0 stride 0 flags 0 FD 31 captureDataReady 0
onStreamProcess:292 screenId#35[loc(0,0) size(1920x1200)] data ready
Java_sun_awt_screencast_ScreencastHelper_getRGBPixelsImpl:856 
all data ready
Java_sun_awt_screencast_ScreencastHelper_getRGBPixelsImpl:864 screenId#35[loc(0,0) size(1920x1200)] @@@ copying screen data 0, captureData 0x7f03d40042a0
	||	x    50 y    50 w    10 h    10 requested area
	||	x     0 y     0 w  1920 h  1200 screen bound
	||	x    50 y    50 w    10 h    10 in-screen coords capture area

onStreamParamChanged:197 screenId#35[loc(0,0) size(1920x1200)] param event id 4
onStreamStateChanged:303 screenId#35[loc(0,0) size(1920x1200)] state 3 (streaming) -> 2 (paused) err (null)
onStreamStateChanged:303 screenId#35[loc(0,0) size(1920x1200)] state 2 (paused) -> 0 (unconnected) err (null)
doCleanup:109 STOPPING loop

src/java.desktop/unix/classes/sun/awt/screencast/TokenStorage.java
```diff
@@ -0,0 +1,410 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static java.nio.file.StandardWatchEventKinds.ENTRY_CREATE;
+import static java.nio.file.StandardWatchEventKinds.ENTRY_DELETE;
+import static java.nio.file.StandardWatchEventKinds.ENTRY_MODIFY;
+import static java.nio.file.StandardWatchEventKinds.OVERFLOW;
+import static sun.awt.screencast.ScreencastHelper.SCREENCAST_DEBUG;
+
+final class TokenStorage {
```
turbanoff, line 56: Would be nice to have a few words about purpose of this class in the javadoc

src/java.desktop/unix/classes/sun/awt/screencast/TokenItem.java
```diff
@@ -0,0 +1,143 @@
...
+import java.awt.Rectangle;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+import static sun.awt.screencast.ScreencastHelper.SCREENCAST_DEBUG;
+
+final class TokenItem {
```
turbanoff, line 38: add couple of words to the javadoc

src/java.desktop/unix/classes/sun/awt/screencast/TokenItem.java
```diff
@@ -0,0 +1,143 @@
...
+    String token;
+    List<Rectangle> allowedScreensBounds;
```
turbanoff, line 40: make fields final
```suggestion
    final String token;
    final List<Rectangle> allowedScreensBounds;
```

make/modules/java.desktop/lib/Awt2dLibraries.gmk
azvegint: Fixed

src/java.desktop/unix/classes/sun/awt/screencast/TokenItem.java
azvegint: Sure, updated

src/java.desktop/unix/classes/sun/awt/screencast/TokenItem.java
azvegint: It won't hurt.

prrace:
Regarding the failed capture
> cropTo:163 Unexpected stride / 4: 0 srcW: 1920

I've confirmed that on my system we DO enter FSEM using Xrandr
If I just comment out the call to XSendEvent(..) in X11GD_SetFullscreenMode(..)
then the capture works fine.

So something about this code, or pipewire or XWayland isn't handling this properly.

azuev-java (approved):
Looks good, seems to be working reasonably well on Ubuntu 20 and 22.

src/java.desktop/unix/classes/sun/awt/screencast/TokenStorage.java
```diff
@@ -0,0 +1,416 @@
...
+
+        return path;
+    }
+
+    private static boolean setFilePermission(Path path) {
+        try {
+            Files.setPosixFilePermissions(
+                    path,
+                    Set.of(PosixFilePermission.OWNER_READ,
+                            PosixFilePermission.OWNER_WRITE)
```
prsadhuk, line 145: alignment issue..Posix..OWNER_WRITE should be placed below Posix..OENER_READ

Also, why we aren't adding OWNER_EXECUTE here too like done above?

src/java.desktop/unix/classes/sun/awt/screencast/TokenStorage.java
```diff
@@ -0,0 +1,416 @@
...
+        // with exact matches at the beginning.
+        LinkedHashSet<TokenItem> result = new LinkedHashSet<>();
+
+        Set<String> malformed = new HashSet<>();
+        List<TokenItem> allTokenItems;
+
+        synchronized (PROPS) {
+            allTokenItems = PROPS.entrySet()
+                    .stream()
+                    .map(o -> {
```
prsadhuk, line 316: align below PROPS

src/java.desktop/unix/classes/sun/awt/screencast/TokenStorage.java
```diff
@@ -0,0 +1,416 @@
...
+                    .toList();
+        }
+
+        removeMalformedRecords(malformed);
+
+        // 1. Try to find exact matches
+        for (TokenItem tokenItem : allTokenItems) {
+            if (tokenItem != null
+                    && tokenItem
+                    .hasAllScreensWithExactMatch(affectedScreenBounds)) {
```
prsadhuk, line 335: align below `tokenItem`

src/java.desktop/unix/classes/sun/awt/screencast/TokenStorage.java
```diff
@@ -0,0 +1,416 @@
...
+            System.out.println("// getTokens exact matches 1. " + result);
+        }
+
+        // 2. Try screens of the same size but in different locations,
+        // screens may have been moved while the token is still valid
+        List<Dimension> dimensions = affectedScreenBounds
+                .stream()
+                .map(rectangle -> new Dimension(
+                        rectangle.width,
+                        rectangle.height
```
prsadhuk, line 350: align below `affectedScreenBounds`

src/java.desktop/unix/classes/sun/awt/screencast/TokenStorage.java
```diff
@@ -0,0 +1,416 @@
...
+                ))
+                .toList();
+
+        for (TokenItem tokenItem : allTokenItems) {
+            if (tokenItem != null
+                    && tokenItem.hasAllScreensOfSameSize(dimensions)) {
```
prsadhuk, line 356: align below `tokenItem`

src/java.desktop/unix/classes/sun/awt/screencast/TokenStorage.java
```diff
@@ -0,0 +1,416 @@
...
+            System.out.println("// getTokens same sizes 2. " + result);
+        }
+
+        return result;
+    }
+
+    private static void removeMalformedRecords(Set<String> malformedRecords) {
+        if (!isWritable()
+                || malformedRecords == null
+                || malformedRecords.isEmpty()) {
```
prsadhuk, line 371: align below `isWritable`

src/java.desktop/unix/classes/sun/awt/screencast/TokenStorage.java
```diff
@@ -0,0 +1,416 @@
...
+                    System.err.printf(
+                            "Token storage: unable to %s\n%s\n", failMsg, e);
+                }
+            }
+        }
+    }
+
+    private static boolean isWritable() {
+        if (PROPS_PATH == null
+                || (Files.exists(PROPS_PATH) && !Files.isWritable(PROPS_PATH))) {
```
prsadhuk, line 406: align below PROPS_PATH

src/java.desktop/unix/native/libawt_xawt/awt/gtk2_interface.h
```diff
@@ -86,13 +86,6 @@ typedef struct {
   gushort   revents;
 } GPollFD;
 
-typedef struct {
-  gint x;
-  gint y;
-  gint width;
-  gint height;
-} GdkRectangle;
```
prsadhuk, line 94: Any reason why it is removed? I think it's needed in gtk2_interface.c, right? 
Probably we can delete from gtk3_interface.h as it is not used in gtk3_interface.c

src/java.desktop/unix/native/libawt_xawt/awt/screencast_pipewire.c
```diff
@@ -0,0 +1,910 @@
...
+            (*env)->DeleteLocalRef(env, jOldToken);
+            return;
+        }
+
+        jintArray allowedBounds = NULL;
+        if (screenSpace.screenCount > 0) {
+            allowedBounds = (*env)->NewIntArray(env, screenSpace.screenCount*4);
+            EXCEPTION_CHECK_DESCRIBE();
+            if (!allowedBounds) {
+                (*env)->ExceptionClear(env);
```
prsadhuk, line 674: I guess as mentioned above, DESCRIBE clears exception also so no need of calling  explicit Clear
https://docs.oracle.com/en/java/javase/17/docs/specs/jni/functions.html#exceptiondescribe

src/java.desktop/unix/native/libawt_xawt/awt/screencast_pipewire.c
```diff
@@ -0,0 +1,910 @@
...
+            for (int i = 0; i < screenSpace.screenCount; ++i) {
+                GdkRectangle bounds = screenSpace.screens[i].bounds;
+                elements[4 * i] = bounds.x;
+                elements[4 * i + 1] = bounds.y;
+                elements[4 * i + 2] = bounds.width;
+                elements[4 * i + 3] = bounds.height;
+
+                if ((*env)->ExceptionCheck(env)) {
+                    (*env)->ExceptionDescribe(env);
+                    (*env)->ExceptionClear(env);
```
prsadhuk, line 694: same here

prsadhuk (approved):
Mostly alignment issue with couple of code comments..

src/java.desktop/unix/classes/sun/awt/screencast/TokenStorage.java
prrace: Because execute is required on directories but not files unless they are programs.

src/java.desktop/unix/native/libawt_xawt/awt/gtk2_interface.h
prsadhuk: ok, I see it's used in screencast c file so moved to gtk_interface.h

src/java.desktop/unix/classes/sun/awt/screencast/TokenStorage.java
azvegint: Exactly, without an execute bit on the directory we cannot access/add/remove its files.

Our file is not a script or executable that needs to be run.

src/java.desktop/unix/native/libawt_xawt/awt/screencast_pipewire.c
azvegint: Actually, these checks are not needed here at all, this is an oversight, there was a Java function call before.
The check after ReleaseIntArrayElements is also unnecessary.

PR #13805, opened 2023-05-04
8307423: [s390x] Represent Registers as values

offamitkumar:
The recent Register implementation uses wild pointer (including null pointer) dereferences which exhibit undefined behavior. We should migrate away from pointer-based representation of Register values as it was done for x86 ([JDK-8292153](https://bugs.openjdk.org/browse/JDK-8292153)). Problems exist when trying to build with GCC 11.3.0 ([JDK-8307093](https://bugs.openjdk.org/browse/JDK-8307093)).

Tested `fastdebug, slowdebug, optimized, release build` , `tier1-test` on fastdebug build and build with GCC-9.5.0 as well.

---------
### Progress

### Issue
 * [JDK-8307423](https://bugs.openjdk.org/browse/JDK-8307423): [s390x] Represent Registers as values

### Reviewers
 * [Martin Doerr](https://openjdk.org/census#mdoerr) (@TheRealMDoerr - **Reviewer**) ⚠️ Review applies to [6fe675e3](https://git.openjdk.org/jdk/pull/13805/files/6fe675e321c05bdd749abc6983d5815cb9d08d3b)
 * [Lutz Schmidt](https://openjdk.org/census#lucy) (@RealLucy - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13805/head:pull/13805` \
`$ git checkout pull/13805`

`$ git checkout pull/13805` \
`$ git pull https://git.openjdk.org/jdk.git pull/13805/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13805`

`$ git pr show -t 13805`

</details>

<a href="https://git.openjdk.org/jdk/pull/13805.diff">https://git.openjdk.org/jdk/pull/13805.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13805#issuecomment-1534963515)

### DIFF
--- a/src/hotspot/cpu/s390/assembler_s390.hpp
+++ b/src/hotspot/cpu/s390/assembler_s390.hpp
@@ -189,11 +189,6 @@ class Address {
     _index(noreg),
     _disp(0) {}
 
-  Address(Register base, Register index, intptr_t disp = 0) :
-    _base(base),
-    _index(index),
-    _disp(disp) {}
-
   Address(Register base, intptr_t disp = 0) :
     _base(base),
     _index(noreg),
--- a/src/hotspot/cpu/s390/register_s390.hpp
+++ b/src/hotspot/cpu/s390/register_s390.hpp
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 2016, 2019, Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2016, 2019 SAP SE. All rights reserved.
+ * Copyright (c) 2016, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2023 SAP SE. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -29,9 +29,8 @@
 #include "asm/register.hpp"
 #include "runtime/vm_version.hpp"
 
-class Address;
+// forward declaration
 class VMRegImpl;
-
 typedef VMRegImpl* VMReg;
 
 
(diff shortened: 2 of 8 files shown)


offamitkumar:
Hi @RealLucy, @TheRealMDoerr 

Please review the changes.
Thank you

src/hotspot/cpu/s390/register_s390.hpp
```diff
@@ -485,98 +389,57 @@ class ConcreteRegisterImpl : public AbstractRegisterImpl {
...
-#define Z_fp              AS_REGISTER(Register, Z_R9)
-#define Z_locals          AS_REGISTER(Register, Z_R12)
-#define Z_bcp             AS_REGISTER(Register, Z_R13)
-#define Z_bytecode        AS_REGISTER(Register, Z_R14)
-#endif
+constexpr Register      Z_bytecode     = Z_R14;
 
 // Temporary registers to be used within frame manager. We can use
-// the nonvolatiles because the call stub has saved them.
+// the nonvolatile because the call stub has saved them.
```
TheRealMDoerr, line 433: Better: "nonvolatile ones"

src/hotspot/cpu/s390/assembler_s390.hpp
```diff
@@ -189,11 +189,6 @@ class Address {
     _index(noreg),
     _disp(0) {}
 
-  Address(Register base, Register index, intptr_t disp = 0) :
-    _base(base),
-    _index(index),
-    _disp(disp) {}
-
```
TheRealMDoerr, line 196: I can live with the removal, but I guess it may be useful at some point of time. s390 supports specifying both, index and disp (unlike PPC64).

src/hotspot/cpu/s390/register_s390.hpp
```diff
@@ -57,235 +56,172 @@ typedef VMRegImpl* VMReg;
...
-  }
+  constexpr int encoding() const { assert(is_valid(), "invalid register"); return _encoding; }
+  inline VMReg as_VMReg() const;
+  FloatRegister successor() const { return FloatRegister(encoding() + 1); }
 
-  bool  is_valid() const          { return 0 <= value() && value() < number_of_registers; }
-  bool is_volatile() const        { return (0 <= (value()&0x7F) && (value()&0x7F) <= 7); }
-  bool is_nonvolatile() const     { return (8 <= (value()&0x7F) && (value()&0x7F) <= 15); }
+  // tester
+  constexpr bool  is_valid()      const { return 0 <= _encoding && _encoding < number_of_registers; }
```
TheRealMDoerr, line 186: Indentation is different than for the other register types. I suggest to adapt it, here.

TheRealMDoerr (approved):
LGTM. Please consider my minor suggestions.

src/hotspot/cpu/s390/assembler_s390.hpp
offamitkumar: This constructor was causing ambiguity with this one :
```
  Address(Register base, RegisterOrConstant roc, intptr_t disp = 0) :
    _base(base),
    _index(noreg),
    _disp(disp) {
    if (roc.is_constant()) _disp += roc.as_constant(); else _index = roc.as_register();
  }
```

that's why I removed it. But yeah I can try to find another workaround, any suggestion from your side ?

src/hotspot/cpu/s390/register_s390.hpp
offamitkumar: Done; I did some other formatting changes, Please let me know if they looks good, I can happily revert them otherwise.

src/hotspot/cpu/s390/vmreg_s390.inline.hpp
```diff
@@ -26,18 +26,15 @@
 #ifndef CPU_S390_VMREG_S390_INLINE_HPP
 #define CPU_S390_VMREG_S390_INLINE_HPP
 
-inline VMReg RegisterImpl::as_VMReg() {
-  if (this == noreg) {
-    return VMRegImpl::Bad();
-  }
```
offamitkumar, line 32: Although @TheRealMDoerr  you've reviewed the changes. Do you think this getting rid from this check is okay? No other arch have this & this was an Typecasting error for us, after these changes.

src/hotspot/cpu/s390/register_s390.hpp
```diff
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 2016, 2019, Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2016, 2019 SAP SE. All rights reserved.
+ * Copyright (c) 2016, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2023 SAP SE. All rights reserved.
```
offamitkumar, line 3: Hi All, is it okay If I add IBM header as well ?

offamitkumar: Thanks Lutz, Let's skip this part then :-)

src/hotspot/cpu/s390/assembler_s390.hpp
TheRealMDoerr: Ah, got it. I'm fine with the removal. The other constructor can be used as replacement.

src/hotspot/cpu/s390/register_s390.hpp
TheRealMDoerr: I agree. You can add IBM Copyright headers when you contribute new files or when they contain substantial contributions.

offamitkumar: My thoughts were that we could update header once we do substantials changes in the file, I haven't considered the port as whole. But I'll take care of it next time.

src/hotspot/cpu/s390/vmreg_s390.inline.hpp
TheRealMDoerr: Fine with me. PPC64 still has it, but if it's not used, I think it's ok to remove it. We have the `assert(is_valid(), "invalid register")` in `encoding()` to catch it.
If you prefer to keep it, you need to use `*this`.

RealLucy (changes requested):
Your changes look good to me. 

There is one fact I do not like. It was not introduced by this PR, but could easily be fixed now.

The value to indicate "this is not a register" is "-1", used as a literal in multiple places. I would rather see
`#define NOREG_ENCODING -1`
and then 
`constexpr Register(int encoding = NOREG_ENCODING) : _encoding(encoding) {}`

Searching for "-1" in register_s390.hpp will reveal all affected locations.

offamitkumar:
@TheRealMDoerr, do you think applying `#define NOREG_ENCODING -1`  change to PPC as well, will be a good idea ?

RealLucy:
> @TheRealMDoerr, do you think applying `#define NOREG_ENCODING -1` change to PPC as well, will be a good idea ?

If so, it should be done in a separate PR.

RealLucy (changes requested):
Even more requests - sorry.

RealLucy (approved):
LGTM
Thank you for fixing, and for taking my requests into consideration.

RealLucy (approved):
LGTM
Thank you for fixing, and for taking my requests into consideration.

offamitkumar:
Thanks @RealLucy @TheRealMDoerr 

/integrate

PR #13806, opened 2023-05-04
8306326: [BACKOUT] 8277573: VmObjectAlloc is not generated by intrinsics methods which allocate objects

lmesnik:
8277573: VmObjectAlloc is not generated by intrinsics methods which allocate objects

caused significant regressions in some benchmarks and should be reverted.

This fix backout changes and update problemlist bugs to new issue.
Tier1 passed
Running also tier5 to check other builds and more svc testing

### DIFF
--- a/src/hotspot/share/opto/runtime.hpp
+++ b/src/hotspot/share/opto/runtime.hpp
@@ -136,7 +136,6 @@ class OptoRuntime : public AllStatic {
   static address _slow_arraycopy_Java;
   static address _register_finalizer_Java;
 #if INCLUDE_JVMTI
-  static address _notify_jvmti_object_alloc;
   static address _notify_jvmti_vthread_start;
   static address _notify_jvmti_vthread_end;
   static address _notify_jvmti_vthread_mount;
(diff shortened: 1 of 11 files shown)


src/hotspot/share/opto/runtime.hpp
```diff
@@ -216,7 +215,6 @@ class OptoRuntime : public AllStatic {
   static address slow_arraycopy_Java()                   { return _slow_arraycopy_Java; }
   static address register_finalizer_Java()               { return _register_finalizer_Java; }
 #if INCLUDE_JVMTI
-  static address notify_jvmti_object_alloc()             { return _notify_jvmti_object_alloc; }
```
sspitsyn, line 219: This line has to be also removed:
`312   static const TypeFunc* notify_jvmti_object_alloc_Type();`

sspitsyn (commented):
The `notify_jvmti_object_alloc_Type` declaration needs to be also removed from the runtime.hpp file. Other than that the BACKOUT looks clean.
Thanks,.
Serguei

sspitsyn (approved):

TobiHartmann (approved):
Looks good.

vnkozlov (approved):
Agree.

PR #13809, opened 2023-05-04
8307779: Relax the java.awt.Robot specification

azvegint:
We need to relax the java.awt.Robot specification according to the latest operating system trends. 
This should at least cover the case of Wayland, which has changed many familiar concepts in Linux.

https://bugs.openjdk.org/browse/JDK-8280982 [Wayland] [XWayland] java.awt.Robot taking screenshots
https://bugs.openjdk.org/browse/JDK-8280995 [XWayland] Robot.mouseMove does not visually move mouse cursor
https://bugs.openjdk.org/browse/JDK-8280990 [XWayland] XTest emulated mouse click does not bring window to front.
https://bugs.openjdk.org/browse/JDK-8280988 [XWayland] Click on title to request focus test failures

---------
### Progress
- [x] Change requires CSR request [JDK-8308012](https://bugs.openjdk.org/browse/JDK-8308012) to be approved

### Issues
 * [JDK-8307779](https://bugs.openjdk.org/browse/JDK-8307779): Relax the java.awt.Robot specification
 * [JDK-8308012](https://bugs.openjdk.org/browse/JDK-8308012): Relax the java.awt.Robot specification (**CSR**)

### Reviewers
 * [Maxim Kartashev](https://openjdk.org/census#mkartashev) (@mkartashev - no project role) ⚠️ Review applies to [53145ec4](https://git.openjdk.org/jdk/pull/13809/files/53145ec40a30afa9b7f2570db2efa95ca22f2fa7)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13809/head:pull/13809` \
`$ git checkout pull/13809`

`$ git checkout pull/13809` \
`$ git pull https://git.openjdk.org/jdk.git pull/13809/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13809`

`$ git pr show -t 13809`

</details>

<a href="https://git.openjdk.org/jdk/pull/13809.diff">https://git.openjdk.org/jdk/pull/13809.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13809#issuecomment-1540807717)

### DIFF
--- a/src/java.desktop/share/classes/java/awt/Robot.java
+++ b/src/java.desktop/share/classes/java/awt/Robot.java
@@ -400,7 +400,7 @@ public class Robot {
      * from a user to capture the screens via system dialog.
      * When selected in this dialog, the permission to capture a set of screens
      * can be saved for later reuse.<br>
-     * Use the {@link #revokeScreenCapturePermission()} to reset the stored
+     * Use the {@link #revokeScreenCapturePermission()} to revoke the stored
      * permission.
      * <br>
      * The pixel color may be black for screens for which no capture permission
@@ -431,7 +431,7 @@ public class Robot {
      * from a user to capture the screens via system dialog.
      * When selected in this dialog, the permission to capture a set of screens
      * can be saved for later reuse.<br>
-     * Use the {@link #revokeScreenCapturePermission()} to reset the stored
+     * Use the {@link #revokeScreenCapturePermission()} to revoke the stored
      * permission.
      * <br>
      * The image may be black for screens for which no capture permission
--- a/src/java.desktop/share/classes/java/awt/Robot.java
+++ b/src/java.desktop/share/classes/java/awt/Robot.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1999, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1999, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -68,6 +68,14 @@ import static sun.java2d.SunGraphicsEnvironment.toDeviceSpaceAbs;
  * <p>
  * Applications that use Robot for purposes other than self-testing should
  * handle these error conditions gracefully.
+ *
+ * @implNote on Linux systems using Wayland, the robot runs in
+ * so-called X11 compatibility mode. This means that generated mouse and
+ * keyboard events can only be delivered within the X11 server and between
+ * their clients.<br>
+ * Trying to interact with the Wayland area of responsibility may have no effect
+ * (e.g. interacting with window decorations, trying to change the
+ * window stacking order with the mouse or keyboard shortcuts)
  *
  * @author      Robi Khan
  * @since       1.3
@@ -189,6 +197,10 @@ public class Robot {
 
     /**
      * Moves mouse pointer to given screen coordinates.
+     * @implNote the mouse pointer may not visually move on Linux systems
+     *           using Wayland, while the subsequent mousePress and mouseRelease
+     *           can be delivered to the correct location
+     *
      * @param x         X position
      * @param y         Y position
      */
@@ -383,8 +395,24 @@ public class Robot {
 
     /**
      * Returns the color of a pixel at the given screen coordinates.
+     *
+     * @implNote On Linux systems using Wayland, permission may be requested
+     * from a user to capture the screens via system dialog.
+     * When selected in this dialog, the permission to capture a set of screens
+     * can be saved for later reuse.<br>
+     * Use the {@link #resetScreenCapturePermission()} to reset the stored
+     * permission.
+     * <br>
+     * The pixel color may be black if permission to capture a particular screen
+     * has not been granted.
+     * <br>
+     * It is not recommended to call this method on EDT, as it may block the UI
+     * update until the user confirms his choice in the system dialog.
+     *
      * @param   x       X position of pixel
      * @param   y       Y position of pixel
+     * @throws  SecurityException if {@code readDisplayPixels} permission
+     *          is not granted, or user has denied screen capturing
      * @return  Color of the pixel
      */
     public synchronized Color getPixelColor(int x, int y) {
@@ -397,10 +425,26 @@ public class Robot {
     /**
      * Creates an image containing pixels read from the screen.  This image does
      * not include the mouse cursor.
+     *
+     * @implNote On Linux systems using Wayland, permission may be requested
+     * from a user to capture the screens via system dialog.
+     * When selected in this dialog, the permission to capture a set of screens
+     * can be saved for later reuse.<br>
+     * Use the {@link #resetScreenCapturePermission()} to reset the stored
+     * permission.
+     * <br>
+     * The image may be black if permission to capture a particular screen
+     * has not been granted.
+     * <br>
+     * It is not recommended to call this method on EDT, as it may block the UI
+     * update until the user confirms his choice in the system dialog.
+     *
      * @param   screenRect      Rect to capture in screen coordinates
      * @return  The captured image
-     * @throws  IllegalArgumentException if {@code screenRect} width and height are not greater than zero
-     * @throws  SecurityException if {@code readDisplayPixels} permission is not granted
+     * @throws  IllegalArgumentException if {@code screenRect} width and height
+     *          are not greater than zero
+     * @throws  SecurityException if {@code readDisplayPixels} permission
+     *          is not granted, or user has denied screen capturing
      * @see     SecurityManager#checkPermission
      * @see     AWTPermission
      */
@@ -583,6 +627,14 @@ public class Robot {
         if (security != null) {
             security.checkPermission(AWTPermissions.READ_DISPLAY_PIXELS_PERMISSION);
         }
+    }
+
+    /**
+     * Resets the stored screen data capture permission for a set of screens.
+     * Is a no-op if not supported by the platform.
+     */
+    public void resetScreenCapturePermission() {
+        peer.resetScreenCapturePermission();
     }
 
     /*
--- a/src/java.desktop/share/classes/java/awt/Robot.java
+++ b/src/java.desktop/share/classes/java/awt/Robot.java
@@ -434,8 +434,8 @@ public class Robot {
      * Use the {@link #revokeScreenCapturePermission()} to reset the stored
      * permission.
      * <br>
-     * The image may be black if permission to capture a particular screen
-     * has not been granted.
+     * The image may be black for screens for which no capture permission
+     * has been granted.
      * <br>
      * It is not recommended to call this method on EDT, as it may block the UI
      * update until the user confirms his choice in the system dialog.
@@ -445,7 +445,8 @@ public class Robot {
      * @throws  IllegalArgumentException if {@code screenRect} width and height
      *          are not greater than zero
      * @throws  SecurityException if {@code readDisplayPixels} permission
-     *          is not granted, or user has denied screen capturing
+     *          is not granted, or user has denied screen capture for all
+     *          screens
      * @see     SecurityManager#checkPermission
      * @see     AWTPermission
      */
--- a/src/java.desktop/share/classes/java/awt/Robot.java
+++ b/src/java.desktop/share/classes/java/awt/Robot.java
@@ -400,11 +400,11 @@ public class Robot {
      * from a user to capture the screens via system dialog.
      * When selected in this dialog, the permission to capture a set of screens
      * can be saved for later reuse.<br>
-     * Use the {@link #resetScreenCapturePermission()} to reset the stored
+     * Use the {@link #revokeScreenCapturePermission()} to reset the stored
      * permission.
      * <br>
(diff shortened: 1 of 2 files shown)


src/java.desktop/share/classes/java/awt/Robot.java
```diff
@@ -585,6 +629,14 @@ private static void checkScreenCaptureAllowed() {
         }
     }
 
+    /**
+     * Resets the stored screen data capture permission for a set of screens.
```
mkartashev, line 633: Does "reset" here mean that the next attempt to capture screen pixels will require another permission? It's not clear what the effects of this "reset" are.

Also, the part "for a set of screens" kind of implies that for some other set of screens the permission will not be reset, but how can one specify for which set of screens the permission is to be reset? Perhaps, simply
"Revokes the stored permission to capture screen data. Subsequent calls to getPixelColor and createScreenCapture may request a new permission from the user on applicable platforms"?

src/java.desktop/share/classes/java/awt/peer/RobotPeer.java
```diff
@@ -127,4 +127,10 @@
     default boolean useAbsoluteCoordinates() {
         return false;
     }
+
+    /**
+     * Resets the stored screen data capture permission for a set of screens.
+     * Is a no-op if not supported by the platform.
```
mkartashev, line 133: Same here as in `Robot.resetScreenCapturePermission()`.

src/java.desktop/share/classes/java/awt/Robot.java
azvegint: Yes, this sounds much better, updated.

src/java.desktop/share/classes/java/awt/Robot.java
```diff
@@ -383,8 +395,24 @@ private static void checkKeycodeArgument(int keycode) {
...
+     *
+     * @implNote On Linux systems using Wayland, permission may be requested
+     * from a user to capture the screens via system dialog.
+     * When selected in this dialog, the permission to capture a set of screens
+     * can be saved for later reuse.<br>
+     * Use the {@link #resetScreenCapturePermission()} to reset the stored
+     * permission.
+     * <br>
+     * The pixel color may be black if permission to capture a particular screen
+     * has not been granted.
```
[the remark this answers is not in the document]
azvegint: Updated.

mkartashev (approved):
LGTM, although I am not an OpenJDK reviewer.

prrace (changes requested):

src/java.desktop/share/classes/java/awt/Robot.java
```diff
@@ -383,8 +395,25 @@ private static void checkKeycodeArgument(int keycode) {
...
+     * has been granted.
+     * <br>
+     * It is not recommended to call this method on EDT, as it may block the UI
+     * update until the user confirms his choice in the system dialog.
+     *
      * @param   x       X position of pixel
      * @param   y       Y position of pixel
+     * @throws  SecurityException if {@code readDisplayPixels} permission
+     *          is not granted, or user has denied screen capture for all
+     *          screens
```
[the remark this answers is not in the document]
azvegint: > Is there some subtlety about "ALL" screens ?

It is more like "the user has not allowed any of his screens to be captured."

In fact, we may not throw out this exception and just provide black pixels(and call it  "unspecified" behavior).

> Do you really mean all screens which are in the area being captured ?

> I added the word REQUIRED here, since if we only need screen 1 contents, surely it
doesn't matter if we don't have screen 2 permission ?

That's not exactly true.

There are screens that the user may or may not have given permission to capture. 
This permission(specifically `restore_token`) is one at a single moment in time, not several separate permissions for each screen. This restore_token is associated with the permissions for specific screens.

There is an area that the user wants to capture.

The intersection of the capture area and the screens allowed for capture will have actual data from the screen, otherwise - black pixels.

So if a user allows access only to screen A and then asks for pixels from screen B he will get black pixels.

This is about the same when we have a single monitor and we capture an area outside of it (e.g. -100, -100, 50, 50) - we get a black square.

> Does that also imply you can get a partial capture if it spans 2 screens and you only have permission for one of them ?

Yes, the plan is to provide what we can and leave the other parts black.

> We do also need to think forward to a pure wayland scenario - any differences there. ?

As a screen capture - should work as it is.

The only concern is the impossibility to specify and get the window coordinates.
This way we won't know where the window we are displaying is located.
It will affect the testing approach and we will have to add a method like `captureWindowContents` to check the window contents.

src/java.desktop/share/classes/java/awt/Robot.java
```diff
@@ -585,6 +631,16 @@ private static void checkScreenCaptureAllowed() {
         }
     }
 
+    /**
+     * Revokes the stored permission to capture screen data.
+     * Subsequent calls to {@link #getPixelColor(int, int)}
+     * and {@link #createScreenCapture(Rectangle)} may request
+     * a new permission from the user on applicable platforms.
+     */
+    public void revokeScreenCapturePermission() {
```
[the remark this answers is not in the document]
azvegint: Updated according to your comments.
I still left `revokeScreenCapturePermission` here for now, so that its documentation can be easily compared and aligned with the other documentation changes. 
After finalization it will be moved to #13803 (if we decide to keep the method)

src/java.desktop/share/classes/java/awt/Robot.java
```diff
@@ -397,10 +455,27 @@ public synchronized Color getPixelColor(int x, int y) {
     /**
      * Creates an image containing pixels read from the screen.  This image does
      * not include the mouse cursor.
```
prsadhuk, line 457: We have seen in macOS specifically that absence of mouse cursor is not guaranteed and screen capture can contain mouse cursor, so I guess we should remove this line..

src/java.desktop/share/classes/java/awt/Robot.java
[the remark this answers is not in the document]
kevinrushforth: I agree with Phil (and Sergey, who mentioned the same earlier). I'm not sure I see enough motivation for this method to justify adding it. It seems more like a workaround for an implementation quirk of Wayland to me.

kevinrushforth (commented):
I left a few comments inline.

src/java.desktop/share/classes/java/awt/Robot.java
```diff
@@ -189,6 +226,10 @@ private static void checkIsScreenDevice(GraphicsDevice device) {
 
     /**
      * Moves mouse pointer to given screen coordinates.
+     * @implNote the mouse pointer may not visually move on Linux systems
+     *           using Wayland, while the subsequent mousePress and mouseRelease
+     *           can be delivered to the correct location
```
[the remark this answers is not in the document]
azvegint: Updated.
Probably this can be covered by the line 80:
`<li> ignoring or limiting specific requests for Robot generated (synthesized) events related to keyboard and mouse etc.</li>`

src/java.desktop/share/classes/java/awt/Robot.java
azvegint: Sure, this is a good time to do it.

src/java.desktop/share/classes/java/awt/Robot.java
```diff
@@ -383,8 +424,25 @@ private static void checkKeycodeArgument(int keycode) {
 
     /**
      * Returns the color of a pixel at the given screen coordinates.
+     * <p>
+     * If the desktop environment requires that permissions be granted
+     * to capture screen content, and the required permissions are not granted,
+     * then a {@code SecurityException} may be thrown,
+     * or the content of the returned {@code Color} is undefined.
```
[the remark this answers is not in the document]
azvegint: I agree, I'll try to come up with something less confusing.

src/java.desktop/share/classes/java/awt/Robot.java
```diff
@@ -189,6 +226,11 @@ private static void checkIsScreenDevice(GraphicsDevice device) {
 
     /**
      * Moves mouse pointer to given screen coordinates.
+     * @implNote the mouse pointer may not visually move on some platforms,
```
prrace, line 229: This doesn't seem to be an implNote in the way that the JEP
intends since it has nothing to do with anything we did in our implementation. I think it would have to be just spec, in which case I am not sure we should call out Linux+Wayland

See  https://openjdk.org/jeps/8068562

"Implementation Notes. This section contains informative notes about the implementation, such as advice to implementors, or performance characteristics that are specific to the implementation in this class of this version of the JDK. The information in this section is subject to change from release to release. These characteristics are also allowed to vary across platforms, vendors, and versions."

src/java.desktop/share/classes/java/awt/Robot.java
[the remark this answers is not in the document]
prrace: If there's no exception, how does the application know the colours aren't valid ?
I think throwing the exception should be preferred behaviour and undefined colours would be just if we had no way to detect that there was a problem.

src/java.desktop/share/classes/java/awt/Robot.java
```diff
@@ -395,12 +454,28 @@ public synchronized Color getPixelColor(int x, int y) {
...
+     * Creates an image containing pixels read from the screen.
+     * <p>
+     * If the desktop environment requires that permissions be granted
+     * to capture screen content, and the required permissions are not granted,
+     * then a {@code SecurityException} may be thrown,
+     * or the contents of the returned {@code BufferedImage} are undefined.
+     * <p>
+     * The {@link #revokeScreenCapturePermission()} can be used to revoke
+     * a previously granted permission.
+     * <p>
```
prrace, line 466: I would like to completely separate any decision on adding revokeScreenCapturePermission from
this spec. relaxation.
So we should remove the reference to it here into a separate PR about adding that API although I'm still not convinced we should add it.

src/java.desktop/share/classes/java/awt/Robot.java
```diff
@@ -395,12 +454,28 @@ public synchronized Color getPixelColor(int x, int y) {
...
+     *
      * @param   screenRect      Rect to capture in screen coordinates
      * @return  The captured image
-     * @throws  IllegalArgumentException if {@code screenRect} width and height are not greater than zero
-     * @throws  SecurityException if {@code readDisplayPixels} permission is not granted
+     * @throws  IllegalArgumentException if {@code screenRect} width and height
+     *          are not greater than zero
+     * @throws  SecurityException if {@code readDisplayPixels} permission
+     *          is not granted, or user has not allowed any of his screens
+     *          to be captured.
```
prrace, line 478: "user" -> "the user"
"his" -> gender neutral "their".
Or maybe better ", or access to the screen is denied by the desktop environment"

src/java.desktop/share/classes/java/awt/Robot.java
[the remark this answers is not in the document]
prrace: Like I wrote elsewhere, I am not sure how the app would know when to call this.
And it is easy enough to use the preferences API to clear this out whilst we think more on this. BTW "reset" or "clear" might be better than "revoke" .. for all we know the current token denies access anyway and what we really mean is more like those other suggestions.

prrace (changes requested):

src/java.desktop/share/classes/java/awt/Robot.java
azvegint: Updated.

src/java.desktop/share/classes/java/awt/Robot.java
azvegint: Removed from this PR

src/java.desktop/share/classes/java/awt/Robot.java
azvegint: Updated to the former for now.

src/java.desktop/share/classes/java/awt/Robot.java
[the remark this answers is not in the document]
prrace: BTW in the implementation PR, Alexander wrote


The [DENY case](https://github.com/azvegint/jdk/blob/482471df61490120590bc49886f76dc99c3d7dd8/src/java.desktop/unix/native/libawt_xawt/awt/screencast_portal.c#L586-L591) propagated to java and we are throwing the security exception.

So I interpret this as there is a use case for the exception.

src/java.desktop/share/classes/java/awt/peer/RobotPeer.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1999, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1999, 2023, Oracle and/or its affiliates. All rights reserved.
```
prrace, line 2: This file no longer needed in this PR

src/java.desktop/share/classes/java/awt/Robot.java
prrace: I think you should draft the CSR now based on the current text.
We may still need to tweak words like
"or the user has not allowed any of their screens to be captured."
based on developing understanding, or perhaps being careful not to be describe in the docs what happens to be the pipewire behaviour.
In fact that is why I suggested my other words, because (1) there may be no user involved and
(2) screen vs screens seems to be a pipewire / wayland concept. I'm pretty sure macOS just treats it as "the desktop" no matter how many monitors are involved.

PR #13813, opened 2023-05-04
8263464: NMT: assert in gtest os.release_multi_mappings_vm

gerard-ziemski:
This fix allows NMT to account for released memory that was allocated in chunks, as long as the final pointer and size refer to total contiguous regions of the requested size.

The main idea here is to iterate over the "synthetic" region that was given to us and split it up into the underlaying existing regions, and then recursively handle them one at a time as we normally do.

Tested via `Mach5 tier1,tier2,tier3` and locally via `gtest:NMT*:os*`

---------
### Progress

### Issue
 * [JDK-8263464](https://bugs.openjdk.org/browse/JDK-8263464): NMT: assert in gtest os.release_multi_mappings_vm (**Bug** - P4) ⚠️ Issue is not open.

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13813/head:pull/13813` \
`$ git checkout pull/13813`

`$ git checkout pull/13813` \
`$ git pull https://git.openjdk.org/jdk.git pull/13813/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13813`

`$ git pr show -t 13813`

</details>

<a href="https://git.openjdk.org/jdk/pull/13813.diff">https://git.openjdk.org/jdk/pull/13813.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13813#issuecomment-1535318040)


gerard-ziemski:
@tstuefe You might be interested in reviewing this one.

src/hotspot/share/services/virtualMemoryTracker.cpp
```diff
@@ -523,28 +523,46 @@ bool VirtualMemoryTracker::remove_released_region(address addr, size_t size) {
...
+      assert(remove_rgn->base()+remove_rgn->size()<=end, "not contained");
+
+      remaining -= remove_rgn->size();
+      node_rgn = node_rgn->next();
+
+      assert(remove_released_region(remove_rgn), "error in remove_released_region");
+    }
+
+    return true;
   }
```
tstuefe, line 566: - you don't need to carry/calculate remaining. The end pointer, and comparing each region's end with it, should be sufficient.
- Some sort of assert that addresses are monotonously ascending would be assuring
- The last region may not be completely part of the released memory range. See above, lines 534..546. Needs to be treated the same way here. Possibly you can fold both approaches together roughly like this:
```
while (release range end is behind current region end) {
  release whole region
  region = next region
}
// last region may not be fully contained
if (last region not fully contained) {
  split region, release lower half
}
```

test/hotspot/gtest/runtime/test_os.cpp
```diff
@@ -458,11 +458,6 @@ struct NUMASwitcher {
 #ifndef _AIX // JDK-8257041
 TEST_VM(os, release_multi_mappings) {
 
-  // With NMT enabled, this will trigger JDK-8263464. For now disable the test if NMT=on.
-  if (MemTracker::tracking_level() > NMT_off) {
-    return;
-  }
-
```
tstuefe, line 465: Since appearantly it is possible to release part of a range, we need also a test that tests multi-region-release but with the last region only partly covered.

src/hotspot/share/services/virtualMemoryTracker.cpp
gerard-ziemski: Thank you for the feedback!

I prefer:

```
    while (remaining > 0) {
    }
```

since the size if one of the primary arguments, which the existing logic uses to figure out how to handle it, so using `remaining` fits nicely in my opinion.

You suggested: 

```
// last region may not be fully contained
if (last region not fully contained) {
  split region, release lower half
}
```

but I'd rather not duplicate the logic in the lines 534..546

Hopefully, I accomplished the same functionality within the `while` loop using `remove_size` variable, that you are OK with?

test/hotspot/gtest/runtime/test_os.cpp
gerard-ziemski: Added `partial_release_multi_mappings` test to cover this.

test/hotspot/gtest/runtime/test_os.cpp
```diff
@@ -510,6 +510,47 @@ TEST_VM(os, release_multi_mappings) {
...
+  PRINT_MAPPINGS("C");
+
+  // Clean up. Release all mappings except a 1/2 stripe sized portion at the very end.
+  {
+    WINDOWS_ONLY(NUMASwitcher b(true);) // allow release_memory to release multiple regions
+    ASSERT_TRUE(os::release_memory((char*)p, partial_release_len));
+  }
+}
+#endif // !AIX
+
```
tstuefe, line 553: - This tests case B, and we need case C and D too.
- You should make the first - middle - release the testing release with the partwise releasing. The final release is just to clean up the mess. For that to work, the stripes must be larger and a multiple of os::allocation_granularity, since otherwise you cannot release them partwise.

Would be cool if you fit all of these cases into a single function, e.g. 
```
static void test_multi_release(size_t stripe_size, int num_stripes, size_t release_range_offset, size_t release_range_size)
```

And then call them 4 times in 4 tests, for each of the test cases. E.g. with a stripe len set to 1M could be like this:

```
TEST_VM(TEST_VM, release_multi_mappings_full_regions) {
  test_multi_release(M, 4, M, 2*M); // release stripes 2 and 3 fully
}

TEST_VM(TEST_VM, release_multi_mappings_partial_regions) {
  test_multi_release(M, 4, M + 512*K, 512*K + M); // release stripe 2 partly, stripe 3 fully
}

TEST_VM(TEST_VM, release_multi_mappings_partial_regions) {
  test_multi_release(M, 4, M, M + 512*K); // release stripe 2 fully, stripe 3 partly
}

TEST_VM(TEST_VM, release_multi_mappings_partial_regions) {
  test_multi_release(M, 4, M + 256K, 512 * K); // release a 512K hole into stripe 2
}

```

tstuefe (changes requested):

test/hotspot/gtest/runtime/test_os.cpp
gerard-ziemski: I tried to make a picture of the memory and the parts that should get released for testing:

```
//  allocate              [aaaa][bbbb][cccc][dddd]

//  1 release partial     [....][bbbb][cccc][....]
//  2 release partial     [....][..bb][cccc][....]
//  3 release partial     [....][bb..][cccc][....]
//  4 release partial     [....][bbbb][cc..][....]
//  5 release partial     [....][bbbb][..cc][....]
//  6 release partial     [....][..bb][cc..][....]
//  7 release partial     [....][bb..][..cc][....]
//  8 release partial     [....][bb..][cc..][....]
//  9 release partial     [....][..bb][..cc][....]
// 10 release partial     [....][.bb.][....][....]
// 11 release partial     [....][.bb.][.cc.][....]
```

Are we still missing any cases?

Did I add too much? (i.e. drop case 11?)

gerard-ziemski: I think I see 3 more cases for the release pattern that we can add:

```
  [aaaa][bbbb][cccc][dddd]

1 [....][xxxx][xxxx][....]
2 [....][..xx][xxxx][....]
3 [....][xxxx][xx..][....]
4 [....][..xx][xx..][....]

5 [....][..xx][....][....]
6 [....][....][xx..][....]
7 [....][.xx.][....][....]
```

Aren't cases 5,6 and 7 unique?

tstuefe: > Aren't cases 5,6 and 7 unique?

Are they not already covered by existing tests? Since that functionality already exists?

gerard-ziemski: I might be missing it, but `grep -crn "release_memory(" test/hotspot/gtest` shows:

```
test/hotspot/gtest/memory/test_virtualspace.cpp:2
test/hotspot/gtest/runtime/test_os.cpp:20
test/hotspot/gtest/runtime/test_committed_virtualmemory.cpp:1
test/hotspot/gtest/nmt/test_nmt_locationprinting.cpp:1
```

And when I look at the files I see:

`test/hotspot/gtest/runtime/test_os.cpp`
- tests bad ranges on Windows
- tests striped reserve, followed by one full release, one full re-reserve, then release again
- tests os::print_memory_mappings()
- tests os::win32::find_mapping()
- tests `stripe-stripe-stripe-stripe-stripe-stripe` reserve, followed by `stripe-______-______-______-______-stripe` release, then re-reserve middle stripes, then full release

`test/hotspot/gtest/memory/test_virtualspace.cpp`
- tests different sizes, alignments, one at a time

`test/hotspot/gtest/runtime/test_committed_virtualmemory.cpp`
- tests whether the memory is resident

`test/hotspot/gtest/nmt/test_nmt_locationprinting.cpp`
- tests NMT print_containing_region mechanism

So, no, I don't see dedicated partial release tests of single ranges.

tstuefe: You may be right, then. If you feel up for it, cases 5-7 would also be good to test. With the proposed setup, it would not be much more work. Thanks!

gerard-ziemski:
Added support for more partial regions, made test more flexible and used it to cover more cases.

gerard-ziemski:
ping

src/hotspot/share/services/virtualMemoryTracker.cpp
```diff
@@ -523,29 +525,58 @@ bool VirtualMemoryTracker::remove_released_region(address addr, size_t size) {
...
+    // Segment the region into stripes to be further processed recursively
+    while (remaining > 0) {
+      ReservedMemoryRegion* remove_rgn = node_rgn->data();
+      assert(remove_rgn != nullptr, "NULL region");
+
+      node_rgn = node_rgn->next();
+      assert(remove_rgn->base() < node_rgn->data()->base(), "not ascending bases");
+
+      // Allow for the segment to be partially released
+      long offset = remove_rgn->base() - addr;
```
tstuefe, line 563: Just a fly-by (no time to look at this closer, sorry): long is incorrect for Windows (LLP64). ptrdiff_t is better but has a number of caveats. hotspot usually uses pointer_delta(), with a preceding test for order.

gerard-ziemski: Thank you! Done.

src/hotspot/share/services/virtualMemoryTracker.cpp
```diff
@@ -523,29 +525,56 @@ bool VirtualMemoryTracker::remove_released_region(address addr, size_t size) {
...
-      return false;
-    } else {
-      reserved_rgn->move_committed_regions(addr, *new_rgn->data());
-      return true;
+      remaining -= remove_size;
+      addr += remove_size;
     }
+    return true;
   }
+  return false;
```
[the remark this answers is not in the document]
gerard-ziemski: It's not, but I left it there for any code checkers not to get confused, such as parfait.

This used to be needed, not 100% if such tools got smarter, not to complain, so easier to just make sure we have it covered.

src/hotspot/share/services/virtualMemoryTracker.cpp
```diff
@@ -523,29 +525,56 @@ bool VirtualMemoryTracker::remove_released_region(address addr, size_t size) {
...
-    address high_base = addr + size;
-    ReservedMemoryRegion high_rgn(high_base, top - high_base,
-      *reserved_rgn->call_stack(), reserved_rgn->flag());
+    address end = addr+size;
+    size_t remaining = size;
+    LinkedListNode<ReservedMemoryRegion>* node_rgn = _reserved_regions->find_node(rgn);
+    // Segment the region into stripes to be further processed recursively
+    while (remaining > 0) {
+      ReservedMemoryRegion* remove_rgn = node_rgn->data();
+      assert(remove_rgn != nullptr, "NULL region");
```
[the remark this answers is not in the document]
gerard-ziemski: Thanks, done.

src/hotspot/share/services/virtualMemoryTracker.cpp
```diff
@@ -523,29 +525,56 @@ bool VirtualMemoryTracker::remove_released_region(address addr, size_t size) {
...
+      reserved_rgn->exclude_region(addr, top - addr);
+      LinkedListNode<ReservedMemoryRegion>* new_rgn = _reserved_regions->add(high_rgn);
+      if (new_rgn == nullptr) {
+        return false;
+      } else {
+        reserved_rgn->move_committed_regions(addr, *new_rgn->data());
+        return true;
+      }
+    }
   } else {
```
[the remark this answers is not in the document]
gerard-ziemski: This branch will indeed be taken in the multiple regions case (i.e. the synthetic region that we can just  make up, that doesn't align with underlaying real regions and spawns across multiple of them)

gerard-ziemski:
Thank you Johan for feedback!

src/hotspot/share/services/virtualMemoryTracker.cpp
```diff
@@ -523,29 +525,72 @@ bool VirtualMemoryTracker::remove_released_region(address addr, size_t size) {
     }
   }
 
-  VirtualMemorySummary::record_released_memory(size, reserved_rgn->flag());
-
-  assert(reserved_rgn->contain_region(addr, size), "Not completely contained");
-  if (reserved_rgn->base() == addr ||
-      reserved_rgn->end() == addr + size) {
+  if (size <= reserved_rgn->size() && reserved_rgn->contain_region(addr, size)) {
```
jdksjolen, line 528: A region which contains another must necessarily have at least the size of the contained region, therefore the size check is unnecessary.

src/hotspot/share/services/virtualMemoryTracker.cpp
```diff
@@ -523,29 +525,72 @@ bool VirtualMemoryTracker::remove_released_region(address addr, size_t size) {
...
+    // The requested chunk to be freed is smaller or same size as the real
+    // underlaying region and is fully contained within one (i.e. single region)
+    VirtualMemorySummary::record_released_memory(size, reserved_rgn->flag());
+    assert(reserved_rgn->contain_region(addr, size), "Not completely contained");
+    if (reserved_rgn->base() == addr || reserved_rgn->end() == addr + size) {
```
jdksjolen, line 533: This is the one I wanted a comment for! "If the two regions start at the same place, or ends at the same place, then we can shrink the region from the left or right. Otherwise, we'll have to split it up"

src/hotspot/share/services/virtualMemoryTracker.cpp
```diff
@@ -523,29 +525,72 @@ bool VirtualMemoryTracker::remove_released_region(address addr, size_t size) {
...
       reserved_rgn->exclude_region(addr, size);
```
jdksjolen, line 534: This is unfortunately not named "shrink_region_from" or something similar :-(. I don't expect you to rename the function, to be clear :).

src/hotspot/share/services/virtualMemoryTracker.cpp
```diff
@@ -523,29 +525,72 @@ bool VirtualMemoryTracker::remove_released_region(address addr, size_t size) {
...
-    return true;
+      return true;
+    } else {
```
jdksjolen, line 536: OK! We split the region into two, that looks good. What happens to the committed regions that are part of both the removed reserved region and one of the surviving regions (or why not both?).

```
          *   &
          |   |
[---------------]
       [----]
~>
[-----]     [---]
```

What if we have a committed region starting by the `*` and ending at the `&`, what happens to it?

jdksjolen: Easier example:

```
- = reserved
C = committed
D = to be deleted
[---------------]
  [CCCCCCCCC]
     [DDD]
Are the results:
[----]   [-------]
  [CC]   [CC]
Where the left C is part of the left - and the right C is part of the right -
Because it looks like the results might be:
[----]   [-------]
  [CCCCCCCCC]
where the left - owns the committed region
```

jdksjolen: It looks like the tests only check the results of reserved regions.

src/hotspot/share/services/virtualMemoryTracker.cpp
gerard-ziemski: Ah, I see now, you are correct. I will fix it.

src/hotspot/share/services/virtualMemoryTracker.cpp
gerard-ziemski: We still have to split in some cases where the two regions start at the same place, i.e.:

```
aaaa bbbb cccc
xxxx xxxx
```

Where a,b,c are the underlaying strips, and x denotes the chunk we want to free.

src/hotspot/share/services/virtualMemoryTracker.cpp
gerard-ziemski: I'm not 100% liking the name "shrink_region_from", but if we were to find one that we can all agree on, I can factor out that code into its own method.

src/hotspot/share/services/virtualMemoryTracker.cpp
gerard-ziemski: Done.

PR #13814, opened 2023-05-04
8304074: [JMX] Add an approximation of total bytes allocated on the Java heap by the JVM

phohensee:
Please review this addition to com.sun.management.ThreadMXBean that returns the total number of bytes allocated on the Java heap since JVM launch by both terminated and live threads.

Because this PR adds a new interface method, I've updated the JMM_VERSION to 4, but would be happy to update it to 3_1 instead.

---------
### Progress
- [x] Change requires CSR request [JDK-8307396](https://bugs.openjdk.org/browse/JDK-8307396) to be approved

### Issues
 * [JDK-8304074](https://bugs.openjdk.org/browse/JDK-8304074): [JMX] Add an approximation of total bytes allocated on the Java heap by the JVM
 * [JDK-8307396](https://bugs.openjdk.org/browse/JDK-8307396): [JMX] Add an approximation of total bytes allocated on the Java heap by the JVM (**CSR**)

### Reviewers
 * [Mandy Chung](https://openjdk.org/census#mchung) (@mlchung - **Reviewer**) ⚠️ Review applies to [5f2f86bb](https://git.openjdk.org/jdk/pull/13814/files/5f2f86bb50b644fb589dd3da548b0ba8f14a2006)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13814/head:pull/13814` \
`$ git checkout pull/13814`

`$ git checkout pull/13814` \
`$ git pull https://git.openjdk.org/jdk.git pull/13814/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13814`

`$ git pr show -t 13814`

</details>

<a href="https://git.openjdk.org/jdk/pull/13814.diff">https://git.openjdk.org/jdk/pull/13814.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13814#issuecomment-1535343762)


src/hotspot/share/services/management.cpp
```diff
@@ -2095,7 +2095,25 @@ jlong Management::ticks_to_ms(jlong ticks) {
...
-#endif // INCLUDE_MANAGEMENT
+
+// Gets the amount of memory allocated on the Java heap since JVM launch.
+JVM_ENTRY(jlong, jmm_GetAllThreadAllocatedMemory(JNIEnv *env))
+    // There is a race between threads that exit during the loop and calling
+    // exited_allocated_bytes. If the  result is initialized with exited_allocated_bytes,
+    // the final result may be "too small" because a thread might be retired before
+    // the loop gets to it and thus not be counted. If, on the other hand and done
+    // here, exited_allocated_bytes is added after the loop, the final result might be
+    // "too large" because a thread might be counted twice, once in the loop and agsin
```
dholmes-ora, line 2106: typo agsin

src/hotspot/share/services/management.cpp
```diff
@@ -2095,7 +2095,25 @@ jlong Management::ticks_to_ms(jlong ticks) {
   return (jlong)(((double)ticks / (double)os::elapsed_frequency())
                  * (double)1000.0);
 }
-#endif // INCLUDE_MANAGEMENT
+
+// Gets the amount of memory allocated on the Java heap since JVM launch.
+JVM_ENTRY(jlong, jmm_GetAllThreadAllocatedMemory(JNIEnv *env))
+    // There is a race between threads that exit during the loop and calling
+    // exited_allocated_bytes. If the  result is initialized with exited_allocated_bytes,
```
dholmes-ora, line 2102: If you want a stable and accurate value did you try holding the Threads_lock while you iterate the threads? Or do it as a safepoint VMop?

src/java.management/share/classes/sun/management/ThreadImpl.java
```diff
@@ -525,6 +532,7 @@ private static native void getThreadInfo1(long[] ids,
     private static native void getThreadUserCpuTime1(long[] ids, long[] result);
     private static native long getThreadAllocatedMemory0(long id);
     private static native void getThreadAllocatedMemory1(long[] ids, long[] result);
+    private static native long getThreadAllocatedMemory2();
```
shipilev, line 535: We can call this one `getAllThreadAllocatedMemory`, which obviates the need for `2` as the suffix.

src/jdk.management/share/classes/com/sun/management/ThreadMXBean.java
```diff
@@ -121,7 +156,7 @@ public interface ThreadMXBean extends java.lang.management.ThreadMXBean {
      * </pre></blockquote>
      *
      * @return an approximation of the total memory allocated, in bytes, in
-     * heap memory for the current thread
+     * heap memory for the current thread,
```
shipilev, line 159: I am not sure if typos changes in the public API requires a CSR (albeit trivial one). Maybe skip these updates?

test/jdk/com/sun/management/ThreadMXBean/ThreadAllocatedMemory.java
```diff
@@ -201,38 +206,102 @@ private static void testGetThreadsAllocatedBytes()
...
+            }
+        }
+    }
+
+    private static void testGetAllThreadAllocatedBytes()
+        throws Exception {
+
+        // baseline should be positive
+        Thread curThread = Thread.currentThread();
+        long cumulative_size = mbean.getAllThreadAllocatedBytes();
```
shipilev, line 221: Java style for variables is camel-case, `cumulativeSize`.

test/jdk/com/sun/management/ThreadMXBean/ThreadAllocatedMemoryArray.java
```diff
@@ -118,18 +116,8 @@ public static void main(String[] argv)
...
-            if (sizes[i] > newSize) {
-                throw new RuntimeException("TEST FAILED: " +
-                    threads[i].getName() +
-                    " previous allocated bytes = " + sizes[i] +
-                    " > current allocated bytes = " + newSize);
-            }
-            System.out.println(threads[i].getName() +
-                " Previous allocated bytes = " + sizes[i] +
-                " Current allocated bytes = " + newSize);
+            checkResult(threads[i], sizes[i], sizes1[i]);
```
shipilev, line 120: Since we are cleaning up the test anyway, can we / should we rename `sizes` -> `before`, `size1` -> `after`?

test/jdk/com/sun/management/ThreadMXBean/ThreadAllocatedMemoryArray.java
```diff
@@ -173,11 +160,25 @@ public static void main(String[] argv)
     }
 
 
+    private static void checkResult(Thread curThread,
+                                    long prev_size, long curr_size) {
```
shipilev, line 164: camelCase arguments.

test/jdk/com/sun/management/ThreadMXBean/ThreadAllocatedMemory.java
```diff
@@ -298,22 +367,22 @@ public void run() {
...
+            System.out.println(getName() +
+                " ThreadAllocatedBytes before = " + size1 +
+                " ThreadAllocatedBytes after = " + size2);
 
             if (size1 > size2) {
                 throw new RuntimeException(getName() +
-                    " ThreadAllocatedBytes = " + size1 +
-                    " > ThreadAllocatedBytes = " + size2);
+                    " ThreadAllocatedBytes before = " + size1 +
+                    " > ThreadAllocatedBytes after = " + size2);
```
shipilev, line 377: Is this replaceable with `checkResult(...)`?

test/jdk/com/sun/management/ThreadMXBean/ThreadAllocatedMemory.java
```diff
@@ -201,38 +206,102 @@ private static void testGetThreadsAllocatedBytes()
...
         if (size < 0) {
             throw new RuntimeException(
-                "Invalid allocated bytes returned = " + size);
+                "Invalid allocated bytes returned for thread " +
+                curThread.getName() + " = " + size);
         }
     }
 
-    private static void checkResult(Thread curThread,
+    private static long checkResult(Thread curThread,
```
shipilev, line 286: There is another `checkResult` below? Should they be replaced by a single method?

src/hotspot/share/services/management.cpp
shipilev: I agree we should strive to get the value as accurate as possible. I think for operational use at scale, we need to avoid doing safepoints. Holding a `ThreadLock` might also penalize other code that (ab)uses threading (we frequently see thousands of threads coming and going, don't ask).

But I have a fundamental question here: since SMR/TLH gives us a snapshot of currently live threads, and it also protects us from seeing an exiting thread in bad state (ultimately, a `delete`-d one), why can't we just trust its `cooked_allocated_bytes`, and avoid adding allocated bytes on exit path? If we cannot trust that, can we make it trustable while thread is protected by SMR/TLH?

src/hotspot/share/include/jmm.h
```diff
@@ -52,7 +52,9 @@ enum {
   JMM_VERSION_1_2_2 = 0x20010202,
   JMM_VERSION_2   = 0x20020000, // JDK 10
   JMM_VERSION_3   = 0x20030000, // JDK 14
-  JMM_VERSION     = JMM_VERSION_3
+  JMM_VERSION_3_0 = 0x20030000,
```
shipilev, line 55: What's `JMM_VERSION_3_0`?

src/hotspot/share/services/management.cpp
```diff
@@ -2095,7 +2095,25 @@ jlong Management::ticks_to_ms(jlong ticks) {
...
+    // "too large" because a thread might be counted twice, once in the loop and agsin
+    // in exited_allocated_bytes if it's retired after it's encountered in the loop but
+    // before the call to exited_allocated_bytes. A user might use this method to
+    // trigger some sort of alarm, so it was felt best to err on the high side.
+    jlong result = 0;
+    for (JavaThreadIteratorWithHandle jtiwh; JavaThread *thread = jtiwh.next();) {
+      jlong size = thread->cooked_allocated_bytes();
+      result += size;
+    }
+    return result + ThreadService::exited_allocated_bytes();;
```
shipilev, line 2115: Double `;;`.

shipilev (commented):
Some comments follow.

simonis (commented):
Looks good in general. Please find my comments inline.

src/hotspot/share/services/management.cpp
phohensee: I thought about doing those, but both would slow down the app/JVM and very likely introduce p99.9/p100 latency outliers that we'd rather not see just because we're sampling. Also,

1. The existing thread allocated bytes implementation isn't particularly accurate either, in the sense that reality quickly gets away from the method return values. That was a conscious decision to go with speed and efficiency over stability/accuracy, so I went the same way for this implementation.
2. Cloud services sample at fairly long intervals in order to avoid overhead: 1 minute is common and 5 minutes not unheard of. Stability/accuracy within short time frames is unneeded with such long sampling intervals.

test/jdk/com/sun/management/ThreadMXBean/ThreadAllocatedMemory.java
phohensee: No, there's only one checkResult method. I changed the result type from void to long.

src/hotspot/share/services/management.cpp
phohensee: Fixed.

src/hotspot/share/services/management.cpp
phohensee: Afaiu, SMR/TLH keeps a terminated thread's TLS accessible, but doesn't stop the termination process.

If we initialize result with exited_allocated_bytes, the "too small" possibility is still there. We still have a window between that initialization and the creation of the threads iterator where threads may terminate and thus not be included in the iteration. And new threads may become active during the iteration and not be included. As Aleksey observes, an advantage is that threads that terminate during the iteration will be counted only once and we don't have to check for terminated threads.

If we stick with adding exited_allocated_bytes to result after the iteration, Volker is correct that we shouldn't include terminated threads in the sum because their allocated bytes values will have been added to exited_allocated_bytes. We have the same undercount possibility that initializing result with exited_allocated_bytes has, so this (fixed) approach can result in "too small" also, but "too large" goes away.

Looks like both approaches are equivalent, so let's go with initializing result with exited_allocated_bytes because it avoids a comparison in the loop.

src/hotspot/share/services/threadService.hpp
```diff
@@ -102,6 +106,11 @@ class ThreadService : public AllStatic {
   static jlong get_live_thread_count()        { return _atomic_threads_count; }
   static jlong get_daemon_thread_count()      { return _atomic_daemon_threads_count; }
 
+  static jlong exited_allocated_bytes()       { return _exited_allocated_bytes; }
+  static void incr_exited_allocated_bytes(jlong size) {
+    Atomic::add(&_exited_allocated_bytes, size);
```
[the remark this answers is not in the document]
phohensee: Fixed.

src/java.management/share/classes/sun/management/ThreadImpl.java
phohensee: Fixed.

src/jdk.management/share/classes/com/sun/management/ThreadMXBean.java
phohensee: Fixed.

test/jdk/com/sun/management/ThreadMXBean/ThreadAllocatedMemory.java
phohensee: Fixed.

test/jdk/com/sun/management/ThreadMXBean/ThreadAllocatedMemoryArray.java
phohensee: Fixed.

test/jdk/com/sun/management/ThreadMXBean/ThreadAllocatedMemory.java
phohensee: Yes. Done.

src/hotspot/share/include/jmm.h
phohensee: Removed.

src/hotspot/share/services/management.cpp
phohensee: Fixed.

src/hotspot/share/include/jmm.h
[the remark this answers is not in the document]
phohensee: Removed.

src/hotspot/share/include/jmm.h
```diff
@@ -316,7 +318,8 @@ typedef struct jmmInterface_1_ {
   void         (JNICALL *SetVMGlobal)            (JNIEnv *env,
                                                   jstring flag_name,
                                                   jvalue  new_value);
-  void*        reserved6;
+  jlong        (JNICALL *GetAllThreadAllocatedMemory)
```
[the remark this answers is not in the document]
phohensee: Done.

src/hotspot/share/services/management.cpp
```diff
@@ -2264,7 +2279,7 @@ const struct jmmInterface_1_ jmm_interface = {
   jmm_DumpHeap0,
   jmm_FindDeadlockedThreads,
   jmm_SetVMGlobal,
-  nullptr,
+  jmm_GetAllThreadAllocatedMemory,
```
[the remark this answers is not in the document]
phohensee: Done.

src/hotspot/share/services/management.cpp
dholmes-ora: > Afaiu, SMR/TLH keeps a terminated thread's TLS accessible, but doesn't stop the termination process.

Incorrect. A thread cannot complete the termination process if it is contained by a TLH - see ` ThreadsSMRSupport::smr_delete` and the call to `wait_until_not_protected`.

But not sure that helps with the zero-else-double accounting problem. Any read of the "total accumulated bytes written to date" value is racing with terminating threads.

src/hotspot/share/services/management.cpp
```diff
@@ -2095,7 +2095,21 @@ jlong Management::ticks_to_ms(jlong ticks) {
...
+
+// Gets the amount of memory allocated on the Java heap since JVM launch.
+JVM_ENTRY(jlong, jmm_GetAllThreadAllocatedMemory(JNIEnv *env))
+    // The result may be "too small" because the threads iterator takes a
+    // snapshot and new threads may be created after result is initialized.
+    // There's no check for terminated threads in the loop because their
+    // allocated byte counts aren't included in exited_allocated_bytes
+    // when result is initialized.
+    jlong result = ThreadService::exited_allocated_bytes();
+    for (JavaThreadIteratorWithHandle jtiwh; JavaThread *thread = jtiwh.next();) {
```
dholmes-ora, line 2107: If you call `exited_allocated_bytes` whilst you have an active `ThreadsListHandle` then you at least ensure you don't miss accounting for threads that are just about to terminate.

src/hotspot/share/services/threadService.cpp
```diff
@@ -167,6 +170,7 @@ void ThreadService::remove_thread(JavaThread* thread, bool daemon) {
     // We did not get here via JavaThread::exit() so current_thread_exiting()
     // was not called, e.g., JavaThread::cleanup_failed_attach_current_thread().
     decrement_thread_counts(thread, daemon);
+    ThreadService::incr_exited_allocated_bytes(thread->cooked_allocated_bytes());
```
dholmes-ora, line 173: By doing this here you increase the likelihood of double-accounting for this thread. If you do this after the thread is no longer on any threads-list you may miss its contribution entirely, but you won't double-count it.

src/jdk.management/share/classes/com/sun/management/ThreadMXBean.java
```diff
@@ -105,6 +105,41 @@ public interface ThreadMXBean extends java.lang.management.ThreadMXBean {
      */
     public long[] getThreadUserTime(long[] ids);
 
+    /**
+     * Returns an approximation of the total amount of memory, in bytes,
+     * allocated in heap memory since the Java virtual machine was launched,
+     * including the amount allocated by terminated threads.
```
dholmes-ora, line 111: This "including ..." part seems redundant - it is the value allocated since JVM launch.

src/hotspot/share/services/management.cpp
simonis: Do you mean something like:
```cpp
    JavaThreadIteratorWithHandle jtiwh;
    jlong result = ThreadService::exited_allocated_bytes();
    while (JavaThread *thread = jtiwh.next()) {
      ...
```
That would be fine for me. Otherwise I agree with the current compromise between accuracy and speed.

src/hotspot/share/services/threadService.cpp
dcubed-ojdk: If the `thread` is no longer on any ThreadsList, then it is not safe to access `thread`.

dholmes-ora: @dcubed-ojdk This is the current thread acting on itself

src/hotspot/share/services/management.cpp
dholmes-ora: Yes. Though I realize my comment needed to make clear that you need a thread to update the count after it has removed itself from the thread-list, for this to be an improvement.

src/hotspot/share/runtime/atomic.hpp
```diff
@@ -307,8 +307,8 @@ class Atomic : AllStatic {
   struct AddImpl;
 
   // Platform-specific implementation of add.  Support for sizes of 4
-  // bytes and (if different) pointer size bytes are required.  The
-  // class must be default constructable, with these requirements:
+  // and 8 bytes are required.  The class must be default constructable,
```
dholmes-ora, line 310: Comment change seems unnecessary.

src/jdk.management/share/classes/com/sun/management/ThreadMXBean.java
```diff
@@ -106,15 +106,19 @@ public interface ThreadMXBean extends java.lang.management.ThreadMXBean {
...
+     * in heap memory by all threads since the Java virtual machine started.
      * The returned value is an approximation because some Java virtual machine
      * implementations may use object allocation mechanisms that result in a
      * delay between the time an object is allocated and the time its size is
      * recorded. An implementation that walks an active Java thread list will
      * necessarily also deliver an approximate result.
      *
+     * @implSpec The default implementation throws UnsupportedOperationException
+     * if the Java virtual machine implementation does not support thread
+     * memory allocation measuremust, and otherwise acts as though thread
```
dholmes-ora, line 119: Typo: measuremust

src/hotspot/share/services/management.cpp
```diff
@@ -2097,14 +2097,18 @@ jlong Management::ticks_to_ms(jlong ticks) {
...
     // The result may be "too small" because the threads iterator takes a
     // snapshot and new threads may be created after result is initialized.
     // There's no check for terminated threads in the loop because their
     // allocated byte counts aren't included in exited_allocated_bytes
-    // when result is initialized.
-    jlong result = ThreadService::exited_allocated_bytes();
-    for (JavaThreadIteratorWithHandle jtiwh; JavaThread *thread = jtiwh.next();) {
+    // when result is initialized. The iterator is created and initialized
+    // before the call to exited_allocated_bytes to ensure we don't miss
+    // accounting for threads that are just about to terminate.
```
dholmes-ora, line 2107: Sorry this is only accurate if a thread updates `exited_allocated_bytes` after it removes itself from the Threads-list.

src/hotspot/share/services/threadService.cpp
```diff
@@ -217,6 +221,7 @@ void ThreadService::current_thread_exiting(JavaThread* jt, bool daemon) {
   assert(!jt->is_terminated() && jt->is_exiting(), "must be exiting");
 
   decrement_thread_counts(jt, daemon);
+  ThreadService::incr_exited_allocated_bytes(jt->cooked_allocated_bytes());
```
dholmes-ora, line 224: Again this is too soon. This should be deferred until after the thread has removed itself from the threads-list.

dholmes-ora:
Also note that the current placement of the `incr_exited_allocated_bytes` skips accounting for hidden threads because those ThreadService methods skip hidden threads. You need to do this update as one of the last actions of a terminating/detaching thread.

src/hotspot/share/runtime/atomic.hpp
phohensee: I found that I had to add 64-bit atomic add support on 32-bit platforms (working on it), so I changed the comment to match the similar cmpxchg one.

src/hotspot/share/services/threadService.cpp
simonis: This method (i.e. `ThreadService::remove_thread()`) is called from `Threads::remove()` *after* the thread was removed from the thread list:
```cpp
void Threads::remove(JavaThread* p, bool is_daemon) {
  ...
    // Maintain fast thread list
    ThreadsSMRSupport::remove_thread(p);
    ...
    ThreadService::remove_thread(p, is_daemon);
```

But if we reach here from `JavaThread::cleanup_failed_attach_current_thread()` as the comment implies (`JavaThread::cleanup_failed_attach_current_thread()` calls `Threads::remove()` in the case of an attach failure), calling `ThreadService::incr_exited_allocated_bytes()` is probably irrelevant, because a thread which failed to attach can't allocate anyway.

src/hotspot/share/services/management.cpp
```diff
@@ -2098,17 +2094,32 @@ jlong Management::ticks_to_ms(jlong ticks) {
...
+    // the threads list (in ThreadService::remove_thread), which means there's
+    // a race between threads that exit during the loop and reading
+    // exited_allocated_bytes. If result is initialized with exited_allocated_bytes,
+    // the final result may be "too small" because a thread might be removed
+    // from the list before the loop gets to it and thus not be counted. If,
+    // on the other hand, exited_allocated_bytes is added after the loop,
+    // the final result might be "too large" because a thread might be counted
+    // twice, once in the loop and agsin in exited_allocated_bytes if it's
+    // removed from the list after it's encountered in the loop but before
+    // adding exited_allocated_bytes.
```
dholmes-ora, line 2107: This is not an accurate description.

Once we have the TLH we have a fixed set of threads and none of them can update exited_allocated_bytes. If we read exited_allocated_bytes while the TLH is held then the count can be short for two reasons:
1. Newly started threads
2. Terminating threads that already escaped the TLH but which hadn't updated exited_allocated_bytes by the time we read it.
AFAICS with the current logic there is no possibility of double-accounting.

dholmes-ora: Sorry I thought you had made the change to only update exited_allocated_bytes after the thread removed itself from the threads-list. It really makes things so much simpler.

phohensee: I made that change. The comment now reads:
    // A thread increments exited_allocated_bytes in ThreadService::remove_thread                                                                                                                                                                                                                                                                                           
    // only after it removes itself from the threads list, and once a TLH is                                                                                                                                                                                                                                                                                                
    // created, no thread it references can remove itself from the threads                                                                                                                                                                                                                                                                                                  
    // list, so none can update exited_allocated_bytes. We therefore initialize                                                                                                                                                                                                                                                                                             
    // result with exited_allocated_bytes after after we create the TLH so that                                                                                                                                                                                                                                                                                             
    // the final result can only be short due to (1) threads that start after                                                                                                                                                                                                                                                                                               
    // the TLH is created, or (2) terminating threads that escape TLH creation                                                                                                                                                                                                                                                                                              
    // and don't update exited_allocated_bytes before we initialize result.

src/hotspot/share/services/threadService.cpp
phohensee: Done.

src/hotspot/share/services/management.cpp
phohensee: Fixed, see below.

src/jdk.management/share/classes/com/sun/management/ThreadMXBean.java
phohensee: Fixed.

src/hotspot/share/services/management.cpp
phohensee: Resolved, see below.

src/hotspot/share/services/management.cpp
phohensee: Done as Volker suggests.

src/hotspot/share/services/threadService.cpp
phohensee: Done.

dholmes-ora (changes requested):
Sorry to be a pain but can you split the Atomic changes out of this PR into their own issue please.

Though having said that, now that you call incr_exited_allocayed_bytes() where you do, it is protected by the Threads_lock so no atomic add is needed now.

simonis:
Looks good now except for the points already mentioned by @dholmes-ora .

src/jdk.management/share/classes/com/sun/management/ThreadMXBean.java
phohensee: Removed.

src/hotspot/share/runtime/atomic.hpp
phohensee: Reverted, since 64-bit atomic add is no longer needed.

src/hotspot/share/services/#allocationSite.hpp#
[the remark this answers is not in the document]
phohensee: Removed.

src/hotspot/share/services/.#allocationSite.hpp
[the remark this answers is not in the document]
phohensee: Removed.

src/hotspot/share/services/threadService.hpp
```diff
@@ -61,6 +61,10 @@ class ThreadService : public AllStatic {
   static PerfVariable* _peak_threads_count;
   static PerfVariable* _daemon_threads_count;
 
+  // As could this...
+  // Number of heap bytes allocated by termianted threads.
```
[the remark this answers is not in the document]
phohensee: Fixed.

phohensee:
Makes things very much simpler without needing 64-bit atomic add. Hopefully all cleaned up.

src/hotspot/share/services/threadService.hpp
```diff
@@ -102,6 +106,12 @@ class ThreadService : public AllStatic {
   static jlong get_live_thread_count()        { return _atomic_threads_count; }
   static jlong get_daemon_thread_count()      { return _atomic_daemon_threads_count; }
 
+  static jlong exited_allocated_bytes()       { return _exited_allocated_bytes; }
+  static void incr_exited_allocated_bytes(jlong size) {
+    // No need for atomicity, method is called under the Threads_lock
+    _exited_allocated_bytes += size;
+  }
```
dholmes-ora, line 113: As there is a lock-free read however you need to use Atomic::load and Atomic::store - epsecially for 32-bit systems.

test/jdk/com/sun/management/ThreadMXBean/ThreadAllocatedMemory.java
```diff
@@ -201,38 +206,102 @@ private static void testGetThreadsAllocatedBytes()
...
+        // baseline should be positive
+        Thread curThread = Thread.currentThread();
+        long cumulativeSize = mbean.getTotalThreadAllocatedBytes();
+        if (cumulativeSize <= 0) {
+            throw new RuntimeException(
+                "Invalid allocated bytes returned for " + curThread.getName() + " = " + cumulativeSize);
+        }
+
+        // start threads
+        done = false; done1 = false;
```
dholmes-ora, line 228: Nit: separate lines please

dholmes-ora (changes requested):

src/hotspot/share/services/threadService.hpp
phohensee: Changed. Not sure atomic load is needed.

test/jdk/com/sun/management/ThreadMXBean/ThreadAllocatedMemory.java
phohensee: Fixed.

src/hotspot/share/services/threadService.hpp
phohensee: Even though this version passed the pre-submit tests, which include the new versions of ThreadAllocatedBytes.java, running it more often on a host with many hardware threads resulted in failures where getTotalAllocatedBytes did not monotonically increase. I believe what happened was that without an acquire/release, the hardware can float earlier stores to _exited_allocated_bytes above later ones. Adding acquire/release fixed the issue.

dholmes-ora: That should not be possible if the update is always under the lock. acquire/release control what other memory updates you see when you see "this" update - they have no affect on the update itself.

src/hotspot/share/services/threadService.hpp
```diff
@@ -108,8 +108,13 @@ class ThreadService : public AllStatic {
 
   static jlong exited_allocated_bytes()       { return _exited_allocated_bytes; }
```
dholmes-ora, line 109: This needs to be an atomic::load too .

src/hotspot/share/services/threadService.hpp
```diff
@@ -108,8 +108,13 @@ class ThreadService : public AllStatic {
...
   static void incr_exited_allocated_bytes(jlong size) {
-    // No need for atomicity, method is called under the Threads_lock
-    _exited_allocated_bytes += size;
+    // No need for an atomic add because called under the Threads_lock,
+    // but because _exited_allocated_bytes is read concurrently, need
+    // atomic store to avoid readers seeing a partial update. Also need
+    // acquire/release because the hardware might move a subsequent
+    // store to _exited_allocated_bytes above this one.
+    jlong old = Atomic::load_acquire(&_exited_allocated_bytes);
+    Atomic::release_store(&_exited_allocated_bytes, old + size);
```
dholmes-ora, line 117: First yes Atomic::load is not needed here due to the lock, it is needed in the accessor above - sorry that wasn't clear. The Atomic store is needed of course.

But the load_acquire is definitely not needed as there is an implicit acquire when the lock is acquired, so any load of the field here must see the last value written under the lock. The release_store is also not needed because a release is for ensuring visibility of previous stores of which there are none, and has no affect on the visibility of the store done by the `release_store`.

So lets imagine a really weakly-ordered piece of hardware where all of the writing threads are completely properly synchronized by the use of the lock (else the hw is broken), but a lock-free reader can potentially see any of those writes out-of-order.  I think the only way to guarantee seeing the most recent write would be to issue a full fence() before the load. But I would be very surprised if any of our hardware actually operates in such a fashion.

dholmes-ora (changes requested):

src/hotspot/share/services/threadService.hpp
phohensee: Thanks for the explanation. I reverted the acquire/release and the atomic load in incr_exited_allocated_bytes and added an atomic load in exited_allocated_bytes.

src/hotspot/share/services/threadService.hpp
phohensee: Done.

dholmes-ora (commented):
Updates look good - thanks.

But presumably you still have those test failures now the acquire/release has been removed again?

phohensee:
Nope, test failures went away with the latest revision. I'm at a loss on that one. Perhaps it was the atomic load.

dholmes-ora (commented):
I'm happy with the basic implementation of this now - thanks. I'll leave any issues about spec and JMM version for serviceability folk to approve.

dholmes-ora (approved):
I will click Approve to get rid of the "requested changes" indicator but please wait for serviceability reviews.

src/hotspot/share/services/management.cpp
```diff
@@ -2095,7 +2091,25 @@ jlong Management::ticks_to_ms(jlong ticks) {
...
+// Gets the amount of memory allocated on the Java heap since JVM launch.
+JVM_ENTRY(jlong, jmm_GetTotalThreadAllocatedMemory(JNIEnv *env))
+    // A thread increments exited_allocated_bytes in ThreadService::remove_thread
+    // only after it removes itself from the threads list, and once a TLH is
+    // created, no thread it references can remove itself from the threads
+    // list, so none can update exited_allocated_bytes. We therefore initialize
+    // result with exited_allocated_bytes after after we create the TLH so that
+    // the final result can only be short due to (1) threads that start after
+    // the TLH is created, or (2) terminating threads that escape TLH creation
+    // and don't update exited_allocated_bytes before we initialize result.
```
dholmes-ora, line 2104: Okay this is why you can't do monotonicity check in the tests. Imagine we have 10 threads all live and we call this function and so tally up all 10 threads. The next time we call this 5 of the threads are in the process of terminating - they have escaped the TLH but not yet updated the exited_allocated_bytes - so now we only tally up 5 threads and so the total appears to be far less than last time.

src/jdk.management/share/classes/com/sun/management/ThreadMXBean.java
```diff
@@ -105,6 +105,42 @@ public interface ThreadMXBean extends java.lang.management.ThreadMXBean {
...
 
+    /**
+     * Returns an approximation of the total amount of memory, in bytes, allocated
+     * in heap memory by all threads since the Java virtual machine started.
+     * The returned value is an approximation because some Java virtual machine
+     * implementations may use object allocation mechanisms that result in a
+     * delay between the time an object is allocated and the time its size is
+     * recorded.
+     *
+     * @implSpec The default implementation throws UnsupportedOperationException
```
mlchung, line 116:
```suggestion
     * @implSpec The default implementation throws {@code UnsupportedOperationException}
```

test/jdk/com/sun/management/ThreadMXBean/ThreadAllocatedMemory.java
```diff
@@ -152,7 +158,7 @@ private static void testGetThreadAllocatedBytes()
         try {
             curThread.join();
         } catch (InterruptedException e) {
```
mlchung, line 160: should it just let `InterruptedException` be thrown rather than swallowing it?  This pattern appears multiple places in this test and ThreadAllocatedMemoryArray.java test.

mlchung (commented):
The specification looks good.   A minor comment in the test.

phohensee:
Yes, it seems to have happened when I did a merge of this branch with jdk master, which latter must have been in a strange state. Unfortunately, git merge didn't complain at all, so I didn't check before pushing.

test/jdk/com/sun/management/ThreadMXBean/ThreadAllocatedMemory.java
phohensee: The original implementation grabbed this code from, e.g., test/jdk/java/lang/management/ThreadMXBean/ThreadUserTime.java. In the latter, InterruptedException causes the test to fail, but for some reason now lost, ThreadAllocatedMemory.java doesn't do that, but ThreadAllocatedMemoryArray.java does.

dholmes-ora:
@phohensee not clear if you saw my comment a couple of days ago about why the monotonic increasing check in the test can't work.

src/hotspot/share/services/management.cpp
kevinjwalls: The tests and all apps might thank us for keeping last known value, and not returning anything lower, ensuring montonicity.

phohensee:
Yes, I saw your comment and have been thinking about how to fix it, but haven't come up with anything. Ramki (ysr) pointed out that walking the thread list using SMR is very expensive when there are 1000s of threads, and that there isn't much overhead to collecting per-thread statistics: a thread local counter is incremented when TLABs are retired or there are direct allocations of large objects in eden or the old gen. So, he suggested adding an atomic increment of a global counter whenever a thread local counter is incremented. Atomic operations have been completing in the LLC (i.e., on-chip) for some time now, so the overhead is pretty small. What do you think, shall I reimplement that way?

test/jdk/com/sun/management/ThreadMXBean/ThreadAllocatedMemory.java
mlchung: Since you are on this file, it may be good to clean them up.

dholmes-ora:
A global allocation counter sounds like a potential bottleneck to me and everyone has to pay for it. I'd want to see broad performance measurements for that and hear what the GC folk have to say.

dholmes-ora:
A global counter would also need to resurrect Atomic::add(jlong) for all platforms.

dholmes-ora:
The monotonicity issue can be addressed by tracking the last total and never reporting a smaller value. (Much as we used to cache the current time to avoid it going backwards.)

src/hotspot/share/services/management.cpp
phohensee: Fixed using a high water mark.

phohensee:
I added a high water mark to getTotalThreadAllocatedBytes (under a lock, lest we have a monotonicity issue on the high water mark), and updated the tests per Mandy's request.

test/jdk/com/sun/management/ThreadMXBean/ThreadAllocatedMemory.java
phohensee: Done.

mlchung:
Thanks for updating the test.   The spec looks good.   I'll leave the hotspot implementation for David and others to approve.

mlchung (approved):

phohensee:
Thanks for your review, Mandy.

test/jdk/com/sun/management/ThreadMXBean/ThreadAllocatedMemory.java
```diff
@@ -300,11 +302,16 @@ private static long checkResult(Thread curThread,
         return currSize;
     }
 
+    private static void reportUnexpected(Exception e, String reason) {
```
dholmes-ora, line 305: nit: 'when' or 'where' seems a more appropriate name than 'reason'

dholmes-ora (commented):
High watermark changes seem fine in principle but not sure about borrowing the `MonitoringSupport_lock ` to do the locking. At a minimum this comment should be expanded:
```
./share/runtime/mutexLocker.hpp:extern Mutex*   MonitoringSupport_lock;          // Protects updates to the serviceability memory pools.
```

test/jdk/com/sun/management/ThreadMXBean/ThreadAllocatedMemory.java
phohensee: Changed to "when".

phohensee:
Changed the MonitoringSupport_lock comment to "Protects updates to the serviceability memory pools and allocated memory high water mark". The lock was introduced to serialize updates to the G1 memory pool stats, but given that doesn't happen very often, and the name, imo it's reasonable to use it in a more general fashion.

dholmes-ora (approved):
Nothing further from me. Thanks.

phohensee:
Thanks for your in-depth review!

src/hotspot/share/services/management.cpp
simonis: A simple fix for guaranteeing monotonicity, would be to add another field (e.g. `max_allocated_bytes`) which keeps the last result returned by `getTotalThreadAllocatedBytes()` and the latter would then return the maximum of the current computation and `max_allocated_bytes`.

phohensee: That's what I did. :) Thanks for your review!

PR #13818, opened 2023-05-04
8306843: JVMTI tag map extremely slow after JDK-8292741

coleenp:
The ResourceHashtable conversion for JDK-8292741 didn't add the resizing code.  The old hashtable code was tuned for resizing in anticipation of large hashtables for JVMTI tags.  This patch ports over the old hashtable resizing code.  It also adds a ResourceHashtable::put_fast() function that prepends to the bucket list, which is also reclaims the performance of the old hashtable for this test with 10M tags.  The ResourceHashtable put function is really a put_if_absent. This can be cleaned up in a future change.  Also, the remove function needed a lambda to destroy the WeakHandle, since resizing requires copying entries.

Tested with JVMTI and JDI tests locally, and tier1-4 tests.

---------
### Progress

### Issue
 * [JDK-8306843](https://bugs.openjdk.org/browse/JDK-8306843): JVMTI tag map extremely slow after JDK-8292741

### Reviewers
 * [Ioi Lam](https://openjdk.org/census#iklam) (@iklam - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13818/head:pull/13818` \
`$ git checkout pull/13818`

`$ git checkout pull/13818` \
`$ git pull https://git.openjdk.org/jdk.git pull/13818/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13818`

`$ git pr show -t 13818`

</details>

<a href="https://git.openjdk.org/jdk/pull/13818.diff">https://git.openjdk.org/jdk/pull/13818.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13818#issuecomment-1535495126)

### DIFF
--- a/src/hotspot/share/classfile/stringTable.cpp
+++ b/src/hotspot/share/classfile/stringTable.cpp
@@ -631,12 +631,11 @@ class VerifyCompStrings : StackObj {
     return java_lang_String::equals(a, b);
   }
 
-  ResizeableResourceHashtable<oop, bool,
-                              AnyObj::C_HEAP, mtInternal,
+  ResizeableResourceHashtable<oop, bool, AnyObj::C_HEAP, mtInternal,
                               string_hash, string_equals> _table;
  public:
   size_t _errors;
-  VerifyCompStrings() : _table(unsigned(_items_count / 8) + 1), _errors(0) {}
+  VerifyCompStrings() : _table(unsigned(_items_count / 8) + 1, 0 /* do not resize */), _errors(0) {}
   bool operator()(WeakHandle* val) {
     oop s = val->resolve();
     if (s == nullptr) {
--- a/src/hotspot/share/prims/jvmtiTagMapTable.cpp
+++ b/src/hotspot/share/prims/jvmtiTagMapTable.cpp
@@ -127,7 +127,6 @@ void JvmtiTagMapTable::remove(oop obj) {
   JvmtiTagMapKey jtme(obj);
   auto clean = [] (const JvmtiTagMapKey& entry, jlong tag) {
     entry.release_weak_handle();
-    return true;
   };
   _table.remove(jtme, clean);
 }
--- a/src/hotspot/share/prims/jvmtiTagMapTable.cpp
+++ b/src/hotspot/share/prims/jvmtiTagMapTable.cpp
@@ -109,7 +109,7 @@ void JvmtiTagMapTable::add(oop obj, jlong tag) {
   bool is_added;
   if (obj->fast_no_hash_check()) {
     // Can't be in the table so add it fast.
-    is_added = _table.put_fast(new_entry, tag);
+    is_added = _table.put_when_absent(new_entry, tag);
   } else {
     jlong* value = _table.put_if_absent(new_entry, tag, &is_added);
     *value = tag; // assign the new tag
--- a/src/hotspot/share/prims/jvmtiTagMapTable.cpp
+++ b/src/hotspot/share/prims/jvmtiTagMapTable.cpp
@@ -35,25 +35,23 @@ JvmtiTagMapKey::JvmtiTagMapKey(oop obj) : _obj(obj) {}
 
 JvmtiTagMapKey::JvmtiTagMapKey(const JvmtiTagMapKey& src) {
   // move object into WeakHandle when copying into the table
-  assert(src._obj != nullptr, "must be set");
+  if (src._obj != nullptr) {
 
-  // obj was read with AS_NO_KEEPALIVE, or equivalent, like during
-  // a heap walk.  The object needs to be kept alive when it is published.
-  Universe::heap()->keep_alive(src._obj);
+    // obj was read with AS_NO_KEEPALIVE, or equivalent, like during
+    // a heap walk.  The object needs to be kept alive when it is published.
+    Universe::heap()->keep_alive(src._obj);
 
-  _wh = WeakHandle(JvmtiExport::weak_tag_storage(), src._obj);
-  _obj = nullptr;
+    _wh = WeakHandle(JvmtiExport::weak_tag_storage(), src._obj);
+    _obj = nullptr;
+  } else {
+    // resizing needs to create a copy.
+    _wh = src._wh;
+    _obj = nullptr;
+  }
 }
 
-JvmtiTagMapKey::~JvmtiTagMapKey() {
-  // If obj is set null it out, this is called for stack object on lookup,
-  // and it should not have a WeakHandle created for it yet.
-  if (_obj != nullptr) {
-    _obj = nullptr;
-    assert(_wh.is_null(), "WeakHandle should be null");
-  } else {
-    _wh.release(JvmtiExport::weak_tag_storage());
-  }
+void JvmtiTagMapKey::release_weak_handle() const {
+  _wh.release(JvmtiExport::weak_tag_storage());
 }
 
 oop JvmtiTagMapKey::object() const {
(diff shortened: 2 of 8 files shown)


src/hotspot/share/utilities/resizeableResourceHash.hpp
```diff
@@ -85,19 +85,44 @@ class ResizeableResourceHashtable : public ResourceHashtableBase<
                                      K, V, ALLOC_TYPE, MEM_TYPE, HASH, EQUALS>;
   using Node = ResourceHashtableNode<K, V>;
   NONCOPYABLE(ResizeableResourceHashtable);
+
+  // Calculate next "good" hashtable size based on requested count
+  int calculate_resize(bool use_large_table_sizes) const {
+    const int resize_factor    = 2.0;     // by how much we will resize using current number of entries
```
sspitsyn, line 91: Nit: extra spaces brefore the '=' sign.
Q: Why is a FP constant assigned to the integer variable?

src/hotspot/share/utilities/resourceHash.hpp
```diff
@@ -202,14 +223,15 @@ class ResourceHashtableBase : public STORAGE {
...
-  bool remove(K const& key) {
+  template<typename Function>
+  bool remove(K const& key, Function function) {
     unsigned hv = HASH(key);
     Node** ptr = lookup_node(hv, key);
 
     Node* node = *ptr;
     if (node != nullptr) {
       *ptr = node->_next;
+      bool cont = function(node->_key, node->_value);
```
sspitsyn, line 234: Q: The local `cont` is not used. Just wanted to check if anything is missed here.
     Also, what does this name mean? Should it be named `cond` instead?

sspitsyn (commented):
Thank you fore taking care about these performance issue!
I've posted a couple of comments but am still looking at it.
It is hard to make sure the changes are fully correct.

src/hotspot/share/utilities/resourceHash.hpp
coleenp: The 'cont' variable was because I cut/pasted the lambda from iterate and in that case means to continue.  That's also not needed for 'remove' so I removed the return variable for the lambda function.

coleenp:
Serguei, thank you for doing a first pass.

src/hotspot/share/utilities/resizeableResourceHash.hpp
```diff
@@ -85,19 +85,44 @@ class ResizeableResourceHashtable : public ResourceHashtableBase<
                                      K, V, ALLOC_TYPE, MEM_TYPE, HASH, EQUALS>;
   using Node = ResourceHashtableNode<K, V>;
   NONCOPYABLE(ResizeableResourceHashtable);
+
+  // Calculate next "good" hashtable size based on requested count
+  int calculate_resize(bool use_large_table_sizes) const {
+    const int resize_factor = 2;     // by how much we will resize using current number of entries
```
iklam, line 91: Does this function depend on the template parameters? If not, I think it can be made a static function -- you may need to pass `BASE::number_of_entries()` in as a parameter.

src/hotspot/share/classfile/stringTable.cpp
```diff
@@ -631,12 +631,11 @@ class VerifyCompStrings : StackObj {
...
   }
 
-  ResizeableResourceHashtable<oop, bool,
-                              AnyObj::C_HEAP, mtInternal,
+  ResizeableResourceHashtable<oop, bool, AnyObj::C_HEAP, mtInternal,
                               string_hash, string_equals> _table;
  public:
   size_t _errors;
-  VerifyCompStrings() : _table(unsigned(_items_count / 8) + 1), _errors(0) {}
+  VerifyCompStrings() : _table(unsigned(_items_count / 8) + 1, 0 /* do not resize */), _errors(0) {}
```
iklam, line 638: Shouldn't this use a regular ResourceHashtable instead?

src/hotspot/share/utilities/resourceHash.hpp
```diff
@@ -136,6 +138,25 @@ class ResourceHashtableBase : public STORAGE {
     }
   }
 
+ /**
+  * Inserts a value in the front of the table.
+  * Since the table is locked, this is ok.
+  * @return: true: a new item is always added
+  */
+  bool put_fast(K const& key, V const& value) {
+    unsigned hv = HASH(key);
```
iklam, line 147: I think `put_fast` is not clear enough. Maybe `put_must_be_absent()` or something more concise.

iklam (commented):
I can't comment on the JVMTI changes, but the changes in the hashtable code seems OK to me.

src/hotspot/share/utilities/resourceHash.hpp
dholmes-ora: I would suggest `put_when_absent` to complement `put_if_absent` - with suitable descriptive comments of course.

iklam: I cannot tell the difference between `put_when_absent` and `put_if_absent`. Grammatically they mean the same thing to me.

coleenp: My preference is to eventually make 'put' be 'put-ifwhen-absent', so I don't care which name you two pick.

src/hotspot/share/classfile/stringTable.cpp
coleenp: I filed an RFE then closed it, see JDK-8307623.

src/hotspot/share/utilities/resourceHash.hpp
dholmes-ora: `put_when_known_absent`? 

A basic `put` should either add or replace; a `put_if_absent` should only add else do nothing.

sspitsyn:
Not sure, if `put_if_absent_fast` is worth to consider. :)

src/hotspot/share/prims/jvmtiTagMapTable.cpp
```diff
@@ -35,27 +35,25 @@ JvmtiTagMapKey::JvmtiTagMapKey(oop obj) : _obj(obj) {}
...
-  if (_obj != nullptr) {
+    _wh = WeakHandle(JvmtiExport::weak_tag_storage(), src._obj);
     _obj = nullptr;
-    assert(_wh.is_null(), "WeakHandle should be null");
   } else {
-    _wh.release(JvmtiExport::weak_tag_storage());
+    // resizing needs to create a copy.
+    _wh = src._wh;
+    _obj = nullptr;
   }
```
sspitsyn, line 50: There can be just one line at 51 instead of two lines at 45 and 49.
Then, I do not see where in the class `JvmtiTagMapKey` the `_obj` can obtain non nullptr value.

src/hotspot/share/utilities/resourceHash.hpp
coleenp: put_when_absent is what I have and it's fine.  I don't think we need more sentence names or changing doesn't materially improve this patch.  I was comparing to the std::unordered_map class which we want to minimally emulate and insert does insert if absent, so we shouldn't rewrite "put" to mean put_if/when_absent, but the existing behavior was surprising and unexpected to me.

https://en.cppreference.com/w/cpp/container/unordered_map/insert

src/hotspot/share/prims/jvmtiTagMapTable.cpp
coleenp: Yes, _obj should always be null after copying either because we've transferred the oop over to the WeakHandle or that it's a copy so already a WeakHandle.  I admit that having _obj assigned to null in both places looks odd, but it was for two different reasons.  I added a comment too.

iklam (approved):
LGTM

coleenp:
Thank you Serguei and Ioi.

src/hotspot/share/utilities/resourceHash.hpp
dholmes-ora: Well ... many of our API's are more Java oriented in naming rather than C++ containers. And unordered_map does not strike me as something we even want to minimally emulate when it comes to method naming. YMMV.

PR #13821, opened 2023-05-05
8291065: Creating a VarHandle for a static field triggers class initialization

liach:
This patch implements lazy initialization for VarHandle working on static fields. It has a good initial call performance.

We introduce a new internal API, `target()` to unpack a lazy VarHandle in VH implementation methods. If called via MethodHandle, a barrier is added in the MethodHandle instead.

The new test ensures the correctness of Lazy VH for both direct and indirect invocation; the performance of MethodHandle version of lazy VH is not yet tested.

```
Benchmark                                            Mode  Cnt    Score    Error  Units
LazyStaticColdStart.methodHandleCreateEager            ss   10   41.490 ± 12.331  us/op
LazyStaticColdStart.methodHandleCreateLazy             ss   10   21.810 ± 16.964  us/op
LazyStaticColdStart.methodHandleInitializeCallEager    ss   10   57.860 ± 13.738  us/op
LazyStaticColdStart.methodHandleInitializeCallLazy     ss   10   93.300 ± 18.858  us/op
LazyStaticColdStart.varHandleCreateEager               ss   10   39.860 ±  9.362  us/op
LazyStaticColdStart.varHandleCreateLazy                ss   10   17.630 ±  1.111  us/op
LazyStaticColdStart.varHandleInitializeCallEager       ss   10  123.170 ± 62.468  us/op
LazyStaticColdStart.varHandleInitializeCallLazy        ss   10  105.390 ± 41.815  us/op
```

---------
### Progress
- [x] Change requires CSR request [JDK-8311075](https://bugs.openjdk.org/browse/JDK-8311075) to be approved

### Issues
 * [JDK-8291065](https://bugs.openjdk.org/browse/JDK-8291065): Creating a VarHandle for a static field triggers class initialization (**Bug** - P4)
 * [JDK-8311075](https://bugs.openjdk.org/browse/JDK-8311075): Creating a VarHandle for a static field no longer triggers class initialization (**CSR**)

### Reviewers
 * [Mandy Chung](https://openjdk.org/census#mchung) (@mlchung - **Reviewer**) ⚠️ Review applies to [3c5d380b](https://git.openjdk.org/jdk/pull/13821/files/3c5d380b6b518db010bc26e0a1eaf9097a7154a7)
 * [Paul Sandoz](https://openjdk.org/census#psandoz) (@PaulSandoz - **Reviewer**) ⚠️ Review applies to [3c5d380b](https://git.openjdk.org/jdk/pull/13821/files/3c5d380b6b518db010bc26e0a1eaf9097a7154a7)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13821/head:pull/13821` \
`$ git checkout pull/13821`

`$ git checkout pull/13821` \
`$ git pull https://git.openjdk.org/jdk.git pull/13821/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13821`

`$ git pr show -t 13821`

</details>

<a href="https://git.openjdk.org/jdk/pull/13821.diff">https://git.openjdk.org/jdk/pull/13821.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13821#issuecomment-1535693915)


src/java.base/share/classes/java/lang/invoke/IndirectVarHandle.java
```diff
@@ -61,33 +60,31 @@ private IndirectVarHandle(VarHandle target, Class<?> value, Class<?>[] coordinat
...
+    MethodType accessModeTypeUncached(AccessType at) {
+        return at.accessModeType(null, value, coordinates);
     }
 
     @Override
-    public List<Class<?>> coordinateTypes() {
-        return List.of(coordinates);
-    }
+    public boolean isAccessModeSupported(AccessMode accessMode) {
+        var directTarget = this.directTarget;
```
mcimadamore, line 74: Why is this not defined in the superclass, and then use `asDirect` (as done in other cases) ?

src/java.base/share/classes/java/lang/invoke/VarHandles.java
```diff
@@ -106,66 +107,66 @@ else if (type == double.class) {
...
         else {
-            // TODO This is not lazy on first invocation
-            // and might cause some circular initialization issues
-
-            // Replace with something similar to direct method handles
-            // where a barrier is used then elided after use
-
-            if (UNSAFE.shouldBeInitialized(refc))
-                UNSAFE.ensureClassInitialized(refc);
+            if (UNSAFE.shouldBeInitialized(refc)) {
```
mcimadamore, line 110: This seems unrelated to the issue this PR is really about (if I understand correctly). Would it make sense to address this as part of another PR?

mcimadamore: Also, perhaps @PaulSandoz knows more history on this one (I believe this code predates FFM).

src/java.base/share/classes/java/lang/invoke/IndirectVarHandle.java
liach: I aim to avoid eager evaluation of directTarget (which is lazy in LazyStaticVarHandle, computing it requires initializing the holder class). This code path checks a stable field, so it should allow constant-folding when the direct target is available.

mcimadamore: I see - so, since this is just a "query" you would like to avoid initialization just for this.

liach:
@mlchung Would you mind taking a look at this?

mlchung:
I'm aware of these PRs.  I'll get to it when I get to it.

src/java.base/share/classes/java/lang/invoke/VarHandles.java
```diff
@@ -106,66 +107,66 @@ else if (type == double.class) {
...
+                return new LazyStaticVarHandle(refc, f, isWriteAllowedOnFinalFields, false);
```
ExE-Boss, line 111: This should probably call `maybeAdapt` on the result:
```suggestion
                return maybeAdapt(new LazyStaticVarHandle(refc, f, isWriteAllowedOnFinalFields, false));
```

mlchung:
I studied the patch and it seems a clever idea.   I will do another pass to check if `VarHandle::asDirect` is only invoked when the var handle is doing field access.  

As for the implementation of `LazyStaticVarHandle`, I expect that `makeFieldHandle` can always create the direct var handle (i.e. `FieldStaticReadXXX` var handle) and then wrapped by `LazyStaticVarHandle` similar to how `IndirectVarHandle` does; something like this:

```
    LazyStaticVarHandle(VarHandle target, Class<?> refc, boolean exact) {
        super(target.vform, exact);
        this.target = target;
        this.refc = refc;
    }
```

Some overridden methods for example `describeConstable` can simply retrun `target.describeConstable()`.   Do I miss some issue that it can't be implemented that way?

src/java.base/share/classes/java/lang/invoke/LazyStaticVarHandle.java
```diff
@@ -0,0 +1,185 @@
...
+        return delegate == null ? (hasInvokeExactBehavior() ? this
+                : new LazyStaticVarHandle(refc, field, writeAllowedOnFinalFields, false))
+                : delegate.withInvokeExactBehavior();
+    }
+
+    @Override
+    public VarHandle withInvokeBehavior() {
+        var delegate = this.delegate;
+        return delegate == null ? (!hasInvokeExactBehavior() ? this
+                : new LazyStaticVarHandle(refc, field, writeAllowedOnFinalFields, true))
```
mlchung, line 134:
```suggestion
                : new LazyStaticVarHandle(refc, field, writeAllowedOnFinalFields, false))
```

src/java.base/share/classes/java/lang/invoke/LazyStaticVarHandle.java
```diff
@@ -0,0 +1,185 @@
...
+            long offset = MethodHandleNatives.staticFieldOffset(field);
+            return this.delegate = VarHandles.makeInitializedStaticFieldVarHandle(field, refc, base, offset, writeAllowedOnFinalFields);
+        }
+    }
+
+    @Override
+    public VarHandle withInvokeExactBehavior() {
+        var delegate = this.delegate;
+        return delegate == null ? (hasInvokeExactBehavior() ? this
+                : new LazyStaticVarHandle(refc, field, writeAllowedOnFinalFields, false))
```
mlchung, line 126:
```suggestion
                : new LazyStaticVarHandle(refc, field, writeAllowedOnFinalFields, true))
```

mlchung (commented):
I think the exact value for `withInvokeExactBehavior` and `withInvokeBehavior` is swapped.

liach:
> As for the implementation of `LazyStaticVarHandle`, I expect that `makeFieldHandle` can always create the direct var handle (i.e. `FieldStaticReadXXX` var handle) and then wrapped by `LazyStaticVarHandle` similar to how `IndirectVarHandle` does; 

Since I am not familiar with hotspot internals, I assumed that staticFieldBase and staticFieldOffset shouldn't be used until the class is initialized (rather than linked). This for certain can be done if they are ready once a class is linked, and we should probably update unsafe's comment as well. Can you confirm that they staticFieldBase/Offset are safe to be called before initialization?

If this change is safe, we can significantly reduce the initial-call overhead of the lazy VH significantly as well.

mlchung:
`staticFieldBase` and `staticFieldOffset` can be called before initialization when a class has been linked.   They are unsafe operations and so the code has to ensure class initialization before it accesses a static field.

liach:
In addition, should we initialize the ref class or the declaring class only? I don't think we need to initialize the ref class when declaring class initialization should be sufficient, but that is a change in behavior.

mlchung:
> In addition, should we initialize the ref class or the declaring class only? I don't think we need to initialize the ref class when declaring class initialization should be sufficient, but that is a change in behavior.

Declaring class will be initialized [1].

refc should be equals to the declaring class because the static field to be looked up can only be found in the declaring class.  There should be no behavioral change.

[1] https://download.java.net/java/early_access/jdk21/docs/api/java.base/java/lang/invoke/MethodHandles.Lookup.html#findStaticVarHandle(java.lang.Class,java.lang.String,java.lang.Class)

liach:
Not quite. I mean scenarios like this:
```java
import java.lang.invoke.*;

class Parent {
	public static final int[] value = {34};
}

class Child extends Parent {}

public class Test {
	public static void main(String... args) throws Throwable {
		System.out.println(Child.value[0]);
		var lookup = MethodHandles.lookup();
		var getter = lookup.findStaticVarHandle(Child.class, "value", int[].class);
		System.out.println(((int[]) getter.getAcquire())[0]);
	}
}
```

Here, the `value` field is in parent, but it's accessed via a subclass and is resolved successfully. Are the refc and declaring class still equal here? I think we only need to initialize Parent without initializing Child in this case.

PaulSandoz (commented):
The latest revision looks a much better arrangement.

I wonder if it is possible to transition from an indirect VH to a direct VH once initialized i.e., `checkAccessModeThenIsDirect` returns the value of `initialized`. Thereby we reduce the potential footprint cost of inflating further MHs and take the more efficient route through the guard method on subsequent invocations.

Not thought through all the implications of this, but since this will all occur in the interpreter it may well work out. We then get closer, conceptually at least, to the behavior of static field access using direct MHs.

mlchung:
> I think we only need to initialize Parent without initializing Child in this case.
 
The bytecode behavior to access the field via `lookup.findStaticVarHandle(Child.class, "value", int[].class)` is `Child.value;`.  The class or interface that declared the resolved field is initialized per JVMS `getStatic` and `putStatic`.   So I think `Parent` is initialized but not `Child`.    @PaulSandoz can confirm.

PaulSandoz:
> > I think we only need to initialize Parent without initializing Child in this case.
> 
> The bytecode behavior to access the field via `lookup.findStaticVarHandle(Child.class, "value", int[].class)` is `Child.value;`. The class or interface that declared the resolved field is initialized per JVMS `getStatic` and `putStatic`. So I think `Parent` is initialized but not `Child`. @PaulSandoz can confirm.

Yes, confirmed (verified both with explicit field access, `Child.value`,  and access using a direct method handle).

mlchung:
> I wonder if it is possible to transition from an lazy (indirect) VH to a direct VH once initialized i.e., `checkAccessModeThenIsDirect` returns the value of `initialized`. 

Sounds a good idea.  Once the class is initialized, the lazy VH becomes a direct VH that may help the performance.  In that case, the VarHandleGuards would need to ensure it links with the direct VH not the lazy VH.

src/java.base/share/classes/java/lang/invoke/LazyInitializingVarHandle.java
```diff
@@ -0,0 +1,116 @@
...
+    @ForceInline
+    boolean checkAccessModeThenIsDirect(AccessDescriptor ad) {
+        super.checkAccessModeThenIsDirect(ad);
+        return false;
+    }
+
+    @Override
+    public VarHandle withInvokeExactBehavior() {
+        var updatedDelegate = target.withInvokeExactBehavior();
+        return initialized ? new LazyInitializingVarHandle(updatedDelegate, refc) : updatedDelegate;
```
mlchung, line 75:
```suggestion
        return initialized ? updatedDelegate: new LazyInitializingVarHandle(updatedDelegate, refc);
```

If initialized, should return `updatedDelegate` instead.   Same for `withInvokeBehavior`

test/jdk/java/lang/invoke/VarHandles/LazyStaticTest.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class LazyStaticTest {
+    static Set<Class<?>> initialized = ConcurrentHashMap.newKeySet();
+
+    public static void main(String... args) throws Throwable {
+        assert initialized.isEmpty() : "Incorrect initial state";
+
+        class Sample1 {
```
mlchung, line 42: Nit: suggest to move this inner class outside of the method body.  I think it helps readability.   maybe name it as `Sample` or `Test` or sth.   I expect `Sample2` when I read `Sample1`.

test/jdk/java/lang/invoke/VarHandles/LazyStaticTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+        var lookup = MethodHandles.lookup();
+        var sample1AppleVh = lookup.findStaticVarHandle(Sample1.class, "apple", int.class);
+
+        assert !initialized.contains(Sample1.class) : "Sample1 class initialized on VarHandle creation";
+
+        sample1AppleVh.set(42);
+
+        assert initialized.contains(Sample1.class) : "Sample1 class initialized after VarHandle use";
+        assert Sample1.apple == 42 : "The value is not set correctly to Sample1.apple";
```
mlchung, line 59: It may be good to extend the test case to invoke the lazy VH multiple times that will exercise in the case if lazy VH transitions to direct VH.

liach:
> I wonder if it is possible to transition from an lazy (indirect) VH to a direct VH once initialized i.e., `checkAccessModeThenIsDirect` returns the value of `initialized`. Thereby we reduce the potential footprint cost of inflating further MHs and take the more efficient route through the guard method on subsequent invocations.

Currently, the VH passed to linkToStatic-linked implementation methods must be of the correct implementation type. As a result, we have to at least add an `asDirect` call when we pass the VH instance, like the invocation for indirect VH's MHs. Would that be an acceptable refactor, or is that too costly as after constant folding, the two VHs already have similar performance?

mlchung:
what performance results do you see with this change?

liach:
> what performance results do you see with this change?

The indirect VH invocation already has a performance similar to the direct invocation in JMH. Do you mean that I should add `asDirect` to the direct paths in VH and check the benchmark results?

mlchung:
> I've also updated the benchmark for the latest patch. Unfortunately the regular invocations and lazy var handle initial call performance saw a decrease.

Maybe I misunderstood what you meant about "a decrease in performance".

> The indirect VH invocation already has a performance similar to the direct invocation in JMH. Do you mean that I should add asDirect to the direct paths in VH and check the benchmark results?

I guess there would be no difference then.

liach:
Yeah, I was referring the performance drop from 0.5ns/op to 12ns/op for botg direct and indirect; don't know what caused the problems of general slowdown on my end.

Also, is there anything I can do to fix the windows build failures?

src/java.base/share/classes/java/lang/invoke/LazyInitializingVarHandle.java
```diff
@@ -0,0 +1,117 @@
...
+
+    @Override
+    public Optional<VarHandleDesc> describeConstable() {
+        return target.describeConstable();
+    }
+
+    private void ensureInitialized() {
+        UNSAFE.ensureClassInitialized(refc);
+        this.initialized = true;
+    }
```
ExE-Boss, line 106: This should probably have a fast‑path when `this.initialized` is `true`, so that it can be better constant folded by the **JIT**:
```suggestion
    private void ensureInitialized() {
        if (this.initialized) {
            return;
        }
        UNSAFE.ensureClassInitialized(refc);
        this.initialized = true;
    }
```

liach: I think I should probably update the MH lambda form as well, to skip the ensure call entirely if JVM has not constant-fold the inefficient form yet.

liach: Indeed, updating lambda form and cached mh works: the average invocation performance becomes ~0.8ns/op, equal to the direct version, compared to the original ~1ms/op. See the updated benchmark results in the main PR body.

src/java.base/share/classes/java/lang/invoke/LazyInitializingVarHandle.java
```diff
@@ -0,0 +1,137 @@
...
+                var callTarget = target.getMethodHandle(i);
+                mh.updateForm(new Function<>() {
+                    @Override
+                    public LambdaForm apply(LambdaForm lambdaForm) {
+                        return callTarget.form;
+                    }
+                });
+                cache[i] = callTarget;
+            }
+        }
```
ExE-Boss, line 125: Unlike calling `Unsafe::ensureClassInitialized(…)` (which performs its own synchronisation) or writing `true` to `this.initialized` (which is a benign race), updating the lambda forms this way is probably not quite correct to do under a race and probably needs some synchronisation.

liach: The form update is a benign race. The array update might need some fences, but I'm not sure, as the read MH should be complete when the updated value is already observed.

PaulSandoz:
> > I wonder if it is possible to transition from an lazy (indirect) VH to a direct VH once initialized i.e., `checkAccessModeThenIsDirect` returns the value of `initialized`.
> 
> Sounds a good idea. Once the class is initialized, the lazy VH becomes a direct VH that may help the performance. In that case, the VarHandleGuards would need to ensure it links with the direct VH not the lazy VH.

I believe instead of changing the guards we can change the implementation of the static methods on the concrete static VHs to call `asDirect()` (lets also add `@ForceInline` on `asDirect`), since `linkToStatic` will pass the args to the static method described by the member name (since the lazy and direct VH handles share the same var form).

That has the nice effect of pushing the complexity closer to where it is needed. Its subtle but i think can be explained. Thereby we can reduce the memory footprint for subsequent access modes. Perhaps we can even "zero" out the MH cache on initialization?

liach:
> I believe instead of changing the guards we can change the implementation of the static methods on the concrete static VHs to call `asDirect()` (lets also add `@ForceInline` on `asDirect`), since `linkToStatic` will pass the args to the static method described by the member name (since the lazy and direct VH handles share the same var form).
> 
> That has the nice effect of pushing the complexity closer to where it is needed. Its subtle but i think can be explained. Thereby we can reduce the memory footprint for subsequent access modes. Perhaps we can even "zero" out the MH cache on initialization?

Is this aimed at decreasing long-term invocation cost or the initial invocation cost? FYI long-term invocation is already efficient enough; if this is for initial invocation, which can potentially bypass costly MH construction, I recommend creating an alternative method to `asDirect` for the purpose to be called and casted by X-VarHandle implementations.

PaulSandoz:
> Is this aimed at decreasing long-term invocation cost or the initial invocation cost? FYI long-term invocation is already efficient enough; if this is for initial invocation, which can potentially bypass costly MH construction, I recommend creating an alternative method to `asDirect` for the purpose to be called and casted by X-VarHandle implementations.

It aims to reduce the cost of getting to peak performance (the work that C2 has to perform) and reduce the overall memory footprint of such VHs (as long as the user does not explicitly request MHs from them). Assuming of course that the peak performance is good!

For this pattern i not sure we need to create another method returning the target that effectively does the same as `asDirect`, and instead it may come down to better naming and documentation of the concepts.

There is already a tight coupling between the `LazyInitializingVarHandle` and its target via the sharing of the var forms. Renaming `asDirect` to `target` might do it, and we refine a what a direct handle is to mean the guards just pass along the handle that is given to it, otherwise we use a (lazily created) method handle to its target. Thereby a VH implementation free to switch between indirect and direct and is further free to deal with any implementation specific indirection in its own implementation.

liach:
@PaulSandoz Thank you for your suggestion! It might be harder to measure the time to get to peak performance, but at least for the initial invocation, the lazy VH is almost no different from the eager VH now. The old overhead was mostly from LambdaForm spinning, same for the lazy DMH overhead. Is this what you've envisioned?

PaulSandoz:
> @PaulSandoz Thank you for your suggestion! It might be harder to measure the time to get to peak performance, but at least for the initial invocation, the lazy VH is almost no different from the eager VH now. The old overhead was mostly from LambdaForm spinning, same for the lazy DMH overhead. 

Ok, i did not mean to suggest measuring time to peak performance, it's hard as you say. I was more thinking of an intuitive understanding of the cost (work performed by C2 and runtime memory footprint).

> Is this what you've envisioned?

Close, I was suggesting the following:
- `LazyInitializingVarHandle::target` just returns `target` and does not call `ensureInitialized`.
- Override `checkAccessModeThenIsDirect` as you did before, but returns `initialized`.
- Change only the static field var handle implementations to call `target()`, since it is unnecessary on the others.

However, i think your approach is better as it avoids any implicit MH creation. Suggest refining as follows:
- Change only the static field var handle implementations to call `target()`, since it is unnecessary on the others.
- Split `ensureInitialized` in two:
```
    @ForceInline
    private void ensureInitialized() {
        if (this.initialized)
            return;
        
        initialize();
    }

    private void initialize() {
        UNSAFE.ensureClassInitialized(refc);
        this.initialized = true;
        ...
    }
```

I am wondering if the simplest approach in `initialize` is to set the `methodHandleTable` to null?

liach:
> Change only the static field var handle implementations to call `target()`, since it is unnecessary on the others.

Unfortunately, I don't think there is a convenient way for VH implementations to easily unwrap the correct VH; a `target()` call is more convenient than an extra instanceof check and cast and get.

> I am wondering if the simplest approach in `initialize` is to set the `methodHandleTable` to null?

I update the entries in the table because they are already created and potentially live on the heap or stack. Those entries should benefit from the post-initialization performance benefits as well.

PaulSandoz:
> > Change only the static field var handle implementations to call `target()`, since it is unnecessary on the others.
> 
> Unfortunately, I don't think there is a convenient way for VH implementations to easily unwrap the correct VH; a `target()` call is more convenient than an extra instanceof check and cast and get.
> 

We don't need to update *all* VH implementations to call target. We only need to update the static field VH implementations which we know may be passed an instance of `LazyInitializingVarHandle` or its concrete instance.


> > I am wondering if the simplest approach in `initialize` is to set the `methodHandleTable` to null?
> 
> I update the entries in the table because they are already created and potentially live on the heap or stack. Those entries should benefit from the post-initialization performance benefits as well.

Ok, yes, it's less likely now but possible a MH has been explicitly obtained before the defining class has been intialized. In this case I still think we can clear the table out after processing any existing entries, since subsequent requests for MHs after initialization will defer to the target's method handle table.

We will need some tests covering this edge case.

liach:
> Ok, yes, it's less likely now but possible a MH has been explicitly obtained before the defining class has been intialized. In this case I still think we can clear the table out after processing any existing entries, since subsequent requests for MHs after initialization will defer to the target's method handle table.

I think I can just replace the lazy VH's MH table with that of the target, as whatever this table's cache has the target table does as well (due to our getMethodHandleUncached delegation). This is effectively GCing the old table as well.

PaulSandoz:
Something was bothering me about the current complexity with method handles and the potential interactions with indirect var handles. Its hard to reason about all the interactions so i needed to take a deeper dive, which i really should have done earlier on. Apologies if we are circling around this many times.

Since we are now leaning on `LazyInitializingVarHandle::target` to initialize I think we can just do this:
```
    private void initialize() {
        UNSAFE.ensureClassInitialized(refc);
        this.initialized = true;
    }

    @Override
    public MethodHandle toMethodHandle(AccessMode accessMode) {
        if (initialized) {
            return target.toMethodHandle(accessMode);
        } else {
            if (isAccessModeSupported(accessMode)) {
                MethodHandle mh = target.getMethodHandle(accessMode.ordinal());
                // Ensure target on this VH is called to initialize the class
                return mh.bindTo(this);
            }
            else {
                // Ensure an UnsupportedOperationException is thrown
                return MethodHandles.varHandleInvoker(accessMode, accessModeType(accessMode)).
                        bindTo(this);
            }
        }
    }
```

However, construction of an indirect VH will result in initialization:
```
    private IndirectVarHandle(VarHandle target, Class<?> value, Class<?>[] coordinates,
                      BiFunction<AccessMode, MethodHandle, MethodHandle> handleFactory, VarForm form, boolean exact) {
        super(form, exact);
        this.handleFactory = handleFactory;
        this.target = target;
        this.directTarget = target.target();
        this.value = value;
        this.coordinates = coordinates;
    }
```

Since this is not performance sensitive code we could check if target is an instance of `LazyInitializingVarHandle` then conditionally get it's target if initialized e.g.,
```
   this.target = target instanceof LazyInitializingVarHandle lazyTarget ? lazyTarget.lazyTarget() : target.target();
```
?

liach:
/issue remove 8307508

Will fix the indirect var handle NPE in a separate patch as recommended by Mandy.

liach:
@PaulSandoz I have thought about VarHandle a bit more. Why does getMethodHandle accept the direct target of a VarHandle when getMethodHandle itself is already specific to each VarHandle instance? I thought maybe the method handle cache can be elevated to var form so it requires a known receiver VH to be passed, but the MH is not erased, so it doesn't belong to the var form. If we can remove the VarHandle parameter to getMethodHandle, we can simplify the implementations of getMethodHandleUncached and toMethodHandle significantly.

If such a getMethodHandle refactor is desirable, I can accomplish it in a separate patch and make this patch dependent on it.

liach:
@PaulSandoz I think this patch as of now is what I want:
1. Lazy VH works both when obtained directly or delegated by an Indirect VH
2. In the most common usage scenario (direct calls on VH), the lazy VH has a good initial performance and a similar long-term performance compared to the eager VH
3. The invasive changes to existing structures are minimized.

For the performance concerns of `getMethodHandle` and `toMethodHandle` (mainly creation and initial invocation), I find that the VH invocation semantics itself has some problems: for example, it's quite pointless to pass a VH-specific direct VH to a VH-specific `getMethodHandle`. My focus on this patch will be to ensure the correctness of MH-related content, but caring about their performance is out of scope for me.

PaulSandoz:
> @PaulSandoz I have thought about VarHandle a bit more. Why does getMethodHandle accept the direct target of a VarHandle when getMethodHandle itself is already specific to each VarHandle instance?

IIRC its to avoid creating an extra bound MH when used with the guards.

src/java.base/share/classes/java/lang/invoke/VarHandle.java
```diff
@@ -2186,6 +2206,14 @@ final MethodHandle getMethodHandle(int mode) {
         return mh;
     }
 
+    /**
+     * Computes a method handle that can be passed the {@linkplain #target() direct}
```
PaulSandoz, line 2210:
```suggestion
     * Computes a method handle that can be passed the {@linkplain #asDirect() direct}
```
?

test/jdk/java/lang/invoke/VarHandles/LazyInitializingTest.java
```diff
@@ -0,0 +1,204 @@
...
+        vh.isAccessModeSupported(VarHandle.AccessMode.GET_AND_BITWISE_XOR_ACQUIRE);
+        vh.withInvokeExactBehavior();
+        vh.withInvokeBehavior();
+        vh.toMethodHandle(VarHandle.AccessMode.GET_AND_BITWISE_XOR_ACQUIRE);
+        vh.hasInvokeExactBehavior();
+        vh.accessModeType(VarHandle.AccessMode.GET_AND_BITWISE_XOR_ACQUIRE);
+    }
+
+    @Test
+    public void testInitializationOnVarHandleUse() {
```
PaulSandoz, line 85: I think we will need to expand this to also test operating on a MH to a VH, an indirect VH, and on a MH to indirect VH. We could do this separately.

test/micro/org/openjdk/bench/java/lang/invoke/LazyStaticColdStart.java
```diff
@@ -0,0 +1,155 @@
...
+                    cob.constantInstruction(100L);
+                    cob.invokestatic(CD_Blackhole, "consumeCPU", MTD_void_long);
+                    cob.invokestatic(CD_ThreadLocalRandom, "current", MTD_ThreadLocalRandom);
+                    cob.invokevirtual(CD_ThreadLocalRandom, "nextLong", MTD_long);
+                    cob.putstatic(describedClass, "v", CD_long);
+                    cob.return_();
+                });
+            });
+
+            static {
```
PaulSandoz, line 92: Is this static block required?

liach: I use this block to remove the overhead around MethodHandle and VarHandle infrastructure initialization (such as LF spinning) that is not related to any particular class; so that I measure only the impact of encountering a new class.

test/jdk/java/lang/invoke/VarHandles/LazyInitializingTest.java
liach: The MH on VH case is worth considering; the indirect VH cases are already covered by `-Djava.lang.invoke.VarHandle.VAR_HANDLE_IDENTITY_ADAPT=true`.

liach: Done; added a test case to test initialization on toMethodHandle use in latest push, passed locally.

mlchung (commented):
Good work!   I try to see how to improve the documentation to help readers to understand.  See if the suggestion helps.

test/micro/org/openjdk/bench/java/lang/invoke/LazyStaticColdStart.java
PaulSandoz: Ah, implied by "... and initializes the lazy/non-lazy handles to prevent further creation costs" Could you also add a small comment above this block?

test/jdk/java/lang/invoke/VarHandles/LazyInitializingTest.java
PaulSandoz: Ok. I just realized you could also run using the system property `java.lang.invoke.VarHandle.VAR_HANDLE_GUARDS` to disables the guards, and it will go through the MH invokers, achieving the same thing. Up to you if you want to keep the current test or use the system property.

PaulSandoz (approved):

test/jdk/java/lang/invoke/VarHandles/LazyInitializingTest.java
```diff
@@ -0,0 +1,214 @@
...
+ * questions.
+ */
+
+/*
+ * @test
+ * @bug 8291065
+ * @summary Checks interaction of static field VarHandle with class
+ *          initialization mechanism..
+ * @run junit LazyInitializingTest
+ * @run junit/othervm -Djava.lang.invoke.VarHandle.VAR_HANDLE_IDENTITY_ADAPT=true LazyInitializingTest
```
mlchung, line 30: I wonder if this new test should run with all combination of 
`java.lang.invoke.VarHandle.VAR_HANDLE_GUARDS` and `java.lang.invoke.VarHandle.VAR_HANDLE_IDENTITY_ADAPT` as done in other `VarHandleTestXXX`.

mlchung (approved):
Looks good in general.  Some minor comments.

I suggest to create a CSR to document the behavioral change to match the specification.

test/jdk/java/lang/invoke/VarHandles/LazyInitializingTest.java
liach: Done, now the test runs with all 4 combinations and passes on my end.

test/micro/org/openjdk/bench/java/lang/invoke/LazyStaticColdStart.java
liach: Done. Also commented on each line about what each call initializes.

liach:
> I suggest to create a CSR to document the behavioral change to match the specification.

CSR filed for this behavior change; no spec updates but there's minor class-loading compatibility risk. I would be grateful if you can review the CSR.

liach:
The CSR has been approved and I have drafted a release note at https://bugs.openjdk.org/browse/JDK-8311878 following the instructions on the openjdk guide. @psandoz @mlchung Can you please review this release note since you've reviewed this patch?

liach:
Can I integrate this patch now? Seems release notes are usually delivered after patch integration.

mlchung:
yes, go ahead.

psandoz:
Hi guys. I’m Peyton sandoz or psandoz on GitHub. I think you must have
meant to mention someone else, but just wanted to let you know!

liach:
Sorry for my incorrect ping and bothering.

psandoz:
Don’t worry about it, it happens. Y’all have a good day

PR #13824, opened 2023-05-05
8307517: Add VMErrorCallback infrastructure to extend hs_err dumping

stefank:
Sometimes when we crash in the GC we'd like to get some more information about what was going on the crashing thread. One example is when Generational ZGC crashes during store barrier flushing. From https://github.com/openjdk/zgc/blob/349cf9ae38664991879402a90c5e23e291f1c1c3/src/hotspot/share/gc/z/zStoreBarrierBuffer.cpp#L245

```
class ZStoreBarrierBuffer::OnError : public VMErrorCallback {
private:
  ZStoreBarrierBuffer* _buffer;

public:
  OnError(ZStoreBarrierBuffer* buffer) :
      _buffer(buffer) {}

  virtual void call(outputStream* st) {
    _buffer->on_error(st);
  }
};

void ZStoreBarrierBuffer::on_error(outputStream* st) {
  st->print_cr("ZStoreBarrierBuffer: error when flushing");
  st->print_cr(" _last_processed_color: " PTR_FORMAT, _last_processed_color);
  st->print_cr(" _last_installed_color: " PTR_FORMAT, _last_installed_color);

  for (int i = current(); i < (int)_buffer_length; ++i) {
    st->print_cr(" [%2d]: base: " PTR_FORMAT " p: " PTR_FORMAT " prev: " PTR_FORMAT,
        i,
        untype(_base_pointers[i]),
        p2i(_buffer[i]._p),
        untype(_buffer[i]._prev));
  }
}

void ZStoreBarrierBuffer::flush() {
  if (!ZBufferStoreBarriers) {
    return;
  }

  OnError on_error(this);
  VMErrorCallbackMark mark(&on_error);

  for (int i = current(); i < (int)_buffer_length; ++i) {
    const ZStoreBarrierEntry& entry = _buffer[i];
    const zaddress addr = ZBarrier::make_load_good(entry._prev);
    ZBarrier::mark_and_remember(entry._p, addr);
  }

  clear();
}
```

If we crash in ZStoreBarrierBuffer::flush, we print the information above into the hs_err file.

We've found this information to be useful and would like to upstream the infrastructure separately from the much larger Generational ZGC PR.

Testing: this has been brewing and been used in the Generational ZGC repository for a long time.

### DIFF
--- a/src/hotspot/share/utilities/vmError.hpp
+++ b/src/hotspot/share/utilities/vmError.hpp
@@ -213,4 +213,27 @@ public:
   static int prepare_log_file(const char* pattern, const char* default_pattern, bool overwrite_existing, char* buf, size_t buflen);
 
 };
+
+class VMErrorCallback {
+  friend class VMError;
+  friend class VMErrorCallbackMark;
+
+  // Link through all callbacks active on a thread
+  VMErrorCallback* _next;
+
+  // Called by VMError reporting
+  virtual void call(outputStream* st) = 0;
+
+public:
+  VMErrorCallback() : _next(nullptr) {}
+};
+
+class VMErrorCallbackMark : public StackObj {
+  Thread* _thread;
+
+public:
+  VMErrorCallbackMark(VMErrorCallback* callback);
+  ~VMErrorCallbackMark();
+};
+
 #endif // SHARE_UTILITIES_VMERROR_HPP
--- a/src/hotspot/share/runtime/thread.cpp
+++ b/src/hotspot/share/runtime/thread.cpp
@@ -98,6 +98,7 @@ Thread::Thread() {
   _jvmti_env_iteration_count = 0;
   set_allocated_bytes(0);
   _current_pending_raw_monitor = nullptr;
+  _vm_error_callbacks = nullptr;
 
   // thread-specific hashCode stream generator state - Marsaglia shift-xor form
   _hashStateX = os::random();
--- a/src/hotspot/share/runtime/thread.hpp
+++ b/src/hotspot/share/runtime/thread.hpp
@@ -55,6 +55,7 @@ class SafeThreadsListPtr;
 class ThreadClosure;
 class ThreadsList;
 class ThreadsSMRSupport;
+class VMErrorCallback;
 
 class OopClosure;
 class CodeBlobClosure;
@@ -104,6 +105,8 @@ class JavaThread;
 //       - this->entry_point()  // set differently for each kind of JavaThread
 
 class Thread: public ThreadShadow {
+  friend class VMError;
+  friend class VMErrorCallbackMark;
   friend class VMStructs;
   friend class JVMCIVMStructs;
  private:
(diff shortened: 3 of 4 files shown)


xmas92 (approved):
lgtm.

I have also experimented with using this functionality to only record the last unloading events for a specific unloading cycle. (And have the callback print the events if a crash occurred during the cycle). As I found that in release builds the majority of the time is spent inside `Events::log_class_unloading` for do_unloading.

dholmes-ora (commented):
This looks quite neat but I'm not clear on the need for the VMErrorCallbackMark - can't the callback link/unlink itself at construction/destruction?

src/hotspot/share/utilities/vmError.hpp
```diff
@@ -213,4 +213,27 @@ class VMError : public AllStatic {
   static int prepare_log_file(const char* pattern, const char* default_pattern, bool overwrite_existing, char* buf, size_t buflen);
 
 };
+
```
tstuefe, line 216: pre-existing, can you please add a prototype decl for outputStream?

src/hotspot/share/utilities/vmError.hpp
```diff
@@ -213,4 +213,27 @@ class VMError : public AllStatic {
...
+  // Called by VMError reporting
+  virtual void call(outputStream* st) = 0;
+
+public:
+  VMErrorCallback() : _next(nullptr) {}
+};
+
+class VMErrorCallbackMark : public StackObj {
+  Thread* _thread;
```
tstuefe, line 232: Why would we need the thread here? Why not use Thread::current in dtor? This object is only used as stack object, right?

tstuefe (approved):
Nice, I like it.

tstuefe:
> This looks quite neat but I'm not clear on the need for the VMErrorCallbackMark - can't the callback link/unlink itself at construction/destruction?

I like it better this way. Otherwise you dictate that the callback obj itself has to live on the stack. It may be large, or it may be shared between different threads.

src/hotspot/share/utilities/vmError.hpp
stefank: I was treading in Runtime code and Coleen usually wants to use cached-away Thread pointers instead of calling Thread::current() repeatedly. I'm fine with either solution.

tstuefe:
> 
> I can go either way, so it would be good if the reviewers could chime in with their preference. This is what it would look like:
> 
> https://github.com/openjdk/jdk/compare/master...stefank:jdk:8307517_VMErrorCallback_2

I prefer the explicit RAII object, separate from the callback.

src/hotspot/share/utilities/vmError.hpp
dholmes-ora: Given the context it would have to be `Thread::current_or_null_safe()`. But yes we prefer not to re-materialize the current thread if we already have it at hand.

stefank: Could you explain why it would have to be `Thread::current_or_null_safe()`? The constructor and destructor are run in "normal" JVM code and not in the error handler.

dholmes-ora: Sorry my mistake.

stefank:
Thanks for reviewing! In the interest of getting this pushed before the Generational ZGC, I'm going to integrate it now. FWIW, I'm not opposed to doing some follow-up style changes if we decide that this should be further tweaked.
/integrate

PR #8195, opened 2022-04-12
8284732: FFI_GO_CLOSURES macro not defined but required for zero build on Mac OS X

parttimenerd:
It just codifies the current behavior as it defines `FFI_GO_CLOSURES` to be `0`.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed

### Issue
 * [JDK-8284732](https://bugs.openjdk.java.net/browse/JDK-8284732): FFI_GO_CLOSURES macro not defined but required for zero build on Mac OS X

### Reviewers
 * [David Holmes](https://openjdk.java.net/census#dholmes) (@dholmes-ora - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk pull/8195/head:pull/8195` \
`$ git checkout pull/8195`

Update a local copy of the PR: \
`$ git checkout pull/8195` \
`$ git pull https://git.openjdk.java.net/jdk pull/8195/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 8195`

View PR using the GUI difftool: \
`$ git pr show -t 8195`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.java.net/jdk/pull/8195.diff">https://git.openjdk.java.net/jdk/pull/8195.diff</a>

</details>

PR #13827, opened 2023-05-05
8307527: MacOS Zero builds fail with undefined FFI_GO_CLOSURES after JDK-8304265

shipilev:
See the bug. Actually, I am not sure why JDK-8304265 changed the `#ifndef FFI_GO_CLOSURES` to `#ifdef _APPLE_`. That seems too intrusive if `FFI_GO_CLOSURES` *is* enabled. So I rewrote the block to something more safe.

Additional testing:
 - [x] macos-aarch64-zero-fastdebug `make images` passes

### DIFF
--- a/src/hotspot/cpu/zero/globalDefinitions_zero.hpp
+++ b/src/hotspot/cpu/zero/globalDefinitions_zero.hpp
@@ -32,7 +32,7 @@
 
 #define SUPPORT_MONITOR_COUNT
 
-#ifdef __APPLE__
+#if defined(__APPLE__) && !defined(FFI_GO_CLOSURES)
 #define FFI_GO_CLOSURES 0
 #endif
 
--- a/src/java.base/share/native/libfallbackLinker/fallbackLinker.c
+++ b/src/java.base/share/native/libfallbackLinker/fallbackLinker.c
@@ -25,6 +25,12 @@
 
 #include "jdk_internal_foreign_abi_fallback_LibFallback.h"
 
+// This is similar to globalDefinitions_zero.hpp, which is not accessible
+// from JDK libraries. Copy-paste it here.
+#if defined(__APPLE__) && !defined(FFI_GO_CLOSURES)
+#define FFI_GO_CLOSURES 0
+#endif
+
 #include <ffi.h>
 
 #include <errno.h>


JornVernee:
Hi Aleksey. The original change came in from: https://github.com/openjdk/panama-foreign/pull/770 with the motivation:

> Finally, I had to slightly modify globalDefinitions_zero.hpp to conditionally define FFI_GO_CLOSURES, since I was getting a redefinition error using GCC. The macro is also defined in ffitarget.h which is included by ffi.h. The define in globalDefinitions comes from this PR: https://github.com/openjdk/jdk/pull/8195 which indicates that the define is only needed on Mac Os X. So I switched out the guard to check for __APPLE__ instead. (the check whether it is already defined doesn't really do anything, since FFI_GO_CLOSURES is defined by including ffi.h).

I kept the check, but I'm still not sure how this define is supposed to work, since as far as I can tell `ffitarget.h` will define `FFI_GO_CLOSURES` (which is included by `ffi.h`). I guess the `#ifndef` is needed in case it is defined on the command line?

I'll run this patch through our CI as well.

JornVernee:
After thinking a bit, I think I'd prefer this to be addressed in the build system instead. e.g. something like:

```
diff --git a/make/autoconf/lib-ffi.m4 b/make/autoconf/lib-ffi.m4
index 0905c3cd225..83de5a4abf7 100644
--- a/make/autoconf/lib-ffi.m4
+++ b/make/autoconf/lib-ffi.m4
@@ -106,6 +106,13 @@ AC_DEFUN_ONCE([LIB_SETUP_LIBFFI],
       AC_MSG_ERROR([Could not find libffi! $HELP_MSG])
     fi

+    if test "x${OPENJDK_TARGET_CPU}" = "xaarch64" && test "x${OPENJDK_TARGET_OS}" = xmacosx; then
+      # ffi.h checks '#if FFI_GO_CLOSURES' which throws a warning in xcode on aarch64 because the aarch64
+      # ffitarget.h (included from ffi.h) doesn't explicitly define FFI_GO_CLOSURES (like it does on e.g. x64).
+      # define it explicitly here to avoid compilation errors
+      LIBFFI_CFLAGS="$LIBFFI_CFLAGS -DFFI_GO_CLOSURES=0"
+    fi
+
     AC_MSG_CHECKING([if libffi works])
     AC_LANG_PUSH(C)
     OLD_CFLAGS="$CFLAGS"
```

And then remove the workaround from the source code. (`LIBFFI_CFLAGS` is used to build both relevant libraries, and should also be used when a new library is added that needs libffi. So this would avoid a repeat of this issue)

Either way, let's thoroughly document the issue this time around, so future editors won't make the same mistake I did.

JornVernee:
Thanks! This looks like a good solution, but I'll leave the Review to one of the build people.

erikj79 (approved):

JornVernee:
@shipilev Ran this through our CI again, but it failing there. It seems like the check in the configuration is erroneously failing (`checking for FFI_GO_CLOSURES definition... no, defining`) and then the compilation of hotspot fails because `FFI_GO_CLOSURES` is already defined, when `ffitarget.h` tries to define it again.

JornVernee:
@shipilev While looking at something else, I noticed that the fallback linker was not handling variadic arguments correctly (and some of our tests were linking against the wrong descriptors). On macos/aarch64 with zero/fallback linker this could cause crashes in some of the java/foreign tests that use varidadic arguments (since the ABI is different for variadic args).

I have a patch here to fix this: https://github.com/openjdk/jdk/compare/master...JornVernee:jdk:VAFixes

That helps, but there are still some failing tests:

```
java/foreign/TestUpcallStack.java                                   Failed. Unexpected exit from test [exit code: 134]
java/foreign/arraystructs/TestArrayStructs.java#interpreted         Failed. Execution failed: `main' threw exception: java.lang.Exception: failures: 14
java/foreign/arraystructs/TestArrayStructs.java#specialized         Failed. Execution failed: `main' threw exception: java.lang.Exception: failures: 14
```

I think this might be an issue with libffi, but I'm not sure (I'm using version 3.4.2). Note that all of these failures are cases where arguments are being passed on the stack.

Also, FYI, I'll be on vacation the next to weeks, so I will likely not be available to reply here.

JornVernee:
Coming back to this: It looks like your fix works if I put it right after the `AC_LINK_IFELSE` use in `lib-ffi.m4`, but before the `CFLAGS` and `LIBS` are reset.

I have been looking at this script as well to get it working on mac and windows: https://github.com/openjdk/jdk/compare/master...JornVernee:jdk:LibFFI_Mac_Win (I have it working on both those platforms)

If you're not able to test, I could also take your fix and create a PR for the patch I have with your fix included (and add you as a contributor). WDYT?

shipilev:
> If you're not able to test, I could also take your fix and create a PR for the patch I have with your fix included (and add you as a contributor). WDYT?

Yes, please! Open your PR, I'll close my as dup.

JornVernee:
> > If you're not able to test, I could also take your fix and create a PR for the patch I have with your fix included (and add you as a contributor). WDYT?
> 
> Yes, please! Open your PR, I'll close my as dup.

Okay, I've gone ahead and made a PR here: https://github.com/openjdk/jdk/pull/14446

shipilev:
Closing in favor of #14446.

PR #13830, opened 2023-05-05
8280993: [XWayland] Popup is not closed on click outside of area controlled by XWayland

azvegint:
On Linux systems, we rely on XGrabPointer (X11 API) to capture mouse input and dismiss popup menus on mouse clicks outside the popup menu.
Unfortunately, on Linux systems using the [Wayland](https://wayland.freedesktop.org/) session this only works inside [XWayland(Wayland's X11 server implementation)](https://wayland.freedesktop.org/xserver.html).
This means if a user clicks on a part of the screen not controlled by XWayland (e.g. window decorations, other non X11 applications) the popup menu will not be hidden.

As a workaround, we can hide this menu when the parent popup menu window loses focus.
However, it does have its drawbacks, which should be described in the documentation.The focus does not change when clicking on the header of its own parent window or on non-focusable windows, .e.g., empty space in system dock, so in this case the popup menu is not hidden.

Third-party applications use a similar approach.

I also have doubts about the need to change the documentation, as I can't find where it is described that the popup menu should be hidden when clicked outside the menu.

<del>CSR: https://bugs.openjdk.org/browse/JDK-8307529</del>

---------
### Progress

### Issues
 * [JDK-8280993](https://bugs.openjdk.org/browse/JDK-8280993): [XWayland] Popup is not closed on click outside of area controlled by XWayland
 * [JDK-8307529](https://bugs.openjdk.org/browse/JDK-8307529): [XWayland] Popup is not closed on click outside of area controlled by XWayland (**CSR**) (Withdrawn)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13830/head:pull/13830` \
`$ git checkout pull/13830`

`$ git checkout pull/13830` \
`$ git pull https://git.openjdk.org/jdk.git pull/13830/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13830`

`$ git pr show -t 13830`

</details>

<a href="https://git.openjdk.org/jdk/pull/13830.diff">https://git.openjdk.org/jdk/pull/13830.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13830#issuecomment-1536196100)

### DIFF
--- a/src/java.desktop/share/classes/java/awt/PopupMenu.java
+++ b/src/java.desktop/share/classes/java/awt/PopupMenu.java
@@ -43,6 +43,10 @@ import sun.awt.AWTAccessor;
  * (e.g., you add it to a {@code MenuBar}), then you <b>cannot</b>
  * call {@code show} on that {@code PopupMenu}.
  *
+ * @implNote On Linux systems using Wayland, the popup menu may not be dismissed
+ * by clicking on the decorations of its own parent window
+ * and on some system panels.
+ *
  * @author      Amy Fowler
  */
 public class PopupMenu extends Menu {
--- a/src/java.desktop/share/classes/java/awt/PopupMenu.java
+++ b/src/java.desktop/share/classes/java/awt/PopupMenu.java
@@ -43,10 +43,6 @@ import sun.awt.AWTAccessor;
  * (e.g., you add it to a {@code MenuBar}), then you <b>cannot</b>
  * call {@code show} on that {@code PopupMenu}.
  *
- * @implNote On Linux systems using Wayland, the popup menu may not be dismissed
- * by clicking on the decorations of its own parent window
- * and on some system panels.
- *
  * @author      Amy Fowler
  */
 public class PopupMenu extends Menu {
--- a/src/java.desktop/share/classes/javax/swing/JPopupMenu.java
+++ b/src/java.desktop/share/classes/javax/swing/JPopupMenu.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1997, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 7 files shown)


mkartashev:
> I also have doubts about the need to change the documentation, as I can't find where it is described that the popup menu should be hidden when clicked outside the menu.

FWIW, I agree that you only need to change the documentation if someone can actually point to where it says that popup hides when there's a mouse click outside its bounds. Otherwise the implementation notes seem to do more harm (by confusing the reader) than good.

src/java.desktop/share/classes/java/awt/PopupMenu.java
```diff
@@ -43,6 +43,10 @@
  * (e.g., you add it to a {@code MenuBar}), then you <b>cannot</b>
  * call {@code show} on that {@code PopupMenu}.
  *
+ * @implNote On Linux systems using Wayland, the popup menu may not be dismissed
```
mrserb, line 46: I guess the XWayland is affected, but implementation of the toolkit for Wayland will not have this issue?

azvegint: It should, and this is another point about not changing the documentation.

mrserb: Then let's not update it.

src/java.desktop/share/classes/javax/swing/JPopupMenu.java
```diff
@@ -758,6 +764,62 @@ public void pack() {
         }
     }
 
+    // We rely on the X11 input grab mechanism, but for the Wayland session
```
prrace, line 767: Whilst we need to do this, having all this code discussing Wayland here in JPopupMenu is really the wrong place.
This needs to be re-done in a way that is about the specific problem which is something like
SunToolkit.clickOutsideWillDismissPopup() 
and the word "Wayland" shouldn't appear anywhere here.

IIUC this fix is needed so that a number of tests that fail now because of this problem will be able to pass, so it would be good to get it in soon so I can get a clearer picture of what remaining test issues we have.

azvegint: Sure, moved to toolkit.

> IIUC this fix is needed so that a number of tests that fail now because of this problem will be able to pass, so it would be good to get it in soon so I can get a clearer picture of what remaining test issues we have.

This only affects manual tests, as automatic tests use the XTest API, which works correctly with X11 Input Grab.

Besides, this is a real use case that can be annoying. This fix doesn't fix the problem completely (we still can't hide the menu by clicking on the title bar of our own window or on non-focusable windows such as the desktop), but it definitely makes the user experience better.

src/java.desktop/share/classes/javax/swing/JPopupMenu.java
```diff
@@ -758,6 +759,16 @@ public void pack() {
         }
     }
 
+    private Window getMenuInvoker() {
+        if (invoker instanceof Window menuInvoker) {
```
azvegint, line 763: [Pattern matching for instanceof](https://openjdk.org/jeps/394) is used.
Maybe we shouldn't use this if we plan to backport this change to the JDK < 16

prrace: I wouldn't worry about that.

prrace (approved):

PR #13831, opened 2023-05-05
JDK-8305958: Use links instead of buttons for auto-generated header links

hns:
Please review a change to use plain HTML links instead of buttons to provide the URL to page section/heading anchors. Special care was taken to retain the accessibility features added in JDK-8299718, so section links can be reached and activated by keyboard and have an `aria-label` attribute. The generated documentation was tested on a range of desktop and mobile browsers (Firefox, Chrome on Mac OS and Linux, Safari on Mac OS and iOS, Chrome on Android) and can be viewed here:

https://cr.openjdk.org/~hannesw/8305958/api.00/index.html

---------
### Progress

### Issue
 * [JDK-8305958](https://bugs.openjdk.org/browse/JDK-8305958): Use links instead of buttons for auto-generated header links

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13831/head:pull/13831` \
`$ git checkout pull/13831`

`$ git checkout pull/13831` \
`$ git pull https://git.openjdk.org/jdk.git pull/13831/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13831`

`$ git pr show -t 13831`

</details>

<a href="https://git.openjdk.org/jdk/pull/13831.diff">https://git.openjdk.org/jdk/pull/13831.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13831#issuecomment-1536291351)

### DIFF
--- a/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/resources/search.js.template
+++ b/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/resources/search.js.template
@@ -414,7 +414,7 @@ $(function() {
         if (id) {
             hdr.append(" <a href='#" + id + "' class='anchor-link' aria-label='" + messages.linkToSection
                 + "'><img src='" + pathtoroot + "link.svg' alt='" + messages.linkIcon +"' tabindex='0'"
-                + " style='width: 0.9em; height: 0.9em; vertical-align: baseline;'></a>");
+                + " width='16' height='16'></a>");
         }
     });
     $(window).on("orientationchange", collapse).on("resize", function(e) {
--- a/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/resources/search.js.template
+++ b/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/resources/search.js.template
@@ -408,7 +408,7 @@ $(function() {
     $("input#search-input").focus(collapse);
     $("main").click(collapse);
     $("section[id] > :header, :header[id], :header:has(a[id])").each(function(idx, el) {
-        // Create anchor liniks for headers with an associated id attribute
+        // Create anchor links for headers with an associated id attribute
         var hdr = $(el);
         var id = hdr.attr("id") || hdr.parent("section").attr("id") || hdr.children("a").attr("id");
         if (id) {
--- a/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/resources/search.js.template
+++ b/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/resources/search.js.template
@@ -31,9 +31,8 @@ const messages = {
     loading: "##REPLACE:doclet.search.loading##",
     searching: "##REPLACE:doclet.search.searching##",
     redirecting: "##REPLACE:doclet.search.redirecting##",
-    copyToClipboard: "##REPLACE:doclet.Copy_to_clipboard##",
-    copyUrlToClipboard: "##REPLACE:doclet.Copy_url_to_clipboard##",
-    copiedToClipboard: "##REPLACE:doclet.Copied_to_clipboard##"
+    linkIcon: "##REPLACE:doclet.Link_icon##",
+    linkToSection: "##REPLACE:doclet.Link_to_section##"
 }
 const categories = {
     modules: "##REPLACE:doclet.search.modules##",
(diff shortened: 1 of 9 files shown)


liach:
This link icon does look much more informative compared to the existing copy icon.

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/resources/search.js.template
```diff
@@ -409,14 +408,13 @@ $(function() {
     $("input#search-input").focus(collapse);
     $("main").click(collapse);
     $("section[id] > :header, :header[id], :header:has(a[id])").each(function(idx, el) {
-        // Create copy-to-clipboard buttons for headers with an associated id attribute
+        // Create anchor liniks for headers with an associated id attribute
```
jonathan-gibbons, line 411: typo: `linik`

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/resources/search.js.template
```diff
@@ -409,14 +408,13 @@ $(function() {
...
         var hdr = $(el);
         var id = hdr.attr("id") || hdr.parent("section").attr("id") || hdr.children("a").attr("id");
         if (id) {
-            hdr.append($("<button class='copy copy-header' onclick='copyUrl(this)' aria-label='"
-                + messages.copyUrlToClipboard +"'> " + "<img src='" + pathtoroot + "copy.svg' alt='" +
-                messages.copyUrlToClipboard + "'> " + "<span data-copied='" + messages.copiedToClipboard +
-                "'>" + messages.copyToClipboard + "</span></button>"));
+            hdr.append(" <a href='#" + id + "' class='anchor-link' aria-label='" + messages.linkToSection
+                + "'><img src='" + pathtoroot + "link.svg' alt='" + messages.linkIcon +"' tabindex='0'"
+                + " style='width: 0.9em; height: 0.9em; vertical-align: baseline;'></a>");
```
jonathan-gibbons, line 417: Should this be in a CSS class, not an inline `style` ?

jonathan-gibbons (commented):
* a typo
* suggest a CSS class, not `style` attribute for the link
* I like the new icon

jonathan-gibbons:
Playing with the posted demo docs, I have the following comments, not necessarily criticism

* I'm mildly surprised the links show up on some of the "standard" "internal" headings, like "Constructor Details". I guess there may be cases where this is wanted, and maybe if you didn't do it, we would get a RFE to do it.

* the change accentuates the difference between "real" headings and "heading-like labels" (like **See Also:**) which is just a bold label.   But given my surprise about the use of links in my previous bullet, I'm not necessarily suggesting we should change this.

hns:
> 
>     * I'm mildly surprised the links show up on some of the "standard" "internal" headings, like "Constructor Details". I guess there may be cases where this is wanted, and maybe if you didn't do it, we would get a RFE to do it.
> 

This was already the case with the previous copy-to-clipboard button. All HTML headings with an `id` attribute as well as sections with an `id` attribute and a heading as first child element are given a link. This not only includes standard page sections, but also individual member details. I think this is quite useful when looking at a member detail without having gone through the link in the member summary section.

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/resources/search.js.template
hns: I used inline style to make sure the styles are always available, as the default size of the SVG image is so large that it blows up the layout of the page. My initial choice was to use the `height` and `width` attributes in the HTML `img` element, but after some research came to the conclusion that that was not the proper use of these attributes. (I'm not 100% sure this is the correct conclusion but that's another problem.)

This is how I ended up with the inline styles. It could probably be moved to the stylesheet, at the risk of blowing up the layout when the page is loaded without stylesheet for some reason (such as network problems).

hns: I think I found a good solution for this. I moved the CSS rules from inline to the stylesheet and added `width` and `height` attributes to the `img` element as fallback / safety net. Both HTML attributes and CSS rules result in the same computed values, but the CSS rules override the HTML attributes. The `vertical-align: baseline` was not needed as that is the default value for that property.

jonathan-gibbons (approved):

PR #13833, opened 2023-05-05
8302670: use-after-free related to PhaseIterGVN interaction with Unique_Node_List and Node_Stack

eme64:
**Motivation**

- Generally: we copy containers by value: the consequence is that we copy all internals by value, including size, capacity and pointers. From there on, the containers can diverge, and make each other inconsistent. One may be destructed and free the memory of the other one. In theory this could cause a bug on the main-branch. In practice, we probably (maybe?) use the correct one of the many copies that is currently supposed to be alive. If one pushes to the wrong copy, then one will most likely eventually hit a SIGSEGV - which has happened to me and @TobiHartmann a few times - it is very annoying. Plus: copy by value of containers is very bad design, and makes it difficult to understand which one is the "live" copy.
- We also overwrite igvn phases. One case is particularly hairy: `igvn = ccp` (truncate ccp, and store it into igvn variable. Aka `object slicing in c++`)

@jcking 's first version https://github.com/openjdk/jdk/pull/12703. He dropped it as a "discussion or starting point for somebody else". I took a lot of ideas from him, but went a bit more aggressive with refactoring instead of the `replace_with` move-like approach.

**Changes**

- Make many containers `NONCOPYABLE`:
  - `Dict`
  - `VectorSet`
  - `Node_Array`, `Node_List`, `Unique_Node_List`
  - `Node_Stack`
  - `NodeHash`
  - `Type_Array`
  - `Phase`
  - Note: for many classes I still allow the `A(A&&) = default;` constructor. This allows implicit moves (rvalues) so that we can return containers from functions and capture them.
- Create "global" containers for `Compile`:
  - `C->igvn_worklist()` (renamed from `for_igvn`, referenced to by `PhaseIterGVN._worklist`)
  - `C->type_array()` (referenced to by `PhaseValues._types`)
  - `C->node_hash_table()` (referenced to by `PhaseValues._table`)
  - They are created in the `Compile` constructor. The phases can then hold a reference (`&`) to them.
  - Note: before, these were located in the phases, and passed back and forth by value. They were passed downward via the phase constructor, where the corresponding fields were taken over from the previous phase. Then they were passed upward by `PhaseGVN.replace_with` (for `_table` and `_types`), or by simply overwriting the old `igvn` variable with a newly constructed igvn that has the containers passed into its constructor from the previous phase. I imagine it as "weaving" the containers from phase to phase, where the ownership travels. The messy part was that others still held a outdated value copy of it.
  - `_table` was passed around from phase to phase, and stored by value in `PhaseValue._table`. It was then `clear()`'ed via the destructors. But since there was only really ever one "owner", the "live" container was only cleared once the last phase was over (it was passed via `replace-with` or the `NodeHash(NodeHash *use_this_state)` constructor). Now that we create the `_table` via `new`, and never actively deconstruct it, we need to `clear()` it explicitly after the last igvn goes out of scope just before `final_graph_reshaping`.
  - I would have liked these containers to be allocated inside the `Compile` object directly (as values). But that would have lead to cpp-header-file cyclic dependencies between `compile.hpp` and `node.hpp`. So I had to take pointers.
- Moved things from `PhaseTransform` to `PhaseValues`:
  - `_types` (now only by reference) and all type related functions
  - `ConNode caches`: related fields and functions (needed to move it because I moved `_types`)
  - `saturate / saturate_and_maybe_push_to_igvn_worklist`
  - They thematically make more sense there, the class is called `PhaseValues` after all and the comments suggest it was meant for this stuff. And other subclasses of `PhaseTransform` do not use these things anyway. For example `PhaseIdealLoop` always accesses them via the `igvn` that it is passed.
  - Had to change lots of interfaces from `PhaseTransform*` to `PhaseValues*` because of value/type functionality only available in `PhaseValues`. I considered moving them directly to `PhaseGVN`. That would have worked, but maybe eventually we want to refactor the CCP/IGVN/GVN phases and it is better to have `PhaseValues` as the superclass of all of them.
- To make sure that the phase containers were copied around, there are a few cases where we used to overwrite a phase by value. Since we should disable copy by value of phases, I had to find a solution. The solution that @jcking proposed was to destruct/re-construct. So we now use `reset_from_gvn` and `reset_from_igvn`. This does the same as copy by value, but explicitly. We may want to refactor this in the future, maybe there is a better way.
- `PhaseGVN.replace_with` made sure that the containers were passed back from igvn to gvn. I was able to remove it since the containers are now at `Compile`, and do not have to be passed any more.
- Refactoring around `PhaseRenumberLive`:
  - New worklists were generated and overwrote the `for_igvn/igvn_worklist`, this looked extremely nasty and used copy by value extensively. Now it is much simpler.
  - `Unique_Node_List.recompute_idx_set`: after re-numbering, the old worklist has the `VectorSet` invalid (the bits are set for the old idx, but should be set for the new idx now). Instead of creating a new worklist, we can just recompute the `VectorSet`.
  - The only thing I am not 100% happy with: `gvn->types().swap(_new_type_array);`. We need to re-order the `_types`, because the types need to be at the index of the new idx. I implemented a safe `swap` method for that. Still, it means we lose some memory in the `comp_arena()`. An alternative would be to have one in a local array, and copy it back. Open to suggestions.
- `PhaseTransform._nodes` was used by different phases in various and non-consistent ways. I moved it into the subclasses, and renamed it according to what it is actually used for:
  - `PhaseIdealLoop._loop_ctrl`
  - `Matcher._new_nodes`
  - `node_map` in `haseCCP::do_transform`
  - I removed some old dump functions, which did not explicitly state for what usecase they were: `PhaseTransform::dump_old2new_map, PhaseTransform::dump_new, PhaseTransform::dump_types, PhaseTransform::dump_nodes_and_types`. Most of the functionality can easily be done through other ways. Let me know if the removal is problematic. I would have to move them to the phase that you wish it should work for.
- Made `_stack` local to `PhaseIterGVN::remove_globally_dead_node`.
- At many places we check if `igvn_worklist()` is empty and clear it, just for good measure. I packaged this into `igvn_worklist().ensure_empty()`.

**Future Work**
- `igvn.reset...` - can we remove it? The destruct/re-construct could possibly be replaced by calling the proper init-functions to reset the old igvn.
- Refactor Phases:
  - `transform` functions have a very confusing and inconsistent naming, implementation and usage. Many have asserts that ensure they are either never used, or only used the right way. A better design could make it more readable and would allow removal of the asserts, as they would become trivial.
  - `Value -> GVN -> IGVN -> CCP` nesting does not really make sense. We should probably have `CCP` next to `GVN`. Maybe `IGVN` should also be next to `GVN`, or a subclass.
  - `init_con_caches` is this really something local, or could it live "globally" at `Compile`?
  - `Phase._pnum (PhaseNumber)`: do we really need this? Is there not a better solution?
  - `PhaseValues._iterGVN` used as flag for `PhaseValues.is_IterGVN()`. I guess this is to avoid having a virtual function. But is that really worth it?
  - Make it clearer which methods are `virtual` (some are tagged `virtual`, but are never overridden), and which ones `override` others.
  - Generally, the comments/descriptions of the `Phase` classes could be better (or removed) - they seem to be out of date.
  - Consider moving `PhaseIdealLoop::_loop_or_ctrl` to its own arena, so that we can use `ResouceMarks` and not have conflicts with it.
- This change here also gets us a step closer to modular Phases, which can be easily enabled/disabled and reordered.

I filed: [JDK-8307815](https://bugs.openjdk.org/browse/JDK-8307815) C2 Phase structure cleanup
(and linked this PR in it)

**Testing**

Passes up to tier5 and stress testing. Just for good measure I checked it with and without `-XX:VerifyIterativeGVN=10`.
**TODO**: performance testing

**Discussion**

This is a bit of a tricky refactoring, it took me quite some time, and I am still not 100% sure about it. I am open to suggestions.

---------
### Progress

### Issue
 * [JDK-8302670](https://bugs.openjdk.org/browse/JDK-8302670): use-after-free related to PhaseIterGVN interaction with Unique_Node_List and Node_Stack

### Reviewers
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - **Reviewer**) ⚠️ Review applies to [28946dc1](https://git.openjdk.org/jdk/pull/13833/files/28946dc178f1d149d30c8954e43978fda7ee6023)
 * [Christian Hagedorn](https://openjdk.org/census#chagedorn) (@chhagedorn - **Reviewer**) ⚠️ Review applies to [dc2d49a9](https://git.openjdk.org/jdk/pull/13833/files/dc2d49a93667f48ec7ab82312446136e389f46a9)
 * [Justin King](https://openjdk.org/census#jcking) (@jcking - Committer) ⚠️ Review applies to [dc2d49a9](https://git.openjdk.org/jdk/pull/13833/files/dc2d49a93667f48ec7ab82312446136e389f46a9)

### Contributors
 * Justin King `<jcking@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13833/head:pull/13833` \
`$ git checkout pull/13833`

`$ git checkout pull/13833` \
`$ git pull https://git.openjdk.org/jdk.git pull/13833/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13833`

`$ git pr show -t 13833`

</details>

<a href="https://git.openjdk.org/jdk/pull/13833.diff">https://git.openjdk.org/jdk/pull/13833.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13833#issuecomment-1541440125)


src/hotspot/share/opto/compile.hpp
```diff
@@ -948,7 +969,7 @@ class Compile : public Phase {
 
   void              identify_useful_nodes(Unique_Node_List &useful);
   void              update_dead_node_list(Unique_Node_List &useful);
-  void              disconnect_useless_nodes(Unique_Node_List &useful, Unique_Node_List* worklist);
+  void              disconnect_useless_nodes(Unique_Node_List &useful, Unique_Node_List &worklist);
```
TobiHartmann, line 972:
```suggestion
  void              disconnect_useless_nodes(Unique_Node_List& useful, Unique_Node_List& worklist);
```

src/hotspot/share/opto/lcm.cpp
```diff
@@ -1269,7 +1269,7 @@ Node* PhaseCFG::catch_cleanup_find_cloned_def(Block *use_blk, Node *def, Block *
   if( j == def_blk->_num_succs ) {
     // Block at same level in dom-tree is not a successor.  It needs a
     // PhiNode, the PhiNode uses from the def and IT's uses need fixup.
-    Node_Array inputs = new Node_List();
+    Node_Array inputs;
```
TobiHartmann, line 1272: Urgh, looks like we had object slicing here before your changes.

src/hotspot/share/opto/matcher.hpp
```diff
@@ -88,6 +88,8 @@ class Matcher : public PhaseTransform {
   // Private arena of State objects
   ResourceArea _states_arena;
 
+  Node_List   _new_nodes;
```
TobiHartmann, line 91: Please add a comment.

src/hotspot/share/libadt/vectset.hpp
```diff
@@ -53,7 +53,9 @@ class VectorSet : public AnyObj {
 public:
   VectorSet();
   VectorSet(Arena* arena);
-  ~VectorSet() {}
+
+  // Allow move constructor for && (eg. capture return of function)
```
TobiHartmann, line 58: It's not completely clear yet to me why this is required and how it correlates with `NONCOPYABLE` but I leave this to the experts :)

src/hotspot/share/opto/node.hpp
```diff
@@ -1627,10 +1641,35 @@ class Unique_Node_List : public Node_List {
...
+  // If the idx of the Nodes change, we must recompute the VectorSet
+  void recompute_idx_set() {
+    _in_worklist.clear();
+    for (uint i = 0; i < size(); i++) {
+      Node* n = at(i);
+      _in_worklist.set(n->_idx);
+    }
+  }
+
+#ifndef PRODUCT
```
TobiHartmann, line 1661:
```suggestion
#ifdef ASSERT
```

We don't need this in the optimized build.

src/hotspot/share/opto/phaseX.cpp
```diff
@@ -477,15 +421,12 @@ PhaseRenumberLive::PhaseRenumberLive(PhaseGVN* gvn,
     C->grow_node_notes(C->node_note_array(), new_size);
   }
 
+  assert(worklist.is_subset_of(_useful), "sanity");
```
TobiHartmann, line 432: Please add a more useful error message.

src/hotspot/share/opto/node.hpp
```diff
@@ -1627,10 +1641,35 @@ class Unique_Node_List : public Node_List {
...
+  bool is_subset_of(Unique_Node_List &other) {
```
TobiHartmann, line 1662:
```suggestion
  bool is_subset_of(Unique_Node_List& other) {
```

src/hotspot/share/opto/phaseX.cpp
```diff
@@ -383,30 +352,10 @@ Node *NodeHash::find_index(uint idx) { // For debugging
...
-
-
-#endif
-
-
 //=============================================================================
 //------------------------------PhaseRemoveUseless-----------------------------
 // 1) Use a breadthfirst walk to collect useful nodes reachable from root.
-PhaseRemoveUseless::PhaseRemoveUseless(PhaseGVN* gvn, Unique_Node_List* worklist, PhaseNumber phase_num) : Phase(phase_num) {
+PhaseRemoveUseless::PhaseRemoveUseless(PhaseGVN* gvn, Unique_Node_List &worklist, PhaseNumber phase_num) : Phase(phase_num) {
```
TobiHartmann, line 366:
```suggestion
PhaseRemoveUseless::PhaseRemoveUseless(PhaseGVN* gvn, Unique_Node_List& worklist, PhaseNumber phase_num) : Phase(phase_num) {
```

src/hotspot/share/opto/phaseX.cpp
```diff
@@ -440,21 +389,17 @@ PhaseRemoveUseless::PhaseRemoveUseless(PhaseGVN* gvn, Unique_Node_List* worklist
...
+//     information is returned in PhaseGVN::_types.
 //
 // Other data structures used by the compiler are not updated. The hash table for value
 // numbering (the field PhaseGVN::_table) is not updated because computing the hash
-// values is not based on node IDs. The field PhaseGVN::_nodes is not updated either
-// because it is empty wherever PhaseRenumberLive is used.
+// values is not based on node IDs.
 PhaseRenumberLive::PhaseRenumberLive(PhaseGVN* gvn,
-                                     Unique_Node_List* worklist, Unique_Node_List* new_worklist,
+                                     Unique_Node_List &worklist,
```
TobiHartmann, line 410:
```suggestion
                                     Unique_Node_List& worklist,
```

src/hotspot/share/opto/phaseX.cpp
```diff
@@ -499,16 +440,14 @@ PhaseRenumberLive::PhaseRenumberLive(PhaseGVN* gvn,
...
-      new_worklist->push(n);
-    }
-
     if (update_embedded_ids(n) < 0) {
       _delayed.push(n); // has embedded IDs; handle later
     }
   }
 
-  assert(worklist_size == new_worklist->size(), "the new worklist must have the same size as the original worklist");
+  // VectorSet in Unique_Node_Set must be recomputed, since ID's have changed.
```
TobiHartmann, line 456:
```suggestion
  // VectorSet in Unique_Node_Set must be recomputed, since IDs have changed.
```

src/hotspot/share/opto/phaseX.hpp
```diff
@@ -25,6 +25,7 @@
 #ifndef SHARE_OPTO_PHASEX_HPP
 #define SHARE_OPTO_PHASEX_HPP
 
+#include "utilities/globalDefinitions.hpp"
```
TobiHartmann, line 28: Should be added in alphabetical order.

src/hotspot/share/opto/phaseX.hpp
```diff
@@ -471,14 +458,36 @@ class PhaseIterGVN : public PhaseGVN {
   // improvement, such that it would take many (>>10) steps to reach 2**32.
 
 public:
+
   PhaseIterGVN(PhaseIterGVN* igvn); // Used by CCP constructor
   PhaseIterGVN(PhaseGVN* gvn); // Used after Parser
 
+  // Reset IGVN from GVN: call deconstructor, and placement new.
+  // Acheives the same as the following (but without move constructors):
```
TobiHartmann, line 468:
```suggestion
  // Achieves the same as the following (but without move constructors):
```

src/hotspot/share/opto/phaseX.hpp
```diff
@@ -471,14 +458,36 @@ class PhaseIterGVN : public PhaseGVN {
...
+  // igvn = PhaseIterGVN(gvn);
+  void reset_from_gvn(PhaseGVN* gvn) {
+    if (this != gvn) {
+      this->~PhaseIterGVN();
+      ::new (static_cast<void*>(this)) PhaseIterGVN(gvn);
+    }
+  }
+
+  // Reset IGVN with another: call deconstructor, and placement new.
+  // Acheives the same as the following (but without move constructors):
```
TobiHartmann, line 478:
```suggestion
  // Achieves the same as the following (but without move constructors):
```

TobiHartmann (approved):
Great work, Emanuel! I only have a few minor comments.

> Phase._pnum (PhaseNumber): do we really need this? Is there not a better solution?

Since you removed the least two usages outside of the `Phase` constructor, let's file a follow-up RFE to investigate if we can simply remove it.

src/hotspot/share/opto/compile.hpp
```diff
@@ -415,7 +417,16 @@ class Compile : public Phase {
...
   // Parsing, optimization
   PhaseGVN*             _initial_gvn;           // Results of parse-time PhaseGVN
-  Unique_Node_List*     _for_igvn;              // Initial work-list for next round of Iterative GVN
+
+  // Shared worklist for all IGVN rounds. Nodes can be pushed to it at any time.
+  // If pushed outside IGVN, the Node is processed in the next IGVN round.
+  Unique_Node_List*     _igvn_worklist;
+
+  // Shared type array for GVN, IGVN and CCP. It maps node ID -> Type*.
+  Type_Array*           _type_array;
```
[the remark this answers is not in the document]
eme64: will change to `_types`

src/hotspot/share/opto/compile.hpp
```diff
@@ -415,7 +417,16 @@ class Compile : public Phase {
...
+
+  // Shared node hash table for GVN, IGVN and CCP.
+  NodeHash*             _node_hash_table;
```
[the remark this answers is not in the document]
eme64: will change to `_node_hash`

src/hotspot/share/opto/compile.hpp
```diff
@@ -935,11 +946,21 @@ class Compile : public Phase {
...
-  Unique_Node_List* for_igvn()                  { return _for_igvn; }
+  Unique_Node_List& igvn_worklist() {
+    assert(_igvn_worklist != nullptr, "must be created in Compile::Compile");
+    return *_igvn_worklist;
+  }
+  Type_Array& type_array() {
+    assert(_type_array != nullptr, "must be created in Compile::Compile");
+    return *_type_array;
+  }
+  NodeHash& node_hash_table() {
```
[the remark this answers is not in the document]
eme64: Will move change from pointer to reference to be in Phase constructor

src/hotspot/share/opto/lcm.cpp
eme64: Yup, there were a few nasty things happening :sweat_smile:

src/hotspot/share/libadt/vectset.hpp
eme64: I took this from @jcking . From what I understand:
`NONCOPYABLE` disables the copy constructor (`&`) and move operator. Somehow, this also disables the move constructor (`&&`). Re-enabling that one allows things like returning local containers, and capturing them via that move constructor.
```
Unique_Node_List some_function() {
  Unique_Node_List local_worklist;
  // do stuff
  return local_worklist;
}

void other_function() {
  Unique_Node_List capture_worklist = some_function();
  // capture_worklist has its scope widened to this function
}
```
But if someone has a more detailed explanation, I'm glad to hear it ;)

chhagedorn (commented):
Nice work! That's a great cleanup and makes things much easier to follow.

I only have some comments.

> Value -> GVN -> IGVN -> CCP nesting does not really make sense. We should probably have CCP next to GVN. Maybe IGVN should also be next to GVN, or a subclass.

That was always confusing. I like the idea of having them as separate subclasses of Value.

Maybe we could also investigate in a future RFE to place `_loop_ctrl` in a separate arena as well. I've run into some problems when using a `ResourceMark` for a local `Node_List`. I've called `get_ctrl()` for some unmapped node which initiated the `_loop_ctrl` list to be grown/reallocated. When the local `ResourceMark` went out of scope, it also reverted the newly allocated space for `_loop_ctrl` which was hard to foresee.

This would also allow us to add some more `ResourceMark`s throughout the code (could also be part of this future RFE).

eme64:
Thanks @TobiHartmann for the suggestions and review!

Thanks @chhagedorn for the suggestions and ideas!

@justin would you mind reviewing this too, since quite a bit of material is from your original PR?

/contributor add @jcking

chhagedorn (approved):
Thanks for doing the updates, looks good!

src/hotspot/share/libadt/dict.hpp
```diff
@@ -61,6 +61,9 @@ class Dict : public AnyObj { // Dictionary structure
   Dict(const Dict &base, Arena* arena); // Deep-copy
   ~Dict();
 
+  // Allow move constructor for && (eg. capture return of function)
+  Dict(Dict&&) = default;
```
jcking, line 65: Nit: You might consider invalidating the other dict being moved from, to catch accidental use-after-move. Could be punted to a future change.

src/hotspot/share/libadt/vectset.hpp
```diff
@@ -55,6 +55,9 @@ class VectorSet : public AnyObj {
   VectorSet(Arena* arena);
   ~VectorSet() {}
 
+  // Allow move constructor for && (eg. capture return of function)
+  VectorSet(VectorSet&&) = default;
```
jcking, line 59: Same as the other, consider invalidating the moved from `VectorSet` by setting the data to nullptr or something similar to catch misbehaving code.

jcking (approved):

src/hotspot/share/libadt/dict.hpp
kimbarrett: The only way to get use-after-move is with an explicit `std::move` or equivalent.

But shouldn't there also be a move-assign operator?  I think at some point the standard mandates
some amount of the Rule of Five (previously the Rule of Three), and some violations are already
deprecated in C++14 (and gcc warns about them).

OTOH, I think the rationale of needing a move constructor to permit returning noncopyable objects from functions is eliminated by C++17's guaranteed copy elision.  So it might be that these move constructors are only needed until we upgrade the language standard we use.  (I hope that will be soon-ish, but it's not imminent.)

src/hotspot/share/libadt/vectset.hpp
kimbarrett: pedantic: s/explicitly default it or define it/explicitly define it (possibly with a default definition)/.  But yes, that's
why the explicit definition is needed if you want to permit move  of noncopyable (e.g. move-only).

kimbarrett (commented):
Not really a review, just some drive-by comments.

src/hotspot/share/libadt/dict.hpp
kimbarrett: The main reason to do some cleanup of the moved-from object is if it owns resources that are being transferred
by the move.  You don't want the destruction of the moved-from object trashing stuff in the moved-to object.
All of the classes having move-constructors added by this PR have default or empty destructors.

eme64:
@kimbarrett @jcking I wonder if it is not better to just avoid any move constructors/assign. We would have to convert the "return value optimization" cases (i.e. return contrainer by value, and it being captured by the move constructor), and instead create containers outside, and pass them in as references ("pseudo-output"). It is a bit ugly, but maybe more understandable than the move semantics?

eme64:
This is the patch, there was only a hand full of cases. Should I apply this patch?
```
diff --git a/src/hotspot/share/libadt/dict.hpp b/src/hotspot/share/libadt/dict.hpp
index c021536c402..02c634ff1d1 100644
--- a/src/hotspot/share/libadt/dict.hpp
+++ b/src/hotspot/share/libadt/dict.hpp
@@ -61,9 +61,6 @@ class Dict : public AnyObj { // Dictionary structure
   Dict(const Dict &base, Arena* arena); // Deep-copy
   ~Dict();
 
-  // Allow move constructor for && (eg. capture return of function)
-  Dict(Dict&&) = default;
-
   // Return # of key-value pairs in dict
   uint32_t Size(void) const { return _cnt; }
 
diff --git a/src/hotspot/share/libadt/vectset.hpp b/src/hotspot/share/libadt/vectset.hpp
index a82046f2ba9..5a58afc276b 100644
--- a/src/hotspot/share/libadt/vectset.hpp
+++ b/src/hotspot/share/libadt/vectset.hpp
@@ -55,9 +55,6 @@ public:
   VectorSet(Arena* arena);
   ~VectorSet() {}
 
-  // Allow move constructor for && (eg. capture return of function)
-  VectorSet(VectorSet&&) = default;
-
   void insert(uint elem);
   bool is_empty() const;
   void reset() {
diff --git a/src/hotspot/share/opto/loopPredicate.cpp b/src/hotspot/share/opto/loopPredicate.cpp
index 8511b3da897..04577ef35e4 100644
--- a/src/hotspot/share/opto/loopPredicate.cpp
+++ b/src/hotspot/share/opto/loopPredicate.cpp
@@ -229,7 +229,8 @@ ProjNode* PhaseIdealLoop::create_new_if_for_predicate(ProjNode* cont_proj, Node*
 // Update ctrl and control inputs of all data nodes starting from 'node' to 'new_ctrl' which have 'old_ctrl' as
 // current ctrl.
 void PhaseIdealLoop::set_ctrl_of_nodes_with_same_ctrl(Node* node, ProjNode* old_ctrl, Node* new_ctrl) {
-  Unique_Node_List nodes_with_same_ctrl = find_nodes_with_same_ctrl(node, old_ctrl);
+  Unique_Node_List nodes_with_same_ctrl;
+  find_nodes_with_same_ctrl(node, old_ctrl, nodes_with_same_ctrl);
   for (uint j = 0; j < nodes_with_same_ctrl.size(); j++) {
     Node* next = nodes_with_same_ctrl[j];
     if (next->in(0) == old_ctrl) {
@@ -240,8 +241,8 @@ void PhaseIdealLoop::set_ctrl_of_nodes_with_same_ctrl(Node* node, ProjNode* old_
 }
 
 // Recursively find all input nodes with the same ctrl.
-Unique_Node_List PhaseIdealLoop::find_nodes_with_same_ctrl(Node* node, const ProjNode* ctrl) {
-  Unique_Node_List nodes_with_same_ctrl;
+void PhaseIdealLoop::find_nodes_with_same_ctrl(Node* node, const ProjNode* ctrl, Unique_Node_List& nodes_with_same_ctrl) {
+  nodes_with_same_ctrl.ensure_empty();
   nodes_with_same_ctrl.push(node);
   for (uint j = 0; j < nodes_with_same_ctrl.size(); j++) {
     Node* next = nodes_with_same_ctrl[j];
@@ -252,15 +253,16 @@ Unique_Node_List PhaseIdealLoop::find_nodes_with_same_ctrl(Node* node, const Pro
       }
     }
   }
-  return nodes_with_same_ctrl;
 }
 
 // Clone all nodes with the same ctrl as 'old_ctrl' starting from 'node' by following its inputs. Rewire the cloned nodes
 // to 'new_ctrl'. Returns the clone of 'node'.
 Node* PhaseIdealLoop::clone_nodes_with_same_ctrl(Node* node, ProjNode* old_ctrl, Node* new_ctrl) {
   DEBUG_ONLY(uint last_idx = C->unique();)
-  Unique_Node_List nodes_with_same_ctrl = find_nodes_with_same_ctrl(node, old_ctrl);
-  Dict old_new_mapping = clone_nodes(nodes_with_same_ctrl); // Cloned but not rewired, yet
+  Unique_Node_List nodes_with_same_ctrl;
+  find_nodes_with_same_ctrl(node, old_ctrl, nodes_with_same_ctrl);
+  Dict old_new_mapping(cmpkey, hashkey);
+  clone_nodes(nodes_with_same_ctrl, old_new_mapping); // Cloned but not rewired, yet
   rewire_cloned_nodes_to_ctrl(old_ctrl, new_ctrl, nodes_with_same_ctrl, old_new_mapping);
   Node* clone_phi_input = static_cast<Node*>(old_new_mapping[node]);
   assert(clone_phi_input != nullptr && clone_phi_input->_idx >= last_idx, "must exist and be a proper clone");
@@ -268,15 +270,14 @@ Node* PhaseIdealLoop::clone_nodes_with_same_ctrl(Node* node, ProjNode* old_ctrl,
 }
 
 // Clone all the nodes on 'list_to_clone' and return an old->new mapping.
-Dict PhaseIdealLoop::clone_nodes(const Node_List& list_to_clone) {
-  Dict old_new_mapping(cmpkey, hashkey);
+void PhaseIdealLoop::clone_nodes(const Node_List& list_to_clone, Dict& old_new_mapping) {
+  assert(old_new_mapping.Size() == 0, "must be empty");
   for (uint i = 0; i < list_to_clone.size(); i++) {
     Node* next = list_to_clone[i];
     Node* clone = next->clone();
     _igvn.register_new_node_with_optimizer(clone);
     old_new_mapping.Insert(next, clone);
   }
-  return old_new_mapping;
 }
 
 // Rewire inputs of the unprocessed cloned nodes (inputs are not updated, yet, and still point to the old nodes) by
diff --git a/src/hotspot/share/opto/loopnode.hpp b/src/hotspot/share/opto/loopnode.hpp
index 0f876042eda..ae61fb87a6c 100644
--- a/src/hotspot/share/opto/loopnode.hpp
+++ b/src/hotspot/share/opto/loopnode.hpp
@@ -1345,9 +1345,9 @@ public:
  private:
   // Helper functions for create_new_if_for_predicate()
   void set_ctrl_of_nodes_with_same_ctrl(Node* node, ProjNode* old_ctrl, Node* new_ctrl);
-  Unique_Node_List find_nodes_with_same_ctrl(Node* node, const ProjNode* ctrl);
+  void find_nodes_with_same_ctrl(Node* node, const ProjNode* ctrl, Unique_Node_List& nodes_with_same_ctrl);
   Node* clone_nodes_with_same_ctrl(Node* node, ProjNode* old_ctrl, Node* new_ctrl);
-  Dict clone_nodes(const Node_List& list_to_clone);
+  void clone_nodes(const Node_List& list_to_clone, Dict& old_new_mapping);
   void rewire_cloned_nodes_to_ctrl(const ProjNode* old_ctrl, Node* new_ctrl, const Node_List& nodes_with_same_ctrl,
                                    const Dict& old_new_mapping);
   void rewire_inputs_of_clones_to_clones(Node* new_ctrl, Node* clone, const Dict& old_new_mapping, const Node* next);
diff --git a/src/hotspot/share/opto/node.hpp b/src/hotspot/share/opto/node.hpp
index 43375187abc..b85459130d5 100644
--- a/src/hotspot/share/opto/node.hpp
+++ b/src/hotspot/share/opto/node.hpp
@@ -1539,9 +1539,6 @@ public:
   }
   Node_Array() : Node_Array(Thread::current()->resource_area()) {}
 
-  // Allow move constructor for && (eg. capture return of function)
-  Node_Array(Node_Array&&) = default;
-
   Node *operator[] ( uint i ) const // Lookup, or null for not mapped
   { return (i<_max) ? _nodes[i] : (Node*)nullptr; }
   Node* at(uint i) const { assert(i<_max,"oob"); return _nodes[i]; }
@@ -1568,9 +1565,6 @@ public:
   Node_List(uint max = OptoNodeListSize) : Node_Array(Thread::current()->resource_area(), max), _cnt(0) {}
   Node_List(Arena *a, uint max = OptoNodeListSize) : Node_Array(a, max), _cnt(0) {}
 
-  // Allow move constructor for && (eg. capture return of function)
-  Node_List(Node_List&&) = default;
-
   bool contains(const Node* n) const {
     for (uint e = 0; e < size(); e++) {
       if (at(e) == n) return true;
@@ -1607,9 +1601,6 @@ public:
   Unique_Node_List() : Node_List(), _clock_index(0) {}
   Unique_Node_List(Arena *a) : Node_List(a), _in_worklist(a), _clock_index(0) {}
 
-  // Allow move constructor for && (eg. capture return of function)
-  Unique_Node_List(Unique_Node_List&&) = default;
-
   void remove( Node *n );
   bool member( Node *n ) { return _in_worklist.test(n->_idx) != 0; }
   VectorSet& member_set(){ return _in_worklist; }
diff --git a/src/hotspot/share/opto/stringopts.cpp b/src/hotspot/share/opto/stringopts.cpp
index 399f18ce9aa..a2589d0f6fb 100644
--- a/src/hotspot/share/opto/stringopts.cpp
+++ b/src/hotspot/share/opto/stringopts.cpp
@@ -369,8 +369,8 @@ void StringConcat::eliminate_initialize(InitializeNode* init) {
   init->disconnect_inputs(C);
 }
 
-Node_List PhaseStringOpts::collect_toString_calls() {
-  Node_List string_calls;
+void PhaseStringOpts::collect_toString_calls(Node_List& string_calls) {
+  assert(string_calls.size() == 0, "output list must be empty");
   Node_List worklist;
 
   _visited.clear();
@@ -405,7 +405,6 @@ Node_List PhaseStringOpts::collect_toString_calls() {
 #ifndef PRODUCT
   Atomic::add(&_stropts_total, encountered);
 #endif
-  return string_calls;
 }
 
 // Recognize a fluent-chain of StringBuilder/Buffer. They are either explicit usages
@@ -647,7 +646,8 @@ PhaseStringOpts::PhaseStringOpts(PhaseGVN* gvn):
   // if it's possible to fuse the usage of the SB into a single String
   // construction.
   GrowableArray<StringConcat*> concats;
-  Node_List toStrings = collect_toString_calls();
+  Node_List toStrings;
+  collect_toString_calls(toStrings);
   while (toStrings.size() > 0) {
     StringConcat* sc = build_candidate(toStrings.pop()->as_CallStaticJava());
     if (sc != nullptr) {
diff --git a/src/hotspot/share/opto/stringopts.hpp b/src/hotspot/share/opto/stringopts.hpp
index 21be4109c7d..0ad349c32d1 100644
--- a/src/hotspot/share/opto/stringopts.hpp
+++ b/src/hotspot/share/opto/stringopts.hpp
@@ -47,7 +47,7 @@ class PhaseStringOpts : public Phase {
   VectorSet _visited;
 
   // Collect a list of all SB.toString calls
-  Node_List collect_toString_calls();
+  void collect_toString_calls(Node_List& string_calls);
 
   // Examine the use of the SB alloc to see if it can be replace with
   // a single string construction.
diff --git a/src/hotspot/share/opto/superword.cpp b/src/hotspot/share/opto/superword.cpp
index b1e4677a7cb..823bd05b67b 100644
--- a/src/hotspot/share/opto/superword.cpp
+++ b/src/hotspot/share/opto/superword.cpp
@@ -2502,8 +2502,6 @@ private:
   GrowableArray<int> _incnt;               // number of (implicit) in-edges
   int _max_pid = 0;
 
-  bool _schedule_success;
-
   SuperWord* _slp;
 public:
   PacksetGraph(SuperWord* slp)
@@ -2547,7 +2545,6 @@ public:
   int incnt(int pid) { return _incnt.at(pid - 1); }
   void incnt_set(int pid, int cnt) { return _incnt.at_put(pid - 1, cnt); }
   GrowableArray<int>& out(int pid) { return _out.at(pid - 1); }
-  bool schedule_success() const { return _schedule_success; }
 
   // Create nodes (from packs and scalar-nodes), and add edges, based on DepPreds.
   void build() {
@@ -2626,10 +2623,10 @@ public:
 
   // Schedule nodes of PacksetGraph to worklist, using topsort: schedule a node
   // that has zero incnt. If a PacksetGraph node corresponds to memops, then add
-  // those to the memops_schedule. At the end, we return the memops_schedule, and
-  // note if topsort was successful.
-  Node_List schedule() {
-    Node_List memops_schedule;
+  // those to the memops_schedule. We return true if topsort was successful, and
+  // false if there was a cycle.
+  bool schedule(Node_List& memops_schedule) {
+    assert(memops_schedule.size() == 0, "output list must be empty");
     GrowableArray<int> worklist;
     // Directly schedule all nodes without precedence
     for (int pid = 1; pid <= _max_pid; pid++) {
@@ -2668,8 +2665,8 @@ public:
     }
 
     // Was every pid scheduled? If not, we found some cycles in the PacksetGraph.
-    _schedule_success = (worklist.length() == _max_pid);
-    return memops_schedule;
+    bool schedule_success = (worklist.length() == _max_pid);
+    return schedule_success;
   }
 
   // Print the PacksetGraph.
@@ -2722,7 +2719,8 @@ void SuperWord::schedule() {
   graph.build();
 
   // (2) Schedule the PacksetGraph.
-  Node_List memops_schedule = graph.schedule();
+  Node_List memops_schedule;
+  bool schedule_success = graph.schedule(memops_schedule);
 
   // (3) Check if the PacksetGraph schedule succeeded (had no cycles).
   // We now know that we only have independent packs, see verify_packs.
@@ -2730,7 +2728,7 @@ void SuperWord::schedule() {
   // graph (DAG) after scheduling. Thus, we must check if the packs have
   // introduced a cycle. The SuperWord paper mentions the need for this
   // in "3.7 Scheduling".
-  if (!graph.schedule_success()) {
+  if (schedule_success) {
     if (TraceSuperWord) {
       tty->print_cr("SuperWord::schedule found cycle in PacksetGraph:");
       graph.print(true, false);
```

eme64:
@kimbarrett @jcking So if we never use `std::move` (currently not used at all in the HotSpot code), do I actually need to have a custom implementation of the move-constructor, or is the `default` enough?

kimbarrett:
> @kimbarrett @jcking So if we never use `std::move` (currently not used at all in the HotSpot code), do I actually need to have a custom implementation of the move-constructor, or is the `default` enough?

It depends on the class whether a simple shallow-copy (the default) is an
adequate move. The main question is whether the destruction of the moved-from
object might damage/delete something that was pilfered for use in the moved-to
object. Since from a quick skim it looked like the destructors for all of the
classes being given move constructors are empty or (implicitly) defaulted, I
think that shouldn’t be a problem. But someone should check them more
carefully. In particular, are there any bases or nested members with
non-trivial destructors?

The problem with std::move is that one can convert an lvalue to an rvalue,
move construct from the rvalue, and then try to use both the old lvalue
(potential UB) and the newly moved-to object, even though they share state
(source of UB). (Also remember that std::move is basically just a convenience
wrapper over a cast of an lvalue to an rvalue - there’s nothing magic about
it. But we’re not doing that anywhere either.)

eme64:
@kimbarrett Thanks for the explanations.

All of the containers in question are subclasses from `AnyObj`:
```
AnyObj -> Dict
AnyObj -> VectorSet
AnyObj -> Node_Array -> Node_List -> Unique_Node_List
```

`AnyObj` does have a destructor, it overwrites some internal addresses with `badHeapOopVal`. I'm not sure about this, but maybe that is ok? `VectorSet` has an empty destructor, so that should be ok. `Dict` as well. `Node_Array`, `Node_List` and `Unique_Node_List` have the implicit destroctor. And I think shallow copies of everything is ok.

I tried to still implement proper move-constructors for the 5 containers. `Dict`, `VectorSet` and `Node_Array` was relatively straight forward. But for `Node_List` and `Unique_Node_List` I think I would have to call the suberclass (baseclass) move-constructor, so that the corresponding fields of the superclasses are also handled. But for that, I would have to explicitly use `std::move` on the baseclass. That would be a first in the whole code-base. And I'd have to `#include <utility>`.

So is it really worth it to implement the move-constructors, just to `nullptr` out some fileds? Or should I just do it where it is easy, and leave the rest? Or just leave them all default?
@kimbarrett @jcking do you have any better alternative?

eme64:
This is what I have so far, using `std::move` explicitly. @kimbarrett @jcking I'm not sure this is very nice. Any suggestions?
```
diff --git a/src/hotspot/share/libadt/dict.hpp b/src/hotspot/share/libadt/dict.hpp
index c021536c402..0a9ff85c192 100644
--- a/src/hotspot/share/libadt/dict.hpp
+++ b/src/hotspot/share/libadt/dict.hpp
@@ -62,7 +62,20 @@ class Dict : public AnyObj { // Dictionary structure
   ~Dict();
 
   // Allow move constructor for && (eg. capture return of function)
-  Dict(Dict&&) = default;
+  Dict(Dict&& other) : _arena(other._arena),
+                       _bin(other._bin),
+                       _size(other._size),
+                       _cnt(other._cnt),
+                       _hash(other._hash),
+                       _cmp(other._cmp) {
+    // Other is invalidated: SIGSEGV upon use.
+    other._arena = nullptr;
+    other._bin = nullptr;
+    other._size = 0;
+    other._cnt = 0;
+  };
+  Dict& operator=(Dict&&) = delete;
+  NONCOPYABLE(Dict);
 
   // Return # of key-value pairs in dict
   uint32_t Size(void) const { return _cnt; }
@@ -78,8 +91,6 @@ class Dict : public AnyObj { // Dictionary structure
 
   // Print out the dictionary contents as key-value pairs
   void print();
-
-  NONCOPYABLE(Dict);
 };
 
 // Hashing functions
diff --git a/src/hotspot/share/libadt/vectset.hpp b/src/hotspot/share/libadt/vectset.hpp
index a82046f2ba9..c7f08da9e43 100644
--- a/src/hotspot/share/libadt/vectset.hpp
+++ b/src/hotspot/share/libadt/vectset.hpp
@@ -25,6 +25,7 @@
 #ifndef SHARE_LIBADT_VECTSET_HPP
 #define SHARE_LIBADT_VECTSET_HPP
 
+#include <utility>
 #include "memory/allocation.hpp"
 #include "utilities/copy.hpp"
 
@@ -56,7 +57,18 @@ public:
   ~VectorSet() {}
 
   // Allow move constructor for && (eg. capture return of function)
-  VectorSet(VectorSet&&) = default;
+  VectorSet(VectorSet&& other) : _size(other._size),
+                                 _data(other._data),
+                                 _data_size(other._data_size),
+                                 _set_arena(other._set_arena) {
+    // Other is invalidated: reads as if empty, SIGSEGV upon write.
+    other._size = 0;
+    other._data = nullptr;
+    other._data_size = 0;
+    other._set_arena = nullptr;
+  };
+  VectorSet& operator=(VectorSet&&) = delete;
+  NONCOPYABLE(VectorSet);
 
   void insert(uint elem);
   bool is_empty() const;
@@ -113,8 +125,6 @@ public:
     uint32_t mask = 1U << (elem & bit_mask);
     _data[word] |= mask;
   }
-
-  NONCOPYABLE(VectorSet);
 };
 
 #endif // SHARE_LIBADT_VECTSET_HPP
diff --git a/src/hotspot/share/opto/compile.cpp b/src/hotspot/share/opto/compile.cpp
index 536a4e7b46d..3f93406e18f 100644
--- a/src/hotspot/share/opto/compile.cpp
+++ b/src/hotspot/share/opto/compile.cpp
@@ -2192,6 +2192,60 @@ void Compile::remove_root_to_sfpts_edges(PhaseIterGVN& igvn) {
 void Compile::Optimize() {
   TracePhase tp("optimizer", &timers[_t_optimizer]);
 
+  // TODO remove playground
+  VectorSet sss;
+  sss.insert(10000);
+  tty->print_cr("VectorSet 1: %d", sss.test(10000));
+  VectorSet s2(std::move(sss));
+  tty->print_cr("VectorSet 2: %d", s2.test(10000));
+  tty->print_cr("VectorSet 3: %d", sss.test(10000));
+  s2.insert(20000);
+  tty->print_cr("VectorSet 4: %d", s2.test(20000));
+  // SIGSEGV upon insert:
+  // sss.insert(10000);
+
+  Dict d1(cmpkey, hashkey);
+  d1.Insert((void*)1000, (void*)42);
+  tty->print_cr("Dict 1: %ld", (long)d1[(void*)1000]);
+  Dict d2(std::move(d1));
+  tty->print_cr("Dict 2: %ld", (long)d2[(void*)1000]);
+  // SIGSEGV upon read:
+  //tty->print_cr("Dict 3: %ld", (long)d1[(void*)1000]);
+
+  Node_Array a1;
+  a1.map(10000, (Node*)42);
+  tty->print_cr("Node_Array 1: %ld", (long)a1.at(10000));
+  Node_Array a2(std::move(a1));
+  tty->print_cr("Node_Array 2: %ld", (long)a2.at(10000));
+  tty->print_cr("Node_Array 3: max %ld", (long)a1.max());
+  // oob assert on read:
+  // tty->print_cr("Node_Array 4: %ld", (long)a1.at(10000));
+  // asserts because _max not larger than zero:
+  // a1.map(10000, (Node*)666);
+
+  Node_List l1;
+  l1.push((Node*)42);
+  tty->print_cr("Node_List 1: %ld", (long)l1.at(0));
+  Node_List l2(std::move(l1));
+  tty->print_cr("Node_List 2: %ld", (long)l2.at(0));
+  tty->print_cr("Node_List 3: size %ld", (long)l1.size());
+  // asserts because _max not larger than zero:
+  // l1.push((Node*)666);
+
+  Unique_Node_List w1;
+  Node* nt0 = new Node(2);
+  w1.push(nt0);
+  w1.dump();
+  Node_List w2(std::move(w1));
+  w2.dump();
+  tty->print_cr("Unique_Node_List 1: size %ld", (long)w1.size());
+  w1.dump();
+  // does not quite behave correctly when write to w1
+  Node* nt1 = new Node(2);
+  w1.push(nt0);
+  tty->print_cr("Unique_Node_List 2: size %ld", (long)w1.size());
+  w1.dump();
+
 #ifndef PRODUCT
   if (env()->break_at_compile()) {
     BREAKPOINT;
diff --git a/src/hotspot/share/opto/node.hpp b/src/hotspot/share/opto/node.hpp
index 43375187abc..e65915acbba 100644
--- a/src/hotspot/share/opto/node.hpp
+++ b/src/hotspot/share/opto/node.hpp
@@ -1540,7 +1540,16 @@ public:
   Node_Array() : Node_Array(Thread::current()->resource_area()) {}
 
   // Allow move constructor for && (eg. capture return of function)
-  Node_Array(Node_Array&&) = default;
+  Node_Array(Node_Array&& other) : _a(other._a),
+                                   _max(other._max),
+                                   _nodes(other._nodes) {
+    // Other is invalidated: reads as if empty, asserts on write.
+    other._a = nullptr;
+    other._max = 0;
+    other._nodes = nullptr;
+  };
+  Node_Array& operator=(Node_Array&&) = delete;
+  NONCOPYABLE(Node_Array);
 
   Node *operator[] ( uint i ) const // Lookup, or null for not mapped
   { return (i<_max) ? _nodes[i] : (Node*)nullptr; }
@@ -1557,8 +1566,6 @@ public:
 
   uint max() const { return _max; }
   void dump() const;
-
-  NONCOPYABLE(Node_Array);
 };
 
 class Node_List : public Node_Array {
@@ -1569,7 +1576,12 @@ public:
   Node_List(Arena *a, uint max = OptoNodeListSize) : Node_Array(a, max), _cnt(0) {}
 
   // Allow move constructor for && (eg. capture return of function)
-  Node_List(Node_List&&) = default;
+  Node_List(Node_List&& other) : Node_Array(std::move(other)), _cnt(other._cnt) {
+    // Other is invalidated: reads as if empty, asserts on write.
+    other._cnt = 0;
+  };
+  Node_List& operator=(Node_List&&) = delete;
+  NONCOPYABLE(Node_List);
 
   bool contains(const Node* n) const {
     for (uint e = 0; e < size(); e++) {
@@ -1594,8 +1606,6 @@ public:
   uint size() const { return _cnt; }
   void dump() const;
   void dump_simple() const;
-
-  NONCOPYABLE(Node_List);
 };
 
 //------------------------------Unique_Node_List-------------------------------
@@ -1608,7 +1618,12 @@ public:
   Unique_Node_List(Arena *a) : Node_List(a), _in_worklist(a), _clock_index(0) {}
 
   // Allow move constructor for && (eg. capture return of function)
-  Unique_Node_List(Unique_Node_List&&) = default;
+  Unique_Node_List(Unique_Node_List&& other) : Node_List(std::move(other)),
+                                               _in_worklist(std::move(other._in_worklist)),
+                                               _clock_index(std::move(other._clock_index)) {
+    // Other is invalidated: reads as if empty, may behave incorrectly on write.
+    other._clock_index = 0;
+  };
 
   void remove( Node *n );
   bool member( Node *n ) { return _in_worklist.test(n->_idx) != 0; }
```

eme64:
@kimbarrett I quote you from the slack conversation:
> “Invalidating” the moved-from object to protect against someone using std::move (which at least for now we don’t do) and then doing something they aren’t supposed to do doesn’t seem like a good path to me. Anyone using std::move is taking responsibility for not doing anything they shouldn’t with the moved-from object. Also remember that the move constructors in question are a workaround for not having C++17.  I think it is likely that we’ll have C++17 before we start making any use (or at least any significant use, e.g. beyond things like the constructor chaining in your example) of std::move.

I will hence just keep the `default` move constructors, but and `delete` the move assign operators.

eme64:
@jcking @TobiHartmann @chhagedorn @kimbarrett 
Thanks for all the reviewing!

I'm running testing again up to tier6 and stress testing.
I will integrate in 24h, unless you bring up some concerns.

eme64:
Thanks @jcking for the original PR and the review!
Thanks @kimbarrett for the guidance on move semantics!
And of course thanks @TobiHartmann  @chhagedorn for the review and the many suggestions!

/integrate

PR #13843, opened 2023-05-05
8307533: Use atomic bitset functions for metadata flags

coleenp:
Replace the bit set copies from metadata to use the Atomic functions.
Tested with tier1-4.

---------
### Progress

### Issue
 * [JDK-8307533](https://bugs.openjdk.org/browse/JDK-8307533): Use atomic bitset functions for metadata flags

### Reviewers
 * [Calvin Cheung](https://openjdk.org/census#ccheung) (@calvinccheung - **Reviewer**) ⚠️ Review applies to [7009b524](https://git.openjdk.org/jdk/pull/13843/files/7009b524fc326740c92040dde87544849f3a2c7d)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13843/head:pull/13843` \
`$ git checkout pull/13843`

`$ git checkout pull/13843` \
`$ git pull https://git.openjdk.org/jdk.git pull/13843/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13843`

`$ git pr show -t 13843`

</details>

<a href="https://git.openjdk.org/jdk/pull/13843.diff">https://git.openjdk.org/jdk/pull/13843.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13843#issuecomment-1536720673)


calvinccheung (approved):
LGTM.
Looks like we can do similar change to the `set_defined_by_cds_in_class_path()` function in packageEntry.hpp. I can file a bug to take care of that.

src/hotspot/share/oops/fieldInfo.inline.hpp
```diff
@@ -151,23 +152,12 @@ inline FieldInfoReader& FieldInfoReader::set_position_and_next_index(int positio
...
 inline void FieldStatus::atomic_clear_bits(u1& flags, u1 mask) {
-  // Atomically update the flags with the bits given
-  u1 old_flags, new_flags, witness;
-  do {
-    old_flags = flags;
-    new_flags = old_flags & ~mask;
-    witness = Atomic::cmpxchg(&flags, old_flags, new_flags);
-  } while (witness != old_flags);
+  u1 val = (~mask);
+  Atomic::fetch_then_and(&flags, val);
```
kimbarrett, line 160: u1 is not a supported type for Atomic bitops.  This only happens to work right now because all
platforms are currently using a cmpxchg-based implementation and aren't enforcing the documented
limitation of only providing support for size of an int or size of a pointer (if different).

src/hotspot/share/oops/instanceKlassFlags.hpp
```diff
@@ -121,8 +123,8 @@ class InstanceKlassFlags {
   IK_STATUS_DO(IK_STATUS_GET_SET)
 #undef IK_STATUS_GET_SET
 
-  void atomic_set_bits(u1 bits);
-  void atomic_clear_bits(u1 bits);
+  void atomic_set_bits(u1 bits)   { Atomic::fetch_then_or(&_status, bits); }
+  void atomic_clear_bits(u1 bits) { u1 val = (~bits); Atomic::fetch_then_and(&_status, val); }
```
kimbarrett, line 127: Again here, u1 is not a supported type for Atomic bitops.

src/hotspot/share/oops/methodFlags.hpp
```diff
@@ -86,8 +87,8 @@ class MethodFlags {
 #undef M_STATUS_GET_SET
 
   int as_int() const { return _status; }
-  void atomic_set_bits(u4 bits);
-  void atomic_clear_bits(u4 bits);
+  void atomic_set_bits(u4 bits)   { Atomic::fetch_then_or(&_status, bits); }
+  void atomic_clear_bits(u4 bits) { u4 val = (~bits); Atomic::fetch_then_and(&_status, val); }
```
kimbarrett, line 91: Why introduce a new variable (and why the extra parens).  Just
```
Atomic::fetch_then_and(&_status, ~bits);
```

src/hotspot/share/oops/instanceKlassFlags.hpp
kimbarrett, line 127: Why introduce a new variable?

src/hotspot/share/oops/fieldInfo.inline.hpp
```diff
@@ -151,23 +152,12 @@ inline FieldInfoReader& FieldInfoReader::set_position_and_next_index(int positio
...
 
 inline void FieldStatus::atomic_clear_bits(u1& flags, u1 mask) {
-  // Atomically update the flags with the bits given
-  u1 old_flags, new_flags, witness;
-  do {
-    old_flags = flags;
-    new_flags = old_flags & ~mask;
-    witness = Atomic::cmpxchg(&flags, old_flags, new_flags);
-  } while (witness != old_flags);
+  u1 val = (~mask);
```
kimbarrett, line 159: Why introduce a new variable?

kimbarrett (changes requested):

src/hotspot/share/oops/fieldInfo.inline.hpp
coleenp: But I need u1! I thought that was the point of having the templates?  Do I change this back to my own CAS loop?

src/hotspot/share/oops/methodFlags.hpp
coleenp: The template didn't like this, I suppose I could add some casting.

kimbarrett:
The failures to build when the extra variables are removed is because of integral arithmetic promotion by the `~bits` expression.  That causes the resulting value of that expression to be of a different type from the atomic value.

kimbarrett:
The Atomic bitops aren't intended to support other sizes; only the same sizes as Atomic::add and friends.  That narrower
types are currently supported by the default implementation is an accident.  Platform specializations might not have such
support, since the underlying platform might not have it.

If support for narrower types is a (not previously known to me) requirement, some non-trivial changes may be needed.
Among other things, I think the current very simple platform specialization mechanism won't be sufficient.

coleenp:
@kimbarrett thank you for adding support for u1 types.  I merged with the latest changes.

coleenp:
Thanks Calvin and Kim
/integrate

PR #11044, opened 2022-11-08
8139457: Relax alignment of array elements

rkennke:
See [JDK-8139457](https://bugs.openjdk.org/browse/JDK-8139457) for details.

Basically, when running with -XX:-UseCompressedClassPointers, arrays will have a gap between the length field and the first array element, because array elements will only start at word-aligned offsets. This is not necessary for smaller-than-word elements.

Also, while it is not very important now, it will become very important with Lilliput, which eliminates the Klass field and would always put the length field at offset 8, and leave a gap between offset 12 and 16.

Testing:
 - [x] runtime/FieldLayout/ArrayBaseOffsets.java (x86_64, x86_32, aarch64, arm, riscv, s390)
 - [x] bootcycle (x86_64, x86_32, aarch64, arm, riscv, s390)
 - [x] tier1 (x86_64, x86_32, aarch64, riscv)
 - [x] tier2 (x86_64, aarch64, riscv)
 - [x] tier3 (x86_64, riscv)

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Change requires CSR request [JDK-8314882](https://bugs.openjdk.org/browse/JDK-8314882) to be approved
- [x] Commit message must refer to an issue

### Issues
 * [JDK-8139457](https://bugs.openjdk.org/browse/JDK-8139457): Relax alignment of array elements (**Enhancement** - P4)
 * [JDK-8314882](https://bugs.openjdk.org/browse/JDK-8314882): Relax alignment of array elements (**CSR**)

### Reviewers
 * [Thomas Stuefe](https://openjdk.org/census#stuefe) (@tstuefe - **Reviewer**) ⚠️ Review applies to [e8c1e408](https://git.openjdk.org/jdk/pull/11044/files/e8c1e4081cb3945c8f79c204caec2c1c77bff6e8)
 * [Stefan Karlsson](https://openjdk.org/census#stefank) (@stefank - **Reviewer**)
 * [Aleksey Shipilev](https://openjdk.org/census#shade) (@shipilev - **Reviewer**) ⚠️ Review applies to [3e37e785](https://git.openjdk.org/jdk/pull/11044/files/3e37e7852780273b7c72f8bf14693a3703b3ff2d)
[...]

PR #10907, opened 2022-10-28
8291555: Implement alternative fast-locking scheme

rkennke:
This change adds a fast-locking scheme as an alternative to the current stack-locking implementation. It retains the advantages of stack-locking (namely fast locking in uncontended code-paths), while avoiding the overload of the mark word. That overloading causes massive problems with Lilliput, because it means we have to check and deal with this situation when trying to access the mark-word. And because of the very racy nature, this turns out to be very complex and would involve a variant of the inflation protocol to ensure that the object header is stable. (The current implementation of setting/fetching the i-hash provides a glimpse into the complexity).

What the original stack-locking does is basically to push a stack-lock onto the stack which consists only of the displaced header, and CAS a pointer to this stack location into the object header (the lowest two header bits being 00 indicate 'stack-locked'). The pointer into the stack can then be used to identify which thread currently owns the lock.

This change basically reverses stack-locking: It still CASes the lowest two header bits to 00 to indicate 'fast-locked' but does *not* overload the upper bits with a stack-pointer. Instead, it pushes the object-reference to a thread-local lock-stack. This is a new structure which is basically a small array of oops that is associated with each thread. Experience shows that this array typcially remains very small (3-5 elements). Using this lock stack, it is possible to query which threads own which locks. Most importantly, the most common question 'does the current thread own me?' is very quickly answered by doing a quick scan of the array. More complex queries like 'which thread owns X?' are not performed in very performance-critical paths (usually in code like JVMTI or deadlock detection) where it is ok to do more complex operations (and we already do). The lock-stack is also a new set of GC roots, and would be scanned during thread scanning, possibly concurrently, via the normal protocols.

The lock-stack is fixed size, currently with 8 elements. According to my experiments with various workloads, this covers the vast majority of workloads (in-fact, most workloads seem to never exceed 5 active locks per thread at a time). We check for overflow in the fast-paths and when the lock-stack is full, we take the slow-path, which would inflate the lock to a monitor. That case should be very rare.

In contrast to stack-locking, fast-locking does *not* support recursive locking (yet). When that happens, the fast-lock gets inflated to a full monitor. It is not clear if it is worth to add support for recursive fast-locking.

[...]

PR #13844, opened 2023-05-05
8305895: Implementation: JEP 450: Compact Object Headers (Experimental)

rkennke:
This is the main body of the JEP 450: Compact Object Headers (Experimental).

Main changes:
 - Introduction of the (experimental) flag UseCompactObjectHeaders. All changes in this PR are protected by this flag.
 - The compressed Klass* can now be stored in the mark-word of objects. In order to be able to do this, we are building on #10907, #13582 and #13779 to protect the relevant (upper 32) bits of the mark-word. Significant parts of this PR deal with loading the compressed Klass* from the mark-word, and dealing with (monitor-)locked objects. When the object is monitor-locked, we load the displaced mark-word from the monitor, and load the compressed Klass* from there. This PR also changes some code paths (mostly in GCs) to be more careful when accessing Klass* (or mark-word or size) to be able to fetch it from the forwardee in case the object is forwarded, and/or reach through to the monitor when the object is locked by a monitor.
 - The identity hash-code is narrowed to 25 bits.
 - Instances can now have their base-offset (the offset where the field layouter starts to place fields) at offset 8 (instead of 12 or 16).
 - Arrays will can now store their length at offset 8. Due to alignment restrictions, array elements will still start at offset 16. #11044 will resolve that restriction and allow array elements to start at offset 12 (except for long, double and uncompressed oops, which are still required to start at an element-aligned offset).
 - CDS can now write and read archives with the compressed header. However, it is not possible to read an archive that has been written with an opposite setting of UseCompactObjectHeaders.

Testing:
 (+UseCompactObjectHeaders tests are run with the flag hard-patched into the build, to also catch @flagless tests, and to avoid mismatches with CDS - see above.)
 - [x] tier1 (x86_64)
 - [x] tier2 (x86_64)
 - [x] tier3 (x86_64)
 - [ ] tier4 (x86_64)
 - [x] tier1 (aarch64)
 - [x] tier2 (aarch64)
 - [x] tier3 (aarch64)
 - [ ] tier4 (aarch64)
 - [ ] tier1 (x86_64) +UseCompactObjectHeaders
 - [ ] tier2 (x86_64) +UseCompactObjectHeaders
 - [ ] tier3 (x86_64) +UseCompactObjectHeaders
 - [ ] tier4 (x86_64) +UseCompactObjectHeaders
 - [ ] tier1 (aarch64) +UseCompactObjectHeaders
 - [ ] tier2 (aarch64) +UseCompactObjectHeaders
 - [ ] tier3 (aarch64) +UseCompactObjectHeaders
 - [ ] tier4 (aarch64) +UseCompactObjectHeaders

---------
### Progress
- [ ] Change requires CSR request [JDK-8306000](https://bugs.openjdk.org/browse/JDK-8306000) to be approved

### Integration blocker
&nbsp;⚠️ Dependency #13779 must be integrated first

### Issues
 * [JDK-8305895](https://bugs.openjdk.org/browse/JDK-8305895): Implementation: JEP 450: Compact Object Headers (Experimental)
 * [JDK-8306000](https://bugs.openjdk.org/browse/JDK-8306000): Add experimental -XX:+UseCompactObjectHeaders flag (**CSR**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13844/head:pull/13844` \
`$ git checkout pull/13844`

`$ git checkout pull/13844` \
`$ git pull https://git.openjdk.org/jdk.git pull/13844/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13844`

`$ git pr show -t 13844`

</details>

<a href="https://git.openjdk.org/jdk/pull/13844.diff">https://git.openjdk.org/jdk/pull/13844.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13844#issuecomment-1538551476)


merykitty:
I'm not sure if this is trivial or significant, but if you limit the class pointer to 30 bit, and use the upper 2 bits for locking, then you can obtain the class pointer in less instructions:

    movl dst, [obj + 4]
    andl dst, 0xBFFFFFFF
    jl slow_path

This exploits the fact that the most significant bit represents a negative number, so it clears the unrelated bit and checks for valid header at the same time, the sequence is only 2 instructions long after macro fusion, compared to the current value of 3.

This also allows quick class comparisons against constants, assuming that most instance is in unlock state, the comparison when equality is likely can be done:

    cmpl [obj + 4], con | 0x40000000
    jne slow_path

This can be matched on an `If` so that the `slow_path` can branch to the `IfTrue` label directly, and the fast path has only 1 comparison and 1 conditional jump.

Thanks.

rkennke:
> I'm not sure if this is trivial or significant, but if you limit the class pointer to 30 bit, and use the upper 2 bits for locking, then you can obtain the class pointer in less instructions:
> 
> ```
> movl dst, [obj + 4]
> andl dst, 0xBFFFFFFF
> jl slow_path
> ```
> 
> This exploits the fact that the most significant bit represents a negative number, so it clears the unrelated bit and checks for valid header at the same time, the sequence is only 2 instructions long after macro fusion, compared to the current value of 3.
> 
> This also allows quick class comparisons against constants, assuming that most instance is in unlock state, the comparison when equality is likely can be done:
> 
> ```
> cmpl [obj + 4], con | 0x40000000
> jne slow_path
> ```
> 
> This can be matched on an `If` so that the `slow_path` can branch to the `IfTrue` label directly, and the fast path has only 1 comparison and 1 conditional jump.
> 
> Thanks.

These are great suggestions! I would shy away from doing it in this PR, though, because this also affects the locking subsystem and would cause quite intrusive changes and invalidate all the testing that we've done. Let's consider this in the Lilliput project and upstream the optimization separately, ok?

Thanks!
Roman

forax:
This seems great as an intermediary step toward a 32 bits header.

src/hotspot/share/oops/typeArrayKlass.cpp
```diff
@@ -228,7 +228,6 @@ Klass* TypeArrayKlass::array_klass_or_null() {
 }
 
 size_t TypeArrayKlass::oop_size(oop obj) const {
-  assert(obj->is_typeArray(),"must be a type array");
```
shipilev, line 231: So, these checks are removed because they reach for class, which we cannot use with `UseCompactObjectHeaders`? Better to disable the assert at this time, I think, instead of removing it.

src/hotspot/share/oops/oop.hpp
```diff
@@ -108,6 +121,17 @@ class oopDesc {
   // to be able to figure out the size of an object knowing its klass.
   inline size_t size_given_klass(Klass* klass);
 
+  // The following set of methods is used to access the mark-word and related
+  // properties when the object may be forwarded. Be careful where and when
+  // using this method. It assumes that the forwardee is installed in
+  // the header as a plain pointer (or self-forwarded). In particular,
+  // those methods can not deal with the sliding-forwarding that is used
+  // in Serial, G1 and Shenandoah full-GCs.
+  inline markWord forward_safe_mark() const;
```
shipilev, line 130: `forward_safe_mark` seems to be only used in `forward_safe_klass`, can be inlined/simplified there?

src/hotspot/share/oops/oop.hpp
```diff
@@ -108,6 +121,17 @@ class oopDesc {
   // to be able to figure out the size of an object knowing its klass.
   inline size_t size_given_klass(Klass* klass);
 
+  // The following set of methods is used to access the mark-word and related
```
shipilev, line 124: So, these are done to avoid introducing branches on the paths where objects are definitely _not_ forwarded? Are there fewer places than where we expect forwardings? Maybe the better way would be to make all methods handle the occasional forwarding, and then provide the methods that provide the _fast-path_, like `fast_mark`, `fast_class`, etc?

src/hotspot/share/gc/parallel/psOldGen.cpp
```diff
@@ -395,7 +395,7 @@ class VerifyObjectStartArrayClosure : public ObjectClosure {
     _start_array(start_array) { }
 
   virtual void do_object(oop obj) {
-    HeapWord* test_addr = cast_from_oop<HeapWord*>(obj) + 1;
+    HeapWord* test_addr = cast_from_oop<HeapWord*>(obj);
```
shipilev, line 398: I thought this `+1` is specifically to test that `object_start` is able to find the object header when given the interior pointer. See the `guarantee`-s in the next lines.

src/hotspot/share/gc/parallel/psPromotionManager.inline.hpp
```diff
@@ -247,9 +247,11 @@ inline oop PSPromotionManager::copy_unmarked_to_survivor_space(oop o,
   // Copy obj
   Copy::aligned_disjoint_words(cast_from_oop<HeapWord*>(o), cast_from_oop<HeapWord*>(new_obj), new_obj_size);
 
-  // Parallel GC claims with a release - so other threads might access this object
-  // after claiming and they should see the "completed" object.
-  ContinuationGCSupport::transform_stack_chunk(new_obj);
+  if (!new_obj->mark().is_marked()) {
```
shipilev, line 250: Oh, this is the same as for Shenandoah. See the comment there: probably want to condition this on `UseCompactObjectHeaders`.

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
```diff
@@ -322,8 +322,13 @@ inline oop ShenandoahHeap::evacuate_object(oop p, Thread* thread) {
 
   // Try to install the new forwarding pointer.
   oop copy_val = cast_to_oop(copy);
-  ContinuationGCSupport::relativize_stack_chunk(copy_val);
-
+  if (!copy_val->mark().is_marked()) {
+    // If we copied a mark-word that indicates 'forwarded' state, then
```
shipilev, line 326: Ouch. This is only the problem with `UseCompactObjectHeaders`, right? Can additionally conditionalize on that, so that legacy code path stays the same.

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
```diff
@@ -305,7 +305,8 @@ class ShenandoahVerifyOopClosure : public BasicOopIterateClosure {
    */
   void verify_oops_from(oop obj) {
     _loc = obj;
-    obj->oop_iterate(this);
+    Klass* klass = obj->forward_safe_klass();
+    obj->oop_iterate_backwards(this, klass);
```
shipilev, line 309: Why `backwards`?

src/hotspot/share/gc/serial/markSweep.cpp
```diff
@@ -172,12 +172,12 @@ void MarkSweep::mark_object(oop obj) {
     _string_dedup_requests->add(obj);
   }
 
+  ContinuationGCSupport::transform_stack_chunk(obj);
```
shipilev, line 175: Add a comment, something like: "// Do the transform while we still have the header intact, which might include important class information".

shipilev (commented):
Another round!

src/hotspot/share/gc/shared/memAllocator.cpp
```diff
@@ -399,18 +399,23 @@ void MemAllocator::mem_clear(HeapWord* mem) const {
...
 oop MemAllocator::finish(HeapWord* mem) const {
   assert(mem != nullptr, "null object pointer");
-  // May be bootstrapping
-  oopDesc::set_mark(mem, markWord::prototype());
   // Need a release store to ensure array/class length, mark word, and
   // object zeroing are visible before setting the klass non-null, for
   // concurrent collectors.
-  oopDesc::release_set_klass(mem, _klass);
+  if (UseCompactObjectHeaders) {
+    oopDesc::release_set_mark(mem, _klass->prototype_header());
```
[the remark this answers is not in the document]
rkennke: I like _klass->prototype_header() more, and would argue that we should use that instead, here. An object's prototype mark really depends on the Klass of the object, with compact headers, and we would always get the correct prototype out of _klass->prototype_header(). Also, perhaps more importantly, the Klass::prototype_header() is useful because we can load it in generated code with a single instruction, while fetching the markWord::prototype() and amending it *at runtime* would require a whole sequence of instructions.

src/hotspot/share/runtime/globals.hpp
```diff
@@ -1060,7 +1064,7 @@ const int ObjectAlignmentInBytes = 8;
   product(bool, ErrorFileToStdout, false,                                   \
           "If true, error data is printed to stdout instead of a file")     \
                                                                             \
-  develop(bool, UseHeavyMonitors, false,                                    \
+  product(bool, UseHeavyMonitors, false,                                    \
```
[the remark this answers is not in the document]
rkennke: This slipped in for some testing. Will revert.

src/hotspot/share/gc/shared/collectedHeap.hpp
```diff
@@ -309,7 +309,7 @@ class CollectedHeap : public CHeapObj<mtGC> {
   }
 
   virtual void fill_with_dummy_object(HeapWord* start, HeapWord* end, bool zap);
-  static constexpr size_t min_dummy_object_size() {
+  static size_t min_dummy_object_size() {
```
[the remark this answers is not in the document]
rkennke: That's because oopDesc::header_size() can no longer be constexpr, because it depends on UseCompactObjectHeaders.

src/hotspot/share/gc/shared/collectedHeap.cpp
```diff
@@ -227,7 +227,9 @@ bool CollectedHeap::is_oop(oop object) const {
     return false;
   }
 
-  if (is_in(object->klass_raw())) {
+  // With compact headers, we can't safely access the class, due
+  // to possibly forwarded objects.
+  if (!UseCompactObjectHeaders && is_in(object->klass_raw())) {
```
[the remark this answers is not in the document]
rkennke: I have no idea *shrugs*

src/hotspot/share/gc/parallel/psOldGen.cpp
rkennke: Yes, but with compact headers, we could now have 1-word-sized objects, in which case this would fail. I am not sure how to deal with that, TBH. Maybe do the whole test only when !UseCompactObjectHeaders or when object-size is > 1?

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
rkennke: Because that's the only oop_iterate() variant that takes a Klass* :-) It seems trivial to add such a forward-variant, though. Prefer that?

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
shipilev: Yes, upstreaming this separately would be cleaner for this PR. It would also highlight any problems with adjusting the lifecycle for relativization of stack chunks.

src/hotspot/share/runtime/arguments.cpp
```diff
@@ -3116,6 +3116,29 @@ jint Arguments::finalize_vm_init_args(bool patch_mod_javabase) {
   UNSUPPORTED_OPTION(ShowRegistersOnAssert);
 #endif // CAN_SHOW_REGISTERS_ON_ASSERT
 
+#ifdef _LP64
+  if (!FLAG_IS_DEFAULT(UseCompactObjectHeaders)) {
```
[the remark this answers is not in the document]
rkennke: I've done this on purpose.
When the default for UseCompactObjectHeaders is false, then CDS archives will be written with legacy headers, and we could not read this when running with +UseCompactObjectHeaders.
When the default for UseCompactObjectHeaders is true, then CDS archives will be written with compact headers, and we could not read this when running with -UseCompactObjectHeaders.
I (and others) are changing the default of this flag regularily for testing, because that also catches tests that require flagless, and the way this is written, would not require changing this line in arguments.cpp too.

I guess it would be even more useful if we could detect which setting of the flag has been used when writing a CDS archive, and don't read it if it's not compatible.

It would be *even* better, if we could detect the setting of the flag when archive has been written, and transform it into whatever the JVM is running with, but that would be too much to ask for this PR, I think.

feilongjiang:
Hi, I'm trying to build JDK binary base on this pr on Linux-riscv64 (with GCC-11), but I got the following error:

```
=== Output from failing command(s) repeated here ===
* For target buildtools_interim_langtools_modules_java.compiler.interim__the.BUILD_java.compiler.interim_batch:
warning: unknown enum constant Feature.STRING_TEMPLATES
warning: unknown enum constant Feature.STRING_TEMPLATES
warning: unknown enum constant Feature.STRING_TEMPLATES
warning: unknown enum constant Feature.STRING_TEMPLATES
error: warnings found and -Werror specified
1 error
4 warnings

* All command lines available in /home/ubuntu/workspace/jdk/build/linux-riscv64-server-release/make-support/failure-logs.
=== End of repeated output ===
```

Here is my gcc info:
```
ubuntu@ubuntu-93:~/workspace/jdk$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/riscv64-linux-gnu/11/lto-wrapper
Target: riscv64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 11.3.0-1ubuntu1~22.04' --with-bugurl=file:///usr/share/doc/gcc-11/README.Bugs --enable-languages=c,ada,c++,go,d,fortran,objc,obj-c++,m2 --prefix=/usr --with-gcc-major-version-only --program-suffix=-11 --program-prefix=riscv64-linux-gnu- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-libitm --disable-libquadmath --disable-libquadmath-support --enable-plugin --enable-default-pie --with-system-zlib --enable-libphobos-checking=release --with-target-system-zlib=auto --enable-objc-gc=auto --enable-multiarch --disable-werror --disable-multilib --with-arch=rv64gc --with-abi=lp64d --enable-checking=release --build=riscv64-linux-gnu --host=riscv64-linux-gnu --target=riscv64-linux-gnu --with-build-config=bootstrap-lto-lean --enable-link-serialization=4
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 11.3.0 (Ubuntu 11.3.0-1ubuntu1~22.04) 
``

**Comment by shipilev** [comment]
> === Output from failing command(s) repeated here ===
> * For target buildtools_interim_langtools_modules_java.compiler.interim__the.BUILD_java.compiler.interim_batch:
> warning: unknown enum constant Feature.STRING_TEMPLATES
> warning: unknown enum constant Feature.STRING_TEMPLATES
> warning: unknown enum constant Feature.STRING_TEMPLATES
> warning: unknown enum constant Feature.STRING_TEMPLATES
> error: warnings found and -Werror specified

This is not a gcc problem, but a javac problem.

Please check your boot JDK is correct, that you have ran `make clean`, and that mainline JDK on the same machine does not fail with the same error?

File: src/hotspot/share/runtime/arguments.cpp (hunk 13 above)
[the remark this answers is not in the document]
Comment (shipilev) [reply src/hotspot/share/runtime/arguments.cpp:3120]: OK, I see. This makes sense.

**Comment by feilongjiang** [comment]
> > === Output from failing command(s) repeated here ===
> > 
> > * For target buildtools_interim_langtools_modules_java.compiler.interim__the.BUILD_java.compiler.interim_batch:
> >   warning: unknown enum constant Feature.STRING_TEMPLATES
> >   warning: unknown enum constant Feature.STRING_TEMPLATES
> >   warning: unknown enum constant Feature.STRING_TEMPLATES
> >   warning: unknown enum constant Feature.STRING_TEMPLATES
> >   error: warnings found and -Werror specified
> 
> This is not a gcc problem, but a javac problem.
> 
> Please check your boot JDK is correct, that you have ran `make clean`, and that mainline JDK on the same machine does not fail with the same error?

I have tried the same boot JDK with the mainline JDK code base, and it looks good.
P.S. I also tried cross-compiling with this PR, and the building passed without failures.

**Comment by shipilev** [comment]
> I have tried the same boot JDK with the mainline JDK code base, and it looks good. P.S. I also tried cross-compiling with this PR, and the building passed without failures.

OK, but this still points to incompatibility with boot/build JDK. The state of the source tree at this PR does not have any `STRING_TEMPLATES` symbol, it was added with JDK-8285932, so there is no way interim java.compiler in current PR references it, it must come from somewhere else. Do you have a self-built boot JDK that carries JDK-8285932? I'd suggest using a more stable JDK 20 as boot JDK.

**Comment by feilongjiang** [comment]
> > I have tried the same boot JDK with the mainline JDK code base, and it looks good. P.S. I also tried cross-compiling with this PR, and the building passed without failures.
> 
> 
> 
> OK, but this still points to incompatibility with boot/build JDK. The state of the source tree at this PR does not have any `STRING_TEMPLATES` symbol, it was added with JDK-8285932, so there is no way interim java.compiler in current PR references it, it must come from somewhere else. Do you have a self-built boot JDK that carries JDK-8285932? I'd suggest using a more stable JDK 20 as boot JDK.

Yes, the boot JDK I used was based on mainline JDK. That the problem here.

File: src/hotspot/share/gc/parallel/psPromotionManager.inline.hpp (hunk 5 above)
Comment (fisk) [inline src/hotspot/share/gc/parallel/psPromotionManager.inline.hpp:250]: For this check to work correctly, we are assuming that Copy::aligned_disjoint_words respects word level atomicity, even though we are using one of the non-atomic copying functions. That doesn't feel safe.

**Review by fisk — CHANGES_REQUESTED** [review]

File: src/hotspot/share/gc/parallel/psPromotionManager.inline.hpp (hunk 5 above)
Comment (shipilev) [reply src/hotspot/share/gc/parallel/psPromotionManager.inline.hpp:250]: True, it is not exactly safe. I wonder if we can plug this particular leak by doing the 

```
  if (UseCompactObjectHeaders) {
    // The copy above is not atomic. Make sure we have seen the proper mark
    // and re-install it into the copy, so that Klass* is guaranteed to be correct.
    markWord mark = o->mark_acquire();
    if (!mark.is_marked()) {
      new_obj->set_mark(mark);
      ContinuationGCSupport::transform_stack_chunk(new_obj);
    } else {
      // If we copied a mark-word that indicates 'forwarded' state, the object
      // installation would not succeed. We cannot access Klass* anymore either.
      // Skip the transformation.
    }
  } else {
    ContinuationGCSupport::transform_stack_chunk(new_obj);
  }
```

Comment (fisk) [reply src/hotspot/share/gc/parallel/psPromotionManager.inline.hpp:250]: The load in mark_acquire can float up above the copying. So I don't think that will work either.

Comment (rkennke) [reply src/hotspot/share/gc/parallel/psPromotionManager.inline.hpp:250]: Hmm, right. I guess this is not only about atomicity. It's also possible that we see that it's not marked/forwarded, then ignore the transform_stack_chunk() call, which would be wrong.
The problem is that transform_stack_chunk() wants to access the Klass* to check is_stackChunk(). So maybe we need to extract the Klass* from the test_mark and pass it to (a new variant of) ContinuationSupport::transform_stack_chunk() which only uses that class? That should work, right?

Comment (shipilev) [reply src/hotspot/share/gc/parallel/psPromotionManager.inline.hpp:250]: I don't quite see the problem. So the `mark_acquire` floats before the copy, what does it break? 

If we read the "forwarded" mark early, we know it would fail installation, because there is already a forwarding. No transformation is needed then.

If we read the "not forwarded" mark, then there a chance we are going to install the copy as forwardee, so we need to transform. But for the case of compact object headers, we need to guarantee for `transform_stack_chunk` is seeing a consistent `Klass*` from the mark, which specially atomic read of mark provides. We store that mark in the copy too, thus overwriting any atomicity violations the copy routine might have introduced, and giving `transform_chunk_call` a green light to proceed with transformation. (Aside: This thing could even be just `mark`, because even the relaxed atomic load would suffice.)

IOW, if the copy routine is not atomic, and it is a problem for mark, then we can overwrite mark in the copy with the special atomic read.

Comment (fisk) [reply src/hotspot/share/gc/parallel/psPromotionManager.inline.hpp:250]: > Hmm, right. I guess this is not only about atomicity. It's also possible that we see that it's not marked/forwarded, then ignore the transform_stack_chunk() call, which would be wrong.
> The problem is that transform_stack_chunk() wants to access the Klass* to check is_stackChunk(). So maybe we need to extract the Klass* from the test_mark and pass it to (a new variant of) ContinuationSupport::transform_stack_chunk() which only uses that class? That should work, right?

Yes, that would work. Given of course that nobody is reloading the klass somewhere in there.

Comment (fisk) [reply src/hotspot/share/gc/parallel/psPromotionManager.inline.hpp:250]: > I don't quite see the problem. So the `mark_acquire` floats before the copy, what does it break?
> 
> If we read the "forwarded" mark early, we know it would fail installation, because there is already a forwarding. No transformation is needed then.
> 
> If we read the "not forwarded" mark, then there a chance we are going to install the copy as forwardee, so we need to transform. But for the case of compact object headers, we need to guarantee for `transform_stack_chunk` is seeing a consistent `Klass*` from the mark, which specially atomic read of mark provides. We store that mark in the copy too, thus overwriting any atomicity violations the copy routine might have introduced, and giving `transform_chunk_call` a green light to proceed with transformation. (Aside: This thing could even be just `mark`, because even the relaxed atomic load would suffice.)
> 
> IOW, if the copy routine is not atomic, and it is a problem for mark, then we can overwrite mark in the copy with the value we got with special atomic load.

Ah. Yes, I think you are right.

File: src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp [hunk 14]
```diff
@@ -4314,9 +4314,37 @@ void MacroAssembler::load_method_holder(Register holder, Register method) {
...
 
+// Loads the obj's Klass* into dst.
+// Preserves all registers (incl src, rscratch1 and rscratch2).
+void MacroAssembler::load_nklass(Register dst, Register src) {
+  assert(UseCompressedClassPointers, "expects UseCompressedClassPointers");
+
+  if (!UseCompactObjectHeaders) {
+    ldrw(dst, Address(src, oopDesc::klass_offset_in_bytes()));
+    return;
+  }
```
shipilev, line 4325: Long-shot: maybe we should be doing `MacroAssembler::load_compact_klass` specifically for `UseCompactObjectHeaders`, and keep the legacy `UseCompressedClassPointers` paths in the callers. This would limit the exposure to the new code here.

src/hotspot/cpu/x86/c1_LIRAssembler_x86.cpp
```diff
@@ -3534,12 +3525,28 @@ void LIR_Assembler::emit_load_klass(LIR_OpLoadKlass* op) {
...
+    // Check if we can take the (common) fast path, if obj is unlocked.
+    __ movq(result, Address(obj, oopDesc::mark_offset_in_bytes()));
+    __ testb(result, markWord::monitor_value);
+    __ jcc(Assembler::notZero, *op->stub()->entry());
+    __ bind(*op->stub()->continuation());
+    // Fast-path: shift and decode Klass*.
+    __ shrq(result, markWord::klass_shift);
+    __ decode_klass_not_null(result, tmp);
+  } else
   if (UseCompressedClassPointers) {
```
shipilev, line 3542:
```suggestion
  } else if (UseCompressedClassPointers) {
```

src/hotspot/share/gc/parallel/psOldGen.cpp
shipilev: I think we should keep the original code semantics intact, and `guarantees` might need to check stuff even with zero offset.

```
// With compact headers, the objects can be one-word sized.
size_t int_off = UseCompactObjectHeaders ? MIN2(1, obj->size() - 1) : 1;
HeapWord* test_addr = cast_from_oop<HeapWord*>(obj) + int_off;
```

src/hotspot/share/gc/shared/collectedHeap.cpp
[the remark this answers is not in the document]
shipilev: I am dealing with this separately, this is not a concern for this PR.

src/hotspot/share/gc/shared/collectedHeap.hpp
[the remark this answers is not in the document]
shipilev: OK, fine.

src/hotspot/share/gc/shared/gc_globals.hpp
```diff
@@ -689,7 +689,7 @@
           range(128, NOT_LP64(512) LP64_ONLY(1024))                         \
           constraint(GCCardSizeInBytesConstraintFunc,AtParse)               \
                                                                             \
-  develop(bool, UseAltGCForwarding, false,                                  \
+  product(bool, UseAltGCForwarding, false,                                  \
```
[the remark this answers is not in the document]
shipilev: Comment is still open :)

src/hotspot/share/gc/shared/memAllocator.cpp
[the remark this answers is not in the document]
shipilev: OK, I see, makes sense.

src/hotspot/share/oops/objArrayKlass.inline.hpp
```diff
@@ -70,7 +70,8 @@ void ObjArrayKlass::oop_oop_iterate_elements_bounded(
 
 template <typename T, typename OopClosureType>
 void ObjArrayKlass::oop_oop_iterate(oop obj, OopClosureType* closure) {
-  assert (obj->is_array(), "obj must be array");
+  // We cannot safely access the Klass* with compact headers.
```
shipilev, line 73: "// In this assert, ...

src/hotspot/share/oops/oop.hpp
shipilev: OK

src/hotspot/share/oops/oop.inline.hpp
```diff
@@ -79,37 +83,81 @@ markWord oopDesc::cas_set_mark(markWord new_mark, markWord old_mark, atomic_memo
   return Atomic::cmpxchg(&_mark, old_mark, new_mark, order);
 }
 
+markWord oopDesc::resolve_mark() const {
+  assert(LockingMode != LM_LEGACY, "Not safe with legacy stack-locking");
+  markWord hdr = mark();
```
shipilev, line 88: Convention `hdr` -> `m`? Everywhere in new code in this file, I think.

shipilev (commented):
Another quick read.

src/hotspot/share/gc/parallel/psPromotionManager.inline.hpp
```diff
@@ -247,9 +247,21 @@ inline oop PSPromotionManager::copy_unmarked_to_survivor_space(oop o,
   // Copy obj
   Copy::aligned_disjoint_words(cast_from_oop<HeapWord*>(o), cast_from_oop<HeapWord*>(new_obj), new_obj_size);
 
-  // Parallel GC claims with a release - so other threads might access this object
-  // after claiming and they should see the "completed" object.
-  ContinuationGCSupport::transform_stack_chunk(new_obj);
+  if (UseCompactObjectHeaders) {
+    // The copy above is not atomic. Make sure we have seen the proper mark
+    // and re-install it into the copy, so that Klass* is guaranteed to be correct.
+    markWord mark = o->mark_acquire();
```
fisk, line 253: I don't think we need the acquire here, do we?

fisk (changes requested):

src/hotspot/share/gc/parallel/psPromotionManager.inline.hpp
rkennke: Right. An atomic load would be sufficient, which is what oopDesc::mark() already does. I change those code paths to use just mark().

src/hotspot/share/oops/oop.inline.hpp
```diff
@@ -79,37 +83,81 @@ markWord oopDesc::cas_set_mark(markWord new_mark, markWord old_mark, atomic_memo
...
+  } else
+#endif
+  {
     return _metadata._klass;
   }
 }
 
 Klass* oopDesc::klass_or_null() const {
-  if (UseCompressedClassPointers) {
+#ifdef _LP64
```
coleenp, line 126: I don't like all these #ifdef _LP64 here.  Maybe markWord.inline.hpp can be refactored to not require callers to have this conditional inclusion.

src/hotspot/share/runtime/arguments.cpp
```diff
@@ -3116,6 +3116,27 @@ jint Arguments::finalize_vm_init_args(bool patch_mod_javabase) {
...
+  }
+  if (UseCompactObjectHeaders && LockingMode == LM_LEGACY) {
+    FLAG_SET_DEFAULT(LockingMode, LM_LIGHTWEIGHT);
+  }
+  if (UseCompactObjectHeaders && !UseAltGCForwarding) {
+    FLAG_SET_DEFAULT(UseAltGCForwarding, true);
+  }
+  if (UseCompactObjectHeaders && !UseCompressedClassPointers) {
+    FLAG_SET_DEFAULT(UseCompressedClassPointers, true);
+  }
```
coleenp, line 3136: Make this a function like set_compact_object_headers_flags(), that checks for FLAG_IS_CMDLINE for the related options and give a warning for them too, and add a test.

coleenp (changes requested):
I don't have any comments on the compiler code, but some comments and questions.  Some of the LP64 preprocessor conditionals are inconsistent in the assembly code.

coleenp:
@rkennke Can you merge up with the GenerationalZGC changes because some of our test definitions need it.

src/hotspot/cpu/x86/c1_MacroAssembler_x86.cpp
```diff
@@ -163,24 +163,35 @@ void C1_MacroAssembler::try_allocate(Register obj, Register var_size_in_bytes, i
...
-  {
-    movptr(Address(obj, oopDesc::klass_offset_in_bytes()), klass);
+    {
+      movptr(Address(obj, oopDesc::klass_offset_in_bytes()), klass);
+    }
   }
 
   if (len->is_valid()) {
     movl(Address(obj, arrayOopDesc::length_offset_in_bytes()), len);
+    if (UseCompactObjectHeaders) {
```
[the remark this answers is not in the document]
rkennke: Ok I will put it in _LP64 (even though it is not strictly needed - UseCompactObjectHeaders is hard-wired constant false, so compiler will not include the code, I would expect), but why not check UseCompactObjectHeaders here? The new code is only sensible with compact headers.

src/hotspot/share/gc/parallel/psPromotionManager.cpp
```diff
@@ -290,7 +290,7 @@ void PSPromotionManager::process_array_chunk(PartialArrayScanTask task) {
   assert(PSChunkLargeArrays, "invariant");
 
   oop old = task.to_source_array();
-  assert(old->is_objArray(), "invariant");
+  assert(old->forward_safe_klass()->is_objArray_klass(), "invariant");
```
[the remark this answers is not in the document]
rkennke: There are situations in GC where the object can be forwarded or not. This is where those methods are useful and required for compact-headers, because the mark-word and therefore Klass* could only be loaded from the forwardee. Originally I put this code in places where it happened in each GC, but eventually realized that it is a somewhat common pattern, so I put it in oopDesc. But we need to be careful there, because we have two different ways to deal with forwarding full-GC uses sliding-forwarding (which preserves the Klass* in the mark-word) and normal-GCs use the normal-forwarding which only preserves the Klass* in the forwardee. The forward_safe_* methods only work on the latter, and are only required in the latter. But that is not very difficult to sort out, IMO, because full-GCs are in fully separate code paths anyway.

src/hotspot/share/oops/klass.cpp
```diff
@@ -195,11 +195,22 @@ void* Klass::operator new(size_t size, ClassLoaderData* loader_data, size_t word
...
+static markWord make_prototype(Klass* kls) {
+  markWord prototype = markWord::prototype();
+#ifdef _LP64
+  if (UseCompactObjectHeaders) {
+    prototype = prototype.set_klass(kls);
+  }
+#endif
+  return prototype;
+}
+
```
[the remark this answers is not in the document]
rkennke: Why? This code used to be there with BiasedLocking, and has been removed. I've re-instated it for compact object headers, because the prototype mark for an object now depends on its Klass, but other than that, why would it be useful? The prototype would be just markWord::prototype().

src/hotspot/share/oops/objArrayKlass.cpp
```diff
@@ -156,7 +156,8 @@ ObjArrayKlass::ObjArrayKlass(int n, Klass* element_klass, Symbol* name) : ArrayK
 }
 
 size_t ObjArrayKlass::oop_size(oop obj) const {
-  assert(obj->is_objArray(), "must be object array");
+  // In this assert, we cannot safely access the Klass* with compact headers.
+  assert(UseCompactObjectHeaders || obj->is_objArray(), "must be object array");
```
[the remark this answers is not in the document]
rkennke: Yes, there is. We're a bit excessive with asserting the klass here. I tried to remain as close as possible with that, so I disabled it only for compact object headers.

src/hotspot/share/oops/oop.inline.hpp
rkennke: The problem is with 32bits, in markWord, we only have 32bits in the header, and no place to stick in the Klass* in the upper 32 bits. That's why I put all those #ifdefs, there.
If you take a step back, you'll notice that compact object headers mostly aligns the layout headers of 64bit and 32bit JVMs. There would be a great opportunity here to consolidate all this code, make the whole header a union/struct/bitfield that looks the same both on 32bit and 64bit builds. But this conflicts with the current implementation where we want to be able to switch between compact and legacy header layout.

Also, going forward, we want to shrink the header even more to just 32bits, and still have it switchable with the old layout. Eventually all this stuff will be the same in 32bit and 64bit JVMs, but for the time being I think we need to keep it slightly messy to support the legacy layout.

rkennke:
I'm sorry, I think I butchered this PR while trying to merge latest upstream through all the dependent PRs. Let's continue the discussion the new PR #13961. I hope I haven't caused anything breakage (for some reason, this PR now shows as "Merged" which worries me. I believe the Skara bot did that. I wonder where it has been merged to.)

src/hotspot/share/oops/klass.cpp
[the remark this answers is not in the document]
coleenp: I thought not to waste a 64 bit field in Klass and to maybe eliminate some if (CompactObjectHeaders) use the one in Klass else use MarkWord::prototype(), just always use the one in Klass.  Minimizing if (CompactObjectHeaders) would be a good thing. At any case, this isn't for this change, just an idea to try to use this field unconditionally.  I recognize it from BiasedLocking.

PR #13851, opened 2023-05-06
8307523: [vectorapi] Optimize MaskFromLongBenchmark.java

changpeng1997:
To avoid dead code elimination, a use-point laneIsSet() is added in each benchmark method in MaskFromLongBenchmark.java.

However, currently laneIsSet() [1] is implemented by toLong(). So it may generate a fromLong-toLong pair [2], making this benchmark to be noneffective after inlining laneIsSet() into the outer method. The assembly of maskFromLong_byte128 benchmark on SVE2 is shown in [3]. We cannot see the bdep instruction used by fromLong on AArch64 [4]. So, in this case, we cannot measure fromLong()'s performance by using this benchmark.

This patch uses trueCount() [5] instead of toLong() to measure the fromLong()'s performance effectively. After this patch, we can see the bdep instruction in the hot loop [6] of maskFromLong_byte128 benchmark.

Since using Blackhole to consume VectorMask will generate a heavy vector box, we don't use Blackhole to fix this benchmark.

[1]: https://github.com/openjdk/jdk/blob/96fa2751e8bbc05d6d064d80c07720cc9db05c54/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/AbstractMask.java#L70
[2]: https://github.com/openjdk/jdk/blob/ff368d504e9101e11c7182185f56255f429d31e3/src/hotspot/share/opto/vectornode.cpp#L1736
[3]: https://gist.github.com/changpeng1997/467f6056f78d99c055030fa5888b6baa
[4]: https://github.com/openjdk/jdk/blob/787832a58677205c9a11ae100dd8a2fbddb30a4a/src/hotspot/cpu/aarch64/c2_MacroAssembler_aarch64.cpp#L1099
[5]: https://docs.oracle.com/en/java/javase/16/docs/api/jdk.incubator.vector/jdk/incubator/vector/VectorMask.html#trueCount()
[6]: https://gist.github.com/changpeng1997/79bea0a9f80530bec89978950897000d

---------
### Progress

### Issue
 * [JDK-8307523](https://bugs.openjdk.org/browse/JDK-8307523): [vectorapi] Optimize MaskFromLongBenchmark.java

### Reviewers
 * [Quan Anh Mai](https://openjdk.org/census#qamai) (@merykitty - Committer)
 * [Xiaohong Gong](https://openjdk.org/census#xgong) (@XiaohongGong - Committer)
 * [Nick Gasson](https://openjdk.org/census#ngasson) (@nick-arm - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13851/head:pull/13851` \
`$ git checkout pull/13851`

`$ git checkout pull/13851` \
`$ git pull https://git.openjdk.org/jdk.git pull/13851/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13851`

`$ git pr show -t 13851`

</details>

<a href="https://git.openjdk.org/jdk/pull/13851.diff">https://git.openjdk.org/jdk/pull/13851.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13851#issuecomment-1536984510)

### DIFF
--- a/test/micro/org/openjdk/bench/jdk/incubator/vector/MaskFromLongBenchmark.java
+++ b/test/micro/org/openjdk/bench/jdk/incubator/vector/MaskFromLongBenchmark.java
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
+// Copyright (c) 2021, 2023, Oracle and/or its affiliates. All rights reserved.
 // DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
 //
 // This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


merykitty:
Storing into a boolean array should be safer as `trueCount` can be implemented as `bitCount(toLong())`. Thanks.

changpeng1997:
> Storing into a boolean array should be safer as `trueCount` can be implemented as `bitCount(toLong())`. Thanks.

I think intoArray() may have heavier cost.

merykitty:
@changpeng1997 Yes that's right, I think ideally we can make `Blackhole` receives a `VectorNode`, but this patch is good as it is

merykitty (approved):

changpeng1997:
> @changpeng1997 Yes that's right, I think ideally we can make `Blackhole` receive a `VectorNode`, but this patch is good as it is

Thanks for your review.

changpeng1997:
> @changpeng1997 Yes that's right, I think ideally we can make `Blackhole` receive a `VectorNode`, but this patch is good as it is

I have updated commit message to explain why we should not use blackhole to fix this benchmark.

XiaohongGong (approved):
Looks good to me!

changpeng1997:
@jatin-bhateja Could you please help to review this patch? Thanks.

nick-arm (approved):

PR #13854, opened 2023-05-06
8307567: Avoid relocating global roots to metaspaceObjs in CDS dump

iklam:
The functions `ArchiveBuilder::relocate_roots()` and `ArchiveBuilder::relocate_vm_classes()` relocate global roots such as `Universe::_typeArrayKlassObjs[]` and `vmClasses::_klasses[]` to point to locations inside the CDS "buffer" space.

As a result, the VM is no longer usable after CDS dumping has finished.

Previously, we tried to save some of the VM states and restore them after CDS dump has finished. See [JDK-8264735](https://bugs.openjdk.org/browse/JDK-8264735), which added functions like `FileMapInfo::clone_shared_path_table()`.

However, this is too cumbersome and doesn't cover all cases. The intention of the global root relocation is to make it easy to write the roots using `WriteClosure::do_ptr()`, which could handle only pointers into the "buffer" space. If we call `get_buffered_addr()` inside `WriteClosure::do_ptr()`, then there's no need to modify the global roots.

Notes for reviewers:

- Start with changes in `WriteClosure::do_ptr()`
- Some code in SystemDictionaryShared also needed to call `get_buffered_addr()`
- I added new function `ArchiveBuilder::write_pointer_in_buffer()` to simplify writing pointers into the CDS buffer
-  `ArchiveBuilder::relocate_roots()`, `ArchiveBuilder::relocate_vm_classes()` and `FileMapInfo::clone_shared_path_table()` are removed.
    - I am leaving `CloneDumpTimeClassTable` and `CloneDumpTimeLambdaProxyClassTable` for now. I plan to remove them in a follow up RFE after more testing.

---------
### Progress

### Issue
 * [JDK-8307567](https://bugs.openjdk.org/browse/JDK-8307567): Avoid relocating global roots to metaspaceObjs in CDS dump

### Reviewers
 * [Matias Saavedra Silva](https://openjdk.org/census#matsaave) (@matias9927 - Committer) ⚠️ Review applies to [d304dde7](https://git.openjdk.org/jdk/pull/13854/files/d304dde7a5ec182b8ba8221bbc070f77cc17272a)
 * [Calvin Cheung](https://openjdk.org/census#ccheung) (@calvinccheung - **Reviewer**) ⚠️ Review applies to [5a1c2df1](https://git.openjdk.org/jdk/pull/13854/files/5a1c2df11b0e557a037125151b4eb9bfe6938290)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13854/head:pull/13854` \
`$ git checkout pull/13854`

`$ git checkout pull/13854` \
`$ git pull https://git.openjdk.org/jdk.git pull/13854/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13854`

`$ git pr show -t 13854`

</details>

<a href="https://git.openjdk.org/jdk/pull/13854.diff">https://git.openjdk.org/jdk/pull/13854.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13854#issuecomment-1537186761)


src/hotspot/share/cds/archiveBuilder.cpp
```diff
@@ -461,8 +454,16 @@ bool ArchiveBuilder::gather_one_source_obj(MetaspaceClosure::Ref* enclosing_ref,
   }
 }
 
-void ArchiveBuilder::remember_embedded_pointer_in_copied_obj(MetaspaceClosure::Ref* enclosing_ref,
-                                                             MetaspaceClosure::Ref* ref) {
+void ArchiveBuilder::remember_embedded_pointer_in_gathered_obj(address src_obj, address* field_addr) {
+  // field_addr must be a valid field inside src_obj.
```
matias9927, line 458: Is there a way to check this with an assert?

src/hotspot/share/classfile/systemDictionaryShared.cpp
```diff
@@ -1166,22 +1166,19 @@ class AdjustLambdaProxyClassInfo : StackObj {
...
-        assert(ArchiveBuilder::current()->is_in_buffer_space(ok0), "must be");
-        assert(ArchiveBuilder::current()->is_in_buffer_space(ok1), "must be");
-        InstanceKlass* bk0 = ok0;
-        InstanceKlass* bk1 = ok1;
-        assert(bk0->next_link() == 0, "must be called after Klass::remove_unshareable_info()");
-        assert(bk1->next_link() == 0, "must be called after Klass::remove_unshareable_info()");
-        bk0->set_next_link(bk1);
-        bk1->set_lambda_proxy_is_available();
-        ArchivePtrMarker::mark_pointer(bk0->next_link_addr());
+    InstanceKlass* last_buff_k = nullptr;
```
matias9927, line 1169: Nice, this is a much cleaner approach!

src/hotspot/share/memory/iterator.hpp
```diff
@@ -355,6 +355,10 @@ class SerializeClosure : public Closure {
   bool writing() {
     return !reading();
   }
+
+  // Useful aliases
```
matias9927, line 359: This is a nice addition but it might be buried under the rest of this PR. Would it be worth making this a separate enhancement?

matias9927 (approved):
I just have a few questions but otherwise this looks good!

src/hotspot/share/cds/archiveBuilder.cpp
```diff
@@ -471,22 +472,27 @@ void ArchiveBuilder::remember_embedded_pointer_in_copied_obj(MetaspaceClosure::R
...
+
 void ArchiveBuilder::gather_source_objs() {
   ResourceMark rm;
   log_info(cds)("Gathering all archivable objects ... ");
   gather_klasses_and_symbols();
   GatherSortedSourceObjs doit(this);
-  iterate_sorted_roots(&doit, /*is_relocating_pointers=*/false);
+  iterate_sorted_roots(&doit);
   doit.finish();
+  FileMapInfo::remember_embedded_pointers();
```
calvinccheung, line 495: What is the above call for?
I thought `FileMapInfo::metaspace_pointers_do()` already called from `iterate_roots()` which is called from `iterate_sorted_roots()`.

src/hotspot/share/cds/archiveBuilder.hpp
```diff
@@ -241,22 +241,21 @@ class ArchiveBuilder : public StackObj {
...
-  void iterate_sorted_roots(MetaspaceClosure* it, bool is_relocating_pointers);
+  void iterate_sorted_roots(MetaspaceClosure* it);
   void sort_symbols_and_fix_hash();
   void sort_klasses();
   static int compare_symbols_by_address(Symbol** a, Symbol** b);
   static int compare_klass_by_name(Klass** a, Klass** b);
 
   void make_shallow_copies(DumpRegion *dump_region, const SourceObjList* src_objs);
   void make_shallow_copy(DumpRegion *dump_region, SourceObjInfo* src_info);
-
```
calvinccheung, line 252: Blank line removed by accident?

calvinccheung (approved):
I have a question and a comment. Looks good otherwise.

src/hotspot/share/memory/iterator.hpp
iklam: I filed [JDK-8307959](https://bugs.openjdk.org/browse/JDK-8307959): "Remove explicit type casts from SerializeClosure::do_xxx() calls"

 I will integrate that before integrate this PR.

src/hotspot/share/cds/archiveBuilder.cpp
iklam: The assert is done inside `ArchiveBuilder::SourceObjList::remember_embedded_pointer()`, which knows the size of `src_obj`:

```
assert(field_offset_in_bytes >= 0, "must be");
assert(field_offset_in_bytes + intx(sizeof(intptr_t)) <= intx(src_obj_size), "must be");
```

src/hotspot/share/cds/archiveBuilder.cpp
iklam: This call was a hack. It was necessary because `SharedClassPathEntry` was not a regular `MetaspaceObj` type. 

I thought rewriting `SharedClassPathEntry` to subclass from `MetaspaceObj` would be too complicated, and wanted to delay that in a future REF. However, it turned out to be fairly straight-forward. I was able to further simplify the code. See commit [46aaec7](https://github.com/openjdk/jdk/pull/13854/commits/46aaec77036982e1454203e8466ccf04d5ff3575)

Now the pointers inside them are remembered during `ArchiveBuilder::gather_source_objs()`, just like other types of objects.

src/hotspot/share/cds/archiveBuilder.hpp
iklam: Removed.

calvinccheung (approved):
Thanks for making `SharedClassPathEntry` a regular `MetaspaceObj`. It looks a lot cleaner now.

iklam:
Thanks @calvinccheung and @matias9927 for the review.
Passed tiers 1-4 and builds-tier5
/integrate

PR #13860, opened 2023-05-08
JDK-8133773: clarify specification of Spliterator.tryAdvance

viktorklang-ora:
Attempting to make the "at-most once" nature, of invoking the supplied action to tryAdvance, clearer in the JavaDoc.

---------
### Progress

### Issue
 * [JDK-8133773](https://bugs.openjdk.org/browse/JDK-8133773): clarify specification of Spliterator.tryAdvance

### Reviewers
 * [Tagir F. Valeev](https://openjdk.org/census#tvaleev) (@amaembo - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13860/head:pull/13860` \
`$ git checkout pull/13860`

`$ git checkout pull/13860` \
`$ git pull https://git.openjdk.org/jdk.git pull/13860/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13860`

`$ git pr show -t 13860`

</details>

<a href="https://git.openjdk.org/jdk/pull/13860.diff">https://git.openjdk.org/jdk/pull/13860.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13860#issuecomment-1538156663)

### DIFF
--- a/src/java.base/share/classes/java/util/Spliterator.java
+++ b/src/java.base/share/classes/java/util/Spliterator.java
@@ -295,7 +295,7 @@ import java.util.function.LongConsumer;
  */
 public interface Spliterator<T> {
     /**
-     * If a remaining element exists, performs the given action on it,
+     * If a remaining element exists: performs the given action on it,
      * returning {@code true}; else returns {@code false}.  If this
      * Spliterator is {@link #ORDERED} the action is performed on the
      * next element in encounter order.  Exceptions thrown by the
@@ -304,7 +304,7 @@ public interface Spliterator<T> {
      * Subsequent behavior of a spliterator is unspecified if the action throws
      * an exception.
      *
-     * @param action The action
+     * @param action The action which will be invoked at-most once
      * @return {@code false} if no remaining elements existed
      * upon entry to this method, else {@code true}.
      * @throws NullPointerException if the specified action is null
--- a/src/java.base/share/classes/java/util/Spliterator.java
+++ b/src/java.base/share/classes/java/util/Spliterator.java
@@ -304,7 +304,7 @@ public interface Spliterator<T> {
      * Subsequent behavior of a spliterator is unspecified if the action throws
      * an exception.
      *
-     * @param action The action which will be invoked at-most once
+     * @param action The action whose operation is performed at-most once
      * @return {@code false} if no remaining elements existed
      * upon entry to this method, else {@code true}.
      * @throws NullPointerException if the specified action is null


src/java.base/share/classes/java/util/Spliterator.java
```diff
@@ -304,7 +304,7 @@
      * Subsequent behavior of a spliterator is unspecified if the action throws
      * an exception.
      *
-     * @param action The action
+     * @param action The action which will be invoked at-most once
```
PaulSandoz, line 307: If you want to align with the language specified in `Consumer` i recommend:
```suggestion
     * @param action The action whose operation is performed at-most once
```

viktorklang-ora:
@PaulSandoz Great suggestion, thank you Paul!

PaulSandoz (approved):

amaembo (approved):

src/java.base/share/classes/java/util/Spliterator.java
```diff
@@ -295,7 +295,7 @@
  */
 public interface Spliterator<T> {
     /**
-     * If a remaining element exists, performs the given action on it,
+     * If a remaining element exists: performs the given action on it,
```
RogerRiggs, line 298: I'm not sure the ":" is an improvement on the ",". 
"If the remaining element exists, the action is performed on it, ..."

viktorklang-ora: @RogerRiggs "the" instead of "a" gives me the impression that the Spliterator only has a single element left. The choice of colon instead of a comma is because "If a remaining element exists" introduces a list of 2 elements (and introducing a list is a good place for a colon) whereas the "else returns {@code false}" is a closely connected sentence to the topic of the list, which seems like a place to use the semi-colon.

But, not being a native English-speaker, I will happily change it to something which is more correct. :)

RogerRiggs: The switch from "a" to "the" was unintentional.  
"If a remaining element exists, the action is performed on it, returning true;..."

RogerRiggs (approved):

src/java.base/share/classes/java/util/Spliterator.java
viktorklang-ora: @RogerRiggs Got it. :)

PR #13873, opened 2023-05-08
8307486: ProcessTools.java should wait until vthread is completed before checking exceptions

lmesnik:
Updated processtools to check exception after join().

Tested with running CI virtual thread tests.

---------
### Progress

### Issue
 * [JDK-8307486](https://bugs.openjdk.org/browse/JDK-8307486): ProcessTools.java should wait until vthread is completed before checking exceptions

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13873/head:pull/13873` \
`$ git checkout pull/13873`

`$ git checkout pull/13873` \
`$ git pull https://git.openjdk.org/jdk.git pull/13873/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13873`

`$ git pr show -t 13873`

</details>

<a href="https://git.openjdk.org/jdk/pull/13873.diff">https://git.openjdk.org/jdk/pull/13873.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13873#issuecomment-1539187140)

### DIFF
--- a/test/lib/jdk/test/lib/process/ProcessTools.java
+++ b/test/lib/jdk/test/lib/process/ProcessTools.java
@@ -895,10 +895,10 @@ public final class ProcessTools {
                         tg.uncaughtThrowable = error;
                     }
                 });
-            if (tg.uncaughtThrowable != null) {
-                throw new RuntimeException(tg.uncaughtThrowable);
-            }
             vthread.join();
+            if (tg.uncaughtThrowable != null) {
+                throw tg.uncaughtThrowable;
+            }
         } else if (wrapper.equals("Kernel")) {
             MainThreadGroup tg = new MainThreadGroup();
             Thread t = new Thread(tg, () -> {
@@ -913,7 +913,7 @@ public final class ProcessTools {
             t.start();
             t.join();
             if (tg.uncaughtThrowable != null) {
-                throw new RuntimeException(tg.uncaughtThrowable);
+                throw tg.uncaughtThrowable;
             }
         } else {
             mainMethod.invoke(null, new Object[] { classArgs });


test/lib/jdk/test/lib/process/ProcessTools.java
```diff
@@ -895,10 +895,10 @@ public static void main(String[] args) throws Throwable {
                         tg.uncaughtThrowable = error;
                     }
                 });
+            vthread.join();
             if (tg.uncaughtThrowable != null) {
-                throw new RuntimeException(tg.uncaughtThrowable);
```
dholmes-ora, line 899: I think the wrapping with RuntimeException should be preserved because the uncaughtThrowable was thrown in a different thread and we want an exception that is thrown in the current thread.

dholmes-ora (changes requested):
Moving the join() is fine but I don't think the other changes are wanted.

Thanks.

test/lib/jdk/test/lib/process/ProcessTools.java
lmesnik: I removed 'new RuntimeException' to make the exception chain more similar to the original one. This exception is thrown by method main() and not going to be handled but printed only. The test which tries to search the exception in the output of a spawned process by classname might be confused if found 'RuntimeException' instead of the expected one.

dholmes-ora: Surely any such tests would have failed by now if that were the case? Propagating an exception in one thread when the exception was actually generated and thrown in another is normally considered a source of confusion.

lmesnik: Really, I am not sure there are currently tests that search the exception name in process output, but they would fail because of wrapping exceptions with RuntimeException(...).
Please note, that thread factory mode. The ProcessTools.main() in this tries to run the main()method of test in another thread. Agree for human might be more confusing, however in this mode is should be something expected. However, I just to reduce positive false positive failures related to changed exception name.

If the test normally throws an exception then it is printed like:
Exception in thread "main" java.lang.RuntimeException: java.lang.Error: Expected
instead the original
Exception in thread "main" java.lang.Error: Expected
and it might confuse the output parser. The goal is to return to the original printing.

dholmes-ora: If the parser is looking for `java.lang.Error: Expected` via `output.shouldContain` then it will still work fine.

But okay

dholmes-ora (approved):

AlanBateman (approved):

PR #13876, opened 2023-05-09
8047998: Abort the vm if MaxNewSize is not the same as NewSize when MaxHeapSize is the same as InitialHeapSize

lgxbslgx:
Hi all,

When `MaxHeapSize` is equal to `InitialHeapSize` and `NewSize` is set in command line,
currently, the `max_young_size/MaxNewSize` will be set to the value of `NewSize`.

Considering the document of the `NewSize` (shown below), someone may set the `NewSize` 
to a very small value and expect the JVM to adjust the value dynamically.
Then when the `MaxHeapSize` is equal to `InitialHeapSize` (set by user or ergonomics),
the `MaxNewSize` is set to the value of `NewSize`, which is small unexpectedly.

```
  product(size_t, NewSize, ScaleForWordSize(1*M),                           \
          "Initial new generation size (in bytes)")                         \
          constraint(NewSizeConstraintFunc,AfterErgo)                       \
```

This patch fixes the issue by setting the `MaxNewSize` to `NewSize` only when the `NewSize` 
is larger than the original `max_young_size/MaxNewSize`.

The title of JDK-8047998 may need to adjusted.

Thanks for the review.

Best Regards,
-- Guoxiong

---------
### Progress
- [x] Change requires CSR request [JDK-8309304](https://bugs.openjdk.org/browse/JDK-8309304) to be approved

### Issues
 * [JDK-8047998](https://bugs.openjdk.org/browse/JDK-8047998): Abort the vm if MaxNewSize is not the same as NewSize when MaxHeapSize is the same as InitialHeapSize (**Bug** - P4)
 * [JDK-8309304](https://bugs.openjdk.org/browse/JDK-8309304): Abort the VM if MaxNewSize is not the same as NewSize when MaxHeapSize is the same as InitialHeapSize (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13876/head:pull/13876` \
`$ git checkout pull/13876`

`$ git checkout pull/13876` \
`$ git pull https://git.openjdk.org/jdk.git pull/13876/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13876`

`$ git pr show -t 13876`

</details>

<a href="https://git.openjdk.org/jdk/pull/13876.diff">https://git.openjdk.org/jdk/pull/13876.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13876#issuecomment-1539336502)


albertnetymk:
Can't say which behavior is "more" unexpected -- `NewSize` means the initial young-size at start-up. Assuming one prints the young-size at start-up, the developer might be surprised if the printed value is diff from `NewSize`.

lgxbslgx:
> Can't say which behavior is "more" unexpected -- `NewSize` means the initial young-size at start-up. Assuming one prints the young-size at start-up, the developer might be surprised if the printed value is diff from `NewSize`.

Another solution: only change the document of the corresponding flags to remind the user and don't change the code.
The draft patch:

```
diff --git a/src/hotspot/share/gc/shared/gc_globals.hpp b/src/hotspot/share/gc/shared/gc_globals.hpp
index 1bf74a81706..3ffe8ddeb74 100644
--- a/src/hotspot/share/gc/shared/gc_globals.hpp
+++ b/src/hotspot/share/gc/shared/gc_globals.hpp
@@ -604,7 +604,12 @@
                                                                             \
   product(size_t, MaxNewSize, max_uintx,                                    \
           "Maximum new generation size (in bytes), max_uintx means set "    \
-          "ergonomically")                                                  \
+          "ergonomically."                                                  \
+          "If the InitialHeapSize is equal to MaxHeapSize "                 \
+          "(either of them may be set at command line or "                  \
+          "be set ergonomically), and the NewSize is set at command line, " \
+          "MaxNewSize would be set to the value of NewSize, "               \
+          "even the MaxNewSize is also set at command line.")               \
           range(0, max_uintx)                                               \
                                                                             \
   product_pd(size_t, HeapBaseMinAddress,                                    \

```

What do you think about it?

albertnetymk:
Could you provide a concrete example (specific JVM flags and values) to illustrate the problem (unexpected behavior due to inadequate/misleading/confusing doc)? I don't get what the problem is, even after reading the original ticket description and your revised solution.

albertnetymk:
Thank you for breaking down the problem.

It's important to note that `-XX:InitialRAMPercentage=1.5625` is also included implicitly.

I can't say I'm entirely surprised by the following:

`-XX:MaxHeapSize=256M -XX:NewSize=1M -XX:MaxNewSize=80M` + `-XX:InitialRAMPercentage=1.5625` on a large RAM system leads to a constant young-size.

(Of course, one could argue that the default value of `InitialRAMPercentage` is surprising. Would `0` be any better? After all, its counterpart, `InitialHeapSize`, has `0` as the default value.)

albertnetymk:
As far as I understand it, there are potentially two issues here:

1. `InitialRAMPercentage` and `InitialHeapSize` control the same attribute but they have diff default value.

2. `MaxNewSize` can be silently ignored in certain cases. For example:

```
java -XX:+UseSerialGC -XX:InitialHeapSize=256m -XX:MaxHeapSize=256M -XX:NewSize=8M -XX:MaxNewSize=80M '-Xlog:gc,gc+heap=trace' --version
```

Partial output: `Initial young 8388608  Maximum young 8388608` meaning that `MaxNewSize=80M` is silently ignored/discarded/overwritten, which can be surprising, especially when `-XX:InitialHeapSize=256m` (or even larger value) is derived from the default and implicit `InitialRAMPercentage`.

Regardless of issue 1, maybe emitting a warning/error would be less surprising in the case of issue 2. (Your concern is mostly issue 2, right?)

lgxbslgx:
> Regardless of issue 1, maybe emitting a warning/error would be less surprising in the case of issue 2. (Your concern is mostly issue 2, right?)

Yes, the main problem is `issue 2`.

lgxbslgx:
> OK, maybe sth along these lines:
> 
> ```c++
> if (MaxHeapSize == InitialHeapSize) {
>   if (FLAG_IS_CMDLINE(NewSize) && FLAG_IS_CMDLINE(MaxNewSize) && NewSize != MaxNewSize) {
>     vm_exit_during_initialization(...);
>   }
>   ...
> }
> ```
> 
> I think exiting here is fine because this is an impossible constraint to satisfy, variable young-gen size in constant whole-heap setup. What do you think?

The condition `MaxHeapSize == InitialHeapSize` is often meet, I don't think it is good to abort the vm. I do worry that such exiting may break some current online systems, which is unacceptable for us (Do nothing is better to break the online systems). Considering the JDK21 is a LTS version, I don't want to do such breaking change.

albertnetymk:
> The condition MaxHeapSize == InitialHeapSize is often meet

True, but additionally specifying both `NewSize` and `MaxNewSize` is uncommon, IMO.

> I do worry that such exiting may break some current online systems

Actually, I think it's desirable to fail loudly if the JVM cmd flags contain errors. Hard to say whether this is an error or not, but debugging suboptimal performance (from incorrect heap/generation-size) is much harder. Therefore, notifying developers at jvm-startup when cmd flags can't be satisfied might be appreciated. Again, this setup should be rare.

(I agree this is subjective.)

lgxbslgx:
> > The condition MaxHeapSize == InitialHeapSize is often meet
> 
> True, but additionally specifying both `NewSize` and `MaxNewSize` is uncommon, IMO.
> 
> > I do worry that such exiting may break some current online systems
> 
> Actually, I think it's desirable to fail loudly if the JVM cmd flags contain errors. Hard to say whether this is an error or not, but debugging suboptimal performance (from incorrect heap/generation-size) is much harder. Therefore, notifying developers at jvm-startup when cmd flags can't be satisfied might be appreciated. Again, this setup should be rare.
> 
> (I agree this is subjective.)

I will be ok with your previous suggestion (exiting the vm) if this PR is intergrated into JDK22, which means we should integrate it into mainline after June 08, 2023 [1] . What do you think about it?

[1] https://openjdk.org/projects/jdk/21/

albertnetymk (approved):

tschatzl (approved):

lgxbslgx:
@albertnetymk Thanks for creating the CSR. It may be good to mark it as `PROPOSED` now.

albertnetymk:
@lgxbslgx The CSR has been approved.

lgxbslgx:
@albertnetymk @tschatzl Thanks for your reviews. Integrating.

/integrate

dcubed-ojdk:
Two tests are failing many times in Tier2 after [JDK-8047998](https://bugs.openjdk.org/browse/JDK-8047998): 

  10 gc/arguments/TestNewSizeFlags.java 
  10 gc/arguments/TestVerifyBeforeAndAfterGCFlags.java 

so 20 failures per Tier2. 

Three tests are failing many times in Tier3 after [JDK-8047998](https://bugs.openjdk.org/browse/JDK-8047998): 

   4 compiler/runtime/Test8010927.java 
   6 jdk/jfr/event/gc/detailed/TestPromotionFailedEventWithDefNew.java 
   6 jdk/jfr/event/gc/detailed/TestPromotionFailedEventWithParallelScavenge.java 

so 16 failures per Tier3.

This fix has been backed out via:

[JDK-8311215](https://bugs.openjdk.org/browse/JDK-8311215) [BACKOUT] JDK-8047998 Abort the vm if MaxNewSize is not the same as NewSize when MaxHeapSize is the same as InitialHeapSize

and the following bug created to track the failures caused by the original
JDK-8047998 and to use as a [REDO]:

[JDK-8311214](https://bugs.openjdk.org/browse/JDK-8311214) [REDO] JDK-8047998 Abort the vm if MaxNewSize is not the same as NewSize when MaxHeapSize is the same as InitialHeapSize

dcubed-ojdk:
Before this work is reintegrated, please make sure that someone runs the
revised fix thru Mach5 Tier2 and Tier3 testing (at a minimum).

PR #13878, opened 2023-05-09
8307653: Adjust delay time and gc log argument in TestAbortOnVMOperationTimeout

lgxbslgx:
Hi all,

This patch increases the delay time of the test `TestAbortOnVMOperationTimeout` 
so that it can pass on the low performance devices (such as my local riscv64 dev board).

And I adjust the gc log options to get more gc log message which is useful to
inspect the issue [JDK-8296819](https://bugs.openjdk.org/browse/JDK-8296819).

Thanks for the review.

Best Regards,
-- Guoxiong

---------
### Progress

### Issue
 * [JDK-8307653](https://bugs.openjdk.org/browse/JDK-8307653): Adjust delay time and gc log argument in TestAbortOnVMOperationTimeout

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13878/head:pull/13878` \
`$ git checkout pull/13878`

`$ git checkout pull/13878` \
`$ git pull https://git.openjdk.org/jdk.git pull/13878/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13878`

`$ git pr show -t 13878`

</details>

<a href="https://git.openjdk.org/jdk/pull/13878.diff">https://git.openjdk.org/jdk/pull/13878.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13878#issuecomment-1539440887)

### DIFF
--- a/test/hotspot/jtreg/runtime/Safepoint/TestAbortOnVMOperationTimeout.java
+++ b/test/hotspot/jtreg/runtime/Safepoint/TestAbortOnVMOperationTimeout.java
@@ -26,13 +26,13 @@ import jdk.test.lib.process.*;
 
 /*
  * @test TestAbortOnVMOperationTimeout
- * @bug 8181143 8269523
+ * @bug 8181143 8269523 8307653
  * @summary Check abort on VM timeout is working
  * @requires vm.flagless
  * @library /test/lib
  * @modules java.base/jdk.internal.misc
  *          java.management
- * @run driver TestAbortOnVMOperationTimeout
+ * @run driver/timeout=600 TestAbortOnVMOperationTimeout
  */
 
 public class TestAbortOnVMOperationTimeout {
@@ -53,9 +53,7 @@ public class TestAbortOnVMOperationTimeout {
 
         // These should definitely pass: more than 3 minutes is enough for Serial to act.
         // The values are deliberately non-round to trip off periodic task granularity.
-        for (int delay : new int[]{183423}) {
-            testWith(delay, true);
-        }
+        testWith(183423, true);
 
         // These should fail: Serial is not very fast but we have seen the test
         // execute as quickly as 2ms!
--- a/test/hotspot/jtreg/runtime/Safepoint/TestAbortOnVMOperationTimeout.java
+++ b/test/hotspot/jtreg/runtime/Safepoint/TestAbortOnVMOperationTimeout.java
@@ -51,9 +51,9 @@ public class TestAbortOnVMOperationTimeout {
             return;
         }
 
-        // These should definitely pass: more than a minute is enough for Serial to act.
+        // These should definitely pass: more than 3 minutes is enough for Serial to act.
         // The values are deliberately non-round to trip off periodic task granularity.
-        for (int delay : new int[]{63423, 12388131}) {
+        for (int delay : new int[]{183423}) {
             testWith(delay, true);
         }
 
@@ -72,7 +72,7 @@ public class TestAbortOnVMOperationTimeout {
                 "-Xmx256m",
                 "-XX:+UseSerialGC",
                 "-XX:-CreateCoredumpOnCrash",
-                "-Xlog:gc",
+                "-Xlog:gc*=debug",
                 "TestAbortOnVMOperationTimeout",
                 "foo"
         );
--- a/test/hotspot/jtreg/runtime/Safepoint/TestAbortOnVMOperationTimeout.java
+++ b/test/hotspot/jtreg/runtime/Safepoint/TestAbortOnVMOperationTimeout.java
@@ -32,7 +32,7 @@ import jdk.test.lib.process.*;
  * @library /test/lib
  * @modules java.base/jdk.internal.misc
  *          java.management
- * @run driver/timeout=600 TestAbortOnVMOperationTimeout
+ * @run driver TestAbortOnVMOperationTimeout
  */
 
 public class TestAbortOnVMOperationTimeout {
@@ -51,8 +51,8 @@ public class TestAbortOnVMOperationTimeout {
             return;
         }
 
-        // These should definitely pass: more than 3 minutes is enough for Serial to act.
-        // The values are deliberately non-round to trip off periodic task granularity.
+        // This should definitely pass: more than 3 minutes is enough for Serial to act.
+        // The value is deliberately non-round to trip off periodic task granularity.
         testWith(183423, true);
 
         // These should fail: Serial is not very fast but we have seen the test
@@ -70,7 +70,7 @@ public class TestAbortOnVMOperationTimeout {
                 "-Xmx256m",
                 "-XX:+UseSerialGC",
                 "-XX:-CreateCoredumpOnCrash",
-                "-Xlog:gc*=debug",
+                "-Xlog:gc*=info",
                 "TestAbortOnVMOperationTimeout",
                 "foo"
         );


test/hotspot/jtreg/runtime/Safepoint/TestAbortOnVMOperationTimeout.java
```diff
@@ -51,9 +51,9 @@ public static void main(String[] args) throws Exception {
             return;
         }
 
-        // These should definitely pass: more than a minute is enough for Serial to act.
+        // These should definitely pass: more than 3 minutes is enough for Serial to act.
         // The values are deliberately non-round to trip off periodic task granularity.
-        for (int delay : new int[]{63423, 12388131}) {
+        for (int delay : new int[]{183423}) {
             testWith(delay, true);
         }
```
dholmes-ora, line 58: With only one value there is no point having the loop. Why not just replace the shorter, failing, delay with the new longer one?

dholmes-ora (changes requested):

test/hotspot/jtreg/runtime/Safepoint/TestAbortOnVMOperationTimeout.java
lgxbslgx: > With only one value there is no point having the loop

Good caught. I just obey the previous code style and forget to refactor it. Will fix it.

> Why not just replace the shorter, failing, delay with the new longer one?

The longer one is useless actually. If the shorter one, `63423` previously, passed,
it means the gc time is less than `63423`. Then the longer one will always pass.
And jtreg would report `time out` in seraval minutes normally,
so the longer time `12388131` is useless.
So I want to remove the longer one and just remain the shorter one.

test/hotspot/jtreg/runtime/Safepoint/TestAbortOnVMOperationTimeout.java
```diff
@@ -26,13 +26,13 @@
...
  * @test TestAbortOnVMOperationTimeout
- * @bug 8181143 8269523
+ * @bug 8181143 8269523 8307653
  * @summary Check abort on VM timeout is working
  * @requires vm.flagless
  * @library /test/lib
  * @modules java.base/jdk.internal.misc
  *          java.management
- * @run driver TestAbortOnVMOperationTimeout
+ * @run driver/timeout=600 TestAbortOnVMOperationTimeout
```
dholmes-ora, line 35: Is this change really needed? Do you see timeouts without it?

lgxbslgx: Yes, local env shows `time out`. The jtreg's default timeout is `480s`, so I increase it to `600s`.

dcubed-ojdk: Not quite. The default JTREG timeout is 120 seconds and the default timeoutFactor
value that we use is 4 for a total timeout value of 480 seconds. If your test runs
indicate the need for a total timeout value of 600 seconds, then you should use:
```
 * @run driver/timeout=150
```

lgxbslgx: I had tried `@run driver/timeout=480`, but the test failed. Then I used `timeout=600`, the test passed. So I set it to `600`.

dcubed-ojdk: What is your environment's timeoutFactor value?
Also, how are you running the test?

lgxbslgx: Maybe I should ignore such `time out` tests in my low performance dev board and don't fix it?

But such `time out` indeed pullutes the test result, and if I don't fix it,
I always need to read the failed test result to judge which tests fail.
In other words, my local CI will never be green.

dcubed-ojdk: When I run the test via:

lgxbslgx: > What is your environment's timeoutFactor value?

I haven't changed the default value. So it should be 4.

> Also, how are you running the test?

I used the command `make test Test=runtime/Safepoint/TestAbortOnVMOperationTimeout.java`.

dcubed-ojdk: Please search your .jtr file for your run for the pattern: "test.timeout.factor"
and report the results here.

lgxbslgx: I see the text `-J-Dtest.timeout.factor=4.0` in the log.

dcubed-ojdk: When you get a timeout with 'timeout=480', what 'elapsed time' values are in the '.jtr' file.
For my 'fastdebug' run, I see:
```
$ grep 'elapsed time' build/macosx-x86_64-normal-server-fastdebug/test-support/jtreg_open_test_hotspot_jtreg_runtime_Safepoint_TestAbortOnVMOperationTimeout_java/runtime/Safepoint/TestAbortOnVMOperationTimeout.jtr
elapsed time (seconds): 3.458
elapsed time (seconds): 3.455
elapsed time (seconds): 10.176
```

test/hotspot/jtreg/runtime/Safepoint/TestAbortOnVMOperationTimeout.java
```diff
@@ -51,11 +51,9 @@ public static void main(String[] args) throws Exception {
             return;
         }
 
-        // These should definitely pass: more than a minute is enough for Serial to act.
+        // These should definitely pass: more than 3 minutes is enough for Serial to act.
```
dcubed-ojdk, line 54: "These" is no longer accurate since this has been changed to test a single value.

test/hotspot/jtreg/runtime/Safepoint/TestAbortOnVMOperationTimeout.java
```diff
@@ -51,11 +51,9 @@ public static void main(String[] args) throws Exception {
...
         // The values are deliberately non-round to trip off periodic task granularity.
-        for (int delay : new int[]{63423, 12388131}) {
-            testWith(delay, true);
-        }
+        testWith(183423, true);
```
dcubed-ojdk, line 56: "The values are ..." is no longer accurate since this has been changed to test a single value.

dcubed-ojdk (changes requested):

lgxbslgx:
> The proper way to address a slow piece of hardware is to use a larger timeoutFactor value.

Got it. I will remove the `/timeout=600` of the test.

dcubed-ojdk:
BTW, my first experiment for [JDK-8296819](https://bugs.openjdk.org/browse/JDK-8296819) would
have been switching from (default) timeout=120 to timeout=180 and running that thru the stress
kit in my lab.

test/hotspot/jtreg/runtime/Safepoint/TestAbortOnVMOperationTimeout.java
lgxbslgx: I can't reproduce the failed test by using `@run driver/timeout=480` now. The test passes locally now with `timeout=480`. The `elapsed time` values, which belong to the current passed test, are shown below.

```
elapsed time (seconds): 115.718
elapsed time (seconds): 115.699
elapsed time (seconds): 555.602
```

I try to recall my previous steps:
- The test failed locally with `@run driver`,  without `timeout` option.
- I changed it to `@run driver/timeout=900`. It passed.
- I changed it to `@run driver/timeout=480`. It failed.
- I changed it to `@run driver/timeout=600`. It passed.

So I push the code with `@run driver/timeout=600`.

Now I try `@run driver/timeout=480` locally, it can't reproduce. So strange! Such intermittent fail may be related to [JDK-8296819](https://bugs.openjdk.org/browse/JDK-8296819)?

test/hotspot/jtreg/runtime/Safepoint/TestAbortOnVMOperationTimeout.java
lgxbslgx: Fixed.

test/hotspot/jtreg/runtime/Safepoint/TestAbortOnVMOperationTimeout.java
lgxbslgx: Fixed.

dholmes-ora (approved):
I am okay with these changes.

I do worry that your slow dev board may expose a lot of internal timeout issues in tests - but tests should be using values modified by the timeoutFactor to compensate for slow systems where possible. Not sure how that would apply in the current case though.

dcubed-ojdk (approved):

lgxbslgx:
@dholmes-ora @dcubed-ojdk Thanks for the review. Integrating.
/integrate

PR #13888, opened 2023-05-09
8288619: Unexpected parsing for @see

lgxbslgx:
Hi all,

When invoking the method `DocCommentParser::quotedString` with field `newline` as `true` 
and the quoted string has the character `@`, `quotedString` will break the loop and return `null`.
This patch changes the field `newline` to `false` at the beginning of the method `quotedString`
to solve this issue. And a corresponding test case is added.

Thanks for the review.

Best Regards,
-- Guoxiong

---------
### Progress

### Issue
 * [JDK-8288619](https://bugs.openjdk.org/browse/JDK-8288619): Unexpected parsing for @see

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13888/head:pull/13888` \
`$ git checkout pull/13888`

`$ git checkout pull/13888` \
`$ git pull https://git.openjdk.org/jdk.git pull/13888/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13888`

`$ git pr show -t 13888`

</details>

<a href="https://git.openjdk.org/jdk/pull/13888.diff">https://git.openjdk.org/jdk/pull/13888.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13888#issuecomment-1540206811)

### DIFF
--- a/test/langtools/tools/javac/doctree/SeeTest.java
+++ b/test/langtools/tools/javac/doctree/SeeTest.java
@@ -75,6 +75,7 @@ DocComment[DOC_COMMENT, pos:1
      * @see
      *    "{@code}"
      */
+    @PrettyCheck(false)
     void new_line_before_quoted_string() { }
 /*
 DocComment[DOC_COMMENT, pos:1
--- a/test/langtools/tools/javac/doctree/SeeTest.java
+++ b/test/langtools/tools/javac/doctree/SeeTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -23,7 +23,7 @@
 
 /*
  * @test
- * @bug 7021614 8031212 8273244 8284908 8200337
+ * @bug 7021614 8031212 8273244 8284908 8200337 8288619
  * @summary extend com.sun.source API to support parsing javadoc comments
  * @modules jdk.compiler/com.sun.tools.javac.api
  *          jdk.compiler/com.sun.tools.javac.file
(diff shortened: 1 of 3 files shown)


jonathan-gibbons:
> > Overall, I would expect `newline` to be set `false`earlier, maybe in `blockTag`, so that the setting applies to all block tags. That might still leave the possibility of a newline happening after the tag name and before the quoted string, but your test case does not cover that.
> 
> If there is something between the tag name and the quoted string, the method `quotedString` is never invoked by method `TagParser::parse`. So such case seems not related to this bug. Do I miss anything?

I was just trying to imagine a case where `newline` might be true after the tag name has been read.

jonathan-gibbons (changes requested):
I see the problem reported in the JBS issue, but the proposed fix is not the best one.

It would be better to set the `newline` flag to false earlier, perhaps at the beginning of `blockTag`

test/langtools/tools/javac/doctree/SeeTest.java
```diff
@@ -50,6 +50,24 @@ void quoted_text() { }
         Text[TEXT, pos:12, "String"]
     ]
 ]
+*/
+
+    /**
+     * Test '@' in quoted string.
+     * @see "{@code}"
+     */
+    void at_sign_in_quote_string() { }
```
jonathan-gibbons, line 59: (minor) add `d` after `quote` in method name

jonathan-gibbons: Also, add a test case for this, with a newline after the tag:
```

    /**
     * abc.
     * @see
     *    "{@code ...}"
     */
```

lgxbslgx: > (minor) add `d` after `quote` in method name

Fixed.

lgxbslgx:
> It would be better to set the `newline` flag to false earlier, perhaps at the beginning of `blockTag`

Fixed.

> Here is another test case:
> 
> ```
>     /**
>      * case 2.
>      * @see
>      *    "{@code ...}"
>      */
>     public void m2() { }
> ```

I add this test case. But it fails with the following error message. The `DocCommentTester#PrettyChecker` seems not work as expected?

```
METHOD new_line_before_quoted_string PrettyChecker
Error: mismatch
*** expected:
Test_new_line_before_quoted_string.
@see
____"{@code}"
*** found:
Test_new_line_before_quoted_string.
@see_"{@code}"
```

jonathan-gibbons (approved):

lgxbslgx:
@jonathan-gibbons Thanks for the review. Integrating.

/integrate

PR #13889, opened 2023-05-09
8307766: Linux: Provide the option to override the timer slack

shipilev:
See the description in the RFE.

On `c6.8xlarge`, the sleep accuracy is improved the following way:

```
% CONF=linux-x86_64-server-release make test TEST=micro:ThreadSleep.millisNanos

Benchmark                   (sleep)  Mode  Cnt           Score      Error  Units

# Default
ThreadSleep.millisNanos           0  avgt   15         473.727 ±    3.542  ns/op
ThreadSleep.millisNanos           1  avgt   15         585.868 ±    2.010  ns/op
ThreadSleep.millisNanos          10  avgt   15         583.475 ±    1.426  ns/op
ThreadSleep.millisNanos         100  avgt   15       57833.073 ±  146.406  ns/op
ThreadSleep.millisNanos        1000  avgt   15       59324.901 ±  620.924  ns/op
ThreadSleep.millisNanos       10000  avgt   15       68071.990 ±  537.880  ns/op
ThreadSleep.millisNanos      100000  avgt   15      158069.170 ±  329.991  ns/op
ThreadSleep.millisNanos     1000000  avgt   15     1059044.770 ±  370.982  ns/op
ThreadSleep.millisNanos    10000000  avgt   15    10060017.354 ±  399.309  ns/op
ThreadSleep.millisNanos   100000000  avgt   15   100062969.073 ± 2399.053  ns/op
ThreadSleep.millisNanos  1000000000  avgt   15  1000068474.467 ± 2669.980  ns/op

# -XX:TimerSlack=1
ThreadSleep.millisNanos           0  avgt   15         469.632 ±    1.500  ns/op
ThreadSleep.millisNanos           1  avgt   15         587.562 ±    8.966  ns/op
ThreadSleep.millisNanos          10  avgt   15         584.241 ±    1.997  ns/op
ThreadSleep.millisNanos         100  avgt   15        7907.160 ±   12.476  ns/op
ThreadSleep.millisNanos        1000  avgt   15        7913.984 ±   56.996  ns/op
ThreadSleep.millisNanos       10000  avgt   15       17929.107 ±  432.083  ns/op
ThreadSleep.millisNanos      100000  avgt   15      108154.851 ±  403.717  ns/op
ThreadSleep.millisNanos     1000000  avgt   15     1008854.095 ±  418.444  ns/op
ThreadSleep.millisNanos    10000000  avgt   15    10009902.829 ±  279.617  ns/op
ThreadSleep.millisNanos   100000000  avgt   15   100014000.020 ± 2143.820  ns/op
ThreadSleep.millisNanos  1000000000  avgt   15  1000020734.333 ± 4846.754  ns/op
```

---------
### Progress

### Issue
 * [JDK-8307766](https://bugs.openjdk.org/browse/JDK-8307766): Linux: Provide the option to override the timer slack (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13889/head:pull/13889` \
`$ git checkout pull/13889`

`$ git checkout pull/13889` \
`$ git pull https://git.openjdk.org/jdk.git pull/13889/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13889`

`$ git pr show -t 13889`

</details>

<a href="https://git.openjdk.org/jdk/pull/13889.diff">https://git.openjdk.org/jdk/pull/13889.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13889#issuecomment-1554181631)


src/hotspot/os/linux/globals_linux.hpp
```diff
@@ -77,7 +77,18 @@
           "Use CPU_ALLOC code path in os::active_processor_count ")     \
                                                                         \
   product(bool, DumpPerfMapAtExit, false, DIAGNOSTIC,                   \
-          "Write map file for Linux perf tool at exit")
+          "Write map file for Linux perf tool at exit")                 \
+                                                                        \
+  product(intx, TimerSlack, -1, EXPERIMENTAL,                           \
```
dholmes-ora, line 82: Why experimental? Is this not something that these users want to use in production?

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -4646,6 +4647,13 @@ jint os::init_2(void) {
     FLAG_SET_DEFAULT(UseCodeCacheFlushing, false);
   }
 
+  // Override the timer slack value if needed.
```
dholmes-ora, line 4650: This needs more explanation. This is a per-thread setting, so by doing it here we change the main JVM thread so that subsequent threads created by this thread will also inherit the new value. Any native threads attaching to the JVM will have their own values.

dholmes-ora (commented):
If this is really of use to people then I expect it would also be useful to know what the current setting is for a thread - perhaps either logged at thread creation, or perhaps included in one of the thread `print_on` functions?

dholmes-ora:
ThreadSleep.millisNanos          10  avgt   15         583.475 ±    1.426  ns/op
ThreadSleep.millisNanos         100  avgt   15       57833.073 ±  146.406  ns/op
ThreadSleep.millisNanos        1000  avgt   15       59324.901 ±  620.924  ns/op
ThreadSleep.millisNanos       10000  avgt   15       68071.990 ±  537.880  ns/op

Not sure what is being measured here but this looks odd. The first 10x increase to 100 results in a 100x increase. But then  the next 10x is almost no difference, and the 10x after that not that much difference.

shipilev:
> ```
> ThreadSleep.millisNanos          10  avgt   15         583.475 ±    1.426  ns/op
> ThreadSleep.millisNanos         100  avgt   15       57833.073 ±  146.406  ns/op
> ThreadSleep.millisNanos        1000  avgt   15       59324.901 ±  620.924  ns/op
> ThreadSleep.millisNanos       10000  avgt   15       68071.990 ±  537.880  ns/op
> ```
> 
> Not sure what is being measured here but this looks odd. The first 10x increase to 100 results in a 100x increase. But then the next 10x is almost no difference, and the 10x after that not that much difference.

Odd, right? This is how the 50us default timer slack value introduces jitter: you can turn around fast, or you can get stalled up to the timer slack. And this is precisely (pun intended) why I'd like to have some control over the timer slack :)

src/hotspot/os/linux/globals_linux.hpp
shipilev: I would say it is an option that can be used sparingly in selected production/experimental workloads. Like other low-level VM options, it stays experimental to avoid a stronger contract for supportability. I can turn that bona fide `product`, if you want.

shipilev:
> If this is really of use to people then I expect it would also be useful to know what the current setting is for a thread - perhaps either logged at thread creation, or perhaps included in one of the thread `print_on` functions?

Yes, that would be a good thing to add, see new commit. I tried to limit the fan-out for this change by printing the timer slack in Linux-specific thread init block, near the stack printing.

src/hotspot/os/linux/os_linux.cpp
shipilev: Done.

src/hotspot/os/linux/globals_linux.hpp
dholmes-ora: What do your "customers" want? Are they happy to deploy with an experimental flag on the command-line?

dholmes-ora:
> And this is precisely (pun intended) why I'd like to have some control over the timer slack

I'm still not really understanding what this timer slack actually means in practice and how it relates to the sleep benchmark.

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -4646,6 +4656,15 @@ jint os::init_2(void) {
     FLAG_SET_DEFAULT(UseCodeCacheFlushing, false);
   }
 
+  // Override the timer slack value if needed. The adjustment for the main
+  // thread would inherit the setting the children threads, which would be
+  // most threads in JDK/JVM.
```
dholmes-ora, line 4661: s/would inherit the setting the children threads/will establish the setting for child threads/

shipilev: Done, thanks.

src/hotspot/os/linux/globals_linux.hpp
shipilev: Yes, experimental flag is fine for limited internal use. We can promote this flag to product if/when it proves to be useful for the generic low-latency cases.

dholmes-ora:
> Perhaps this LWN article would help ...

Thanks. So are timer events only coalesced within a process, or is this system wide? I'm guessing the basic operation is that if a thread requests a "timed wait" that would require a timer event at time T1, but there already exists a timer event for time T2, (T2 > T1) then if T2-T1 < timer-slack then an event for T1 is not created and we piggy-back on the event for T2 instead. So the lower the slack the more likely we get our own timer-event and so get a resolution closer to that of the h/w, but it also means we have more timer-events to process so that could lower the resolution again.

shipilev:
> > Perhaps this LWN article would help ...
> 
> Thanks. So are timer events only coalesced within a process, or is this system wide? I'm guessing the basic operation is that if a thread requests a "timed wait" that would require a timer event at time T1, but there already exists a timer event for time T2, (T2 > T1) then if T2-T1 < timer-slack then an event for T1 is not created and we piggy-back on the event for T2 instead. So the lower the slack the more likely we get our own timer-event and so get a resolution closer to that of the h/w, but it also means we have more timer-events to process so that could lower the resolution again.

I have not read the kernel sources for this, but my recollection is that timer lists are maintained per-CPU + there is a migration mechanics, which is effectively make timers system-wide. So timer slack allows piggy-backing on hi-res timers already pending on the system, which improves efficiency at the accuracy cost. I suspect that even if we are the only timer currently pending, the kernel would still apply the timer slack to it, hoping that some _other_ timer would be able to piggy-back on us. Which explains why a single-threaded `Thread.sleep` workload is sensitive to timer slack even on a rather quiet system.

Yes, I do wonder if we can blow out the kernel time subsystem with very low timer slack value, but that is something that needed to be checked for the concrete workload. Linux helpfully provides us the API to tune the heuristics either way, so this PR gives us the handy ability to experiment with it :)

shipilev:
(Trying to revive this PR) Let me know if there are more comments to this.

dholmes-ora (commented):
As this is experimental and seems non-intrusive I can approve it. But how will this be tested? Can we at least sanity check setting it and reading back the value we set?

Thanks.

shipilev:
> As this is experimental and seems non-intrusive I can approve it. But how will this be tested? Can we at least sanity check setting it and reading back the value we set?

Yes, and even more, see the new commit with a test.

src/hotspot/os/linux/globals_linux.hpp
```diff
@@ -77,7 +77,18 @@
...
+          "(Linux-specific) Override the timer slack value to this "    \
+          "number of nanoseconds. Lower value provides more accurate "  \
+          "high-precision timers, at the expense of (possibly) worse "  \
+          "power efficiency. Set -1 to disable the override. In current "\
+          "Linux, 0 means using the system-wide default, which is the " \
+          "same as disabled override. See prctl(PR_SET_TIMERSLACK) for "\
+          "more info.")                                                 \
+                                                                        \
```
tstuefe, line 90: None of the other Linux specifc args write "(Linux specific)". I'd leave that out.

The first sentence is difficult to parse. How about "Overrides timer slack value to the given number of nanoseconds [1..n]."

Why do we need two values (-1, 0) for oob? Is the "In current Linux, 0..." info useful for anything? I'd just use "0" as "default".

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -988,6 +989,15 @@ bool os::create_thread(Thread* thread, ThreadType thr_type,
...
 
+    {
+      // Print current timer slack, if available
+      int slack = prctl(PR_GET_TIMERSLACK);
+      if (slack >= 0) {
+        log_info(os, thread)("Thread \"%s\" (pthread id: " UINTX_FORMAT ") timer slack: %dns",
+                             thread->name(), (uintx) tid, slack);
+      }
+    }
+
```
tstuefe, line 1000: Maybe also in create_attached_thread? And then it would make sense to factor this out into a small static helper.

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -4701,6 +4711,15 @@ jint os::init_2(void) {
...
 
+  // Override the timer slack value if needed. The adjustment for the main
+  // thread will establish the setting for child threads, which would be
+  // most threads in JDK/JVM.
+  if (TimerSlack > 0) {
+    if (prctl(PR_SET_TIMERSLACK, TimerSlack) < 0) {
+      vm_exit_during_initialization("Setting timer slack failed: %s", os::strerror(errno));
+    }
+  }
+
```
tstuefe, line 4722: Is this enough? The API reads as if this should happen for every thread individually. 

Or do threads inherit this override from their parent threads? So you only set it once for main thread? 

But even then, you'd need to do it at least for every thread that gets attached, since their parentage is unknown.

This raises another question, if a thread is detached, should we restore the original timer slack?

dholmes-ora: > Or do threads inherit this override from their parent threads? So you only set it once for main thread?

Exactly. And my understanding is that we are only concerned with the main set of Java threads not any natively attached threads. If you have native threads then they can set their own values. We should not be second-guessing or trying to save/restore.

src/hotspot/os/linux/globals_linux.hpp
dholmes-ora: IIUC using zero as default would be incorrect as that may reset an inherited customized value to the platform default. A value of -1 says "ignore this".

test/hotspot/jtreg/runtime/os/TestTimerSlack.java
```diff
@@ -0,0 +1,128 @@
...
+
+            OutputAnalyzer output = new OutputAnalyzer(pb.start());
+            output.shouldHaveExitValue(0);
+            output.shouldContain("timer slack:");
+
+            defaultSlack = parseSlackValue(output);
+
+            if (defaultSlack == 0) {
+                output.reportDiagnosticSummary();
+                throw new IllegalStateException("Default slack value (" + defaultSlack + ") is unexpected");
```
dholmes-ora, line 56: This pattern is repeated quite a few times. Pity we can't make use of something in OutputAnalyzer directly. A small helper method `fail(String msg)` would avoid the repetition.

test/hotspot/jtreg/runtime/os/TestTimerSlack.java
```diff
@@ -0,0 +1,128 @@
...
+
+            int disabledSlack = parseSlackValue(output);
+            if (disabledSlack != defaultSlack) {
+                output.reportDiagnosticSummary();
+                throw new IllegalStateException("Disabled slack value (" + disabledSlack +
+                    ") is not default (" + defaultSlack + ") when requested (" + slack + ")");
+            }
+        }
+
+        // Check the timer slack value is accepted by all threads
```
dholmes-ora, line 79: I don't see anything thread related in the logic that follows. I expected to see the test program create threads and read back their time-slack values.

src/hotspot/os/linux/os_linux.cpp
tstuefe: > > Or do threads inherit this override from their parent threads? So you only set it once for main thread?
> 
> Exactly. And my understanding is that we are only concerned with the main set of Java threads not any natively attached threads. If you have native threads then they can set their own values. We should not be second-guessing or trying to save/restore.

That seems an strange omission and at odds with other runtime decisions we force upon attached threads, e.g. stack size / guard pages.

If that decision is deliberate, we should at least document it in the flag description.

src/hotspot/os/linux/globals_linux.hpp
tstuefe: We only call pctrl if the flag is >0. So, 0 has already the meaning of "ignore".

src/hotspot/os/linux/os_linux.cpp
dholmes-ora: ??? We can't affect either the stacksize, nor guard pages of an existing native thread!

src/hotspot/os/linux/globals_linux.hpp
dholmes-ora: That seems a bug then. We should pass zero to pctrl if that is the value set (even if the current implementation does nothing for a value of zero).

src/hotspot/os/linux/os_linux.cpp
tstuefe: > ??? We can't affect either the stacksize, nor guard pages of an existing native thread!

No, but we place guard pages into the stacks of attached threads. We also do other things, e.g. set the signal mask, or set the FPU control word.

My point is that we attempt to minimize the differences between attached threads and created threads. Which makes sense. Therefore it feels odd that Java APIs that depend on blocking syscalls should show different behavior depending on whether they are called in a created or in an attached thread. The developer of the Java code may not even know. You may run with hotspot embedded into some custom launcher, and the launcher doing its own thread pooling.

dholmes-ora: We only do to attached threads what we need to do to enable them to run Java - and those things should not affect them once they have detached. We don't mess with attached thread names for example. You can make the case either way for this - you may want the attached thread to behave exactly like a native Java thread, or you may be upset that the behaviour of an attached threads changes when executing non Java code because of some action attaching took. So I don't have a concern that native threads are not affected by this experimental flag at least at this initial stage of experimentation.

tstuefe: You are saying we do the minimum needed to run java, but leave them alone otherwise? Okay, I can see the logic in that.

src/hotspot/os/linux/os_linux.cpp
shipilev: As discussed in other thread, I'd prefer to keep the attached threads out of the picture here.

test/hotspot/jtreg/runtime/os/TestTimerSlack.java
shipilev: Done so in the new commit.

test/hotspot/jtreg/runtime/os/TestTimerSlack.java
shipilev: Right. I was mostly concentrating on main thread. New commit spawns a few threads explicitly. Alas, we cannot check if those are the same threads, because our printout happens on thread creation path, when `setName` did not yet have a chance to set the thread name.

src/hotspot/os/linux/globals_linux.hpp
shipilev: Right. Thinking more about it, I think I am going to use `-1` to mean that all new paths are completely disabled, so that we don't call `prctl` by default, and do not print the slack values at all. With `0`, we rely on system default, and then print out the actual values. With `> 0`, we both override the value and print the current slack values.

This insulates current code from this experimental addition better.

dholmes-ora (approved):
This all seems reasonable to me.

Thanks.

tstuefe (approved):
Okay.

shipilev:
RISC-V build failure looks like an infrastructure issue.

I re-tested current patch on `ThreadSleep` bench with various `TimerSlack` values, and they still make sense.

/integrate

PR #13895, opened 2023-05-10
8307572: AArch64: Vector registers are clobbered by some macroassemblers

nsjian:
I found that MacroAssembler::arrays_equals() would call stubcode, which may use vector registers. However, the call site in match rule does not claim the use of vector registers. Since c2 will allocate v16-v31 first [1], it's rare that using of v0-v7 will cause problem, but I did create a test case to expose the bug.

Apart from arrays_equals, I also checked other macroassemblers, and found several similar issues. Fixed by claiming those vector register being killed in match rules call sites, which should have minimal performance impact compared to always saving/restoring those vector registers, since those V0-Vx registers are rarely allocated and live cross the macroassembler call.

A jtreg test case is also added to demonstrate the failure. Test will fail without this patch, and pass with this patch.

Test: I tried to update the allocation order in [1] to allocate V0-V15 first and then V16-V31, and full jtreg tests passed with the allocation order changed. (I did found some test failures with this allocation order change without this patch). I have also eyeballed and checked other macroassembler calls, and others seemed fine.

[1] https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/aarch64/aarch64.ad#L424

---------
### Progress

### Issue
 * [JDK-8307572](https://bugs.openjdk.org/browse/JDK-8307572): AArch64: Vector registers are clobbered by some macroassemblers

### Reviewers
 * [Andrew Dinn](https://openjdk.org/census#adinn) (@adinn - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13895/head:pull/13895` \
`$ git checkout pull/13895`

`$ git checkout pull/13895` \
`$ git pull https://git.openjdk.org/jdk.git pull/13895/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13895`

`$ git pr show -t 13895`

</details>

<a href="https://git.openjdk.org/jdk/pull/13895.diff">https://git.openjdk.org/jdk/pull/13895.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13895#issuecomment-1541436203)

### DIFF
--- a/src/hotspot/cpu/aarch64/aarch64.ad
+++ b/src/hotspot/cpu/aarch64/aarch64.ad
@@ -17078,14 +17078,17 @@ instruct string_compareUU_sve(iRegP_R1 str1, iRegI_R2 cnt1, iRegP_R3 str2, iRegI
 %}
 
 instruct string_indexofUU(iRegP_R1 str1, iRegI_R4 cnt1, iRegP_R3 str2, iRegI_R2 cnt2,
-       iRegI_R0 result, iRegINoSp tmp1, iRegINoSp tmp2, iRegINoSp tmp3,
-       iRegINoSp tmp4, iRegINoSp tmp5, iRegINoSp tmp6, rFlagsReg cr)
+                          iRegI_R0 result, iRegINoSp tmp1, iRegINoSp tmp2,
+                          iRegINoSp tmp3, iRegINoSp tmp4, iRegINoSp tmp5, iRegINoSp tmp6,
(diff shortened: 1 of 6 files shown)


theRealAph (approved):
Great catch, thanks. Does this one need backports?

adinn (approved):
Nice work Ningsheng!

nsjian:
Thanks for the review! @theRealAph @adinn

PR #13899, opened 2023-05-10
8307478: Implementation of Prepare to Restrict The Dynamic Loading of Agents

AlanBateman:
This is the implementation for JEP 451. There are two parts to this:

1. A multi-line warning is printed when a JVM TI or Java agent is loaded into a running VM. For JVM TI, the message is printed to stderr from JvmtiAgent::load. For Java agents, it is printed to System.err (as that may be redirected) in the JPLIS (j.l.instrumentation) implementation. This part includes an update to the JVM TI spec and API docs to require the warning.

2. If running with -Djdk.instrument.traceUsage or -Djdk.instrument.traceUsage=true, the calls to the Instrumentation API print a trace message and stack trace.

Testing: tier1-6

---------
### Progress
- [x] Change requires CSR request [JDK-8307479](https://bugs.openjdk.org/browse/JDK-8307479) to be approved

### Issues
 * [JDK-8307478](https://bugs.openjdk.org/browse/JDK-8307478): Implementation of Prepare to Restrict The Dynamic Loading of Agents
 * [JDK-8307479](https://bugs.openjdk.org/browse/JDK-8307479): Implementation of Prepare to Restrict The Dynamic Loading of Agents (**CSR**)

### Reviewers
 * [Serguei Spitsyn](https://openjdk.org/census#sspitsyn) (@sspitsyn - **Reviewer**) ⚠️ Review applies to [cbd9bb1d](https://git.openjdk.org/jdk/pull/13899/files/cbd9bb1d56283167975d5c31e16692361edb2d80)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13899/head:pull/13899` \
`$ git checkout pull/13899`

`$ git checkout pull/13899` \
`$ git pull https://git.openjdk.org/jdk.git pull/13899/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13899`

`$ git pr show -t 13899`

</details>

<a href="https://git.openjdk.org/jdk/pull/13899.diff">https://git.openjdk.org/jdk/pull/13899.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13899#issuecomment-1543970472)


src/java.instrument/share/classes/sun/instrument/InstrumentationImpl.java
```diff
@@ -59,6 +64,15 @@
  * processing behind native methods.
  */
 public class InstrumentationImpl implements Instrumentation {
+    private static final String TRACE_USAGE_PROP_NAME = "jdk.instrument.traceUsage";
+    private final static boolean TRACE_USAGE;
```
turbanoff, line 68:
```suggestion
    private static final boolean TRACE_USAGE;
```

src/java.instrument/share/classes/java/lang/instrument/package-info.java
```diff
@@ -169,36 +189,9 @@
...
- *
- * <p> The {@code agentmain} method should do any necessary initialization
- * required to start the agent and return. If the agent cannot be started, for
- * example the agent class cannot be loaded, the agent class does not define a
- * conformant {@code agentmain} method, or the {@code agentmain} method throws
- * an uncaught exception or error, the JVM will abort.
- *
+ * <p> The agent class may also have a {@code premain} method for use when the agent
+ * is started using a command-line option. When the agent is started in a running JVM
+ * the {@code premain} method is not invoked.
```
plummercj, line 194:
```suggestion
 * is started using a command-line option. When the agent is started in a running JVM,
 * the {@code premain} method is not invoked.
```

src/java.instrument/share/classes/java/lang/instrument/package-info.java
```diff
@@ -248,31 +241,33 @@
...
  *
- * <p> The following manifest attributes are defined for an agent JAR file:
+ * <p> The following attributes in the main section of the application or agent
+ * JAR file manifest are defined for Java agents:
  *
  * <blockquote><dl>
  *
+ * <dt>{@code Launcher-Agent-Class}</dt>
+ * <dd> If an implementation supports a mechanism to start an application in an
+ * executable JAR file then this attribute, if present, specifies the binary name
```
plummercj, line 253:
```suggestion
 * <dd> If an implementation supports a mechanism to start an application in an
 * executable JAR file, then this attribute, if present, specifies the binary name
```

src/java.instrument/share/classes/java/lang/instrument/package-info.java
```diff
@@ -248,31 +241,33 @@
...
+ * of the agent class that is packaged with the application.
+ * The agent is started, by invoking the agent class {@code agentmain} method,
+ * before the application {@code main} method is invoked. </dd>
```
plummercj, line 256:
```suggestion
 * The agent is started by invoking the agent class {@code agentmain} method
 * before the application {@code main} method is invoked. </dd>
```

test/jdk/com/sun/tools/attach/warnings/DynamicLoadWarningTest.java
```diff
@@ -0,0 +1,210 @@
...
+                    // shutdown
+                    s.getOutputStream().write(0);
+                } catch (Exception e) {
+                    e.printStackTrace();
+                }
+            });
+
+            // launch application with the given VM options, waiting for it to terminate
+            Stream<String> s1 = Stream.of(vmopts);
+            Stream<String> s2 = Stream.of("Application", ""+listener.getLocalPort());
```
sspitsyn, line 197: Nit: Spaces are missed around '+' sign at lines: 121, 180 and 197.

sspitsyn (approved):
The fix looks good. Posted one nit though.
Thanks,
Serguei

src/java.instrument/share/native/libinstrument/InvocationAdapter.c
```diff
@@ -486,17 +491,18 @@ jint loadAgent(JNIEnv* env, jstring path) {
...
-    }
-
     // get path to JAR file as UTF-8 string
     jarfile = (*env)->GetStringUTFChars(env, path, NULL);
     if (jarfile == NULL) {
         return JNI_ERR;
     }
 
+    // create JPLISAgent with JVMTI environment
+    if (createNewJPLISAgent(vm, &agent, jarfile, JNI_FALSE) != JPLIS_INIT_ERROR_NONE) {
```
plummercj, line 501: No warning is ever printed in this case since this is an agent being loaded from an executable JAR. Correct?

test/jdk/com/sun/tools/attach/warnings/DynamicLoadWarningTest.java
AlanBateman: Okay, we can use Integer.toString and Long.toString instead.

src/java.instrument/share/native/libinstrument/InvocationAdapter.c
AlanBateman: That's right, it's just for agents that are loaded into a running VM.

test/jdk/com/sun/tools/attach/warnings/DynamicLoadWarningTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+        // get absolute path to JVM TI agent
+        String libname;
+        if (Platform.isWindows()) {
+            libname = JVMTI_AGENT_LIB + ".dll";
+        } else if (Platform.isOSX()) {
+            libname = "lib" + JVMTI_AGENT_LIB + ".dylib";
+        } else {
+            libname = "lib" + JVMTI_AGENT_LIB + ".so";
+        }
```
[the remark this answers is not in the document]
AlanBateman: I wasn't aware of Platform.sharedLibraryExt(), it reduces the code a bit here.

test/jdk/com/sun/tools/attach/warnings/DynamicLoadWarningTest.java
```diff
@@ -0,0 +1,210 @@
...
+    @Test
+    void testLoadJvmtiAgent() throws Exception {
+        String message = "WARNING: A JVM TI agent has been dynamically loaded";
+
+        // warning should be printed
+        test((pid, vm) -> vm.loadAgentLibrary(JVMTI_AGENT_LIB)).shouldContain(message);
+
+        // no warning should be printed
+        test((pid, vm) -> vm.loadAgentLibrary(JVMTI_AGENT_LIB), "-XX:+EnableDynamicAgentLoading")
+                .shouldNotContain(message);
```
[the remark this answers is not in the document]
AlanBateman: We can't currently test `jcmd JVMTI.agent_load` with `-XX:-EnableDynamicAgentLoading` due to JDK-8304438.

plummercj (approved):
Looks good.

apangin (changes requested):
When testing the proposed patch, I stumbled upon two (related) issues:

1. There are no checks that an agent have been already loaded. When I invoke `agent_load` multiple times with exactly the same library path, the warning is printed every time. The repeated warnings are not helpful; they are neither good for user experience nor for integrity, since agent developers will be tempted to suppress excessive warnings on the first load by hacking the JDK internals and thus breaking integrity even more.
2. If an agent is loaded at JVM startup using `-agentlib/agentpath` option, the subsequent communication with the agent through jcmd results in the warning being printed again and again. This contradicts the intentions described in JEP 451:

    > Tools that employ agents loaded at startup are unaffected by these changes.

    As @pron mentioned, the presence of `-agentlib/agentpath` option serves as an explicit user consent to use the tool, and disallowing such agents (or issuing a warning about them) is a non-goal of the JEP.

    With the current behavior, users will be tempted to add `-XX:+EnableDynamicAgentLoading` option instead of putting one particular agent on the command line. This again does not serve the purpose of strengthen the integrity.

The use cases I mentioned here are quite natural and supported by popular profilers: a user may want to use a tool some time after the JVM startup, or reconfigure it later at runtime similarly to how `jcmd JFR.configure` works.

AlanBateman:
> When testing the proposed patch, I stumbled upon two (related) issues:

Thanks for bringing this up. The question is whether dynamically loading the same same agent library N times is treated as one or N agents. Similarly, if an agent library is specified on the command line, then later the same agent is dynamically loaded and started in the running VM, is it the same agent or not?  Right now, they are treated as different agents, they may have different agent options, they accumulate in the "agent list", each load/start generates a JFR event, and with the changes here, emit a warning. I think we'll need to think a bit more about that scenario as the notion of agent identity doesn't surface in the API docs (the canonical path to the agent library or JAR might do). I agree it could be annoying to have multiple warnings if the profiler is attaching and loading the same agent library many times.

AlanBateman:
>  I agree it could be annoying to have multiple warnings if the profiler is attaching and loading the same agent library many times.

I've updated the changes to allow it be implementation specific as to whether a warning is printed when attempting to start the same agent a second or subsequent time. I think to sync up with Ron/Alex on this point on whether the description in the JEP needs to say anything on this. 

For now, attempting to start a JVM TI agent that was previously loaded will not print a warning. We could potentially do the same for Java agents.

src/hotspot/share/prims/jvmtiAgent.cpp
```diff
@@ -504,6 +507,15 @@ static bool invoke_Agent_OnAttach(JvmtiAgent* agent, outputStream* st) {
     agent->set_os_lib(library);
     agent->set_loaded();
   }
+
+  // Print warning if EnableDynamicAgentLoading not enabled on the command line
+  if (!FLAG_IS_CMDLINE(EnableDynamicAgentLoading) && !agent->is_instrument_lib() && !JvmtiAgentList::is_loaded(library)) {
```
plummercj, line 512: While looking at some code related to this, I noticed a couple of typos in the pre-existing load_agent_from_executable() comment. See lines 265 ("cant't") and 268 (".&&"). Maybe you could clean them up.

plummercj, line 512: The use of `!JvmtiAgentList::is_loaded(library)` here is a bit confusing. Isn't the library always already loaded by the time we get here (the assert below seems to imply that)? If so, wouldn't it already be in the list? If it's not in the list yet, perhaps a comment explaining why would be helpful here.

test/jdk/com/sun/tools/attach/warnings/DynamicLoadWarningTest.java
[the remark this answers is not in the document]
plummercj: @AlanBateman  @sspitsyn Can this be revisited now that JDK-8304438 is being worked on? Perhaps the changes to JDK-8304438 can include an update to this test, although it all depends on which is being pushed first.

src/hotspot/share/prims/jvmtiAgent.cpp
sspitsyn: Nit: I'd suggest to add dot at the end of comment at 511.

sspitsyn: It looks like you are right.
There is also and assert at line 519:
`519   assert(agent->is_loaded(), "invariant");`

So, the agent has to be loaded if we got to the line 512.
Also, there is a statement at line 507 (before line 512):
`507    agent->set_os_lib(library);`

AlanBateman: > The use of `!JvmtiAgentList::is_loaded(library)` here is a bit confusing. Isn't the library always already loaded by the time we get here (the assert below seems to imply that)? 

JvmtiAgentList::load_agent creates the JvmtiAgent, attempts to load it, and adds it to the agent list if is succeeds. The test that you are looking is checking the list of already loaded agents. Maybe the function name "is_loaded" is the confusion here, maybe a better name is needed.

src/hotspot/share/prims/jvmtiAgentList.cpp
```diff
@@ -220,6 +220,22 @@ void JvmtiAgentList::unload_agents() {
...
 
+// Return true if a library is a loaded agent library
+bool JvmtiAgentList::is_loaded(void* os_lib) {
+  JvmtiAgentList::Iterator it = JvmtiAgentList::agents();
+  while (it.has_next()) {
+    JvmtiAgent* const agent = it.next();
+    if (os_lib == nullptr) {
+       if (agent->is_static_lib() && agent->is_loaded()) {
+         return true;
+       }
```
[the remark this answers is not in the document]
AlanBateman: load_agent_from_executable has a comment to explain how statically linked agents are started, that's why it needs to use agent->is_static_lib().&& agent->is_loaded() here. There isn't currently a way to test this but there is other work going to support static builds so it might be possible to write some automated tests at that point.

test/jdk/com/sun/tools/attach/warnings/DynamicLoadWarningTest.java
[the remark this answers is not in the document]
AlanBateman: The tests have different concerns. The test here is checking that warnings are printed or can be suppressed. Tests for -EnableDynamicAgentLoading are a different concern. I see Serguei is expanding test coverage as part of JDK-8304438 so I think we are good.

src/hotspot/share/prims/jvmtiAgent.cpp
sspitsyn: I see the source of my confusion.
The function `invoke_Agent_OnAttach` is called from the `JvmtiAgent::load`.
But at the time of `invoke_Agent_OnAttach` the agent has not been added to the list yet.
It is added after the `JvmtiAgent::load` with the function `JvmtiAgentList::add`:
```
jint JvmtiAgentList::load_agent(const char* agent_name, const char* absParam,
                           const char* options, outputStream* st) {
  // The abs parameter should be "true" or "false"
  const bool is_absolute_path = (absParam != nullptr) && (strcmp(absParam, "true") == 0);
  JvmtiAgent* const agent = new JvmtiAgent(agent_name, options, is_absolute_path, /* dynamic agent */ true);
  if (agent->load(st)) {
    add(agent);
```
Now I see the code is correct.

test/jdk/com/sun/tools/attach/warnings/DynamicLoadWarningTest.java
```diff
@@ -0,0 +1,308 @@
...
+
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+import static org.junit.jupiter.api.Assertions.*;
+
+class DynamicLoadWarningTest {
+    private static final String JVMTI_AGENT_WARNING = "WARNING: A JVM TI agent has been dynamically loaded";
+    private static final String JAVA_AGENT_WARNING = "WARNING: A Java agent has been loaded dynamically";
```
sspitsyn, line 67: Nit: The warnings can be unified to say "loaded dynamically" or "dynamically loaded" the same way.
Of course, it also impacts the files: `src/hotspot/share/prims/jvmtiAgent.cpp` and `src/java.instrument/share/classes/sun/instrument/InstrumentationImpl.java`.

sspitsyn (approved):
The update looks good to me.
Posted one nit though.
Thanks,
Serguei

src/hotspot/share/prims/jvmtiAgentList.cpp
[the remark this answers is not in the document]
AlanBateman: I've updated this so that is_loaded is split into one function to test if a statically linked agent is already loaded and another for agent libraries. That might be clearer to understand.

src/hotspot/share/prims/jvmtiAgent.cpp
plummercj: > JvmtiAgentList::load_agent creates the JvmtiAgent, attempts to load it, and adds it to the agent list if is succeeds. The test that you are looking is checking the list of already loaded agents. Maybe the function name "is_loaded" is the confusion here, maybe a better name is needed.

Yes, I did notice that there were two different `is_loaded` checks going on. What's not clear is the relation between agent->is_loaded() and JvmtiAgentList::is_loaded(library). I assume that at this point in time the library is not on the list but the agent is loaded, so that's why it works, but it's not clear to me why this is the case. I think a comment here explaining why would be helpful so the reader doesn't need to track down when each of these "loaded" conditions becomes true.

plummercj (approved):
Except for the possible addition of some comments regarding what is meant by "loaded", the changes look good.

plummercj (approved):

PR #13901, opened 2023-05-10
8305189: C2 failed "assert(_outcnt==1) failed: not unique"

rwestrel:
pre/main/post loops are created for an inner loop of a loop nest but
assert predicates cause the main and post loops to be removed. The
OpaqueZeroTripGuard nodes for the loops are not removed: there's no
logic to trigger removal of the opaque nodes once the loops are no
longer there. With the inner loops gone, the outer loop becomes
candidate for optimizations and is unrolled which causes the zero trip
guards of the now removed loops to be duplicated and the opaque nodes
to have more than one use.

The fix I propose is, using logic similar to
`PhaseIdealLoop::eliminate_useless_predicates()`, to check during loop
opts if every OpaqueZeroTripGuard node guards a loop and if not,
remove it.

---------
### Progress

### Issue
 * [JDK-8305189](https://bugs.openjdk.org/browse/JDK-8305189): C2 failed "assert(_outcnt==1) failed: not unique"

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13901/head:pull/13901` \
`$ git checkout pull/13901`

`$ git checkout pull/13901` \
`$ git pull https://git.openjdk.org/jdk.git pull/13901/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13901`

`$ git pr show -t 13901`

</details>

<a href="https://git.openjdk.org/jdk/pull/13901.diff">https://git.openjdk.org/jdk/pull/13901.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13901#issuecomment-1542031469)


src/hotspot/share/opto/loopnode.cpp
```diff
@@ -6145,6 +6175,11 @@ void PhaseIdealLoop::build_loop_late_post_work(Node *n, bool pinned) {
   IdealLoopTree *chosen_loop = get_loop(least);
   if( !chosen_loop->_child )   // Inner loop?
     chosen_loop->_body.push(n);// Collect inner loops
+
+  if (!_verify_only && n->Opcode() == Op_OpaqueZeroTripGuard) {
+    _zero_trip_guard_opaque_nodes.push(n);
```
chhagedorn, line 6180: That's a good idea to collect them newly here for each loop opts pass.

chhagedorn (approved):
That looks reasonable.

src/hotspot/share/opto/opaquenode.cpp
```diff
@@ -54,6 +55,37 @@ Node* Opaque3Node::Identity(PhaseGVN* phase) {
   return this;
 }
 
+CountedLoopNode* OpaqueZeroTripGuardNode::guarded_loop() const {
```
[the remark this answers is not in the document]
rwestrel: I made that change.

src/hotspot/share/opto/loopnode.cpp
chhagedorn: No action required. I was first trying to suggest to move it to `Compile` to the other predicate opaque node lists but then I've thought that this solution here is cleaner. So it was more of a :+1: (I will refactor these predicate lists in the assertion predicate changes - otherwise, I would have suggested to move them here as well at some point).

chhagedorn (approved):
Update looks good!

src/hotspot/share/opto/loopnode.cpp
rwestrel: Thanks for the clarification.

rwestrel:
> That looks reasonable.

Thanks for reviewing this.

TobiHartmann:
I see the following failure with `TestMissingMulLOptimization` from JDK-8299546 and `-XX:StressLongCountedLoop=2000000`:

```
#
#  Internal Error (c:\sb\prod\1684151031\workspace\open\src\hotspot\share\opto\loopnode.cpp:4157), pid=5368, tid=836
#  Error: assert(loop == nullptr) failed

Current CompileTask:
C2:    267   15    b  4       compiler.ccp.TestMissingMulLOptimization::test (101 bytes)

Stack: [0x0000002f4f600000,0x0000002f4f700000]
V  [jvm.dll+0xc53091]  os::win32::platform_print_native_stack+0xf1  (os_windows_x86.cpp:236)
V  [jvm.dll+0xee2a99]  VMError::report+0x1019  (vmError.cpp:815)
V  [jvm.dll+0xee4775]  VMError::report_and_die+0x645  (vmError.cpp:1596)
V  [jvm.dll+0xee4e84]  VMError::report_and_die+0x64  (vmError.cpp:1361)
V  [jvm.dll+0x55053b]  report_vm_error+0x5b  (debug.cpp:191)
V  [jvm.dll+0xadcab2]  PhaseIdealLoop::eliminate_useless_zero_trip_guard+0x2f2  (loopnode.cpp:4157)
V  [jvm.dll+0xad0fb1]  PhaseIdealLoop::build_and_optimize+0x971  (loopnode.cpp:4455)
V  [jvm.dll+0x4ebc51]  Compile::optimize_loops+0x1d1  (compile.cpp:2155)
V  [jvm.dll+0x4de2e8]  Compile::Optimize+0xef8  (compile.cpp:2391)
V  [jvm.dll+0x4db378]  Compile::Compile+0x1458  (compile.cpp:840)
V  [jvm.dll+0x3f05ba]  C2Compiler::compile_method+0x11a  (c2compiler.cpp:121)
V  [jvm.dll+0x4f6a81]  CompileBroker::invoke_compiler_on_method+0x881  (compileBroker.cpp:2268)
V  [jvm.dll+0x4f3ea6]  CompileBroker::compiler_thread_loop+0x396  (compileBroker.cpp:1945)
V  [jvm.dll+0x7f2ff9]  JavaThread::thread_main_inner+0x279  (javaThread.cpp:720)
V  [jvm.dll+0xe5434d]  Thread::call_run+0x1cd  (thread.cpp:222)
V  [jvm.dll+0xc519c2]  os::win32::thread_native_entry+0xa2  (os_windows.cpp:551)
```

chhagedorn:
Would it still be required/useful to keep an `OpaqueZeroTripGuardNode` for such a counted loop inside an infinite loop?

rwestrel:
> Would it still be required/useful to keep an `OpaqueZeroTripGuardNode` for such a counted loop inside an infinite loop?

You're right that it's probably safer to be conservative and rather than remove the opaque node anyway, keep it if we can't tell if it's useful or not. I updated the change with a new commit that does that.

src/hotspot/share/opto/loopnode.cpp
```diff
@@ -4139,6 +4139,45 @@ void PhaseIdealLoop::eliminate_useless_predicates() {
...
+    if (!useful_zero_trip_guard_opaques_nodes.member(opaque)) {
+      IfNode* iff = opaque->if_node();
+      IdealLoopTree* loop = get_loop(iff);
+      while (loop != _ltree_root && loop != nullptr) {
+        loop = loop->_parent;
+      }
+      if (loop == nullptr) {
+        // unreachable from _ltree_root: zero trip guard is in a newly discovered infinite loop.
+        // We can't tell if the opaque node is useful or not
+       assert(guarded_loop == nullptr || guarded_loop->is_in_infinite_subgraph(), "");
```
chhagedorn, line 4170: Indentation:
```suggestion
        assert(guarded_loop == nullptr || guarded_loop->is_in_infinite_subgraph(), "");
```

chhagedorn (approved):
I agree that it is better to be safe here - update looks good!

I'll resubmit some testing.

chhagedorn (approved):
I agree that it is better to be safe here - update looks good!

I'll resubmit some testing.

chhagedorn:
All testing passed!

chhagedorn (approved):

rwestrel:
> All testing passed!

Thanks for the review and testing!

TobiHartmann (approved):
Looks reasonable to me.

rwestrel:
@TobiHartmann thanks for the review

PR #13904, opened 2023-05-10
8307816: Add missing STS to ZGC

rkennke:
Testing in project Lilliput has revealed that ZGC is lacking one STS. Without it, ZGC could reach to already-deflated monitor when trying to fetch a displaced header, in order to get to an object's Klass* (e.g. to get its size).

Testing:
 - [x] hotspot_gc

---------
### Progress

### Issue
 * [JDK-8307816](https://bugs.openjdk.org/browse/JDK-8307816): Add missing STS to ZGC

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13904/head:pull/13904` \
`$ git checkout pull/13904`

`$ git checkout pull/13904` \
`$ git pull https://git.openjdk.org/jdk.git pull/13904/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13904`

`$ git pr show -t 13904`

</details>

<a href="https://git.openjdk.org/jdk/pull/13904.diff">https://git.openjdk.org/jdk/pull/13904.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13904#issuecomment-1542255157)

### DIFF
--- a/src/hotspot/share/gc/z/zRelocate.cpp
+++ b/src/hotspot/share/gc/z/zRelocate.cpp
@@ -23,6 +23,7 @@
 
 #include "precompiled.hpp"
 #include "gc/shared/gc_globals.hpp"
+#include "gc/shared/suspendibleThreadSet.hpp"
 #include "gc/z/zAbort.inline.hpp"
 #include "gc/z/zAddress.inline.hpp"
 #include "gc/z/zBarrier.inline.hpp"
@@ -332,6 +333,10 @@ private:
       _target->reset_for_in_place_relocation();
       _forwarding->set_in_place();
     }
+
+    if (SuspendibleThreadSet::should_yield()) {
+      SuspendibleThreadSet::yield();
+    }
   }
 
 public:
@@ -403,6 +408,7 @@ public:
     ZRelocateClosure<ZRelocateSmallAllocator> small(&_small_allocator);
     ZRelocateClosure<ZRelocateMediumAllocator> medium(&_medium_allocator);
 
+    SuspendibleThreadSetJoiner sts_joiner;
     for (ZForwarding* forwarding; _iter.next(&forwarding);) {
       if (is_small(forwarding)) {
         small.do_forwarding(forwarding);


fisk (changes requested):
So far single generation ZGC has not had any need for the STS joiner, so we have barely used it at all. I think only some class unloading code uses it. With generational ZGC we have added such mechanisms all over the place to allow the generations to synchronize w.r.t. each other in safepoints. So I think in generational ZGC there are no changes required in this regard.
However, in single generation ZGC, the change you have made will cause the JVM to hang if a load barrier executed in a safepoint hits a relocation stall due to memory shortage. That code assumes the concurrent GC thread will make progress with compaction from within the safepoint. In generational ZGC we went through the trouble of adding an STS like mechanism that is aware of relocation stalls, and allows them to be solved from within safepoints, on the mutator's behalf. You would need something like that for single generation ZGC as well.
Other than that, there are probably other places requiring STS that reads the klass pointer, such as during marking and probably reference processing, etc.
Perhaps since this is going to be experimental and mostly aiming towards hitting production in the future, a pragmatic solution is to support only generational ZGC, which shouldn't really require any changes like this. What do you think?

rkennke:
> So far single generation ZGC has not had any need for the STS joiner, so we have barely used it at all. I think only some class unloading code uses it. With generational ZGC we have added such mechanisms all over the place to allow the generations to synchronize w.r.t. each other in safepoints. So I think in generational ZGC there are no changes required in this regard. However, in single generation ZGC, the change you have made will cause the JVM to hang if a load barrier executed in a safepoint hits a relocation stall due to memory shortage. That code assumes the concurrent GC thread will make progress with compaction from within the safepoint. In generational ZGC we went through the trouble of adding an STS like mechanism that is aware of relocation stalls, and allows them to be solved from within safepoints, on the mutator's behalf. You would need something like that for single generation ZGC as well. Other than that, there are probably other places requiring STS that reads the klass pointer, such as during marking and probably reference processing, etc. Perhaps since this is going to be experimental and mostly aiming towards hitting production in the future, a pragmatic solution is to support only generational ZGC, which shouldn't really require any changes like this. What do you think?

Ok, if generational ZGC comes with all the relevant synchronizations, then I will wait for that, and withdraw this PR. Thanks for the explanations.

tschatzl:
I already added a comment to the CR about that this is likely only a problem with regular ZGC+Lilliput and for now the consensus is that are not intending to support that combination.

PR #13914, opened 2023-05-10
JDK-8305457 java.io.SimpleIO - common I/O tasks simplified

JimLaskey:
In unison with the Amber onramp project, there are many mundane I/O tasks that leave developer searching StackOverflow for the appropriate idiom. The intent of the java.io.SimpleIO class is to package these idioms as easy to use static methods. 

- read/write from/to the console 
- read/write from/to (UTF-8) text files 

These methods should not require wrapping in try blocks, but allow the handling of UncheckedIOExceptions.

---------
### Progress

### Issue
 * [JDK-8305457](https://bugs.openjdk.org/browse/JDK-8305457): java.io.SimpleIO - common I/O tasks simplified (**Enhancement** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13914/head:pull/13914` \
`$ git checkout pull/13914`

`$ git checkout pull/13914` \
`$ git pull https://git.openjdk.org/jdk.git pull/13914/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13914`

`$ git pr show -t 13914`

</details>

<a href="https://git.openjdk.org/jdk/pull/13914.diff">https://git.openjdk.org/jdk/pull/13914.diff</a>

</details>


src/java.base/share/classes/java/io/SimpleIO.java
```diff
@@ -0,0 +1,1037 @@
...
+     * @implSpec Each line will be followed by the platform line separator.
+     *
+     * @param path  file name string of file to be written
+     * @param lines  list of lines to be written to the file
+     * @throws UncheckedIOException wrapping an IOException if an io error occurs.
+     */
+    public static void writeLines(Path path, List<String> lines) {
+        Objects.requireNonNull(path, "path must not be null");
+        Objects.requireNonNull(lines, "lines must not be null");
+        try (PrintWriter out = new PrintWriter(path.toFile())) {
```
AlanBateman, line 666: I assume this can use Files.write(path, lines).

src/java.base/share/classes/java/io/SimpleIO.java
```diff
@@ -0,0 +1,1037 @@
...
+public final class SimpleIO {
+    /**
+     * Private constructor.
+     */
+    private SimpleIO() {
+        throw new AssertionError("private constructor");
+    }
+
+    /**
+     * Fetch jline LineReader lazily.
```
AlanBateman, line 110: This coupling to jline is very messy, have you looked at using System.console().reader() which will use jline when available?

src/java.base/share/classes/java/io/SimpleIO.java
```diff
@@ -0,0 +1,1037 @@
...
+     */
+    public static String input(String prompt) {
+        Objects.requireNonNull(prompt, "prompt must not be null");
+
+        if (LineReader.hasLineReader()) {
+            String input = LineReader.readLine(prompt);
+            return input != null ? input : "";
+        } else {
+            System.out.print(prompt);
+            Scanner scanner = new Scanner(System.in);
```
AlanBateman, line 222: You might need Console.charset here, or just look at the example in the Scanner docs that gets a reader.

src/java.base/share/classes/java/io/SimpleIO.java
```diff
@@ -0,0 +1,1037 @@
...
+     * @throws UncheckedIOException wrapping an IOException if an io error occurs.
+     */
+    public static void write(String filename, String content) {
+        Objects.requireNonNull(filename, "filename must not be null");
+        Objects.requireNonNull(content, "content must not be null");
+        write(Path.of(filename), content);
+    }
+
+    /**
+     * The contents of the file at {@code path} are read as a string.
```
AlanBateman, line 571: This method writes, it doesn't read.

src/java.base/share/classes/java/io/SimpleIO.java
```diff
@@ -0,0 +1,1037 @@
...
+     * @implSpec Line terminators are normalized the platform line separator.
+     *
+     * @param path  path of file to be written
+     * @param content  string content of the file
+     * @throws UncheckedIOException wrapping an IOException if an io error occurs.
+     */
+    public static void write(Path path, String content) {
+        Objects.requireNonNull(path, "path must not be null");
+        Objects.requireNonNull(content, "content must not be null");
+        try (PrintWriter out = new PrintWriter(path.toFile())) {
```
AlanBateman, line 603: Could this use Files.writeString(Path, String)?

src/java.base/share/classes/java/io/SimpleIO.java
```diff
@@ -0,0 +1,1037 @@
...
+     * will print on the output stream as:
+     * {@snippet lang="text":
+     * "Hope" is the thing with feathers
+     * That perches in the soul
+     * And sings the tune without the words
+     * And never stops - at all
+     *
+     * }
+     *
+     * @param filename  file name of file to be read
```
AlanBateman, line 427: The audience may be new developers but "file name" could be mislead as it can be a full file path, might have to tweak this a bit to get right.

src/java.base/share/classes/java/io/SimpleIO.java
```diff
@@ -0,0 +1,1037 @@
...
+     * @param path  path of file to be read
+     * @return Content read from a file
+     * @throws UncheckedIOException wrapping an IOException if an io error occurs.
+     *
+     * @implSpec Line terminators are normalized to '\n'.
+     */
+    public static String read(Path path) {
+        Objects.requireNonNull(path, "path must not be null");
+        return readLines(path).stream()
+                .collect(Collectors.joining("\n", "", "\n"));
```
AlanBateman, line 471: See also Files.readString.

src/java.base/share/classes/java/io/SimpleIO.java
JimLaskey: Thx

src/java.base/share/classes/java/io/SimpleIO.java
JimLaskey: This code was the model for `System.console().reader()`, just haven't caught up yet.

JimLaskey: Actually, console does not provide the same functionality, that is not full jline support.

src/java.base/share/classes/java/io/SimpleIO.java
JimLaskey: Switching to using `System.console().readLine()` as fallback.

src/java.base/share/classes/java/io/SimpleIO.java
JimLaskey: Thx

src/java.base/share/classes/java/io/SimpleIO.java
JimLaskey: Thx

src/java.base/share/classes/java/io/SimpleIO.java
JimLaskey: Thx

src/java.base/share/classes/java/io/SimpleIO.java
JimLaskey: Does `Files.readString` normalize the line terminators?

src/java.base/share/classes/java/io/SimpleIO.java
JimLaskey:
```
    public static void main(String... args) {
        String line = input("*> ");
        System.out.println(line);
        line = input("*> ");
        System.out.println(line);
    }
```

```
>> java Test.java
*> This is a line.
This is a line.
*> This is a line.
This is a line.
```

```
    public static void main(String... args) {
        String line = System.console().readLine("*> ");
        System.out.println(line);
        line = System.console().readLine("*> ");
        System.out.println(line);
    }
```

```
>> java Test.java
>> This is a line.^[[D^[[D^[[D^[[D
This is a line.
*> ^[[A^[[A^[[A^[[A^[[A
```

src/java.base/share/classes/java/io/SimpleIO.java
bplb: Also, it could be named `writeLines` instead of `write` as the signature will disambiguate (which cannot be done for `readLines` vs. `read`).

bplb:
> The intent of the java.io.Simplified class

I think `s/Simplified/SimpleIO/` here, no?

src/java.base/share/classes/java/io/SimpleIO.java
```diff
@@ -0,0 +1,1031 @@
...
+     * }
+     * will interact with the console as:
+     * {@snippet lang="text":
+     * Name>> Jean
+     * Jean
+     * }
+     *
+     * @param prompt string contain prompt for input, may be the empty string
+     * @return a string of characters read in from input or empty string if user aborts input
+     * @throws NullPointerException if prompt is null
```
bplb, line 212: Throwing `NullPointerException` is indicated only here and in `printLines`, but I think it would be thrown in numerous other places, right? If so, then perhaps add a statement to that effect to the class level specification and remove the various `@throws NullPointerException`. Either that or add the latter everywhere it is needed.

src/java.base/share/classes/java/io/SimpleIO.java
```diff
@@ -0,0 +1,1031 @@
...
+            if (scanner.hasNextBoolean()) {
+                return scanner.nextBoolean();
+            }
+        }
+    }
+
+    private static final Object[] NULL_ARRAY = new Object[1];
+
+    /**
+     * Single space print values to the output stream.
```
bplb, line 336: Should this instead state "to the console's output stream."?

src/java.base/share/classes/java/io/SimpleIO.java
```diff
@@ -0,0 +1,1031 @@
...
+     *
+     * }
+     *
+     * @param path  path of file to be read
+     * @return Content read from a file
+     * @throws UncheckedIOException wrapping an IOException if an io error occurs.
+     *
+     * @implSpec Line terminators are normalized to '\n'.
+     */
+    public static String read(Path path) {
```
bplb, line 466: Unless I missed something, this would be the first class in the `java.io` package to have a `java.nio.file.Path` parameter. The only other use of `Path` is by `File::toPath`. Should `Path` be replaced with `File` for parameters?

src/java.base/share/classes/java/io/SimpleIO.java
JimLaskey: Need -Djdk.console=jdk.internal.le to activate.

AlanBateman: > Need -Djdk.console=jdk.internal.le to activate.

We had to disable this by default because the jline code was forking processes to get tty settings. Jan is fixing that and once that is in then it should be possible to re-enable the provider by default. Then, I hope, the code in SimpleIO can be removed.

src/java.base/share/classes/java/io/SimpleIO.java
AlanBateman: > Unless I missed something, this would be the first class in the `java.io` package to have a `java.nio.file.Path` parameter. The only other use of `Path` is by `File::toPath`. Should `Path` be replaced with `File` for parameters?

I think we should expect new APIs to use Path rather than File but for this API, there is an argument that all the methods should take a path string rather than having two methods for each op.

src/java.base/share/classes/java/io/SimpleIO.java
JimLaskey: Switched to console and forcing `jdk.console=jdk.internal.le` for the time being.

src/java.base/share/classes/java/io/SimpleIO.java
JimLaskey: Fallback no longer required

src/java.base/share/classes/java/io/SimpleIO.java
JimLaskey: The concern is to make sure students don't get surprised by platform terminators.

src/java.base/share/classes/java/io/SimpleIO.java
bplb: Also, the `SimpleIO::{read,write}*(Path...)` methods do not add any value over their `java.nio.file.Files` counterparts.

src/java.base/share/classes/java/io/SimpleIO.java
```diff
@@ -0,0 +1,962 @@
...
+     *
+     * @param prompt string contain prompt for input, may be the empty string
+     * @return a token entered by user
+     * @throws NullPointerException if prompt is null
+     */
+    public static String inputNext(String prompt) {
+        Objects.requireNonNull(prompt, "prompt must not be null");
+        while (true) {
+            String input = input(prompt);
+            var scanner = scanner(input);
```
agluszak, line 147: why use `var` in some places and not others?

src/java.base/share/classes/java/io/SimpleIO.java
```diff
@@ -0,0 +1,962 @@
...
+ *         printLine("Hello", name);
+ *     }
+ * }
+ * }
+ * The file I/O method methods are designed to work with UTF-8 text files. The developer can
+ * specify the file by either a simple file name string or by using a {@link Path}.
+ * The developer also has the choice of working with the whole content as a string or
+ * as a string list representing lines of content.
+ * <p>
+ * File I/O is futher simplifed by throwing {@link UncheckedIOException} for caught
```
buerkle, line 67:
```suggestion
 * File I/O is further simplified by throwing {@link UncheckedIOException} for caught
```

hoat4:
What is the benefit of creating a new class instead of putting simple console read/write operations in the existing java.io.Console class as static methods? 

Furthermore, I observed many people learning to program don't understand what does "print" and "IO" mean. "write" is usually better understandable by beginners than "print", because print is commonly associated with physical printing for those who don't have prior coding experience.

src/java.base/share/classes/java/io/SimpleIO.java
```diff
@@ -0,0 +1,961 @@
...
+ * }
+ * }
+ * @since  21
+ */
+public final class SimpleIO {
+    /**
+     * Private constructor.
+     */
+    private SimpleIO() {
+        throw new AssertionError("private constructor");
```
stsypanov, line 102: Wouldn't it be to more reasonable to throw `UnsupportedOperationException` here?

Adyel: In his 'Effective Java' book, Joshua Bloch suggests using `AssertionError` because it sends a clear message that the exception should not be caught.

PR #13917, opened 2023-05-10
8307547: Support variant collations

naotoj:
The fix to https://bugs.openjdk.org/browse/JDK-8306927 switched the default collation for Swedish to the modern one. In order to provide a means for users who need the old collation, this PR intends to make `Collator` recognize the `co` Unicode locale extension so that multiple implementations for a locale can be provided. I would also like reviews for the corresponding CSR.

---------
### Progress
- [x] Change requires CSR request [JDK-8307565](https://bugs.openjdk.org/browse/JDK-8307565) to be approved

### Issues
 * [JDK-8307547](https://bugs.openjdk.org/browse/JDK-8307547): Support variant collations
 * [JDK-8307565](https://bugs.openjdk.org/browse/JDK-8307565): Support variant collations (**CSR**)

### Reviewers
 * [Steven Loomis](https://openjdk.org/census#srl) (@srl295 - Committer) ⚠️ Review applies to [4ab39664](https://git.openjdk.org/jdk/pull/13917/files/4ab39664620ca0ff9bc39be23426da9c9489dfb6)
 * [Justin Lu](https://openjdk.org/census#jlu) (@justin-curtis-lu - Committer) ⚠️ Review applies to [4ab39664](https://git.openjdk.org/jdk/pull/13917/files/4ab39664620ca0ff9bc39be23426da9c9489dfb6)
 * [Alan Bateman](https://openjdk.org/census#alanb) (@AlanBateman - **Reviewer**) ⚠️ Review applies to [4ab39664](https://git.openjdk.org/jdk/pull/13917/files/4ab39664620ca0ff9bc39be23426da9c9489dfb6)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13917/head:pull/13917` \
`$ git checkout pull/13917`

`$ git checkout pull/13917` \
`$ git pull https://git.openjdk.org/jdk.git pull/13917/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13917`

`$ git pr show -t 13917`

</details>

<a href="https://git.openjdk.org/jdk/pull/13917.diff">https://git.openjdk.org/jdk/pull/13917.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13917#issuecomment-1542759904)


justin-curtis-lu:
LGTM, does it make sense to add some other locale's collation variants?

naotoj:
Thanks, Justin

> does it make sense to add some other locale's collation variants?

Maybe, but I am not seeing any requests for adding more collation variants. This enhancement is mainly for those who need the old Swedish collation as a workaround

src/java.base/share/classes/java/text/Collator.java
```diff
@@ -114,10 +113,22 @@
...
+ * <a href="https://www.unicode.org/reports/tr35/#UnicodeCollationIdentifier">
+ * Unicode collation identifier</a> in the specified {@code Locale}.
+ * For example:
+ * <blockquote>
+ * {@snippet lang = java:
+ * Collator.getInstance(Locale.forLanguageTag("sv-u-co-trad"));
+ * }
+ * </blockquote>
+ * may return a {@code Collator} instance with the traditional sorting, which
+ * gives 'v' and 'w' the same sorting order.
```
AlanBateman, line 131: The overall intent and wording looks okay but I wonder if it should be in the getInstance methods instead. Also wondering whether impNote is right because it's both informational for developers using the API and useful to implementors.

naotoj: Thanks. Will move it out of the class description into each `getInstance()` method, changing @implNote to @apiNote

AlanBateman:
The update, and API note in Collator.getInstance(Locale), looks okay to me. I'm just wondering where a developer might go to get a definitive list, i.e. aside from this API note, how would they know that "-trad" or "-traditional" can be used to configure the ordering. Locale.forLanguageTag supports more than BCP 47 language tag strings so is this considered a private use language tag.

naotoj:
> I'm just wondering where a developer might go to get a definitive list, i.e. aside from this API note, how would they know that "-trad" or "-traditional" can be used to configure the ordering. Locale.forLanguageTag supports more than BCP 47 language tag strings so is this considered a private use language tag.

I think those should go into Oracle JDK's `Supported Locales` document. Created a task to include them (https://bugs.openjdk.org/browse/JDK-8308018)

srl295:
> > I'm just wondering where a developer might go to get a definitive list, i.e. aside from this API note, how would they know that "-trad" or "-traditional" can be used to configure the ordering. Locale.forLanguageTag supports more than BCP 47 language tag strings so is this considered a private use language tag.
> 
> I think those should go into Oracle JDK's `Supported Locales` document. Created a task to include them (https://bugs.openjdk.org/browse/JDK-8308018)

`-u-co` is defined [here](https://www.unicode.org/reports/tr35/#UnicodeCollationIdentifier) (UTS#35) and the keys are in the datafile [here (link to `main` branch!)](https://github.com/unicode-org/cldr/blob/main/common/bcp47/collation.xml) - note the `since=` attribute, these are very stable.

naotoj:
Thanks @srl295 
I am just curious how CLDR handles the default switch of Swedish collation. Now the `traditional` collation used to be `standard`, and `standard` used to be `reformed`. How do apps specify their desired collation in Swedish, regardless of CLDR versions?

srl295:
> Thanks @srl295
> I am just curious how CLDR handles the default switch of Swedish collation. Now the `traditional` collation used to be `standard`, and `standard` used to be `reformed`. How do apps specify their desired collation in Swedish, regardless of CLDR versions?

I don't know off the top of my head but digging around I found https://unicode-org.atlassian.net/browse/CLDR-7088 and cross linked some tickets. Can you ask around on those?

justin-curtis-lu (approved):

AlanBateman:
> > so is this considered a private use language tag
> 
> Not private use at all. The `-u-` subtag is registered, and the links above are from the registrar, see
> 
> * https://www.rfc-editor.org/info/bcp47
> * https://www.rfc-editor.org/rfc/rfc6067

Good, I don't think I have any other questions on this now.

AlanBateman (approved):

naotoj:
Thanks, @AlanBateman for the review.
@srl295 thanks. I found CLDR's collation documents are a bit outdated. Filed an issue here: https://unicode-org.atlassian.net/jira/software/c/projects/CLDR/issues/CLDR-16638

src/java.base/share/classes/java/text/Collator.java
```diff
@@ -232,6 +227,18 @@ public static synchronized Collator getInstance() {
 
     /**
      * Gets the Collator for the desired locale.
+     * @apiNote Implementations of {@code Collator} class may produce
```
jddarcy, line 230: For less l10n savvy readers, I suggest providing a bit more context in the apiNote, such as by explicitly mentioning Swedish and showing what getting the default collator would look like.

PR #13923, opened 2023-05-11
8307885

plummercj:
When using -Xcomp, the 5 second sleep in the debuggee main thread is not long enough to guarantee using SIGTERM on the debuggee will make it exit(143) before the main thread exits cleanly with exitValue 0. The easiest solution is to just allow exitValue 0. See the CR for more details on other solutions I rejected.

### DIFF
--- a/test/jdk/com/sun/jdi/ConnectedVMs.java
+++ b/test/jdk/com/sun/jdi/ConnectedVMs.java
@@ -31,9 +31,6 @@
  * @library /test/lib
  * @run build TestScaffold VMConnection TargetListener TargetAdapter
  * @run compile -g InstTarg.java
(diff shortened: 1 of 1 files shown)


kevinjwalls (approved):
Reasonable compromise 8-)

passName doesn't get passed to InstTarg, otherwise it could wait at the end in the "Kill" case, as if we want to see a process exit in response to a SIGTERM, then that process needs to wait around.  But tolerating the signal not terminating it due to the race should be reliable also. 8-)

plummercj:
I'm going to push this a bit early since it is creating noise in our tier4 CI.

plummercj:
Thanks for the reviews Kevin and Serguei!

/integrate

PR #13924, opened 2023-05-11
8302673: [SuperWord] MaxReduction and MinReduction should vectorize for int

robcasloz:
The canonicalization of MinI/MaxI chains into right-spline graphs within `MinINode/MaxINode::Ideal()` inhibits the vectorization of reductions of these nodes. This changeset reworks `MinINode/MaxINode::Ideal()` to perform the same algebraic optimizations without the need for canonicalization, re-enabling auto-vectorization of MinI/MaxI reductions. This is achieved by handling all four permutations of the targeted Ideal subgraph induced by the commutativity of MinI/MaxI directly. The algorithm (for the MaxI case, the MinI case is analogous) tries to apply two Ideal graph rewrites in the following order, where `c0` and `c1` are constants and `MAX` is a compile-time operation:
1. `max(x + c0, max(x + c1, z))` (or a permutation of it) to `max(x + MAX(c0, c1), z)`.
2. `max(x + c0, x + c1)` (or a permutation of it) to `x + MAX(c0, c1)`.

Here is an example of the four permutations handled in step 1 with `x = RShiftI`, `c0 = 100` or `150`, `c1 = 150` or `100`, and `z = ConI (#int:200)`:

two-level-idealization

Here is an example of the two permutations handled in step 2 with `x = RShiftI`, `c0 = 10` or `11`, and `c1 = 11` or `10`:

one-level-idealization

The changeset implements `MinINode/MaxINode::Ideal()` in a common method `MaxNode::IdealI()`, since the algorithm is symmetric for both node types. The changeset also extends the existing MinI/MaxI Idealization tests with positive tests for all targeted permutations and negative tests, and adds a new test (contributed by @jbhateja) to assert that MinI/MaxI reductions are vectorized.

##### Functionality

- tier1-5, stress test, fuzzing (windows-x64, linux-x64, linux-aarch64, macosx-x64, macosx-aarch64; release and debug mode).

##### Performance

- Tested performance on a set of standard benchmark suites (DaCapo, SPECjbb2015, SPECjvm2008). No significant change was observed.

---------
### Progress

### Issue
 * [JDK-8302673](https://bugs.openjdk.org/browse/JDK-8302673): [SuperWord] MaxReduction and MinReduction should vectorize for int

### Reviewers
 * [Emanuel Peter](https://openjdk.org/census#epeter) (@eme64 - Committer)

### Contributors
 * Jatin Bhateja `<jbhateja@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13924/head:pull/13924` \
`$ git checkout pull/13924`

`$ git checkout pull/13924` \
`$ git pull https://git.openjdk.org/jdk.git pull/13924/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13924`

`$ git pr show -t 13924`

</details>

<a href="https://git.openjdk.org/jdk/pull/13924.diff">https://git.openjdk.org/jdk/pull/13924.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13924#issuecomment-1545337007)


test/hotspot/jtreg/compiler/loopopts/superword/MinMaxRed_Int.java
```diff
@@ -0,0 +1,107 @@
...
+            System.out.println("Success");
+        } else {
+            throw new AssertionError("Failed");
+        }
+    }
+
+    public static void ReductionInit(int[] a, int[] b) {
+        for (int i = 0; i < a.length; i++) {
+            a[i] = -i;
+            b[i] = i;
```
eme64, line 82: That means that `a[i] * b[i] == -i*i`, and get increasingly smaller. I think it would be better if this was a bit more random, and not biased to the maximum always being at the beginning and the minimum at the end.

eme64: Plus, we should try to cover the whole int range, or at least as much as possible.
One solution: just pick two random ints, and then add/subtract them before min/max.

src/hotspot/share/opto/addnode.cpp
```diff
@@ -1086,166 +1086,98 @@ static bool can_overflow(const TypeInt* t, jint c) {
...
+  if (n->Opcode() == Op_AddI && n->in(2)->is_Con()) {
+    const Type* t = n->in(2)->bottom_type();
+    if (t == Type::TOP) {
+      return nullptr;
+    }
+    *con = t->is_int()->get_con();
+    n = n->in(1);
   }
+  return n;
+}
```
eme64, line 1103: This was confusing to read.

Why not make it a `as_add_constant`, and explicitly always set the `con`:
```
case AddI(x, top): return <nullptr, 0>
case AddI(x, int_con): return <x, int_con>
default: return <n, 0>
```
That would make it easier to argue what the value of `con` is after the call - it does certainly not depend on what it was before.

eme64: Add a comment that explains that for `top` we will bail out - for that we can check `nullptr`.
In the other cases, we know that `n == AddI(x, int_con)`.

src/hotspot/share/opto/addnode.cpp
```diff
@@ -1086,166 +1086,98 @@ static bool can_overflow(const TypeInt* t, jint c) {
...
+  jint x_off = 0;
+  Node* x = constant_add_input(l, &x_off);
+  if (x == nullptr) return nullptr;
+  jint y_off = 0;
+  Node* y = constant_add_input(r, &y_off);
+  if (y == nullptr) return nullptr;
+  // Try to transform opcode(x + x_off, opcode(y + y_off, z)) (in any of its
+  // four possible permutations given by opcode's commutativity) into
+  // opcode(x + opcode(x_off, y_off), z), where opcode is either MinI or MaxI,
+  // if x == y and the additions can't overflow.
```
eme64, line 1132: Ok, effectively we have 5, not just 4 cases here:
```
opcode(x + x_off, opcode(y + y_off, z))
opcode(x + x_off, opcode(z, y + y_off))
opcode(opcode(y + y_off, z), x + x_off)
opcode(opcode(z, y + y_off), x + x_off)
opcode(x + x_off, y + y_off)
```

I find the nested for-loop quite confusing. Maybe packing the inner stuff into a separate function could work?

Maybe doing it explicit could also do the trick?
```
// Check for opcode(x + x_con,  y + y_con), no z
if (in(1)->Opcode() == Op_AddI && in(2)->Opcode() == Op_AddI) {
  Node* ret = try_fold(opcode, in(1), in(2), nullptr);
  if (ret != nullptr) { return ret; }
}

// Check for these 4 cases, equivalent to opcode3(addx, addy, z)
// opcode(x + x_con, opcode(y + y_con, z))
// opcode(x + x_con, opcode(z, y + y_con))
// opcode(opcode(y + y_con, z), x + x_con)
// opcode(opcode(z, y + y_con), x + x_con)
for (uint i = 1; i < 2; i++) {
  Node* addx = in(i);
  Node* other = in(i == 1 ? 2 : 1); // or just "2-i"
  if (addx->Opcode() != Op_AddI || other->Opcode() != opcode) { continue; }
  for (uint j = 1; i < 2; j++) {
    Node* addy = other->in(j);
    Node* z = other->in(j == 1 ? 2 : 1);
    if (addy->Opcode() != Op_AddI) { continue; }
    // We have opcode3(addx, addy, z)
    Node* ret = try_fold(opcode, addx, addy, z);
    if (ret != nullptr) { return ret; }
  }
}
```
Where we have
```
Node* try_fold(int opcode, Node* addx, Node* addy, Node* z = nullptr) {
  jint addx_con = 0;
  jint addy_con = 0;
  Node* addx_var = as_add_constant(addx, &addx_con);
  Node* addy_var = as_add_constant(addy, &addy_con);
  if (addx_var == nullptr || addy_var == nullptr) {
    // found a top
    return nullptr;
  }
  // could even check addx_var != addy_var, then we don't have to do that inside...
  Node* folded = extract_addition(phase, addx_var, addx_con, addy_var, addy_con, opcode);
  if (z != nullptr) {
    folded = opcode(folded, z);
  }
  return folded;
}
```

Maybe this does a few more calls to `as_add_constant` than strictly necessary, but it is a bit easier to understand, right?

eme64:
Nice work with the tests, it's good to have some specific IR tests there!

I hope we can also generalize this for `MaxL/MinL` (once we do this [JDK-8307513](https://bugs.openjdk.org/browse/JDK-8307513)) - I think that is now also going to be easier with your refactoring towards `MaxNode::IdealI`.

src/hotspot/share/opto/addnode.cpp
robcasloz: Thanks, I went with the tuple return option as suggested and also simplified the semantics of the function. Hope it is clearer now!

src/hotspot/share/opto/addnode.cpp
robcasloz: Thanks for the feedback, I agree the first refactor was not as clear as it should be, possibly because I derived it quite mechanically from the existing code. I have reworked `MaxNode::IdealI()` using more explicit naming, a functional and simplified version of `as_add_with_constant()`, and a clearer separation of the different optimizations that are applied. Hope it is more readable now.

robcasloz:
@eme64 Sorry for the delay, I have addressed your feedback now! Please let me know if you find the new version more readable.

src/hotspot/share/opto/addnode.cpp
```diff
@@ -1095,185 +1096,154 @@ Node* MaxNode::build_min_max_diff_with_zero(Node* a, Node* b, bool is_max, const
...
+      Node* inner_op = in(inner_op_index);
+      Node* inner_add = inner_op->in(inner_add_index);
+      ConstAddOperands inner_add_operands = as_add_with_constant(inner_add);
+      if (inner_add_operands.first == nullptr) {
+        continue;
+      }
+      Node* inner = inner_add_operands.first;
+      jint inner_off = inner_add_operands.second;
+      // Try to extract the inner add.
+      Node* add_extracted = extract_add(phase, inner, inner_off, outer, outer_off);
```
eme64, line 1182: Optional: You could also leave `outer_add_operands` and `inner_add_operands` packed in the `Pair`, and pass it as such into `extract_add`. Could reduce the number of lines a bit here.

src/hotspot/share/opto/addnode.cpp
```diff
@@ -1095,185 +1096,154 @@ Node* MaxNode::build_min_max_diff_with_zero(Node* a, Node* b, bool is_max, const
...
+      if (add_extracted == nullptr) {
+        continue;
+      }
+      Node* add_transformed = phase->transform(add_extracted);
+      Node* inner_other = inner_op->in(inner_add_index == 1 ? 2 : 1);
+      if (opcode == Op_MinI) {
+        return new MinINode(add_transformed, inner_other);
+      } else {
+        return new MaxINode(add_transformed, inner_other);
+      }
```
eme64, line 1192: Could you make use of `MaxNode::build_min_max`?

src/hotspot/share/opto/addnode.cpp
```diff
@@ -1095,185 +1096,154 @@ Node* MaxNode::build_min_max_diff_with_zero(Node* a, Node* b, bool is_max, const
...
+  Node* c = n->in(2);
+  if (!c->is_Con()) {
+    return null;
+  }
+  const Type* c_type = c->bottom_type();
+  if (c_type == Type::TOP) {
+    return null;
+  }
+  return ConstAddOperands(x, c_type->is_int()->get_con());
+}
```
eme64, line 1141: This is what it was on my last review:

```
// Return:
// <x, C>,       if n is of the form x + C, where 'C' is a non-TOP constant;
// <nullptr, _>, if n is of the form x + C, where 'C' is a TOP constant;
// <n, con>      otherwise.
static Node* constant_add_input(Node* n, jint* con) {
  if (n->Opcode() == Op_AddI && n->in(2)->is_Con()) {
    const Type* t = n->in(2)->bottom_type();
    if (t == Type::TOP) {
      return nullptr;
    }
    *con = t->is_int()->get_con();
    n = n->in(1);
  }
  return n;
}
```
Here, you used to also allow packing just a single `n`, and leave the constant as `zero`. Did you remove this possibility on purpose? Now `n` must be an `AddI`.

This used to allow cases like this to be folded:
`max(max(a, b),  a + 1)      ->      max(a + max(0, 1), b)`

Or am I missing something? Do you have tests for this case?

eme64:
@robcasloz it looks much better, thanks for refactoring :)
I have left a few more comments.

src/hotspot/share/opto/addnode.cpp
robcasloz: Done (extracting common functionality into `build_min_max_int()`), thanks!

src/hotspot/share/opto/addnode.cpp
robcasloz: Good catch, thanks Emanuel! The simplification of `as_add_with_constant()` was too aggressive, and the lost optimizations were not caught by any test or noticeable regression on the standard Java benchmark suites. I added more test cases to catch them now and reverted `as_add_with_constant()` to return `<n, 0>`  for non-additions.

robcasloz:
@eme64 Thanks for your thorough review, I addressed your new comments and suggestions. Please let me know if there is anything else you would like to change.

src/hotspot/share/opto/addnode.cpp
```diff
@@ -1095,185 +1098,130 @@ Node* MaxNode::build_min_max_diff_with_zero(Node* a, Node* b, bool is_max, const
...
+        continue;
+      }
+      // Try to extract the inner add.
+      Node* add_extracted = extract_add(phase, inner_add_operands, outer_add_operands);
+      if (add_extracted == nullptr) {
+        continue;
+      }
+      Node* add_transformed = phase->transform(add_extracted);
+      Node* inner_other = inner_op->in(inner_add_index == 1 ? 2 : 1);
+      return build_min_max_int(add_transformed, inner_other, opcode == Op_MaxI);
```
eme64, line 1186: Did something prevent you from directly using `MaxNode::build_min_max`?

eme64 (approved):
Now it looks much cleaner, thanks for the new changes!

vnkozlov (approved):
Looks good to me too.

robcasloz:
Thanks for reviewing, Emanuel and Vladimir! I noticed that commit 29922ea in this PR accidentally fixes a latent issue in `MaxNode::build_min_max()`. I have reported and fixed the issue separately in [JDK-8309295](https://bugs.openjdk.org/browse/JDK-8309295). Will rebase this PR and submit a new version after re-testing.

src/hotspot/share/opto/addnode.cpp
robcasloz: Technically I think that should be possible, but I find that extracting the core logic into a separate function is more readable (makes it straightforward to understand the effect of the call in `MaxNode::IdealI()`) and efficient (avoids a redundant application of `PhaseGVN::transform()` to the newly created `MinI`/`MaxI` nodes).

vnkozlov (approved):
Update is good.

eme64 (approved):
Looks good, thanks for all the refactoring!

robcasloz:
Emanuel, Vladimir, thanks again for reviewing, and particularly thanks Emanuel for the useful feedback!

PR #13927, opened 2023-05-11
Read length of SNIServerName as unsigned.

jamsilva:
The RFC allows Server Names to have up to 255 characters. The current code fails for names with length > 127. See https://www.rfc-editor.org/rfc/rfc6066#section-3


jamsilva:
Can I get some eyes on this PR and related ones for jdk17u and jdk20u?
This shouldn't be difficult to review: a single line change, justified by the relevant RFC. Attribution shouldn't even be a real concern for such a small change...

PR #13928, opened 2023-05-11
8306806: JMX agent with JDP enabled won't start when PerfData is disabled

kevinjwalls:
Java Discovery Protocol (perhaps a hidden feature, but maybe should be more widely known!) and -XX:-UsePerfData together cause a failure to startup.

PerfData is the mechanism for communicating the URL and other properties for remote management, so disabling PerfData clearly breaks this.  But there should be a clearer message, and not a fatal error.

e.g.
$ java -XX:-UsePerfData  -Dcom.sun.management.jmxremote.autodiscovery=true -Dcom.sun.management.jmxremote.port=0 -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false --version
Warning: Exception thrown by the agent : JDP not starting, PerfData not available: Empty PerfData buffer
java 21-internal 2023-09-19 LTS
...etc...

The use of PerfData is superior to the previous hard-coding of knowledge of the URL protocol and structure, and is required to communicate a dynamically assigned port (JDK-8167337).

---------
### Progress

### Issue
 * [JDK-8306806](https://bugs.openjdk.org/browse/JDK-8306806): JMX agent with JDP enabled won't start when PerfData is disabled

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13928/head:pull/13928` \
`$ git checkout pull/13928`

`$ git checkout pull/13928` \
`$ git pull https://git.openjdk.org/jdk.git pull/13928/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13928`

`$ git pr show -t 13928`

</details>

<a href="https://git.openjdk.org/jdk/pull/13928.diff">https://git.openjdk.org/jdk/pull/13928.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13928#issuecomment-1543883538)

### DIFF
--- a/src/jdk.management.agent/share/classes/jdk/internal/agent/Agent.java
+++ b/src/jdk.management.agent/share/classes/jdk/internal/agent/Agent.java
@@ -35,6 +35,7 @@ import java.lang.reflect.Method;
 import java.net.InetAddress;
 import java.net.MalformedURLException;
 import java.net.UnknownHostException;
+import java.nio.BufferUnderflowException;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.text.MessageFormat;
@@ -502,8 +503,8 @@ public class Agent {
             Map<String,String> remoteProps = null;
             try {
                 remoteProps = ConnectorAddressLink.importRemoteFrom(0);
-            } catch (IOException ioe) {
-                warning(AGENT_EXCEPTION, "JDP not starting, PerfData not available: " + ioe.getMessage());
+            } catch (BufferUnderflowException bue) {
+                warning(AGENT_EXCEPTION, "JDP not starting, PerfData not available");
             }
             if (remoteProps != null) {
                 String jmxUrlStr = remoteProps.get("sun.management.JMXConnectorServer.0.remoteAddress");
--- a/src/jdk.management.agent/share/classes/jdk/internal/agent/Agent.java
+++ b/src/jdk.management.agent/share/classes/jdk/internal/agent/Agent.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2003, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2003, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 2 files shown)


src/jdk.management.agent/share/classes/jdk/internal/agent/Agent.java
```diff
@@ -499,12 +499,17 @@ private static void startDiscoveryService(Properties props)
...
-            String jmxUrlStr = remoteProps.get("sun.management.JMXConnectorServer.0.remoteAddress");
-
-            String instanceName = props.getProperty("com.sun.management.jdp.name");
-
-            JdpController.startDiscoveryService(address, port, instanceName, jmxUrlStr);
+            Map<String,String> remoteProps = null;
+            try {
+                remoteProps = ConnectorAddressLink.importRemoteFrom(0);
+            } catch (IOException ioe) {
+                warning(AGENT_EXCEPTION, "JDP not starting, PerfData not available: " + ioe.getMessage());
```
dholmes-ora, line 506: Is this the only possible reason?

As per the JBS link I added there is a general problem about the Java side of the "Perf" code not knowing/realizing that `UsePerfData` can be disabled. Ideally there would be a way to query this, or a specific type of exception that could be thrown to report it.

Will the user understand what this message means? Should it explicitly say "please ensure -XX:-UsePerfData" is not set?

kevinjwalls: > Is this the only possible reason?

Yes I think it is - although IOException is already declared by importRemoteFrom, and that calls perf.attach(id) which declares IOException, the native Perf_Attach doesn't look like it throws it (OOM and IllegalArgument can be thrown).

BUT maybe a better alternative is to catch the BufferUnderflowException, this is the unchecked Exception which currently terminates us.  This is more specific and safer in case we ever do make more use of IOException in the attach - I wil update...

(Use of ConnectorAddressLink is all about communication within the same process so it doesn't expect any problems...)


> Will the user understand what this message means?

If I use the BufferUnderflowException, there's no detail message so it will say:

Warning: Exception thrown by the agent : JDP not starting, PerfData not available

If you have added -UsePerfData you should see the connection, it's not too chatty, and isn't saying too much in case it's wrong and the perfdata failed to allocate for some other reason.

dholmes-ora (approved):
Okay - seems reasonable.

Thanks.

plummercj (approved):
Looks good.

kevinjwalls:
Thanks for the reviews!

PR #13931, opened 2023-05-11
JDK-8307935: Class space argument processing can be simplified

tstuefe:
[JDK-8299229](https://bugs.openjdk.org/browse/JDK-8299229) removed the last scenario that tied compressed klass pointers to compressed oops.

In argument.cpp, we can therefore remove comments that refer to this old dependency.

Furthermore, in `Arguments::set_use_compressed_klass_ptrs()`, we can remove the following code pieces:

```
  if (FLAG_IS_DEFAULT(UseCompressedClassPointers)) {
    FLAG_SET_ERGO(UseCompressedClassPointers, true);
  }
```

I think this is a remnant from the time where compressed class pointers depended on compressed oops. We can now just set UseCompressedClassPointers by default to true for 64-bit platforms, which has the same effect as above code and is more in line of how we maintain default values normally.

```
  // Check the CompressedClassSpaceSize to make sure we use compressed klass ptrs.
  if (UseCompressedClassPointers) {
    if (CompressedClassSpaceSize > KlassEncodingMetaspaceMax) {
      warning("CompressedClassSpaceSize is too large for UseCompressedClassPointers");
      FLAG_SET_DEFAULT(UseCompressedClassPointers, false);
    }
  }
```

This coding is unnecessary since the CompressedClassSpaceSize option is range-checked and capped at 3GB. KlassEncodingMetaspaceMax is 32GB, so we will never hit the above condition. Moreover, even if that check were effective, it would not be sufficient, since the encoding range size is shared between CDS and class space and therefore the space available depends on the (yet unknown) CDS archive size, modulo internal alignments. To reduce complexity, we may just remove this part.

(It makes sense to leave an assert here to counter-check the range check which is supposedly done beforehand).

Tests:
- manually ran runtime/CommandLine, runtime/Metaspace, gc/Metaspace and all gtest variants on Linux x64 and x86.

---------
### Progress

### Issue
 * [JDK-8307935](https://bugs.openjdk.org/browse/JDK-8307935): Class space argument processing can be simplified

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13931/head:pull/13931` \
`$ git checkout pull/13931`

`$ git checkout pull/13931` \
`$ git pull https://git.openjdk.org/jdk.git pull/13931/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13931`

`$ git pr show -t 13931`

</details>

<a href="https://git.openjdk.org/jdk/pull/13931.diff">https://git.openjdk.org/jdk/pull/13931.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13931#issuecomment-1543973001)


tstuefe:
> Looks good. Maybe consider renaming `set_use_compressed_klass_ptrs` now that it doesn't _set_ anything.

Thanks @stefank !

I will fill this function with a new purpose setting stuff shortly, in a future patch. So I would prefer it kept this name, otherwise I would have to name it back later.

stefank:
> > Looks good. Maybe consider renaming `set_use_compressed_klass_ptrs` now that it doesn't _set_ anything.
> 
> Thanks @stefank !
> 
> I will fill this function with a new purpose setting stuff shortly, in a future patch. So I would prefer it kept this name, otherwise I would have to name it back later.

OK. Sounds good.

coleenp (approved):
This is ok.  I was going to object thinking that users/tests might have -XX:+UseCompressedOops and not remember to also set UseCompressedClassPointers, but if the latter is default that's fine.

tstuefe:
Thank you @coleenp 

/integrate

PR #13932, opened 2023-05-11
8306647: Implementation of Structured Concurrency (Preview)

AlanBateman:
This is the implementation of:

- JEP 453: Structured Concurrency (Preview)
- JEP 446: Scoped Values (Preview)

For the most part, this is just moving code and tests.  StructuredTaskScope moves to j.u.concurrent as a preview API, ScopedValue moves to j.lang as a preview API, and module jdk.incubator.concurrent has been removed. The significant API changes since incubator are:

- StructuredTaskScope.fork returns Subtask instead of Future (JEP 453 has a section on this)
- ScopedValue.where methods are replaced with runWhere, callWhere and getWhere

---------
### Progress
- [x] Change requires CSR request [JDK-8306573](https://bugs.openjdk.org/browse/JDK-8306573) to be approved
- [x] Change requires CSR request [JDK-8306916](https://bugs.openjdk.org/browse/JDK-8306916) to be approved

### Issues
 * [JDK-8306647](https://bugs.openjdk.org/browse/JDK-8306647): Implementation of Structured Concurrency (Preview) (**Enhancement** - `"3"`)
 * [JDK-8306572](https://bugs.openjdk.org/browse/JDK-8306572): Implementation of Scoped Values (Preview) (**Enhancement** - `"4"`)
 * [JDK-8306916](https://bugs.openjdk.org/browse/JDK-8306916): Implementation of Structured Concurrency (Preview) (**CSR**)
 * [JDK-8306573](https://bugs.openjdk.org/browse/JDK-8306573): Implementation of Scoped Values (Preview) (**CSR**)

### Reviewers
 * [Daniel Fuchs](https://openjdk.org/census#dfuchs) (@dfuch - **Reviewer**) ⚠️ Review applies to [cc902ce6](https://git.openjdk.org/jdk/pull/13932/files/cc902ce63d59ee484a5e2c2496bcc089f66d6631)
 * [Mandy Chung](https://openjdk.org/census#mchung) (@mlchung - **Reviewer**) ⚠️ Review applies to [cc902ce6](https://git.openjdk.org/jdk/pull/13932/files/cc902ce63d59ee484a5e2c2496bcc089f66d6631)

### Contributors
 * Alan Bateman `<alanb@openjdk.org>`
 * Andrew Haley `<aph@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13932/head:pull/13932` \
`$ git checkout pull/13932`

`$ git checkout pull/13932` \
`$ git pull https://git.openjdk.org/jdk.git pull/13932/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13932`

`$ git pr show -t 13932`

</details>

<a href="https://git.openjdk.org/jdk/pull/13932.diff">https://git.openjdk.org/jdk/pull/13932.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13932#issuecomment-1559673598)


src/java.base/share/classes/java/lang/ScopedValue.java
```diff
@@ -370,6 +372,48 @@ public <R> R call(Callable<? extends R> op) throws Exception {
...
+            var newSnapshot = new Snapshot(this, prevSnapshot);
+            return runWith(newSnapshot, new CallableAdapter<R>(op));
+        }
+
+        // A lightweight adapter from Supplier to Callable. This is
+        // used here to create the Callable which is passed to
+        // Carrier#call() in this thread because it needs neither
+        // runtime bytecode generation nor any release fencing.
+        private static final class CallableAdapter<V> implements Callable<V> {
+            private Supplier<? extends V> s;
```
PaulSandoz, line 408:
```suggestion
            private final Supplier<? extends V> s;
```

src/java.base/share/classes/java/lang/ScopedValue.java
```diff
@@ -488,12 +532,46 @@ public static <T> Carrier where(ScopedValue<T> key, T value) {
...
+     * StructuredTaskScope#close() close} it, then exiting {@code op} causes the
+     * underlying construct of each {@code StructuredTaskScope} created in the
+     * dynamic scope to be closed. This may require blocking until all child threads
+     * have completed their sub-tasks. The closing is done in the reverse order that
+     * they were created. Once closed, {@link StructureViolationException} is thrown.
+     *
+     * @implNote
+     * This method is implemented to be equivalent to:
+     * {@snippet lang=java :
+     *     // @link substring="call" target="Carrier#call(Callable)" :
```
PaulSandoz, line 558:
```suggestion
     *     // @link substring="get" target="Carrier#get(Supplier)" :
```
?

src/java.base/share/classes/java/lang/ScopedValue.java
```diff
@@ -248,7 +249,7 @@ Object find(ScopedValue<?> key) {
      * mapping bound to values. The following example runs an operation with {@code k1}
      * bound (or rebound) to {@code v1}, and {@code k2} bound (or rebound) to {@code v2}.
      * {@snippet lang=java :
-     *     // @link substring="where" target="#where(ScopedValue, Object)" :
+     *     // @link substring="runWhere" target="#runWhere(ScopedValue, Object)" :
```
PaulSandoz, line 252: Is this correct?

AlanBateman: Good catch, might have been the victim of search and replace, it should continue to link to ScopedValue.where.

src/java.base/share/classes/java/lang/ScopedValue.java
```diff
@@ -370,6 +371,48 @@ public <R> R call(Callable<? extends R> op) throws Exception {
...
+         * @param <R> the type of the result of the operation
+         * @return the result
+         * @see ScopedValue#getWhere(ScopedValue, Object, Supplier)
+         */
+        public <R> R get(Supplier<? extends R> op) {
+            Objects.requireNonNull(op);
+            Cache.invalidate(bitmask);
+            var prevSnapshot = scopedValueBindings();
+            var newSnapshot = new Snapshot(this, prevSnapshot);
+            return runWith(newSnapshot, new CallableAdapter<R>(op));
```
PaulSandoz, line 399: Can we just do this instead?
```suggestion
            return runWith(newSnapshot, op::get);
```
IIUC the current approach is to avoid the dynamic creation of a class via the invoke dynamic? I don't fully understand the comment about release fencing.

AlanBateman: @theRealAph will want to comment on this as it very performance sensitive. I think CallableAdapter.s is non-final to avoid the release fence.

src/java.base/share/classes/java/util/concurrent/StructuredTaskScope.java
```diff
@@ -0,0 +1,1281 @@
...
+ * }
+ * The second example creates a ShutdownOnFailure object to capture the exception of the
+ * first subtask to fail, cancelling the other by way of shutting down the task scope. The
+ * main task waits in {@link #joinUntil(Instant)} until both subtasks complete with a
+ * result, either fails, or a deadline is reached. It invokes {@link
+ * ShutdownOnFailure#throwIfFailed(Function) throwIfFailed(Function)} to throw an exception
+ * if either subtask fails. This method is a no-op if both subtasks complete successfully.
+ * The example uses {@link Supplier#get()} to get the result of each subtask. Using
+ * {@code Supplier} instead of {@code Subtask} is preferred for common cases where the
+ * the object returned by fork is only used to get the result of a subtask that completed
```
PaulSandoz, line 159:
```suggestion
 * {@code Supplier} instead of {@code Subtask} is preferred for common cases where
 * the object returned by fork is only used to get the result of a subtask that completed
```

src/java.base/share/classes/java/util/concurrent/StructuredTaskScope.java
```diff
@@ -0,0 +1,1281 @@
...
+                T r = (T) result;
+                return r;
+            }
+
+            Throwable ex = firstException;
+            if (ex != null) {
+                throw new ExecutionException(ex);
+            }
+
+            throw new IllegalStateException("No completed subtasks");
```
PaulSandoz, line 1077: I believe it may be possible to implement as the following if you so wish:
```suggestion
            return result(ExecutionException::new);
```

src/java.base/share/classes/java/util/concurrent/StructuredTaskScope.java
```diff
@@ -0,0 +1,1281 @@
...
+         * @throws ExecutionException if a subtask failed
+         * @throws WrongThreadException if the current thread is not the task scope owner
+         * @throws IllegalStateException if the task scope owner did not invoke join after
+         * forking
+         */
+        public void throwIfFailed() throws ExecutionException {
+            ensureOwnerAndJoined();
+            Throwable exception = firstException;
+            if (exception != null)
+                throw new ExecutionException(exception);
```
PaulSandoz, line 1251:
```suggestion
            throwIfFailed(ExecutionException::new);
```

src/java.base/share/classes/java/util/concurrent/StructuredTaskScope.java
AlanBateman: Good, avoids duplication.

src/java.base/share/classes/java/lang/ScopedValue.java
theRealAph: That's right. The problem is that we can never get rid of the release fence, apparently even when the instance of the adapter is scalar replaced. I imagine that'll get fixed one day, but this is internal JDK code.

PaulSandoz: Oh, and i guess that has some performance implications in some cases? more so on the set of instructions produced on ARM say rather than limiting C2 optimizations?

Given that the Supplier is likely to be the target of a lambda expression which may also capture I was surprised there would be much of an increased impact. (HotSpot may not strength reduce multiple fences in this case.)

Can we update to add say "/* non-final */ to the field and update the class doc to say the release fence inserted by HotSpot as a result of constructing a class with final fields has performance implications <insert loose quantification of those implications> ?

test/jdk/java/util/concurrent/StructuredTaskScope/StructuredTaskScopeTest.java
```diff
@@ -0,0 +1,1621 @@
...
+    @MethodSource("factories")
+    void testHandleCompleteBeforeShutdown(ThreadFactory factory) throws Exception {
+        try (var scope = new CollectAll<String>(factory)) {
+            Callable<String> task1 = () -> "foo";
+            Callable<String> task2 = () -> { throw new FooException(); };
+            scope.fork(task1);
+            scope.fork(task2);
+            scope.join();
+
+            var subtask1  = scope.find(task1);
```
turbanoff, line 1016:
```suggestion
            var subtask1 = scope.find(task1);
```

src/java.base/share/classes/java/util/concurrent/StructuredTaskScope.java
```diff
@@ -0,0 +1,1275 @@
...
+     * Starts a new thread in this task scope to execute a value-returning task, thus
+     * creating a <em>subtask</em> of this task scope.
+     *
+     * <p> The value-returning task is provided to this method as a {@link Callable}, the
+     * thread executes the task's {@link Callable#call() call} method. The thread is
+     * created with the task scope's {@link ThreadFactory}. It inherits the current thread's
+     * {@linkplain ScopedValue scoped value} bindings. The bindings must match the bindings
+     * captured when the task scope was created.
+     *
+     * <p> This method returns a {@link Subtask Subtask}to represent the <em>forked
```
PaulSandoz, line 521:
```suggestion
     * <p> This method returns a {@link Subtask Subtask} to represent the <em>forked
```

src/java.base/share/classes/java/util/concurrent/StructuredTaskScope.java
```diff
@@ -0,0 +1,1275 @@
...
+     */
+    private void ensureJoinedIfOwner(int phase) {
+        if (Thread.currentThread() == flock.owner() && (phase > lastJoinCompleted)) {
+            throw newIllegalStateExceptionNoJoin();
+        }
+    }
+
+    /**
+     * Ensures that the current thread is the owner of this task scope and that it joined
+     * (with {@link #join()} or {@link #joinUntil(Instant)} after {@linkplain #fork(Callable)
```
PaulSandoz, line 473:
```suggestion
     * (with {@link #join()} or {@link #joinUntil(Instant)}) after {@linkplain #fork(Callable)
```

PaulSandoz (commented):
This looks good (both `ScopedValue` and `StructuredTaskScope`).

I recall some discussion on the loom email list noting that `ScopedValue` instances are like capabilities and should be shared carefully (like `MethodHandle`s in some sense). I think it would be useful to follow up on that in the specification.

I am uncertain on the name `Subtask` but cannot think of anything better, perhaps during this round of preview we might find something else.

For a small API `StructureTaskScope` is rather powerful and deftly combines language, library, and platform features (virtual threads). 

I suggest revisiting the constraints on operating on `SubTask` instances. I get the motivation but i wonder if the additional specification and implementation work required to properly describe the behaviour and enforce it is really worth it. Such a restriction tends to feel artificial given there are no side-effects to such operations, and further we cannot enforce such restrictions in the compiler. 
The constraints of `fork`, `join`, `shutdown`, and `close` are key with the idiomatic use with try-with-resources. Just let subtasks be accessible by any thread for all its methods and provide clear documentation on the recommended use.

forax:
I've rewritten all my tests to use the new API, it works !
https://github.com/forax/loom-fiber/tree/java21

One surprising thing is that Subtask.get() give less leeway to the owner thread than to the other virtual threads so in onComplete() storing a Subtask to use it later by the owner thread does not work well if join() is not called yet.
This is perhaps a bit too much.

Anyway, good for me !

AlanBateman:
> One surprising thing is that Subtask.get() give less leeway to the owner thread than to the other virtual threads so in onComplete() storing a Subtask to use it later by the owner thread does not work well if join() is not called yet. This is perhaps a bit too much.

Maybe but it would be a bug if the owner were to call Subtask::get before joining. This is easy to detect for subtasks that are forked by the owner. Time will tell how common it will be for subtasks to fork additional subtasks.

dfuch (approved):
I reviewed the API documentation of `ScopedValue` (mostly the class level API documentation, but I had a look at the methods API too) and it looks good. It reads well, and when I had question they were usually answered in the next paragraph that followed what I was reading.

mlchung (approved):
I reviewed the implementation changes to promote an incubating API to a preview API.  That part looks good.

PaulSandoz (approved):

PR #13945, opened 2023-05-11
8303465: KeyStore of type KeychainStore, provider Apple does not show all trusted certificates

RealCLanger:
With this PR we try to be better in loading certificates from the MacOS Keychain into a JDK Trust store.

The current implementation after JDK-8278449 would only load/trust certificates from an identity (with private key available) and certificates that have explicit trust set in the user domain (as shown by security dump-trust-settings). This, however is not sufficient and does not match the MacOS system behavior, e.g. if you compare with tools like curl or Safari.

This change does the following:
1. The native method that reads trust settings will call the API SecTrustSettingsCopyTrustSettings on a certificate for both, User and Admin domain.
2. We now trust self-signed certificates that have an explicit trust entry with no sub-records or no sub-records that would deny the certificate usage for any purpose.
3. The check for double aliases has been augmented by comparing whether the certificate to be added is the same as the one that is already present. This can happen if a certificate is contained in both, the user and the system keychain, for instance.

I have added a test that verifies whether certificates that should be trusted from "security dump-trust-settings" are contained in the keystore and those that should be disallowed are absent.

---------
### Progress
- [x] Change requires CSR request [JDK-8308690](https://bugs.openjdk.org/browse/JDK-8308690) to be approved

### Issues
 * [JDK-8303465](https://bugs.openjdk.org/browse/JDK-8303465): KeyStore of type KeychainStore, provider Apple does not show all trusted certificates
 * [JDK-8308690](https://bugs.openjdk.org/browse/JDK-8308690): KeyStore of type KeychainStore, provider Apple does not show all trusted certificates (**CSR**)

### Reviewers
 * [Matthias Baesken](https://openjdk.org/census#mbaesken) (@MBaesken - **Reviewer**) ⚠️ Review applies to [b14e5f2c](https://git.openjdk.org/jdk/pull/13945/files/b14e5f2c78ff4aded84410a2b58d83138349d9ab)
 * [Weijun Wang](https://openjdk.org/census#weijun) (@wangweij - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13945/head:pull/13945` \
`$ git checkout pull/13945`

`$ git checkout pull/13945` \
`$ git pull https://git.openjdk.org/jdk.git pull/13945/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13945`

`$ git pr show -t 13945`

</details>

<a href="https://git.openjdk.org/jdk/pull/13945.diff">https://git.openjdk.org/jdk/pull/13945.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13945#issuecomment-1544732815)


MBaesken:
Hi Christoph, I do not see any reference to kSecTrustSettingsDomainSystem in your coding.  Handling at least kSecTrustSettingsDomainUser  and  kSecTrustSettingsDomainAdmin is good but I am not sure about kSecTrustSettingsDomainSystem . Did you find some documentation why it should be omitted ?

MBaesken:
Additionally, the return value of SecTrustSettingsCopyTrustSettings (https://developer.apple.com/documentation/security/1400261-sectrustsettingscopytrustsetting) should  better be checked.

RealCLanger:
> Hi Christoph, I do not see any reference to kSecTrustSettingsDomainSystem in your coding. Handling at least kSecTrustSettingsDomainUser and kSecTrustSettingsDomainAdmin is good but I am not sure about kSecTrustSettingsDomainSystem . Did you find some documentation why it should be omitted ?

Hi Matthias,
yes, I think it is not nicely documented. I've seen in testing, that kSecTrustSettingsDomainSystem merely holds information for trusted root CAs. So in theory, we could add this. However, other code in that area that we've found out in the wild doesn't do it as well. Let's see what others think about this.

MBaesken:
> > Hi Christoph, I do not see any reference to kSecTrustSettingsDomainSystem in your coding. Handling at least kSecTrustSettingsDomainUser and kSecTrustSettingsDomainAdmin is good but I am not sure about kSecTrustSettingsDomainSystem . Did you find some documentation why it should be omitted ?
> 
> Hi Matthias, yes, I think it is not nicely documented. I've seen in testing, that kSecTrustSettingsDomainSystem merely holds information for trusted root CAs. So in theory, we could add this. However, other code in that area that we've found out in the wild doesn't do it as well. Let's see what others think about this.

Yes this seems to be the case.  Could you maybe add a one liner comment to  libosxsecurity/KeystoreImpl.m (near to the user and admin domain handling)  summarizing what you said? And I still prefer checking the return values of the calls to SecTrustSettingsCopyTrustSettings .

RealCLanger:
> Yes this seems to be the case. Could you maybe add a one liner comment to libosxsecurity/KeystoreImpl.m (near to the user and admin domain handling) summarizing what you said? And I still prefer checking the return values of the calls to SecTrustSettingsCopyTrustSettings .

Done. 😄

src/java.base/macosx/native/libosxsecurity/KeystoreImpl.m
MBaesken, line 1: Regarding CFArrayGetValueAtIndex, when looking at other usages of the function in the codebase the result is NULL checked usually. Should we do this here too? I admit the old coding does not have it as well.

MBaesken (approved):
So please check the CFArrayGetValueAtIndex usage, but otherwise looks okay to me now, thanks for the adjustments.

seanjmullan:
Please don't integrate this until I or someone from my team reviews it. Thanks.

RealCLanger:
> Please don't integrate this until I or someone from my team reviews it. Thanks.

Sure.

RealCLanger:
> No matter what `SecTrustSettingsCopyTrustSettings` returns, you will always call `jm_createTrustedCertEntry`. This means if I add a self-signed certificate but has not added any trusted settings onto it, it will be always trusted. Is this intended?

Yes, I will call `jm_createTrustedCertEntry` for every certificate, at least independent from the results of the '`SecTrustSettingsCopyTrustSettings` calls.
As I outlined in my initial PR description, point 3, the actual check whether a certificate is self-signed is done in the `createTrustedCertEntry` Java method. So, yes, when there is a self-signed certificate without explicit trust settings, it is always trusted. I thought that this was the intentional behavior even before my changes. However, the difference to the code before is that I look at the certificate and check whether it is a real plain self-signed certificate that can be used for TLS communication - which would be trusted. But what's not trusted now are CA root certificates which also means self-signed but additionally have key usage 'keyCertSign' or 'cRLSign'. See here: https://github.com/RealCLanger/jdk/blob/b14e5f2c78ff4aded84410a2b58d83138349d9ab/src/java.base/macosx/classes/apple/security/KeychainStore.java#L857

Makes sense?

src/java.base/macosx/native/libosxsecurity/KeystoreImpl.m
RealCLanger: Hm, can it really be NULL? I mean before we check the array size and then iterate the valid range.

RealCLanger:
> Before your new change, such a certificate is not trusted, because `SecTrustSettingsCopyTrustSettings` returns `errSecItemNotFound` so `jm_createTrustedCertEntry` is not called at all.
> 
> I am not sure if such a certificate is meant to be always trusted. Note that you can create such an entry with only `security add-certificates` but not `security add-trusted-cert`. macOS allows anyone to run the first command but prompts you for an administrator password when running the second. The name of the second command also implies that it's the only way to assign trust to a certificate, IMHO.

Hm, after thinking about this again and also comparing with behavior of curl, I think you're right. A self-signed certificate should only be trusted if it has a trust entry (e.g. added by `security add-trusted-cert`). Somehow I was under the impression that self-signed certificates should be trusted when they exist. But after reading comments etc. again I'm not sure why I thought so at all. 😜 Will update the PR...

src/java.base/macosx/classes/apple/security/KeychainStore.java
```diff
@@ -804,8 +803,26 @@ private void createTrustedCertEntry(String alias, List<String> inputTrust,
             tce.cert = cert;
             tce.certRef = keychainItemRef;
 
+            // Check whether a certificate with same alias already exists and is the same
+            // If yes, we can return here - the existing entry must have the same
+            // properties and trust settings
```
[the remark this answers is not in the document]
RealCLanger: This handles the case, when a certificate is in both, the login (user) and system keychain.

RealCLanger:
> Since you removed the key usage checks, can you update the PR description please?

Done.

RealCLanger:
> The code change looks fine to me. Thanks.
> 
> Since this is rather a big behavior change, I think a CSR and a release note are required. The previous release note on this is at https://www.oracle.com/java/technologies/javase/19-relnote-issues.html#JDK-8278449.

Thanks for the review. As for CSR and release note, I guess one could do both, sure. However, it's quite a bit of overhead. And, for JDK-8278449, I don't see a CSR either. 😉 I would consider this hange rather as a bugfix for JDK-8278449. So, if you really want me to do it, I'll do it. But please help me to get both processed quickly as I want to get this integrated ideally through this week. Please let me know. Thanks.

RealCLanger:
> I've started the CSR at https://bugs.openjdk.org/browse/JDK-8308690. Please edit if there is any issue. At the same time, please write a release note. See https://openjdk.org/guide/#release-notes for the process.

Thanks. I've created a release note and Edited the CSR. Please review both. When you are fine with the CSR, I can finalize it.

RealCLanger:
> I think you can finalize the CSR now.

Thx for the hint, done.

RealCLanger:
@wangweij The CSR is approved. Would you like to review/approve the PR now?

RealCLanger:
Thx.
/integrate

PR #13949, opened 2023-05-12
8307365: JvmtiStressModule hit SIGSEGV in JvmtiEventControllerPrivate::recompute_thread_enabled

pchilano:
The following patch fixes a bug introduced while refactoring the VirtualThreadStart/End events. Specifically, the code to delete the JvmtiThreadState of a terminating vthread was moved before we start the VTMS transition. That allowed said code to run concurrently with recompute_enabled() leading to different crashing modes. I wrote the detailed sequence of events leading to the crash in the bug comments. 

To fix it I moved the cleanup code back after the call to VTMS_unmount_begin(). Now, since the rebinding of the JvmtiThreadState to that of the carrier has to be done after this cleanup code is executed, otherwise we would delete the wrong JvmtiThreadState state, I had to add a boolean argument to VTMS_unmount_begin() to differentiate the last unmount call from the other ones. This is unfortunate since ideally VTMS_unmount_begin() would be oblivious to these two cases as with VTMS_mount_end() where we don't need to check if this is the first mount.
I looked for other ways to solve it instead of the extra boolean argument but wasn't convinced. One way would be to have another JvmtiExport::cleanup_thread() that would handle this case. Another way which is very simple is to move the rebind_to_jvmti_thread_state_of() call to VTMS_unmount_end() instead. But that means during the transition the _jvmti_thread_state field of the carrier would be either null or that of the vthread, unlike today which is always that of the carrier during the transitions. I didn't want to change that behavior in this fix but I can also explore that route.

I tested the patch with the reproducer I attached to the bug, plus I also run tiers1-3 in mach5.

Thanks,
Patricio

---------
### Progress

### Issue
 * [JDK-8307365](https://bugs.openjdk.org/browse/JDK-8307365): JvmtiStressModule hit SIGSEGV in JvmtiEventControllerPrivate::recompute_thread_enabled

### Reviewers
 * [Serguei Spitsyn](https://openjdk.org/census#sspitsyn) (@sspitsyn - **Reviewer**) ⚠️ Review applies to [f22cc954](https://git.openjdk.org/jdk/pull/13949/files/f22cc954511376a567d22878868aa54fa6aaa6a8)
 * [Daniel D. Daugherty](https://openjdk.org/census#dcubed) (@dcubed-ojdk - **Reviewer**) ⚠️ Review applies to [f22cc954](https://git.openjdk.org/jdk/pull/13949/files/f22cc954511376a567d22878868aa54fa6aaa6a8)
 * [Leonid Mesnik](https://openjdk.org/census#lmesnik) (@lmesnik - **Reviewer**) ⚠️ Review applies to [78b1eaf3](https://git.openjdk.org/jdk/pull/13949/files/78b1eaf3cc363dbf58fdc857ce09950a9d23cc6b)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13949/head:pull/13949` \
`$ git checkout pull/13949`

`$ git checkout pull/13949` \
`$ git pull https://git.openjdk.org/jdk.git pull/13949/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13949`

`$ git pr show -t 13949`

</details>

<a href="https://git.openjdk.org/jdk/pull/13949.diff">https://git.openjdk.org/jdk/pull/13949.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13949#issuecomment-1545041739)

### DIFF
--- a/src/hotspot/share/prims/jvmtiThreadState.cpp
+++ b/src/hotspot/share/prims/jvmtiThreadState.cpp
@@ -554,13 +554,14 @@ JvmtiVTMSTransitionDisabler::VTMS_vthread_end(jobject vthread) {
       JvmtiExport::post_vthread_end(vthread);
     }
   }
+  VTMS_unmount_begin(vthread, /* last_unmount */ true);
   if (thread->jvmti_thread_state() != nullptr) {
+    assert(thread->jvmti_thread_state()->is_virtual(), "wrong JvmtiThreadState");
     JvmtiExport::cleanup_thread(thread);
-    thread->set_jvmti_thread_state(nullptr);
-    oop vt = JNIHandles::resolve(vthread);
-    java_lang_Thread::set_jvmti_thread_state(vt, nullptr);
+    assert(thread->jvmti_thread_state() == nullptr, "should be null");
+    assert(java_lang_Thread::jvmti_thread_state(JNIHandles::resolve(vthread)) == nullptr, "should be null");
   }
-  VTMS_unmount_begin(vthread);
+  thread->rebind_to_jvmti_thread_state_of(thread->threadObj());
 }
 
 void
@@ -581,7 +582,7 @@ JvmtiVTMSTransitionDisabler::VTMS_vthread_unmount(jobject vthread, bool hide) {
     if (JvmtiExport::should_post_vthread_unmount()) {
       JvmtiExport::post_vthread_unmount(vthread);
     }
-    VTMS_unmount_begin(vthread);
+    VTMS_unmount_begin(vthread, /* last_unmount */ false);
   } else {
     VTMS_unmount_end(vthread);
   }
@@ -615,14 +616,16 @@ JvmtiVTMSTransitionDisabler::VTMS_mount_end(jobject vthread) {
 }
 
 void
-JvmtiVTMSTransitionDisabler::VTMS_unmount_begin(jobject vthread) {
+JvmtiVTMSTransitionDisabler::VTMS_unmount_begin(jobject vthread, bool last_unmount) {
   JavaThread* thread = JavaThread::current();
 
   assert(!thread->is_in_tmp_VTMS_transition(), "sanity check");
   assert(!thread->is_in_VTMS_transition(), "sanity check");
 
   start_VTMS_transition(vthread, /* is_mount */ false);
-  thread->rebind_to_jvmti_thread_state_of(thread->threadObj());
+  if (!last_unmount) {
+    thread->rebind_to_jvmti_thread_state_of(thread->threadObj());
+  }
 }
 
 void
--- a/src/hotspot/share/prims/jvmtiThreadState.cpp
+++ b/src/hotspot/share/prims/jvmtiThreadState.cpp
@@ -556,7 +556,6 @@ JvmtiVTMSTransitionDisabler::VTMS_vthread_end(jobject vthread) {
   }
   VTMS_unmount_begin(vthread, /* last_unmount */ true);
   if (thread->jvmti_thread_state() != nullptr) {
-    assert(thread->jvmti_thread_state()->is_virtual(), "wrong JvmtiThreadState");
     JvmtiExport::cleanup_thread(thread);
     assert(thread->jvmti_thread_state() == nullptr, "should be null");
     assert(java_lang_Thread::jvmti_thread_state(JNIHandles::resolve(vthread)) == nullptr, "should be null");
--- a/test/hotspot/jtreg/serviceability/jvmti/vthread/ThreadStateTest/ThreadStateTest.java
+++ b/test/hotspot/jtreg/serviceability/jvmti/vthread/ThreadStateTest/ThreadStateTest.java
@@ -24,10 +24,10 @@
 /**
  * @test
  * @bug 8307365
- * @summary Exercise JVMTI state creation concurrently with terminating vthreads
+ * @summary Exercise JvmtiThreadState creation concurrently with terminating vthreads
  * @requires vm.continuations
  * @modules java.base/java.lang:+open
- * @run main/othervm/native -agentlib:JVMTIStateSanityTest JVMTIStateSanityTest
+ * @run main/othervm/native -agentlib:ThreadStateTest ThreadStateTest
  */
 
 import java.util.concurrent.*;
@@ -37,8 +37,8 @@ import java.util.List;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 
-public class JVMTIStateSanityTest {
-    private static final String agentLib = "JVMTIStateSanityTest";
+public class ThreadStateTest {
+    private static final String agentLib = "ThreadStateTest";
     static final int VTHREAD_COUNT = 64;
 
     private static native void SetSingleSteppingMode(boolean enable);
@@ -51,7 +51,7 @@ public class JVMTIStateSanityTest {
     private void runTest() throws Exception {
         int tryCount = 150;
 
-        // Force creation of JVMTI state on vthread start.
+        // Force creation of JvmtiThreadState on vthread start.
         SetMonitorContendedMode(true);
 
         while (tryCount-- > 0) {
@@ -70,7 +70,7 @@ public class JVMTIStateSanityTest {
             // Give some time for vthreads to finish.
             Thread.sleep(10);
 
-            // Trigger race of JVMTI state creation with terminating vthreads.
+            // Trigger race of JvmtiThreadState creation with terminating vthreads.
             SetMonitorContendedMode(false);
             SetMonitorContendedMode(true);
 
@@ -81,7 +81,7 @@ public class JVMTIStateSanityTest {
             scheduler.shutdown();
             Thread.sleep(20);
 
-            // Check that looping over all JVMTI states works fine.
+            // Check that looping over all JvmtiThreadStates works fine.
             SetSingleSteppingMode(true);
 
             // Reset for next iteration
@@ -97,7 +97,7 @@ public class JVMTIStateSanityTest {
             System.err.println("java.library.path: " + System.getProperty("java.library.path"));
             throw ex;
         }
-        JVMTIStateSanityTest obj = new JVMTIStateSanityTest();
+        ThreadStateTest obj = new ThreadStateTest();
         obj.runTest();
     }
 
--- a/test/hotspot/jtreg/serviceability/jvmti/vthread/ThreadStateTest/ThreadStateTest.java
+++ b/test/hotspot/jtreg/serviceability/jvmti/vthread/ThreadStateTest/ThreadStateTest.java
@@ -38,11 +38,10 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 
 public class ThreadStateTest {
-    private static final String agentLib = "ThreadStateTest";
     static final int VTHREAD_COUNT = 64;
 
-    private static native void SetSingleSteppingMode(boolean enable);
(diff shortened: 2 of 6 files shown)


src/hotspot/share/prims/jvmtiThreadState.cpp
```diff
@@ -554,13 +554,14 @@ JvmtiVTMSTransitionDisabler::VTMS_vthread_end(jobject vthread) {
       JvmtiExport::post_vthread_end(vthread);
     }
   }
+  VTMS_unmount_begin(vthread, /* last_unmount */ true);
   if (thread->jvmti_thread_state() != nullptr) {
+    assert(thread->jvmti_thread_state()->is_virtual(), "wrong JvmtiThreadState");
```
sspitsyn, line 559: We agreed with you to temporarily remove this assert as it triggers the bug:
[8308124](https://bugs.openjdk.org/browse/JDK-8308124) dynamic loading of a JVMTI agent has a race with JvmtiThreadState cleanup

A fix of the [8308124](https://bugs.openjdk.org/browse/JDK-8308124) will add this assert back.

sspitsyn (approved):
Thank you for taking care about this issue!
Yes, clearing the `JvmtiThreadState` of a virtual thread has to be done
while in transition as it provides a needed synchronization.
This makes it a little bit ugly but I hope it can be simplified again after getting rid of the `rebind_to_jvmti_thread_state_of()` which is still on my TODO list.
Thanks,
Serguei

pchilano:
> Thank you for taking care about this issue! Yes, clearing the `JvmtiThreadState` of a virtual thread has to be done while in transition as it provides a needed synchronization. This makes it a little bit ugly but I hope it can be simplified again after getting rid of the `rebind_to_jvmti_thread_state_of()` which is still on my TODO list. Thanks, Serguei
>
Thanks for the review Serguei!

Patricio

src/hotspot/share/prims/jvmtiThreadState.cpp
pchilano: Yes, thanks for finding and filing a bug for that. I removed the assert.

sspitsyn (approved):

dcubed-ojdk:
@pchilano - This bug has had reported sightings in Tier[34568] so Tier[1-3]
testing is probably not enough. However, since you've been able to create a
tight reproducer, your testing might be sufficient. Of course, as @dholmes-ora
likes to say: only time will tell!

dcubed-ojdk (approved):
Summary: Thumbs up. This is most definitely NOT a trivial fix.

Normally the way I like to review these kinds of fixes is to map the failure
modes back to the fix just to make sure I understand how each of the
failure modes is covered by the changes. For this particular bug that has
NOT been an easy thing to do. I believe the failure modes are complicated
to follow because we "lost" the synchronization of being in the
`VTMS_unmount_begin` transition which allowed a thread calling
`recompute_enabled()` to race with our ending/exiting vthread/cthread
combo. Ouch.

The fix does restore the synchronization of being in the
`VTMS_unmount_begin` transition so I can see how we would no longer
be racing with the `recompute_enabled()` calling thread.

pchilano:
> Summary: Thumbs up. This is most definitely NOT a trivial fix.
> 
> Normally the way I like to review these kinds of fixes is to map the failure modes back to the fix just to make sure I understand how each of the failure modes is covered by the changes. For this particular bug that has NOT been an easy thing to do. I believe the failure modes are complicated to follow because we "lost" the synchronization of being in the `VTMS_unmount_begin` transition which allowed a thread calling `recompute_enabled()` to race with our ending/exiting vthread/cthread combo. Ouch.
> 
Exactly. In the reproducer I aimed for the ThreadsListHandle crash, but while working on it I also hit the other ones.

pchilano:
> @pchilano - This bug has had reported sightings in Tier[34568] so Tier[1-3] testing is probably not enough. However, since you've been able to create a tight reproducer, your testing might be sufficient. Of course, as @dholmes-ora likes to say: only time will tell!
>
I actually also tested it in mach5 tiers 4-6 after the initial post.  : )

src/hotspot/share/prims/jvmtiThreadState.cpp
```diff
@@ -581,7 +581,7 @@ JvmtiVTMSTransitionDisabler::VTMS_vthread_unmount(jobject vthread, bool hide) {
     if (JvmtiExport::should_post_vthread_unmount()) {
       JvmtiExport::post_vthread_unmount(vthread);
     }
-    VTMS_unmount_begin(vthread);
+    VTMS_unmount_begin(vthread, /* last_unmount */ false);
```
lmesnik, line 584: I think it would be better just to add 
thread->rebind_to_jvmti_thread_state_of(thread->threadObj());
after this call instead of adding a parameter to  VTMS_unmount_begin.
Just suggestion.

pchilano:
Thanks for the review Dan!

lmesnik (approved):
The changes looks good. Just small suggestion.
Any reasons why didn't add the test?

src/hotspot/share/prims/jvmtiThreadState.cpp
pchilano: So we need to execute the thread->rebind_to_jvmti_thread_state_of(thread->threadObj()) call after we do the cleanup to avoid removing the wrong state. So the other option to avoid the extra parameter would be to remove the rebind call from VTMS_unmount_begin(), which I also thought. But that would require to also add it to VTMS_vthread_unmount() for the normal unmount case, which already breaks the symmetry of all those methods, given that for the mount case the rebind call is encapsulated in VTMS_mount_end().

pchilano:
> The changes looks good. Just small suggestion. Any reasons why didn't add the test?
>
Thanks for the review Leonid!
I guess I think about this test as too contrived to expose this particular bug and might not be that useful otherwise. But I'm not sure what the policy is about adding reproducer tests.

src/hotspot/share/prims/jvmtiThreadState.cpp
lmesnik: Thanks for the explanation.

lmesnik:
> > The changes looks good. Just small suggestion. Any reasons why didn't add the test?
> 
> Thanks for the review Leonid! I guess I think about this test as too contrived to expose this particular bug and might not be that useful otherwise. But I'm not sure what the policy is about adding reproducer tests.

Our policy is to include all regression test/cases/examples which could be implemented as jtreg tests.  A lot of regression tests cover some specific bugs and corner cases. It is pretty fine. You could exclude the test from tier1 if feel that we don't need. to run it every time. But generally, any test coverage improvements are very welcome.

pchilano:
> > > The changes looks good. Just small suggestion. Any reasons why didn't add the test?
> > 
> > 
> > Thanks for the review Leonid! I guess I think about this test as too contrived to expose this particular bug and might not be that useful otherwise. But I'm not sure what the policy is about adding reproducer tests.
> 
> Our policy is to include all regression test/cases/examples which could be implemented as jtreg tests. A lot of regression tests cover some specific bugs and corner cases. It is pretty fine. You could exclude the test from tier1 if feel that we don't need. to run it every time. But generally, any test coverage improvements are very welcome.
>
The only thing about the repro is that as I wrote in the bug comments it needs a specific artificial delay inside the vm to make it crash. Without it I couldn't reproduce it, at least running the test locally. Maybe running it several times in mach5 will show some crashes. That's why I also doubt if this test will actually be useful. But do you think it's still worth it to add it?

lmesnik:
I think it is a good stress test for finishing virtual threads while some events are coming. 
Please just remove it's name from  Repro8307365 to something more descriptive and add it to the vthread subdirectory.

I am pretty sure that it is a useful test.

lmesnik:
BTW, you could file separate RFE for this just to don't delay current fix.

pchilano:
@lmesnik I added the new test, please check it out. You might also want to suggest a better name for it.

sspitsyn:
I'd suggest to name new test as `ThreadStateTest`, `JvmtiThreadStateTest` or `ThreadStateSanityTest`.
One more quick suggestion is to replace JVMTI state in the comments to `JvmtiThreadState`.

pchilano:
> I'd suggest to name new test as `ThreadStateTest`, `JvmtiThreadStateTest` or `ThreadStateSanityTest`. One more quick suggestion is to replace JVMTI state in the comments to `JvmtiThreadState`.
>
Good suggestions, done.

test/hotspot/jtreg/serviceability/jvmti/vthread/ThreadStateTest/ThreadStateTest.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+
+public class ThreadStateTest {
+    private static final String agentLib = "ThreadStateTest";
+    static final int VTHREAD_COUNT = 64;
+
+    private static native void SetSingleSteppingMode(boolean enable);
```
lmesnik, line 44: Could you please rename them to 's'etSingleSteppingMode and 's'etMonitorContendedMode.

test/hotspot/jtreg/serviceability/jvmti/vthread/ThreadStateTest/ThreadStateTest.java
```diff
@@ -0,0 +1,121 @@
...
+            // Check that looping over all JvmtiThreadStates works fine.
+            SetSingleSteppingMode(true);
+
+            // Reset for next iteration
+            SetSingleSteppingMode(false);
+        }
+    }
+
+    public static void main(String[] args) throws Exception {
+        try {
```
lmesnik, line 93: You could just remove try/catch.

test/hotspot/jtreg/serviceability/jvmti/vthread/ThreadStateTest/ThreadStateTest.java
```diff
@@ -0,0 +1,121 @@
...
+    };
+
+    private void runTest() throws Exception {
+        int tryCount = 150;
+
+        // Force creation of JvmtiThreadState on vthread start.
+        SetMonitorContendedMode(true);
+
+        while (tryCount-- > 0) {
+            ExecutorService scheduler = Executors.newFixedThreadPool(8);
```
lmesnik, line 58: Is it possible to configure pool using any of these parameters?
-Djdk.defaultScheduler.parallelism= 
 -Djdk.defaultScheduler.maxPoolSize=

lmesnik (approved):

test/hotspot/jtreg/serviceability/jvmti/vthread/ThreadStateTest/ThreadStateTest.java
pchilano: Done.

test/hotspot/jtreg/serviceability/jvmti/vthread/ThreadStateTest/ThreadStateTest.java
pchilano: The reason I'm not using the default scheduler is not to control the size of the pool but to be able to control its shutdown, since the carriers going away was part of the ThreadsListHandle crash.

lmesnik:
@pchilano Thank you for explanations and adding the test.

pchilano:
Thanks for the reviews @sspitsyn, @dcubed-ojdk and @lmesnik!

sspitsyn (commented):
New test looks good.
Thank you for adding it!
Serguei

PR #13950, opened 2023-05-12
8296920: Regression Test DialogOrient.java fails on MacOS

prsadhuk:
Test was failing with PrinterException: Wrong Orientation since printing was done with default orientation.
Fix is made to set the printable with correct PageFormat

It was not seen prior to [JDK-8262731](https://bugs.openjdk.org/browse/JDK-8262731) probably because the exception was being swallowed..

### DIFF
--- a/test/jdk/java/awt/print/Dialog/DialogOrient.java
+++ b/test/jdk/java/awt/print/Dialog/DialogOrient.java
@@ -69,8 +69,8 @@ public class DialogOrient implements Printable {
       Sysout.printInstructions( instructions );
 
         PrinterJob job = PrinterJob.getPrinterJob();
-        job.setPrintable(new DialogOrient());
         PageFormat landscape = job.pageDialog(job.defaultPage());
+        job.setPrintable(new DialogOrient(), landscape);
 
         if (job.printDialog()) {
             job.print();


test/jdk/java/awt/print/Dialog/DialogOrient.java
```diff
@@ -69,8 +69,8 @@ private static void init() throws Exception {
       Sysout.printInstructions( instructions );
 
         PrinterJob job = PrinterJob.getPrinterJob();
-        job.setPrintable(new DialogOrient());
         PageFormat landscape = job.pageDialog(job.defaultPage());
+        job.setPrintable(new DialogOrient(), landscape);
```
alisenchung, line 73: why doesn't job.setPrintable(DialogOrient) work here? doesn't job.pageDialog already set the job orientation to landscape?

prsadhuk: https://docs.oracle.com/en/java/javase/20/docs/api/java.desktop/java/awt/print/PrinterJob.html#setPrintable(java.awt.print.Printable)
says "The [PageFormat](file:///D:/jdk-17.0.1_doc-all/docs/api/java.desktop/java/awt/print/PageFormat.html) for each page is the default page format."
which is Portrait so it seems prudent to pass the pageformat modified in pageDialog to setPrintable

honkar-jdk: @prsadhuk 

The test works with the fix. The redundant if block can be removed, it is repeated twice in the test Ln#78.

```
 if (job.printDialog()){
            job.print();
        }
```

honkar-jdk (commented):
This can be taken up as a separate issue - The Standard Test Machinery can be removed  and PassFailJFrame framework can be used to clean-up this test.

honkar-jdk (approved):

DamonGuy (approved):
Fix makes sense as the landscape var wasn't even used prior, which doesn't make sense as to why it was added in the first place. I tested with and without the change on my local MacOS machine and it now passes.

alisenchung (approved):

jayathirthrao (approved):
Please add latest copyright year before pushing.

Was able to reproduce the issue and its fix.
I can see why we need to pass our PageFormat to setPrintable().

Strange that, we have this "landscape" variable and it is not used to pass appropriate page format. Test update seems fine.

PR #13956, opened 2023-05-12
8307990: jspawnhelper must close its writing side of a pipe before reading from it

simonis:
Since JDK13, executing commands in a sub-process defaults to the so called `POSIX_SPAWN` launching mechanism (i.e. `-Djdk.lang.Process.launchMechanism=POSIX_SPAWN`) on Linux. This works by using `posix_spawn(3)` to firstly start a tiny helper program called `jspawnhelper` in a subprocess. In a second step, `jspawnhelper` reads the command data from the parent Java process over a Unix pipe and finally executes (i.e. `execvp(3)`) the requested command.

In cases where the parent process terminates abnormally before `jspawnhelper` has read all the expected data from the pipe, `jspawnhelper` will block indefinitely on the reading end of the pipe. This is especially harmful if the parent process had open sockets, because in that case, the forked `jspawnhelper` process will inherit them and keep all the corresponding ports open effectively preventing other processes to bind to them. Notice that this is not an abstract scenario. We've observed this regularly in production with services which couldn't be restarted after a crash after migrating to JDK 17.

The fix of the issue is rather trivial. `jspawnhelper` has to close its writing end of the pipe which connects it with the parent Java process *before* starting to read from that pipe such that reads from the pipe can immediately return with EOF if the parent process terminates abnormally.

Also did some cleanup:
 - in `jspawnhelper.c` correctly chek the result of `readFully()`
 - in `ProcessImpl_md.c` use `restartableWrite()` instead of `write()` to write the command data from the parent process to `jspawnhelper`
 - in `childproc.{c,h}` remove remaining traces of `MODE_CLONE` because that's long gone.

---------
### Progress
- [x] Change must be properly reviewed (2 reviews required, with at least 2 [Reviewers](https://openjdk.org/bylaws#reviewer))

### Issue
 * [JDK-8307990](https://bugs.openjdk.org/browse/JDK-8307990): jspawnhelper must close its writing side of a pipe before reading from it

### Reviewers
 * [Roger Riggs](https://openjdk.org/census#rriggs) (@RogerRiggs - **Reviewer**) ⚠️ Review applies to [13fb0bf7](https://git.openjdk.org/jdk/pull/13956/files/13fb0bf74702381f1e86933758ac481393dfe163)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13956/head:pull/13956` \
`$ git checkout pull/13956`

`$ git checkout pull/13956` \
`$ git pull https://git.openjdk.org/jdk.git pull/13956/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13956`

`$ git pr show -t 13956`

</details>

<a href="https://git.openjdk.org/jdk/pull/13956.diff">https://git.openjdk.org/jdk/pull/13956.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13956#issuecomment-1545929898)


RogerRiggs:
Looks ok.

Is it practical to write a test for this situation? 
Can I assume you've validated the improvement as a remedy for the observed hangs?

tstuefe:
Trying to understand this, and the scope of the problem:

- Parent process opens "in" pipe. 
- Now owns read and write end of pipe. 
- Parent forks jspawnhelper. 
- jspawnhelper inherits handles. Now there are four open file descriptors to this pipe.
- Parent dies and releases its handles
- jspawnhelper gets thrown out of its first "readFully", but enters the next "readFully". It still owns both read and write end of the pipe. So it will wait endlessly. We only get an EOF from a pipe read when all write ends are closed.

Is this the problem?

I wonder whether the parent may not have the same problem at the other end. It may be prudent to close the child's ends of the three pipes right after calling startChild() (https://github.com/openjdk/jdk/blob/020a60e6449749ca979c1ace3af7db57f4c53a5f/src/java.base/unix/native/libjava/ProcessImpl_md.c#LL681C17-L681C27).

simonis:
> Looks ok.
> 
> Is it practical to write a test for this situation? Can I assume you've validated the improvement as a remedy for the observed hangs?

I thought about a test but couldn't come up with a practical way to write it. The JVM has to exit in the time frame after the `posix_spawn()` and before `jspawnhelper` has read its data from the parent. In production this usually happens due to memory constraints on the particular host which let the OOM-killer kill the JVM process because it is the biggest memory consumer.

The production issue could be mitigated by going back to using vfork (i.e. `-Djdk.lang.Process.launchMechanism=VFORK`). I could locally reproduce the problem manually in gdb and with the help of a small C program. I've verified that the proposed change of closing the child's write end of the pipe fixes the problem both in the small C example as well as in the JDK if the issue is reproduced manually in gdb.

simonis:
> Trying to understand this, and the scope of the problem:
> 
>     * Parent process opens "in" pipe.
> 
>     * Now owns read and write end of pipe.
> 
>     * Parent forks jspawnhelper.
> 
>     * jspawnhelper inherits handles. Now there are four open file descriptors to this pipe.
> 
>     * Parent dies and releases its handles
> 
>     * jspawnhelper gets thrown out of its first "readFully", but enters the next "readFully". It still owns both read and write end of the pipe. So it will wait endlessly. We only get an EOF from a pipe read when all write ends are closed.
> 
> Is this the problem? 

Exactly, except that `jspawnhelper` doesn't even gets "thrown out" of the first `readFully()` it is waiting in, as long as there's an open write end for that pipe.

> I wonder whether the parent may not have the same problem at the other end. It may be prudent to close the child's ends of the three pipes right after calling startChild() (https://github.com/openjdk/jdk/blob/020a60e6449749ca979c1ace3af7db57f4c53a5f/src/java.base/unix/native/libjava/ProcessImpl_md.c#LL681C17-L681C27).

But the parent first only reads from `fail[0]` (both, the keep-alive message as well as the EOF on a successfull `execvp()`). And it does close its write end of that pipe before doing the first read, see:  https://github.com/openjdk/jdk/blob/020a60e6449749ca979c1ace3af7db57f4c53a5f/src/java.base/unix/native/libjava/ProcessImpl_md.c#L698

Once it made sure the child process was executed, it also closes the corresponding ends of the in/out/err pipes which connect it with the child process (see https://github.com/openjdk/jdk/blob/020a60e6449749ca979c1ace3af7db57f4c53a5f/src/java.base/unix/native/libjava/ProcessImpl_md.c#L738-L740).

tstuefe:
Thank you Volker for the explanations.

Some more notes:
	
1) I think the child should probably close (but without error handling) the *read* end of the *fail* pipe too before sending out the liveness ping.

https://github.com/openjdk/jdk/blob/020a60e6449749ca979c1ace3af7db57f4c53a5f/src/java.base/unix/native/libjava/childproc.c#L335

Because the same problem could arise if the parent were to terminate. Child writes to the pipe, and theoretically could block on a full pipe since it holds open an fd to the read end. I admit this is highly theoretical since the pipe would have to be full for this to happen, and nobody wrote to the pipe yet, and the aliveness ping is just an integer (4 bytes). But to be more correct, and since the fix is so trivial, the fail pipe read end should be closed in the child process before writing to fail pipe.

2) I think you don't actually have to hand in the in-pipe-read-end fd number via command line arg, just to have the child to close it. You could just, in the parent, set the fd to FD_CLOEXEC. Since posix_spawn() exec's the spawn helper, this would close the file descriptor for you. Extending this thought, you could do this with all pipe ends you want to cauterize in the child process.

tstuefe:
> > Looks ok.
> > Is it practical to write a test for this situation? Can I assume you've validated the improvement as a remedy for the observed hangs?
> 
> I thought about a test but couldn't come up with a practical way to write it. The JVM has to exit in the time frame after the `posix_spawn()` and before `jspawnhelper` has read its data from the parent. In production this usually happens due to memory constraints on the particular host which let the OOM-killer kill the JVM process because it is the biggest memory consumer.
> 

A regression test would be good.

This can be very simply a runtime switch that kills the parent process at vulnerable times. See my example here:

https://github.com/openjdk/jdk/compare/master...tstuefe:jdk:test-for-parent-premature-death

Using that (KILLTEST=1), I was able to reproduce your problem with the hanging child. Using this to build a jtreg test is not hard.

Cheers, Thomas

simonis:
>     1. I think the child should probably close (but without error handling) the _read_ end of the fail pipe too _before_ sending out the liveness ping.
> 
> 
> https://github.com/openjdk/jdk/blob/020a60e6449749ca979c1ace3af7db57f4c53a5f/src/java.base/unix/native/libjava/childproc.c#L335
> 
> Because the same problem could arise if the parent were to terminate. Child writes to the pipe, and theoretically could block on a full pipe since it holds open an fd to the read end. I admit this is highly theoretical since the pipe would have to be full for this to happen, and nobody wrote to the pipe yet, and the aliveness ping is just an integer (4 bytes). But to be more correct, and since the fix is so trivial, the fail pipe read end should be closed in the child process before writing to fail pipe.

I thought about this, but as you said, I think it can not happen as we only write at most two times 4 bytes. Also, if we close the read end of the fail pipe in the child and the parent crashes (i.e. there's no read end at all), we will get a SIGPIPE which we would then have to handle in `jspawnhelper`. In the end, I don't think that's worth it for an event which should never happen (at least not in this PR).

> 
>     2. I think you don't actually have to hand in the in-pipe-read-end fd number via command line arg, just to have the child to close it. You could just, in the parent, set the fd to FD_CLOEXEC. Since posix_spawn() exec's the spawn helper, this would close the file descriptor for you. Extending this thought, you could do this with all pipe ends you want to cauterize in the child process.

I've deliberately not used `FD_CLOEXEC`  because the file descriptor closing code in `childProcess()` is shared by all launch mechanisms. So to make it work, I'd had to reset the corresponding file descriptors to `-1` in the child anyway.

I therefor felt the current fix is smaller, simpler and easier to understand.

simonis:
> > > Looks ok.
> > > Is it practical to write a test for this situation? Can I assume you've validated the improvement as a remedy for the observed hangs?
> > 
> > 
> > I thought about a test but couldn't come up with a practical way to write it. The JVM has to exit in the time frame after the `posix_spawn()` and before `jspawnhelper` has read its data from the parent. In production this usually happens due to memory constraints on the particular host which let the OOM-killer kill the JVM process because it is the biggest memory consumer.
> 
> A regression test would be good.
> 
> This can be very simply a runtime switch that kills the parent process at vulnerable times. See my example here:
> 
> [master...tstuefe:jdk:test-for-parent-premature-death](https://github.com/openjdk/jdk/compare/master...tstuefe:jdk:test-for-parent-premature-death)
> 
> Using that (KILLTEST=1), I was able to reproduce your problem with the hanging child. Using this to build a jtreg test is not hard.
> 
> Cheers, Thomas

Do you really think it makes sense to add such test code to the native runtime libraries?

I'm not sure. What do others think?

tstuefe:
> > > > Looks ok.
> > > > Is it practical to write a test for this situation? Can I assume you've validated the improvement as a remedy for the observed hangs?
> > > 
> > > 
> > > I thought about a test but couldn't come up with a practical way to write it. The JVM has to exit in the time frame after the `posix_spawn()` and before `jspawnhelper` has read its data from the parent. In production this usually happens due to memory constraints on the particular host which let the OOM-killer kill the JVM process because it is the biggest memory consumer.
> > 
> > 
> > A regression test would be good.
> > This can be very simply a runtime switch that kills the parent process at vulnerable times. Sely be my example here:
> > [master...tstuefe:jdk:test-for-parent-premature-death](https://github.com/openjdk/jdk/compare/master...tstuefe:jdk:test-for-parent-premature-death)
> > Using that (KILLTEST=1), I was able to reproduce your problem with the hanging child. Using this to build a jtreg test is not hard.
> > Cheers, Thomas
> 
> Do you really think it makes sense to add such test code to the native runtime libraries?

Debug only? Sure. The fact that this error is in there since JDK 13 is scary.

> 
> I'm not sure. What do others think?

simonis:
> The fact that this error is in there since JDK 13 is scary.

The error is there since ten years since the `posix_spawn` mechanism was initially introduced with [JDK-5049299](https://bugs.openjdk.org/browse/JDK-5049299) for JDK 8 and downported to JDK 7 :)

In JDK 13 it only became the default mechanism on Linux.

And although LTS releases are not different from any other releases (reference pending :) it was only detected when migrations to JDK 17 finally started to ramp-up.

RogerRiggs:
I think writing a test is worth a little bit of time but getting a clean test run using the KILLTEST might be a bit tricky. The exit(-1) of the parent would be detected by jtreg as a failure. And verifying that the spawned child runs to completion might be difficult without writing a driver app to monitor the child; though it may be sufficient to know just that it does not hang and timeout. When the parent dies, the child will be re-parented.
And it would only be run in debug builds.

ecki:
Independent of the actual fix, it was mentioned that it can block listening sockets, aren’t those close on exec? Also should a write and read timeout be used in addition? At least to call the close descriptor code before retrying?

dholmes-ora:
I wonder if @Martin-Buchholz is able to look at this one?

My concern with changes like this is that they fix an issue but then have unexpected side-effects themselves.

tstuefe:
@RogerRiggs @simonis 

Okay, I wrote a jtreg test case for the issue, just to see if I could. Rebased atop of Volkers change, you can just take the commit if you want:

https://github.com/tstuefe/jdk/tree/Added-TestCase-For-Volker
https://github.com/openjdk/jdk/commit/e7e0d2067cedbb750859ee04361e15a98d1e517c

Test case fails immediately (so, no relying on timeouts) if Volkers patch is missing. Succeeds immediately with Volkers patch. I also tried commenting out the fixing close() call, test fails immediately.

Patch relies on jspawnhelper to write a small marker file on communication error with parent if KILLTEST is set. Only in debug code. I think this is very reliable. No second-guessing aliveness of the child and worrying about PID reuse. Re-parenting is no issue either. And I think such a file make be a good idea anyway to analyze handshake errors with jspawnhelper.

Tested on Linux x64.

@simonis, feel free to add me as contributor if you take this test. I spent way to much time on this :-)

tstuefe:
> I wonder if @Martin-Buchholz is able to look at this one?
> 
> My concern with changes like this is that they fix an issue but then have unexpected side-effects themselves.

Are there any specific concerns you have?

dholmes-ora:
> > I wonder if @Martin-Buchholz is able to look at this one?
> > My concern with changes like this is that they fix an issue but then have unexpected side-effects themselves.
> 
> Are there any specific concerns you have?

There seems to be more going on here than just closing the write side of the pipe. We will now error out if any of the readFully's do a short read, rather than just when reporting an error - which on the surface seems like a good thing, but what if harmless short-reads can actually happen in some contexts?

tstuefe:
> > > I wonder if @Martin-Buchholz is able to look at this one?
> > > My concern with changes like this is that they fix an issue but then have unexpected side-effects themselves.
> > 
> > 
> > Are there any specific concerns you have?
> 
> There seems to be more going on here than just closing the write side of the pipe. We will now error out if any of the readFully's do a short read, rather than just when reporting an error - which on the surface seems like a good thing, but what if harmless short-reads can actually happen in some contexts?

I agree, it would be clearer to just to the close.

simonis:
> > > > I wonder if @Martin-Buchholz is able to look at this one?
> > > > My concern with changes like this is that they fix an issue but then have unexpected side-effects themselves.
> > > 
> > > 
> > > Are there any specific concerns you have?
> > 
> > 
> > There seems to be more going on here than just closing the write side of the pipe. We will now error out if any of the readFully's do a short read, rather than just when reporting an error - which on the surface seems like a good thing, but what if harmless short-reads can actually happen in some contexts?
> 
> I agree, it would be clearer to just to the close.

Sorry, but I don't understand this argument. If we do a short read we will work with corrupted `ChildStuff` and `SpawnInfo` structures. This can in the extreme case execute arbitrary code (e.g. if `ChildStuff.argv` is not fully read from the parent). You are basically saying it is better to work on corrupted data rather than reporting an error.

simonis:
> @RogerRiggs @simonis
> 
> Okay, I wrote a jtreg test case for the issue, just to see if I could. Rebased atop of Volkers change, you can just take the commit if you want:
> 
> https://github.com/tstuefe/jdk/tree/Added-TestCase-For-Volker [9119b44](https://github.com/openjdk/jdk/commit/9119b442dac8dd95228d47d70871ee59862649ce)
> 
> Test case fails immediately (so, no relying on timeouts) if Volkers patch is missing. Succeeds immediately with Volkers patch. I also tried commenting out the fixing close() call, test fails immediately.
> 
> Patch relies on jspawnhelper to write a small marker file on communication error with parent if KILLTEST is set. Only in debug code. I think this is very reliable. No second-guessing aliveness of the child and worrying about PID reuse. Re-parenting is no issue either. And I think such a file make be a good idea anyway to analyze handshake errors with jspawnhelper.
> 
> Tested on Linux x64.
> 
> @simonis, feel free to add me as contributor if you take this test. I spent way to much time on this :-)

Thank you Thomas. Unfortunately I saw this too late yesterday, when I already almost finished my version :)

I've now also added some "simulted crashes" to `jspawnhelper` at different points of the handshake. This didn't revealed any new issues but might be helpful for the future.

simonis:
> Should this issue have a CSR for any behavioral changes?

Well, you can certainly argue that every bug fix is a behavioral changes, right :)

But seriously, I don't see how this PR could require a CSR. The only behavioral change is really that `jspawnhelper` can now no longer block and deadlock. But that's exactly how it should have behaved in the first place.

simonis:
> Independent of the actual fix, it was mentioned that it can block listening sockets, aren’t those close on exec? Also should a write and read timeout be used in addition? At least to call the close descriptor code before retrying?

While the JDK opens files with `FD_CLOEXEC` by default, this is not the case for sockets. There have been attempts to fix this (see "[JDK-8207335: Use SOCK_CLOEXEC when creating sockets and channels](https://bugs.openjdk.org/browse/JDK-8207335)", [mailing list thread](https://mail.openjdk.org/pipermail/net-dev/2018-July/thread.html#11622)) but they never made it into mainline.

Notice that this is usually only a problem if you fork from native code (e.g. in a JNI library). Otherwise the JDK takes care to close *all* file descriptors before `execvp()`ing a new process (see [this code in `childProcess()` in `childproc.c`](https://github.com/openjdk/jdk/blob/5763be726700be322de3bbaf345d80e11936b472/src/java.base/unix/native/libjava/childproc.c#L363-L369)).

In this specific case the sockets were not closed because `jspawnhelper` blocked very early, *before* cleaning up the file descriptors in `childProcess()`. But that's now fixed with this PR :)

src/java.base/unix/native/libjava/ProcessImpl_md.c
```diff
@@ -487,14 +487,14 @@ static pid_t
 spawnChild(JNIEnv *env, jobject process, ChildStuff *c, const char *helperpath) {
     pid_t resultPid;
     int i, offset, rval, bufsize, magic;
-    char *buf, buf1[16];
+    char *buf, buf1[24];
```
tstuefe, line 490: Since you are here could you please increase buffer size to something safe? Max len of INT_MIN is 11 bytes, so lets have at least 3*11 + 3 bytes.

tstuefe (commented):
The fixes look reasonable, though I would prefer them to be split over individual RFEs. Would make backporting easier too.

I would still like an automated test as part of the jtreg suite. Please see my test proposal, that thing should be adaptable to work with your fix.

RogerRiggs:
> > Should this issue have a CSR for any behavioral changes?
> 
> Well, you can certainly argue that every bug fix is a behavioral changes, right :)
> 
> But seriously, I don't see how this PR could require a CSR. The only behavioral change is really that `jspawnhelper` can now no longer block and deadlock. But that's exactly how it should have behaved in the first place.

Not quite the bar needed for a CSR, but it might be useful to write a release note with a more problem oriented descriptive title. The current title describes the solution, not the problem.
Something that includes the ideas in "process launch can hang if parent process dies prematurely".

Anyone observing this behavior in their own operations would not find this this issue.
A release note would provide an explanation and closure for issues they might have been seeing.

ecki:
> If we do a short read we will work with corrupted `ChildStuff` and `SpawnInfo` structures.

i would agree failing after readFully returns short reads is fine. It won’t stop the method from retrying/completing endlessly in a pathological condition (like the missing close), though).

src/java.base/unix/native/libjava/childproc.c
```diff
@@ -396,3 +399,19 @@ childProcess(void *arg)
     _exit(-1);
     return 0;  /* Suppress warning "no return value from function" */
 }
+
+#ifdef DEBUG
+/* This method is only used in debug builds for testing MODE_POSIX_SPAWN
+ * in the light of abnormal program termination of either the parent JVM
+ * or the newly created jspawnhelper child process during the execution of
+ * Java_java_lang_ProcessImpl_forkAndExec().
+ * See: test/jdk/java/lang/ProcessBuilder/JspawnhelperProtocol.java
```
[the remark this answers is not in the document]
simonis: Sorry, I could say I wanted to see if somebody is really looking at this, but in reality I just missed to add the new test file to the commit :)

src/java.base/unix/native/jspawnhelper/jspawnhelper.c
```diff
@@ -132,12 +137,15 @@ int main(int argc, char *argv[]) {
     ChildStuff c;
     struct stat buf;
     /* argv[0] contains the fd number to read all the child info */
-    int r, fdin, fdout;
+    int r, fdinr, fdinw, fdout;
```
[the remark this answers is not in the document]
simonis: What would be the right value to initialize them to and what would it help? We use `sscanf()` right in the following line and exit if `sscanf()` is not assigning all three values. I think that should be good enough.

src/java.base/unix/native/libjava/ProcessImpl_md.c
simonis: OK, so let it be (until we get 64-bit integers and file descriptors :)

simonis:
> if parent process dies prematurely

I've created a release note for the issue based on your suggestions. Please feel free to proof-read and enhance it :)

src/java.base/unix/native/jspawnhelper/jspawnhelper.c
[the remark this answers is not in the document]
tstuefe: I usually init fds with -1. It's mostly a protection against bitrot and general cleanliness.

test/jdk/java/lang/ProcessBuilder/JspawnhelperProtocol.java
```diff
@@ -0,0 +1,176 @@
...
+        } catch (Exception e) {
+            System.out.println(e.toString());
+            System.exit(ERROR);
+        }
+        if (!p.waitFor(TIMEOUT, TimeUnit.SECONDS)) {
+            System.exit(ERROR + 1);
+        }
+        if (p.exitValue() == 0) {
+            String pwd = p.inputReader().readLine();
+            if (!Path.of("").toAbsolutePath().toString().equals(pwd)) {
```
RogerRiggs, line 63: It would be useful to print the unexpected string; it might help isolate what went wrong.
And make it easier to understand what ERROR+2 means.

test/jdk/java/lang/ProcessBuilder/JspawnhelperProtocol.java
```diff
@@ -0,0 +1,176 @@
...
+            System.out.println("  Failed to executed '" + CMD[0] + "' (exitValue=" + p.exitValue() + ")");
+            System.exit(ERROR + 3);
+        }
+    }
+
+    private static void normalExec() throws Exception {
+        ProcessBuilder pb;
+        pb = ProcessTools.createJavaProcessBuilder("-Djdk.lang.Process.launchMechanism=posix_spawn",
+                                                   "JspawnhelperProtocol",
+                                                   "normalExec");
```
RogerRiggs, line 78: I would just redirect the output to the parent.
`pb.inheritIO()`  and avoid the extra code at line 88.

test/jdk/java/lang/ProcessBuilder/JspawnhelperProtocol.java
```diff
@@ -0,0 +1,176 @@
...
+        BufferedReader br = p.inputReader();
+        String line = br.readLine();
+        boolean foundCrashInfo = false;
+        while (line != null) {
+            System.out.println(line);
+            if (line.equals("posix_spawn:0")) {
+                foundCrashInfo = true;
+            }
+            line = br.readLine();
+        }
```
RogerRiggs, line 106: Try-with-resources works well in cases like this.  (Moving foundCrashInfo out of the t-w-r).

test/jdk/java/lang/ProcessBuilder/JspawnhelperProtocol.java
```diff
@@ -0,0 +1,176 @@
...
+        if (!foundCrashInfo) {
+            throw new Exception("Wrong output from child process");
+        }
+        if (!p.waitFor(TIMEOUT, TimeUnit.SECONDS)) {
+            System.exit(ERROR + 5);
+        }
+
+        int ret = p.exitValue();
+        if (ret == 0) {
+            throw new Exception("Expected error in child execution");
```
RogerRiggs, line 116: Mixing the two styles of error reporting seems a bit odd, some throw exceptions and others just exit.
Being consistent throwing an exception with a message would be better.

jddarcy:
> > Should this issue have a CSR for any behavioral changes?
> 
> Well, you can certainly argue that every bug fix is a behavioral changes, right :)
> 
> But seriously, I don't see how this PR could require a CSR. The only behavioral change is really that `jspawnhelper` can now no longer block and deadlock. But that's exactly how it should have behaved in the first place.

Yes, some judgement is needed on behavioral changes; this was broached in portions of my recent talk to the JCP EC:

"Contributing to OpenJDK: Participating in stewardship for the long-term"
https://jcp.org/aboutJava/communityprocess/ec-public/materials/2023-04-12/Contributing_to_OpenJDK_2023_04_12.pdf

The process spawning code is apparently tricky and subtle, so much so that this will be apparently the third attempt at it since JDK 12. (I still have memories of various platform-specific issues in process handling from my time maintaining the launcher.)

So I don't think the applicability of a CSR for this change should be dismissed and personally I would prefer a change like this be made early in JDK 22 rather than a few weeks before JDK 21 rampdown; it is easier to address any bug tail early in 22 rather than late in 21.

src/java.base/unix/native/libjava/ProcessImpl_md.c
tstuefe: Ok, if you let it be, I fix it in a separate RFE. Cleaner that way anyway. 

I filed https://bugs.openjdk.org/browse/JDK-8308350 and https://github.com/openjdk/jdk/pull/14045. 

Should I happen to push before you, this should be easily adaptable for you.

dholmes-ora:
> Sorry, but I don't understand this argument. If we do a short read we will work with corrupted ChildStuff and SpawnInfo 
> structures. This can in the extreme case execute arbitrary code (e.g. if ChildStuff.argv is not fully read from the parent). You are 
> basically saying it is better to work on corrupted data rather than reporting an error.

No I am simply pointing out that this has changed more than just the issue with close. Any maybe a short-read does indicate data "corruption" and maybe it should be a fatal error. But I don't know exactly how this might manifest so perhaps there are benign short-reads that actually do happen. Regardless it might be better to split this part out and focus on the close issue here.

RogerRiggs:
> > Sorry, but I don't understand this argument. If we do a short read we will work with corrupted ChildStuff and SpawnInfo
> > structures. This can in the extreme case execute arbitrary code (e.g. if ChildStuff.argv is not fully read from the parent). You are
> > basically saying it is better to work on corrupted data rather than reporting an error.
> 
> No I am simply pointing out that this has changed more than just the issue with close. And maybe a short-read does indicate data "corruption" and maybe it should be a fatal error. But I don't know exactly how this might manifest so perhaps there are benign short-reads that actually do happen. Regardless it might be better to split this part out and focus on the close issue here.

Given the purpose and implementation of the `readFully` function, I don't see how it can return anything other than an error or the full requested read length.

test/jdk/java/lang/ProcessBuilder/JspawnhelperProtocol.java
simonis: Good point. Changed the exits to exceptions except in `childCode()` where we need `System.exit()` in order to communicate different error states back to the parent process.

test/jdk/java/lang/ProcessBuilder/JspawnhelperProtocol.java
simonis: Done.

test/jdk/java/lang/ProcessBuilder/JspawnhelperProtocol.java
simonis: Done.

dholmes-ora:
> Given the purpose and implementation of the readFully function, I don't see how it can return anything other than an error or the full requested read length.

@RogerRiggs it will return < requested read length if it encounters EOF - i.e. a short read.

simonis:
> > > Should this issue have a CSR for any behavioral changes?
> > 
> > 
> > Well, you can certainly argue that every bug fix is a behavioral changes, right :)
> > But seriously, I don't see how this PR could require a CSR. The only behavioral change is really that `jspawnhelper` can now no longer block and deadlock. But that's exactly how it should have behaved in the first place.
> 
> Yes, some judgement is needed on behavioral changes; this was broached in portions of my recent talk to the JCP EC:
> 
> "Contributing to OpenJDK: Participating in stewardship for the long-term" https://jcp.org/aboutJava/communityprocess/ec-public/materials/2023-04-12/Contributing_to_OpenJDK_2023_04_12.pdf
> 
> The process spawning code is apparently tricky and subtle, so much so that this will be apparently the third attempt at it since JDK 12. (I still have memories of various platform-specific issues in process handling from my time maintaining the launcher.)
> 
> So I don't think the applicability of a CSR for this change should be dismissed

I've created a [Release Note](https://bugs.openjdk.org/browse/JDK-8308297) (please feel to review it :) which outlines the "*behavioral change*". But this change is really just a simple bug fix for the case where the JVM leaves a hanging process (`jspawnhelper`) behind when it crashes. I don't really see see how a CSR will fit in here unless we start to treat crashes as "defined behavior" and start  tracking them across releases.

If you still think a CSR will be needed, please advice how it should look like?
 
>  and personally I would prefer a change like this be made early in JDK 22 rather than a few weeks before JDK 21 rampdown; it is easier to address any bug tail early in 22 rather than late in 21.

First, this is a bug fix for an issue which impacts production usage of the JDK. Second, tt's a "first-day" bug introduced by [JDK-5049299](https://bugs.openjdk.org/browse/JDK-5049299) for JDK 8 (initially only active on Solaris). It started to manifest on Linux in JDK 13 when the `POSIX_SPAWN` launching mechanism was ported to Linux as well, **but** it was only detected some time after JDK 17 was released when a higher rate of production services started to migrate to JDK 17.

I really don't want to go into a bike-shedding discussion about "*LTS releases are not special*", but I think this is a serious enough bug with a simple enough fix to resolve it as quick as possible (and downport it to JDK 17 as fast as possible). We are still more than two weeks before RDP 1 and we still have more than three month before GA so I don't see a reason not to fix this now.

RogerRiggs:
/reviewers 2 @tstuefe should have a chance to comment/approve

jddarcy:
> 

Without further investigation on my part, I don't know if this fix falls more squarely in the "obviously no bugs" or "no obvious bugs" category.

I left a comment on the release note issue suggesting a description of the solution or implications of the solution be added to the note. It could be also be helpful to state under what conditions the problem being solved is more likely to manifest, if that is known.

For whenever this goes back, highlighting the change on a quality-discuss message could help validate the change doesn't have unexpected side-effects.

mlichtblau:
Hey, chiming on this conversation, because I believe it will fix an issue we encountered since, upgrading from JDK 11 to 17.
In our application we spawn a lot of processes over time and noticed that more and more JVM threads get stuck in the native `ProcessImpl.forkAndExec`.
<details>
<summary>Thread Dump Excerpt</summary>

Thread dump contains multiple threads in this state.
```
"redacted" #954 prio=5 os_prio=0 cpu=59718.38ms elapsed=286286.30s tid=0x00007f62f80201e0 nid=0x3c16 runnable  [0x00007f609ec56000]
   java.lang.Thread.State: RUNNABLE
	at java.lang.ProcessImpl.forkAndExec(java.base@17.0.7/Native Method)
	at java.lang.ProcessImpl.<init>(java.base@17.0.7/ProcessImpl.java:314)
	at java.lang.ProcessImpl.start(java.base@17.0.7/ProcessImpl.java:244)
	at java.lang.ProcessBuilder.start(java.base@17.0.7/ProcessBuilder.java:1110)
	at java.lang.ProcessBuilder.start(java.base@17.0.7/ProcessBuilder.java:1073)
```
</details>

After some investigation we saw that we have `jspawnhelper` processes corresponding to the number of the stuck threads.
`/proc/*/task/{id}/syscall` shows that all of these processes are currently reading from the pipe.

Our theory is that the write calls from the parent thread were interrupted resulting in an incomplete write. 
https://github.com/openjdk/jdk/blob/69f508a2ac344eb61cef7be985348873b8265171/src/java.base/unix/native/libjava/ProcessImpl_md.c#L556-L559
This would cause the `readFully` on the `jspawnhelper` side to wait for the remaining data.
As the number of bytes written is not checked on the writer side it would continue and wait for the alive ping from the `jspawnhelper` which results in a deadlock.
I would be interested in a reproducer e.g. using `gdb`, but to be honest I wouldn't know where to start.
But regardless of whether this fixes our issue I think incomplete writes should be handled according to specification.
I think the `restartableWrites` in this PR do that?

We also mitigated by falling back to `VFORK` as was also mentioned in [this comment](https://github.com/openjdk/jdk/pull/13956#issuecomment-1547777042). 

Maybe it would be worth it to mention our bug situation in the release notes as well?
Thank you and please let me know if I got something wrong or there is anything I could provide to help!

tstuefe (approved):
Looks good. 

Please enable for muslc (see remark). Otherwise, all other remarks are nits and I leave it up to you whether you accept them. I don't need another review.

test/jdk/java/lang/ProcessBuilder/JspawnhelperProtocol.java
```diff
@@ -0,0 +1,183 @@
...
+        pb.inheritIO();
+        Process p = pb.start();
+        if (!p.waitFor(TIMEOUT, TimeUnit.SECONDS)) {
+            throw new Exception("Child process timed out");
+        }
+        if (p.exitValue() != 0) {
+            throw new Exception("Child process exited with " + p.exitValue());
+        }
+    }
+
```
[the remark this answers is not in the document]
simonis: This sounds reasonable. I've renamed `childCode()` to `parentCode()`, streamlined the wording for "parent" and "child" in the various log messages and exceptions and added a comment to the `main()` method which explains the working of the test and the meaning of "parent" and "child" in the output.

I've not merged `simulateCrashInChild()` and `simulateCrashInParent()` because there are some subtle differences between the two (and now with the improved "parent"/"child" naming even more :) and I don't think that a merged version will improve the readability.

RogerRiggs (approved):
ok with latest doc and naming updates.

simonis:
@RogerRiggs , @tstuefe sorry for bothering you one more time, but I think @mlichtblau brought up an interesting case yesterday which isn't fully resolved by the current fix. In @mlichtblau example, `Java_java_lang_ProcessImpl_forkAndExec()` got stuck waiting on the ping from `jspawnhelper` which itself blocks in `readFully()` because of a truncated `write()` from the parent.

The current fix already replaces `write()` with `restartableWrite()` which handles the case where a `write()` call was interrupted **before** writing a single byte. But there's a second case, namely when `write()` was interrupted after it already wrote some (but not all) of the requested bytes. The `write()` man page states:

> Note  that a successful `write()` may transfer fewer than `count` bytes. Such partial writes can occur for various reasons; for example, because .., or because a blocked `write()` to  a  socket, pipe, or similar was interrupted by a signal handler after it had transferred some, but before it had transferred all of the requested bytes.  In the event of a partial write, the caller can make another `write()` call to transfer the remaining bytes.

So in order to safely handle the second case as well, we have to replace `restartableWrite()` with something like `writeFully()` and that's exactly what this new commit does. I've also added a new test case which reproduces the issue by simulating a truncated `write()`.

For your convenience I've tried to explain all the additional changes (except for trivial cleanups and renamings in the test) below:

Thanks in advance,
Volker

##### `jspawnhelper.c`:

```
+    // The file descriptor for reporting errors back to our parent we got on the command
+    // line should be the same like the one in the ChildStuff struct we've just read.
+    assert(c.fail[1] == fdout);
```

Just trying to be overly cautious.

##### childproc.{c,h}

```
-ssize_t restartableWrite(int fd, const void *buf, size_t count);
+ssize_t writeFully(int fd, const void *buf, size_t count);
```

```
-ssize_t
-restartableWrite(int fd, const void *buf, size_t count)
...
-}

+ssize_t
+writeFully(int fd, const void *buf, size_t nbyte)
+#ifdef DEBUG
...
+#endif
...
+}
```

Replace `restartableWrite()` with `writeFully()` which is basically a copy of `readFully()` (with additional testing code for jtreg). This handles both cases, when `write()` is interrupted before having written a single byte and returns EINTR, as well as the case where `write()` already wrote some (but not all) bytes before it was interrupted. 

```
-        restartableWrite(fail_pipe_fd, &code, sizeof(code));
+        if (writeFully(fail_pipe_fd, &code, sizeof(code)) != sizeof(code)) {
+            goto WhyCantJohnnyExec;
+        }
```

Treat an incomplete write as error and bail out.

```
-        restartableWrite(fail_pipe_fd, &errnum, sizeof(errnum));
+        writeFully(fail_pipe_fd, &errnum, sizeof(errnum));
```

We're already in the error handling code here (i.e. `WhyCantJohnnyExec`) so not much we can do. We'll close the pipe and exit with an error anyway.

##### `ProcessImpl_md.c`

```
-    restartableWrite(c->childenv[1], (char *)&magic, sizeof(magic)); // magic number first
+    if (writeFully(c->childenv[1], (char *)&magic, sizeof(magic)) != sizeof(magic)) { // magic number first
+        return -1;
+    }
...
-    restartableWrite(c->childenv[1], (char *)c, sizeof(*c));
-    restartableWrite(c->childenv[1], (char *)&sp, sizeof(sp));
-    restartableWrite(c->childenv[1], buf, bufsize);
+    if (writeFully(c->childenv[1], (char *)c, sizeof(*c)) != sizeof(*c) ||
+        writeFully(c->childenv[1], (char *)&sp, sizeof(sp)) != sizeof(sp) ||
+        writeFully(c->childenv[1], buf, bufsize) != bufsize) {
+        return -1;
+    }
```

Make sure we've written all the required data to jspawnhelper or return an error otherwise. The `-1` return value will be handled in `Java_java_lang_ProcessImpl_forkAndExec()` by throwing an `IOException("posix_spawn failed")`.

```
+    /* We're done. Let jspwanhelper know he can't expect any more data from us. */
+    close(c->childenv[1]);
+    c->childenv[1] = -1;
```

Close the write end of the pipe to jspawnhelper after we've written all data. This will prevent `jspawnhelper` to block on the reading side if somthing goes wrong (e.g. we've not written enough data).

```
+    // Reset errno to protect against bogus error messages
+    errno = 0;
```

In `Java_java_lang_ProcessImpl_forkAndExec()` we might call `throwIOException(env, errno, ..)` in error situations not caused by a failing system call (e.g. if `startChild()` returned `-1` because of a bad response from `jspawnhelper`). It is therefor better to reset `errno` at the beginning of the method to avoid bogus error messages in the generated `IOExceptions`s.

```
-            assert(errnum == CHILD_IS_ALIVE);
             if (errnum != CHILD_IS_ALIVE) {
-                /* Should never happen since the first thing the spawn
-                 * helper should do is to send an alive ping to the parent,
-                 * before doing any subsequent work. */
+                /* This can happen if the spawn helper encounters an error
+                 * before or during the handshake with the parent. */
                 throwIOException(env, 0, "Bad code from spawn helper "
```

This assertion isn't useful. We'd rather want to see the generated exception if `errnum != CHILD_IS_ALIVE` instead of crashing the process. Also, the comment is wrong, because `jspawnhelper` does quite some stuff (e.g. allocating, reading from parent) before sending the alive ping and if these actions fail, `jspawnhelper` will send an error code (e.g. `ERR_MALLOC`, `ERR_PIPE`) which is not equal to `CHILD_IS_ALIVE`.

##### JspawnhelperProtocol.java

```
+    private static void simulateTruncatedWriteInParent(int stage) throws Exception {
...
+    }
```

Added a new test to simualte a truncated write in the parent process to jspawnhelper communication.

```
-            Optional<String> cmd = ph.info().command();
-            if (cmd.isPresent() && cmd.get().endsWith("jspawnhelper")) {
-                throw new Exception("jspawnhelper still alive after parent Java process terminated");
+            try {
+                // Give jspawnhelper a chance to exit gracefully
+                ph.onExit().get(TIMEOUT, TimeUnit.SECONDS);
+            } catch (TimeoutException te) {
+                Optional<String> cmd = ph.info().command();
+                if (cmd.isPresent() && cmd.get().endsWith("jspawnhelper")) {
+                    throw new Exception("jspawnhelper still alive after parent Java process terminated");
+                }
```

Make the `simulateCrashInParent()` sub-test more robust by allowing `jspawnhelper` up to `TIMEOUT` seconds to terminate.

tstuefe:
@simonis I try to look at this before the weekend, but no guarantees. Your analysis looks sound, but the code is sensitive and I want to give it my full attention.

One remark though, you should change JBS title and issue text to reflect the growing number of issues this patch is fixing.

tstuefe:
> Fixing the last bug :)
> 
> If we close the writing end of the pipe to jspawnhelper early in `spawnChild()` right after the last write (which I still think is a good idea), we have to use `c->childenv` rather than `childenv` when closing the pipe descriptors at the end of `Java_java_lang_ProcessImpl_forkAndExec()` in order to avoid double closing:
> 
> ```c++
> -    closeSafely(childenv[0]);
> -    closeSafely(childenv[1]);
> +    /* We use 'c->childenv' here rather than 'childenv' because 'spawnChild()' might have
> +     * already closed 'c->childenv[1]' and signaled this by setting 'c->childenv[1]' to '-1'.
> +     * Otherwise 'c->childenv' and 'childenv' are the same because we just copied 'childenv'
> +     * to 'c->childenv' (with 'copyPipe()') before calling 'startChild()'. */
> +    closeSafely(c->childenv[0]);
> +    closeSafely(c->childenv[1]);
> ```
> 
> This also fixes `test/jdk/java/lang/ProcessBuilder/CloseRace.java` which could failed sporadically the previous version of the change.

I missed that. Christ this stuff is complex :( 

Still think it would be cleaner and simpler to set the FD in the parent to CLOEXEC, before doing posix_spawn, and at the same time set the childStuff variable to -1 to prevent the child from attempting to re-close it. Reconsider?

RogerRiggs (approved):
Shaping up nicely, thanks for the careful attention to detail.

tstuefe (approved):
I gave this a cursory eye, unfortunately i'm too snowed in atm for more. It looks good to me, your reasoning makes sense. Very good job finding and fixing this, and describing the issue. 

About FD_Cloexec , thus can be done in a later change.

simonis:
Thanks @RogerRiggs and @tstuefe for your help and patience with this PR. I've now updated the [Release Note](https://bugs.openjdk.org/browse/JDK-8308297) for this change to also include the error scenario described by @mlichtblau .

mlichtblau:
Hey @simonis, thank you for taking the time to include our problem in the fix. I appreciate the detailed explanation of your fix!

PR #13961, opened 2023-05-12
8305895: Implementation: JEP 450: Compact Object Headers (Experimental)

rkennke:
This is the main body of the JEP 450: Compact Object Headers (Experimental).

Main changes:
 - Introduction of the (experimental) flag UseCompactObjectHeaders. All changes in this PR are protected by this flag.
 - The compressed Klass* can now be stored in the mark-word of objects. In order to be able to do this, we are building on #10907, #13582 and #13779 to protect the relevant (upper 32) bits of the mark-word. Significant parts of this PR deal with loading the compressed Klass* from the mark-word, and dealing with (monitor-)locked objects. When the object is monitor-locked, we load the displaced mark-word from the monitor, and load the compressed Klass* from there. This PR also changes some code paths (mostly in GCs) to be more careful when accessing Klass* (or mark-word or size) to be able to fetch it from the forwardee in case the object is forwarded, and/or reach through to the monitor when the object is locked by a monitor.
 - The identity hash-code is narrowed to 25 bits.
 - Instances can now have their base-offset (the offset where the field layouter starts to place fields) at offset 8 (instead of 12 or 16).
 - Arrays will can now store their length at offset 8. Due to alignment restrictions, array elements will still start at offset 16. #11044 will resolve that restriction and allow array elements to start at offset 12 (except for long, double and uncompressed oops, which are still required to start at an element-aligned offset).
 - CDS can now write and read archives with the compressed header. However, it is not possible to read an archive that has been written with an opposite setting of UseCompactObjectHeaders.

Testing:
 (+UseCompactObjectHeaders tests are run with the flag hard-patched into the build, to also catch @flagless tests, and to avoid mismatches with CDS - see above.)
 - [x] tier1 (x86_64)
 - [x] tier2 (x86_64)
 - [x] tier3 (x86_64)
 - [ ] tier4 (x86_64)
 - [x] tier1 (aarch64)
 - [x] tier2 (aarch64)
 - [x] tier3 (aarch64)
 - [ ] tier4 (aarch64)
 - [ ] tier1 (x86_64) +UseCompactObjectHeaders
 - [ ] tier2 (x86_64) +UseCompactObjectHeaders
 - [ ] tier3 (x86_64) +UseCompactObjectHeaders
 - [ ] tier4 (x86_64) +UseCompactObjectHeaders
 - [ ] tier1 (aarch64) +UseCompactObjectHeaders
 - [ ] tier2 (aarch64) +UseCompactObjectHeaders
 - [ ] tier3 (aarch64) +UseCompactObjectHeaders
 - [ ] tier4 (aarch64) +UseCompactObjectHeaders

---------
### Progress
- [ ] Change requires CSR request [JDK-8306000](https://bugs.openjdk.org/browse/JDK-8306000) to be approved

### Integration blockers
&nbsp;⚠️ Dependency #13582 must be integrated first
&nbsp;⚠️ Too few reviewers with at least role reviewer found (have 0, need at least 1) (failed with updated jcheck configuration in pull request)
&nbsp;⚠️ Whitespace errors (failed with updated jcheck configuration in pull request)

### Issues
 * [JDK-8305895](https://bugs.openjdk.org/browse/JDK-8305895): Implementation: JEP 450: Compact Object Headers (Experimental) (**Enhancement** - P4)
 * [JDK-8306000](https://bugs.openjdk.org/browse/JDK-8306000): Add experimental -XX:+UseCompactObjectHeaders flag (**CSR**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13961/head:pull/13961` \
`$ git checkout pull/13961`

`$ git checkout pull/13961` \
`$ git pull https://git.openjdk.org/jdk.git pull/13961/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13961`

`$ git pr show -t 13961`

</details>

<a href="https://git.openjdk.org/jdk/pull/13961.diff">https://git.openjdk.org/jdk/pull/13961.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13961#issuecomment-1546164508)


src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -5120,8 +5120,8 @@ void MacroAssembler::load_method_holder(Register holder, Register method) {
 }
 
 #ifdef _LP64
-void MacroAssembler::load_nklass(Register dst, Register src) {
-  assert(UseCompressedClassPointers, "expect compressed class pointers");
+void MacroAssembler::load_nklass_compact(Register dst, Register src) {
+  assert(UseCompactObjectHeaders, "expect compact object headers");
 
   if (!UseCompactObjectHeaders) {
```
coleenp, line 5126: Now this isn't needed, right?

coleenp (commented):
These changes are an improvement.

src/hotspot/cpu/x86/c1_MacroAssembler_x86.cpp
```diff
@@ -163,24 +163,37 @@ void C1_MacroAssembler::try_allocate(Register obj, Register var_size_in_bytes, i
...
   if (len->is_valid()) {
     movl(Address(obj, arrayOopDesc::length_offset_in_bytes()), len);
+#ifdef _LP64
+    if (UseCompactObjectHeaders) {
+      // With compact headers, arrays have a 32bit alignment gap after the length.
+      assert(arrayOopDesc::length_offset_in_bytes() == 8, "check length offset");
+      xorptr(t1, t1);
+      movl(Address(obj, arrayOopDesc::length_offset_in_bytes() + sizeof(jint)), t1);
+    }
+#endif
```
coleenp, line 193: This endif should go after UseCompressedClassPointers conditional, and consolidate to one set of #ifdef _LP64.

src/hotspot/share/cds/archiveBuilder.cpp
```diff
@@ -717,6 +717,13 @@ void ArchiveBuilder::make_klasses_shareable() {
     const char* generated = "";
     Klass* k = klasses()->at(i);
     k->remove_java_mirror();
+#ifdef _LP64
+    if (UseCompactObjectHeaders) {
+      Klass* requested_k = to_requested(k);
+      narrowKlass nk = CompressedKlassPointers::encode_not_null(requested_k, _requested_static_archive_bottom);
+      k->set_prototype_header(markWord::prototype().set_narrow_klass(nk));
+    }
+#endif //_LP64
```
coleenp, line 726: If CDS is turned off for UseCompactObjectHeaders, I don't understand this change or the one to archiveHeapWriter.  -Xshare:dump objects would be the wrong size.  If CDS is not supported, then there should be something in arguments.cpp that gives an error for that.  And write a test for that error of mixing and matching.

rkennke: Yeah, we do have code in arguments.cpp that turns off CDS if the wrong setting is used (i.e. the opposite of the default setting). If you hard-code UseCompactObjectHeaders to be true, then the archives will be written in the compact layout, and can be read. That's what the changes in share/cds implement. (Note: I regularily hard-patch the UseCompactObjectHeaders flag to be true for testing, because that also catches all the @flagless tests, and I know that Daniel does that, too.)
We disabled CDS jtreg tests when passing +UseCompactObjectHeaders via cmd line, because we would see a lot of test failures because of archive format mismatch. I am not sure if that's a good way to deal with that.

iklam: You should add a check similar to this:

```
bool FileMapHeader::validate() {
  if (_obj_alignment != ObjectAlignmentInBytes) {
    log_info(cds)("The shared archive file's ObjectAlignmentInBytes of %d"
                  " does not equal the current ObjectAlignmentInBytes of %d.",
                  _obj_alignment, ObjectAlignmentInBytes);
    return false;
  }
```

Then, if the user wants to create a CDS archive that uses UseCompactObjectHeaders, they can do:

```
java -Xshare:dump -XX:+UseCompactObjectHeaders
java -Xshare:on -XX:+UseCompactObjectHeaders -cp . MyApp
```

Also, please create a new test case similar to this existing one:

- test/hotspot/jtreg/runtime/cds/CdsDifferentObjectAlignment.java

The checks you added in arguments.cpp should be removed.

rkennke: > You should add a check similar to this:
> 
> ```
> bool FileMapHeader::validate() {
>   if (_obj_alignment != ObjectAlignmentInBytes) {
>     log_info(cds)("The shared archive file's ObjectAlignmentInBytes of %d"
>                   " does not equal the current ObjectAlignmentInBytes of %d.",
>                   _obj_alignment, ObjectAlignmentInBytes);
>     return false;
>   }
> ```
> 
> Then, if the user wants to create a CDS archive that uses UseCompactObjectHeaders, they can do:
> 
> ```
> java -Xshare:dump -XX:+UseCompactObjectHeaders
> java -Xshare:on -XX:+UseCompactObjectHeaders -cp . MyApp
> ```
> 
> Also, please create a new test case similar to this existing one:
> 
> * test/hotspot/jtreg/runtime/cds/CdsDifferentObjectAlignment.java
> 
> The checks you added in arguments.cpp should be removed.

Thanks, Ioi, this is exactly what I've been looking for!
What about the tests-disabler, would that also not be needed anymore? It's been useful to disable CDS tests when somebody runs this with -XX:+UseCompactObjectHeaders on the cmd line.
https://github.com/rkennke/jdk/commit/e6607312b2c7e77ccd03d4d933b07b2e935d53ba

iklam: The CDS tests generate their own archive files. The VM options specified by JTREG are included in front of the command-line for running "java -Xshare:dump", etc. So in most cases, thee options are applied equally to both dump time and run time. This means that the CDS tests will not fail with differences in ObjectAlignmentInBytes, etc.

However, some tests may be more sensitive to the JTREG VM options. These tests may need to be fixed, or tagged with `@requires vm.flagless`.

I tested with `-vmoptions:-XX:ObjectAlignmentInBytes=16` and only 3 tests failed. We'll try to fix them in [JDK-8308034](https://bugs.openjdk.org/browse/JDK-8308034)

My suggestion is to run the CDS tests with -XX:+UseCompactObjectHeaders in the command-line (we will probably add such a configuration in our CI). This will ensure better coverage. I think most tests will pass, and we can deal with the ones that might fail.

src/hotspot/share/oops/markWord.hpp
```diff
@@ -107,12 +111,23 @@ class markWord {
   static const int self_forwarded_bits            = 1;
   static const int max_hash_bits                  = BitsPerWord - age_bits - lock_bits - self_forwarded_bits;
   static const int hash_bits                      = max_hash_bits > 31 ? 31 : max_hash_bits;
+  static const int hash_bits_compact              = max_hash_bits > 25 ? 25 : max_hash_bits;
```
merykitty, line 114: Maybe these constants should be `private` and be accessed through accessors instead.

coleenp:
This PR has a merged version of https://github.com/openjdk/jdk/pull/13582 that doesn't have UseAltGCForwarding as an experimental flag so fails tests with the flag because
Error: VM option 'UseAltGCForwarding' is experimental and must be enabled via -XX:+UnlockExperimentalVMOptions.
Other than that, testing tier1-8 with these options off on linux-x64 pass.

rkennke:
> This PR has a merged version of #13582 that doesn't have UseAltGCForwarding as an experimental flag so fails tests with the flag because Error: VM option 'UseAltGCForwarding' is experimental and must be enabled via -XX:+UnlockExperimentalVMOptions. Other than that, testing tier1-8 with these options off on linux-x64 pass.

Good about the testing! I'll fix the tests that use +UseAltGCForwarding.

src/hotspot/share/oops/markWord.hpp
```diff
@@ -44,6 +44,10 @@
 //  --------
 //  unused:25 hash:31 -->| unused_gap:1  age:4  unused_gap:1  lock:2 (normal object)
 //
+//  64 bits (with compact headers):
+//  -------------------------------
+//  nklass:32 hash:25 -->| unused_gap:1  age:4  self-fwded:1  lock:2 (normal object)
```
gerard-ziemski, line 49:
```
  static const int unused_gap_bits                = LP64_ONLY(1) NOT_LP64(0);
  static const int age_bits                       = 4;
  static const int lock_bits                      = 2;
  static const int self_forwarded_bits            = 1;
```

We have `32 + 25 + 1 + 4 + 1 + 2 == 65`, so how are fitting all these fields in 64 bits?

src/hotspot/share/oops/markWord.hpp
coleenp: Good suggestion.  Also have an accessor function so that the if (UseCompactHeaders) isn't exposed in the callers.

src/hotspot/share/oops/markWord.hpp
```diff
@@ -107,12 +111,23 @@ class markWord {
   static const int self_forwarded_bits            = 1;
   static const int max_hash_bits                  = BitsPerWord - age_bits - lock_bits - self_forwarded_bits;
   static const int hash_bits                      = max_hash_bits > 31 ? 31 : max_hash_bits;
+  static const int hash_bits_compact              = max_hash_bits > 25 ? 25 : max_hash_bits;
+  // Used only without compact headers.
   static const int unused_gap_bits                = LP64_ONLY(1) NOT_LP64(0);
```
gerard-ziemski, line 116: Is this bit reserved for future use? If so, we should call it reserved, if not why not give that bit back to the hash code?

src/hotspot/share/oops/markWord.hpp
rkennke: No wait, why? Those are constants. What do we win if we make accessors around them?

src/hotspot/share/oops/markWord.hpp
rkennke: That's a mistake. The unused-gap should not be there with compact headers. I'll fix that.

src/hotspot/share/oops/markWord.hpp
merykitty: What I mean is that we have an accessor like:

    static int hash_bits() {
        return UseCompactObjectHeaders ? _hash_bits_compact : _hash_bits_non_compact;
    }

rkennke: Right! That is a good suggestion! I'll do that in the next updated. Thank you!

fisk:
Looking at the generated code, the assumption seems to be that we can find the klass bits in either the object header, or the object monitor DMW. However, it doesn't appear that the inflation protocol has been changed away from using the INFLATING value (aka all zeros). Then the klass pointer is seemingly just nuked, together with all other bits. That means the various fast path checks that check for the monitor bit are going to not find a monitor bit (0b10). Therefore it will just shift the value resulting in a null klass pointer. Will we just crash if someone racingly reads the klass() during inflation? I can't see how that is supposed to work, anyway.

rkennke:
> Looking at the generated code, the assumption seems to be that we can find the klass bits in either the object header, or the object monitor DMW. However, it doesn't appear that the inflation protocol has been changed away from using the INFLATING value (aka all zeros). Then the klass pointer is seemingly just nuked, together with all other bits. That means the various fast path checks that check for the monitor bit are going to not find a monitor bit (0b10). Therefore it will just shift the value resulting in a null klass pointer. Will we just crash if someone racingly reads the klass() during inflation? I can't see how that is supposed to work, anyway.

The new lightweight locking (which is always used when enabling compact headers) does not use INFLATING. It can directly CAS the pointer to the monitor into the header.

coleenp:
You need this patch for JCK tests
diff --git a/src/hotspot/share/gc/z/zObjArrayAllocator.cpp b/src/hotspot/share/gc/z/zObjArrayAllocator.cpp
index 62a1d828859..f2cf07b7a7d 100644
--- a/src/hotspot/share/gc/z/zObjArrayAllocator.cpp
+++ b/src/hotspot/share/gc/z/zObjArrayAllocator.cpp
@@ -139,7 +139,11 @@ oop ZObjArrayAllocator::initialize(HeapWord* mem) const {
   ZThreadLocalData::clear_invisible_root(_thread);
 
   // Signal to the ZIterator that this is no longer an invisible root
-  oopDesc::release_set_mark(mem, markWord::prototype());
+  if (UseCompactObjectHeaders) {
+    oopDesc::release_set_mark(mem, _klass->prototype_header());
+  } else {
+    oopDesc::release_set_mark(mem, markWord::prototype());
+  }
 
   return cast_to_oop(mem);
 }

rkennke:
@fisk according to @coleenp's testing, ZGC is still failing some tests reliably with this JEP, and it looks like there's STS sync missing in ZRelocateTask::work() call tree. See reports, instructions and hs_err files here:

https://bugs.openjdk.org/browse/JDK-8305895

I tried something similar to what I proposed in #13904 but it seems to hang, just as you said it would. Could you help me to resolve this? In essence, the worker must not fetch the object-size (and thus the Klass*) without sync'ing with STS to ensure we don't grab dangling monitors, but also without deadlocking/hanging.

fisk:
> @fisk according to @coleenp's testing, ZGC is still failing some tests reliably with this JEP, and it looks like there's STS sync missing in ZRelocateTask::work() call tree. See reports, instructions and hs_err files here:
> 
> https://bugs.openjdk.org/browse/JDK-8305895
> 
> I tried something similar to what I proposed in #13904 but it seems to hang, just as you said it would. Could you help me to resolve this? In essence, the worker must not fetch the object-size (and thus the Klass*) without sync'ing with STS to ensure we don't grab dangling monitors, but also without deadlocking/hanging.

Sure, I'll have a look.

src/hotspot/cpu/aarch64/templateTable_aarch64.cpp
```diff
@@ -3568,12 +3568,25 @@ void TemplateTable::_new() {
...
+      __ sub(r3, r3, sizeof(oopDesc));
+    }
     __ cbz(r3, initialize_header);
 
     // Initialize object fields
     {
-      __ add(r2, r0, sizeof(oopDesc));
+      if (UseCompactObjectHeaders) {
+        __ add(r2, r0, oopDesc::base_offset_in_bytes());
+        if (!is_aligned(oopDesc::base_offset_in_bytes(), BytesPerLong)) {
```
sunny868, line 3582: I see `oopDesc::base_offset_in_bytes()` must be 8 bytes, so codes here `if(){}` may be useless, right?

src/hotspot/share/opto/memnode.cpp
```diff
@@ -1869,6 +1869,13 @@ Node *LoadNode::Ideal(PhaseGVN *phase, bool can_reshape) {
 const Type*
 LoadNode::load_array_final_field(const TypeKlassPtr *tkls,
                                  ciKlass* klass) const {
+  if (UseCompactObjectHeaders) {
+    if (tkls->offset() == in_bytes(Klass::prototype_header_offset())) {
+      // The field is Klass::_prototype_header.  Return its (constant) value.
+      assert(this->Opcode() == Op_LoadX, "must load a proper type from _prototype_header");
+      return TypeX::make(klass->prototype_header());
```
mmyxym, line 1876: Looks like directly getting the constant value is not specific to compact object headers, isn't it?

mmyxym:
Hi Roman,

I found a crash with G1. The stack trace is like:
#7  <signal handler called>
#8  Klass::layout_helper (this=0x0)
#9  oopDesc::size_given_klass (this=0xfff7d6899f00, klass=0x0)
#10 0x0000ffffb039dda8 in oopDesc::oop_iterate_size<G1RootRegionScanClosure>
#11 G1ConcurrentMark::scan_root_region
#12 G1CMRootRegionScanTask::work

The G1 root region scanning encountered a NULL klass pointer while iterating a object with NULL klass.  The crash is reproduced with a ratio of  ~1/1000 with vmTestbase/nsk/monitoring/stress/lowmem/lowmem023/TestDescription.java with UseCompactObjectHeaders in JDK11.

The root cause is cleanup in safepoint will deflate object monitor with :
ObjectSynchronizer::deflate_monitor
...
obj->release_set_mark(mid->header());           // set back mark to object
mid->clear();                                                    // set NULL to header in object monitor
...
while oopDesc::klass -> oopDesc::resolve_mark will first get the object monitor and then *maybe* read the null MARK. G1ConcurrentMark::scan_root_regions cannot join STS and the cleanup/deflating monitor in safepoint can happen simultaneously with scan root regions, then the crash happens. But we cannot try to join STS in scan_root_regions because the young gc has to wait for the finish of scan_root_regions. Joining STS in scan_root_regions will make the VM dead-locked.

My fix for now is to cancel the cleanup in safepoint while in scan_root_regions. The scan_root_regions is very short period of time so it should be ok and the crashed test case is fixed. You or others may provide a better approach.

--- a/src/hotspot/share/gc/g1/g1CollectedHeap.hpp
+++ b/src/hotspot/share/gc/g1/g1CollectedHeap.hpp

+  inline bool root_region_scanning();

--- a/src/hotspot/share/gc/g1/g1CollectedHeap.inline.hpp
+++ b/src/hotspot/share/gc/g1/g1CollectedHeap.inline.hpp

+inline bool G1CollectedHeap::root_region_scanning() {
+  return _cm->root_regions()->scan_in_progress();
+}

--- a/src/hotspot/share/runtime/safepoint.cpp
+++ b/src/hotspot/share/runtime/safepoint.cpp
+#if INCLUDE_G1GC
+#include "gc/g1/g1CollectedHeap.inline.hpp"
+#endif // INCLUDE_G1GC

@@ -732,6 +735,13 @@ void SafepointSynchronize::do_cleanup_tasks() {
 
   TraceTime timer("safepoint cleanup tasks", TRACETIME_LOG(Info, safepoint, cleanup));
 
+#if INCLUDE_G1GC
+  if (UseCompactObjectHeaders && UseG1GC && G1CollectedHeap::heap()->root_region_scanning()) {
+    log_debug(safepoint)("Safepoint cancels cleanup due to G1 root region scanning.");
+    return;
+  }
+#endif // INCLUDE_G1GC

mmyxym:
Hi Roman,

I ran into a crash of G1ConcurrentMark::scan_root_regions in JDK11 with compact header that the object iteration in root region scanning encountered an NULL mark from an object monitor because of monitor deflation which cleared the remote header on monitor while the concurrent mark thread still hold the monitor to get the mark header. There's a very special point in scan_root_regions that the concurrent mark thread cannot join STS so the race could happen. 

The problem should be very rare in tip or JDK17+ because the monitor deflation is moved into a concurrent thread and there're handshake and VM_RendezvousGCThreads to seperate deflation and monitor deletion. But theoratically the concurrent mark thread in scan_root_regions can still have a monitor which might be deleted at anytime.

rkennke:
> Hi Roman,
> 
> I ran into a crash of G1ConcurrentMark::scan_root_regions in JDK11 with compact header that the object iteration in root region scanning encountered an NULL mark from an object monitor because of monitor deflation which cleared the remote header on monitor while the concurrent mark thread still hold the monitor to get the mark header. There's a very special point in scan_root_regions that the concurrent mark thread cannot join STS so the race could happen.
> 
> The problem should be very rare in tip or JDK17+ because the monitor deflation is moved into a concurrent thread and there're handshake and VM_RendezvousGCThreads to seperate deflation and monitor deletion. But theoratically the concurrent mark thread in scan_root_regions can still have a monitor which might be deleted at anytime.

Hi Liang,

Thanks for the report! This seems to be a hard problem, indeed! On the one hand, scan_root_regions() would need to coordinate with the deflation thread, otherwise it could crash as you described, because oop_iterate() needs to access the Klass* and therefore the header, which might be displaced in a monitor. If that monitor gets deflated without coordination, the G1 thread could reach to a bad header and crash. That coordination is normally ensured by letting the concurrent GC thread join the STS. On the other hand, the concurrent GC thread must not join the STS until scan_root_regions() is finished, because of the reasons described in G1ConcurrentMarkThread::concurrent_mark_cycle_do(), otherwise a subsequent GC request could block scan_root_regions() from joining the STS.

I think that the best answer for now is to wait until external ObjectMonitor mapping arrives, which should be soon, AFAIK (@xmas92 is working on it and should know more). I am not sure if there is a good other solution to the problem that would work with monitor deflation (e.g. in Lilliput/JDK17 or 11), maybe @tschatzl has a good idea?

mmyxym:
> > Hi Roman,
> > I ran into a crash of G1ConcurrentMark::scan_root_regions in JDK11 with compact header that the object iteration in root region scanning encountered an NULL mark from an object monitor because of monitor deflation which cleared the remote header on monitor while the concurrent mark thread still hold the monitor to get the mark header. There's a very special point in scan_root_regions that the concurrent mark thread cannot join STS so the race could happen.
> > The problem should be very rare in tip or JDK17+ because the monitor deflation is moved into a concurrent thread and there're handshake and VM_RendezvousGCThreads to seperate deflation and monitor deletion. But theoratically the concurrent mark thread in scan_root_regions can still have a monitor which might be deleted at anytime.
> 
> Hi Liang,
> 
> Thanks for the report! This seems to be a hard problem, indeed! On the one hand, scan_root_regions() would need to coordinate with the deflation thread, otherwise it could crash as you described, because oop_iterate() needs to access the Klass* and therefore the header, which might be displaced in a monitor. If that monitor gets deflated without coordination, the G1 thread could reach to a bad header and crash. That coordination is normally ensured by letting the concurrent GC thread join the STS. On the other hand, the concurrent GC thread must not join the STS until scan_root_regions() is finished, because of the reasons described in G1ConcurrentMarkThread::concurrent_mark_cycle_do(), otherwise a subsequent GC request could block scan_root_regions() from joining the STS.
> 
> I think that the best answer for now is to wait until external ObjectMonitor mapping arrives, which should be soon, AFAIK (@xmas92 is working on it and should know more). I am not sure if there is a good other solution to the problem that would work with monitor deflation (e.g. in Lilliput/JDK17 or 11), maybe @tschatzl has a good idea?

So far I just skip the deflation while in scan_root_regions in 11u which is practically working but not solid theoretically. Maybe it would ok without side effects if deflation wait until scan_root_regions finish because the G1 young GC will wait for scan_root_regions in the same way.

rkennke:
Will re-open when ready in Lilliput

src/hotspot/share/gc/serial/markSweep.cpp
```diff
@@ -163,12 +163,14 @@ void MarkSweep::mark_object(oop obj) {
     _string_dedup_requests->add(obj);
   }
 
+  // Do the transform while we still have the header intact,
+  // which might include important class information.
+  ContinuationGCSupport::transform_stack_chunk(obj);
```
mmyxym, line 168: Not sure if position of  calling ContinuationGCSupport::transform_stack_chunk will have side effect. Had better guard with option...

PR #13966, opened 2023-05-12
8306593

plummercj:
nsk/jdi/stop/stop001 is problem listed due to [JDK-7034630](https://bugs.openjdk.org/browse/JDK-7034630), but that issue doesn't seem to reproduce anymore. It's suspected that it was only an issue on Solaris. For this reason stop001 should be removed from the problem list. However, it will fail due to lacking virtual thread support, so this issue also needs to be fixed at the same time. The issue is that for virtual threads, after work for [JDK-8306434](https://bugs.openjdk.org/browse/JDK-8306434) and [JDK-8306471](https://bugs.openjdk.org/browse/JDK-8306471) is complete, the test needs to expect IllegalThreadStateException rather than the stop() call succeeding.

### DIFF
--- a/test/hotspot/jtreg/vmTestbase/nsk/jdi/ThreadReference/stop/stop001.java
+++ b/test/hotspot/jtreg/vmTestbase/nsk/jdi/ThreadReference/stop/stop001.java
@@ -269,6 +269,7 @@ public class stop001 {
                 line = pipe.readln();
                 log2("       returned string is: " + line);
                 if (vthreadMode) {
+                    // Because the stop() called failed, stop001a.tObj should be "null".
                     if (!line.equals("null")) {
                         log3("ERROR: 'stop001a.tObj' is not 'null'");
                         expresult = returnCode1;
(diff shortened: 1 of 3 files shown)


lmesnik (approved):

test/hotspot/jtreg/vmTestbase/nsk/jdi/ThreadReference/stop/stop001.java
```diff
@@ -256,16 +268,24 @@ private int runThis (String argv[], PrintStream out) {
...
                 log2("       returned string is: " + line);
-                if (line.equals("null")) {
-                    log3("ERROR: 'stop001a.tObj = e1;' was not assigned");
-                    expresult = returnCode1;
-                } else if (line.equals("equal")) {
-                } else if (line.equals("NOT_equal")) {
-                    log3("ERROR: in the debugee, e1 is not 'LineUnavailableException'");
-                    expresult = returnCode1;
+                if (vthreadMode) {
+                    // Because the stop() called failed, stop001a.tObj should be "null".
```
sspitsyn, line 272: Nit: Is it a typo: `top() called failed` => `stop() call failed` ?

sspitsyn (approved):
Looks good.
Thanks,
Serguei

test/hotspot/jtreg/vmTestbase/nsk/jdi/ThreadReference/stop/stop001.java
plummercj: Fixed.

plummercj:
Thanks for the reviews Serguei and Leonid!

/integrate

PR #13967, opened 2023-05-12
8306467: Fix nsk/jdb/kill/kill001 to work with new JVMTI StopThread support for virtual threads.

plummercj:
Currently kill001 assumes that JVMTI StopThread (via JDI ThreadReference.stop) is not supported for virtual threads. [JDK-8306034](https://bugs.openjdk.org/browse/JDK-8306034) is adding support for StopThread on a virtual thread as long as it is suspended and mounted. This means, for example, it will work for virtual threads in the following conditions:
 - Debuggee in a loop and suspended
 - Debuggee blocked on a java monitor and suspended (because it is mounted when blocked)
 - Debuggee at a breakpoint and suspended

But will continue to not work in the following situations:
- Debuggee in a loop but not suspended
- Debuggee blocked on a java monitor and not suspended
- Debuggee suspended but unmounted, such as during a call the Thread.sleep()

kill001 suspends all threads when a breakpoint is hit in the main thread, and then does a "jdb kill" on each thread, which translate to `ThreadReference.stop()`. Si this is expected to work now since the additional threads are all blocked on a java monitor, and therefore mounted (and the breakpoint also suspended them).

Most of the changes involve undoing the virtual thread specific code that was added to the test as part of [JDK-8282385](https://bugs.openjdk.org/browse/JDK-8282385). However, there is an additional issue that also needs fixing. The test relies on the fact that the async exception is normally not caught, and that jdb normally stops when an uncaught exception is thrown. With virtual threads there ends up being an exception handler in `java.lang.VirtualThread.run()`, resulting in jdb not stopping when the async exception is thrown. This is fixed by having the test issue a jdb "catch all <classname>" command for each async exception type that the test throws.

---------
### Progress

### Issue
 * [JDK-8306467](https://bugs.openjdk.org/browse/JDK-8306467): Fix nsk/jdb/kill/kill001 to work with new JVMTI StopThread support for virtual threads.

### Reviewers
 * [Alex Menkov](https://openjdk.org/census#amenkov) (@alexmenkov - **Reviewer**) ⚠️ Review applies to [c2e0f7b1](https://git.openjdk.org/jdk/pull/13967/files/c2e0f7b13fcb163e80980fcf63940d5c2a2e1bc7)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13967/head:pull/13967` \
`$ git checkout pull/13967`

`$ git checkout pull/13967` \
`$ git pull https://git.openjdk.org/jdk.git pull/13967/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13967`

`$ git pr show -t 13967`

</details>

<a href="https://git.openjdk.org/jdk/pull/13967.diff">https://git.openjdk.org/jdk/pull/13967.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13967#issuecomment-1546300382)

### DIFF
--- a/test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001.java
+++ b/test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001.java
@@ -88,17 +88,13 @@ public class kill001 extends JdbTest {
     static final String LAST_BREAK      = DEBUGGEE_CLASS + ".breakHere";
     static final String MYTHREAD        = "MyThread";
     static final String DEBUGGEE_THREAD = PACKAGE_NAME + "." + MYTHREAD;
-    static final String DEBUGGEE_RESULT = DEBUGGEE_CLASS + ".notKilled";
+    static final String DEBUGGEE_RESULT = DEBUGGEE_CLASS + ".killed";
     static final String DEBUGGEE_EXCEPTIONS = DEBUGGEE_CLASS + ".exceptions";
 
     static int numThreads = nsk.jdb.kill.kill001.kill001a.numThreads;
 
     protected void runCases() {
         String[] reply;
-        Paragrep grep;
-        int count;
-        Vector v;
-        String found;
         String[] threads;
 
         jdb.setBreakpointInMethod(LAST_BREAK);
@@ -147,11 +143,11 @@ public class kill001 extends JdbTest {
 
         reply = jdb.receiveReplyForWithMessageWait(JdbCommand.eval + DEBUGGEE_RESULT,
                                                    DEBUGGEE_RESULT + " =");
-        grep = new Paragrep(reply);
-        found = grep.findFirst(DEBUGGEE_RESULT + " =" );
-        if (found.length() > 0) {
-            if (found.indexOf(DEBUGGEE_RESULT + " = 0") < 0) {
-                log.complain("Not all " + MYTHREAD + "s were killed. " + found + " remaining");
+        Paragrep grep = new Paragrep(reply);
+        String killed = grep.findFirst(DEBUGGEE_RESULT + " =" );
+        if (killed.length() > 0) {
+            if (killed.indexOf(DEBUGGEE_RESULT + " = " + numThreads) < 0) {
+                log.complain("Only " + killed + "out of " + numThreads + MYTHREAD + "s were killed.");
                 success = false;
             }
         } else {
--- a/test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001.java
+++ b/test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001.java
@@ -39,7 +39,7 @@
  * to the number of additional threads created. The value of
  * the "killed" variable is checked by "eval <expr>" command.
  * The test passes if the value is equal to the number of
- * additional thread created and fails otherwise.
+ * additional threads created and fails otherwise.
  * COMMENTS
  *  Modified due to fix of the test bug:
  *  4940902 TEST_BUG: race in nsk/jdb/kill/kill001
--- a/test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001.java
+++ b/test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001.java
@@ -38,7 +38,7 @@
  * the value of the special "killed" variable should to the set
  * to the number of additional threads created. The value of
  * the "killed" variable is checked by "eval <expr>" command.
- * The test passes if the value is equal to the nunber of
+ * The test passes if the value is equal to the number of
  * additional thread created and fails otherwise.
  * COMMENTS
  *  Modified due to fix of the test bug:
--- a/test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001.java
+++ b/test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001.java
@@ -149,7 +149,7 @@ public class kill001 extends JdbTest {
         String killed = grep.findFirst(DEBUGGEE_RESULT + " =" );
         if (killed.length() > 0) {
             if (killed.indexOf(DEBUGGEE_RESULT + " = " + numThreads) < 0) {
-                log.complain("Only " + killed + "out of " + numThreads + MYTHREAD + "s were killed.");
+                log.complain("Only " + killed + " out of " + numThreads + " " + MYTHREAD + "s were killed.");
                 success = false;
             }
         } else {
--- a/test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001.java
+++ b/test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001.java
@@ -35,9 +35,11 @@
  * suspends debuggee at moment when additional threads try to obtain
  * lock on synchronized object (previously locked in main thread)
  * and then tries to kill them. If these threads are killed then
- * a value of the special "notKilled" variable should not be modified.
- * Value of "notKilled" variable is checked by "eval <expr>" command.
- * The test passes if the value is equal to 0 and fails otherwise..
+ * the value of the special "killed" variable should to the set
+ * to the number of additional threads created. The value of
+ * the "killed" variable is checked by "eval <expr>" command.
+ * The test passes if the value is equal to the nunber of
+ * additional thread created and fails otherwise.
  * COMMENTS
  *  Modified due to fix of the test bug:
  *  4940902 TEST_BUG: race in nsk/jdb/kill/kill001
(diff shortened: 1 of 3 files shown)


test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001a.java
```diff
@@ -159,18 +157,27 @@ public void run() {
         synchronized (kill001a.lock) {}
 
         // Sleep during waitTime to give debugger a chance to kill debugee's thread.
-        // Note vthreads need a short sleep because they will never receive the kill,
-        // and therefore sleep the full time, resulting in a test timeout if too long.
         try {
-            Thread.currentThread().sleep(kill001a.vthreadMode ? 10000 : kill001a.waitTime);
```
alexmenkov, line 165: The PR description says:
`kill001 suspends all threads at a breakpoint and then does a "jdb kill" on each thread, which translate to ThreadReference.stop(), so this is expected to work now.`
As far as I see only main thread is suspended at "breakHere", all test threads wait in Thread.sleep().
I'd expect test threads to be unmounted in Thread.sleep(), so it's unclear how the test works in vthread mode

test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001a.java
```diff
@@ -122,34 +114,40 @@ public int runIt(String args[], PrintStream out) {
...
     }
 
     public void run() {
+        boolean killed = false;
+
         // Concatenate strings in advance to avoid lambda calculations later
-        String ThreadFinished = "WARNING: Thread finished: " + this.name;
+        String ThreadFinished = "Thread finished: " + this.name;
         String ThreadInterrupted = "WARNING: Thread was interrupted while waiting for killing: " + this.name;
+        String CaughtExpected ="Thread " + this.name + " caught expected async exception: " + expectedException;
```
sspitsyn, line 148: Nit: Space is missed after '=' sigh.

sspitsyn (approved):
Looks good.
Thanks,
Serguei

test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001a.java
plummercj: You're right. I misunderstood what the test is doing. At the time of the `kill` commands, the threads are actually blocked here:

        synchronized (kill001a.lock) {}

Currently blocking in this manner pins the virtual thread to the carrier thread, so it is mounted. However, we are not yet in the try block at the time the async exception is thrown. After all the threads have had the async exception thrown at them, the main thread is resumed and the sychronized block is exited. The log shows that the exception is thrown on the sleep() line, which is the first statement after the synchronized, which just so happens to be within the try and the right thing happens.

However, even if I add another statement outside the synchronized and just before the try (which ends up being the line where the exception is thrown), it still passes. This is due to yet another bug. The exception is blowing us out of the thread, so we never hit the following block:

```
        if (!killed) {
            // Need to make sure the increment is atomic
            synchronized (kill001a.lock) {
                kill001a.notKilled++;
            }
        }
```

The debugger side checks for the proper notKilled value (it should be 0). So the test still passes even though it never confirmed that the correct async exception was thrown and caught. It just knows that the thread did not exit cleanly. Probably instead of tracking `notKilled`, the test should track `killed` and only increment when the correct exception is actually caught.

And one more thing, the following comment before the sleep() wrong:

        // Sleep during waitTime to give debugger a chance to kill debugee's thread.

I'm not so sure the sleep is even necessary since we block in the synchronized indefinitely until the async exception is thrown. I think just a simple statement at the start of the try block is good enough. Maybe a println(). But this is somewhat fragile because we need to make sure there is no code between the synchronized and the try, so we probably need to move the synchronized to within the try also.

alexmenkov: > The debugger side checks for the proper notKilled value (it should be 0). So the test still passes even though it never confirmed that the correct async exception was thrown and caught. It just knows that the thread did not exit cleanly. Probably instead of tracking `notKilled`, the test should track `killed` and only increment when the correct exception is actually caught.

I think this will make the test more reliable.

> I'm not so sure the sleep is even necessary since we block in the synchronized indefinitely until the async exception is thrown. I think just a simple statement at the start of the try block is good enough. Maybe a println(). But this is somewhat fragile because we need to make sure there is no code between the synchronized and the try, so we probably need to move the synchronized to within the try also.

looks like we just need to move synchronized (or some other code to force vthread to be mounted) to inside the try and remove Thread.sleep, no other statements are needed.

plummercj: > looks like we just need to move synchronized (or some other code to force vthread to be mounted) to inside the try and remove Thread.sleep, no other statements are needed.

That should work. At first I was unsure because the async exception is not actually thrown until executing the next statement after the synchronized. But there should be a branch at the end of the try block, and the exception should be thrown before it is executed. Hopefully the exception table covers that bytecode as being in the try block.

plummercj: This became a lot trickier to understand than I expected. The JVM does not have to throw the async exception at the next executed bytecode. The interpreter seems to do it at an invoke or after a goto (so the target of the goto is the instruction that will appear to have thrown the async exception). I think JIT'd code is similar, although maybe throws on a backwards branch and not after a forward branch. For this reason we need some sort of statement in place after the synchronized block to trigger the throwing of the exception before we leave the try block.

Another thing that messes up the test is that within the synchronized block, you can't have anything that will trigger the throw of the async exception. For example:

        try {
            synchronized (kill001a.lock) {
                kill001a.log.display("entered synchronized");
            }
            kill001a.log.display("exited synchronized");
        } catch (Throwable t) {

This messes up the test because now the exception will be thrown from within the synchronized block. This triggers a catch of the exception by an implicit catch clauses whose only purpose is to do the monitorexit and then rethrow the exception (which the explicit catch clause will then catch). The problem is that the test only expects one throw per thread, and now it gets 2. Although fixable on the debugger side, it is probably best to avoid.

I've pushed an update. Let me know if you think it is ok.

alexmenkov (approved):

test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001.java
```diff
@@ -35,9 +35,11 @@
  * suspends debuggee at moment when additional threads try to obtain
  * lock on synchronized object (previously locked in main thread)
  * and then tries to kill them. If these threads are killed then
- * a value of the special "notKilled" variable should not be modified.
- * Value of "notKilled" variable is checked by "eval <expr>" command.
- * The test passes if the value is equal to 0 and fails otherwise..
+ * the value of the special "killed" variable should to the set
+ * to the number of additional threads created. The value of
+ * the "killed" variable is checked by "eval <expr>" command.
+ * The test passes if the value is equal to the nunber of
```
alexmenkov, line 41: nunber -> number

test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001a.java
plummercj: I went to edit the PR description, but found it is not actually incorrect. It says "kill001 suspends all threads at a breakpoint". That is actually true. Although it is the main thread that hits the breakpoint, all threads are suspended. Since the additional threads are suspended and mounted, the TR.stop() can be done on them. If they are blocked (as they are) but not suspended, the TR.stop() would fail (for virtual threads). I'll try to clean it up a bit though. The description could be clearer.

test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001.java
```diff
@@ -35,9 +35,11 @@
  * suspends debuggee at moment when additional threads try to obtain
  * lock on synchronized object (previously locked in main thread)
  * and then tries to kill them. If these threads are killed then
- * a value of the special "notKilled" variable should not be modified.
- * Value of "notKilled" variable is checked by "eval <expr>" command.
- * The test passes if the value is equal to 0 and fails otherwise..
+ * the value of the special "killed" variable should to the set
+ * to the number of additional threads created. The value of
```
sspitsyn, line 39: `If these threads are killed then the value of the special "killed" variable should to the set to the number of additional threads created.`

Typo? : It feels like there is a typo in the fragment: `variable should to the set to the number`.

plummercj: Yes, it should be "should be set to the number"

test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001a.java
alexmenkov: Yes, I realized this too. By default breakpoints have SUSPEND_ALL policy.

sspitsyn (approved):

PR #13974, opened 2023-05-15
8307795: AArch64: Optimize VectorMask.truecount() on Neon

changpeng1997:
In Vector API Java level, vector mask is represented as a boolean array
with 0x00/0x01 (8 bits of each element) as values, aka in-memory format.
When it is loaded into vector register, e.g. Neon, the in-memory format
will be converted to in-register format with 0/-1 value for each lane
(lane width aligned to its type) by VectorLoadMask [1] operation, and
convert back to in-memory format by VectorStoreMask[2]. In Neon, a
typical VectorStoreMask operation will first narrow given vector
registers by xtn insn [3] into byte element type, and then do a vector
negate to convert to 0x00/0x01 value for each element.

For most of the vector mask operations, the input mask is in-register
format. And a vector mask also works in-register format all through
the compilation. But for some operations like VectorMask.trueCount()[4]
which counts the elements of true value, the expected input mask is
in-memory format. So a VectorStoreMask is generated to convert the
mask from in-register format to in-memory format before those operations.

However, for trueCount() these xtn instructions in VectorStoreMask can
be saved, since the narrowing operations will not influence the number
of active lane (value of 0x01) of its input.

This patch adds an optimized rule
`VectorMaskTrueCount (VectorStoreMask mask)` to save the unnecessary
narrowing operations.

For example,

```
var m = VectorMask.fromArray(IntVector.SPECIES_PREFERRED, ba, 0);
m.not().trueCount();
```

will produce following assembly on a Neon machine before this patch:

```
...
mvn     v16.16b, v16.16b           // VectorMask.not()
xtn     v16.4h, v16.4s
xtn     v16.8b, v16.8h
neg     v16.8b, v16.8b             // VectorStoreMask
addv    b17, v16.8b
umov    w0, v17.b[0]               // VectorMask.trueCount()
...
```

After this patch:

```
...
mvn     v16.16b, v16.16b           // VectorMask.not()
addv    s17, v16.4s
smov    x0, v17.b[0]
neg     x0, x0                     // Optimized VectorMask.trueCount()
...
```

In this case, we can save two xtn insns.

Performance:

Benchmark     Before               After                          Unit
testInt           723.822 ± 1.029  1182.375 ± 12.363    ops/ms
testLong       632.154 ± 0.197  1382.74 ± 2.188        ops/ms
testShort      788.665 ± 1.852   1152.38 ± 3.77         ops/ms

[1]: https://github.com/openjdk/jdk/blob/e1e758a7b43c29840296d337bd2f0213ab0ca3c9/src/hotspot/cpu/aarch64/aarch64_vector.ad#L4740
[2]: https://github.com/openjdk/jdk/blob/f968da97a5a5c68c28ad29d13fdfbe3a4adf5ef7/src/hotspot/cpu/aarch64/aarch64_vector.ad#L4841
[3]: https://developer.arm.com/documentation/dui0801/h/A64-SIMD-Vector-Instructions/XTN--XTN2--vector-
[4]: https://docs.oracle.com/en/java/javase/16/docs/api/jdk.incubator.vector/jdk/incubator/vector/VectorMask.html#trueCount()

### DIFF
--- a/src/hotspot/cpu/aarch64/aarch64_vector_ad.m4
+++ b/src/hotspot/cpu/aarch64/aarch64_vector_ad.m4
@@ -3793,6 +3793,31 @@ instruct vmask_truecount_sve(iRegINoSp dst, pReg src) %{
   ins_pipe(pipe_slow);
 %}
 
+// Combined rule for VectorStoreMask + VectorMaskTrueCount when the vector element type is not T_BYTE.
+
+instruct vstoremask_truecount(iRegINoSp dst, vReg src, immI_gt_1 size, vReg vtmp) %{
+  predicate(UseSVE == 0);
+  match(Set dst (VectorMaskTrueCount (VectorStoreMask src size)));
+  effect(TEMP vtmp);
+  format %{ "vstoremask_truecount $dst, $src\t# KILL $vtmp" %}
+  ins_encode %{
+    // Input "src" is a vector mask represented as lanes with
+    // 0/-1 as element values.
+    uint esize = (uint)$size$$constant;
+    if (esize == 8) {
+      __ addpd($vtmp$$FloatRegister, $src$$FloatRegister);
+    } else {
+      uint length_in_bytes = Matcher::vector_length_in_bytes(this, $src);
+      Assembler::SIMD_Arrangement arrangement = Assembler::esize2arrangement(esize,
+                                                                             /* isQ */ length_in_bytes == 16);
+      __ addv($vtmp$$FloatRegister, arrangement, $src$$FloatRegister);
+    }
+    __ smov($dst$$Register, $vtmp$$FloatRegister, __ B, 0);
+    __ neg($dst$$Register, $dst$$Register);
+  %}
+  ins_pipe(pipe_slow);
+%}
+
 // first true
 
 instruct vmask_firsttrue_lt8e(iRegINoSp dst, vReg src, rFlagsReg cr) %{
--- a/test/micro/org/openjdk/bench/jdk/incubator/vector/StoreMaskTrueCount.java
+++ b/test/micro/org/openjdk/bench/jdk/incubator/vector/StoreMaskTrueCount.java
@@ -1,0 +1,84 @@
+//
+// Copyright (c) 2023, Arm Limited. All rights reserved.
+// DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+//
+// This code is free software; you can redistribute it and/or modify it
+// under the terms of the GNU General Public License version 2 only, as
+// published by the Free Software Foundation.
+//
+// This code is distributed in the hope that it will be useful, but WITHOUT
+// ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// version 2 for more details (a copy is included in the LICENSE file that
+// accompanied this code).
+//
+// You should have received a copy of the GNU General Public License version
+// 2 along with this work; if not, write to the Free Software Foundation,
+// Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+//
+// Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+// or visit www.oracle.com if you need additional information or have any
+// questions.
+//
+//
+package org.openjdk.bench.jdk.incubator.vector;
+
+import java.util.concurrent.TimeUnit;
+import java.util.Random;
+import jdk.incubator.vector.*;
+import org.openjdk.jmh.annotations.*;
+
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@State(Scope.Thread)
+@Fork(jvmArgsPrepend = {"--add-modules=jdk.incubator.vector"})
+public class StoreMaskTrueCount {
+    private static final VectorSpecies<Short> S_SPECIES = ShortVector.SPECIES_PREFERRED;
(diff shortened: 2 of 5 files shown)


theRealAph (approved):
That makes sense. Is it likely that there are more of these combined operations on vector masks that could be matched? if so, it might make sense to do the job earlier, in the C2 optimizer.

test/micro/org/openjdk/bench/jdk/incubator/vector/StoreMaskTrueCount.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Benchmark
+    public static int testLong() {
+        VectorMask<Long> m = VectorMask.fromArray(L_SPECIES, ba, 0);
+        int res = 0;
+        for (int i = 0; i < LENGTH; i++) {
+            m = m.not();
+            res += m.trueCount();
+        }
```
theRealAph, line 80: This looks like it might be removed by loop opts. I think you might need a blackhole somewhere.

changpeng1997:
> That makes sense. Is it likely that there are more of these combined operations on vector masks that could be matched? if so, it might make sense to do the job earlier, in the C2 optimizer.

Thanks for your review.

I have tried to optimize ```VectorMask.firstTrue()``` [1] and ```VectorMask.lastTrue()``` [2] in the same way as this patch, but these two operations are strong correlated with xtn, we cannot simply remove it.

I didn't find a way to optimize these two operations, they are hignly optimized in C2 backend and there are not extra instructions.

[1]: https://github.com/openjdk/jdk/blob/8d49ba9e8d3095f850b3007b56488a0c0cf8ddff/src/hotspot/cpu/aarch64/aarch64_vector.ad#L5517
[2]: https://github.com/openjdk/jdk/blob/8d49ba9e8d3095f850b3007b56488a0c0cf8ddff/src/hotspot/cpu/aarch64/aarch64_vector.ad#L5624

test/micro/org/openjdk/bench/jdk/incubator/vector/StoreMaskTrueCount.java
changpeng1997: > This looks like it might be removed by loop opts. I think you might need a blackhole somewhere.

```m``` will be updated in every iteration of this loop, so ```m``` is not a loop-invariants actually.
I can see the assembly code of this loop by using JMH perfasm.

Comment (theRealAph) [reply test/micro/org/openjdk/bench/jdk/incubator/vector/StoreMaskTrueCount.java:80]: > > This looks like it might be removed by loop opts. I think you might need a blackhole somewhere.
> 
> `m` will be updated in every iteration of this loop, so `m` is not a loop-invariants actually. I can see the assembly code of this loop by using JMH perfasm.

Isn't it? Looks to me like all it does is flip `m` each time. Whether or not this code is optimized today isn't relevant.

So it's the same as

```
        for (int i = 0; i < LENGTH/2; i++) {
            res += m.trueCount();
        }
        m = m.not();
        for (int i = 0; i < LENGTH/2; i++) {
            res += m.trueCount();
        } 
```

... which is trivially optimizable, no?

Comment (changpeng1997) [reply test/micro/org/openjdk/bench/jdk/incubator/vector/StoreMaskTrueCount.java:77]: > > > This looks like it might be removed by loop opts. I think you might need a blackhole somewhere.
> > 
> > 
> > `m` will be updated in every iteration of this loop, so `m` is not a loop-invariants actually. I can see the assembly code of this loop by using JMH perfasm.
> 
> Isn't it? Looks to me like all it does is flip `m` each time. Whether or not this code is optimized today isn't relevant.
> 
> So it's the same as
> 
> ```
>         for (int i = 0; i < LENGTH/2; i++) {
>             res += m.trueCount();
>         }
>         m = m.not();
>         for (int i = 0; i < LENGTH/2; i++) {
>             res += m.trueCount();
>         } 
> ```
> 
> ... which is trivially optimizable, no?

Sorry for the delay.

Yes, actually they do the same thing, though current C2 compiler cannot do such optimization so far. Anyway, I have updated this benchmark to avoid potential optimization and ensure that we can measure performance effectively.

**Review by e1iu — APPROVED** [review]
LGTM.

File: src/hotspot/cpu/aarch64/aarch64_vector_ad.m4 [hunk 2]
```diff
@@ -3822,6 +3822,30 @@ instruct vmask_truecount_sve(iRegINoSp dst, pReg src) %{
   ins_pipe(pipe_slow);
 %}
 
+// Combined rule for VectorStoreMask + VectorMaskTrueCount when the vector element type is not T_BYTE.
```
theRealAph, line 3825:
```suggestion
// Combined rule for VectorMaskTrueCount (VectorStoreMask) when the vector element type is not T_BYTE.
```
Using `+` is unnecessarily confusing, as is swapping the operands.

theRealAph (approved):

e1iu (approved):
Still looks good to me.

TobiHartmann:
This change broke the builds: [JDK-8309110](https://bugs.openjdk.org/browse/JDK-8309110).

dholmes-ora:
What testing was done on this fix before integration? I don't even see Git Hub Actions being run.

changpeng1997:
> What testing was done on this fix before integration? I don't even see Git Hub Actions being run.

@dholmes-ora 
I did see earlier that Github Action ran (In the 'Checks' tab) and finished, and I believed the Windows failure is not related to my patch. Perhaps GHA does not cover the jmh build. 

I ran the full jtreg tests in our internal ci with my first draft patch, but I realized that I forgot to have another run of full jtreg test after updating my patch to the current version. I only ran my new jtreg test before integration. Lesson learnt!

Thanks @DamonFool for the quick fix. I am working on JDK-8309129.

PR #13976, opened 2023-05-15
8308076: X86_64: make rheapbase register allocatable in zero based compressedOops mode

kuaiwei:
In x86 64 mode, decode heap oop could use SIB without base if heap base is zero. like
```asm
0d1     movl    R11, [,R9 << 3 + #72] (zero base compressed oop addressing) # compressed ptr ! Field: java/lang/ClassLoader.classAssertionStatus 
```
So rheapbase( r12 )  can be allocated as general register.

Tier 1/2 tests are passed without new failure.

---------
### Progress

### Integration blocker
&nbsp;⚠️ Title mismatch between PR and JBS for issue [JDK-8308076](https://bugs.openjdk.org/browse/JDK-8308076)

### Issue
 * [JDK-8308076](https://bugs.openjdk.org/browse/JDK-8308076): make rheapbase register allocatable in zero based compressedOops mode for x86_64 (**Enhancement** - P4) ⚠️ Title mismatch between PR and JBS.

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13976/head:pull/13976` \
`$ git checkout pull/13976`

`$ git checkout pull/13976` \
`$ git pull https://git.openjdk.org/jdk.git pull/13976/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13976`

`$ git pr show -t 13976`

</details>

<a href="https://git.openjdk.org/jdk/pull/13976.diff">https://git.openjdk.org/jdk/pull/13976.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13976#issuecomment-1547275191)

### DIFF
--- a/src/hotspot/cpu/x86/x86_64.ad
+++ b/src/hotspot/cpu/x86/x86_64.ad
@@ -379,7 +379,8 @@ RegMask _STACK_OR_LONG_REG_mask;
 RegMask _STACK_OR_INT_REG_mask;
 
 static bool need_r12_heapbase() {
-  return UseCompressedOops;
+  // return UseCompressedOops && (PreserveHeapbaseReg || CompressedOops::ptrs_base() != nullptr);
+  return CompressedOops::need_heapbase_reg();
 }
 
 void reg_mask_init() {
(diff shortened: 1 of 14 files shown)


merykitty:
Please see [JDK-8221249](https://bugs.openjdk.org/browse/JDK-8221249). A possible further improvement you can try is to allocate a vector register to be a dedicated zero register, this also helps other operations such as clear memory of newly created objects.

Thanks.

kuaiwei:
> Please see [JDK-8221249](https://bugs.openjdk.org/browse/JDK-8221249). A possible further improvement you can try is to allocate a vector register to be a dedicated zero register, this also helps other operations such as clear memory of newly created objects.
> 
> Thanks.

Thanks for your suggestion. Before my work, I searched previous discussion but not find this one.

It looks no performance benefit in previous tests. But I find some performance gain in renaissance benchmark. And I also noticed some regression, in general, it looks a better score. I'm still running it multiple times to get a stable result. , I think the regression may has 2 causes
1 JVM can not use r12 as immediate 0
2 new heap decode instruction must has 32bit offset, it will enlarge the code size and increase icache pressure.

It's the tradeoff. It may dependents on app behavior. So I added a new option PreserveHeapbaseReg to switch.

src/hotspot/share/oops/compressedOops.hpp
```diff
@@ -97,6 +98,12 @@ class CompressedOops : public AllStatic {
   static address  ptrs_base_addr()           { return (address)&_narrow_oop._base; }
   static address  ptrs_base()                { return _narrow_oop._base; }
 
+#if defined(X86) && !defined(ZERO)
```
merykitty, line 101: This should not be leaked into shared code

src/hotspot/cpu/x86/x86_64.ad
```diff
@@ -635,68 +635,8 @@ void encode_RegMem(CodeBuffer &cbuf,
...
-        if (disp_reloc != relocInfo::none) {
-          emit_d32_reloc(cbuf, disp, relocInfo::oop_type, RELOC_DISP32);
-        } else {
-          emit_d32(cbuf, disp);
-        }
-      }
-    }
-  }
+  MacroAssembler masm(&cbuf);
+  masm.emit_regmem(reg, base, index, (Address::ScaleFactor)scale, disp, RelocationHolder::none);
```
merykitty, line 639: This change is not really related, right? I think a separate change to cleanup these would be more preferrable

merykitty (commented):
Thanks for your explanations, my proposal can alleviate the usage of immediate 0 but will not solve the increased code size from base-less addresses. I have left some small comments.

Also, there is a `reinit_heapbase` in `MacroAssembler` that you may consider modifying as well.

src/hotspot/cpu/x86/x86_64.ad
kuaiwei: The assembler function in adfile can not handle the case SIB without base reg. And emit_operand_helper in assembler_x86.cpp can cover more cases do more check ,like cpu type check in emit_compressed_disp_byte. So I think it's better to reuse the function in assembler_x86.cpp

src/hotspot/share/oops/compressedOops.hpp
kuaiwei: Fixed.

kuaiwei:
> 

Thanks for the comments. reinit_heapbase is fixed.

src/hotspot/cpu/x86/x86_64.ad
```diff
@@ -379,7 +379,7 @@ RegMask _STACK_OR_LONG_REG_mask;
 RegMask _STACK_OR_INT_REG_mask;
 
 static bool need_r12_heapbase() {
-  return UseCompressedOops;
+  return CompressedOops::need_heapbase_reg();
```
merykitty, line 382: Why not just use `PreserveHeapbaseReg`, this is a machine-dependent flag so I don't find it appealing to introduce a machine-independent check.

theRealAph:
> It's the tradeoff. It may dependents on app behavior. So I added a new option PreserveHeapbaseReg to switch.

This looks like [JDK-8221249](https://bugs.openjdk.org/browse/JDK-8221249). Rather than starting a new patch, you should try to reopen that one.

PR #13983, opened 2023-05-15
8308091: Remove unused iRegIHeapbase() matching operand

zhengxiaolinX:
The `iRegIHeapbase()` matching operand has no usage on both AArch64 and RISC-V platforms after [JDK-8242449](https://bugs.openjdk.org/browse/JDK-8242449) and [JDK-8306667](https://bugs.openjdk.org/browse/JDK-8306667), respectively. As the following-up action discussed in the code review process of [JDK-8306667](https://bugs.openjdk.org/browse/JDK-8306667), this is a small cleanup for the `iRegIHeapbase()` matching operand.

Passed fastdebug/release build on both AArch64/RISC-V platforms.

### DIFF
--- a/src/hotspot/cpu/aarch64/aarch64.ad
+++ b/src/hotspot/cpu/aarch64/aarch64.ad
@@ -5320,17 +5320,6 @@ operand iRegNNoSp()
   interface(REG_INTER);
 %}
 
-// heap base register -- used for encoding immN0
-
-operand iRegIHeapbase()
-%{
-  constraint(ALLOC_IN_RC(heapbase_reg));
-  match(RegI);
-  op_cost(0);
-  format %{ %}
-  interface(REG_INTER);
-%}
-
 // Float Register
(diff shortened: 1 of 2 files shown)


RealFYang (commented):
Maybe we should remove `reg_class heapbase_reg` at the same time?

zhengxiaolinX:
> Maybe we should remove `reg_class heapbase_reg` at the same time?

Thanks, that's reasonable. Changes passed AArch64 hotspot tier1\~4 (fastdebug). RISC-V hotspot tier1 is running.

RealFYang (approved):
Updated change LGTM. Thanks.

TobiHartmann (approved):
Looks good to me too.

zhengxiaolinX:
Thanks for reviewing! @RealFYang @TobiHartmann 

This is a trivial patch, so integrate then.

/integrate

PR #13990, opened 2023-05-15
JDK-8292157: Incorrect error: "block element not allowed within inline element <a>"

hns:
Please review a change to downgrade from error to warning if the label of a `<a>` element or reference tag contains an HTML block element. Before HTML 5, elements were clearly divided into block and inline elements, and the latter were not allowed to contain the former. 

With HTML5, the division between block and inline elements is delegated to CSS, and although elements have default styles, styles can be freely customized by the author. This applies especially true for the `<a>` element which is commonly used both with inline and block layout. That was the rationale for downgrading from error to warning for the `<a>` element only. See the JBS issue comments for details and references.

The error message for the warning was enhanced to refer to the default style for the `<a>` element.

---------
### Progress

### Issue
 * [JDK-8292157](https://bugs.openjdk.org/browse/JDK-8292157): Incorrect error: "block element not allowed within inline element &lt;a&gt;" (**Bug** - `"4"`)

### Reviewers
 * [Jonathan Gibbons](https://openjdk.org/census#jjg) (@jonathan-gibbons - **Reviewer**) ⚠️ Review applies to [bb6228bc](https://git.openjdk.org/jdk/pull/13990/files/bb6228bc4dca46362f5ea0ca4b2dd81db130723c)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13990/head:pull/13990` \
`$ git checkout pull/13990`

`$ git checkout pull/13990` \
`$ git pull https://git.openjdk.org/jdk.git pull/13990/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13990`

`$ git pr show -t 13990`

</details>

<a href="https://git.openjdk.org/jdk/pull/13990.diff">https://git.openjdk.org/jdk/pull/13990.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13990#issuecomment-1548015568)


src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclint/resources/doclint.properties
```diff
@@ -77,8 +77,8 @@ dc.tag.nested.tag=nested tag: {0}
 dc.tag.not.allowed.here = tag not allowed here: <{0}>
 dc.tag.not.allowed = element not allowed in documentation comments: <{0}>
 dc.tag.not.allowed.inline.element = block element not allowed within inline element <{1}>: {0}
-dc.tag.not.allowed.inline.tag = block element not allowed within @{1}: {0}
-dc.tag.not.allowed.inline.other = block element not allowed here: {0}
+dc.tag.not.allowed.element.default.style = block element not allowed within element <{1}> with default style: {0}
```
jonathan-gibbons, line 80: Is it possible to improve the wording?

As written, it is ambiguous/confusing: it seems like `{0}` is the default style, which I don't think is what you intend.
Also, inconsistent pointy brackets `<{1}>` but just `{0}`

jonathan-gibbons: That looks good

Optionally, we could file a followup RFE to try and make more resources more stylistically consistent

jonathan-gibbons (approved):

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclint/resources/doclint.properties
hns: I changed the messages to the new format.

PR #8906, opened 2022-05-26
8271406: Kitchensink failed with "assert(early->flag() == current->flag()) failed: Should be the same"

zhengyu123:
There is a narrow race window if nmt baseline is executed between reserving a virtual region and setting it nmt type, that can result the assertion failure.

Thanks @coleenp for testing the patch, as it is an Oracle internal test.

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.java.net/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8271406](https://bugs.openjdk.java.net/browse/JDK-8271406): Kitchensink failed with "assert(early->flag() == current->flag()) failed: Should be the same"

### Reviewers
 * [Coleen Phillimore](https://openjdk.java.net/census#coleenp) (@coleenp - **Reviewer**)
 * [David Holmes](https://openjdk.java.net/census#dholmes) (@dholmes-ora - **Reviewer**)
 * [Thomas Stuefe](https://openjdk.java.net/census#stuefe) (@tstuefe - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk pull/8906/head:pull/8906` \
`$ git checkout pull/8906`

Update a local copy of the PR: \
`$ git checkout pull/8906` \
`$ git pull https://git.openjdk.java.net/jdk pull/8906/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 8906`

View PR using the GUI difftool: \
`$ git pr show -t 8906`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.java.net/jdk/pull/8906.diff">https://git.openjdk.java.net/jdk/pull/8906.diff</a>

</details>

PR #13993, opened 2023-05-15
8306428: RunThese30M.java crashed with assert(early->flag() == current->flag() || early->flag() == mtNone)

gerard-ziemski:
We print the stack of the allocation site to help us figure out which flag is wrong, and help us debug that part of the code.

This fix is in the same spirit as https://github.com/openjdk/jdk/pull/8906 where we are going to add more debug info, in hopes that it will help analyze it successfully.

Tested locally with jcmd PID VM.native_memory baseline/detail.diff to make sure we are not messing anything up.

---------
### Progress

### Issue
 * [JDK-8306428](https://bugs.openjdk.org/browse/JDK-8306428): RunThese30M.java crashed with assert(early-&gt;flag() == current-&gt;flag() || early-&gt;flag() == mtNone)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13993/head:pull/13993` \
`$ git checkout pull/13993`

`$ git checkout pull/13993` \
`$ git pull https://git.openjdk.org/jdk.git pull/13993/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13993`

`$ git pr show -t 13993`

</details>

<a href="https://git.openjdk.org/jdk/pull/13993.diff">https://git.openjdk.org/jdk/pull/13993.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/services/memReporter.cpp
+++ b/src/hotspot/share/services/memReporter.cpp
@@ -842,10 +842,20 @@ void MemDetailDiffReporter::old_virtual_memory_site(const VirtualMemoryAllocatio
 
 void MemDetailDiffReporter::diff_virtual_memory_site(const VirtualMemoryAllocationSite* early,
   const VirtualMemoryAllocationSite* current) const {
-  assert(early->flag() == current->flag() || early->flag() == mtNone,
-    "Expect the same flag, but %s != %s", NMTUtil::flag_to_name(early->flag()),NMTUtil::flag_to_name(current->flag()));
+  if ((early->flag() != current->flag()) || (current->flag() == mtNone) || (early->flag() == mtNone)) {
+    tty->print("Detected an anomaly in MemDetailDiffReporter::diff_virtual_memory_site:");
+    early->call_stack()->print_on(tty);
+    current->call_stack()->print_on(tty);
+  }
+  assert(current->flag() != mtNone && early->flag() != mtNone,
+         "Expect the flag not to be mtNone, but %s or %s is.",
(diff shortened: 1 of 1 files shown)


src/hotspot/share/services/memReporter.cpp
```diff
@@ -842,10 +842,20 @@ void MemDetailDiffReporter::old_virtual_memory_site(const VirtualMemoryAllocatio
 
 void MemDetailDiffReporter::diff_virtual_memory_site(const VirtualMemoryAllocationSite* early,
   const VirtualMemoryAllocationSite* current) const {
-  assert(early->flag() == current->flag() || early->flag() == mtNone,
-    "Expect the same flag, but %s != %s", NMTUtil::flag_to_name(early->flag()),NMTUtil::flag_to_name(current->flag()));
+  if ((early->flag() != current->flag()) || (current->flag() == mtNone) || (early->flag() == mtNone)) {
```
dholmes-ora, line 845: Shouldn't this be checking for `!= mtNone` as that is what would trip the original assert?

Shouldn't this checking code be under `ifdef ASSERT`?

tstuefe: For simplicity, I usually just take the assert condition, bracket it, and prefix with !. Saves me having to think up the negation.

gerard-ziemski:
Thank you David and Thomas for feedback!

Based on Thomas' comment in the bug, we need to re-think this.

gerard-ziemski:
Moved to https://github.com/openjdk/jdk/pull/14052

PR #13995, opened 2023-05-15
8308071: [REDO] update for deprecated sprintf for src/utils

XueleiFan:
Hi,

This is a redo of JDK-8307855, where issues were found after integration.

The sprintf is deprecated in Xcode 14, and Microsoft Virtual Studio, because of security concerns. The issue was addressed in [JDK-8296812](https://bugs.openjdk.org/browse/JDK-8296812) for building failure, and [JDK-8299378](https://bugs.openjdk.org/browse/JDK-8299378)/[JDK-8299635](https://bugs.openjdk.org/browse/JDK-8299635)/[JDK-8301132](https://bugs.openjdk.org/browse/JDK-8301132) for testing issues . This is a break-down update for sprintf uses in the src/utils directory.

Thanks,
Xuelei

---------
### Progress
- [ ] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Issue
 * [JDK-8308071](https://bugs.openjdk.org/browse/JDK-8308071): [REDO] update for deprecated sprintf for src/utils (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13995/head:pull/13995` \
`$ git checkout pull/13995`

`$ git checkout pull/13995` \
`$ git pull https://git.openjdk.org/jdk.git pull/13995/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13995`

`$ git pr show -t 13995`

</details>

<a href="https://git.openjdk.org/jdk/pull/13995.diff">https://git.openjdk.org/jdk/pull/13995.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/13995#issuecomment-1548509336)

### DIFF
--- a/src/utils/hsdis/binutils/hsdis-binutils.c
+++ b/src/utils/hsdis/binutils/hsdis-binutils.c
@@ -242,19 +242,18 @@ static const char* format_insn_close(const char* close,
   case dis_noninsn:     type = "noninsn";    break;
   }
 
-  strcpy(buf, close);
-  size_t used_size = strlen(close);
+  /* the buf capacity has been checked previously */
+  size_t used_size = snprintf(buf, bufsize, "%s", close);
   char* p = buf + used_size;
   bufsize -= used_size;
+
   if (type) {
-    /* the buf capacity has been checked previously */
     used_size = snprintf(p, bufsize, " type='%s'", type);
     p += used_size;
     bufsize -= used_size;
   }
 
   if (dsize) {
-    /* the buf capacity has been checked previously */
     used_size = snprintf(p, bufsize, " dsize='%d'", dsize);
     p += used_size;
     bufsize -= used_size;
--- a/src/utils/hsdis/binutils/hsdis-binutils.c
+++ b/src/utils/hsdis/binutils/hsdis-binutils.c
@@ -242,7 +242,7 @@ static const char* format_insn_close(const char* close,
   case dis_noninsn:     type = "noninsn";    break;
   }
 
-  size_t used_size = snprintf(buf, bufsize, "%s", close);
+  int used_size = snprintf(buf, bufsize, "%s", close);
   if ((used_size < 0) || (used_size >= bufsize)) {
     return close;
   }
--- a/src/utils/hsdis/binutils/hsdis-binutils.c
+++ b/src/utils/hsdis/binutils/hsdis-binutils.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * The Universal Permissive License (UPL), Version 1.0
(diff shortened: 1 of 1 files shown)


src/utils/hsdis/binutils/hsdis-binutils.c
```diff
@@ -243,10 +243,27 @@ static const char* format_insn_close(const char* close,
   }
 
   strcpy(buf, close);
-  char* p = buf;
-  if (type)    sprintf(p += strlen(p), " type='%s'", type);
-  if (dsize)   sprintf(p += strlen(p), " dsize='%d'", dsize);
-  if (delays)  sprintf(p += strlen(p), " delay='%d'", delays);
+  size_t used_size = strlen(close);
+  char* p = buf + used_size;
+  bufsize -= used_size;
```
vidmik, line 248: May not happen in practice, but if `used_size` is larger than `bufsize` this will wrap to a very large value. Perhaps the strcpy above should also be an `snprintf`, and the return value handled the same way as for the subsequent `snprintf` calls?

XueleiFan: I think it is safe as the `buf` size has been checked at around line 230.  However, it may make the code easier to read if replacing `strcpy` with `snprintf`.  The patch was updated accordingly.

kimbarrett: This and all uses of snprintf in this change are incorrect.  If the output is truncated, snprintf returns the
number of characters that would have been written if there had been enough space.  That is, the result
may be larger than bufsize.

kimbarrett (changes requested):

src/utils/hsdis/binutils/hsdis-binutils.c
XueleiFan: > This and all uses of snprintf in this change are incorrect. If the output is truncated, snprintf returns the number of characters that would have been written if there had been enough space. That is, the result may be larger than bufsize.

The correctness of this change depends on the fact that the buffer has sufficient capacity, which has been checked at line 230.  I agreed that this is not a typical use of `snprintf` that the returned value is not checked.   I will make an update to check the returned value of `snprintf`.

src/utils/hsdis/binutils/hsdis-binutils.c
```diff
@@ -242,11 +242,38 @@ static const char* format_insn_close(const char* close,
   case dis_noninsn:     type = "noninsn";    break;
   }
 
-  strcpy(buf, close);
-  char* p = buf;
-  if (type)    sprintf(p += strlen(p), " type='%s'", type);
-  if (dsize)   sprintf(p += strlen(p), " dsize='%d'", dsize);
-  if (delays)  sprintf(p += strlen(p), " delay='%d'", delays);
+  size_t used_size = snprintf(buf, bufsize, "%s", close);
+  if ((used_size < 0) || (used_size >= bufsize)) {
```
kimbarrett, line 246: (used_size < 0) is tautologically false, since used_size is a size_t, so unsigned.  I'm somewhat surprised
this doesn't trigger a warning from some compiler.

src/utils/hsdis/binutils/hsdis-binutils.c
kimbarrett: OK, I missed that.  (The relevant code doesn't show up in the default github diff.  I really ought to know better
than to use that view for reviewing.)  Even having been pointed to the code, I had to do some counting and
such to convince myself that it was safe.  A bit of commentary might save some time for the next reader.

kimbarrett (changes requested):

src/utils/hsdis/binutils/hsdis-binutils.c
kimbarrett: bufsize is size_t, so that's a comparison between signed and unsigned values, which I think some compilers
will warn about.  Maybe the preceding check for negative is getting rid of that?  But will that still occur in
a slowdebug build, or will the lack of optimization lead to a warning?

XueleiFan: As always, this comment helps a lot.  Thank you!

Updated to cast `int` to `size_t` explicitly.

XueleiFan: @kimbarrett Did you have a chance to have another look?  Please let me know if you prefer to the update that the returned value of snprintf() is not checked because the memory size has been checked previously.

PR #14004, opened 2023-05-16
8306119: Many components respond to a mouse event by requesting focus without supplying the MOUSE_EVENT cause

prsadhuk:
Many Swing components request a focus transfer in response to a mouse event, but fail to supply a Cause when requesting focus. A focus event listener will find the cause to be UNKNOWN, but MOUSE_EVENT would be more appropriate. 
Fixed by adding MOUSE_EVENT cause to requestFocus() when it is called for mouse events...

All jtreg/jck tests are ok..

---------
### Progress

### Issue
 * [JDK-8306119](https://bugs.openjdk.org/browse/JDK-8306119): Many components respond to a mouse event by requesting focus without supplying the MOUSE_EVENT cause

### Reviewers
 * [Jayathirth D V](https://openjdk.org/census#jdv) (@jayathirthrao - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14004/head:pull/14004` \
`$ git checkout pull/14004`

`$ git checkout pull/14004` \
`$ git pull https://git.openjdk.org/jdk.git pull/14004/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14004`

`$ git pr show -t 14004`

</details>

<a href="https://git.openjdk.org/jdk/pull/14004.diff">https://git.openjdk.org/jdk/pull/14004.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14004#issuecomment-1549277662)

### DIFF
--- a/src/java.desktop/macosx/classes/com/apple/laf/AquaSpinnerUI.java
+++ b/src/java.desktop/macosx/classes/com/apple/laf/AquaSpinnerUI.java
@@ -603,7 +603,7 @@ public class AquaSpinnerUI extends SpinnerUI {
             final Component child = ftp.getComponentAfter(root, spinner);
 
             if (child != null && SwingUtilities.isDescendingFrom(child, spinner)) {
-                child.requestFocus();
+                child.requestFocus(FocusEvent.Cause.MOUSE_EVENT);
             }
         }
 
--- a/src/java.desktop/share/classes/javax/swing/plaf/basic/BasicButtonListener.java
+++ b/src/java.desktop/share/classes/javax/swing/plaf/basic/BasicButtonListener.java
@@ -261,7 +261,7 @@ public class BasicButtonListener implements MouseListener, MouseMotionListener,
              }
              model.setPressed(true);
              if(!b.hasFocus() && b.isRequestFocusEnabled()) {
-                b.requestFocus();
+                b.requestFocus(FocusEvent.Cause.MOUSE_EVENT);
              }
           }
        }
@@ -322,7 +322,7 @@ public class BasicButtonListener implements MouseListener, MouseMotionListener,
                 model.setArmed(true);
                 model.setPressed(true);
                 if(!b.hasFocus()) {
-                    b.requestFocus();
+                    b.requestFocus(FocusEvent.Cause.MOUSE_EVENT);
                 }
             }
             else if (key == RELEASE) {
--- a/src/java.desktop/share/classes/javax/swing/plaf/basic/BasicButtonListener.java
+++ b/src/java.desktop/share/classes/javax/swing/plaf/basic/BasicButtonListener.java
@@ -322,7 +322,7 @@ public class BasicButtonListener implements MouseListener, MouseMotionListener,
                 model.setArmed(true);
                 model.setPressed(true);
                 if(!b.hasFocus()) {
-                    b.requestFocus(FocusEvent.Cause.MOUSE_EVENT);
+                    b.requestFocus();
                 }
             }
             else if (key == RELEASE) {
--- a/src/java.desktop/share/classes/javax/swing/text/DefaultCaret.java
+++ b/src/java.desktop/share/classes/javax/swing/text/DefaultCaret.java
@@ -582,10 +582,10 @@ public class DefaultCaret extends Rectangle implements Caret, FocusListener, Mou
         if ((component != null) && component.isEnabled() &&
                                    component.isRequestFocusEnabled()) {
             if (inWindow) {
-                component.requestFocusInWindow();
+                component.requestFocusInWindow(FocusEvent.Cause.MOUSE_EVENT);
             }
             else {
-                component.requestFocus();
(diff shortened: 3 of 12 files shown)


src/java.desktop/macosx/classes/com/apple/laf/AquaSpinnerUI.java
```diff
@@ -603,7 +603,7 @@ private void focusSpinnerIfNecessary() {
             final Component child = ftp.getComponentAfter(root, spinner);
 
             if (child != null && SwingUtilities.isDescendingFrom(child, spinner)) {
-                child.requestFocus();
+                child.requestFocus(FocusEvent.Cause.MOUSE_EVENT);
```
jayathirthrao, line 606: I see that this function is called only from mousePressed. This seems fine.

src/java.desktop/share/classes/javax/swing/plaf/basic/BasicButtonListener.java
```diff
@@ -322,7 +322,7 @@ public void actionPerformed(ActionEvent e) {
                 model.setArmed(true);
                 model.setPressed(true);
                 if(!b.hasFocus()) {
-                    b.requestFocus();
+                    b.requestFocus(FocusEvent.Cause.MOUSE_EVENT);
```
jayathirthrao, line 325: Can this action be performed using a keyboard or this can be reached only using mouse press?
If this actionPerformed can be reached using keyboard, i think we should add appropriate checks before throwing the focus event cause.

src/java.desktop/share/classes/javax/swing/text/DefaultCaret.java
```diff
@@ -582,10 +582,10 @@ private void adjustFocus(boolean inWindow) {
         if ((component != null) && component.isEnabled() &&
                                    component.isRequestFocusEnabled()) {
             if (inWindow) {
-                component.requestFocusInWindow();
+                component.requestFocusInWindow(FocusEvent.Cause.MOUSE_EVENT);
```
jayathirthrao, line 585: Looks like this function is also called only in case of mouse events from mousePressed()->adjustCaretAndFocus() and mouseClicked(). Seems fine.

test/jdk/javax/swing/event/FocusEventCauseTest.java
```diff
@@ -0,0 +1,102 @@
...
+    static volatile boolean expected;
+
+    public static void main(String[] args) throws Exception {
+        try {
+            Robot robot = new Robot();
+            robot.setAutoDelay(100);
+            SwingUtilities.invokeAndWait(() -> {
+                frame = new JFrame("FocusEventCauseTest");
+                JPanel panel = new JPanel();
+                button1 = new JButton("Button1");
```
jayathirthrao, line 59: Generic test comment. Its good if we can check focus event for all UI components that are getting updated.

jayathirthrao (commented):
I see concerns related to SwingUtilities2 and Accessibility in https://bugs.openjdk.org/browse/JDK-8306119 description.
If we are not handling them in this fix, do we have separate JBS bug for them?

prsadhuk:
> I see concerns related to SwingUtilities2 and Accessibility in https://bugs.openjdk.org/browse/JDK-8306119 description. If we are not handling them in this fix, do we have separate JBS bug for them?

As of now, I will let Accessibility team handle the change in a separate JBS issue. Also SwingUtilities concern is to add additonal method which will need CSR which should be handled as separate issue again, so not handled here..

test/jdk/javax/swing/event/FocusEventCauseTest.java
prsadhuk: Actually, the fix is trivial, the test is only added to show how the cause whould be presented to the user using a widget.
The other widgets would be same in showing focus cause event to be MOUSE_EVENT so not handled separately..

src/java.desktop/share/classes/javax/swing/plaf/basic/BasicButtonListener.java
prsadhuk: I have removed this contentious change as it probably can be used using keyboard too.  For now, just kept the other mouse only focus event cause in this file..

jayathirthrao:
> > I see concerns related to SwingUtilities2 and Accessibility in https://bugs.openjdk.org/browse/JDK-8306119 description. If we are not handling them in this fix, do we have separate JBS bug for them?
> 
> As of now, I will let Accessibility team handle the change in a separate JBS issue. Also SwingUtilities concern is to add additonal method which will need CSR which should be handled as separate issue again, so not handled here..

Sure please raise appropriate JBS issues once we finalise the fix for this issue.

jayathirthrao (approved):

prsadhuk:
I have raised [JDK-8309041](https://bugs.openjdk.org/browse/JDK-8309041) for SwingUtilities focus upgrade

PR #14006, opened 2023-05-16
7065228: To interpret case-insensitive string locale independently

DarraghClarke:
Updated instances of `toLowerCase` and `toUpperCase` in several net and io files to specify `Locale.ROOT` to ensure that case conversion issues don't occur,

I didn't add any new tests but ran tier 1-3 with no issues


src/java.base/share/classes/java/io/StreamTokenizer.java
```diff
@@ -628,7 +629,7 @@ else if ('0' <= c && c <= '9') {
             peekc = c;
             sval = String.copyValueOf(buf, 0, i);
             if (forceLower)
-                sval = sval.toLowerCase();
+                sval = sval.toLowerCase(Locale.ROOT);
```
dfuch, line 632: This one gave me pause. AFAICS it's probably OK - but it might warant a CSR and an update of the specification to explicitly state how the lower case conversion is performed (update the `lowerCaseMode` method spec). Could we handle that in a separate PR?

dfuch (commented):
Looks generally like a good improvement. In those places `toLowerCase` is called on strings (or substrings) that are expected to be ASCII, so using Locale.ROOT looks appropriate.
Would be good to get @Michael-Mc-Mahon validate these changes too.

naotoj (approved):
LGTM. Nice clean-up!

src/java.base/share/classes/sun/net/www/protocol/http/DigestAuthentication.java
```diff
@@ -513,7 +513,7 @@ private String getHeaderValueImpl (String uri, String method) {
         String ncstring=null;
 
         if (nccount != -1) {
-            ncstring = Integer.toHexString (nccount).toLowerCase();
+            ncstring = Integer.toHexString (nccount).toLowerCase(Locale.ROOT);
```
djelinski, line 516:
```suggestion
            ncstring = Integer.toHexString(nccount);
```
`toHexString` returns lowercase string

djelinski (approved):

dfuch (approved):
LGTM

jaikiran (approved):
These changes look fine to me.

Michael-Mc-Mahon:
Seems like a useful change and I can see how issues could arise if strings were stored somewhere after being upper/lower cased and then reused in a different locale. 

Is it correct to say that the assumption is these strings are all supposed to be US ASCII (eg protocol defined identifiers, or hostnames etc) rather than user generated text strings? That seems to be the case as far as I can see.

Michael-Mc-Mahon (approved):

DarraghClarke:
> Seems like a useful change and I can see how issues could arise if strings were stored somewhere after being upper/lower cased and then reused in a different locale.
> 
> Is it correct to say that the assumption is these strings are all supposed to be US ASCII (eg protocol defined identifiers, or hostnames etc) rather than user generated text strings? That seems to be the case as far as I can see.

Yes

PR #14007, opened 2023-05-16
Tweak alignment of long/double constants on x86.

mcimadamore:
Some text here

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/ValueLayout.java
+++ b/src/java.base/share/classes/java/lang/foreign/ValueLayout.java
@@ -485,7 +485,7 @@ public sealed interface ValueLayout extends MemoryLayout permits
 
     /**
      * A value layout constant whose size is the same as that of a Java {@code long},
-     * bit alignment set to 64, and byte order set to {@link ByteOrder#nativeOrder()}.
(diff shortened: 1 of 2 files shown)


mcimadamore:
Javadoc: https://cr.openjdk.org/~mcimadamore/jdk/8308248/8308248/v1/javadoc/java.base/module-summary.html
Specdiff: https://cr.openjdk.org/~mcimadamore/jdk/8308248/8308248/v1/specdiff_out/overview-summary.html

PaulSandoz (approved):

src/java.base/share/classes/jdk/internal/foreign/layout/ValueLayouts.java
```diff
@@ -295,7 +295,7 @@ OfLongImpl dup(ByteOrder order, long bitAlignment, Optional<String> name) {
         }
 
         public static OfLong of(ByteOrder order) {
-            return new OfLongImpl(order, Long.SIZE, Optional.empty());
+            return new OfLongImpl(order, ADDRESS_SIZE_BITS, Optional.empty());
```
mcimadamore, line 298: Whoops - I forgot these all-important changes :-)

minborg:
Does it make sense to add two test classes, each running on a 64/32-bit platform that verifies the two alignment variants?

PR #13864, opened 2023-05-08
8305634: Renaming predicates, simple cleanups, and adding summary about current predicates

chhagedorn:
This is the first patch in a series of patches to fix the remaining issues with Assertion/Skeleton Predicates ([JDK-8288981](https://bugs.openjdk.org/browse/JDK-8288981)). To achieve this task, we've decided to cleanup the naming scheme of predicates, cleanup the code operating on predicates (matching predicates, skipping over predicates etc.), and finally redesign the Assertion/Skeleton Predicates. While the basic idea stays the same, we are using new nodes and unified, cleaned up classes and methods to do the job. This redesign simplifies and only even makes it possible to fix the remaining known issues in a clean way.

To make reviewing the entire change easier, I've decided to split the work into several PRs.

This first PR includes the following _semantic-preserving_ changes:
- Establishing a new naming scheme for all the predicates found in C2 which makes it easier to talk about the various kinds of predicates:
  - Updating the code (variables, method names etc.) accordingly. 
  - Renaming "Skeleton Predicates" to "Assertion Predicates".
  - Including a summary of all predicates found in C2 in `loopPredicate.cpp`.
  - Capitalizing predicate names to better distinguish them in comments (e.g. "Parse Predicate" instead of "parse predicate").
- Change `class Predicates` -> `class ParsePredicates`. 
- Improving type information (e.g. using `IfProjNode` instead of `ProjNode`, using `ParsePredicateSuccessProj/ParsePredicateUncommonProj` typedefs for Parse Predicates etc.).
- Removing unused variables.
- Removing unnecessary checks.
- Code style fixes in touched code.

Instead of giving more background information about Assertion Predicates and why we need them here, I've decided to write a _blog post_ dedicated to Assertion Predicates and Loop Predication. This provides an overview and introduction which, hopefully, makes reviewing the PRs related to Assertion Predicates easier - at least on a higher level. 

The blog post can be found on my Github page at:
https://chhagedorn.github.io/jdk/2023/05/05/assertion-predicates.html

Thanks to @rwestrel, @eme64, and @TobiHartmann for your help with discussions, brainstormings, and pre-reviewing some of the changes here and in upcoming PRs.

Thanks,
Christian

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
[...]

PR #14017, opened 2023-05-16
8305635: Replace Parse Predicate IfNode with new ParsePredicateNode and route predicate queries through dedicated classes

chhagedorn:
This is the second PR towards fixing the issues with Assertion Predicates ([JDK-8288981](https://bugs.openjdk.org/browse/JDK-8288981)). This patch still does not change anything in the way the old Assertion Predicates work. The only observable change in the IR is the introduction of a new `ParsePredicateNode` instead of using an `IfNode` to better distinguish these dedicated Parse Predicates added during parsing (they still use the same inputs with `Opaque1Nodes` as before).

Changes include:
- New `ParsePredicateNode` as subclass of `IfNode` and related code updates to make this work.
- Moving predicate access code (skipping, matching etc.), including the called predicate methods found in `PhaseIdealLoop`, to dedicated `Predicates/ParsePredicates` classes. This is only a first step and these classes are further updated in the next PR. They can therefore be seen as an intermediate state to make the entire update to predicate classes easier to follow. As a consequence, I've tried to not clean the code up too much in these classes.
  - Cleanup of touched code (dead code, variable renaming, code style)
  -  Added comments (e.g. for some special case in Loop Predication)

For more background, have a look at the first PR: #13864

Thanks,
Christian

---------
### Progress

### Issue
 * [JDK-8305635](https://bugs.openjdk.org/browse/JDK-8305635): Replace Parse Predicate IfNode with new ParsePredicateNode and route predicate queries through dedicated classes

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14017/head:pull/14017` \
`$ git checkout pull/14017`

`$ git checkout pull/14017` \
`$ git pull https://git.openjdk.org/jdk.git pull/14017/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14017`

`$ git pr show -t 14017`

</details>

<a href="https://git.openjdk.org/jdk/pull/14017.diff">https://git.openjdk.org/jdk/pull/14017.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14017#issuecomment-1549906480)


src/hotspot/share/opto/node.hpp
```diff
@@ -1019,6 +1022,11 @@ class Node {
 
   bool for_post_loop_opts_igvn() const { return (_flags & Flag_for_post_loop_opts_igvn) != 0; }
 
+  // Is 'n' possibly a loop entry (i.e. a Parse Predicate projection)?
+  static bool is_maybe_loop_entry(Node* n) {
```
TobiHartmann, line 1026:
```suggestion
  static bool may_be_loop_entry(Node* n) {
```

src/hotspot/share/opto/cfgnode.hpp
```diff
@@ -452,6 +452,26 @@ class RangeCheckNode : public IfNode {
   virtual Node* Ideal(PhaseGVN *phase, bool can_reshape);
 };
 
+// Special node that denotes a Parse Predicate added during parsing. A Parse Predicate serves as placeholder to later
+// create Runtime Predicates above it. They all share the same uncommon trap. The Parse Predicate will follow the
+// Runtime Predicates. Together they form a Regular Predicate Block. There are three kinds of Parse Predicates:
+// Loop Parse Predicate, Profiled Loop Parse Predicate (both used by Loop Predication), and Loop Limit Check Parse
+// Predicate (used for integer overflow checks when creating a counted loop).
+// More information about predicates can be found at loopPredicate.cpp.
```
TobiHartmann, line 460:
```suggestion
// More information about predicates can be found in loopPredicate.cpp.
```

src/hotspot/share/opto/cfgnode.hpp
```diff
@@ -452,6 +452,26 @@ class RangeCheckNode : public IfNode {
...
+class ParsePredicateNode : public IfNode {
+  Deoptimization::DeoptReason _deopt_reason;
```
TobiHartmann, line 462: Don't we need to override `Node::hash()` and `Node::cmp()` here to account for the `_deopt_reason` field?

TobiHartmann (changes requested):

src/hotspot/share/opto/cfgnode.hpp
chhagedorn: Since it is a CFG node, the hash should always be different due to a different control input. If two `If` nodes have the same control input after an optimization, the graph is broken.

TobiHartmann: Right, I missed that. Looks goot then!

TobiHartmann (approved):

vnkozlov (approved):
Looks good.

chhagedorn:
Thanks Vladimir for your review!

PR #14019, opened 2023-05-16
8308231: Faster MemAllocator::Allocation checks for verify/notification

shipilev:
In multi-array allocations benchmarks, there is a hot path through the native VM allocation code, which calls lots of notification methods, even when we would return immediately, because the allocation was satisfied from existing TLAB. Not calling these helper methods from `MemAllocator::Allocation` constructor/destructor looks like an incremental win for the benchmarks.

Example on M1:

```
Benchmark             (size)  Mode  Cnt        Score        Error  Units

# Before
MultiArrayAlloc.full       1  avgt   15       74,053 ±      0,869  ns/op
MultiArrayAlloc.full       2  avgt   15       87,800 ±      0,931  ns/op
MultiArrayAlloc.full       4  avgt   15      124,814 ±      0,615  ns/op
MultiArrayAlloc.full       8  avgt   15      188,562 ±      0,785  ns/op
MultiArrayAlloc.full      16  avgt   15      313,007 ±      1,108  ns/op
MultiArrayAlloc.full      32  avgt   15      640,276 ±      4,560  ns/op
MultiArrayAlloc.full      64  avgt   15     1395,220 ±      5,860  ns/op
MultiArrayAlloc.full     128  avgt   15     3417,848 ±     11,345  ns/op
MultiArrayAlloc.full     256  avgt   15     9955,360 ±    102,057  ns/op
MultiArrayAlloc.full     512  avgt   15    27738,002 ±    244,940  ns/op
MultiArrayAlloc.full    1024  avgt   15   147507,008 ±   1434,085  ns/op

# After
MultiArrayAlloc.full       1  avgt   15       70,434 ±      0,373  ns/op  ;  5% better
MultiArrayAlloc.full       2  avgt   15       82,394 ±      0,137  ns/op  ;  7% better
MultiArrayAlloc.full       4  avgt   15      108,542 ±      0,129  ns/op  ; 15% better
MultiArrayAlloc.full       8  avgt   15      170,697 ±      4,480  ns/op  ; 11% better
MultiArrayAlloc.full      16  avgt   15      272,902 ±      0,877  ns/op  ; 15% better
MultiArrayAlloc.full      32  avgt   15      524,486 ±      1,447  ns/op  ; 22% better
MultiArrayAlloc.full      64  avgt   15     1088,932 ±      2,739  ns/op  ; 17% better
MultiArrayAlloc.full     128  avgt   15     3151,144 ±     14,621  ns/op  ;  8% better
MultiArrayAlloc.full     256  avgt   15     8455,293 ±     12,656  ns/op  ; 18% better
MultiArrayAlloc.full     512  avgt   15    26060,055 ±    116,524  ns/op  ;  6% better
MultiArrayAlloc.full    1024  avgt   15   130824,480 ±    831,703  ns/op  ; 13% better
```

Additional testing:
 - [x] Ad-hoc micro-benchmarks
 - [x] Linux x86_64 fastdebug `serviceability/jvmti`
 - [x] Linux x86_64 fastdebug `jdk/jfr`
 - [x] Linux x86_64 fastdebug `tier1 tier2 tier3`
 - [x] Linux AArch64 fastdebug `tier1 tier2 tier3`

---------
### Progress
- [ ] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Issue
 * [JDK-8308231](https://bugs.openjdk.org/browse/JDK-8308231): Faster MemAllocator::Allocation checks for verify/notification (**Enhancement** - P4)

### Reviewers
 * [Leonid Mesnik](https://openjdk.org/census#lmesnik) (@lmesnik - **Reviewer**) ⚠️ Review applies to [d189a358](https://git.openjdk.org/jdk/pull/14019/files/d189a3587c084177cbed4c4ef30c7c15eb3f6287)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14019/head:pull/14019` \
`$ git checkout pull/14019`

`$ git checkout pull/14019` \
`$ git pull https://git.openjdk.org/jdk.git pull/14019/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14019`

`$ git pr show -t 14019`

</details>

<a href="https://git.openjdk.org/jdk/pull/14019.diff">https://git.openjdk.org/jdk/pull/14019.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14019#issuecomment-1551264337)

### DIFF
--- a/src/hotspot/share/gc/shared/memAllocator.cpp
+++ b/src/hotspot/share/gc/shared/memAllocator.cpp
@@ -53,10 +53,8 @@ class MemAllocator::Allocation: StackObj {
   size_t              _allocated_tlab_size;
   bool                _tlab_end_reset_for_sample;
 
-  bool check_out_of_memory();
+  bool throw_out_of_memory();
   void verify_before();
-  void verify_after();
-  void notify_allocation();
   void notify_allocation_jvmti_sampler();
   void notify_allocation_low_memory_detector();
   void notify_allocation_jfr_sampler();
@@ -77,13 +75,38 @@ public:
       _allocated_tlab_size(0),
       _tlab_end_reset_for_sample(false)
   {
+#ifdef ASSERT
     assert(Thread::current() == allocator._thread, "do not pass MemAllocator across threads");
     verify_before();
+#endif
   }
 
   ~Allocation() {
-    if (!check_out_of_memory()) {
-      notify_allocation();
+    if (obj() != nullptr) {
+      // Real allocation allocates actual memory: either allocates
+      // outside the TLAB, or allocates a new TLAB.
+      const bool is_real_allocation = _allocated_outside_tlab ||
+                                      (_allocated_tlab_size != 0);
+
+      if (is_real_allocation) {
+        notify_allocation_low_memory_detector();
+        notify_allocation_jfr_sampler();
+      }
+
+      if ((is_real_allocation || _tlab_end_reset_for_sample) &&
+          JvmtiExport::should_post_sampled_object_alloc()) {
+        notify_allocation_jvmti_sampler();
+      }
+
+      if (JvmtiExport::should_post_vm_object_alloc()) {
+        JvmtiExport::record_vm_internal_object_allocation(obj());
+      }
+
+      if (DTraceAllocProbes) {
+        notify_allocation_dtrace_sampler();
+      }
+    } else {
+      throw_out_of_memory();
     }
   }
 
@@ -113,13 +136,9 @@ public:
   }
 };
 
-bool MemAllocator::Allocation::check_out_of_memory() {
+bool MemAllocator::Allocation::throw_out_of_memory() {
   JavaThread* THREAD = _thread; // For exception macros.
   assert(!HAS_PENDING_EXCEPTION, "Unexpected exception, will result in uninitialized storage");
-
-  if (obj() != nullptr) {
-    return false;
-  }
 
   const char* message = _overhead_limit_exceeded ? "GC overhead limit exceeded" : "Java heap space";
   if (!_thread->in_retryable_allocation()) {
@@ -140,16 +159,16 @@ bool MemAllocator::Allocation::check_out_of_memory() {
   }
 }
 
+#ifdef ASSERT
 void MemAllocator::Allocation::verify_before() {
   // Clear unhandled oops for memory allocation.  Memory allocation might
   // not take out a lock if from tlab, so clear here.
   JavaThread* THREAD = _thread; // For exception macros.
   assert(!HAS_PENDING_EXCEPTION, "Should not allocate with exception pending");
-  debug_only(check_for_valid_allocation_state());
+  check_for_valid_allocation_state();
   assert(!Universe::heap()->is_gc_active(), "Allocation during gc not allowed");
 }
 
-#ifdef ASSERT
 void MemAllocator::Allocation::check_for_valid_allocation_state() const {
   // How to choose between a pending exception and a potential
   // OutOfMemoryError?  Don't allow pending exceptions.
@@ -162,20 +181,6 @@ void MemAllocator::Allocation::check_for_valid_allocation_state() const {
 #endif
 
 void MemAllocator::Allocation::notify_allocation_jvmti_sampler() {
-  // support for JVMTI VMObjectAlloc event (no-op if not enabled)
-  JvmtiExport::vm_object_alloc_event_collector(obj());
-
-  if (!JvmtiExport::should_post_sampled_object_alloc()) {
-    // Sampling disabled
-    return;
-  }
-
-  if (!_allocated_outside_tlab && _allocated_tlab_size == 0 && !_tlab_end_reset_for_sample) {
-    // Sample if it's a non-TLAB allocation, or a TLAB allocation that either refills the TLAB
-    // or expands it due to taking a sampler induced slow path.
-    return;
-  }
-
   // If we want to be sampling, protect the allocated object with a Handle
   // before doing the callback. The callback is done in the destructor of
   // the JvmtiSampledObjectAllocEventCollector.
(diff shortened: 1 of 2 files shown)


lmesnik (approved):
Still need review from gc team.

src/hotspot/share/gc/shared/memAllocator.cpp
```diff
@@ -66,6 +66,23 @@ class MemAllocator::Allocation: StackObj {
   void check_for_valid_allocation_state() const;
 #endif
 
+  // Should notify allocation when either of these happen:
+  //   - a non-TLAB allocation;
+  //   - a TLAB allocation that refills the TLAB;
+  //   - a TLAB allocation that expands due to taking a sampler induced slow path;
+  //   - (optionally) the enabled JVMTI event that wants to capture all allocations;
+
+  bool should_notify_allocation_no_jvmti_vmobjalloc() {
```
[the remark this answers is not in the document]
stefank: FWIW, I found the name `should_notify_allocation_no_jvmti_vmobjalloc` confusing when I read this patch. Maybe a better name would be `allocated_in_slow_path()` or negate it to `!allocated_in_fast_path()`?

src/hotspot/share/gc/shared/memAllocator.cpp
```diff
@@ -84,7 +101,9 @@ class MemAllocator::Allocation: StackObj {
   ~Allocation() {
     if (!check_out_of_memory()) {
       verify_after();
-      notify_allocation(_thread);
+      if (should_notify_allocation()) {
+        notify_allocation(_thread);
+      }
```
stefank, line 106: `notify_allocation` used to call   `notify_allocation_low_memory_detector` and `notify_allocation_dtrace_sampler` without performing any "should notify" filtering. Is this change in behavior intentional?

src/hotspot/share/gc/shared/memAllocator.cpp
```diff
@@ -187,9 +206,8 @@ void MemAllocator::Allocation::notify_allocation_jvmti_sampler() {
     return;
   }
 
-  if (!_allocated_outside_tlab && _allocated_tlab_size == 0 && !_tlab_end_reset_for_sample) {
-    // Sample if it's a non-TLAB allocation, or a TLAB allocation that either refills the TLAB
-    // or expands it due to taking a sampler induced slow path.
+  if (!should_notify_allocation_no_jvmti_vmobjalloc()) {
+    // Called here only for JVMTI VMObjectAlloc event
```
stefank, line 210: This comment sounds like we are executing the return statement because of `JVMTI VMObjectAlloc event`, but isn't it the opposite? The reason why we are in this function and *not* executing the return statement is because the VMObjectAlloc event is turned on. Would it make more sense to move the comment to after this if block?

stefank (changes requested):

albertnetymk:
low-mem-detector and jfr are interested only in real-allocation (increasing mem usage), while dtrace will capture inside-tlab allocations also.

I think your change alters the behavior of dtrace, from capturing all-allocations to only jvmti-filtered ones. It can be surprising that the result of dtrace depends on some jvmti flags.

I reorganized the notify-methods and break them into three groups:

1. capturing all-allocations -- dtrace and jvmti-vm-obj-alloc
2. capturing real-allocation -- low-mem-detector and jfr
3. capturing real or pseudo alloc -- jvmti-sampler

The most important part looks like:

```cpp
  ~Allocation() {
    if (!check_out_of_memory()) {
      verify_after();

      notify_allocation_dtrace_sampler(_thread);

      // support for JVMTI VMObjectAlloc event (no-op if not enabled)
      JvmtiExport::vm_object_alloc_event_collector(obj());

      const bool is_real_allocation = _allocated_tlab_size != 0
                                   || _allocated_outside_tlab;

      if (is_real_allocation) {
        notify_allocation_low_memory_detector();
        notify_allocation_jfr_sampler();
      }

      if (is_real_allocation || _tlab_end_reset_for_sample) {
        notify_allocation_jvmti_sampler();
      }
    }
  }
```

src/hotspot/share/gc/shared/memAllocator.cpp
shipilev: Not really, but I wonder if we actually need to do these for inside-TLAB allocs. I think I'll go the way Albert leans, and see what happens here.

shipilev:
> I think your change alters the behavior of dtrace, from capturing all-allocations to only jvmti-filtered ones. It can be surprising that the result of dtrace depends on some jvmti flags.

Right. I just committed another variant, inspired by your suggestion. It passes testing for me, but I would be able to do performance tests only tomorrow.

src/hotspot/share/gc/shared/memAllocator.cpp
```diff
@@ -84,7 +83,29 @@ class MemAllocator::Allocation: StackObj {
...
+      const bool is_real_allocation = _allocated_outside_tlab ||
+                                      (_allocated_tlab_size != 0);
+
+      if (is_real_allocation) {
+        notify_allocation_low_memory_detector();
+        notify_allocation_jfr_sampler();
+      }
+
+      if ((is_real_allocation || _tlab_end_reset_for_sample) &&
+          JvmtiExport::should_post_sampled_object_alloc()) {
```
albertnetymk, line 98: The same check is done inside the caller already.

On this note, I think all checks using info outside `class Allocation` should not be present on this level. (Ofc, this is very subjective.)

shipilev: This is a fast-path code, and it is extremely sensitive to the overheads of the actual calls. So I made a point to check the pre-conditions before calling them.

albertnetymk: > the overheads of the actual calls

I was expecting the same perf, since the compiler can see everything in this scope.

After hiding checks from the this level, I see the same number of `call` instructions in `MemAllocator::allocate`, where everything is inlined into. Ofc, still needs full perf-eval.

shipilev:
I force-pushed because there were lots of changes in the related code and the merge was exceedingly tedious. The `MultiArrayAlloc` benchmarks improve, I am going to post more thorough benchmark results next week. Meanwhile, @iklam, do you want to give this a spin with [JDK-8310823](https://bugs.openjdk.org/browse/JDK-8310823) prototype? I think it would be sensitive to this change as well.

iklam:
> I force-pushed because there were lots of changes in the related code and the merge was exceedingly tedious. The `MultiArrayAlloc` benchmarks improve, I am going to post more thorough benchmark results next week. Meanwhile, @iklam, do you want to give this a spin with [JDK-8310823](https://bugs.openjdk.org/browse/JDK-8310823) prototype? I think it would be sensitive to this change as well.

I had one variant of my  [JDK-8310823](https://bugs.openjdk.org/browse/JDK-8310823) prototype that avoids MemAllocator altogether (calling `CollectedHeap::mem_allocate` directly). There was a very minor benefit. See https://github.com/openjdk/jdk/pull/14520#issuecomment-1650961800

PR #14025, opened 2023-05-17
8308252: Refactor line-by-line file reading code

iklam:
I extracted the `get_line()` code from `CompileReplay` and put it in a utility class so that it can be used by `ClassListParser` as well. A few notable changes:

- Simplified the API
- Changed the buffer size to a size_t
- Added size overflow and OOM checks
- Brought over the `fdopen` logic from `ClassListParser` for handling long path names on Windows. (I don't know how valid this is nowadays, but I don't want to drop it in a refactoring PR).

---------
### Progress

### Issue
 * [JDK-8308252](https://bugs.openjdk.org/browse/JDK-8308252): Refactor line-by-line file reading code

### Reviewers
 * [David Holmes](https://openjdk.org/census#dholmes) (@dholmes-ora - **Reviewer**) ⚠️ Review applies to [b8a0f31e](https://git.openjdk.org/jdk/pull/14025/files/b8a0f31e3dda2cfd550941bc406b5861562a7efa)
 * [Thomas Stuefe](https://openjdk.org/census#stuefe) (@tstuefe - **Reviewer**) ⚠️ Review applies to [6530758a](https://git.openjdk.org/jdk/pull/14025/files/6530758ab7bb04a61a9c9c867810a5b6aed71838)
 * [Calvin Cheung](https://openjdk.org/census#ccheung) (@calvinccheung - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14025/head:pull/14025` \
`$ git checkout pull/14025`

`$ git checkout pull/14025` \
`$ git pull https://git.openjdk.org/jdk.git pull/14025/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14025`

`$ git pr show -t 14025`

</details>

<a href="https://git.openjdk.org/jdk/pull/14025.diff">https://git.openjdk.org/jdk/pull/14025.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14025#issuecomment-1550658581)

### DIFF
--- a/src/hotspot/share/cds/classListParser.cpp
+++ b/src/hotspot/share/cds/classListParser.cpp
@@ -58,11 +58,11 @@ ClassListParser::ClassListParser(const char* file, ParseMode parse_mode)
   : _reader(file), _id2klass_table(INITIAL_TABLE_SIZE, MAX_TABLE_SIZE) {
   log_info(cds)("Parsing %s%s", file,
                 (parse_mode == _parse_lambda_forms_invokers_only) ? " (lambda form invokers only)" : "");
-  _classlist_file = file;
   if (!_reader.is_opened()) {
     char errmsg[JVM_MAXPATHLEN];
     os::lasterror(errmsg, JVM_MAXPATHLEN);
-    vm_exit_during_initialization("Loading classlist failed", errmsg); // FIXME
+    log_error(cds)("Loading classlist %s failed: %s", _reader.filename(), errmsg);
+    MetaspaceShared::unrecoverable_writing_error();
   }
   _line_no = 0;
   _interfaces = new (mtClass) GrowableArray<int>(10, mtClass);
@@ -418,7 +418,7 @@ void ClassListParser::error(const char* msg, ...) {
 
   jio_fprintf(defaultStream::error_stream(),
               "An error has occurred while processing class list file %s %d:%d.\n",
-              _classlist_file, _line_no, (error_index + 1));
+              _reader.filename(), _line_no, (error_index + 1));
   jio_vfprintf(defaultStream::error_stream(), msg, ap);
 
   if (_line_len <= 0) {
--- a/src/hotspot/share/cds/classListParser.cpp
+++ b/src/hotspot/share/cds/classListParser.cpp
@@ -58,10 +58,8 @@ ClassListParser::ClassListParser(const char* file, ParseMode parse_mode)
   : _reader(file), _id2klass_table(INITIAL_TABLE_SIZE, MAX_TABLE_SIZE) {
   log_info(cds)("Parsing %s%s", file,
                 (parse_mode == _parse_lambda_forms_invokers_only) ? " (lambda form invokers only)" : "");
-  if (!_reader.is_opened()) {
-    char errmsg[JVM_MAXPATHLEN];
-    os::lasterror(errmsg, JVM_MAXPATHLEN);
-    log_error(cds)("Loading classlist %s failed: %s", _reader.filename(), errmsg);
+  if (!_reader.is_open()) {
+    log_error(cds)("Loading classlist %s failed: %s", _reader.filename(), os::strerror(_reader.last_errno()));
     MetaspaceShared::unrecoverable_writing_error();
   }
   _line_no = 0;
--- a/src/hotspot/share/cds/classListParser.cpp
+++ b/src/hotspot/share/cds/classListParser.cpp
@@ -54,23 +54,15 @@
 volatile Thread* ClassListParser::_parsing_thread = nullptr;
 ClassListParser* ClassListParser::_instance = nullptr;
 
-ClassListParser::ClassListParser(const char* file, ParseMode parse_mode) : _id2klass_table(INITIAL_TABLE_SIZE, MAX_TABLE_SIZE) {
+ClassListParser::ClassListParser(const char* file, ParseMode parse_mode)
+  : _reader(file), _id2klass_table(INITIAL_TABLE_SIZE, MAX_TABLE_SIZE) {
   log_info(cds)("Parsing %s%s", file,
                 (parse_mode == _parse_lambda_forms_invokers_only) ? " (lambda form invokers only)" : "");
   _classlist_file = file;
-  _file = nullptr;
-  // Use os::open() because neither fopen() nor os::fopen()
-  // can handle long path name on Windows.
-  int fd = os::open(file, O_RDONLY, S_IREAD);
-  if (fd != -1) {
-    // Obtain a File* from the file descriptor so that fgets()
-    // can be used in parse_one_line()
-    _file = os::fdopen(fd, "r");
-  }
-  if (_file == nullptr) {
+  if (!_reader.is_opened()) {
     char errmsg[JVM_MAXPATHLEN];
     os::lasterror(errmsg, JVM_MAXPATHLEN);
-    vm_exit_during_initialization("Loading classlist failed", errmsg);
+    vm_exit_during_initialization("Loading classlist failed", errmsg); // FIXME
   }
   _line_no = 0;
   _interfaces = new (mtClass) GrowableArray<int>(10, mtClass);
@@ -88,13 +80,16 @@ bool ClassListParser::is_parsing_thread() {
 }
 
 ClassListParser::~ClassListParser() {
-  if (_file != nullptr) {
-    fclose(_file);
-  }
   Atomic::store(&_parsing_thread, (Thread*)nullptr);
   delete _indy_items;
   delete _interfaces;
   _instance = nullptr;
+}
+
+int ClassListParser::parse_classlist(const char* classlist_path, ParseMode parse_mode, TRAPS) {
+  ResourceMark rm(THREAD);
+  ClassListParser parser(classlist_path, parse_mode);
+  return parser.parse(THREAD); // returns the number of classes loaded.
 }
 
 int ClassListParser::parse(TRAPS) {
@@ -165,7 +160,7 @@ int ClassListParser::parse(TRAPS) {
 
 bool ClassListParser::parse_one_line() {
   for (;;) {
-    if (fgets(_line, sizeof(_line), _file) == nullptr) {
+    if ((_line = _reader.get_line()) == nullptr) {
       return false;
     }
     ++ _line_no;
--- a/src/hotspot/share/utilities/lineReader.cpp
+++ b/src/hotspot/share/utilities/lineReader.cpp
@@ -30,20 +30,20 @@
 #include <errno.h>
 
 LineReader::LineReader(const char* filename, size_t initial_length) :
-  _filename(os::strdup(filename)), _stream(nullptr), _errno(0),
+  _filename(os::strdup(filename)), _stream(nullptr), _last_errno(0),
   _buffer_length(initial_length), _buffer(nullptr)
 {
   // Use os::open() because neither fopen() nor os::fopen()
   // can handle long path name on Windows. HMM, is this still valid today???
   int fd = os::open(_filename, O_RDONLY, S_IREAD);
   if (fd == -1) {
-    _errno = errno;
+    _last_errno = errno;
   } else {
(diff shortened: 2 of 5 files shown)


src/hotspot/share/utilities/lineReader.cpp
```diff
@@ -0,0 +1,101 @@
...
+void LineReader::close() {
+  if (_stream != nullptr) {
+    fclose(_stream);
+    _stream = nullptr;
+  }
+}
+
+// Returns nullptr if we have reached EOF.
+// \n is treated as the line separator
+// All occurrences of \r are stripper.
```
dholmes-ora, line 64: s/stripper/stripped/

iklam: Removed.

src/hotspot/share/utilities/lineReader.hpp
```diff
@@ -0,0 +1,51 @@
...
+public:
+  LineReader(const char* filename);
+  ~LineReader();
+
+  bool is_opened() const {
+    return _stream != nullptr;
+  }
+  const char* filename() const { return _filename; }
+  char* get_line();
+  void close();
```
calvinccheung, line 48: The `close()` doesn't need to be public.

src/hotspot/share/utilities/lineReader.cpp
```diff
@@ -0,0 +1,101 @@
...
+ */
+
+#include "precompiled.hpp"
+#include "memory/allocation.inline.hpp"
+#include "runtime/os.hpp"
+#include "utilities/lineReader.hpp"
+
+LineReader::LineReader(const char* filename) : _filename(filename), _stream(nullptr) {
+  // Use os::open() because neither fopen() nor os::fopen()
+  // can handle long path name on Windows. HMM, is this still valid today???
```
dholmes-ora, line 32: It is not clear to me that any of these API's avoid the MAX_PATH 260 character limitation as the only way around that is to use `\\?\` UNC path naming - which we don't appear to do anywhere. On Windows 10 and later if  the machine/user is configured for long paths by default then the `FindFirstFile` used by `os::open` may avoid the 260 limit, but only if it actually maps to the underlying unicode version of the method.
Anyway something to follow up on in a separate RFE I think.

src/hotspot/share/utilities/lineReader.hpp
```diff
@@ -0,0 +1,51 @@
...
+class LineReader : public StackObj {
+  const char* _filename;
+  FILE* _stream;
+  size_t _buffer_length;
+  char* _buffer;
+public:
+  LineReader(const char* filename);
+  ~LineReader();
+
+  bool is_opened() const {
```
dholmes-ora, line 43: Nit: I'd probably go for `is_open()`

src/hotspot/share/utilities/lineReader.cpp
```diff
@@ -0,0 +1,101 @@
...
+    if (_stream == nullptr) {
+      ::close(fd);
+    } else {
+      // fd will be closed by fclose(_stream)
+    }
+  } else {
+    _stream = nullptr;
+  }
+
+  _buffer_length = 32;
```
dholmes-ora, line 47: This default buffer size may work well for the `ciReplay` case but for a general utility there should probably be a way to set the initial buffer length to avoid unnecessary resizing.

src/hotspot/share/utilities/lineReader.cpp
```diff
@@ -0,0 +1,101 @@
...
+// All occurrences of \r are stripped.
+char* LineReader::get_line() {
+  if (_stream == nullptr) {
+    return nullptr;
+  }
+  size_t buffer_pos = 0;
+  int c;
+  while ((c = getc(_stream)) != EOF) {
+    if (buffer_pos + 1 >= _buffer_length) {
+      size_t new_length = _buffer_length * 2;
```
dholmes-ora, line 73: Again for a general utility this simple doubling of size may not be appropriate. For small size it should be okay though. And if we can set the initial size to avoid the need to grow then this is less of an issue.

dholmes-ora (commented):
Looks useful. I wonder if the argument file processing  logic might benefit from this too?

A few comments below.

Thanks.

tstuefe:
> Looks useful. I wonder if the argument file processing logic might benefit from this too?
> 

We could use this too for platform specific stuff, e.g. code reading /proc in os_linux.cpp. But for this function to be truly useful, allocator must be choosable since RAs don't always work. A parameter to define allocation would be nice (malloc or RA).

src/hotspot/share/cds/classListParser.cpp
```diff
@@ -54,23 +54,15 @@
...
-  int fd = os::open(file, O_RDONLY, S_IREAD);
-  if (fd != -1) {
-    // Obtain a File* from the file descriptor so that fgets()
-    // can be used in parse_one_line()
-    _file = os::fdopen(fd, "r");
-  }
-  if (_file == nullptr) {
+  if (!_reader.is_opened()) {
     char errmsg[JVM_MAXPATHLEN];
     os::lasterror(errmsg, JVM_MAXPATHLEN);
```
tstuefe, line 63: _reader should buffer errno after the failing OS call. We should not have to rely on os::lasterror() being called right after whatever OS API failed inside the reader. Neither is os::lasterror() necessary, we can just use os::strerror since reader only uses Posix file APIs.

src/hotspot/share/utilities/lineReader.cpp
```diff
@@ -0,0 +1,101 @@
...
+ * questions.
+ *
+ */
+
+#include "precompiled.hpp"
+#include "memory/allocation.inline.hpp"
+#include "runtime/os.hpp"
+#include "utilities/lineReader.hpp"
+
+LineReader::LineReader(const char* filename) : _filename(filename), _stream(nullptr) {
```
tstuefe, line 30: Maybe strdup the file name to be sure? Up to you. We usually just feed literals.

src/hotspot/share/utilities/lineReader.cpp
```diff
@@ -0,0 +1,101 @@
...
+    // Obtain a File* from the file descriptor so that getc()
+    // can be used in get_line().
+    _stream = os::fdopen(fd, "r");
+    if (_stream == nullptr) {
+      ::close(fd);
+    } else {
+      // fd will be closed by fclose(_stream)
+    }
+  } else {
+    _stream = nullptr;
```
tstuefe, line 44: unnecessary

src/hotspot/share/utilities/lineReader.cpp
```diff
@@ -0,0 +1,101 @@
...
+// Returns nullptr if we have reached EOF.
+// \n is treated as the line separator.
+// All occurrences of \r are stripped.
+char* LineReader::get_line() {
+  if (_stream == nullptr) {
+    return nullptr;
+  }
+  size_t buffer_pos = 0;
+  int c;
+  while ((c = getc(_stream)) != EOF) {
```
tstuefe, line 71: Lets not read individual characters. Lets use fgets or fread() or just plain read(). Preferably the first.

src/hotspot/share/utilities/lineReader.cpp
```diff
@@ -0,0 +1,101 @@
...
+    if (buffer_pos + 1 >= _buffer_length) {
+      size_t new_length = _buffer_length * 2;
+      if (new_length < _buffer_length) {
+        // This could happen on 32-bit. On 64-bit, the VM would have exited
+        // due to OOM before we ever get to here.
```
tstuefe, line 76: This is scary. I don't like a general utility to use half my address space on 32-bit if bad things happen. I would cap the max. buffer size to something sensible, e.g. 1K or 64K.

src/hotspot/share/utilities/lineReader.cpp
```diff
@@ -0,0 +1,101 @@
...
+  if (_stream != nullptr) {
+    fclose(_stream);
+    _stream = nullptr;
+  }
+}
+
+// Returns nullptr if we have reached EOF.
+// \n is treated as the line separator.
+// All occurrences of \r are stripped.
+char* LineReader::get_line() {
```
tstuefe, line 65: I would return const char* here, this is an internal buffer.

tstuefe (changes requested):
What would cool would be a closure that gets each line, optionally with the ability to stop iteration prematurely. Or even better, a lambda that does the same.

src/hotspot/share/utilities/lineReader.cpp
iklam: I'll add an initial size parameter that defaults to 160. That should be good to avoid expansion for most reasonable text input files, yet small enough. My feeling is that you usually don't care and don't know what size to set.

iklam:
> > Looks useful. I wonder if the argument file processing logic might benefit from this too?
> 
> We could use this too for platform specific stuff, e.g. code reading /proc in os_linux.cpp. But for this function to be truly useful, allocator must be choosable since RAs don't always work. So a parameter to define allocation would be nice (malloc or RA).

I want to keep this PR simple and my main goal is to have a correct version of the line reading code. We can expand it to other cases, and add lambdas if it makes sense, in follow-on RFEs.

src/hotspot/share/utilities/lineReader.cpp
iklam: I've reimplemented using fgets().

src/hotspot/share/utilities/lineReader.cpp
iklam: Removed.

src/hotspot/share/utilities/lineReader.cpp
iklam: Done.

src/hotspot/share/cds/classListParser.cpp
iklam: I added `int LineReader::last_errno()`.

src/hotspot/share/utilities/lineReader.cpp
dholmes-ora: So now this is not just a refactoring it is a re-implementation, so now I have to try and understand the reading logic and can't just treat it as known good code. :(

iklam: I can revert to the getc version and implement the fgets version in a separate RFE. I agree that keeping the logic unchanged in refactoring will make things more manageable.

getc is buffered I/O, so there shouldn't be much difference in performance when compared to fgets. And the code affected by this PR doesn't care too much about I/O performance.

iklam: By the way, I think I should take out the handling of _max_buffer_length, which I actually haven't implemented. I don't know what the failure mode should be when the line width is longer than the specified max, and I don't want to write any test case when the max is set to a small number like 256.

So let's keep the existing 2x expansion and crash when OOM (which is the same as the numerous sites that use GrowableArray).

src/hotspot/share/utilities/lineReader.hpp
iklam: I think it's useful to have a `close()` function that can be called proactively before the LineReader is destroyed.

src/hotspot/share/utilities/lineReader.cpp
tstuefe: Okay, pity since that makes the code less reusable for cases I had in mind. For instance, reading and writing out gigantic mapping files. I recently had a customer with a 4TB heap and ZGC, where /proc/pid/maps alone was 20mio lines.

tstuefe (approved):
Okay. Since the aim was to transfer the code as verbatim as possible without improving on it, this is fine.

dholmes-ora (approved):
Seems fine to me. Thanks for re-simplifying.

tstuefe:
Still looks good to me.

iklam:
@rose00 has a alternative proposal for rewriting the buffed file reading code to support a more capable API. I will withdraw this PR for now.

PR #11943, opened 2023-01-11
8304006: jlink should create the jimage file in the native endian for the target platform

jaikiran:
Can I please get a review for this change which proposes to fix the issue reported in https://bugs.openjdk.org/browse/JDK-8206890?

The `jlink` command allows a `--endian` option to specify the byte order in the generated image. Before this change, when such a image was being launched, the code would assume the byte order in the image to be the native order of the host where the image is being launched. That would result in failure to launch java, as noted in the linked issue.

The commit in this PR, changes relevant places to not assume native order and instead determine the byte order by reading the magic bytes in the image file's header content.

A new jtreg test has been added which reproduces the issue and verifies the fix.

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8304006](https://bugs.openjdk.org/browse/JDK-8304006): jlink should create the jimage file in the native endian for the target platform (**Enhancement** - P4)

### Reviewers
 * [Alan Bateman](https://openjdk.org/census#alanb) (@AlanBateman - **Reviewer**)
 * [Mandy Chung](https://openjdk.org/census#mchung) (@mlchung - **Reviewer**) ⚠️ Review applies to [258a62d7](https://git.openjdk.org/jdk/pull/11943/files/258a62d73338d1bfc57d7a3b66dc4ce97936e7c3)
 * [Roger Riggs](https://openjdk.org/census#rriggs) (@RogerRiggs - **Reviewer**) ⚠️ Review applies to [258a62d7](https://git.openjdk.org/jdk/pull/11943/files/258a62d73338d1bfc57d7a3b66dc4ce97936e7c3)

### Contributors
 * Mandy Chung `<mchung@openjdk.org>`

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/11943/head:pull/11943` \
`$ git checkout pull/11943`

Update a local copy of the PR: \
`$ git checkout pull/11943` \
`$ git pull https://git.openjdk.org/jdk.git pull/11943/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 11943`

View PR using the GUI difftool: \
`$ git pr show -t 11943`

</details>
<details><summary>Using diff file</summary>

[...]

PR #14027, opened 2023-05-17
8308246: PPC64le build broken after JDK-8304913

TheRealMDoerr:
PPC64le needs to get recognized as PPC64. Otherwise, jlink doesn't recognize the platform and throws a PluginException: ModuleTarget is malformed: No enum constant jdk.internal.util.Architecture.PPC64LE.

### DIFF
--- a/src/jdk.jlink/share/classes/jdk/tools/jlink/internal/Platform.java
+++ b/src/jdk.jlink/share/classes/jdk/tools/jlink/internal/Platform.java
@@ -50,8 +50,9 @@ public record Platform(OperatingSystem os, Architecture arch) {
         OperatingSystem os = OperatingSystem.valueOf(osName.toUpperCase(Locale.ROOT));
 
         archName = platformString.substring(index + 1);
-        // Alias architecture "amd64" to "X64"
-        archName = archName.replace("amd64", "X64");
+        // Alias architecture "amd64" to "X64" and "ppc64le" to "ppc64"
+        archName = archName.replace("amd64", "X64")
+                           .replace("ppc64le", "ppc64");
         Architecture arch = Architecture.valueOf(archName.toUpperCase(Locale.ROOT));
 
         return new Platform(os, arch);


src/jdk.jlink/share/classes/jdk/tools/jlink/internal/Platform.java
```diff
@@ -50,8 +50,9 @@ public static Platform parsePlatform(String platformString) {
         OperatingSystem os = OperatingSystem.valueOf(osName.toUpperCase(Locale.ROOT));
 
         archName = platformString.substring(index + 1);
-        // Alias architecture "amd64" to "X64"
-        archName = archName.replace("amd64", "X64");
+        // Alias architecture "amd64" to "X64" and "ppc64le" to "ppc64"
+        archName = archName.replace("amd64", "X64")
+                           .replace("ppc64le", "ppc64");
```
shipilev, line 55: I'd prefer it to be:

```suggestion
        // Alias architecture names, if needed
        archName = archName.replace("amd64", "X64");
        archName = archName.replace("ppc64le", "PPC64");
```

So that other arches would add here more naturally.

(Also convert to upper-case right away, to make the subsequent `archName.toUpperCase` no-op).

shipilev (approved):
Looks fine, but I have a suggestion.

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/Platform.java
AlanBateman, line 55: Is linux-ppc64 big endian? If so, a module with a ModuleTarget value of linux-ppc64le can't be linked with a module that has ModuleTarget value of linux-ppc64.  I'm not say the aliasing here is wrong but I am curious if attempts to link will fail.

TheRealMDoerr: Yes, linux-ppc64 is big endian. The 2 flavors are completely incompatible. linux-ppc64le is basically a new platform.

TheRealMDoerr: Makes sense. Thanks for the review!

MBaesken (approved):

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/Platform.java
AlanBateman: > Yes, linux-ppc64 is big endian. The 2 flavors are completely incompatible. linux-ppc64le is basically a new platform. Endianness should actually get checked in addition to the archName.

Okay, I was only asking because the ModuleTarget attribute has the OS and arch, the endianness is derived from those right now. Maybe in the future there can be more about the architecture and processor features (e.g. soft vs. hard float).  So mulling over this change and wondering what happens if there is cross linking involve these two platforms.

TheRealMDoerr:
Thanks for reviewing and feedback! I'm repairing the build:
/integrate

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/Platform.java
RogerRiggs: Hmm,  most java bytecode (classfiles) aren't going to care about the endian-ness; in int is an int, etc.
Runtime calls to extract int from byte[] is going to check a runtime flag; possibly static and optimization by HotSpot, etc.
The Architecture enum was intending to capture essential build time attributes.
If there is a need for more static attributes of a build, the current API doesn't cover that.

AlanBateman: > Hmm, most java bytecode (classfiles) aren't going to care about the endian-ness; in int is an int, etc.
> Runtime calls to extract int from byte[] is going to check a runtime flag; possibly static and optimization by HotSpot, etc.
> The Architecture enum was intending to capture essential build time attributes.
> If there is a need for more static attributes of a build, the current API doesn't cover that.

The image writer needs to be told the endianness of the target platform, usually the native order but can be overridden. As part of pull/11943, a new resource file is added to map the value of the ModuleTarget class file attribute (essentially $os-$arch) to the endianness to help cross linking.  I don't yet see how this will work with the changes in the this PR.

TheRealMDoerr: Thanks for the hint! I've added a comment to https://github.com/openjdk/jdk/pull/11943 .

PR #14030, opened 2023-05-17
8308281: Java snippets in the FFM API need to be updated

minborg:
As the API has improved over the recent releases, not all `{@snippet ..}` sections have been kept in sync.

This PR suggests all snippets used should be verified against real code that is placed in a new `snippet-files` folder and erroneous snippets are updated.

In this PR, it is suggested duplicating code in the `Snippets.java` class and in the JavaDocs. The benefit of this is that code is directly visible in the code and not only in the generated javadoc.

Another thing to think about is if there should be on single `Snippets.java` class or separate ones for each FFM class.

---------
### Progress

### Issue
 * [JDK-8308281](https://bugs.openjdk.org/browse/JDK-8308281): Java snippets in the FFM API need to be updated

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - **Reviewer**) ⚠️ Review applies to [06d04b24](https://git.openjdk.org/jdk/pull/14030/files/06d04b24e548ddbfbe641c7a8994e19130c3d0da)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14030/head:pull/14030` \
`$ git checkout pull/14030`

`$ git checkout pull/14030` \
`$ git pull https://git.openjdk.org/jdk.git pull/14030/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14030`

`$ git pr show -t 14030`

</details>

<a href="https://git.openjdk.org/jdk/pull/14030.diff">https://git.openjdk.org/jdk/pull/14030.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14030#issuecomment-1551255788)


mcimadamore (commented):
The fixes all are great - thanks for going over it. I'm not sure about the snippet-file folder - e.g. if we're not using it, is there value in having it? I suppose (but I might be wrong) that the IDE cannot perform refactors on such a file, so it will become yet another thing that might go out of date?

src/java.base/share/classes/java/lang/foreign/Arena.java
```diff
@@ -77,12 +76,12 @@
  *
  * {@snippet lang = java:
  * MemorySegment segment = null;
- * try (Arena arena = Arena.ofConfined()) {
+ * try (Arena arena = Arena.ofConfined()) { // @highlight regex='ofConfined()'
  *     segment = arena.allocate(100);
- *     ...
+ *     // ...
```
mcimadamore, line 81: Do we need to add comments here? I believe that, even with hybrid snippets there are ways to handle this (I remember it being discussed)

mcimadamore: Same for all other occurrences of `// ...`. Let's address them when we have to (if we have to).

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -249,12 +249,12 @@
  * Now let's create a method handle for the comparator method defined above:
  *
  * {@snippet lang = java:
- * FunctionDescriptor comparDesc = FunctionDescriptor.of(JAVA_INT,
+ * FunctionDescriptor compareDesc = FunctionDescriptor.of(JAVA_INT,
```
mcimadamore, line 252: note - the use of `compar` here was deliberate, as it matched the name of the function pointer param in qsort - but I'll leave to you (e.g. in case that is too subtle)

mcimadamore (approved):
Looks good - I've left some minor optional comments

PR #14037, opened 2023-05-17
8308293: A linker should expose the layouts it supports

mcimadamore:
Some text here

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/Linker.java
+++ b/src/java.base/share/classes/java/lang/foreign/Linker.java
@@ -202,15 +202,29 @@ import java.util.stream.Stream;
  * </tbody>
  * </table></blockquote>
  * <p>
- * All the native linker implementations limit the function descriptors that they support to those that contain
- * either {@linkplain ValueLayout value layouts} that are also {@linkplain #canonicalLayouts() canonical layouts},
- * or {@linkplain GroupLayout group layouts} that are derived according to the following rules:
+ * All the native linker implementations can only operate on a subset of memory layouts, called <em>supported layouts</em>.
+ * A layout {@code L} is supported iff:
+ * <ul>
+ * <li>{@code L} is a value layout {@code V} and {@code V.withoutName()} is a canonical layout</li>
+ * <li>{@code L} is an address layout {@code A} and {@code A.withoutTargetLayout().withoutName()} is a canonical layout</li>
+ * <li>{@code L} is a sequence layout {@code S} and all the following conditions hold:
  * <ol>
- * <li>The alignment constraint of the group layout is set to its <a href="MemoryLayout.html#layout-align">natural alignment</a></li>
- * <li>The size of the group layout is a multiple of its alignment constraint, and</li>
- * <li>The group layout does not contain padding other than what is strictly required to align its non-padding layout elements,
- * or to satisfy constraint 1</li>
+ * <li>the alignment constraint of {@code S} is set to its <a href="MemoryLayout.html#layout-align">natural alignment</a>, and</li>
+ * <li>{@code S.elementLayout()} is a supported layout.</li>
  * </ol>
+ * </li>
+ * <li>{@code L} is a group layout {@code G} and all the following conditions hold:
+ * <ol>
+ * <li>the alignment constraint of {@code G} is set to its <a href="MemoryLayout.html#layout-align">natural alignment</a>;</li>
+ * <li>the size of {@code G} is a multiple of its alignment constraint;</li>
+ * <li>each member layout in {@code G.memberLayouts()} is either a padding layout or a supported layout, and</li>
+ * <li>{@code G} does not contain padding other than what is strictly required to align its non-padding layout elements, or to satisfy (2).</li>
+ * </ol>
+ * </li>
+ * </ul>
+ *
+ * Native linkers only support function descriptors whose argument/return layouts are supported layouts and
+ * are not sequence layouts.
  *
  * <h3 id="function-pointers">Function pointers</h3>
  *
--- a/src/java.base/share/classes/java/lang/foreign/Linker.java
+++ b/src/java.base/share/classes/java/lang/foreign/Linker.java
@@ -31,12 +31,11 @@ import jdk.internal.foreign.abi.CapturableState;
 import jdk.internal.foreign.abi.SharedUtils;
 import jdk.internal.javac.PreviewFeature;
 import jdk.internal.reflect.CallerSensitive;
-import jdk.internal.reflect.Reflection;
 
 import java.lang.invoke.MethodHandle;
 import java.nio.ByteOrder;
+import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.Set;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
@@ -59,6 +58,11 @@ import java.util.stream.Stream;
  * <li>A linker allows foreign functions to call Java method handles,
  * via the generation of {@linkplain #upcallStub(MethodHandle, FunctionDescriptor, Arena, Option...) upcall stubs}.</li>
  * </ul>
+ * A linker provides a way to lookup up the <em>canonical layouts</em> associated with the data types used by the ABI.
+ * For example, the canonical layout for the C {@code size_t} type is equal to {@link ValueLayout#JAVA_LONG}. The canonical
+ * layouts supported by a linker are exposed via the {@link #canonicalLayouts()} method, which returns a map from
+ * ABI type names to canonical layouts.
+ * <p>
  * In addition, a linker provides a way to look up foreign functions in libraries that conform to the ABI. Each linker
  * chooses a set of libraries that are commonly used on the OS and processor combination associated with the ABI.
  * For example, a linker for Linux/x64 might choose two libraries: {@code libc} and {@code libm}. The functions in these
@@ -103,11 +107,12 @@ import java.util.stream.Stream;
  * defines the layouts associated with the parameter types and return type (if any) of the C function.
  * <p>
  * Scalar C types such as {@code bool}, {@code int} are modelled as {@linkplain ValueLayout value layouts}
- * of a suitable carrier. The mapping between a scalar type and its corresponding layout is dependent on the ABI
- * implemented by the native linker. For instance, the C type {@code long} maps to the layout constant
- * {@link ValueLayout#JAVA_LONG} on Linux/x64, but maps to the layout constant {@link ValueLayout#JAVA_INT} on
- * Windows/x64. Similarly, the C type {@code size_t} maps to the layout constant {@link ValueLayout#JAVA_LONG}
- * on 64-bit platforms, but maps to the layout constant {@link ValueLayout#JAVA_INT} on 32-bit platforms.
+ * of a suitable carrier. The {@linkplain #canonicalLayouts() mapping} between a scalar type and its corresponding
+ * canonical layout is dependent on the ABI implemented by the native linker. For instance, the C type {@code long}
+ * maps to the layout constant {@link ValueLayout#JAVA_LONG} on Linux/x64, but maps to the layout constant
+ * {@link ValueLayout#JAVA_INT} on Windows/x64. Similarly, the C type {@code size_t} maps to the layout constant
+ * {@link ValueLayout#JAVA_LONG} on 64-bit platforms, but maps to the layout constant {@link ValueLayout#JAVA_INT} on
+ * 32-bit platforms.
  * <p>
  * Composite types are modelled as {@linkplain GroupLayout group layouts}. More specifically, a C {@code struct} type
  * maps to a {@linkplain StructLayout struct layout}, whereas a C {@code union} type maps to a {@link UnionLayout union
@@ -198,14 +203,13 @@ import java.util.stream.Stream;
  * </table></blockquote>
  * <p>
  * All the native linker implementations limit the function descriptors that they support to those that contain
- * only so-called <em>canonical</em> layouts. A canonical layout has the following characteristics:
+ * either {@linkplain ValueLayout value layouts} that are also {@linkplain #canonicalLayouts() canonical layouts},
+ * or {@linkplain GroupLayout group layouts} that are derived according to the following rules:
  * <ol>
- * <li>Its alignment constraint is set to its <a href="MemoryLayout.html#layout-align">natural alignment</a></li>
- * <li>If it is a {@linkplain ValueLayout value layout}, its {@linkplain ValueLayout#order() byte order} is
- * the {@linkplain ByteOrder#nativeOrder() native byte order}.
- * <li>If it is a {@linkplain GroupLayout group layout}, its size is a multiple of its alignment constraint, and</li>
- * <li>It does not contain padding other than what is strictly required to align its non-padding layout elements,
- * or to satisfy constraint 3</li>
+ * <li>The alignment constraint of the group layout is set to its <a href="MemoryLayout.html#layout-align">natural alignment</a></li>
+ * <li>The size of the group layout is a multiple of its alignment constraint, and</li>
+ * <li>The group layout does not contain padding other than what is strictly required to align its non-padding layout elements,
+ * or to satisfy constraint 1</li>
  * </ol>
  *
  * <h3 id="function-pointers">Function pointers</h3>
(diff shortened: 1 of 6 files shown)


src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -573,6 +591,22 @@ static Linker nativeLinker() {
      */
     SymbolLookup defaultLookup();
 
+    /**
+     * {@return a mapping between the names of data types used by the ABI implemented by this linker and their
```
mcimadamore, line 595: Much of the verbiage here is carried over from `defaultLookup` as we need to do the usual dance of saying that the set of returned types is not specified, but should be (a) sensible and (b) stable.

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -197,16 +202,29 @@
  * </tbody>
  * </table></blockquote>
  * <p>
- * All the native linker implementations limit the function descriptors that they support to those that contain
- * only so-called <em>canonical</em> layouts. A canonical layout has the following characteristics:
+ * All the native linker implementations can only operate on a subset of memory layouts, called <em>supported layouts</em>.
```
mcimadamore, line 205: I revamped this section as I realized that what we had did not cover things in the recursive case - e.g. a struct layout is only supported if it contains other supported layouts. This new text should hopefully capture everything in a more mathematical form.

mcimadamore:
javadoc: https://cr.openjdk.org/~mcimadamore/jdk/8308293/v1/javadoc/java.base/module-summary.html
specdiff: https://cr.openjdk.org/~mcimadamore/jdk/8308293/v1/specdiff_out/overview-summary.html

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -197,16 +202,29 @@
...
+ * <li>{@code L} is a group layout {@code G} and all the following conditions hold:
  * <ol>
- * <li>Its alignment constraint is set to its <a href="MemoryLayout.html#layout-align">natural alignment</a></li>
- * <li>If it is a {@linkplain ValueLayout value layout}, its {@linkplain ValueLayout#order() byte order} is
- * the {@linkplain ByteOrder#nativeOrder() native byte order}.
- * <li>If it is a {@linkplain GroupLayout group layout}, its size is a multiple of its alignment constraint, and</li>
- * <li>It does not contain padding other than what is strictly required to align its non-padding layout elements,
- * or to satisfy constraint 3</li>
+ * <li>the alignment constraint of {@code G} is set to its <a href="MemoryLayout.html#layout-align">natural alignment</a>;</li>
+ * <li>the size of {@code G} is a multiple of its alignment constraint;</li>
```
PaulSandoz, line 219: Do you think this is a constraint that will hold across all linker implementations?

mcimadamore: All "native linkers" as the text says. Other linkers (e.g. not for C) might obey other rules. These rules are basically constraining  struct layouts to what can come out of a native compiler in the absence of pragma pack directives.

PaulSandoz (commented):
I wonder if we need to be stronger on the compatibility requirements for the supported canonical layouts. It seems we can make stronger claims than symbols made available by the defaultLookup because the ABI supported by the native linker will make strong (specification) claims.

So maybe this comes down to the linker supporting a subset ABI's data types, and that subset might increase over time, but never decrease? In this respect could we present a table for each supported linker ABI listing the ABI type and its canonical layout type? (in practice it might just be one table with noted adjustments.)

Then there is the possibility that a linker might change the layout corresponding to a data type. Ideally the linker support the prior layout *and* the new layout. I don't think this will arise with the current support set of supported ABI data types, but it might if we choose to add support for say the optional `__int128` data type prior to the platform adding support for a primitive value class of Int128?

[1] https://gitlab.com/x86-psABIs/x86-64-ABI

src/java.base/share/classes/java/lang/foreign/Linker.java
PaulSandoz: Ah, i see now, i missed the relevance of that term, introduced earlier in non-modified text.

mcimadamore:
> So maybe this comes down to the linker supporting a subset ABI's data types, and that subset might increase over time, but never decrease? In this respect could we present a table for each supported linker ABI listing the ABI type and its canonical layout type? (in practice it might just be one table with noted adjustments.)

I see what you mean and I'm not sure about this. On the one hand, having a set of "trusted" type names would be handy - but I don't know how much commitment we want to put in there? I'm also a bit skeptical at listing all possible ABIs, since I suspect the set of supported platforms will change quickly.

Is what you are after some kind of guarantee of "at least these type names will be available" ?

As for a linker possibly having multiple different layouts for the same ABI type, that is true, and, in a way, already the case with ValueLayout.OfChar/ValueLayout.OfShort. I worked around that by using different type names - e.g. `int16_t` and `char16_t`. 

For more exotic types which might be modeled initially opaquely with MemorySegment, and later on with some other ValueLayout.OfFooBar, I believe we'd need to provide a way to go from the opaque layout to the less opaque one.

The other option would be to admit that a single ABI type can map to multiple layouts, and have `Map<String, List<MemoryLayout>>`. It seemed a bit on the convoluted side of things (esp. given that we can get away without it, at least in this patch), but if you think that would be more robust, I can change that.

PaulSandoz:
Here's the crux of what i am wondering about. Can we specify native linker support for a subset of the System V Application Binary Interface (e.g., LP64 and ILP32 programming models for all non-optional scalar types, sequences of, and groups of) such that a developer can write code using the FFM API and it will work across all JDK implementations supporting that native linker?

AFAICT the closest we have to that is the table in the Linker doc, and that table references C type names. Perhaps we can use C type name as the ABI type name for the System V Application Binary Interface? (literally copy the name used in Figure 3.1 C column of the ABI specification).

Then can do we the same and specify the equivalent native linker support for ABIs of Windows 64/32 and ARM?

mcimadamore:
> Here's the crux of what i am wondering about. Can we specify native linker support for a subset of the System V Application Binary Interface (e.g., LP64 and ILP32 programming models for all non-optional scalar types, sequences of, and groups of) such that a developer can write code using the FFM API and it will work across all JDK implementations supporting that native linker?
> 
> AFAICT the closest we have to that is the table in the Linker doc, and that table references C type names. Perhaps we can use C type name as the ABI type name for the System V Application Binary Interface? (literally copy the name used in Figure 3.1 C column of the ABI specification).
> 
> Then can do we the same and specify the equivalent native linker support for ABIs of Windows 64/32 and ARM?

Consider that, at the time of writing we support (or might support soon):

* Sysv (Linux and MacOS)
* Aarch64(Linux, MacOs, Windows) - these have actually some ABI differences (e.g. variadic calls)
* PowerPC (which might come in big/little endian flavors)
* RiscV

That's quite a lot of ABIs and tables to have.

Also, if we wanted to tighten up the spec a little bit, what the user cares about is some minimum guarantees about the supported ABI types across platforms. E.g. you don't want a table-per-ABI, precisely because you want to know (I think): "if I call `linker.canonicalLayout().get("int")`" is this call guaranteed to succeed?

So, pulling on the string, IMHO we should:

* define which common subset of C types we support (e.g. list the C type names) - probably the one we use now is a good set;
* Give an example, on x64, of how these types might be realized using layouts (e.g. current table)

More pulling, the `char16_t` type is probably *not* defined on all ABIs (e.g. I can't find it here [1]). Which seems to suggest that perhaps canonical layout should just return mapping from string to lists of layouts, even if inconvenient?

PaulSandoz:
I agree focusing on a subset of C types is the way to go. That avoids the unnecessary verbosity of many tables, and we can enumerate the types differing by data model (e.g., LP64 and ILP32). 

As a developer i would like to know for all C-based native linkers (which is all native linkers? what else would they be based on?) if:

1. the C `int` type etc is supported; and
2. whether i can use the `JAVA_INT` layout in my function descriptors, and know it corresponds to a C `int`. 

It seems obvious that i should be able to but AFAICT the specification is more example based, so it's not clear to me if different Java implementations can deviate in such behaviour.

Requiring the use of the C type names in the canonical mapping does help, because then i can more directly ask the C-based linker "Hey what's your canonical layout for the C `char` type?" .

I don't see `char16_t` defined either and i believe that type is an alias for `uint_least16_t`, which is a type of *at least* 16 bits. We need to use a C type of `unsigned short` or `uint16_t`:

- Java `ValueLayout.JAVA_SHORT` <-> C `short`
- Java `ValueLayout.JAVA_CHAR` <-> C `unsigned short`

?

For canonical type names we may want to prefer types specified by the C language over those defined by the C library in standard headers?

mcimadamore:
> I don't see `char16_t` defined either and i believe that type is an alias for `uint_least16_t`, which is a type of _at least_ 16 bits. We need to use a C type of `unsigned short` or `uint16_t`:
> 
>     * Java `ValueLayout.JAVA_SHORT` <-> C `short`
> 
>     * Java `ValueLayout.JAVA_CHAR` <-> C `unsigned short`
> 
> 
> ?
Good point on "unsigned short". 
> 
> For canonical type names we may want to prefer types specified by the C language over those defined by the C library in standard headers?

Yes, I think better to stick with standard C types - IMHO with the exception of size_t which is very ubiquitous.

mcimadamore:
Updated javadoc and specdiffs:
javadoc: https://cr.openjdk.org/~mcimadamore/jdk/8308293/v2/javadoc/java.base/java/lang/foreign/Linker.html
specdiff: https://cr.openjdk.org/~mcimadamore/jdk/8308293/v2/specdiff_out/overview-summary.html

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -59,6 +58,11 @@
  * <li>A linker allows foreign functions to call Java method handles,
  * via the generation of {@linkplain #upcallStub(MethodHandle, FunctionDescriptor, Arena, Option...) upcall stubs}.</li>
  * </ul>
+ * A linker provides a way to lookup up the <em>canonical layouts</em> associated with the data types used by the ABI.
+ * For example, the canonical layout for the C {@code size_t} type is equal to {@link ValueLayout#JAVA_LONG}. The canonical
```
PaulSandoz, line 62:
```suggestion
 * For example, the canonical layout for the C {@code size_t} type is equal to {@link ValueLayout#JAVA_LONG} on 64-bit platforms. The canonical
```
?

PaulSandoz (commented):
This look much better. Can we strengthen the specification of `canonicalLayouts` in accordance with the class specification

src/java.base/share/classes/java/lang/foreign/Linker.java
mcimadamore: You are correct in calling this out. I think this should be spelled out more (similarly to what we do for default lookup) since we're still in the "general" linker section. E.g.

```
A linker provides a way to lookup up the <em>canonical layouts</em> associated with the data types used by the ABI.
For example, a linker implementing the C ABI might chose to provide a canonical layout for the C {@code size_t} type. On 64-bit platforms, this canonical layout might be equal to {@link ValueLayout#JAVA_LONG}. The canonical
layouts supported by a linker are exposed via the {@link #canonicalLayouts()} method, which returns a map from
ABI type names to canonical layouts.
```

mcimadamore:
> This look much better. Can we strengthen the specification of `canonicalLayouts` in accordance with the class specification

We can't do more in that method javadoc,  think, as that has to be general enough for all linkers. I think the rules set up in that method javadoc are good - e.g. the set of layouts should be stable (both in terms of names and layout types).

What we can do is to sprinkle some wording in the `nativeLinker` factory - e.g. `the native linker is guaranteed to provide canonical layouts for basic C types <link to the class section>`.

PaulSandoz:
> > This look much better. Can we strengthen the specification of `canonicalLayouts` in accordance with the class specification
> 
> We can't do more in that method javadoc, think, as that has to be general enough for all linkers. I think the rules set up in that method javadoc are good - e.g. the set of layouts should be stable (both in terms of names and layout types).
> 
> What we can do is to sprinkle some wording in the `nativeLinker` factory - e.g. `the native linker is guaranteed to provide canonical layouts for basic C types <link to the class section>`.

Yes, that's better.

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -573,6 +610,22 @@ static Linker nativeLinker() {
      */
     SymbolLookup defaultLookup();
 
+    /**
+     * {@return a mapping between the names of data types used by the ABI implemented by this linker and their
```
minborg, line 614: I think we should state we return "an unmodifiable mapping".

minborg (approved):
LGTM. Are there any additional types we might consider apart from the basic ones and `size_t`? Maybe one for an address pointing at `errno`?

mcimadamore:
> LGTM. Are there any additional types we might consider apart from the basic ones and `size_t`? Maybe one for an address pointing at `errno`?

I think there is a certain gravitational pull towards keeping the set of guaranteed canonical layouts as minimal as possible. In that sense pointer to errno seems not to meet the bar IMHO. (also note that one could always ask the captureStateLayout, and figure out how to express the errno type from there).

mcimadamore:
@PaulSandoz after thinking some more, it seems a bit ad-hoc to guarantee a canonical for "unsigned short", but not for other unsigned types? Possible alternatives (beside keeping what we have in this PR):

* guarantee also all the other unsigned types (e.g. char, int, long and long long)
* do not guarantee unsigned short - but provide a mapping for it anyways
* do not guarantee unsigned short, and do not provide a mapping for it - this means that JAVA_CHAR will not be usable when linking

What do you think?

PaulSandoz:
On further reflection i think mapping C `unsigned short` only to Java `char` is a misleading. Although Java `char` is an integral type is not really intended to be used generally as an unsigned 16-bit integer, so arguably Java `char` is not as useful a carrier type for native interoperation as Java `short` might be even though it is signed.
Thus i am inclined to remove that mapping.

What if we say something to the effect of:

> unless explicitly declared in the canonical layouts C's unsigned integral types are mapped to the layouts associated with the required C's signed integral types of the same bit sizes.

?

Arguably C `unsigned short` could map to carriers Java `short` or Java `char`, but i am inclined to say the user should cast between Java `short` to `char` in such cases.

FWIW i checked what the FFM API and jextract does today and it maps unsigned C types to signed Java types.

mcimadamore:
> On further reflection i think mapping C `unsigned short` only to Java `char` is a misleading. Although Java `char` is an integral type is not really intended to be used generally as an unsigned 16-bit integer, so arguably Java `char` is not as useful a carrier type for native interoperation as Java `short` might be even though it is signed. Thus i am inclined to remove that mapping.
> 
> What if we say something to the effect of:
> 
> > unless explicitly declared in the canonical layouts C's unsigned integral types are mapped to the layouts associated with the required C's signed integral types of the same bit sizes.
> 
> ?
> 
> Arguably C `unsigned short` could map to carriers Java `short` or Java `char`, but i am inclined to say the user should cast between Java `short` to `char` in such cases.
> 
> FWIW i checked what the FFM API and jextract does today and it maps unsigned C types to signed Java types.



> On further reflection i think mapping C `unsigned short` only to Java `char` is a misleading. Although Java `char` is an integral type is not really intended to be used generally as an unsigned 16-bit integer, so arguably Java `char` is not as useful a carrier type for native interoperation as Java `short` might be even though it is signed. Thus i am inclined to remove that mapping.
> 
> What if we say something to the effect of:
> 
> > unless explicitly declared in the canonical layouts C's unsigned integral types are mapped to the layouts associated with the required C's signed integral types of the same bit sizes.
> 
> ?
> 
> Arguably C `unsigned short` could map to carriers Java `short` or Java `char`, but i am inclined to say the user should cast between Java `short` to `char` in such cases.
> 
> FWIW i checked what the FFM API and jextract does today and it maps unsigned C types to signed Java types.

I tend to agree with your conclusion. And I confirm that we do not use "char" anywhere in jextract. The only "problem" with that approach is that if we go down that path, JAVA_CHAR is no longer a canonical type, so users cannot mention it in function descriptors. Apart from requiring few test updates, I don't see many other problems with it - if one really really wanted the result of a native call to be converted to `char`, a MH adapter can be used. Effectively, what you suggest amount at saying: we do have a JAVA_CHAR layout, which is mostly there for Java interop. But a native linker (which only cares about native interop) doesn't really care much about that. Does that sound good?

mcimadamore:
> Arguably C `unsigned short` could map to carriers Java `short` or Java `char`, but i am inclined to say the user should cast between Java `short` to `char` in such cases.

(Another advantage of this is that, should we get proper unsigned carriers from Valhalla one day, native linkers could be updated to support those en masse - not just for `short`)

PaulSandoz:
> Effectively, what you suggest amount at saying: we do have a JAVA_CHAR layout, which is mostly there for Java interop. But a native linker (which only cares about native interop) doesn't really care much about that. Does that sound good?

Yes.

mcimadamore:
Updated javadoc and specdiffs (v3):
javadoc: https://cr.openjdk.org/~mcimadamore/jdk/8308293/v3/javadoc/java.base/java/lang/foreign/Linker.html
specdiff: https://cr.openjdk.org/~mcimadamore/jdk/8308293/v3/specdiff_out/overview-summary.html

test/jdk/java/foreign/normalize/TestNormalize.java
```diff
@@ -166,8 +166,7 @@ public static Object[][] cases() {
         return new Object[][] {
             { JAVA_BOOLEAN, booleanToInt(true),     BOOLEAN_HOB_MASK, BOOLEAN_TO_INT, SAVE_BOOLEAN_AS_INT },
             { JAVA_BYTE,    byteToInt((byte) 42),   BYTE_HOB_MASK,    BYTE_TO_INT,    SAVE_BYTE_AS_INT    },
-            { JAVA_SHORT,   shortToInt((short) 42), SHORT_HOB_MASK,   SHORT_TO_INT,   SAVE_SHORT_AS_INT   },
-            { JAVA_CHAR,    charToInt('a'),         CHAR_HOB_MASK,    CHAR_TO_INT,    SAVE_CHAR_AS_INT    }
+            { JAVA_SHORT,   shortToInt((short) 42), SHORT_HOB_MASK,   SHORT_TO_INT,   SAVE_SHORT_AS_INT   }
```
ExE-Boss, line 169: Since arrays support trailing commas, this can use that:
```suggestion
            { JAVA_SHORT,   shortToInt((short) 42), SHORT_HOB_MASK,   SHORT_TO_INT,   SAVE_SHORT_AS_INT   },
```

PaulSandoz (approved):
This all looks reasonable. I say let's soak it and then see if we need refine based on feedback and further research (e.g., if we find we need to declare multiple layouts per ABI type for extensibility reasons).

mcimadamore:
Thanks for taking the time to review. After some more consideration, I will withdraw this PR. While this API is largely not problematic, we need to make sure that this API fits with how the FFM API will be evolved to support other types besides the C basic types we know and love (e.g. `long double`, vectors, `complex` types). So adding this API point now seems premature.

PR #14038, opened 2023-05-17
8308310

dfuch:
Please find here a change that revisits usage of monitors in the HttpClient.

With Virtual Threads now part of the platform it should be possible to pass a newVirtualThreadPerTaskExecutor to the HttpClient. Logging, when called from a synchronized block, can cause the carrier thread to get pinned in case of contention when printing through the underlying PrintStream.

This change aims at avoiding situations where the carrier threads might get pinned.

### DIFF
--- a/src/java.net.http/share/classes/jdk/internal/net/http/ConnectionPool.java
+++ b/src/java.net.http/share/classes/jdk/internal/net/http/ConnectionPool.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -40,6 +40,7 @@ import java.util.ListIterator;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.Flow;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.Collectors;
 import jdk.internal.net.http.common.FlowTube;
 import jdk.internal.net.http.common.Logger;
@@ -57,11 +58,12 @@ final class ConnectionPool {
 
     // Pools of idle connections
 
+    private final ReentrantLock stateLock =  new ReentrantLock();
     private final HashMap<CacheKey,LinkedList<HttpConnection>> plainPool;
     private final HashMap<CacheKey,LinkedList<HttpConnection>> sslPool;
     private final ExpiryList expiryList;
     private final String dbgTag; // used for debug
-    boolean stopped;
+    volatile boolean stopped;
 
     /**
      * Entries in connection pool are keyed by destination address and/or
@@ -139,7 +141,7 @@ final class ConnectionPool {
         return dbgTag;
     }
 
-    synchronized void start() {
+    void start() {
         assert !stopped : "Already stopped";
     }
 
@@ -149,9 +151,21 @@ final class ConnectionPool {
         return new CacheKey(secure, destination, proxy);
     }
 
-    synchronized HttpConnection getConnection(boolean secure,
-                                              InetSocketAddress addr,
-                                              InetSocketAddress proxy) {
+    HttpConnection getConnection(boolean secure,
+                                 InetSocketAddress addr,
+                                 InetSocketAddress proxy) {
+        if (stopped) return null;
+        stateLock.lock();
+        try {
+            return getConnection0(secure, addr, proxy);
+        } finally {
+            stateLock.unlock();
+        }
+    }
+
+    private HttpConnection getConnection0(boolean secure,
+                                  InetSocketAddress addr,
+                                  InetSocketAddress proxy) {
         if (stopped) return null;
         // for plain (unsecure) proxy connection the destination address is irrelevant.
         addr = secure || proxy == null ? addr : null;
@@ -185,7 +199,8 @@ final class ConnectionPool {
 
         // it's possible that cleanup may have been called.
         HttpConnection toClose = null;
-        synchronized(this) {
+        stateLock.lock();
+        try {
             if (cleanup.isDone()) {
                 return;
             } else if (stopped) {
@@ -203,6 +218,8 @@ final class ConnectionPool {
                 putConnection(conn, sslPool);
             }
             expiryList.add(conn, now, keepAlive);
+        } finally {
+            stateLock.unlock();
         }
         if (toClose != null) {
             if (debug.on()) {
@@ -243,7 +260,7 @@ final class ConnectionPool {
     removeFromPool(HttpConnection c,
                    HashMap<CacheKey,LinkedList<HttpConnection>> pool) {
         //System.out.println("cacheCleaner removing: " + c);
-        assert Thread.holdsLock(this);
+        assert stateLock.isHeldByCurrentThread();
         CacheKey k = c.cacheKey();
         List<HttpConnection> l = pool.get(k);
         if (l == null || l.isEmpty()) {
@@ -288,7 +305,8 @@ final class ConnectionPool {
         if (!expiryList.purgeMaybeRequired()) return nextPurge;
 
         List<HttpConnection> closelist;
-        synchronized (this) {
+        stateLock.lock();
+        try {
             closelist = expiryList.purgeUntil(now);
             for (HttpConnection c : closelist) {
                 if (c instanceof PlainHttpConnection) {
@@ -302,6 +320,8 @@ final class ConnectionPool {
             nextPurge = now.until(
                     expiryList.nextExpiryDeadline().orElse(now),
                     ChronoUnit.MILLIS);
+        } finally {
+            stateLock.unlock();
         }
         closelist.forEach(this::close);
         return nextPurge;
@@ -316,7 +336,8 @@ final class ConnectionPool {
     void stop() {
         List<HttpConnection> closelist = Collections.emptyList();
         try {
-            synchronized (this) {
+            stateLock.lock();
+            try {
                 stopped = true;
                 closelist = expiryList.stream()
                     .map(e -> e.connection)
@@ -324,6 +345,8 @@ final class ConnectionPool {
                 expiryList.clear();
                 plainPool.clear();
                 sslPool.clear();
+            } finally {
+                stateLock.unlock();
             }
         } finally {
             closelist.forEach(this::close);
@@ -354,28 +377,25 @@ final class ConnectionPool {
 
         // A loosely accurate boolean whose value is computed
         // at the end of each operation performed on ExpiryList;
-        // Does not require synchronizing on the ConnectionPool.
+        // Does not require holding the ConnectionPool stateLock.
         boolean purgeMaybeRequired() {
             return mayContainEntries;
         }
 
         // Returns the next expiry deadline
-        // should only be called while holding a synchronization
-        // lock on the ConnectionPool
+        // should only be called while holding the ConnectionPool stateLock.
         Optional<Instant> nextExpiryDeadline() {
             if (list.isEmpty()) return Optional.empty();
             else return Optional.of(list.getLast().expiry);
         }
 
-        // should only be called while holding a synchronization
-        // lock on the ConnectionPool
+        // should only be called while holding the ConnectionPool stateLock.
         HttpConnection removeOldest() {
             ExpiryEntry entry = list.pollLast();
             return entry == null ? null : entry.connection;
         }
 
-        // should only be called while holding a synchronization
-        // lock on the ConnectionPool
+        // should only be called while holding the ConnectionPool stateLock.
         void add(HttpConnection conn) {
             add(conn, Instant.now(), KEEP_ALIVE_TIMEOUT);
         }
@@ -408,8 +428,7 @@ final class ConnectionPool {
             mayContainEntries = true;
         }
 
-        // should only be called while holding a synchronization
-        // lock on the ConnectionPool
+        // should only be called while holding the ConnectionPool stateLock.
         void remove(HttpConnection c) {
             if (c == null || list.isEmpty()) return;
             ListIterator<ExpiryEntry> li = list.listIterator();
@@ -423,8 +442,7 @@ final class ConnectionPool {
             }
         }
 
-        // should only be called while holding a synchronization
-        // lock on the ConnectionPool.
+        // should only be called while holding the ConnectionPool stateLock.
         // Purge all elements whose deadline is before now (now included).
         List<HttpConnection> purgeUntil(Instant now) {
             if (list.isEmpty()) return Collections.emptyList();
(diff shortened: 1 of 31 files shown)


test/jdk/java/net/httpclient/AuthFilterCacheTest.java
```diff
@@ -267,19 +270,39 @@ void doClient(List<URI> uris) {
...
+                                                                            String protocol,
+                                                                            String prompt,
+                                                                            String scheme,
+                                                                            URL url,
+                                                                            RequestorType reqType) {
+            PasswordAuthentication passwordAuthentication;
+            int count;
+            synchronized (this) {
+                count = ++this.count;
+                passwordAuthentication =  super.requestPasswordAuthenticationInstance(
```
turbanoff, line 298:
```suggestion
                passwordAuthentication = super.requestPasswordAuthenticationInstance(
```

src/java.net.http/share/classes/jdk/internal/net/http/ConnectionPool.java
```diff
@@ -57,11 +58,12 @@ final class ConnectionPool {
 
     // Pools of idle connections
 
+    private final ReentrantLock stateLock =  new ReentrantLock();
```
turbanoff, line 61:
```suggestion
    private final ReentrantLock stateLock = new ReentrantLock();
```

src/java.net.http/share/classes/jdk/internal/net/http/Http1Response.java
```diff
@@ -114,21 +116,25 @@ private String dbgString() {
...
-        byte state;
+        volatile byte state;
+        ClientRefCountTracker(HttpClientImpl client, Logger logger) {
+            this.client = client;
+            this.debug = logger;
+        }
 
-        public synchronized boolean acquire() {
-            if (state == 0) {
+        public  boolean acquire() {
```
turbanoff, line 130:
```suggestion
        public boolean acquire() {
```

src/java.net.http/share/classes/jdk/internal/net/http/ConnectionPool.java
```diff
@@ -149,9 +151,21 @@ static CacheKey cacheKey(boolean secure, InetSocketAddress destination,
...
+        stateLock.lock();
+        try {
+            return getConnection0(secure, addr, proxy);
+        } finally {
+            stateLock.unlock();
+        }
+    }
+
+    private HttpConnection getConnection0(boolean secure,
+                                  InetSocketAddress addr,
```
turbanoff, line 167: let's align parameters

src/java.net.http/share/classes/jdk/internal/net/http/Http1Exchange.java
```diff
@@ -68,11 +69,11 @@
 
     /** Records a possible cancellation raised before any operation
      * has been initiated, or an error received while sending the request. */
-    private Throwable failed;
+    private AtomicReference<Throwable> failedRef = new AtomicReference<>();
```
turbanoff, line 72: Seems it's only modified under lock. Can we use `volatile` instead?

src/java.net.http/share/classes/jdk/internal/net/http/ConnectionPool.java
```diff
@@ -149,9 +151,21 @@ static CacheKey cacheKey(boolean secure, InetSocketAddress destination,
...
+                                  InetSocketAddress proxy) {
```
dfuch, line 168:
```suggestion
    private HttpConnection getConnection0(boolean secure,
                                                                       InetSocketAddress addr,
                                                                       InetSocketAddress proxy) {
```

src/java.net.http/share/classes/jdk/internal/net/http/Http1Exchange.java
dfuch: I prefer to keep AtomicReference - it's easier to reason about.
Maybe at some time I'll do a pass to use VarHandles in various places but now is not the time.

turbanoff: let's make it `final` at least

dfuch: Oops! Thanks for noticing that. That's a bug. Done.

src/java.net.http/share/classes/jdk/internal/net/http/Stream.java
```diff
@@ -148,7 +150,8 @@
     private boolean requestSent, responseReceived;
 
     // send lock: prevent sending DataFrames after reset occurred.
-    private final Object sendLock = new Object();
+    private final Lock sendLock = new ReentrantLock();
+    final Lock stateLock = new ReentrantLock();
```
turbanoff, line 154: can it be `private` too?

src/java.net.http/share/classes/jdk/internal/net/http/Http1Response.java
```diff
@@ -114,21 +116,25 @@ private String dbgString() {
...
-    private final class ClientRefCountTracker {
-        final HttpClientImpl client = connection.client();
+    private static final class ClientRefCountTracker {
+        final HttpClientImpl client;
+        final Logger debug;
         // state & 0x01 != 0 => acquire called
         // state & 0x02 != 0 => tryRelease called
-        byte state;
+        volatile byte state;
+        ClientRefCountTracker(HttpClientImpl client, Logger logger) {
```
turbanoff, line 125: let's add an empty line before constructor to make distinguishing more easy

src/java.net.http/share/classes/jdk/internal/net/http/Stream.java
dfuch: Good point

src/java.net.http/share/classes/jdk/internal/net/http/ConnectionPool.java
```diff
@@ -450,25 +468,22 @@ List<HttpConnection> purgeUntil(Instant now) {
...
-        // should only be called while holding a synchronization
-        // lock on the ConnectionPool
+        // should only be called while holding the ConnectionPool stateLock.
         java.util.stream.Stream<ExpiryEntry> stream() {
             return list.stream();
         }
 
-        // should only be called while holding a synchronization
-        // lock on the ConnectionPool
+        // should only be called while holding the ConnectionPool stateLock.
```
jaikiran, line 476: Hello Daniel, all these methods which say a lock needs to be held when they are called, should we add a `assert stateLock.isHeldByCurrentThread();` to make it verifiable? I understand we didn't have a similar assert when the comment said the method needs to be called while holding the monitor, but since we are changing this part now, perhaps we can add those asserts?

src/java.net.http/share/classes/jdk/internal/net/http/Http1Exchange.java
```diff
@@ -342,8 +349,11 @@ CompletableFuture<ExchangeImpl<T>> sendHeadersAsync() {
 
                         if (debug.on()) debug.log("requestAction.headers");
                         List<ByteBuffer> data = requestAction.headers();
-                        synchronized (lock) {
+                        lock.lock();
+                        try {
                             state = State.HEADERS;
```
jaikiran, line 354: Can this code be replaced by a call to `switchState(...)` to be consistent with the rest of the places in this class which do a similar call?

src/java.net.http/share/classes/jdk/internal/net/http/Http1Response.java
```diff
@@ -150,12 +157,21 @@ public synchronized void tryRelease() {
                 client.unreference();
             } else if (state == 0) {
                 if (debug.on())
-                    debug.log("Operation finished: releasing ref count for %s", client);
+                    debug.log("Operation not started: releasing ref count for %s", client);
             } else if ((state & 0x02) == 0x02) {
                 if (debug.on())
                     debug.log("ref count for %s already released", client);
             }
-            state |= 0x02;
```
jaikiran, line 158: It looks like previously (before the changes in this PR), the `tryRelease` could incorrectly set this `state` to `0x02` even when the `state` was `0`. From what I see in this change, that seems to have been addressed now.

src/java.net.http/share/classes/jdk/internal/net/http/Http1Response.java
```diff
@@ -139,8 +146,8 @@ public synchronized boolean acquire() {
             }
         }
 
-        public synchronized void tryRelease() {
-            if (state == 0x01) {
+        public void tryRelease() {
+            if (STATE.compareAndSet(this, (byte)0x01, (byte)0x03)) {
```
jaikiran, line 150: Nit - a space is missing between the cast and the value for the second and third parameters.

src/java.net.http/share/classes/jdk/internal/net/http/Http2ClientImpl.java
```diff
@@ -293,6 +314,14 @@ SettingsFrame getClientSettings() {
         return frame;
     }
 
+    private void lock() {
+        lock.lock();
+    }
+
+    private void unlock() {
+        lock.unlock();
+    }
```
jaikiran, line 323: Any specific reason for creating private wrapper methods instead of just using `lock.lock()` and `lock.unlock()` at the call sites?

src/java.net.http/share/classes/jdk/internal/net/http/ConnectionPool.java
dfuch: The `stateLock` variable is not available from within the method:
```
Non-static field 'stateLock' cannot be referenced from a static context
```

src/java.net.http/share/classes/jdk/internal/net/http/Http1Response.java
dfuch: OK - will do - though I usually don't put a space there.

src/java.net.http/share/classes/jdk/internal/net/http/Http2Connection.java
```diff
@@ -556,16 +573,26 @@ private static CompletableFuture<?> checkSSLConfig(AbstractAsyncSSLConnection ac
                 .thenCompose(checkAlpnCF);
     }
 
-    synchronized boolean finalStream() {
-        return finalStream;
+    boolean finalStream() {
+        stateLock.lock();
+        try {
+            return finalStream;
```
jaikiran, line 579: Perhaps we can make `finalStream` volatile and remove the use of locks when accessing and setting it?

src/java.net.http/share/classes/jdk/internal/net/http/ConnectionPool.java
jaikiran: Thank you, that wasn't clearly noticable in the diff.

src/java.net.http/share/classes/jdk/internal/net/http/Http2ClientImpl.java
dfuch: removed private methods

src/java.net.http/share/classes/jdk/internal/net/http/Http1Response.java
jaikiran: It's fine in the current form if this wasn't an oversight.

src/java.net.http/share/classes/jdk/internal/net/http/Http2Connection.java
dfuch: Looks doable

src/java.net.http/share/classes/jdk/internal/net/http/MultiExchange.java
```diff
@@ -211,12 +211,16 @@ HttpClient.Version version() {
...
 
-    private synchronized void setExchange(Exchange<T> exchange) {
-        if (this.exchange != null && exchange != this.exchange) {
-            this.exchange.released();
-            if (cancelled) exchange.cancel();
+    private void setExchange(Exchange<T> exchange) {
+        Exchange<T> previousExchange;
+        synchronized (this) {
+            previousExchange = this.exchange;
+            this.exchange = exchange;
```
jaikiran, line 218: There's a behavioural change here as compared to before this change. Previously, we used to first call `released()` on the old exchange and `cancel()` the new exchange, before setting the `this.exchange` to the new exchange. Now, with this change, we first switch the `this.exchange` to the new exchange and then call `released()` and `cancel()`. Do you think that could have any (odd) issues?

dfuch: Good observation. When we reach here the previous exchange is practically terminated. Looking at the implementation of released(), and at the place where `setExchange` is called, I don't think it matters.

test/jdk/java/net/httpclient/ConcurrentResponses.java
```diff
@@ -168,13 +178,20 @@ void testAsString(String uri) throws Exception {
                 .map(cf -> cf.thenCompose(response -> assertbody(response, requests.get(response.request()))))
                 .toArray(CompletableFuture<?>[]::new))
                 .join();
+        client.close();
+        virtualExecutor.close();
```
jaikiran, line 182: Could we do this in a try/finally block?

test/jdk/java/net/httpclient/ManyRequests.java
```diff
@@ -260,13 +271,13 @@ static void test(HttpsServer server, HttpClient client) throws Exception {
                 done = true;
             } catch (CompletionException e) {
                 if (!Platform.isWindows()) throw e;
-                if (LIMIT.get() < REQUESTS) throw e;
+                if (LIMIT.get() < MAX_LIMIT) throw e;
```
jaikiran, line 274: Previously, `LIMIT.get()` was being compared against `MAX_COUNT`, because `REQUEST` was initialized (as `final`) to that value. Now it's being compared to `MAX_LIMIT`. Is this intentional?

jaikiran:
The tests that have been selectively changed to use virtual threads for the HttpClient (and have been  set with `-Djdk.tracePinnedThreads=full`), I'm guessing were selected based on some of the failures that were noticed when virtual threads were used?

src/java.net.http/share/classes/jdk/internal/net/http/MultiExchange.java
jaikiran: Thank you Daniel for checking.

test/jdk/java/net/httpclient/ManyRequests.java
dfuch: What it says here is that if we already reduce the LIMIT once and it fails again we should fail.

dfuch:
> The tests that have been selectively changed to use virtual threads for the HttpClient (and have been set with -Djdk.tracePinnedThreads=full), I'm guessing were selected based on some of the failures that were noticed when virtual threads were used?

Yes - these were the bigger offenders causing virtual threads to get pinned.

jaikiran (approved):
The latest changes in `f5986f93` look fine to me. There's just one suggestion that's remaining in the `ConcurrentResponses` test about moving the closing of the client and virtual thread executor in a finally block, but other than that this looks good to me.

jaikiran (approved):

PR #14040, opened 2023-05-17
8308108: Support Unicode extension for collation settings

naotoj:
This change intends to interpret the BCP47 U extension wrt collation settings in the given `Locale`, then applies them to the created instances in the 1-arg factory method in `Collator`. A corresponding CSR has also been drafted.

---------
### Progress
- [x] Change requires CSR request [JDK-8308243](https://bugs.openjdk.org/browse/JDK-8308243) to be approved

### Issues
 * [JDK-8308108](https://bugs.openjdk.org/browse/JDK-8308108): Support Unicode extension for collation settings
 * [JDK-8308243](https://bugs.openjdk.org/browse/JDK-8308243): Support Unicode extension for collation settings (**CSR**)

### Reviewers
 * [Iris Clark](https://openjdk.org/census#iris) (@irisclark - **Reviewer**) ⚠️ Review applies to [88a93839](https://git.openjdk.org/jdk/pull/14040/files/88a9383999f70ac43c2d5fbdfefe82a00a750b13)
 * [Roger Riggs](https://openjdk.org/census#rriggs) (@RogerRiggs - **Reviewer**) ⚠️ Review applies to [627b78b3](https://git.openjdk.org/jdk/pull/14040/files/627b78b375a25b3a5b9ab9f580c021bf8c09eb62)
 * [Justin Lu](https://openjdk.org/census#jlu) (@justin-curtis-lu - Committer) ⚠️ Review applies to [627b78b3](https://git.openjdk.org/jdk/pull/14040/files/627b78b375a25b3a5b9ab9f580c021bf8c09eb62)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14040/head:pull/14040` \
`$ git checkout pull/14040`

`$ git checkout pull/14040` \
`$ git pull https://git.openjdk.org/jdk.git pull/14040/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14040`

`$ git pr show -t 14040`

</details>

<a href="https://git.openjdk.org/jdk/pull/14040.diff">https://git.openjdk.org/jdk/pull/14040.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14040#issuecomment-1551959300)

### DIFF
--- a/src/java.base/share/classes/java/text/Collator.java
+++ b/src/java.base/share/classes/java/text/Collator.java
@@ -230,9 +230,39 @@ public abstract class Collator
      * has the "{@code ks}" and/or the "{@code kk}"
      * <a href="https://www.unicode.org/reports/tr35/tr35-collation.html#Setting_Options">
      * Unicode collation settings</a>, the returned {@code Collator} instance
-     * will have the specified strength and/or decomposition respectively. If the specified
-     * setting value is not recognized, strength and/or decomposition will not be
-     * overridden.
+     * will have the specified strength and/or decomposition respectively. The following
+     * table describes the mappings between recognizable BCP47 values and their corresponding
+     * Collator constants.
+     * <table class="striped">
+     * <caption style="display:none">Strength/Decomposition mappings</caption>
+     * <thead>
+     * <tr><th scope="col">BCP 47 values for strength (ks)</th>
+     *     <th scope="col">Collator constants for strength</th></tr>
+     * </thead>
+     * <tbody>
+     * <tr><th scope="row" style="text-align:left">level1</th>
+     *     <td>PRIMARY</td></tr>
+     * <tr><th scope="row" style="text-align:left">level2</th>
+     *     <td>SECONDARY</td></tr>
+     * <tr><th scope="row" style="text-align:left">level3</th>
+     *     <td>TERTIARY</td></tr>
+     * <tr><th scope="row" style="text-align:left">identic</th>
+     *     <td>IDENTICAL</td></tr>
+     * </tbody>
+     * <thead>
+     * <tr><th scope="col">BCP 47 values for normalization (kk)</th>
+     *     <th scope="col">Collator constants for decomposition</th></tr>
+     * </thead>
+     * <tbody>
+     * <tr><th scope="row" style="text-align:left">true</th>
+     *     <td>CANONICAL_DECOMPOSITION</td></tr>
+     * <tr><th scope="row" style="text-align:left">false</th>
+     *     <td>NO_DECOMPOSITION</td></tr>
+     * </tbody>
+     * </table>
+     * If the specified setting value is not recognized, strength and/or
+     * decomposition will not be overridden.
+     *
      * @apiNote Implementations of {@code Collator} class may produce
      * different instances based on the "{@code co}"
      * <a href="https://www.unicode.org/reports/tr35/#UnicodeCollationIdentifier">
--- a/src/java.base/share/classes/java/text/Collator.java
+++ b/src/java.base/share/classes/java/text/Collator.java
@@ -226,7 +226,13 @@ public abstract class Collator
     }
 
     /**
-     * Gets the Collator for the desired locale.
+     * Gets the Collator for the desired locale. If the desired locale
+     * has the "{@code ks}" and/or the "{@code kk}"
+     * <a href="https://www.unicode.org/reports/tr35/tr35-collation.html#Setting_Options">
+     * Unicode collation settings</a>, the returned {@code Collator} instance
+     * will have the specified strength and/or decomposition respectively. If the specified
+     * setting value is not recognized, strength and/or decomposition will not be
+     * overridden.
      * @apiNote Implementations of {@code Collator} class may produce
      * different instances based on the "{@code co}"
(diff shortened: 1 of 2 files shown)


test/jdk/sun/text/resources/Collator/CollationSettingsTests.java
```diff
@@ -0,0 +1,85 @@
...
+import static java.text.Collator.*;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class CollationSettingsTests {
+    private static final Collator ENG_DEF =  Collator.getInstance(Locale.ENGLISH);
```
turbanoff, line 43: nit
```suggestion
    private static final Collator ENG_DEF = Collator.getInstance(Locale.ENGLISH);
```

irisclark (approved):
Looks good.

Associated CSR also Reviewed.

src/java.base/share/classes/java/text/Collator.java
```diff
@@ -226,7 +226,13 @@ public static synchronized Collator getInstance() {
...
     /**
-     * Gets the Collator for the desired locale.
+     * Gets the Collator for the desired locale. If the desired locale
+     * has the "{@code ks}" and/or the "{@code kk}"
+     * <a href="https://www.unicode.org/reports/tr35/tr35-collation.html#Setting_Options">
+     * Unicode collation settings</a>, the returned {@code Collator} instance
+     * will have the specified strength and/or decomposition respectively. If the specified
+     * setting value is not recognized, strength and/or decomposition will not be
+     * overridden.
      * @apiNote Implementations of {@code Collator} class may produce
```
RogerRiggs, line 236: The API should identify the valid values for ks and kk and the corresponding values as passed to setStrength and setDecomposition.

RogerRiggs (approved):
Looks good

src/java.base/share/classes/java/text/Collator.java
```diff
@@ -226,7 +226,43 @@ public static synchronized Collator getInstance() {
...
+     * </thead>
+     * <tbody>
+     * <tr><th scope="row" style="text-align:left">true</th>
+     *     <td>CANONICAL_DECOMPOSITION</td></tr>
+     * <tr><th scope="row" style="text-align:left">false</th>
+     *     <td>NO_DECOMPOSITION</td></tr>
+     * </tbody>
+     * </table>
+     * If the specified setting value is not recognized, strength and/or
+     * decomposition will not be overridden.
```
AlanBateman, line 264: The proposed update looks okay but I don't think "If the specified setting value is not recognized, strength and/or decomposition will not be overridden" could be clearer.
If I understand correctly, if the Unicode locale type for the given Locale might have a strength level and decomposition mode and this method makes a best effort to return a Collator with the expected level and mode. If so, then I think the javadoc needs to make this a bit clearer. Also, if the factory method returns a Collator that doesn't do what I asked, can I call setStrength/setDecomposition to change it?

naotoj: I don't think there is any chance that the factory method returns a Collator that fails to meet the settings in the locale, because setStrength()/setDecompositon() with valid values never fail. So the returned instance will always have the requested strength/decomposition. Does this answer your question?

AlanBateman: > I don't think there is any chance that the factory method returns a Collator that fails to meet the settings in the locale, because setStrength()/setDecompositon() with valid values never fail. So the returned instance will always have the requested strength/decomposition. Does this answer your question?

It's the last sentence that bothers me. I understand the setters can't fail but the last sentence of the proposed text makes it sound like invalid settings will be just ignored. So I think this sentence needs to be re-examined as I don't think I understand what it is trying to say now.

justin-curtis-lu: IIUC that's already the existing behavior for any Unicode extension.

Example,

```
Locale enIsoLoc = Locale.forLanguageTag("en-u-fw-mon");
WeekFields.of(enIsoLoc).getFirstDayOfWeek(); // MONDAY
Locale badExtension = Locale.forLanguageTag("en-u-fw-zzz");
WeekFields.of(badExtension).getFirstDayOfWeek(); // SUNDAY
```

If the value for a Unicode extension is invalid, they are simply ignored. In this case 'zzz' is not recognized and thus can not overhaul the first day of the week.

justin-curtis-lu (approved):
LGTM

AlanBateman:
I went through the latest javadoc but the last sentence is still confusing. I think the part that is missing is the method doesn't make it clear up front that it gets requested locale and then attempts to set its strength and decomposition if recognized. If this were clear in the first part of the method description then it is okay to say that the returned Collator might not have the expected strength and decomposition.

naotoj:
Thanks, Alan. Clarified the method doc based on your suggestion.

AlanBateman (approved):
Thanks for persisting with this to get the wording clear, the latest version looks good.

PR #14045, opened 2023-05-18
JDK-8308350: Increase buffer size for jspawnhelper arguments

tstuefe:
Trivial fix for a small problem.

jspawnhelper gets handed several file descriptors as arguments. The buffer size for this string is too small (7 chars per fd) to print out every conceivable int. This will overun the buffer if we happen to have fds larger than (printed size) 7 characters. This could lead to crashes or malfunctions if the parent VM has opened a large amount of file descriptors.

Note that on Linux, this can normally not happen since the kernel limits the number of open file descriptors per process to 1M, and these fds are still printable within the limits of this buffer. It is possible to get more fds per process, but only via kernel patch. But we still should not rely on that. And there is also still MacOS using the same mechanism.

---------
### Progress

### Issue
 * [JDK-8308350](https://bugs.openjdk.org/browse/JDK-8308350): Increase buffer size for jspawnhelper arguments

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14045/head:pull/14045` \
`$ git checkout pull/14045`

`$ git checkout pull/14045` \
`$ git pull https://git.openjdk.org/jdk.git pull/14045/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14045`

`$ git pr show -t 14045`

</details>

<a href="https://git.openjdk.org/jdk/pull/14045.diff">https://git.openjdk.org/jdk/pull/14045.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14045#issuecomment-1552626932)

### DIFF
--- a/src/java.base/unix/native/libjava/ProcessImpl_md.c
+++ b/src/java.base/unix/native/libjava/ProcessImpl_md.c
@@ -487,7 +487,7 @@ static pid_t
 spawnChild(JNIEnv *env, jobject process, ChildStuff *c, const char *helperpath) {
     pid_t resultPid;
     int i, offset, rval, bufsize, magic;
-    char *buf, buf1[16];
+    char *buf, buf1[(2 * 11) + 2]; // "%d:%d\0"
     char *hlpargs[2];
     SpawnInfo sp;
 


tstuefe:
> jspawnhelper gets handed several file descriptors as arguments. The buffer size for this string is too small (7 chars per fd) to print out every conceivable int. This will overun the buffer if we happen to have fds larger than (printed size) 7 characters. This could lead to crashes (snprintf does not zero terminate on truncation) or malfunctions (if \0 happened to follow and we pass an invalid truncated fd to jspawnhelper) if the parent VM has opened a large amount of file descriptors.
> 
> Note that on Linux, this can normally not happen since the kernel limits the number of open file descriptors per process to 1M, and these fds are still printable within the limits of this buffer. It is possible to get more fds per process, but only via kernel patch. But we still should not rely on that. And there is also still MacOS using the same mechanism.

tstuefe:
GHAs are all suffering from infrastructure problems.

RogerRiggs (approved):
Trivial

tstuefe:
Thanks @RogerRiggs !

/integrate

PR #14049, opened 2023-05-18
8308369: Default decomposition mode in Collator

naotoj:
Amending the description about the default decomposition mode in `Collator.NO_DECOMPOSITION` javadoc.


RogerRiggs (approved):
Looks good.

PR #14056, opened 2023-05-19
8308168: Convenient methods to build switch and try/finally in Classfile API

asotona:
Classfile API provides convenient methods for if/then/else or try/catch, however similar methods to build switch or try/finally are missing.

Here I would like to propose new `CodeBuilder` conveniences:
`lookupswitch(Consumer<SwitchBuilder> switchHandler)`
`tableswitch(Consumer<SwitchBuilder> switchHandler)`
`tryWithFinalizer(Consumer<CodeBuilder> tryHandler,
                                         Consumer<CodeBuilder> finalizerHandler,
                                         Label... externalLabels)`
`tryWithFinalizer(Consumer<CodeBuilder> tryHandler,
                                         Consumer<CodeBuilder> finalizerHandler,
                                         boolean compactForm,
                                         Label... externalLabels)`

and new `CodeBuilder.SwitchBuilder` interface with methods:
`switchCase(int caseValue, Consumer<BlockCodeBuilder> caseHandler)`
`switchCase(List<Integer> caseValues, Consumer<BlockCodeBuilder> caseHandler)`
`defaultCase(Consumer<BlockCodeBuilder> defaultHandler)`

---------
### Progress

### Issue
 * [JDK-8308168](https://bugs.openjdk.org/browse/JDK-8308168): Convenient methods to build switch and try/finally in Classfile API

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14056/head:pull/14056` \
`$ git checkout pull/14056`

`$ git checkout pull/14056` \
`$ git pull https://git.openjdk.org/jdk.git pull/14056/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14056`

`$ git pr show -t 14056`

</details>

<a href="https://git.openjdk.org/jdk/pull/14056.diff">https://git.openjdk.org/jdk/pull/14056.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/classfile/CodeBuilder.java
+++ b/src/java.base/share/classes/jdk/internal/classfile/CodeBuilder.java
@@ -187,10 +187,6 @@ public sealed interface CodeBuilder
          * targeting the "break" label is appended to the built block.
          */
         Label breakLabel();
-
-        default CodeBuilder break_() {
-            return goto_(breakLabel());
-        }
     }
 
     /**
@@ -402,9 +398,10 @@ public sealed interface CodeBuilder
          *
          * @param defaultHandler handler that receives a {@linkplain CodeBuilder} to
          *                       generate the body of the default switch case
+         * @return this builder
          * @throws java.lang.IllegalArgumentException if an existing block handles default switch case.
          */
-        void defaultCase(Consumer<BlockCodeBuilder> defaultHandler);
+        SwitchBuilder defaultCase(Consumer<BlockCodeBuilder> defaultHandler);
     }
 
     /**
--- a/src/java.base/share/classes/jdk/internal/classfile/CodeBuilder.java
+++ b/src/java.base/share/classes/jdk/internal/classfile/CodeBuilder.java
@@ -52,6 +52,7 @@ import jdk.internal.classfile.impl.CatchBuilderImpl;
 import jdk.internal.classfile.impl.ChainedCodeBuilder;
 import jdk.internal.classfile.impl.LabelImpl;
 import jdk.internal.classfile.impl.NonterminalCodeBuilder;
+import jdk.internal.classfile.impl.SwitchBuilderImpl;
 import jdk.internal.classfile.impl.TerminalCodeBuilder;
 import jdk.internal.classfile.instruction.ArrayLoadInstruction;
 import jdk.internal.classfile.instruction.ArrayStoreInstruction;
@@ -186,6 +187,10 @@ public sealed interface CodeBuilder
          * targeting the "break" label is appended to the built block.
          */
         Label breakLabel();
+
+        default CodeBuilder break_() {
+            return goto_(breakLabel());
+        }
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/java.base/share/classes/jdk/internal/classfile/CodeBuilder.java
```diff
@@ -1192,6 +1234,14 @@ default CodeBuilder lookupswitch(Label defaultTarget, List<SwitchCase> cases) {
         return lookupSwitchInstruction(defaultTarget, cases);
     }
 
+    default CodeBuilder lookupswitch(Consumer<SwitchBuilder> switchHandler) {
+        return block(b -> {
```
liach, line 1238: Maybe we should convert this into an anonymous or local class, for these switch building utilities might be useful when migrating the switching MethodHandle's LambdaForm invoker code.

src/java.base/share/classes/jdk/internal/classfile/CodeBuilder.java
```diff
@@ -1378,8 +1428,30 @@ default CodeBuilder tableswitch(Label defaultTarget, List<SwitchCase> cases) {
         return tableSwitchInstruction(low, high, defaultTarget, cases);
     }
 
-    // Structured conveniences:
+    default CodeBuilder tableswitch(Consumer<SwitchBuilder> switchHandler) {
+        return block(b -> {
```
liach, line 1432: Same anonymous class comment for tableswitch

src/java.base/share/classes/jdk/internal/classfile/impl/CodeBuilderFinalizerImpl.java
```diff
@@ -0,0 +1,142 @@
+/*
+ * Copyright (c) 2022, Oracle and/or its affiliates. All rights reserved.
```
liach, line 2:
```suggestion
 * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
```
Other files need copyright year updates to `Copyright (c) 2022, 2023` as well.

src/java.base/share/classes/jdk/internal/classfile/impl/CodeBuilderFinalizerImpl.java
```diff
@@ -0,0 +1,142 @@
...
+
+public final class CodeBuilderFinalizerImpl {
+
+    public static void buildWithFinalizer(CodeBuilder parent,
+                                          Consumer<CodeBuilder> tryHandler,
+                                          Consumer<CodeBuilder> finalizerHandler,
+                                          boolean compactForm,
+                                          Label... externalLabels) {
+
+        var mapping = new HashMap<Label, Label>(externalLabels.length + 1);
```
liach, line 46:
```suggestion
        var mapping = HashMap.<Label, Label>newHashMap(externalLabels.length + 1);
```

src/java.base/share/classes/jdk/internal/classfile/impl/CodeBuilderFinalizerImpl.java
```diff
@@ -0,0 +1,142 @@
...
+        for (var extLabel : externalLabels) {
+            mapping.put(extLabel, null);
+        }
+
+        var endLabel = parent.newLabel();
+        BlockCodeBuilderImpl topBlock = new BlockCodeBuilderImpl(parent, endLabel);
+        topBlock.start();
+        var excTable = new ArrayList<Label>();
+        topBlock.transforming(
+                (cob, coe) -> {
```
liach, line 56: Same local/anonymous class vs lambda comment, MethodHandles does have an intrinsic (in InvokerBytecodeGenerator) for try-finally block

src/java.base/share/classes/jdk/internal/classfile/impl/CodeBuilderFinalizerImpl.java
```diff
@@ -0,0 +1,142 @@
...
+import jdk.internal.classfile.CodeBuilder;
+import jdk.internal.classfile.Instruction;
+import jdk.internal.classfile.Label;
+import jdk.internal.classfile.TypeKind;
+import jdk.internal.classfile.instruction.BranchInstruction;
+import jdk.internal.classfile.instruction.ReturnInstruction;
+
+public final class CodeBuilderFinalizerImpl {
+
+    public static void buildWithFinalizer(CodeBuilder parent,
```
liach, line 40: Should we ensure that this generates code compatible to that generated by javac so it can migrate to this API? If javac plans to migrate to this API, we might need stricter specification and implementation requirements.

src/java.base/share/classes/jdk/internal/classfile/CodeBuilder.java
asotona: Conversion to anonymous classes should be done only when specifically needed or we will end with horrible codebase.
Code used during JDK bootstrap must be very carefully selected, as for example any use of BlockCodeBuilder involves several other classes and would require to expand also several switch expressions to if/then/else sequences.
I would recommend to use bare minimum (only the base builder features) from the Classfile API to construct lambdas and method handles.

src/java.base/share/classes/jdk/internal/classfile/CodeBuilder.java
```diff
@@ -1192,6 +1234,14 @@ default CodeBuilder lookupswitch(Label defaultTarget, List<SwitchCase> cases) {
         return lookupSwitchInstruction(defaultTarget, cases);
     }
 
+    default CodeBuilder lookupswitch(Consumer<SwitchBuilder> switchHandler) {
+        return block(b -> {
+            var switchBuilder = new SwitchBuilderImpl(b, true);
```
ExE-Boss, line 1239: The second parameter should’ve probably been `false` here:
```suggestion
            var switchBuilder = new SwitchBuilderImpl(b, /* tableswitch: */ false);
```

PR #14058, opened 2023-05-19
8308407: libjvm library not reproducibly comparable between vendors

andrew-m-leonard:
This PR fixes issue: https://bugs.openjdk.org/browse/JDK-8308407

It ensures that the C compilation of abstract_vm_version.o generates a minimum 64byte static C const for vm_vendor, thus ensuring for most Vendor builds the object has a 64byte string char array.

Since this is just a nul terminated string as before, just within a fixed 64byte location, and should a longer Vendor string be specified it will be longer, there is no change in behavior. It purely provides a fixed 64byte string array in the compiled object, which for most if not all Vendors will be sufficient.

---------
### Progress

### Issue
 * [JDK-8308407](https://bugs.openjdk.org/browse/JDK-8308407): libjvm library not reproducibly comparable between vendors

### Reviewers
 * [Andrew John Hughes](https://openjdk.org/census#andrew) (@gnu-andrew - **Reviewer**)

### Contributors
 * Simon Tooke `<stooke@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14058/head:pull/14058` \
`$ git checkout pull/14058`

`$ git checkout pull/14058` \
`$ git pull https://git.openjdk.org/jdk.git pull/14058/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14058`

`$ git pr show -t 14058`

</details>

<a href="https://git.openjdk.org/jdk/pull/14058.diff">https://git.openjdk.org/jdk/pull/14058.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14058#issuecomment-1554410072)


andrew-m-leonard:
Re-based due to upstream build break:

gnu-andrew:
> @andrew-m-leonard is 64 chars sufficient for the most common values found in the `vm_vendor` string? Some indication that this covers the major vendors would make it a clearer win.

`sizeof(VENDOR) < VENDOR_PADDING ? VENDOR_PADDING : sizeof(VENDOR)` means it will always be at least `VENDOR_PADDING` but may be longer. An earlier revision - see https://github.com/adoptium/jdk/pull/17 - did start to impose a 64 character limit, but the patch was revised to avoid this.

gnu-andrew (approved):
Looks good to me following the revisions on the original Adoptium PR, but we should also get a review from someone who wasn't involved in that.

gnu-andrew:
> Re-based due to upstream build break:

Did that actually require a code change or was this just to cause a re-run? You should be able to schedule a re-run from the Actions UI without rebasing.

DanHeidinga:
> > @andrew-m-leonard is 64 chars sufficient for the most common values found in the `vm_vendor` string? Some indication that this covers the major vendors would make it a clearer win.
> 
> `sizeof(VENDOR) < VENDOR_PADDING ? VENDOR_PADDING : sizeof(VENDOR)` means it will always be at least `VENDOR_PADDING` but may be longer. An earlier revision - see [adoptium/jdk#17](https://github.com/adoptium/jdk/pull/17) - did start to impose a 64 character limit, but the patch was revised to avoid this.

If  `sizeof(VENDOR)` on most distributions is >64 chars, then using 64 as a minimum provides no value.  I have no idea what the typical length of this string is and I expect most reviewers only know the value their distribution uses.  Confirming the common choices are less then 64 chars makes the reviewers lives easier as it affirms the value of the PR beyond its technical correctness.

gnu-andrew:
> > > @andrew-m-leonard is 64 chars sufficient for the most common values found in the `vm_vendor` string? Some indication that this covers the major vendors would make it a clearer win.
> > 
> > 
> > `sizeof(VENDOR) < VENDOR_PADDING ? VENDOR_PADDING : sizeof(VENDOR)` means it will always be at least `VENDOR_PADDING` but may be longer. An earlier revision - see [adoptium/jdk#17](https://github.com/adoptium/jdk/pull/17) - did start to impose a 64 character limit, but the patch was revised to avoid this.
> 
> If `sizeof(VENDOR)` on most distributions is >64 chars, then using 64 as a minimum provides no value. I have no idea what the typical length of this string is and I expect most reviewers only know the value their distribution uses. Confirming the common choices are less then 64 chars makes the reviewers lives easier as it affirms the value of the PR beyond its technical correctness.

I see your point. The default ("Oracle Corporation") and the ones I'm familiar with ("Red Hat, Inc.", "Adoptium") fit well within 64 characters, and this PR would make all three take up the same amount of space, easing comparisons between binaries from different vendors. I don't know if @andrew-m-leonard has studied more cases, but I believe a length over 64 characters is uncommon. It's also only an issue if comparing binaries with different vendor strings, and even then, it would be no different to the situation prior to this patch. So I do think this change makes common comparisons more viable and doesn't regress any existing ones.

As noted previously, the handling of lengths above 64 characters is more for potential backwards compatibility than because we've actually encountered any with a length that long in the wild i.e. we don't want to break something that was theoretically possible before.

dholmes-ora (approved):
So just to be clear for any VENDOR string > 64 the resulting libjvm still won't be comparable?

Thanks.

andrew-m-leonard:
> So just to be clear for any VENDOR string > 64 the resulting libjvm still won't be comparable?
> 
> Thanks.

@dholmes-ora Yes, correct, should a vendor use a --with-vendor-name "string", that exceeds 64 characters, such a libjvm would not be identically comparable due to the binary object differences caused by non-deterministic offsetting and 32bit padding.

PR #14061, opened 2023-05-19
8308444: LoadStoreNode::result_not_used() is too conservative

merykitty:
Hi,

This patch improves the implementation of `LoadStoreNode::result_not_used()` to be less conservative and verifies that the preferable node is matched for `getAndAdd`. Please kindly review, thanks a lot.


test/hotspot/jtreg/compiler/c2/x86/TestGetAndAdd.java
```diff
@@ -0,0 +1,112 @@
+/*
```
vnkozlov, line 1: Please, use `c2/varhandle` directory for this test. We don't use platform in dir names.
The test could be updated later for other platforms.

src/hotspot/share/opto/memnode.cpp
```diff
@@ -3014,9 +3014,15 @@ uint LoadStoreNode::ideal_reg() const {
 }
 
 bool LoadStoreNode::result_not_used() const {
```
vnkozlov, line 3016: Add comment to this method explaining what it is for and cases when it returns `true` or `false`.

test/hotspot/jtreg/compiler/c2/x86/TestGetAndAdd.java
```diff
@@ -0,0 +1,112 @@
...
+import compiler.lib.ir_framework.*;
+
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.VarHandle;
+
+/*
+ * @test
+ * bug 8308444
+ * @summary verify that the correct node is matched for atomic getAndAdd
+ * @requires os.arch=="amd64" | os.arch=="x86_64"
```
vnkozlov, line 34: You can use os.simpleArch == "x64" to cover both cases.
And put second `requires` after it so they stay together.

vnkozlov (commented):
Implementation looks good to me. I have few comments about test.
You need second review.

TobiHartmann:
I'm seeing failures with `compiler/unsafe/JdkInternalMiscUnsafeAccessTestLong.java` and `JdkInternalMiscUnsafeAccessTestInt.java`:

```
test compiler.unsafe.JdkInternalMiscUnsafeAccessTestLong.testArray(): failure
java.lang.AssertionError: getAndAdd long expected [81985529216486895] but found [33520196056]
	at org.testng.Assert.fail(Assert.java:99)
	at org.testng.Assert.failNotEquals(Assert.java:1037)
	at org.testng.Assert.assertEqualsImpl(Assert.java:140)
	at org.testng.Assert.assertEquals(Assert.java:122)
	at org.testng.Assert.assertEquals(Assert.java:797)
	at compiler.unsafe.JdkInternalMiscUnsafeAccessTestLong.testAccess(JdkInternalMiscUnsafeAccessTestLong.java:361)
	at compiler.unsafe.JdkInternalMiscUnsafeAccessTestLong.testArray(JdkInternalMiscUnsafeAccessTestLong.java:130)
```

merykitty:
@vnkozlov I have addressed the reviews, the test is moved to `irTest`.
@TobiHartmann I have fixed the error, it is because a call node may return a membar on `bottom_type()`, I excluded them in the check.
Thanks a lot.

vnkozlov (approved):
Update looks good.

merykitty:
Thanks very much for the reviews and testing. I will integrate the change.

PR #14063, opened 2023-05-19
8308452: Extend internal Architecture enum with byte order and address size

RogerRiggs:
The internal enum jdk.internal.util.Architecture does not provide information about the big or little endianness or the address size (64 or 32 bits).  The endian-ness and address size are intrinsic to the architecture.

The values of the enum are extended to separately identify the big endian and little-endian uses of the ISA.
For example, `PPC64` and `PPC64LE` for the big and little-endian versions.  
The enum values directly reflect the build-time artifacts and resulting executables.

This information about an architecture will make the enum more useful especially to identify a target platform in a cross-platform use case. A method is added to map well known aliases for the platforms to the Architecture enum.

---------
### Progress

### Issue
 * [JDK-8308452](https://bugs.openjdk.org/browse/JDK-8308452): Extend internal Architecture enum with byte order and address size (**Enhancement** - P4)

### Reviewers
 * [Jaikiran Pai](https://openjdk.org/census#jpai) (@jaikiran - **Reviewer**) ⚠️ Review applies to [2e04a1ca](https://git.openjdk.org/jdk/pull/14063/files/2e04a1cafb3b712fc2be1b3720b6022f78d670be)
 * [Mandy Chung](https://openjdk.org/census#mchung) (@mlchung - **Reviewer**)
 * [Amit Kumar](https://openjdk.org/census#amitkumar) (@offamitkumar - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14063/head:pull/14063` \
`$ git checkout pull/14063`

`$ git checkout pull/14063` \
`$ git pull https://git.openjdk.org/jdk.git pull/14063/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14063`

`$ git pr show -t 14063`

</details>

<a href="https://git.openjdk.org/jdk/pull/14063.diff">https://git.openjdk.org/jdk/pull/14063.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14063#issuecomment-1561338738)


src/java.base/share/classes/jdk/internal/util/Architecture.java
```diff
@@ -33,16 +35,65 @@
...
-    AARCH64,
-    ARM,
-    RISCV64,
-    S390,
-    PPC64,
+    OTHER(0, null),          // An unknown architecture not specifically named
+    X64(64, ByteOrder.LITTLE_ENDIAN),  // Represents AMD64 and X86_64
+    X86(32, ByteOrder.LITTLE_ENDIAN),
+    AARCH64(64, ByteOrder.LITTLE_ENDIAN),
+    ARM(64, ByteOrder.LITTLE_ENDIAN),
```
TheRealMDoerr, line 42: ARM should be 32 bit.

TheRealMDoerr (commented):
I think this makes sense. Thanks for working on it! Most architectures have a fixed endianness.
For PPC64, I think using PPC64LE is fine to distinguish as it is a common name.
Another architecture which supports both endianness flavors is IA64. HPUX is big endian and linux little endian. I don't know how this should be handled. Fortunately, I don't think we need to support IA64. :-)

src/java.base/share/classes/jdk/internal/util/Architecture.java
```diff
@@ -33,16 +35,65 @@
...
 public enum Architecture {
-    OTHER,      // An unknown architecture not specifically named
-    X64,        // Represents AMD64 and X86_64
-    X86,
-    AARCH64,
-    ARM,
-    RISCV64,
-    S390,
-    PPC64,
+    OTHER(0, null),          // An unknown architecture not specifically named
```
AlanBateman, line 38: ByteOrder.nativeOrder() might be a better default for unknown archs.

src/java.base/share/classes/jdk/internal/util/OperatingSystem.java
```diff
@@ -127,16 +127,25 @@ public static OperatingSystem current() {
...
+    }
+
+    /**
+     * {@return Simple conversion of ASCII to upper case; used when too early to use Locale conversions}
+     *
+     * @param str a string
+     * @return the string with lower case ascii converted to upper case
+     */
+    /* package-private */
+    static String toUpperCase(String str) {
```
AlanBateman, line 140: A bit icky to have OperatingSystem to expose a static method to convert Strings as it's nothing to do with OperatingSystem, is there anywhere else?

AlanBateman (commented):
Mapping the arch to endianness seems okay for the current ports but I could imagine needing to re-visit that at some point.

TheRealMDoerr:
I think the mapping from ppc64le to ppc64 should get removed with the introduction of ppc64le as separate architecture:
https://github.com/openjdk/jdk/blob/64f6681cc7bb184112adcf2b4f2c313e1b0a5c4f/src/jdk.jlink/share/classes/jdk/tools/jlink/internal/Platform.java#LL55C5-L55C5
Would you mind removing it with your change?

src/java.base/share/classes/jdk/internal/util/OperatingSystem.java
RogerRiggs: Delaying the initialization of OperatingSystem made it possible to revert to String.toUpperCase(). 
Enabled by the fix in PR [14181](https://github.com/openjdk/jdk/pull/14181).

TheRealMDoerr (commented):
Thanks for the updates! Unfortunately, there's still something going wrong on ppc64le:
- jdk/internal/util/ArchTest.java: System property os.arch: "ppc64le", Architecture.current(): "PPC64"
- tools/jlink/plugins/CDSPluginTest.java: Cannot generate CDS archives: target image platform linux-ppc64le is different from runtime platform linux-ppc64

Do you happen to know why Architecture.current() misses the "le"?

RogerRiggs:
> Do you happen to know why Architecture.current() misses the "le"?

@TheRealMDoerr  The makefile that does the substitution (in GensrcMisc.gmk) for the OPENJDK_TARGET_CPU is using the _CANONICAL form of the architecture (that maps ppc64le to ppc64).
I'll work up a fix; note that we do want the canonical form for s390x and amd/x86_64.

src/java.base/share/classes/jdk/internal/util/OperatingSystem.java
```diff
@@ -26,6 +26,8 @@
 import jdk.internal.util.PlatformProps;
 import jdk.internal.vm.annotation.ForceInline;
 
+import java.util.Locale;
```
naotoj, line 29: Nit: Stray import?

TheRealMDoerr (approved):
Thanks for the updates! LGTM.

src/java.base/share/classes/jdk/internal/util/Architecture.java
```diff
@@ -33,19 +35,81 @@
...
+    PPC64LE(64, ByteOrder.LITTLE_ENDIAN),
+    MIPSEL(32, ByteOrder.LITTLE_ENDIAN),
+    MIPS64EL(64, ByteOrder.LITTLE_ENDIAN)
     ;
 
+    private final int addrSize;
+    private final ByteOrder byteOrder;
+
+    /**
+     * Construct an Arctecture with number of address bits and byte order.
```
jaikiran, line 60: Hello Roger, there is a typo here "Arctecture" instead of "Architecture"

jaikiran (approved):
The changes in this PR look OK to me. The introduction of instance method `byteOrder()` in the `Architecture` enum, as proposed in this PR, will help with the changes being proposed in https://github.com/openjdk/jdk/pull/11943. That way jlink won't have to have its own mapping of osname-architecture to endianness.
Once this PR is integrated, I'll update that other PR to use this new API and remove the endianness mappings being proposed in that PR.

RogerRiggs:
I think this has settled down.
If anyone would care to test it on S390, PPC, LOONGARCH64, or MIPS that would be appreciated.

mlchung (approved):
This change looks fine to me.

TheRealMDoerr:
Still good on PPC64.

TheRealMDoerr (approved):

offamitkumar (approved):
Looks good for s390.

PR #14065, opened 2023-05-19
8300865: C2: product reduction in ProdRed_Double is not vectorized

sviswa7:
This PR fixes the problem with double reduction on x86_64. 

In the test compiler.loopopts.superword.ProdRed_Double, the product reduction loop in prodReductionImplement() was not getting vectorized when run as follows:
jtreg -XX:CompileCommand=PrintAssembly,compiler.loopopts.superword.ProdRed_Double::prodReductionImplement compiler/loopopts/superword/ProdRed_Double.java
The print assembly generated in the pid-xxx.log output in JTwork/scratch directory was not showing any vector_reduction_double node.

This was happening as the ReductionNode::implemented was passed a vector size of one element. For the vector reduction implemented we need to check with at least vector size of two elements.

With this PR the vector_reduction_double node is generated.

Please review.

Best Regards,
Sandhya

---------
### Progress

### Issue
 * [JDK-8300865](https://bugs.openjdk.org/browse/JDK-8300865): C2: product reduction in ProdRed_Double is not vectorized

### Reviewers
 * [Fei Gao](https://openjdk.org/census#fgao) (@fg1417 - Committer) ⚠️ Review applies to [ba3b5dfa](https://git.openjdk.org/jdk/pull/14065/files/ba3b5dfaacb3a3429288ad96ca518609088260ed)
 * [Emanuel Peter](https://openjdk.org/census#epeter) (@eme64 - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14065/head:pull/14065` \
`$ git checkout pull/14065`

`$ git checkout pull/14065` \
`$ git pull https://git.openjdk.org/jdk.git pull/14065/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14065`

`$ git pr show -t 14065`

</details>

<a href="https://git.openjdk.org/jdk/pull/14065.diff">https://git.openjdk.org/jdk/pull/14065.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14065#issuecomment-1557588208)

### DIFF
--- a/src/hotspot/share/opto/superword.cpp
+++ b/src/hotspot/share/opto/superword.cpp
@@ -3718,10 +3718,7 @@ bool SuperWord::construct_bb() {
             // First see if we can map the reduction on the given system we are on, then
             // make a data entry operation for each reduction we see.
             BasicType bt = use->bottom_type()->basic_type();
-            // Matcher::min_vector_size may return 1 in some cases, e.g. double for x86.
-            // For vector reduction implemented check we need atleast two elements.
-            int min_vec_size = MAX2(Matcher::min_vector_size(bt), 2);
-            if (ReductionNode::implemented(use->Opcode(), min_vec_size, bt)) {
+            if (ReductionNode::implemented(use->Opcode(), Matcher::max_vector_size(bt), bt)) {
               reduction_uses++;
             }
           }
--- a/src/hotspot/share/opto/superword.cpp
+++ b/src/hotspot/share/opto/superword.cpp
@@ -3718,7 +3718,7 @@ bool SuperWord::construct_bb() {
             // First see if we can map the reduction on the given system we are on, then
             // make a data entry operation for each reduction we see.
             BasicType bt = use->bottom_type()->basic_type();
-            if (ReductionNode::implemented(use->Opcode(), Matcher::max_vector_size(bt), bt)) {
+            if (ReductionNode::implemented(use->Opcode(), Matcher::superword_max_vector_size(bt), bt)) {
               reduction_uses++;
             }
           }
--- a/src/hotspot/share/opto/superword.cpp
+++ b/src/hotspot/share/opto/superword.cpp
@@ -3718,7 +3718,10 @@ bool SuperWord::construct_bb() {
             // First see if we can map the reduction on the given system we are on, then
             // make a data entry operation for each reduction we see.
             BasicType bt = use->bottom_type()->basic_type();
-            if (ReductionNode::implemented(use->Opcode(), Matcher::min_vector_size(bt), bt)) {
+            // Matcher::min_vector_size may return 1 in some cases, e.g. double for x86.
+            // For vector reduction implemented check we need atleast two elements.
+            int min_vec_size = MAX2(Matcher::min_vector_size(bt), 2);
+            if (ReductionNode::implemented(use->Opcode(), min_vec_size, bt)) {
               reduction_uses++;
             }
           }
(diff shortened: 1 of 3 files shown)


src/hotspot/share/opto/superword.cpp
```diff
@@ -3718,7 +3718,10 @@ bool SuperWord::construct_bb() {
             // First see if we can map the reduction on the given system we are on, then
             // make a data entry operation for each reduction we see.
             BasicType bt = use->bottom_type()->basic_type();
-            if (ReductionNode::implemented(use->Opcode(), Matcher::min_vector_size(bt), bt)) {
+            // Matcher::min_vector_size may return 1 in some cases, e.g. double for x86.
+            // For vector reduction implemented check we need atleast two elements.
+            int min_vec_size = MAX2(Matcher::min_vector_size(bt), 2);
+            if (ReductionNode::implemented(use->Opcode(), min_vec_size, bt)) {
```
fg1417, line 3724: Hi @sviswa7, can we use `superword_max_vector_size()` as the input here? `MAX2(Matcher::min_vector_size(bt), 2);` may not tally with the actual situation on other 64-bit platforms. WDYT?

sviswa7: @fg1417 Thanks a lot for the review. Yes, we could use superword_max_vector_size here. I have made the change.

fg1417 (approved):
LGTM

eme64:
Running testing at commit 4. Will report back.

sviswa7:
Thanks a lot @eme64.

eme64 (approved):
Thanks for the fix! Looks good.

sviswa7:
Thanks a lot @eme64. There was an existing jmh benchmark for vector reduction. I have updated it to add double reduction case.

sviswa7:
The performance numbers on my desktop are:
Base runs, no vectorization happens with superword:
Benchmark                              (COUNT)  (seed)  Mode  Cnt    Score   Error  Units
VectorReduction.NoSuperword.mulRedD        512       0  avgt    4  435.795 ± 0.082  ns/op
VectorReduction.WithSuperword.mulRedD      512       0  avgt    4  434.154 ± 0.042  ns/op

With the PR reduction succeeds and vectorization of the loop happens when superword is enabled: 
Benchmark                              (COUNT)  (seed)  Mode  Cnt    Score   Error  Units
VectorReduction.NoSuperword.mulRedD        512       0  avgt    4  435.897 ± 0.137  ns/op
VectorReduction.WithSuperword.mulRedD      512       0  avgt    4  405.479 ± 1.896  ns/op

eme64:
@sviswa7 Thanks for adding the benchmark. The win is small, but that was to be expected given that the double reduction has to be performed in a linear order, and hence has quite a large latency.

eme64 (approved):

sviswa7:
@vnkozlov Could you also please review this PR?

sviswa7:
> Looks good. You may regress for next case on KNL? when using `max vector size` (code is in x86.ad) which I think is fine.
> 
> ```
>     case Op_MinReductionV:
>     case Op_MaxReductionV:
>       if (UseAVX > 2 && (!VM_Version::supports_avx512dq() && size_in_bits == 512)) {
>         return false;
>       }
> ```

Yes, that should be ok. Thanks a lot for the review @vnkozlov.

PR #14066, opened 2023-05-20
8159023: Engineering notation of DecimalFormat does not work as documented

justin-curtis-lu:
Please review this PR which updates the Scientific Notation section of Decimal Format. It aims to resolve [JDK-8159023](https://bugs.openjdk.org/browse/JDK-8159023) as well as [JDK-6282188](https://bugs.openjdk.org/browse/JDK-6282188).

**Scientific Notation** in Decimal Format contains the definition for a scientific notation formatted number's mantissa as such: _The number of significant digits in the mantissa is the sum of the minimum integer and maximum fraction digits, and is unaffected by the maximum integer digits. For example, 12345 formatted with "##0.##E0" is "12.3E3"._

Both the definition and example are incorrect, as the actual result is "12.E345". 

The following example data show that scientific notation formatted numbers do not adhere to that definition. As, according to the definition, the following numbers should have 3 significant digits, but in reality, they have up to 5.
```
123 formatted by ##0.#E0 is 123E0
1234 formatted by ##0.#E0 is 1.234E3
12345 formatted by ##0.#E0 is 12.34E3
123456 formatted by ##0.#E0 is 123.5E3
1234567 formatted by ##0.#E0 is 1.235E6
12345678 formatted by ##0.#E0 is 12.35E6
123456789 formatted by ##0.#E0 is 123.5E6 
```

The actual definition of the mantissa, as well as examples and further context are included in this change. In addition, a test has been added that tests various patterns to numbers and ensures the format follows the new definition's mathematical expression.

---------
### Progress
- [x] Change requires CSR request [JDK-8308317](https://bugs.openjdk.org/browse/JDK-8308317) to be approved

### Issues
 * [JDK-8159023](https://bugs.openjdk.org/browse/JDK-8159023): Engineering notation of DecimalFormat does not work as documented
 * [JDK-8308317](https://bugs.openjdk.org/browse/JDK-8308317): Engineering notation of DecimalFormat does not work as documented (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14066/head:pull/14066` \
`$ git checkout pull/14066`

`$ git checkout pull/14066` \
`$ git pull https://git.openjdk.org/jdk.git pull/14066/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14066`

`$ git pr show -t 14066`

</details>

<a href="https://git.openjdk.org/jdk/pull/14066.diff">https://git.openjdk.org/jdk/pull/14066.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14066#issuecomment-1555386539)

### DIFF
--- a/test/jdk/java/text/Format/DecimalFormat/MantissaDigits.java
+++ b/test/jdk/java/text/Format/DecimalFormat/MantissaDigits.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/*
+ * @test
+ * @bug 8159023
+ * @summary Confirm behavior of mantissa for scientific notation in Decimal Format
+ * @run junit MantissaDigits
+ */
+
+import java.text.DecimalFormat;
+import java.util.Locale;
+import java.util.stream.Stream;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MantissaDigits {
+    private static final double[] numbers = {
+            1.1, 12.1, 123.1, 1234.1, 12345.1, 123456.1,
+            -1.1, -12.1, -123.1, -1234.1, -12345.1, -123456.1,
+            1, 12, 123, 1234, 12345, 123456, 1234567,
+            -1, -12, -123, -1234, -12345, -123456, -1234567,
+            1.1234, 1.1111, 1.412, 222.333, -771.2222
+
+            // The following numbers when formatted, may produce
+            // results that have leading or trailing 0s (which ultimately,
+            // are ignored, but were "included" as significant digits)
+            // 1.101013, 1.10011, 0.00101, 0.100, -1.00130, -0.00131
+            // ie: 0.00101 formatted with #0.000##E0 gives 10.10E-4
+            // which implies there was only "4" significant digits even
+            // if the mathematical expression produced 6 significant digits.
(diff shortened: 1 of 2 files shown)


test/jdk/java/text/Format/DecimalFormat/MantissaDigits.java
```diff
@@ -0,0 +1,116 @@
...
+            int zeroes = (int) (pattern.chars()
+                    // ignore exponent zero
+                    .filter(ch -> ch == '0').count() - 1);
+            int hashes = (int) pattern.chars()
+                    .filter(ch -> ch == '#').count();
+
+
+            // Test the new definition of the Mantissa
+            Integer calculatedDigits = Math
+                    .min(Math.max(zeroes, originalNumDigits), (hashes+zeroes));
```
naotoj, line 89: Could we make those expected numbers into the `Arguments` so that there is no need to calculate in the test case at runtime? I think that would be more readable.

test/jdk/java/text/Format/DecimalFormat/MantissaDigits.java
```diff
@@ -0,0 +1,116 @@
...
+            1, 12, 123, 1234, 12345, 123456, 1234567,
+            -1, -12, -123, -1234, -12345, -123456, -1234567,
+            1.1234, 1.1111, 1.412, 222.333, -771.2222
+            };
+
+    @ParameterizedTest
+    @MethodSource("patterns")
+    public void testMantissaDefinition(String pattern) {
+        Locale.setDefault(Locale.ENGLISH);
+        DecimalFormat df = new DecimalFormat(pattern);
```
naotoj, line 56: Instead of setting the default locale, you could get the explicit format by
```
new DecimalFormat(pattern, new DecimalFormatSymbols(Locale.US)); // DFS can be a static instance
```

naotoj (commented):
(a bit of background to others) We did some archaeological work to dig into the history of this spec, and found that the spec was retrofitted to the implementation. (https://bugs.openjdk.org/browse/JDK-4241670). We suspect that this wording was incorrectly interpreted from the impl at that time.

test/jdk/java/text/Format/DecimalFormat/MantissaDigits.java
justin-curtis-lu: Thanks for the review, addressed both suggestions in the most recent commit.

test/jdk/java/text/Format/DecimalFormat/MantissaDigits.java
```diff
@@ -0,0 +1,110 @@
...
+
+public class MantissaDigits {
+    private static final double[] numbers = {
+            1.1, 12.1, 123.1, 1234.1, 12345.1, 123456.1,
+            -1.1, -12.1, -123.1, -1234.1, -12345.1, -123456.1,
+            1, 12, 123, 1234, 12345, 123456, 1234567,
+            -1, -12, -123, -1234, -12345, -123456, -1234567,
+            1.1234, 1.1111, 1.412, 222.333, -771.2222
+            };
+    private static final DecimalFormatSymbols dfs = new DecimalFormatSymbols(Locale.US);
```
naotoj, line 52: The `static final` field is better uppercased.

test/jdk/java/text/Format/DecimalFormat/MantissaDigits.java
```diff
@@ -0,0 +1,110 @@
...
+    // Hard coded as 1, since all test patterns only have 1 exponent digit
+    private static final int EXPONENTDIGITS = 1;
+
+    @ParameterizedTest
+    @MethodSource("patterns")
+    public void testMantissaDefinition(String pattern, int minDigits, int maxDigits) {
+        DecimalFormat df = new DecimalFormat(pattern, dfs);
+        for (double number : numbers) {
+            // Count the significant digits in the pre-formatted number
+            int originalNumDigits = String.valueOf(number)
```
naotoj, line 64: It's OK as it stands, but you could write:
```
str.chars().filter(Character::isDigit).count()
```
which might be more readable

naotoj (commented):
Thanks, Justin. Left some super-nits.

test/jdk/java/text/Format/DecimalFormat/MantissaDigits.java
justin-curtis-lu: Like this much better, incorporated!

naotoj (approved):
Looks good

PR #14072, opened 2023-05-21
8308038: java/util/concurrent/ThreadPerTaskExecutor/ThreadPerTaskExecutorTest.java timed out

AlanBateman:
This is a test only change to the unit test for the ExecutorService returned by Executors.newThreadPerTaskExecutor. The tests for interrupting invokeAll assume the threads started to execute the tasks do actually execute the task code. The refresh in JEP 444 changed the implementation to use FutureTask, and FutureTask checks the interrupt status before it executes the task code. So some intermittent timeouts of the tests for interrupting invokeAll as those tests were waiting for the task to complete.

The main change is that the tests for interrupting invokeAll are changed to interrupt when the main thread blocks in invokeAll. They are also changed to check if the task started or not. The tests for interrupting invokeAny already did this, but these are changed to use the same infrastructure to avoid having two styles of tests in the same source file.

---------
### Progress

### Issue
 * [JDK-8308038](https://bugs.openjdk.org/browse/JDK-8308038): java/util/concurrent/ThreadPerTaskExecutor/ThreadPerTaskExecutorTest.java timed out

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14072/head:pull/14072` \
`$ git checkout pull/14072`

`$ git checkout pull/14072` \
`$ git pull https://git.openjdk.org/jdk.git pull/14072/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14072`

`$ git pr show -t 14072`

</details>

<a href="https://git.openjdk.org/jdk/pull/14072.diff">https://git.openjdk.org/jdk/pull/14072.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14072#issuecomment-1556785693)


test/jdk/java/util/concurrent/ThreadPerTaskExecutor/ThreadPerTaskExecutorTest.java
```diff
@@ -603,20 +572,19 @@ void testInvokeAnyWithInterruptSet(ExecutorService executor) throws Exception {
...
-            Callable<String> task2 = () -> {
-                Thread.sleep(Duration.ofMinutes(2));
-                return "bar";
-            };
-            scheduleInterrupt(Thread.currentThread(), Duration.ofMillis(500));
+            var task = new LongRunningTask<Void>();
             try {
-                executor.invokeAny(Set.of(task1, task2));
+                scheduleInterruptAt("java.util.concurrent.ThreadPerTaskExecutor.invokeAny");
+                executor.invokeAny(Set.of(task));
```
dfuch, line 578: Seems strange to test invokeAny when there is only one task?

test/jdk/java/util/concurrent/ThreadPerTaskExecutor/ThreadPerTaskExecutorTest.java
```diff
@@ -856,53 +823,51 @@ void testInvokeAllInterrupt3(ExecutorService executor) throws Exception {
...
     void testInvokeAllInterrupt4(ExecutorService executor) throws Exception {
         try (executor) {
-            Callable<String> task1 = () -> "foo";
-            DelayedResult<String> task2 = new DelayedResult("bar", Duration.ofMinutes(1));
-            scheduleInterrupt(Thread.currentThread(), Duration.ofMillis(500));
+            var task = new LongRunningTask<Void>();
             try {
-                executor.invokeAll(Set.of(task1, task2));
+                scheduleInterruptAt("java.util.concurrent.ThreadPerTaskExecutor.invokeAll");
+                executor.invokeAll(Set.of(task));
```
dfuch, line 835: same remark for invokeAll?

dfuch (commented):
Looks reasonable. Just some question about testing invokeAll/invokeAny when there is only one task.

test/jdk/java/util/concurrent/ThreadPerTaskExecutor/ThreadPerTaskExecutorTest.java
```diff
@@ -261,17 +253,20 @@ void testClose3(ExecutorService executor) throws Exception {
...
         } finally {
             assertTrue(Thread.interrupted());
         }
         assertTrue(executor.isShutdown());
         assertTrue(executor.isTerminated());
-        assertTrue(executor.awaitTermination(10,  TimeUnit.MILLISECONDS));
-        assertThrows(ExecutionException.class, future::get);
+        assertTrue(executor.awaitTermination(10, TimeUnit.MILLISECONDS));
+        Throwable e = assertThrows(ExecutionException.class, future::get);
+        assertTrue(e.getCause() instanceof InterruptedException);
```
jaikiran, line 269: Would using `assertEquals(InterruptedException.class, e.getCause().getClass())` be better? That way if/when the test fails, it even prints the unexpected exception type?

But I then see that this test already uses `assertTrue` for similar cases in some other place, so maybe it's fine in the current form?

test/jdk/java/util/concurrent/ThreadPerTaskExecutor/ThreadPerTaskExecutorTest.java
AlanBateman: > Seems strange to test invokeAny when there is only one task?

There are several tests further up that exercise invokeAny with more than one task. This one is testing interrupt of invokeAny, we could expect it to test interrupting with many tasks if needed.

test/jdk/java/util/concurrent/ThreadPerTaskExecutor/ThreadPerTaskExecutorTest.java
AlanBateman: I don't mind moving this but it would require changing them everywhere (as you point out).

jaikiran (approved):
The change now waits for the current thread to reach a particular code location before interrupting the thread and then also verifies that the submitted task was indeed started, before checking that it is done. This I believe should solve the intermittent failures observed in this test.

test/jdk/java/util/concurrent/ThreadPerTaskExecutor/ThreadPerTaskExecutorTest.java
dfuch: OK - looks fine to me then.

dfuch (approved):

test/jdk/java/util/concurrent/ThreadPerTaskExecutor/ThreadPerTaskExecutorTest.java
AlanBateman: Yes, as we really want an instanceof check here, the exception could be a subclass.

PR #14075, opened 2023-05-22
8308500: ZStatSubPhase::register_start should not call register_gc_phase_start if ZAbort::should_abort()

xmas92:
`ZStatSubPhase::register_start` should not call `register_gc_phase_start` if `ZAbort::should_abort()` is true. This will cause an unbalanced push and pop behaviour of the phase stack as `ZStatSubPhase::register_end` stops popping (and sending events) after the aborting has started. This will create an issue if more subsequent sub-phases are added in-between two abort points as the phase stack may overflow. 

Simply avoid pushing new phases when aborting has started solves this issue.

---------
### Progress

### Issue
 * [JDK-8308500](https://bugs.openjdk.org/browse/JDK-8308500): ZStatSubPhase::register_start should not call register_gc_phase_start if ZAbort::should_abort()

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14075/head:pull/14075` \
`$ git checkout pull/14075`

`$ git checkout pull/14075` \
`$ git pull https://git.openjdk.org/jdk.git pull/14075/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14075`

`$ git pr show -t 14075`

</details>

<a href="https://git.openjdk.org/jdk/pull/14075.diff">https://git.openjdk.org/jdk/pull/14075.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14075#issuecomment-1556706162)

### DIFF
--- a/src/hotspot/share/gc/z/zStat.cpp
+++ b/src/hotspot/share/gc/z/zStat.cpp
@@ -808,7 +808,7 @@ ZStatSubPhase::ZStatSubPhase(const char* name, ZGenerationId id)
   : ZStatPhase(id == ZGenerationId::young ? "Young Subphase" : "Old Subphase", name) {}
 
 void ZStatSubPhase::register_start(ConcurrentGCTimer* timer, const Ticks& start) const {
-  if (timer != nullptr) {
+  if (timer != nullptr && !ZAbort::should_abort()) {
     assert(!Thread::current()->is_Worker_thread(), "Unexpected timer value");
     timer->register_gc_phase_start(name(), start);
   }


src/hotspot/share/gc/z/zStat.cpp
```diff
@@ -808,7 +808,7 @@ ZStatSubPhase::ZStatSubPhase(const char* name, ZGenerationId id)
   : ZStatPhase(id == ZGenerationId::young ? "Young Subphase" : "Old Subphase", name) {}
 
 void ZStatSubPhase::register_start(ConcurrentGCTimer* timer, const Ticks& start) const {
-  if (timer != nullptr) {
+  if (timer != nullptr && !ZAbort::should_abort()) {
```
stefank, line 811: `register_end` also skips the logging part if we should abort. Should we do the same for the `register_start`?

stefank (changes requested):

src/hotspot/share/gc/z/zStat.cpp
xmas92: My initial implementation had that. (just an early return)
But thought it might be interesting to see if there ever is an issue with the abort code to have the phase starts in the log. The end logging is less interesting as we are aborting. This is how single generation also does it.
Given that we only log sub phases on debug, maybe this is less useful as we will miss it in the normal gc* logs.

stefank (approved):
OK. Then let's go with your current proposal.

fisk (approved):
Looks good.

xmas92:
Thanks for the reviews.
/integrate

PR #14077, opened 2023-05-22
8308507: G1: GClocker induced GCs can starve threads requiring memory leading to OOME

walulyai:
Please review this change which fixes the thread starvation problem during allocation for G1.

The starvation problem is not limited to GCLocker, however, currently, it manifests as an OOME only when GCLocker is active. In other cases, the starvation only affects the "starved" thread as it may loop indefinitely. 

Starvation with an active GCLocker happens as below:

1. Thread A tries to allocate memory as normal, and tries to start a GC; the GCLocker is active and so the thread gets stalled waiting for the GC.
2. GCLocker induced GC executes and frees some memory.
3. Thread A does not get any of that memory, but other threads also waiting for memory.
4. Goto 1 until the gclocker retry count has been reached.

In this change, we take the general approach to solving starvation problems with announcement tables (request queues). On slow allocation, a thread that wishes to complete an Allocation GC and then attempt an allocation, announces its allocation request before proceeding to participate in a race to execute a GC safepoint. Whichever thread succeeds in executing the Allocation GC safepoint will be tasked with completing all allocation requests that were announced before the safepoint. This guarantees that all announced allocation requests are either satisfied during the safepoint, or failed in case there is not enough memory to complete all requests. This effectively deals with the starvation issue and reduces the number of allocation GCs triggered.

Note: The change also adopts ZList from ZGC and makes it available under utilities as DoublyLinkedList with slight modifications. 

Testing: Tier 1-7

---------
### Progress

### Integration blocker
&nbsp;⚠️ Whitespace errors (failed with the updated jcheck configuration)

### Issue
 * [JDK-8308507](https://bugs.openjdk.org/browse/JDK-8308507): G1: GClocker induced GCs can starve threads requiring memory leading to OOME (**Bug** - P3)(⚠️ The fixVersion in this issue is [22] but the fixVersion in .jcheck/conf is 21, a new backport will be created when this pr is integrated.)

### Reviewers
 * [Thomas Schatzl](https://openjdk.org/census#tschatzl) (@tschatzl - **Reviewer**) ⚠️ Review applies to [51fcf016](https://git.openjdk.org/jdk/pull/14077/files/51fcf01632bacf8ff618b714b150efb6886fe76e)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14077/head:pull/14077` \
`$ git checkout pull/14077`

`$ git checkout pull/14077` \
`$ git pull https://git.openjdk.org/jdk.git pull/14077/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14077`

`$ git pr show -t 14077`

</details>

<a href="https://git.openjdk.org/jdk/pull/14077.diff">https://git.openjdk.org/jdk/pull/14077.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14077#issuecomment-1557079522)

### DIFF
--- a/src/hotspot/share/gc/g1/g1Allocator.cpp
+++ b/src/hotspot/share/gc/g1/g1Allocator.cpp
@@ -70,6 +70,10 @@ G1Allocator::~G1Allocator() {
 #ifdef ASSERT
 bool G1Allocator::has_mutator_alloc_region() {
   uint node_index = current_node_index();
+  return has_mutator_alloc_region(node_index);
+}
+
+bool G1Allocator::has_mutator_alloc_region(uint node_index) {
   return mutator_alloc_region(node_index)->get() != nullptr;
 }
 #endif
--- a/src/hotspot/share/gc/g1/g1Allocator.cpp
+++ b/src/hotspot/share/gc/g1/g1Allocator.cpp
@@ -67,11 +67,6 @@ G1Allocator::~G1Allocator() {
   FREE_C_HEAP_ARRAY(SurvivorGCAllocRegion, _survivor_gc_alloc_regions);
 }
 
-#ifdef ASSERT
-bool G1Allocator::has_mutator_alloc_region(uint node_index) {
-  return mutator_alloc_region(node_index)->get() != nullptr;
-}
-#endif
 
 void G1Allocator::init_mutator_alloc_regions() {
   for (uint i = 0; i < _num_alloc_regions; i++) {
--- a/src/hotspot/share/gc/g1/g1Allocator.cpp
+++ b/src/hotspot/share/gc/g1/g1Allocator.cpp
@@ -68,11 +68,6 @@ G1Allocator::~G1Allocator() {
 }
 
 #ifdef ASSERT
-bool G1Allocator::has_mutator_alloc_region() {
-  uint node_index = current_node_index();
-  return has_mutator_alloc_region(node_index);
-}
-
 bool G1Allocator::has_mutator_alloc_region(uint node_index) {
   return mutator_alloc_region(node_index)->get() != nullptr;
 }
--- a/src/hotspot/share/gc/g1/g1CollectedHeap.cpp
+++ b/src/hotspot/share/gc/g1/g1CollectedHeap.cpp
@@ -416,8 +416,7 @@ bool G1CollectedHeap::attempt_allocation_after_gc(size_t word_size,
   }
 
   if (should_try_gc) {
-    bool succeeded = false;
-    do_collection_pause(word_size, gc_count_before, &succeeded, gc_cause);
+    do_collection_pause(word_size, gc_count_before, gc_cause);
   } else {
     log_trace(gc, alloc)("%s: Stall until clear", Thread::current()->name());
     // The GCLocker is either active or the GCLocker initiated
@@ -716,11 +715,8 @@ HeapWord* G1CollectedHeap::attempt_allocation_humongous(size_t word_size) {
 }
 
 HeapWord* G1CollectedHeap::attempt_allocation_at_safepoint(size_t word_size,
-                                                           uint node_index,
-                                                           bool expect_null_mutator_alloc_region) {
+                                                           uint node_index) {
   assert_at_safepoint_on_vm_thread();
-  assert(!_allocator->has_mutator_alloc_region(node_index) || !expect_null_mutator_alloc_region,
-         "the current alloc region was unexpectedly found to be non-null");
 
   if (!is_humongous(word_size)) {
     return _allocator->attempt_allocation_locked(word_size, node_index);
@@ -920,13 +916,11 @@ void G1CollectedHeap::resize_heap_if_necessary() {
 }
 
 HeapWord* G1CollectedHeap::satisfy_failed_allocation_helper(size_t word_size,
-                                                            uint node_index,
-                                                            bool expect_null_mutator_alloc_region) {
+                                                            uint node_index) {
   // Let's attempt the allocation first.
   HeapWord* result =
     attempt_allocation_at_safepoint(word_size,
-                                    node_index,
-                                    expect_null_mutator_alloc_region);
+                                    node_index);
   if (result != nullptr) {
     return result;
   }
@@ -937,8 +931,7 @@ HeapWord* G1CollectedHeap::satisfy_failed_allocation_helper(size_t word_size,
   // do something smarter than full collection to satisfy a failed alloc.)
   if (expand(word_size)) {
     return attempt_allocation_at_safepoint(word_size,
-                                           node_index,
-                                           expect_null_mutator_alloc_region);
+                                           node_index);
 
   }
   return nullptr;
@@ -1024,12 +1017,6 @@ void G1CollectedHeap::reset_allocation_requests() {
 bool G1CollectedHeap::handle_allocation_requests(bool expect_null_mutator_alloc_region) {
   assert_at_safepoint_on_vm_thread();
 
-  const uint active_numa_nodes = G1NUMA::numa()->num_active_nodes();
-  bool *expect_null_alloc_regions = (bool*)alloca(active_numa_nodes * sizeof(bool));
-  for (uint i = 0; i < active_numa_nodes; i++) {
-    expect_null_alloc_regions[i] = expect_null_mutator_alloc_region;
-  }
-
   for (StalledAllocReq* alloc_req : _stalled_allocations) {
     if (alloc_req->failed()) {
       // If an allocation request was declared failed, we maintain the failed state.
@@ -1040,19 +1027,19 @@ bool G1CollectedHeap::handle_allocation_requests(bool expect_null_mutator_alloc_
 
     HeapWord* result =
       satisfy_failed_allocation_helper(alloc_req->size(),
-                                       alloc_req->node_index(),
-                                       expect_null_alloc_regions[alloc_req->node_index()]);
+                                       alloc_req->node_index());
 
     if (result == nullptr) {
       // Failed to allocate, give up.
       return false;
     }
 
-    expect_null_alloc_regions[alloc_req->node_index()] = false;
-
     // Allocation succeeded, update the state and result of the allocation request.
     alloc_req->set_state(StalledAllocReq::AllocationState::Success, result);
 
+    // Initialize memory locations with filler objects such that the heap is passable
+    // even when there is a delay between the allocation and object initialization
+    // by the requesting mutator thread.
     if (is_humongous(alloc_req->size())) {
       // Calculate payload size and initialize the humongous object with a fillerArray.
       size_t words = alloc_req->size();
@@ -1068,7 +1055,6 @@ bool G1CollectedHeap::handle_allocation_requests(bool expect_null_mutator_alloc_
       policy()->old_gen_alloc_tracker()->
           record_collection_pause_humongous_allocation(size_in_regions * HeapRegion::GrainBytes);
     } else {
-      // Fill the allocated memory with filler objects.
       CollectedHeap::fill_with_objects(result, alloc_req->size());
     }
   }
@@ -2376,7 +2362,6 @@ void G1CollectedHeap::verify_numa_regions(const char* desc) {
 
 HeapWord* G1CollectedHeap::do_collection_pause(size_t word_size,
                                                uint gc_count_before,
-                                               bool* succeeded,
                                                GCCause::Cause gc_cause) {
   assert_heap_not_locked_and_not_at_safepoint();
   VM_G1CollectForAllocation op(word_size, gc_count_before, gc_cause);
@@ -2386,7 +2371,6 @@ HeapWord* G1CollectedHeap::do_collection_pause(size_t word_size,
   bool ret_succeeded = op.prologue_succeeded() && op.gc_succeeded();
   assert(result == nullptr || ret_succeeded,
          "the result should be null if the VM did not succeed");
-  *succeeded = ret_succeeded;
 
   assert_heap_not_locked();
   return result;
--- a/src/hotspot/share/gc/g1/g1CollectedHeap.cpp
+++ b/src/hotspot/share/gc/g1/g1CollectedHeap.cpp
@@ -442,8 +442,7 @@ bool G1CollectedHeap::attempt_allocation_after_gc(size_t word_size,
     assert(request.state() == StalledAllocReq::AllocationState::Failed, "Sanity check!");
     // VM successfully scheduled a collection which failed to allocate. No
     // point in trying to allocate further. We'll just return null.
-    log_debug(gc, alloc)("%s: Failed to allocate "
-                         SIZE_FORMAT " words", Thread::current()->name(), word_size);
+    log_debug(gc, alloc)("%s: Failed to allocate %zu words", Thread::current()->name(), word_size);
     *result = nullptr;
   }
   return true;
@@ -739,13 +738,6 @@ HeapWord* G1CollectedHeap::attempt_allocation_at_safepoint(size_t word_size,
   ShouldNotReachHere();
 }
 
-HeapWord* G1CollectedHeap::attempt_allocation_at_safepoint(size_t word_size,
-                                                           bool expect_null_mutator_alloc_region) {
-  assert_at_safepoint_on_vm_thread();
-  uint node_index = _allocator->current_node_index();
-  return attempt_allocation_at_safepoint(word_size, expect_null_mutator_alloc_region, node_index);
-}
-
 class PostCompactionPrinterClosure: public HeapRegionClosure {
 private:
   G1HRPrinter* _hr_printer;
@@ -978,8 +970,8 @@ bool G1CollectedHeap::satisfy_failed_allocations(bool* gc_succeeded) {
     return alloc_succeeded;
   }
 
-  // Attempt to satisfy allocation requests failed; reset the requests, execute a full gc,
-  // then try again.
+  // Attempt to satisfy allocation requests failed; reset the satisfied requests,
+  // execute a full gc, then try again.
   reset_allocation_requests();
 
   *gc_succeeded = do_full_collection(false /* clear_all_soft_refs */, false /* do_maximal_compaction */);
@@ -1085,7 +1077,7 @@ bool G1CollectedHeap::handle_allocation_requests(bool expect_null_mutator_alloc_
 
       const size_t payload_size = words - CollectedHeap::filler_array_hdr_size();
       const size_t len = payload_size * HeapWordSize / sizeof(jint);
-      assert((int)len >= 0, "size too large " SIZE_FORMAT " becomes %d", words, (int)len);
+      assert((int)len >= 0, "size too large %zu becomes %d", words, (int)len);
 
       ObjArrayAllocator allocator(Universe::fillerArrayKlassObj(), words, (int)len, /* do_zero */ false);
       allocator.initialize(result);
@@ -1110,8 +1102,7 @@ bool G1CollectedHeap::expand(size_t word_size) {
   _verifier->verify_region_sets_optional();
 
   size_t expand_bytes = MAX2(word_size * HeapWordSize, MinHeapDeltaBytes);
-  log_debug(gc, ergo, heap)("Attempt heap expansion (allocation request failed). Allocation request: " SIZE_FORMAT "B",
-                            word_size * HeapWordSize);
+  log_debug(gc, ergo, heap)("Attempt heap expansion. Expansion request: %zuB", expand_bytes);
 
   if (expand(expand_bytes, _workers)) {
     _hrm.verify_optional();
@@ -1119,16 +1110,6 @@ bool G1CollectedHeap::expand(size_t word_size) {
     return true;
   }
   return false;
-}
-
-HeapWord* G1CollectedHeap::expand_and_allocate(size_t word_size) {
-  assert_at_safepoint_on_vm_thread();
-
-  if (expand(word_size)) {
-    return attempt_allocation_at_safepoint(word_size,
-                                           false /* expect_null_mutator_alloc_region */);
-  }
-  return nullptr;
 }
 
 bool G1CollectedHeap::expand(size_t expand_bytes, WorkerThreads* pretouch_workers, double* expand_time_ms) {
--- a/src/hotspot/share/gc/g1/g1CollectedHeap.cpp
+++ b/src/hotspot/share/gc/g1/g1CollectedHeap.cpp
@@ -916,11 +916,11 @@ void G1CollectedHeap::resize_heap_if_necessary() {
 }
 
 HeapWord* G1CollectedHeap::satisfy_failed_allocation_helper(size_t word_size,
-                                                            uint node_index) {
+                                                            uint node_index,
+                                                            bool expand_heap) {
   // Let's attempt the allocation first.
-  HeapWord* result =
-    attempt_allocation_at_safepoint(word_size, node_index);
-  if (result != nullptr) {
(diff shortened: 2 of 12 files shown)


src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -947,6 +903,9 @@ void G1CollectedHeap::do_full_collection(bool clear_all_soft_refs) {
 
 bool G1CollectedHeap::upgrade_to_full_collection() {
   GCCauseSetter compaction(this, GCCause::_g1_compaction_pause);
+  // Reset any allocated but yet claimed allocation requests.
```
tschatzl, line 906:
```suggestion
  // Reset any allocated but yet unclaimed allocation requests.
```

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -989,82 +946,153 @@ HeapWord* G1CollectedHeap::satisfy_failed_allocation_helper(size_t word_size,
...
-                                     true,  /* do_gc */
-                                     false, /* maximum_collection */
-                                     false, /* expect_null_mutator_alloc_region */
-                                     succeeded);
+  bool success = handle_allocation_requests(false /* expect_null_mutator_alloc_region*/);
 
-  if (result != nullptr || !*succeeded) {
-    return result;
+  if (success) {
+   return success;
```
tschatzl, line 964:
```suggestion
    return success;
```

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -989,82 +946,153 @@ HeapWord* G1CollectedHeap::satisfy_failed_allocation_helper(size_t word_size,
...
-                                            false, /* do_gc */
-                                            false, /* maximum_collection */
-                                            true,  /* expect_null_mutator_alloc_region */
-                                            succeeded);
+  success = handle_allocation_requests(true /* expect_null_mutator_alloc_region*/);
 
-  if (result != nullptr) {
-    return result;
+  if (success) {
+   return success;
```
tschatzl, line 980:
```suggestion
    return success;
```

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -989,82 +946,153 @@ HeapWord* G1CollectedHeap::satisfy_failed_allocation_helper(size_t word_size,
...
   assert_at_safepoint_on_vm_thread();
 
-  // Attempts to allocate followed by Full GC.
-  HeapWord* result =
-    satisfy_failed_allocation_helper(word_size,
-                                     true,  /* do_gc */
-                                     false, /* maximum_collection */
-                                     false, /* expect_null_mutator_alloc_region */
-                                     succeeded);
+  bool success = handle_allocation_requests(false /* expect_null_mutator_alloc_region*/);
```
tschatzl, line 961: Maybe it would be good to rename `success` to something more specific to avoid confusion with `gc_succeeded`. Something like `alloc_succeeded` (compared to succeeding the gc and succeeding the entire method).

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -989,82 +946,153 @@ HeapWord* G1CollectedHeap::satisfy_failed_allocation_helper(size_t word_size,
...
 
-  if (result != nullptr) {
-    return result;
+  if (success) {
+   return success;
+  }
+
+  // Attempt to satisfy allocation requests after full-gc also failed. We reset the allocation requests
+  // then execute a maximal compaction full-gc before retrying the allocations
+
```
tschatzl, line 985: Superfluous newline (compared to other occurrences of similar comments)

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -989,82 +946,153 @@ HeapWord* G1CollectedHeap::satisfy_failed_allocation_helper(size_t word_size,
...
+   return success;
+  }
+
+  // Even after maximal compaction full gc, we failed to satisfy all the pending allocation requests.
+  // We can declare the pending requests as failed.
+  DoublyLinkedList<StalledAllocReq>::RemoveIterator iter(&_stalled_allocations);
+
+  for (StalledAllocReq* alloc_req; iter.next(&alloc_req);) {
+    _satisfied_allocations.insert_last(alloc_req);
+    alloc_req->set_state(StalledAllocReq::AllocationState::Failed);
```
tschatzl, line 1006: Maybe that's just me but I prefer code changing the request first and then inserting it. I.e. process then pass on to something else instead of the other way around.

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -989,82 +946,153 @@ HeapWord* G1CollectedHeap::satisfy_failed_allocation_helper(size_t word_size,
...
+  assert_at_safepoint_on_vm_thread();
 
-HeapWord* G1CollectedHeap::expand_and_allocate(size_t word_size) {
+  const uint active_numa_nodes = G1NUMA::numa()->num_active_nodes();
+  bool *expect_null_alloc_regions = (bool *)alloca(active_numa_nodes * sizeof(bool));
+  for (uint i = 0; i < active_numa_nodes; i++) {
+    expect_null_alloc_regions[i] = expect_null_mutator_alloc_region;
+  }
+
+  while(true) {
```
tschatzl, line 1047:
```suggestion
  while (true) {
```

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -989,82 +946,153 @@ HeapWord* G1CollectedHeap::satisfy_failed_allocation_helper(size_t word_size,
...
-// Attempting to expand the heap sufficiently
-// to support an allocation of the given "word_size".  If
-// successful, perform the allocation and return the address of the
-// allocated block, or else null.
+bool G1CollectedHeap::handle_allocation_requests(bool expect_null_mutator_alloc_region) {
+  assert_at_safepoint_on_vm_thread();
 
-HeapWord* G1CollectedHeap::expand_and_allocate(size_t word_size) {
+  const uint active_numa_nodes = G1NUMA::numa()->num_active_nodes();
+  bool *expect_null_alloc_regions = (bool *)alloca(active_numa_nodes * sizeof(bool));
```
tschatzl, line 1042:
```suggestion
  bool *expect_null_alloc_regions = (bool*)alloca(active_numa_nodes * sizeof(bool));
```

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -989,82 +946,153 @@ HeapWord* G1CollectedHeap::satisfy_failed_allocation_helper(size_t word_size,
...
+    if (alloc_req == nullptr) {
+      // No more pending requests, all allocations succeeded
+      return true;
+    }
+
+    HeapWord* result =
+      satisfy_failed_allocation_helper(alloc_req->size(),
+                                       alloc_req->node_index(),
+                                       expect_null_alloc_regions[alloc_req->node_index()]
+                                       );
```
tschatzl, line 1058:
```suggestion
                                       expect_null_alloc_regions[alloc_req->node_index()]);
```

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -989,82 +946,153 @@ HeapWord* G1CollectedHeap::satisfy_failed_allocation_helper(size_t word_size,
...
+
+    if (result == nullptr) {
+      // Failed to allocate, give up.
+      return false;
+    }
+
+    expect_null_alloc_regions[alloc_req->node_index()] = false;
+
+    // Allocation succeeded, update the state and result of the allocation request
```
tschatzl, line 1067:
```suggestion
    // Allocation succeeded, update the state and result of the allocation request.
```

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -989,82 +946,153 @@ HeapWord* G1CollectedHeap::satisfy_failed_allocation_helper(size_t word_size,
...
+    alloc_req->set_state(StalledAllocReq::AllocationState::Success, result);
+
+    if (is_humongous(alloc_req->size())) {
+      // Calculate payload size and initialize the humongous object with a fillerArray
```
tschatzl, line 1071:
```suggestion
      // Calculate payload size and initialize the humongous object with a fillerArray.
```

It is nowhere explained why this is needed.

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -989,82 +946,153 @@ HeapWord* G1CollectedHeap::satisfy_failed_allocation_helper(size_t word_size,
...
+
+      size_t size_in_regions = humongous_obj_size_in_regions(alloc_req->size());
+      policy()->old_gen_alloc_tracker()->
+          record_collection_pause_humongous_allocation(size_in_regions * HeapRegion::GrainBytes);
+    } else {
+      // Fill the allocated memory with filler objects
+      CollectedHeap::fill_with_objects(result, alloc_req->size());
+    }
+
+    // Move the allocation request from stalled to satisfied list
```
tschatzl, line 1089:
```suggestion
    // Move the allocation request from stalled to satisfied list.
```

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -1073,10 +1101,22 @@ HeapWord* G1CollectedHeap::expand_and_allocate(size_t word_size) {
...
-
   if (expand(expand_bytes, _workers)) {
     _hrm.verify_optional();
     _verifier->verify_region_sets_optional();
+    return true;
+  }
+  return false;
+}
+
+// Attempting to expand the heap sufficiently
```
tschatzl, line 1112:
```suggestion
// Attempts to expand the heap sufficiently
```
* the formatting of this comment paragraph is very "jagged"
* this looks like a comment that should go into the .hpp file

src/hotspot/share/gc/g1/g1CollectedHeap.hpp
```diff
@@ -168,6 +169,53 @@ class G1CollectedHeap : public CollectedHeap {
   // Testing classes.
   friend class G1CheckRegionAttrTableClosure;
 
+  class StalledAllocReq : public DoublyLinkedListNode {
```
tschatzl, line 172: Please add a short comment about it; maybe it would even be useful to describe the lifecycle (state transitions) of the allocation request here.

src/hotspot/share/gc/g1/g1CollectedHeap.hpp
```diff
@@ -168,6 +169,53 @@ class G1CollectedHeap : public CollectedHeap {
...
+  public:
+    enum class AllocationState {
```
tschatzl, line 174:
```suggestion

    enum class AllocationState {
```

src/hotspot/share/gc/g1/g1CollectedHeap.hpp
```diff
@@ -168,6 +169,53 @@ class G1CollectedHeap : public CollectedHeap {
...
+      Success,
+      Failed,
+      Pending,
+    };
+    StalledAllocReq(size_t size, uint numa_node) :
```
tschatzl, line 179:
```suggestion

    StalledAllocReq(size_t size, uint numa_node) :
```

src/hotspot/share/gc/g1/g1ConcurrentMark.cpp
```diff
@@ -1106,6 +1106,10 @@ class G1UpdateRemSetTrackingBeforeRebuildTask : public WorkerTask {
 
       bool selected_for_rebuild;
       if (hr->is_humongous()) {
+        oop obj = cast_to_oop(hr->humongous_start_region()->bottom());
+        if (G1CollectedHeap::is_obj_filler(obj)) { // Object allocated, but not well-formed
```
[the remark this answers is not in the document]
tschatzl: Just an idea: I think it would be nicer and more understandable if the code explicitly checked if the object is part of the stalled allocation requests instead of relying on it being a filler object.
It is a bit ugly that we use filler objects to identify objects in construction here, given that "nobody should have a reference to it" (I see the need). It is somewhat okay since it's GC.
Another option would be having dedicated "object-in-construction" objects, which is more work.

src/hotspot/share/gc/shared/collectedHeap.inline.hpp
```diff
@@ -46,4 +46,12 @@ inline oop CollectedHeap::class_allocate(Klass* klass, size_t size, TRAPS) {
   return allocator.allocate();
 }
 
+size_t CollectedHeap::filler_array_hdr_size() {
+  return align_object_offset(arrayOopDesc::header_size(T_INT)); // align to Long
+}
+
+size_t CollectedHeap::filler_array_min_size() {
+  return align_object_size(filler_array_hdr_size()); // align to MinObjAlignment
```
tschatzl, line 54:
```suggestion
  return align_object_size(filler_array_hdr_size()); // Align to MinObjAlignment.
```
(pre-existing)

test/hotspot/jtreg/gc/TestAllocHumongousFragment.java
```diff
@@ -172,8 +172,7 @@
  * @requires vm.gc.G1
  * @library /test/lib
  *
- * @run main/othervm -Xlog:gc+region=trace -XX:+UnlockDiagnosticVMOptions -XX:+UnlockExperimentalVMOptions -Xmx1g -Xms1g
- *      -XX:VerifyGCType=full -XX:+VerifyDuringGC -XX:+VerifyAfterGC
+ * @run main/othervm -Xlog:gc -XX:+UnlockDiagnosticVMOptions -XX:+UnlockExperimentalVMOptions -Xmx1g -Xms1g
```
tschatzl, line 175: Is this change intentional?

tschatzl (changes requested):
Initial pass/comments.

src/hotspot/share/gc/g1/g1VMOperations.cpp
```diff
@@ -125,33 +127,40 @@ VM_G1CollectForAllocation::VM_G1CollectForAllocation(size_t         word_size,
...
-    _result = g1h->attempt_allocation_at_safepoint(_word_size,
-                                                   false /* expect_null_cur_alloc_region */);
-    if (_result != nullptr) {
-      // If we can successfully allocate before we actually do the
-      // pause then we will consider this pause successful.
-      _gc_succeeded = true;
-      return;
+  // Any allocation requests that were handled during a previous GC safepoint but have not been observed
+  // by the requesting mutator thread should be reset to pending. This makes it easier for the current GC to
+  // treat the unclaimed memory as garbage.
```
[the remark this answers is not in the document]
walulyai: This would create a dependence between handle_allocation_requests, and any collections that happen before. So we would have to deal with how these `fillerObjects` are treated by the collections, some could be invalidated.

src/hotspot/share/gc/g1/g1ConcurrentMark.cpp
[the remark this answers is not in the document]
walulyai: In the worst case, searching the satisfied allocation requests list might have performance implications. However, that is speculation for now, I haven't tried it.

src/hotspot/share/gc/g1/g1VMOperations.cpp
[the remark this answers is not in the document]
tschatzl: In the general case this is true, but here, before GC occurred, the existing allocations can still be considered valid until we know that we need a gc.
Doing so potentially avoids garbage collections because then we are not going to throw away all of them and try to reallocate.

I.e. the suggestion is like:
```
// keep existing allocations, they are still valid here.

if (has_allocations) {
  if (try_satisfy_allocations()) {  // satisfy non-satisfied allocations.
    return; // no gc occurred, both the previously satisfied allocations and the new allocations are valid.
  }
}
// We are about to do a GC. Reset all allocation requests since we are likely going to free regions containing them.
reset_allocation_requests();

// do gc
```
instead of the current code
```
reset_allocation_requests(); // throw away all requests, even already satisfied ones

if (has_allocations) {
  if (try_satisfy_allocations()) { // satisfy all allocations again
    return; // done, all satisfied (no gc occurred)
  }
  reset_allocation_requests(); // throw away all requests, prepare for GC
}
```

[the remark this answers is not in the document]
walulyai: Agreed!

tschatzl (changes requested):

src/hotspot/share/gc/g1/g1Allocator.cpp
```diff
@@ -70,6 +70,10 @@ G1Allocator::~G1Allocator() {
 #ifdef ASSERT
 bool G1Allocator::has_mutator_alloc_region() {
   uint node_index = current_node_index();
+  return has_mutator_alloc_region(node_index);
+}
+
```
tschatzl, line 75: Seems unused now and can be removed.

tschatzl (approved):
Lgtm.

Please fix the remaining additional minor nits before pushing. Some of the method descriptions need to be adapted (maybe add mention of the `node_index` param and such) too.

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -400,6 +402,52 @@ G1CollectedHeap::mem_allocate(size_t word_size,
...
+
+  StalledAllocReq request(word_size, _allocator->current_node_index());
+
+  {
+    MutexLocker ml(&_alloc_request_lock, Mutex::_no_safepoint_check_flag);
+    _stalled_allocations.insert_last(&request);
+  }
+
+  if (should_try_gc) {
+    bool succeeded = false;
```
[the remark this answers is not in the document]
walulyai: forced by existing implementation of `G1CollectedHeap::do_collection_pause`, I sure can change that and remove the need for the variable.

src/hotspot/share/gc/g1/g1CollectedHeap.hpp
```diff
@@ -168,6 +169,60 @@ class G1CollectedHeap : public CollectedHeap {
...
+  private:
+    const size_t _size;
+    HeapWord* _result;
+    const uint _node_index;
+    AllocationState _state;
+  };
+
+  Mutex _alloc_request_lock;
+  DoublyLinkedList<StalledAllocReq> _stalled_allocations;
+  DoublyLinkedList<StalledAllocReq> _satisfied_allocations;
```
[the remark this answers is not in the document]
walulyai: I thought it would be easier to follow with the two lists. I can merge them, If that is easier.

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -451,60 +499,9 @@ HeapWord* G1CollectedHeap::attempt_allocation_slow(size_t word_size) {
...
-    // stalled due to the GC locker. In either can we should retry the
-    // allocation attempt in case another thread successfully
-    // performed a collection and reclaimed enough space. We do the
-    // first attempt (without holding the Heap_lock) here and the
-    // follow-on attempt will be at the start of the next loop
-    // iteration (after taking the Heap_lock).
-    size_t dummy = 0;
-    result = _allocator->attempt_allocation(word_size, word_size, &dummy);
-    if (result != nullptr) {
+    if (attempt_allocation_after_gc(word_size, gc_count_before, should_try_gc, &result, GCCause::_g1_inc_collection_pause)) {
```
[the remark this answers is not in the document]
walulyai: Because the false return will not be due to GCLocker.

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -989,98 +938,178 @@ HeapWord* G1CollectedHeap::satisfy_failed_allocation_helper(size_t word_size,
...
-// to support an allocation of the given "word_size".  If
-// successful, perform the allocation and return the address of the
-// allocated block, or else null.
+bool G1CollectedHeap::handle_allocation_requests(bool expect_null_mutator_alloc_region) {
+  assert_at_safepoint_on_vm_thread();
+
+  const uint active_numa_nodes = G1NUMA::numa()->num_active_nodes();
+  bool *expect_null_alloc_regions = (bool*)alloca(active_numa_nodes * sizeof(bool));
+  for (uint i = 0; i < active_numa_nodes; i++) {
+    expect_null_alloc_regions[i] = expect_null_mutator_alloc_region;
```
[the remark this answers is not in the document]
walulyai: I wasn't in position to determine that, so I went with keeping it. Removing it would make the code a lot cleaner and also avoid the use of alloca, but again, I need guidance on the removing it part

src/hotspot/share/gc/g1/g1CollectedHeap.hpp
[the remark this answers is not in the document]
albertnetymk: I think it's the `state` that determines how element should be processed, regardless which list it lives on. If so, merging them sounds simpler. Could you try and see which approach is shorter/clearer/easier?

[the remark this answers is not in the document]
walulyai: Done! could have more optimisations on early return from some of the list iterations. But pre-mature at this point.

src/hotspot/share/gc/g1/g1VMOperations.cpp
```diff
@@ -125,33 +132,37 @@ VM_G1CollectForAllocation::VM_G1CollectForAllocation(size_t         word_size,
...
-    _result = g1h->attempt_allocation_at_safepoint(_word_size,
-                                                   false /* expect_null_cur_alloc_region */);
-    if (_result != nullptr) {
-      // If we can successfully allocate before we actually do the
-      // pause then we will consider this pause successful.
-      _gc_succeeded = true;
+  // If any allocation has been requested, try to do that first.
+  bool has_allocation_requests = !g1h->_stalled_allocations.is_empty();
+  if (has_allocation_requests) {
+    bool alloc_succeeded = g1h->handle_allocation_requests(false /* expect_null_mutator_alloc_region*/);
```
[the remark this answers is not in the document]
walulyai: Reset to preferring the gc pause over heap expansion.

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -400,6 +402,48 @@ G1CollectedHeap::mem_allocate(size_t word_size,
...
+  }
+
+  {
+    MutexLocker ml(&_alloc_request_lock, Mutex::_no_safepoint_check_flag);
+    _stalled_allocations.remove(&request);
+  }
+
+  if (request.pending()) {
+    // GC Safepoint did not handle our allocation request. We should retry.
+    return false;
```
albertnetymk, line 434: Is it possible for a thread to hit this point indefinitely, i.e. some form of starvation?

walulyai: Only if you have indefinite triggering of non-allocating GCs. I wouldn't consider that starvation, as starvation here should imply some threads being able to allocate, and some other threads being unable to allocate even with allocation GCs triggered.

walulyai:
keep alive

DiegoPino:
keep alive+1

walulyai:
Keep alive!

PR #14081, opened 2023-05-22
8308545

dfuch:
The test `test/jdk/java/net/httpclient/ShutdownNow.java` has been observed failing intermittently (though rarely) with an assertion error: `java.lang.AssertionError: 0: Unexpected exception: java.io.IOException: Stream 1 cancelled` caused by  

```
java.io.IOException: Stream 1 cancelled
at java.net.http/jdk.internal.net.http.Stream.cancel(Stream.java:1285) 
```

Given the nature of the test - which interrupts / cancel all operations asynchronously, then a "stream %d cancelled" exception should count as an acceptable result.


jaikiran (approved):
The change looks OK to me.

PR #14082, opened 2023-05-22
JDK-8298066: java/util/concurrent/locks/Lock/OOMEInAQS.java timed out

viktorklang-ora:
Doubling the max heap size since not all GCs might have compressed OOPs, so this change should make this test more robust.

Cleared with @DougLea


AlanBateman:
Should the test be removed from ProblemList-generational-zgc.txt as part of this?

viktorklang-ora:
@AlanBateman My thinking was to integrate this and make sure that the problem doesn't resurface before removing it from the problem list. Does that make sense? 🤔

AlanBateman:
> @AlanBateman My thinking was to integrate this and make sure that the problem doesn't resurface before removing it from the problem list. Does that make sense? 🤔

No objection, it's just it would leave ProblemList-generational-zgc.txt listing a test against as issue marked as fixed.

viktorklang-ora:
@AlanBateman I wonder, perhaps it makes more sense, as you say, to clear the problem list together with this, and resurrect it in case it spuriously resurfaces. It's always a challenge to prove a negative. :)

viktorklang-ora:
@dholmes-ora 👍 done

viktorklang-ora:
(Addressed conflict, hence the force-push)

AlanBateman:
The bot isn't happy because JDK-8298066 is listed in both ProblemList-zgc.txt and ProblemList-generational-zgc.txt, it will need to be removed from both or else a separate issue if the increased heap size doesn't work for both GCs.

viktorklang-ora:
/integrate/integrate

viktorklang-ora:
@dholmes-ora It takes time to unlearn decades of "avoid merge commits" :) Thanks for the tip!

PR #14083, opened 2023-05-22
8300038: Make new version of JNU_GetStringPlatformChars which checks for null characters

Michael-Mc-Mahon:
This PR creates a new version of the JNI utility function JNU_GetStringPlatformChars called JNU_GetStringPlatformCharsStrict, which performs additional validation of the returned string, namely that it does not contain any embedded NULL characters. If any such characters are found the function returns NULL with an IAE pending. The change also switches usage in the networking native code to use the new function.

This cautious approach was taken rather than changing the behavior of the existing function as each native code area needs to review the effect of making the switch. Otherwise, surprising behavior changes might occur (eg undocumented IAE being thrown to user code instead of some other exception).

---------
### Progress

### Issue
 * [JDK-8300038](https://bugs.openjdk.org/browse/JDK-8300038): Make new version of JNU_GetStringPlatformChars which checks for null characters

### Reviewers
 * [Daniel Fuchs](https://openjdk.org/census#dfuchs) (@dfuch - **Reviewer**) ⚠️ Review applies to [19c36f89](https://git.openjdk.org/jdk/pull/14083/files/19c36f8951033422ee49aa815a1508327b023b40)
 * [Naoto Sato](https://openjdk.org/census#naoto) (@naotoj - **Reviewer**) ⚠️ Review applies to [19c36f89](https://git.openjdk.org/jdk/pull/14083/files/19c36f8951033422ee49aa815a1508327b023b40)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14083/head:pull/14083` \
`$ git checkout pull/14083`

`$ git checkout pull/14083` \
`$ git pull https://git.openjdk.org/jdk.git pull/14083/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14083`

`$ git pr show -t 14083`

</details>

<a href="https://git.openjdk.org/jdk/pull/14083.diff">https://git.openjdk.org/jdk/pull/14083.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14083#issuecomment-1559639489)


dfuch (approved):
Looks good to me, but it would be good to get this reviewed by someone with more knowledge of native character encoding.

AlanBateman:
I skimmed through the changes and mostly look okay. There's a few usage of "NULL" in exception messages that I assume should be "NUL character".  Probably need to bump the copyright date on several files.  You might want to find a better name for the test as NullChar isn't too descriptive, the test is really doing a lookup of a hostname that contains a NUL character. I assume there will be follow up issues to change other parts of the system to use the strict functions.

Michael-Mc-Mahon:
> I skimmed through the changes and mostly look okay. There's a few usage of "NULL" in exception messages that I assume should be "NUL character". Probably need to bump the copyright date on several files. You might want to find a better name for the test as NullChar isn't too descriptive, the test is really doing a lookup of a hostname that contains a NUL character. I assume there will be follow up issues to change other parts of the system to use the strict functions.

Thanks, I've updated per these comments.

Michael-Mc-Mahon:
I'll integrate this today unless there are further comments.

AlanBateman:
> I'll integrate this today unless there are further comments.

Thanks for addressing my comments.

PR #14084, opened 2023-05-22
8307958: Metaspace verification is slow causing extreme class unloading times

xmas92:
From JBS:
>We see very long class unloading times with Generational ZGC in fastdebug builds. MacOS spindumps show that it is caused by Metaspace verification code.
>
>I think the main problem is that many of the verification operations are performed at every operation (add, remove, split, merge) and the individual operation verifies much more than a constant number of chunks. This seems to give a quadratic complexity which blows up if you don't class unload often enough.

The approach here is to resolve this by putting the verification behind the SOMETIMES macro. It is then possible to turn it of completely with `-XX:VerifyMetaspaceInterval=0` while still benefiting from some stochastic verification. 

The goal here is to have the calls to Verify unchanged. Leaving all explicit gtest verify and Universe::verify calls unaffected.  
The verification was also left unchanged when committing the memory. 

All SOMETIMES uses the same frequency, `VerifyMetaspaceInterval`. I looked at the possibility of adding weight to some of the call site, but it did not seem to be worth the complexity. 

Testing: GHA and Oracle CI (tier1-3 ongoing), various long running stress tests.

---------
### Progress

### Issue
 * [JDK-8307958](https://bugs.openjdk.org/browse/JDK-8307958): Metaspace verification is slow causing extreme class unloading times

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14084/head:pull/14084` \
`$ git checkout pull/14084`

`$ git checkout pull/14084` \
`$ git pull https://git.openjdk.org/jdk.git pull/14084/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14084`

`$ git pr show -t 14084`

</details>

<a href="https://git.openjdk.org/jdk/pull/14084.diff">https://git.openjdk.org/jdk/pull/14084.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14084#issuecomment-1557325890)

### DIFF
--- a/src/hotspot/share/memory/metaspace/chunkManager.cpp
+++ b/src/hotspot/share/memory/metaspace/chunkManager.cpp
@@ -49,7 +49,7 @@ namespace metaspace {
 // Return a single chunk to the freelist and adjust accounting. No merge is attempted.
 void ChunkManager::return_chunk_simple_locked(Metachunk* c) {
   assert_lock_strong(Metaspace_lock);
-  DEBUG_ONLY(c->verify());
+  SOMETIMES(c->verify();)
   _chunks.add(c);
   c->reset_used_words();
   // Tracing
@@ -82,7 +82,7 @@ void ChunkManager::split_chunk_and_add_splinters(Metachunk* c, chunklevel_t targ
   assert(c->prev() == nullptr && c->next() == nullptr, "Chunk must be outside of any list.");
 
   DEBUG_ONLY(chunklevel::check_valid_level(target_level);)
-  DEBUG_ONLY(c->verify();)
+  SOMETIMES(c->verify();)
 
   UL2(debug, "splitting chunk " METACHUNK_FORMAT " to " CHKLVL_FORMAT ".",
       METACHUNK_FORMAT_ARGS(c), target_level);
@@ -101,8 +101,8 @@ void ChunkManager::split_chunk_and_add_splinters(Metachunk* c, chunklevel_t targ
   } else {
     assert(c->committed_words() == committed_words_before, "Sanity");
   }
-  c->verify();
-  verify_locked();
+  SOMETIMES(c->verify();)
+  SOMETIMES(verify_locked();)
   SOMETIMES(c->vsnode()->verify_locked();)
 #endif
   InternalStats::inc_num_chunk_splits();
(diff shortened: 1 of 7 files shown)


tstuefe (changes requested):
Seems reasonable.

Could you please crank up the verification interval for metaspace gtests from 3 to 1? (See test/hotspot/jtreg/gtest/MetaspaceGtests.java).

xmas92:
> Seems reasonable.
> 
> Could you please crank up the verification interval for metaspace gtests from 3 to 1? (See test/hotspot/jtreg/gtest/MetaspaceGtests.java).

Yes, that is much better than changing the test helper. 
Done.

coleenp (approved):
This seems fine.  To be clear, was most of the time spent in the ChunkManager::get_chunk() and return_chunk(), and maybe purge() calls, and not much in the split and merge chunks?  I agree with adding SOMETIMES to all for consistency, but I hope we're not doing more of these operations that I assume should be infrequent.

xmas92:
> This seems fine. To be clear, was most of the time spent in the ChunkManager::get_chunk() and return_chunk(), and maybe purge() calls, and not much in the split and merge chunks? I agree with adding SOMETIMES to all for consistency, but I hope we're not doing more of these operations that I assume should be infrequent.

Maybe it is overly defensive to add it to the split and get_chunk paths. The observed slowdown was for `purge()` calls, and more specifically the verification downstream of `~MetaspaceArena()`, which includes `return_chunk` and `merge()` as well as freelist `contains()` verification.

xmas92:
> A regression test would be good, since Metaspace will evolve and so will verifications, and I would like to know what performance costs are acceptable.

Oracles internal testing will most likely provoke regressions in this area. Similarly how this issue was brought to light. To see the really degenerate behaviour, required constant loading and unloading of many class loaders with very few classes over hours with infrequent class unloading (due to generational ZGC not performing many major collections, low old heap pressure). 

While it is probably possible to reduce the variables and create a simpler test which solely stresses metaspace debug timings and little else, it seems like a rather complex task.

Any product performance regression is caught by our normal performance testing and any relevant and degenerate debug regression would be caught by Oracle internal tests. Hopefully this is enough for now?

xmas92:
> > Maybe it is overly defensive to add it to the split and get_chunk paths. The observed slowdown was for purge() calls, and more specifically the verification downstream of ~MetaspaceArena(), which includes return_chunk and merge() as well as freelist contains() verification.
> 
> Ok, thanks for the info. That makes sense.
> 
> I thought there were more GC tests for class unloading but I can't find them. If you create one test with loading/unloading a class 100 times with the SOMETIMES verification on, I think that would run fast enough and exercise the code paths that won't be exercised regularly. You can put it in runtime/ClassUnload with some other tests. Or if you can find tests already in metaspace that do this, that would be sufficient.

`test/jdk/jdk/jfr/event/runtime/TestMetaspaceAllocationFailure.java` and `test/hotspot/jtreg/runtime/Metaspace/FragmentMetaspace.java` seems like relevant candidates for running with `-XX: VerifyMetaspaceInterval=1`

I'll look at creating a more fuzzy stress test which uses `jdk.test.lib.classloader.*` to load and unload class loaders.

tstuefe (approved):
Good.

xmas92:
Thanks for the review. I will integrate this now to improve the CI testing for tests that may stumble on this.
Created [JDK-8308848](https://bugs.openjdk.org/browse/JDK-8308848) for creating more tests as discussed.
/integrate

PR #14085, opened 2023-05-22
8296411: AArch64: Accelerated Poly1305 intrinsics

theRealAph:
This provides a solid speedup of about 3-4x over the Java implementation.

I have a vectorized version of this which uses a bunch of tricks to speed it up, but it's complex and can still be improved. We're getting close to ramp down, so I'm submitting this simple intrinsic so that we can get it reviewed in time.

Benchmarks:

```
ThunderX (2, I think):

Benchmark                        (dataSize)  (provider)   Mode  Cnt         Score         Error  Units
Poly1305DigestBench.updateBytes          64              thrpt    3  14078352.014 ± 4201407.966  ops/s
Poly1305DigestBench.updateBytes         256              thrpt    3   5154958.794 ± 1717146.980  ops/s
Poly1305DigestBench.updateBytes        1024              thrpt    3   1416563.273 ± 1311809.454  ops/s
Poly1305DigestBench.updateBytes       16384              thrpt    3     94059.570 ±    2913.021  ops/s
Poly1305DigestBench.updateBytes     1048576              thrpt    3      1441.024 ±     164.443  ops/s

Benchmark                        (dataSize)  (provider)   Mode  Cnt        Score        Error  Units
Poly1305DigestBench.updateBytes          64              thrpt    3  4516486.795 ± 419624.224  ops/s
Poly1305DigestBench.updateBytes         256              thrpt    3  1228542.774 ± 202815.694  ops/s
Poly1305DigestBench.updateBytes        1024              thrpt    3   316051.912 ±  23066.449  ops/s
Poly1305DigestBench.updateBytes       16384              thrpt    3    20649.561 ±   1094.687  ops/s
Poly1305DigestBench.updateBytes     1048576              thrpt    3      310.564 ±     31.053  ops/s

Apple M1:

Benchmark                        (dataSize)  (provider)   Mode  Cnt         Score        Error  Units
Poly1305DigestBench.updateBytes          64              thrpt    3  33551968.946 ± 849843.905  ops/s
Poly1305DigestBench.updateBytes         256              thrpt    3   9911637.214 ±  63417.224  ops/s
Poly1305DigestBench.updateBytes        1024              thrpt    3   2604370.740 ±  29208.265  ops/s
Poly1305DigestBench.updateBytes       16384              thrpt    3    165183.633 ±   1975.998  ops/s
Poly1305DigestBench.updateBytes     1048576              thrpt    3      2587.132 ±     40.240  ops/s

Benchmark                        (dataSize)  (provider)   Mode  Cnt         Score        Error  Units
Poly1305DigestBench.updateBytes          64              thrpt    3  12373649.589 ± 184757.721  ops/s
Poly1305DigestBench.updateBytes         256              thrpt    3   3112536.605 ±  14436.410  ops/s
Poly1305DigestBench.updateBytes        1024              thrpt    3    777184.018 ±   8774.478  ops/s
Poly1305DigestBench.updateBytes       16384              thrpt    3     50224.072 ±     29.004  ops/s
Poly1305DigestBench.updateBytes     1048576              thrpt    3       776.229 ±      8.086  ops/s

```

---------
### Progress

### Issue
 * [JDK-8296411](https://bugs.openjdk.org/browse/JDK-8296411): AArch64: Accelerated Poly1305 intrinsics

### Reviewers
 * [Claes Redestad](https://openjdk.org/census#redestad) (@cl4es - **Reviewer**) ⚠️ Review applies to [93a03c62](https://git.openjdk.org/jdk/pull/14085/files/93a03c6252f6c09be05f3991725defeb8b89e3ed)
 * [Andrew Dinn](https://openjdk.org/census#adinn) (@adinn - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14085/head:pull/14085` \
`$ git checkout pull/14085`

`$ git checkout pull/14085` \
`$ git pull https://git.openjdk.org/jdk.git pull/14085/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14085`

`$ git pr show -t 14085`

</details>

<a href="https://git.openjdk.org/jdk/pull/14085.diff">https://git.openjdk.org/jdk/pull/14085.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14085#issuecomment-1557526681)

### DIFF
--- a/src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp
@@ -7128,18 +7128,19 @@ class StubGenerator: public StubCodeGenerator {
       // Partial reduction mod 2**130 - 5
       __ adds(U_1, U_0HI, U_1);
       __ adc(U_2, U_1HI, U_2);
-      // Sum now in U_2:U_1, U_0.
+      // Sum now in U_2:U_1:U_0.
       // Dead: U_0HI, U_1HI.
       regs = (regs.remaining() + U_0HI + U_1HI).begin();
 
-      // U_2:U_1:U_0 += (U_1HI >> 2)
+      // U_2:U_1:U_0 += (U_2 >> 2) * 5 in two steps
+
+      // First, U_2:U_1:U_0 += (U_2 >> 2)
       __ lsr(rscratch1, U_2, 2);
       __ andr(U_2, U_2, (u8)3);
       __ adds(U_0, U_0, rscratch1);
       __ adcs(U_1, U_1, zr);
       __ adc(U_2, U_2, zr);
-
-      // U_1HI:U_0HI, U_0 += (U_1HI >> 2) << 2
+      // Second, U_2:U_1:U_0 += (U_2 >> 2) << 2
       __ adds(U_0, U_0, rscratch1, Assembler::LSL, 2);
       __ adcs(U_1, U_1, zr);
       __ adc(U_2, U_2, zr);


src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
```diff
@@ -569,6 +569,10 @@ void VM_Version::initialize() {
   if (FLAG_IS_DEFAULT(AlignVector)) {
     AlignVector = AvoidUnalignedAccesses;
   }
+
+    if (FLAG_IS_DEFAULT(UsePoly1305Intrinsics)) {
```
fg1417, line 573: Incorrect indention: extra space.

src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp
```diff
@@ -7013,6 +7013,147 @@ class StubGenerator: public StubCodeGenerator {
...
+      __ add(S_2, S_2, 1);
+
+      const Register U_0HI = *++regs, U_1HI = *++regs;
+
+      // NB: this logic depends on some of the special properties of
+      // Poly1305 keys. In particular, because we know that the top
+      // four bits of each 32-bit subword of "r" are zero, we can add
+      // together partial products without any risk of needing to
+      // propagate a carry out.
+      wide_mul(U_0, U_0HI, S_0, R_0);  wide_madd(U_0, U_0HI, S_1, RR_1); wide_madd(U_0, U_0HI, S_2, RR_0);
```
cl4es, line 7097: This asserts that 'the top four bits of each 32-bit subword of "r" are zero' but we're packing 26-bit values into `R_0...R_2` above in a way that seem to break this invariant?

theRealAph: No, it doesn't break the invariants.

R is the randomly-chosen 128-bit key. It is generated from an initial 128-bit-log string or random bits, then
`r &= 0x0ffffffc0ffffffc0ffffffc0fffffff`

This 128-bit-long string is split into 26-bit limbs before the intrinsic is called. The zero bits remain zero.
When we repack R into two 64-bit registers those zero bits are still zero.

theRealAph: See https://loup-vaillant.fr/tutorials/poly1305-design for more explanation

cl4es: Thanks for the link! 

So `r` refers to the value passed via `r_start` and it wasn't clear from the immediate context that `r_start` is already split into 26-bit limbs. So the `pack26` takes the 5 26-bit limbs and repacks them so that `R_0` has the low 64-bit of `r`, `R_1` the high bits. Makes sense. 

`R_2` is unused and could be reclaimed. Perhaps an override for `pack26` that only takes two registers and discards the last 2 bits? Might help clarify the setup.

theRealAph: Done.

cl4es (approved):
Thanks for your patience in answering my questions and addressing my comments.

theRealAph:
> Thanks for your patience in answering my questions and addressing my comments.

Thank you for asking questions that made the patch better, and even removed an instruction in what I thought was a tightly-written intrinsic!

src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp
```diff
@@ -7013,6 +7013,170 @@ class StubGenerator: public StubCodeGenerator {
...
+      regs = (regs.remaining() + S_0 + S_1 + S_2).begin();
+
+      // Partial reduction mod 2**130 - 5
+      __ adds(U_1, U_0HI, U_1);
+      __ adc(U_2, U_1HI, U_2);
+      // Sum now in U_2:U_1, U_0.
+      // Dead: U_0HI, U_1HI.
+      regs = (regs.remaining() + U_0HI + U_1HI).begin();
+
+      // U_2:U_1:U_0 += (U_1HI >> 2)
```
adinn, line 7135: This comment and the next one both need correcting. They mention U_0HI and U_1HI and, as the previous comment says, those registers are dead.

What actually happens here is best summarized as

// U_2:U_1:U_0 += (U2 >> 2) * 5

or, if we actually want to be clearer about the current encoding which does it in several steps

// rscratch1 = (U2 >> 2)
// U2 = U2[1:0]
// U_2:U_1:U_0 += rscratch1
// U_2:U_1:U_0 += (rscratch1 << 2)

i.e. any bits that are set from 130 upwards are masked off, treated as an integer in their own right, multiplied by 5 and the result added back in at the bottom to update the 130 bit result U2[1:0]:U1[63:0]:U0[63:0].

I'm not sure whether this provides an opportunity for you to optimize this by doing the multiply by five earlier i.e. replace the code with this version

    // rscratch1 = (U2 >> 2) * 5
    // U2 = U2[1:0]
    __ lsr(rscratch1, U_2, 2);
    __ andr(U_2, U_2, (u8)3);
    __ add(rscratch1, rscratch1, scratch1, Assembler::LSL, 2);
    // U2:U1:U0 += rscratch1
    __ adds(U_0, U_0, rscratch1, Assembler::LSL, 2);
    __ adcs(U_1, U_1, zr);
    __ adc(U_2, U_2, zr);

The obvious concern is that the multiply of rscratch1 by 5 might overflow 64 bits. Is that why you have implemented two add and carry steps? If so then why is it legitimate to do the multiply by 5 up front in the final reduction that follows the loop?

theRealAph: > This comment and the next one both need correcting. They mention U_0HI and U_1HI and, as the previous comment says, those registers are dead.
>
> What actually happens here is best summarized as
>
> // U_2:U_1:U_0 += (U2 >> 2) * 5
>
> or, if we actually want to be clearer about the current encoding which does it in several steps
>
> // rscratch1 = (U2 >> 2) // U2 = U2[1:0] // U_2:U_1:U_0 += rscratch1 // U_2:U_1:U_0 += (rscratch1 << 2)
>
> i.e. any bits that are set from 130 upwards are masked off, treated as an integer in their own right, multiplied by 5 and the result added back in at the bottom to update the 130 bit result U2[1:0]:U1[63:0]:U0[63:0].

OK.

> I'm not sure whether this provides an opportunity for you to optimize this by doing the multiply by five earlier i.e. replace the code with this version

I'm not sure either, which is why it's done in two separate steps. I think you may be right, but it's a bit late to be optimizing this version any further. That would require careful analysis and a redo of all the testing.

> The obvious concern is that the multiply of rscratch1 by 5 might overflow 64 bits. Is that why you have implemented two add and carry steps?

Indeed.

> If so then why is it legitimate to do the multiply by 5 up front in the final reduction that follows the loop?

I assume that you're referring to the multiply by 5 in

```
// Further reduce modulo 2^130 - 5                                                                                                                                                                                                                                                                                                                           
    __ lsr(rscratch1, U_2, 2);
    __ add(rscratch1, rscratch1, rscratch1, Assembler::LSL, 2); // rscratch1 = U_2 * 5                                                                                                                                                                                                                                                                       
```

`

adinn: Yes, of course, you are right that  0<= U_2 < 6 at the point where that second multiply by 5 occurs (i.e. after the loop).

I believe it is safe to use the same optimization inside the loop for reasons given below. Of course it is a bit late to change this now and retest but if my reasoning is correct then we could consider updating this post release and, maybe, a backport.

The only thing that needs to be determined is what value could sit in U2 when we enter the loop. That's the only important case because we already agreed that at the loop back edge that 0 <= U2 < 6.

The incoming value for U2 at loop entry is derived by the following subsequence of the instruction stream

      __ adcs(S_1, U_1, S_1);
      __ adc(S_2, U_2, zr);         // A.1
      __ add(S_2, S_2, 1);          // A.2
     . . .
       wide_mul(U_1, U_1HI, S_0, R_1);  wide_madd(U_1, U_1HI, S_1, R_0);  wide_madd(U_1, U_1HI, S_2, RR_1); // B
     . . .
      __ andr(U_2, R_0, 3);         // C
      __ mul(U_2, S_2, U_2);        // D
      . . .

      __ adc(U_2, U_1HI, U_2);      // E

At A.1 we know that 0 <= U_2 <= 3 (since it was initialized by unpack26)
So, at A.2 we know that 0 <= S2 <= 5

At B we know that 0 <= RR_1 <= (2^60 - 2^2)  = FFFFFFF_FFFFFFFC (top 4 and bottom 2 bits of RR_1 are clear)
So 0 <= U1_HI < 5 * FFFFFFF_FFFFFFFC = 4FFFFFFF_FFFFFFEC

At C we know 0 <= U_2 <= 3

At D we know 0 <= U_2 <= 15

So at E we know that 0 <= U_2 <= 4FFFFFFF_FFFFFFEC + 15 + 1

So, the highest possible value for U_2 at loop entry is 50000000_00000002.

Clearly we can shift this down by two and add without any danger of overflowing

 50000000_00000002 >> 2 + 50000000_00000002 = 64000000_00000002

adinn (approved):
This is ok to push now the comments have been corrected.

src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp
adinn: Ah, no scratch that. I have made a wrong assumption at B. The value of U1_HI is bounded by the sum of the 3 multiplies.

adinn: I think there is still a proof of validity to be salvaged though. We compute a 128 bit product:

  U1_HI:U1 = S_0 * R_1 + S_1 * R_0 + S_2 * RR_1

We know that R_0 and R_1 have four top bits clear and S2 <= 5. So, I think we can guarantee that the top word of the 128 bit product is small enough to not to overflow when we do the shift. Even if we assume S_0, S_1 and RR_1 have the maximum possible value we have

S_0 * R_1 <= (2^64 - 1) * (2^60 - 1)
S_1 * R_0 <= (2^64 - 1) * (2^60 - 1)
S_2 * RR_1  <= 5 * (2^64 - 1)

So, the top word U1_HI is bounded by at most (2 * (2^60 - 1)) + 5. That leaves more than enough room to guarantee that the shift and add will not overflow.

PR #14092, opened 2023-05-23
8308655: Narrow types of ConstantPool and ConstMethod returns

coleenp:
This change uses a number of ways to eliminate -Wconversion warnings in the metadata files in the oops directory. 

1. narrow return types to u2 if the accessor is for a field or value that's u2 (u2 is most common for constMethod fields and constant pool indices)
2. Use checked_cast<type> for places where we know the int value is u2 or s2 but propagating these types is too much fan out.
3. Use plain casts where it's obvious that the int value fits in the casted-to type.
4. Moved KlassKind to be contained in Klass to add the Unknown enum value to use instead of -1.
5. Moved the compute_from_signature function into ConstMethod as it sets values in ConstMethod and the parameters are changed in the set functions.  Removed some pass through functions in Method.

Tested with tier1-4.

---------
### Progress

### Issue
 * [JDK-8308655](https://bugs.openjdk.org/browse/JDK-8308655): Narrow types of ConstantPool and ConstMethod returns

### Reviewers
 * [Frederic Parain](https://openjdk.org/census#fparain) (@fparain - **Reviewer**) ⚠️ Review applies to [7398e225](https://git.openjdk.org/jdk/pull/14092/files/7398e2252a348ab8114916f90aa05fa287b0b81b)
 * [Matias Saavedra Silva](https://openjdk.org/census#matsaave) (@matias9927 - Committer) ⚠️ Review applies to [7398e225](https://git.openjdk.org/jdk/pull/14092/files/7398e2252a348ab8114916f90aa05fa287b0b81b)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14092/head:pull/14092` \
`$ git checkout pull/14092`

`$ git checkout pull/14092` \
`$ git pull https://git.openjdk.org/jdk.git pull/14092/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14092`

`$ git pr show -t 14092`

</details>

<a href="https://git.openjdk.org/jdk/pull/14092.diff">https://git.openjdk.org/jdk/pull/14092.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14092#issuecomment-1559639022)

### DIFF
--- a/src/hotspot/share/classfile/classLoaderExt.cpp
+++ b/src/hotspot/share/classfile/classLoaderExt.cpp
@@ -66,7 +66,8 @@ void ClassLoaderExt::append_boot_classpath(ClassPathEntry* new_entry) {
 
 void ClassLoaderExt::setup_app_search_path(JavaThread* current) {
   Arguments::assert_is_dumping_archive();
-  _app_class_paths_start_index = ClassLoader::num_boot_classpath_entries();
+  int start_index = ClassLoader::num_boot_classpath_entries();
+  _app_class_paths_start_index = checked_cast<jshort>(start_index);
   char* app_class_path = os::strdup_check_oom(Arguments::get_appclasspath(), mtClass);
 
   if (strcmp(app_class_path, ".") == 0) {
@@ -116,8 +117,9 @@ void ClassLoaderExt::process_module_table(JavaThread* current, ModuleEntryTable*
 
 void ClassLoaderExt::setup_module_paths(JavaThread* current) {
   Arguments::assert_is_dumping_archive();
-  _app_module_paths_start_index = ClassLoader::num_boot_classpath_entries() +
-                              ClassLoader::num_app_classpath_entries();
+  int start_index = ClassLoader::num_boot_classpath_entries() +
+                    ClassLoader::num_app_classpath_entries();
+  _app_module_paths_start_index = checked_cast<jshort>(start_index);
   Handle system_class_loader (current, SystemDictionary::java_system_loader());
   ModuleEntryTable* met = Modules::get_module_entry_table(system_class_loader);
   process_module_table(current, met);
--- a/src/hotspot/share/oops/constMethod.hpp
+++ b/src/hotspot/share/oops/constMethod.hpp
@@ -450,18 +450,18 @@ public:
   void compute_from_signature(Symbol* sig, bool is_static);
 
   // max stack
-  int  max_stack() const                         { return _max_stack; }
+  u2   max_stack() const                         { return _max_stack; }
   void set_max_stack(int size)                   { _max_stack = checked_cast<u2>(size); }
 
   // max locals
-  int  max_locals() const                        { return _max_locals; }
+  u2  max_locals() const                         { return _max_locals; }
   void set_max_locals(int size)                  { _max_locals = checked_cast<u2>(size); }
 
   // size of parameters
   u2 size_of_parameters() const                  { return _size_of_parameters; }
 
   // Number of arguments passed on the stack even when compiled
-  int  num_stack_arg_slots() const               { return _num_stack_arg_slots; }
+  u2 num_stack_arg_slots() const                 { return _num_stack_arg_slots; }
 
   // result type (basic type of return value)
   BasicType result_type() const                  { assert(_result_type >= T_BOOLEAN, "Must be set");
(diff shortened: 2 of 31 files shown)


fparain (commented):
Overall, the changes look good, only two points that could be improved (see comments in the code)

src/hotspot/share/oops/klass.cpp
```diff
@@ -261,7 +261,7 @@ void Klass::initialize_supers(Klass* k, Array<InstanceKlass*>* transitive_interf
       // Overflow of the primary_supers array forces me to be secondary.
       super_check_cell = &_secondary_super_cache;
     }
-    set_super_check_offset((address)super_check_cell - (address) this);
+    set_super_check_offset(u4((address)super_check_cell - (address) this));
```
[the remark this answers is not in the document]
coleenp: This casts the expression in the argument to u4 (same as juint), since the address calculation returns 'address' which is 64 bits.

src/hotspot/share/oops/constMethod.hpp
```diff
@@ -441,28 +446,27 @@ class ConstMethod : public MetaspaceObj {
   u2 orig_method_idnum() const                   { return _orig_method_idnum; }
   void set_orig_method_idnum(u2 idnum)           { _orig_method_idnum = idnum; }
 
+  // Derive stuff from the signature at load time.
+  void compute_from_signature(Symbol* sig, bool is_static);
+
   // max stack
   int  max_stack() const                         { return _max_stack; }
```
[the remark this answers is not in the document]
coleenp: Ok, yes they can since that would be consistant.

fparain (approved):

src/hotspot/share/oops/klass.hpp
```diff
@@ -79,7 +66,22 @@ class vtableEntry;
...
+  // Klass Kinds for all subclasses of Klass
+  enum KlassKind {
+    InstanceKlassKind,
+    InstanceRefKlassKind,
+    InstanceMirrorKlassKind,
+    InstanceClassLoaderKlassKind,
+    InstanceStackChunkKlassKind,
+    TypeArrayKlassKind,
+    ObjArrayKlassKind,
+    Unknown
```
matias9927, line 79: Maybe this should be `UnknownKind` to be consistent with the other options.

src/hotspot/share/classfile/classLoaderExt.cpp
```diff
@@ -66,7 +66,8 @@ void ClassLoaderExt::append_boot_classpath(ClassPathEntry* new_entry) {
 
 void ClassLoaderExt::setup_app_search_path(JavaThread* current) {
   Arguments::assert_is_dumping_archive();
-  _app_class_paths_start_index = ClassLoader::num_boot_classpath_entries();
+  int start_index = ClassLoader::num_boot_classpath_entries();
+  _app_class_paths_start_index = checked_cast<jshort>(start_index);
```
matias9927, line 70: This might be a misunderstanding but are these meant to be s2 instead of jshort?

src/hotspot/share/classfile/classLoaderExt.cpp
```diff
@@ -116,8 +117,9 @@ void ClassLoaderExt::process_module_table(JavaThread* current, ModuleEntryTable*
 
 void ClassLoaderExt::setup_module_paths(JavaThread* current) {
   Arguments::assert_is_dumping_archive();
-  _app_module_paths_start_index = ClassLoader::num_boot_classpath_entries() +
-                              ClassLoader::num_app_classpath_entries();
+  int start_index = ClassLoader::num_boot_classpath_entries() +
+                    ClassLoader::num_app_classpath_entries();
+  _app_module_paths_start_index = checked_cast<jshort>(start_index);
```
matias9927, line 122: Same question as above

matias9927 (approved):
Looks good, I just had a few minor comments that might be unimportant. Otherwise, approved!

src/hotspot/share/classfile/classLoaderExt.cpp
coleenp: The declaration is jshort so I left it jshort to not modify the declaration, even though s2 is the same and better.

src/hotspot/share/classfile/classLoaderExt.cpp
coleenp: Same here.  I didn't want it to not match the declaration.

coleenp (commented):
Matias, thank you for the review and comments.

coleenp:
Thanks Fred and Matias for reviewing
/integrate

PR #14093, opened 2023-05-23
8308603: Removing do_pending_ref/enclosing_ref from MetaspaceClosure

iklam:
**The problem**

MetaspaceClosure is used to traverse all *pointers* in the metaspace objects. It works recursively. A pointer is represented by the `MetaspaceClosure::Ref` type. For example, if you have

```
InstanceKlass* a = ..., *b = ...;
a->_super = b;
```

When we are traversing the pointer `a->_super`, we create a `Ref` that records the address of `b` and `&a->super`. However, we currently don't remember the address of `a` in the `Ref`. As a result, when processing the `Ref` inside a `push(Ref* ref)` function, we need to call `MetaspaceClosure::enclousing_ref()` to get to `a`, so that we can mark pointers inside it.

The reason for`do_pending_ref` is even more convoluted. Please see [JDK-8308603](https://bugs.openjdk.org/browse/JDK-8308603) for more details.

**The fix**

The code can be much more readable if we simply remember  `a` in the `Ref`.

**More info**

A more in-depth discussion of the use of the MetaspaceClosure API in CDS can be found at:
- https://github.com/iklam/tools/blob/main/NOTES/cds_status_2023.md#33-copying-of-class-metadata

---------
### Progress

### Issue
 * [JDK-8308603](https://bugs.openjdk.org/browse/JDK-8308603): Removing do_pending_ref/enclosing_ref from MetaspaceClosure (**Enhancement** - P4)

### Reviewers
 * [Calvin Cheung](https://openjdk.org/census#ccheung) (@calvinccheung - **Reviewer**) ⚠️ Review applies to [fecc2a42](https://git.openjdk.org/jdk/pull/14093/files/fecc2a42703af0315b230d6b02408da2279cd3e8)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14093/head:pull/14093` \
`$ git checkout pull/14093`

`$ git checkout pull/14093` \
`$ git pull https://git.openjdk.org/jdk.git pull/14093/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14093`

`$ git pr show -t 14093`

</details>

<a href="https://git.openjdk.org/jdk/pull/14093.diff">https://git.openjdk.org/jdk/pull/14093.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14093#issuecomment-1558377824)


src/hotspot/share/cds/archiveBuilder.cpp
```diff
@@ -440,33 +428,53 @@ bool ArchiveBuilder::gather_one_source_obj(MetaspaceClosure::Ref* enclosing_ref,
...
-    if (src_info == nullptr) {
-      // source objects of point_to_it/set_to_null types are not copied
-      // so we don't need to remember their pointers.
+  address enclosing_obj = ref->enclosing_obj();
+  if (enclosing_obj == nullptr) {
+    return;
+  }
+
+  // We are dealing with 3 addresses:
+  // address o    = ref->obj(): We have found an object whose address is o.
```
coleenp, line 452: Are 'obj' oops ?  We almost always refer to oops as obj but not other things.  Is 'object' an oop?  This operates on metadata right? I realize this terminology is pre-existing.  I suggest changing obj and object in comments at least in the new code to either metadata() or ptr() even, metaptr() ?   Unless this does move oops. I don't know this code.

iklam: All the "objects" in the code related to MetaspaceClosure refer to instances of MetaspaceObj. This code rarely interacts with code that work on oops, so there isn't much confusion in practice.

I'd rather not change part of the code or comments to use a different name than "object". I think that will be even more confusing, causing the reader to think that "metadata", for example, is materially different than "object".

Also, "metadata" is unfortunately not a good name, because not every MetaspaceObj is a Metadata.

If renaming is desirable, that should be done in a separate RFE.

src/hotspot/share/cds/archiveBuilder.cpp
```diff
@@ -440,33 +428,53 @@ bool ArchiveBuilder::gather_one_source_obj(MetaspaceClosure::Ref* enclosing_ref,
...
+  //     Method* o                     = 0x0000abcd;
+  //     Array<Method*>* enclosing_obj = 0x00001000;
+  //     enclosing_obj->at_put(0, o);
+  //
+  // We the MetaspaceClosure iterates on the very first element of this array, we have
+  //     ref->obj()           == 0x0000abcd   (the Method)
+  //     ref->mpp()           == 0x00001008   (the location of the first element in the array)
+  //     ref->enclosing_obj() == 0x00001000   (the Array that contains the Method)
+  //
+  // We use the above informatio to mark the bitmap to indicate that there's a pointer on address 0x00001008.
```
calvinccheung, line 468: typo: informatio -> information

calvinccheung (approved):

iklam:
Thanks @calvinccheung for the review.
/integrate

PR #7310, opened 2022-02-01
8015854: [macosx] JButton's HTML ImageView adding unwanted padding

DamonGuy:
Html does not fit in JButton at certain sizes because default Insets cause html to be displayed off-center. 

Changes made to SwingUtilities.java layoutCompoundLabelImpl method to enable clipping if html does not fit, similar to regular text. AquaButtonUI.java now detects when html does not fit, and an implementation for alternate insets that are recursively tested for regular text inside layoutAndGetText() are now also being used for html content.

Created test (Bug8015854.java) with the same format as the test described on the issue. The button is of size 37x37 with an image of a red square sized 19x19. The test checks for red pixels on the edges of where the square image should be from the center of the button. The test fails with the non-changed jdk, but passes with the changes.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed

### Issue
 * [JDK-8015854](https://bugs.openjdk.java.net/browse/JDK-8015854): [macosx] JButton's HTML ImageView adding unwanted padding

### Reviewers
 * [Prasanta Sadhukhan](https://openjdk.java.net/census#psadhukhan) (@prsadhuk - **Reviewer**)
 * [Alexey Ivanov](https://openjdk.java.net/census#aivanov) (@aivanov-jdk - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk pull/7310/head:pull/7310` \
`$ git checkout pull/7310`

Update a local copy of the PR: \
`$ git checkout pull/7310` \
`$ git pull https://git.openjdk.java.net/jdk pull/7310/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 7310`

View PR using the GUI difftool: \
`$ git pr show -t 7310`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.java.net/jdk/pull/7310.diff">https://git.openjdk.java.net/jdk/pull/7310.diff</a>

</details>

PR #14094, opened 2023-05-23
8294535 : Add screen capture functionality to PassFailJFrame

lawrence-andrew:
1) Used builder pattern 
2) Tested AWT tests and it passed

---------
### Progress

### Issue
 * [JDK-8294535](https://bugs.openjdk.org/browse/JDK-8294535): Add screen capture functionality to PassFailJFrame (**Enhancement** - P4)

### Reviewers

### Contributors

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14094/head:pull/14094` \
`$ git checkout pull/14094`

`$ git checkout pull/14094` \
`$ git pull https://git.openjdk.org/jdk.git pull/14094/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14094`

`$ git pr show -t 14094`

</details>

<a href="https://git.openjdk.org/jdk/pull/14094.diff">https://git.openjdk.org/jdk/pull/14094.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14094#issuecomment-1560594482)


DamonGuy:
I looked at the JBS issue and this PR's description but when I look at the changes, there are changes to tests as well, like the awt TrayIcon one. Is that meant to be included? Or is that a separate change that accidentally got bundled with this change to PassFailJFrame?

aivanov-jdk:
> I looked at the JBS issue and this PR's description but when I look at the changes, there are changes to tests as well, like the awt TrayIcon one. Is that meant to be included? Or is that a separate change that accidentally got bundled with this change to PassFailJFrame?

I believe this is intended to demonstrate the usage of the screenshot feature.

Yet It may be a hindrance for backporting… if the modified tests don't exist in previous versions or there are other differences…

test/jdk/java/awt/regtesthelpers/PassFailJFrame.java
```diff
@@ -184,6 +262,53 @@ public void windowClosing(WindowEvent e) {
         windowList.add(frame);
     }
 
+    private static void captureScreen(Rectangle bounds) {
+        if (robot == null) {
+            try {
+                robot = new Robot();
+            } catch (AWTException e) {
+                String errorMsg = "Failed to create the " +
+                        "instance of Robot.";
```
aivanov-jdk, line 271:
```suggestion
                String errorMsg = "Failed to create an " +
                        "instance of Robot.";
```
Maybe make it a constant?

test/jdk/java/awt/regtesthelpers/PassFailJFrame.java
```diff
@@ -184,6 +262,53 @@ public void windowClosing(WindowEvent e) {
...
+                JOptionPane.showMessageDialog(frame, errorMsg, "Failed",
+                        JOptionPane.ERROR_MESSAGE);
+                forceFail(errorMsg + e.getMessage());
+            }
+        }
+        MultiResolutionImage multiResolutionImage = robot
+                .createMultiResolutionScreenCapture(bounds);
+        List<Image> imageList =
+                multiResolutionImage.getResolutionVariants();
+        Image image = imageList.get(imageList.size() - 1);
```
aivanov-jdk, line 281:
```suggestion
        MultiResolutionImage mri = robot.createMultiResolutionScreenCapture(bounds);
        List<Image> imageList = mri.getResolutionVariants();
        Image image = imageList.get(imageList.size() - 1);
```
`mri` is a common abbreviation of `MultiResolutionImage`, it's much shorter.

Alternatively, you may inline the two calls:
```suggestion
        List<Image> imageList = robot.createMultiResolutionScreenCapture(bounds)
                                     .getResolutionVariants();
        Image image = imageList.get(imageList.size() - 1);
```

test/jdk/java/awt/regtesthelpers/PassFailJFrame.java
```diff
@@ -184,6 +262,53 @@ public void windowClosing(WindowEvent e) {
...
+        File file = new java.io.File("CaptureScreen_" +
+                imgCounter.incrementAndGet() + ".png");
+        try {
+            ImageIO.write((RenderedImage) image, "png", file);
+        } catch (IOException e) {
+            throw new RuntimeException(e.getMessage());
```
aivanov-jdk, line 287: You shouldn't throw away the original exception, it may be useful in diagnosing the problem.
```suggestion
            throw new RuntimeException(e);
```

Show message to the user? Or, add a try-catch inside `captureScreen` to show this message.

test/jdk/java/awt/regtesthelpers/PassFailJFrame.java
```diff
@@ -184,6 +262,53 @@ public void windowClosing(WindowEvent e) {
...
+        }
+    }
+
+    private static void captureScreen(int selectedIndex) {
+        if (selectedIndex == 0) {
+            Arrays.stream(GraphicsEnvironment.getLocalGraphicsEnvironment()
+                    .getScreenDevices())
+                    .map(GraphicsDevice::getDefaultConfiguration)
+                    .map(GraphicsConfiguration::getBounds)
+                    .forEach(PassFailJFrame::captureScreen);
```
aivanov-jdk, line 297:
```suggestion
            Arrays.stream(GraphicsEnvironment.getLocalGraphicsEnvironment()
                                             .getScreenDevices())
                  .map(GraphicsDevice::getDefaultConfiguration)
                  .map(GraphicsConfiguration::getBounds)
                  .forEach(PassFailJFrame::captureScreen);
```
Align the dots of the chained calls, which is especially important for `getScreenDevices` where it's called on the result of `getLocalGraphicsEnvironment()` so that the stream is created on `getScreenDevices`.

test/jdk/java/awt/regtesthelpers/PassFailJFrame.java
```diff
@@ -184,6 +262,53 @@ public void windowClosing(WindowEvent e) {
...
+        }
+
+        if (selectedIndex == 1) {
+            windowList.stream()
+                    .filter(Window::isShowing)
+                    .map(Window::getBounds)
+                    .forEach(PassFailJFrame::captureScreen);
```
aivanov-jdk, line 304:
```suggestion
            windowList.stream()
                      .filter(Window::isShowing)
                      .map(Window::getBounds)
                      .forEach(PassFailJFrame::captureScreen);
```
Align the dots for wrapped calls.

test/jdk/java/awt/regtesthelpers/PassFailJFrame.java
```diff
@@ -184,6 +262,53 @@ public void windowClosing(WindowEvent e) {
...
+        }
+
```
aivanov-jdk, line 306: Maybe use `switch` statement with a `default` case which throws an exception rather than shows success message. (`if else if else` will also do.)

Yeah, it's a private method yet it's better to handle an unexpected value.

test/jdk/java/awt/regtesthelpers/PassFailJFrame.java
```diff
@@ -184,6 +262,53 @@ public void windowClosing(WindowEvent e) {
...
+        JOptionPane.showMessageDialog(frame, "Screen Captured " +
+                        "Successfully", "Screen Capture",
+                JOptionPane.INFORMATION_MESSAGE);
```
aivanov-jdk, line 309:
```suggestion
        JOptionPane.showMessageDialog(frame,
                "Screen captured successfully",
                "Screen Capture",
                JOptionPane.INFORMATION_MESSAGE);
```
It's better not to split strings unnecessarily. The messages usually use regular capitalisation as opposed to title capitalisation.

test/jdk/java/awt/regtesthelpers/PassFailJFrame.java
```diff
@@ -421,4 +546,72 @@ public static void forceFail(String reason) {
...
+        }
+
+        private void validate() {
+            if (this.title == null) {
+                this.title = TITLE;
+            }
+
+            if (this.instructions == null || this.instructions.length() == 0) {
+                throw new RuntimeException("Please provide the test " +
+                        "instruction for this manual test");
```
aivanov-jdk, line 601: Is `IllegalStateException` or `IllegalArgumentException` better than a generic `RuntimeException`?

aivanov-jdk (changes requested):
I really like your idea with *builder pattern* for creating `PassFailJFrame`. Well done!

Is the purpose of screenshot in `PrintLatinCJKTest` to capture how the printed document looks in a PDF viewer?

aivanov-jdk:
@honkar-jdk could you take a look? You've been involved in early development and updating of the manual testing framework, and you've used it extensively.

aivanov-jdk:
> Yet it may be a hindrance for backporting… if the modified tests don't exist in previous versions or there are other differences…

To facilitate backporting, this PR may be broken into two PRs: the first one which updates the API of `PassFailJFrame`, the second, dependent, PR updates the tests.

I faced some difficulties when I was backporting `PassFailJFrame` to 17u and 11u because some changes to the framework were combined with tests. The `PrintLatinCJKTest` test that is modified here isn't backported to 17u or 11u.

I will definitely want to backport these changes to previous versions of JDK, so I am for *splitting the PR*. Thank you, @DamonGuy, for bringing this up.

honkar-jdk (commented):
@lawrence-andrew 

Here is a consolidated review and few suggestions -

- Firstly, the use of builder pattern for screenshot capability looks well structured and clean. It is easy for extensibility and does not break the existing functionality. That being said, the screenshot capability can additionally be helpful for automated tests too.

- Screenshot capability might be more helpful in case of Client CI / Automated Testing since this feature will make it easier to capture screenshots at different instances when running the test. There are few automated tests currently that capture screenshot when the test fails. The screen capture code is repetitive and redundant and can be changed to use the common functionality that you have designed. 

- In case it is decided to extend this functionality to automated tests then having this is `regtesthelpers/Util` instead of PassFailJFrame would be more appropriate since PassFailJFrame is more of a manual framework and moving the functionality to Util will look cleaner.

- Is the screenshot saved to the test source or jtreg scratch directory? Saving it to scratch would be ideal.

- Additionally you could add the screenshot img filename to the JOptionPane that shows up the "Screenshot successfully captured" message.

honkar-jdk:
@aivanov-jdk 

> Yet it may be a hindrance for backporting… if the modified tests don't exist in previous versions or there are other differences…

I agree, framework changes and test changes can be separated for easier back-porting.

aivanov-jdk:
> Here is a consolidated review and few suggestions -
> 
> Firstly, the use of builder pattern for screenshot capability looks well structured and clean. It is easy for extensibility and does not break the existing functionality.

This is the idea. New tests can use newer syntax, the old tests don't need any modifications.

With a couple more updates which are planned, the usage will be even more streamlined.

> That being said, the screenshot capability can additionally be helpful for automated tests too.

It's a good point. However, taking screenshots doesn't require a lot of code and the screenshot could be limited to the relevant portion of the screen or displayed UI.

> Screenshot capability might be more helpful in case of Client CI / Automated Testing since this feature will make it easier to capture screenshots at different instances when running the test. There are few automated tests currently that capture screenshot when the test fails. The screen capture code is repetitive and redundant and can be changed to use the common functionality that you have designed.

This is one of the problems… A really stand-alone test which has _no dependencies_ is easier to run: compile and run; as of Java 17 you can even skip the compile step and pass the `.java` file directly to `java` launcher which compiles it for you without creating the external `.class` file(s).

If there are dependencies, you have to employ _tricks_ to run the test.

Overall, all GUI tests are somewhat repetitive, tests have duplicate code. It is possible to abstract this into helper classes. In fact, we have such classes: [`SwingTestHelper`](https://github.com/openjdk/jdk/blob/master/test/jdk/javax/swing/regtesthelpers/SwingTestHelper.java), [`JRobot`](https://github.com/openjdk/jdk/blob/master/test/jdk/javax/swing/regtesthelpers/JRobot.java), already mentioned [`Util`](https://github.com/openjdk/jdk/blob/master/test/jdk/javax/swing/regtesthelpers/Util.java)…

I don't say it's bad. I'm for avoiding code duplication. Yet I also appreciate the easiest way possible to run a test. The simpler and easier, the better. It's especially important during code reviews.

> In case it is decided to extend this functionality to automated tests then having this is `regtesthelpers/Util` instead of PassFailJFrame would be more appropriate since PassFailJFrame is more of a manual framework and moving the functionality to Util will look cleaner.

This can easily be done in a follow-up PR.

The only thing that can really be abstracted is the call to `Robot.createScreenCapture` or `Robot.createMultiResolutionScreenCapture` and saving to a file. Well, yes, taking the screenshots of the entire screen can be a utility method.

Then, the jtreg framework does make screenshots in the CI if headful test fails. I've seen it multiple times. They're not always useful because the test UI is usually closed by the time when screenshot is taken. Yet you can see if there are other windows open on the system which could interfere with the test.

> Is the screenshot saved to the test source or jtreg scratch directory? Saving it to scratch would be ideal.

It saves the images to _the current directory_. When the test is run by jtreg, the current directory is the `scratch` directory.

> Additionally you could add the screenshot img filename to the JOptionPane that shows up the "Screenshot successfully captured" message.

There could be multiple files. In fact, I question the need for a modal dialog confirming that the operation completed successfully. If an error occurs, a modal dialog is a good way of notifying. If no error occurs, the message is somewhat redundant.

If the displayed message isn't modal, like a tooltip or a notification which disappears automatically after a certain time, it could be better appreciated. But it's harder to implement.

What could be helpful though is an option to open the current directory in File Explorer, Finder… to allow viewing the collected screenshots. It's more helpful than displaying file names, isn't it?

honkar-jdk:
> This can easily be done in a follow-up PR.
> 
> The only thing that can really be abstracted is the call to `Robot.createScreenCapture` or `Robot.createMultiResolutionScreenCapture` and saving to a file. Well, yes, taking the screenshots of the entire screen can be a utility method.

Sounds good.

> It saves the images to _the current directory_. When the test is run by jtreg, the current directory is the `scratch` directory.

For me, the generated screenshot was being saved to the test folder, as I was running it as standalone test. I had issues running it as jtreg test, but this is probably due to some config issue on my end and nothing related to code changes.

aivanov-jdk:
> > It saves the images to _the current directory_. When the test is run by jtreg, the current directory is the `scratch` directory.
> 
> For me, the generated screenshot was being saved to the test folder, as I was running it as standalone test.

Which is expected because you were running the test as a standalone test. Whatever the current directory is, the images are put in the current directory. The jtreg framework sets the current directory to `scratch`.

test/jdk/java/awt/regtesthelpers/PassFailJFrame.java
honkar-jdk: I agree switch statement could be used here.

DamonGuy:
> > > It saves the images to _the current directory_. When the test is run by jtreg, the current directory is the `scratch` directory.
> > 
> > 
> > For me, the generated screenshot was being saved to the test folder, as I was running it as standalone test.
> 
> Which is expected because you were running the test as a standalone test. Whatever the current directory is, the images are put in the current directory. The jtreg framework sets the current directory to `scratch`.

I think the mixup is similar to what I implemented for my test ([https://github.com/openjdk/jdk/pull/7310](https://github.com/openjdk/jdk/pull/7310)) when trying to save an image. I was initially advised to set a testDir path to save the image to. But it sounds like it's not needed here.

honkar-jdk:
> There could be multiple files. In fact, I question the need for a modal dialog confirming that the operation completed successfully. If an error occurs, a modal dialog is a good way of notifying. If no error occurs, the message is somewhat redundant.
> 
> If the displayed message isn't modal, like a tooltip or a notification which disappears automatically after a certain time, it could be better appreciated. But it's harder to implement.
> 
> What could be helpful though is an option to open the current directory in File Explorer, Finder… to allow viewing the collected screenshots. It's more helpful than displaying file names, isn't it?

We can keep it as it is now or show message dialog only on failure, both sound okay. Maybe the File explorer and notification can be explored later.

test/jdk/java/awt/regtesthelpers/PassFailJFrame.java
honkar-jdk: @aivanov-jdk, @lawrence-andrew 

> If the displayed message isn't modal, like a tooltip or a notification which disappears automatically after a certain time, it could be better appreciated. But it's harder to implement.

We could use Timer for disappearing messages, but might be a overkill here.

```
  JOptionPane pane = new JOptionPane("Screen capture successful", JOptionPane.INFORMATION_MESSAGE);
  final JDialog dialog = pane.createDialog("Screen Capture");
  Timer timer = new Timer(2000, e -> dialog.dispose());
  timer.setRepeats(false);
  timer.start();
  dialog.setVisible(true);
```

aivanov-jdk: > > If the displayed message isn't modal, like a tooltip or a notification which disappears automatically after a certain time, it could be better appreciated. But it's harder to implement.
> 
> We could use Timer for disappearing messages, but might be an overkill here.

How else could one implement a disappearing message?

This approach still shows a *modal* dialog box. While it's shown the user can't do anything but dispose the dialog.

In my opinion, if we show the dialog, let the user to dismiss it.

The message confirming screenshots are successfully taken should be non-intrusive, it may be displayed in a small (translucent) window close to the button which will automatically disappear after a specified timeout. An option to dismiss it before the timeout is even better. Yet it shouldn't prevent access to other UI elements so that the tester could click Fail right away.

However, I may be overthinking it… Implementing such a UI is likely an overkill.

Yet I'm still not convinced a modal dialog box is needed. Consider seeing a modal dialog box that says "File has been saved successfully" each time you select Save command in a text editor or any other editor. You expect it to succeed, thus notification is need only if an error occurred. As the result of the error message, you'll try saving the file again.

aivanov-jdk (approved):

honkar-jdk:
Tested by applying PassFailJFrame changes to manual test. Added functionality - Screen capture code works well and since Builder pattern is used, it does not affect existing manual tests. Changes look good.

@lawrence-andrew  Minor: Copyright year needs to be updated for PassFailJFrame.java

test/jdk/java/awt/regtesthelpers/PassFailJFrame.java
```diff
@@ -421,4 +571,72 @@ public static void forceFail(String reason) {
...
         latch.countDown();
     }
+
+    public static class Builder {
+        private String title;
+        private String instructions;
+        private long testTimeOut;
+        private int rows;
+        private int columns;
+        private boolean enableScreenCapture = false;
```
honkar-jdk, line 581: @lawrence-andrew  you missed updating var `enableScreenCapture` to `screenCapture` here. You can either update only the method name  to screenCapture() and retain same booolean var name - enableScreenCapture or update two other places where they are being used.

test/jdk/java/awt/regtesthelpers/PassFailJFrame.java
```diff
@@ -114,16 +129,62 @@ public PassFailJFrame(String title, String instructions,
...
             invokeAndWait(() -> createUI(title, instructions, testTimeOut,
-                    rows, columns));
+                    rows, columns, enableScreenCapture));
         }
     }
 
+    private PassFailJFrame(Builder builder) throws InterruptedException,
+            InvocationTargetException {
+        this(builder.title, builder.instructions, builder.testTimeOut,
+                builder.rows, builder.columns, builder.enableScreenCapture);
```
honkar-jdk, line 182: `builder.enableScreenCapture` - need to update here as well in case you decide to rename the boolean var.

honkar-jdk (changes requested):

test/jdk/java/awt/regtesthelpers/PassFailJFrame.java
lawrence-andrew: Its better to keep method name and var name separately.

test/jdk/java/awt/regtesthelpers/PassFailJFrame.java
```diff
@@ -421,4 +571,72 @@ public static void forceFail(String reason) {
...
+            return this;
+        }
+
+        public Builder columns(int columns) {
+            this.columns = columns;
+            return this;
+        }
+
+        public Builder screenCapture() {
+            this.screenCapture = true;
```
honkar-jdk, line 609: > Its better to keep method name and var name separately.

@lawrence-andrew You need to revert the variable name here then.

honkar-jdk (changes requested):

test/jdk/java/awt/regtesthelpers/PassFailJFrame.java
aivanov-jdk: They had the same name; Java allows the same name for a field and a method. For consistency with other methods and fields, I agree with @honkar-jdk and I'm for renaming it.

test/jdk/java/awt/regtesthelpers/PassFailJFrame.java
aivanov-jdk, line 609: It does not compile:
```text
test\jdk\java\awt\regtesthelpers\PassFailJFrame.java:609: error: cannot find symbol
            this.screenCapture = true;
                ^
  symbol: variable screenCapture
```

aivanov-jdk (changes requested):

lawrence-andrew:
Compiled and tested & I was able to take the screen shot.

aivanov-jdk (approved):
Looks good to me.

honkar-jdk (approved):
LGTM

PR #13322, opened 2023-04-04
8287325: AArch64: fix virtual threads with -XX:UseBranchProtection=pac-ret

shqking:
### Background

1. PAC-RET branch protection was initially implemented on Linux/AArch64 in JDK-8277204 [1].

2. However, it was broken with the introduction of virtual threads [2], mainly because the continuation freeze/thaw mechanism would trigger stack copying to/from memory, whereas the saved and signed LR on the stack doesn't get re-signed accordingly.

3. PR-9067 [3] tried to implement the re-sign part, but it was not accepted because option "PreserveFramePointer" is always turned on by PAC-RET but this would slow down virtual threads by ~5-20x.

4. As a workaround, JDK-8288023 [4] disables PAC-RET when preview language features are enabled. Note that virtual thread is one preview feature then.

5. Virtual thread will become a permanent feature in JDK-21 [5][6].

### Goal

This patch aims to make PAC-RET compatible with virtual threads.

### Requirements of virtual threads

R-1: Option "PreserveFramePointer" should be turned off. That is, PAC-RET implementation should not rely on frame pointer FP. Otherwise, the fast path in stack copying will never be taken.

R-2: Use some invariant values to stack copying as the modifier, so as to avoid the PAC re-sign for continuation thaw, as the fast path in stack copying doesn't walk the frame.

Note that more details can be found in the discussion [3].

### Investigation

We considered to use (relative) stack pointer SP, thread ID, PACStack [7] and value zero as the candidate modifier.

1. SP: In some scenarios, we need to authenticate the return address in places where the current SP doesn't match the SP on function entry. E.g. see the usage in Runtime1::generate_handle_exception(). Hence, neither absolute nor relative SP works.

2. thread ID (tid): It's invariant to virtual thread, but it's nontrivial to access it from the JIT side. We need 1) firstly resolve the address of current thread (See [8] as an example), and 2) get the tid field in the way like java_lang_Thread::thread_id(). I suppose this would introduce big performance overhead. Then can we turn to use "rthread" register (JavaThread object address) as the modifier? Unfortunately, it's not an invariant to virtual threads and PAC re-sign is still needed.

5. PACStack uses the signed return address of caller as the modifier to sign the callee's return address. In this way, we get one PACed call chain. The modifier should be saved into somewhere around the frame record. Inevitably, FP should be preserved to make it easy to find this modifier in case of some exception scenarios (Recall the reason why we fail to use SP as the modifier).

[...]

PR #14095, opened 2023-05-23
8308503: AArch64: SIGILL when running with -XX:UseBranchProtection=pac-ret on hardware without PAC feature

shqking:
When revisiting the behavior of UseBranchProtection [1], we get one SIGILL error when running with -XX:UseBranchProtection=pac-ret on hardware without PAC.

Problem:

We build and run `java --version` with the following configuration matrix `Config X VMoption X Machine`.

```
  Config = {--enable-branch-protection, null}
  VMoption = {-XX:UseBranchProtection=pac-ret, -XX:UseBranchProtection=standard}
  Machine = {w/ PAC, w/o PAC}
```

VM crashes with SIGILL error for configure `Config=null, VMoption=pac-ret, Machine=w/o PAC`. The unrecognized instruction is `pacia x30, x29`, i.e. `pacia(lr, rfp)` generated by function `MacroAssembler::protect_return_address()`. [2]

Root cause:

1. Instruction `pacia` is not in the NOP space. That's why `Config=null, VMoption=pac-ret` passes on `hardware w/ PAC`, but fails on `hardware w/o PAC`.

2. -XX:UseBranchProtection=pac-ret behaves differently from the document [3], i.e.

```
  In order to use Branch Protection features in the VM,
  --enable-branch-protection must be used
```

`_rop_protection` is not turned off for `Config=null`. That's why `VMoption=pac-ret, Machine=w/o PAC` passes with
`Config=--enable-branch-protection` but fails with `Config=null`.

Fix:

This patch refines the parsing of -XX:UseBranchProtection=pac-ret:

1. We handle "pac-ret" and "standard" in the same way, since only one type of branch protection is implemented for now, i.e. "pac-ret". We may update "standard" in the future if "bti" protection is added.

2. `_rop_protection` is not turned on unless all the three conditions are satisfied [4]. Otherwise, it's kept off and one warning message is emitted.

```
// Enable PAC if this code has been built with branch-protection, the
// CPU/OS supports it, and incompatible preview features aren't enabled.
```

[1] https://bugs.openjdk.org/browse/JDK-8287325?focusedCommentId=14581099&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14581099
[2] https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp#L5976
[3] https://github.com/openjdk/jdk/blob/master/doc/building.md#branch-protection
[4] https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp#L457

---------
### Progress

### Issue
 * [JDK-8308503](https://bugs.openjdk.org/browse/JDK-8308503): AArch64: SIGILL when running with -XX:UseBranchProtection=pac-ret on hardware without PAC feature

### Reviewers
 * [Andrew Haley](https://openjdk.org/census#aph) (@theRealAph - **Reviewer**) ⚠️ Review applies to [fbbf86a8](https://git.openjdk.org/jdk/pull/14095/files/fbbf86a85e4069e0f12fa0a8236dc7d1a60a3b16)
 * [Nick Gasson](https://openjdk.org/census#ngasson) (@nick-arm - **Reviewer**)
 * [Dean Long](https://openjdk.org/census#dlong) (@dean-long - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14095/head:pull/14095` \
`$ git checkout pull/14095`

`$ git checkout pull/14095` \
`$ git pull https://git.openjdk.org/jdk.git pull/14095/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14095`

`$ git pr show -t 14095`

</details>

<a href="https://git.openjdk.org/jdk/pull/14095.diff">https://git.openjdk.org/jdk/pull/14095.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14095#issuecomment-1558548225)

### DIFF
--- a/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
@@ -458,14 +458,14 @@ void VM_Version::initialize() {
     // Enable ROP-protection if
     // 1) this code has been built with branch-protection,
     // 2) the CPU/OS supports it, and
-    // 3) incompatible virtual threads feature isn't enabled.
+    // 3) incompatible VMContinuations isn't enabled.
 #ifdef __ARM_FEATURE_PAC_DEFAULT
     if (!VM_Version::supports_paca()) {
       // Disable PAC to prevent illegal instruction crashes.
       warning("ROP-protection specified, but not supported on this CPU. Disabling ROP-protection.");
     } else if (VMContinuations) {
       // Not currently compatible with continuation freeze/thaw.
-      warning("ROP-protection is incompatible with virtual threads feature. Disabling ROP-protection.");
+      warning("ROP-protection is incompatible with VMContinuations. Disabling ROP-protection.");
     } else {
       _rop_protection = true;
     }
--- a/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
@@ -452,29 +452,23 @@ void VM_Version::initialize() {
 
   if (UseBranchProtection == nullptr || strcmp(UseBranchProtection, "none") == 0) {
     _rop_protection = false;
-  } else if (strcmp(UseBranchProtection, "standard") == 0) {
+  } else if (strcmp(UseBranchProtection, "standard") == 0 ||
+             strcmp(UseBranchProtection, "pac-ret") == 0) {
     _rop_protection = false;
     // Enable PAC if this code has been built with branch-protection, the CPU/OS
     // supports it, and incompatible preview features aren't enabled.
 #ifdef __ARM_FEATURE_PAC_DEFAULT
-    if (VM_Version::supports_paca() && !Arguments::enable_preview()) {
-      _rop_protection = true;
-    }
-#endif
-  } else if (strcmp(UseBranchProtection, "pac-ret") == 0) {
-    _rop_protection = true;
-#ifdef __ARM_FEATURE_PAC_DEFAULT
     if (!VM_Version::supports_paca()) {
-      warning("ROP-protection specified, but not supported on this CPU.");
       // Disable PAC to prevent illegal instruction crashes.
-      _rop_protection = false;
+      warning("ROP-protection specified, but not supported on this CPU. Disabling ROP-protection.");
     } else if (Arguments::enable_preview()) {
       // Not currently compatible with continuation freeze/thaw.
-      warning("PAC-RET is incompatible with virtual threads preview feature.");
-      _rop_protection = false;
+      warning("ROP-protection is incompatible with virtual threads preview feature. Disabling ROP-protection.");
+    } else {
+      _rop_protection = true;
     }
 #else
-    warning("ROP-protection specified, but this VM was built without ROP-protection support.");
+    warning("ROP-protection specified, but this VM was built without ROP-protection support. Disabling ROP-protection.");
 #endif
   } else {
     vm_exit_during_initialization(err_msg("Unsupported UseBranchProtection: %s", UseBranchProtection));
--- a/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
@@ -455,15 +455,17 @@ void VM_Version::initialize() {
   } else if (strcmp(UseBranchProtection, "standard") == 0 ||
              strcmp(UseBranchProtection, "pac-ret") == 0) {
     _rop_protection = false;
-    // Enable PAC if this code has been built with branch-protection, the CPU/OS
-    // supports it, and incompatible preview features aren't enabled.
+    // Enable ROP-protection if
+    // 1) this code has been built with branch-protection,
+    // 2) the CPU/OS supports it, and
+    // 3) incompatible virtual threads feature isn't enabled.
 #ifdef __ARM_FEATURE_PAC_DEFAULT
     if (!VM_Version::supports_paca()) {
       // Disable PAC to prevent illegal instruction crashes.
       warning("ROP-protection specified, but not supported on this CPU. Disabling ROP-protection.");
-    } else if (Arguments::enable_preview()) {
+    } else if (VMContinuations) {
       // Not currently compatible with continuation freeze/thaw.
-      warning("ROP-protection is incompatible with virtual threads preview feature. Disabling ROP-protection.");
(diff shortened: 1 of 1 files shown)


shqking:
> This looks good to me, but please read https://bugs.openjdk.org/browse/JDK-8287325 before you commit anything.

Thanks for reviewing this patch.

I personally thought they are two different issues, and should be fixed separately.

JDK-8287325 is the incompatible issue between pac-ret and virtual threads, whereas
this issue is that pac-ret is enabled even on hardware without the support of PAC feature, leading to SIGILL error.

Regarding JDK-8287325, we have proposed the zero modifier solution (See https://github.com/openjdk/jdk/pull/13322). But we currently set that PR as draft, as we're trying to implement another solution of using "relative sp" as the modifier (which was suggested by Dean Long). We just finished one prototype of relative sp modifier, but there're still several jtreg failures. We're trying to fix them. We will upload our prototype for review once it's ready.

As for your comment, do you mean we should fix these two issues in one patch? Thanks.

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
```diff
@@ -452,29 +452,23 @@ void VM_Version::initialize() {
...
     if (!VM_Version::supports_paca()) {
-      warning("ROP-protection specified, but not supported on this CPU.");
       // Disable PAC to prevent illegal instruction crashes.
-      _rop_protection = false;
+      warning("ROP-protection specified, but not supported on this CPU. Disabling ROP-protection.");
     } else if (Arguments::enable_preview()) {
       // Not currently compatible with continuation freeze/thaw.
-      warning("PAC-RET is incompatible with virtual threads preview feature.");
-      _rop_protection = false;
+      warning("ROP-protection is incompatible with virtual threads preview feature. Disabling ROP-protection.");
```
theRealAph, line 466:
```suggestion
      _rop_protection = false;
      warning("ROP-protection is incompatible with virtual threads preview feature. Disabling ROP-protection.");
```

theRealAph: I presume you didn't intend to change the logic here.

AlanBateman: Virtual threads are proposed to be a permanent feature in JDK 21 so I assume predicating this on Arguments:enable_preview check no longer make sense.

shqking: > Virtual threads are proposed to be a permanent feature in JDK 21 so I assume predicating this on Arguments:enable_preview check no longer make sense.

I knew this issue.
Initially I want to make this patch as small as possible, i.e. only focusing on the SIGILL issue.

> Oh, I see. So it really only makes sense either to make ROP protection work in vthreads, or just disable ROP protection in JDK 21.

It may take some more time for me to "make ROP protection work in vthreads", i.e. resolving JDK-8287325. 
Compared to the zero modifier solution (See https://github.com/openjdk/jdk/pull/13322), I personally think relative sp solution is better, but we still have many jtreg failures to fix. I'm not sure if we can make it ready in JDK21 release cycle.

Perhaps, we can firstly disable ROP if `VMContinuations` is on. I mean, change `Arguments::enable_preview()` to `VMContinuations`.
In this way, ROP protection is enabled only with VM options `-XX:UseBranchProtection=standard|pac-ret -XX:+UnlockExperimentalVMOptions -XX:-VMContinuations` on hardware with the support of PAC feature.

theRealAph: > Perhaps, we can firstly disable ROP if `VMContinuations` is on. I mean, change `Arguments::enable_preview()` to `VMContinuations`.
> In this way, ROP protection is enabled only with VM options `-XX:UseBranchProtection=standard|pac-ret -XX:+UnlockExperimentalVMOptions -XX:-VMContinuations` on hardware with the support of PAC feature.

That's probably the best we can do for 21.

shqking:
In the new commit, we disable ROP-protection if `VMContinuations` is on.
Hope this can address your comments.

Test for the new commit:

I built and ran `java --version` with the following configuration matrix `Config X VMoption X Machine`. 
`_rop_protection` behaves as expected.

```
  Config = {--enable-branch-protection, null}
  VMoption = {-XX:UseBranchProtection=pac-ret, -XX:UseBranchProtection=pac-ret -XX:-VMContinuations}
  Machine = {w/ PAC, w/o PAC}
```

I further ran tier1~3 with `Config=--enable-branch-protection, VMoption=-XX:UseBranchProtection=pac-ret -XX:-VMContinuations, Machine=w/ PAC`, and only got one jtreg failiure, i.e.`jdk/incubator/concurrent/ScopedValue/StressStackOverflow.java`.
I don't think it's related to PAC-RET. See https://bugs.openjdk.org/browse/JDK-8308609

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
```diff
@@ -452,29 +452,25 @@ void VM_Version::initialize() {
...
-      warning("ROP-protection specified, but not supported on this CPU.");
       // Disable PAC to prevent illegal instruction crashes.
-      _rop_protection = false;
-    } else if (Arguments::enable_preview()) {
+      warning("ROP-protection specified, but not supported on this CPU. Disabling ROP-protection.");
+    } else if (VMContinuations) {
       // Not currently compatible with continuation freeze/thaw.
-      warning("PAC-RET is incompatible with virtual threads preview feature.");
-      _rop_protection = false;
+      warning("ROP-protection is incompatible with virtual threads feature. Disabling ROP-protection.");
```
theRealAph, line 468:
```suggestion
      warning("ROP-protection is incompatible with VMContinuations. Disabling ROP-protection.");
```

theRealAph (approved):

theRealAph (approved):

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
shqking: Updated in the latest commit. Thanks.

shqking:
Can anyone else help take a look at this patch? 
Maybe @nick-arm @dean-long or @AlanBateman Thanks.

nick-arm (approved):

dean-long (approved):

shqking:
Thanks for your review!
GHA tests are passed. I'm integrating it.

/integrate

PR #14098, opened 2023-05-23
8308645: Javadoc of FFM API needs to be refreshed

mcimadamore:
As the FFM API evolved over time, some parts of the javadoc went out of sync. Now that most of the API is stable, it is a good time to look again at the javadoc as a whole, and bring some more consistency.

While most of the changes in this PR are stylistic, I'd like to call out few things which resulted in API changes:

* `MemorySegment::asSlice(long, long, long)` did not (incorrectly) specified requirement that its alignment parameter must be a power of two.

* `MemoryLayout::sliceHandle` did not require the absence of dereference paths in the provided layout path. While that is technically possible to allow, currently the method is specified as returning a "slice" corresponding to some "nested layout", so following pointers seems incompatible with the spec. I have decided to disallow for now - we can always compatibly enable it later, if required.

* `MemorySegment::copy` - most of the overloads did not specify that `UnsupportedOperationException` is thrown if the destination segment is read-only.

* In several places, an extra `@throws IllegalArgumentException` or `@throws IllegalArgumentException` has been added to capture cases where element size * index computation can overflow. This is true for all the element-wise segment copy methods, for the indexed accessors in memory segment (e.g. `MemorySegment.getAtIndex(ValueLayout.OfInt, long)`), as well as for `SegmentAllocator::allocateArray(MemoryLayout, long)`.

In all these cases (except for overflows), new tests have been added to cover the additional API changes (a CSR will also be filed to cover these).

The class with most changes is `MemoryLayout`. I realized that the javadoc there was a bit sloppy around the definition of "layout paths". Now there are several subsection in the class javadoc, which explain how different kinds of paths can be used. I have introduced the notion of "open path elements" to denote those path elements that are not fully specified, and result in additional var handle coordinates to be added. There is also now a definition of what it means for a layout path to be "well-formed", so that all methods accepting a layout path can simply refer to it (this definition is tricky to give inline in the javadoc of the various path-accepting methods, as it depends on many factors).

Another biggie change was in `MemorySegment` as now all dereference methods state precisely their spatial checks requirements, as well also specifying when the API can throw because of an overflow during offset computation.

Finally, I've left most of the snippets alone, as they are being dealt with in https://github.com/openjdk/jdk/pull/14030

---------
### Progress

### Issues
 * [JDK-8308645](https://bugs.openjdk.org/browse/JDK-8308645): Javadoc of FFM API needs to be refreshed (**Bug** - P3)
 * [JDK-8309398](https://bugs.openjdk.org/browse/JDK-8309398): ValueLayout:: arrayElementVarHandle doesn't throws UnsupportedOperationException - if byteAlignment() &gt; byteSize() (**Bug** - P3)
 * [JDK-8308812](https://bugs.openjdk.org/browse/JDK-8308812): SequenceLayout::withElementCount(long elementCount) doesn't throw IllegalArgumentException - if elementCount &lt; 0 for some cases (**Bug** - P3)

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - **Reviewer**) ⚠️ Review applies to [44b382dd](https://git.openjdk.org/jdk/pull/14098/files/44b382dd387e5817e4d23ec7a9662f93972c7495)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14098/head:pull/14098` \
`$ git checkout pull/14098`

`$ git checkout pull/14098` \
`$ git pull https://git.openjdk.org/jdk.git pull/14098/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14098`

`$ git pr show -t 14098`

</details>

<a href="https://git.openjdk.org/jdk/pull/14098.diff">https://git.openjdk.org/jdk/pull/14098.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14098#issuecomment-1559638558)


mcimadamore:
Javadoc: https://cr.openjdk.org/~mcimadamore/jdk/8308645/javadoc/java.base/java/lang/foreign/package-summary.html

mcimadamore:
I brought over some javadoc improvements from https://github.com/openjdk/jdk/pull/14037 (which has been withdrawn). These improvements should allow us to enable support for Linker in x86 platforms, as they define the notion of "what is a linker supported layout" much more precisely, and in a way that is not dependent on natural alignment (for value layouts).

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -163,6 +170,68 @@
...
+ * {@snippet lang=java :
+ * StructLayout RECTANGLE = MemoryLayout.structLayout(
+ *         ValueLayout.ADDRESS.withTargetLayout(
+ *                 MemoryLayout.sequenceLayout(4,
+ *                         MemoryLayout.structLayout(
+ *                                 ValueLayout.JAVA_INT.withName("x"),
+ *                                 ValueLayout.JAVA_INT.withName("y")
+ *                         ).withName("point")
+*                   )
+*          ).withName("points")
```
ExE-Boss, line 187: Wrong indentation:
```suggestion
 *                 )
 *         ).withName("points")
```

src/java.base/share/classes/java/lang/foreign/Arena.java
```diff
@@ -256,14 +259,14 @@ static Arena ofShared() {
...
+     * @param byteSize the size (in bytes) of the off-heap region of memory backing the native memory segment.
      * @param byteAlignment the alignment constraint (in bytes) of the off-heap region of memory backing the native memory segment.
      * @return a new native memory segment.
-     * @throws IllegalArgumentException if {@code bytesSize < 0}, {@code alignmentBytes <= 0}, or if {@code alignmentBytes}
+     * @throws IllegalArgumentException if {@code bytesSize < 0}, {@code byteAlignment <= 0}, or if {@code byteAlignment}
      * is not a power of 2.
      * @throws IllegalStateException if this arena has already been {@linkplain #close() closed}.
      * @throws WrongThreadException if this arena is confined, and this method is called from a thread {@code T}
-     * other than the arena owner thread.
+     * other than the arena's owner thread.
```
AlanBateman, line 269: "thread T" hints that "T" will be used later, maybe it's not needed.

src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
```diff
@@ -34,9 +34,9 @@
 import jdk.internal.javac.PreviewFeature;
 
 /**
- * A function descriptor models the signature of foreign functions. A function descriptor is made up of zero or more
- * argument layouts and zero or one return layout. A function descriptor is typically used when creating
- * {@linkplain Linker#downcallHandle(MemorySegment, FunctionDescriptor, Linker.Option...) downcall method handles} or
+ * A function descriptor models the signature of a foreign function. A function descriptor is made up of zero or more
+ * argument layouts and zero or one return layout. A function descriptor is used to create
```
AlanBateman, line 38: You might want want to put a comma after "layouts" to avoid any ambiguity.

src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
```diff
@@ -49,21 +49,21 @@
...
 
     /**
-     * {@return the return layout (if any) associated with this function descriptor}
+     * {@return the return layout (if any) of this function descriptor}
      */
     Optional<MemoryLayout> returnLayout();
 
     /**
-     * {@return the argument layouts associated with this function descriptor (as an immutable list)}.
+     * {@return the argument layouts of this function descriptor (as an immutable list)}.
```
AlanBateman, line 57: I assume this should be "unmodifiable" rather than immutable here.

src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
```diff
@@ -126,10 +123,11 @@ static FunctionDescriptor of(MemoryLayout resLayout, MemoryLayout... argLayouts)
     }
 
     /**
-     * Creates a function descriptor with the given argument layouts and no return layout.
+     * Creates a function descriptor with the given argument layouts and no return layout.  This is useful to model functions
+     * which return no values.
```
AlanBateman, line 127: I think I would use "that return" rather than "which return" here.

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -330,8 +353,8 @@
  * unsafely, resize the segment to the desired size (100, in this case). It might also be desirable to
  * attach the segment to some existing {@linkplain Arena arena}, so that the lifetime of the region of memory
  * backing the segment can be managed automatically, as for any other native segment created directly from Java code.
- * Both these operations are accomplished using the restricted {@link MemorySegment#reinterpret(long, Arena, Consumer)}
- * method, as follows:
+ * Both these operations are accomplished using the restricted method {@link MemorySegment#reinterpret(long, Arena, Consumer)},
```
AlanBateman, line 356: I think this should be "Both of these operations".

src/java.base/share/classes/java/lang/foreign/SymbolLookup.java
```diff
@@ -54,7 +54,7 @@
  *     <li>It can be passed to an existing {@linkplain Linker#downcallHandle(FunctionDescriptor, Linker.Option...) downcall method handle}, as an argument to the underlying foreign function.</li>
  *     <li>It can be {@linkplain MemorySegment#set(AddressLayout, long, MemorySegment) stored} inside another memory segment.</li>
  *     <li>It can be used to access the region of memory backing a global variable (this requires
- *     {@link MemorySegment#reinterpret(long)} () resizing} the segment first).</li>
+ *     {@link MemorySegment#reinterpret(long) resizing} the segment first).</li>
```
AlanBateman, line 57: This one probably should be linkplain.

src/java.base/share/classes/java/lang/foreign/SequenceLayout.java
```diff
@@ -70,15 +70,16 @@ public sealed interface SequenceLayout extends MemoryLayout permits SequenceLayo
      * Returns a sequence layout with the same element layout, alignment constraint and name as this sequence layout,
      * but with the specified element count.
      * @param elementCount the new element count.
-     * @return a sequence layout with the given element count.
-     * @throws IllegalArgumentException if {@code elementCount < 0}.
+     * @return a sequence layout with the same characteristics of this layout, but with the given element count.
+     * @throws IllegalArgumentException if {@code elementCount} is negative.
+     * @throws IllegalArgumentException if {@code elementLayout.bitSize() * elementCount} overflows.
```
AlanBateman, line 75: Shouldn't the exception descriptions be combined to avoid IAE being listed twice in the generated javadoc?

src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
```diff
@@ -311,6 +321,7 @@ default MemorySegment allocate(MemoryLayout layout) {
      * @param elementLayout the array element layout.
      * @param count the array element count.
      * @return a segment for the newly allocated memory block.
+     * @throws IllegalArgumentException if {@code elementLayout.byteSize() * count} overflows.
      * @throws IllegalArgumentException if {@code count < 0}.
```
AlanBateman, line 325: Another case where the IAE descriptions should probably be combined.

src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
```diff
@@ -347,8 +358,9 @@ default MemorySegment allocate(long byteSize) {
      * obtained from the provided segment. Each new allocation request will return a new slice starting at the
      * current offset (modulo additional padding to satisfy alignment constraint), with given size.
      * <p>
-     * When the returned allocator cannot satisfy an allocation request, e.g. because a slice of the provided
-     * segment with the requested size cannot be found, an {@link IndexOutOfBoundsException} is thrown.
+     * The returned allocator throws {@link IndexOutOfBoundsException} when a slice of the provided
+     * segment with the requested size and alignment cannot be found.
+     * @implNote A slicing allocator is not <em>thread-safe</em>.
```
AlanBateman, line 363: The implNote about thread safety makes me wonder if the SegmentAllocator should say something about thread safety, e.g. should it specify that the allocate methods are thread safe?

test/jdk/java/foreign/TestLayoutPaths.java
```diff
@@ -122,10 +122,15 @@ public void testIncompleteAccess() {
         seq.varHandle(sequenceElement());
     }
 
+    public void testByteOffsetHandleRange() {
```
JornVernee, line 125: Missing `@Test`?

test/jdk/java/foreign/TestSegmentCopy.java
```diff
@@ -75,6 +77,25 @@ public void testByteCopy(SegmentKind kind1, SegmentKind kind2) {
...
+        MemorySegment s2 = kind2.makeSegment(TEST_BYTE_SIZE);
+        // check failure with read-only dest
+        MemorySegment.copy(s1, Type.BYTE.layout, 0, s2.asReadOnly(), Type.BYTE.layout, 0, 0);
+    }
+
+    @Test(expectedExceptions = IndexOutOfBoundsException.class, dataProvider = "types")
+    public void testBadOverflow(Type type) {
+        if (type.layout.byteSize() > 1) {
+            MemorySegment segment = MemorySegment.ofArray(new byte[100]);
+            // check failure with read-only dest
```
JornVernee, line 92: Comment seems unrelated?
```suggestion
```

JornVernee:
Overall a great cleanup, nice work!

src/java.base/share/classes/java/lang/foreign/SequenceLayout.java
mcimadamore: I think @PaulSandoz  suggested this idiom in other places where the `@throws` clause was very hard to break up, so I thought that would have been preferrable than having a single para. But if there are strong opinions, I could tweak in different ways, perhaps with bullet lists (but that will require a lot of work).

src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
```diff
@@ -365,14 +377,15 @@ static SegmentAllocator slicingAllocator(MemorySegment segment) {
...
-     * This allocator can be useful to limit allocation requests in case a client
+     * The returned allocator throws {@link IndexOutOfBoundsException} when a slice of the provided
+     * segment with the requested size and alignment cannot be found.
+     *
+     * @apiNote A prefix allocator can be useful to limit allocation requests in case a client
      * knows that they have fully processed the contents of the allocated segment before the subsequent allocation request
      * takes place.
-     * <p>
-     * While the allocator returned by this method is <em>thread-safe</em>, concurrent access on the same recycling
+     * @implNote While a prefix allocator is <em>thread-safe</em>, concurrent access on the same recycling
```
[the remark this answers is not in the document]
mcimadamore: I think the term is used pretty much all over the javadoc - e.g. look for this preamble:
```
 * <a href="{@docRoot}/java.base/java/lang/doc-files/ValueBased.html">value-based</a>,
 * immutable and thread-safe
 ```

File: src/java.base/share/classes/java/lang/foreign/MemorySegment.java [hunk 14]
```diff
@@ -930,23 +933,22 @@ default long mismatch(MemorySegment other) {
 
     /**
      * Wraps this segment in a {@link ByteBuffer}. Some properties of the returned buffer are linked to
-     * the properties of this segment. For instance, if this segment is <em>immutable</em>
-     * (e.g. the segment is a read-only segment, see {@link #isReadOnly()}), then the resulting buffer is <em>read-only</em>
-     * (see {@link ByteBuffer#isReadOnly()}). Additionally, if this is a native segment, the resulting buffer is
-     * <em>direct</em> (see {@link ByteBuffer#isDirect()}).
+     * the properties of this segment. For instance, if this segment is a {@linkplain #isReadOnly() read-only segment},
+     * then the resulting buffer is also {@linkplain ByteBuffer#isReadOnly() read-only}. Additionally, if this is a native
+     * segment, the resulting buffer is a {@linkplain ByteBuffer#isDirect() direct buffer}.
```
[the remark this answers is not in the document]
mcimadamore: Good idea - it is not trivial to capture all the properties in a single para.

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -601,7 +639,10 @@ static PathElement dereferenceElement() {
     String toString();
 
     /**
-     * Creates a padding layout with the given byte size and a byte-alignment of one.
+     * Creates a padding layout with the given byte size. The alignment constraint of the returned layout
+     * is 1. As such, regardless of its size, in the absence of an {@linkplain #withByteAlignment(long) explicit}
+     * alignment constraint, a padding layout does not affect the alignment constraint of the group or sequence layout
+     * it is nested into.
```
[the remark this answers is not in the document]
mcimadamore: I let sequence of padding slide on the basis that groups of padding are allowed. But it's a somewhat arbitrary line. That said, we do use stuff like that in jextract, where we model types we don't support (or bitfields) as padding. I think starting to ban padding in certain places would result in a less compositional API - and perhaps tools would have to workaround some of these limitations.

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -355,54 +427,47 @@ default VarHandle varHandle(PathElement... elements) {
...
      * @return the layout selected by the layout path in {@code elements}.
-     * @throws IllegalArgumentException if the layout path does not select any layout nested in this layout,
-     * or if the layout path contains one or more path elements that select one or more sequence element indices
-     * (see {@link PathElement#sequenceElement(long)} and {@link PathElement#sequenceElement(long, long)}).
-     * @throws IllegalArgumentException if the layout path contains one or more dereference path elements
-     * (see {@link PathElement#dereferenceElement()}).
+     * @throws IllegalArgumentException if the layout path is not <a href="#well-formedness">well-formed</a> for this layout.
+     * @throws IllegalArgumentException if the layout path contains one or more <a href=#deref-path-elements>dereference path elements</a>.
+     * @throws IllegalArgumentException if the layout path contains one or more path elements that select one or more
+     * sequence element indices, such as {@link PathElement#sequenceElement(long)} and {@link PathElement#sequenceElement(long, long)}).
```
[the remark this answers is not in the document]
mcimadamore: This is deliberate (but subtle). Basically, for `select` we just want to select a target layout (we don't care which). So the method has a preference for open sequence layout paths: there's no access or offset to model here, the method just needs to end up into some layout at the end of the path. One might argue that these restrictions are unnecessary - but I didn't feel strongly enough to drop them.

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -336,16 +411,13 @@ default MethodHandle byteOffsetHandle(PathElement... elements) {
...
+     * @apiNote The resulting var handle features certain <em>access mode restrictions</em>, which are common to all
+     * {@linkplain MethodHandles#memorySegmentViewVarHandle(ValueLayout) memory segment view handles}.
      *
      * @param elements the layout path elements.
-     * @return a var handle which can be used to access a memory segment at the (possibly nested) layout selected by the layout path in {@code elements}.
-     * @throws UnsupportedOperationException if the layout path has one or more elements with incompatible alignment constraint.
-     * @throws IllegalArgumentException if the layout path in {@code elements} does not select a value layout (see {@link ValueLayout}).
-     * @throws IllegalArgumentException if the layout path in {@code elements} contains a {@linkplain PathElement#dereferenceElement()
-     * dereference path element} for an address layout that has no {@linkplain AddressLayout#targetLayout() target layout}.
+     * @return a var handle that accesses a memory segment at the offset selected by the given layout path.
```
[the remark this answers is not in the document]
mcimadamore: Yeah - I meant what you said - but now that you said it, I also saw how what I've written can be prone to an alternate (and wrong) interpretation. I'll clarify.

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -311,14 +386,14 @@ default MethodHandle byteOffsetHandle(PathElement... elements) {
...
+     * Additionally, the provided dynamic values must conform to bounds which are derived from the layout path, that is,
      * {@code 0 <= x_i < b_i}, where {@code 1 <= i <= n}, or {@link IndexOutOfBoundsException} is thrown.
      * <p>
-     * Multiple paths can be chained, by using {@linkplain PathElement#dereferenceElement() dereference path elements}.
-     * A dereference path element allows to obtain a native memory segment whose base address is the address obtained
-     * by following the layout path elements immediately preceding the dereference path element. In other words,
-     * if a layout path contains one or more dereference path elements, the final address accessed by the returned
-     * var handle can be computed as follows:
+     * Multiple paths can be chained, by using <a href=#deref-path-elements>dereference path elements</a>.
+     * A dereference path element allows to obtain a native memory segment whose base address is the address value
```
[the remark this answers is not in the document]
mcimadamore: I will also look for some other way to say this - typically when I end up with such convoluted sentences there's a better way to write it :-)

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -434,23 +455,22 @@
...
-     * @return a linker for the ABI associated with the underlying native platform.
      * @throws UnsupportedOperationException if the underlying native platform is not supported.
      */
     static Linker nativeLinker() {
         return SharedUtils.getSystemLinker();
     }
 
     /**
-     * Creates a method handle which is used to call a foreign function with the given signature and address.
+     * Creates a method handle which is used to call a foreign function with the given signature and symbol.
```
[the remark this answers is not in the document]
mcimadamore: the idea behind this is to connect with the javadoc of `SymbolLookup` which defines and then uses symbol all over the place.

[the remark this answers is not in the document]
mcimadamore: Maybe a linkplan could help?

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -417,10 +440,8 @@
  * through an invalid linkage request (e.g. by specifying a function descriptor featuring too many argument layouts),
  * the result of such interaction is unspecified and can lead to JVM crashes.
  * <p>
- * When creating upcall stubs the linker runtime validates the type of the target method handle against the provided
- * function descriptor and report an error if any mismatch is detected. As for downcalls, JVM crashes might occur,
- * if the foreign code casts the function pointer associated with an upcall stub to a type
- * that is incompatible with the provided function descriptor. Moreover, if the target method
+ * When an upcall stub is passed to a foreign function, a JVM crash might occur, if the foreign code casts the function pointer
+ * associated with the upcall stub to a type that is incompatible with the type of the upcall stub. Moreover, if the method
```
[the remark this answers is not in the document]
mcimadamore: I agree that's better - I was also puzzled by the text (I think it's there from before, just shuffled?)

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -197,16 +197,39 @@
  * </tbody>
  * </table></blockquote>
  * <p>
- * All the native linker implementations limit the function descriptors that they support to those that contain
- * only so-called <em>canonical</em> layouts. A canonical layout has the following characteristics:
+ * All native linker implementations operate on a subset of memory layouts. More formally, a layout {@code L}
+ * is supported by a native linker {@code NL} iff:
```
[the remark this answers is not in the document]
mcimadamore: I'll think about it - it's a bit problematic to specify in terms of "at least" because we need to be able to refer to "supported by NL" recursively in the text.

src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
mcimadamore: I think SegmentAllocator should be agnostic re. thread safety. Allocation is a world of compromises, where if you give up (thread) safety you can gain more performance (and viceversa). So I think having a "one size fits all" thread-safety blanket might not work very well.

src/java.base/share/classes/java/lang/foreign/Arena.java
mcimadamore: Correct - this is mostly a copy/paste issue for similar throws clauses in methods outside arena.

src/java.base/share/classes/java/lang/foreign/Linker.java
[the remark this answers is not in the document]
JornVernee: Okay, but `SymbolLookup` also says this: "A <em>symbol lookup</em> retrieves the address of a symbol in one or more libraries." and "This symbol lookup, which is known as a <em>default lookup</em>, helps clients to quickly find addresses of well-known symbols."

This seems to imply that "symbol" and "address" are two different things. I get the connection between the address and the SymbolLookup, but I'll note that it's also not necessary for a function address to come from a SymbolLookup (e.g. the address could be returned from native code, or be the address of an upcall stub).

So, I feel like sticking with "address" is better. Maybe a paragraph (or just a `@see` tag) could be added to refer to SymbolLookup instead?

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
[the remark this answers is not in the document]
JornVernee: Thanks, I agree with that analogy

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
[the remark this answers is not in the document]
JornVernee: Ok, thanks for explaining.

src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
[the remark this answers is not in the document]
JornVernee: Right, I guess I'm asking: can/should we do better? ;) It might be nice to define the term once and for all for the entire JDK, similar to how 'reachability' also has a central definition.

But, arguably, that seems like too much scope creep for this PR, so let's save it for a potential followup.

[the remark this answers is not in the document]
minborg: I think there is something we could start off from: https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/annotations/ThreadSafe.html

This could be made under a separate issue.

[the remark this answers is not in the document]
minborg: https://bugs.openjdk.org/browse/JDK-8309342

src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
AlanBateman: > I think SegmentAllocator should be agnostic re. thread safety. Allocation is a world of compromises, where if you give up (thread) safety you can gain more performance (and viceversa). So I think having a "one size fits all" thread-safety blanket might not work very well.

I'm just trying to think about the practical implications of this.  Arena is specified to be thread-safe so I can allocate from a non-confined Arena, and from concurrent threads, without needing to coordinate.  However, if you hand me a SegmentAllocator that is not an Arena then I don't know if I need to coordinate allocation with other parties. I'm not too concerned about the slicingAllocator and prefixAllocator factory methods as the result SegmentAllocators will likely be wrapped.

mcimadamore: Note that, beside thread-safety, there are other things that are inconvenient about a segment allocator:
* an allocator is not constrained to return segments with the same lifetime;
* an allocator does not provide any guarantee re. aliasing memory. Same memory region could be written over and over (this is what prefixAllocator does);

These facts make it hard for clients SegmentAllocator to interact with a SegmentAllocator they don't own (unless they have to allocate precisely _one_ segment, on behalf of the client, and return it to them - which, not surprisingly is also what the Linker does). So, with SegmentAllocators there's always an element of trust: a client gave me an allocator to obtain a slab of memory, I trust the client to have done what it needed to do so that my operation can be safe. If I happen to allocate garbage (as a result of a race) it's on them.

If you need to allocate multiple segments in a single "session", and have some lifetime guarantees of the segments that have been allocated and/or you want to make sure that what you get is not garbage (e.g. thread-safety, aliasing, etc.), then you need Arena.

src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
mcimadamore: This was derived from MethodType::parameterList, which also mentions immutable. I'm fine changing to "unmodifiable".

src/java.base/share/classes/java/lang/foreign/Linker.java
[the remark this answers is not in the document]
mcimadamore: I'll replace "if" with "iff". I think to address your comment on linker options affecting what's supported, we can add a paragraph in that sense (but I think I'd like to do that after we fix https://git.openjdk.org/jdk/pull/14225, so that I can add a link to somewhere concrete.

src/java.base/share/classes/java/lang/foreign/ValueLayout.java
```diff
@@ -66,7 +68,7 @@ public sealed interface ValueLayout extends MemoryLayout permits
      * but with the specified byte order.
      *
      * @param order the desired byte order.
-     * @return a value layout with the given byte order.
+     * @return a value layout with the same characteristics as this layout, but with the given byte order.
```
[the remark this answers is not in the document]
mcimadamore: Possibly. The text is consistent in using different wording between the header and return (as I did a pass in that sense), but perhaps that's a bug.

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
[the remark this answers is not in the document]
mcimadamore: Actually, looking back at this I'm not sure it's correct? We have an input memory segment MS. And a layout path. A layout path identifies an offset O at which MS has to be accessed in order to read a value (of some type). I'm not sure any of this depends on "base address" ?

mcimadamore:
Thanks for the comments so far, I've updated the javadoc in place:
https://cr.openjdk.org/~mcimadamore/jdk/8308645/javadoc/java.base/java/lang/foreign/package-summary.html

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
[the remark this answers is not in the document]
JornVernee: Hmm, that's true. The base address that we read and turn into a memory segment isn't necessarily accessed. So, on second thought, I think the current text is good as it is.

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -930,23 +933,24 @@ default long mismatch(MemorySegment other) {
...
+     * the properties of this segment. More specifically, the resulting buffer has the following characteristics:
+     * <ul>
+     * <li>It is {@linkplain ByteBuffer#isReadOnly() read-only}, if this segment is a
+     * {@linkplain #isReadOnly() read-only segment};</li>
+     * <li>Its {@linkplain ByteBuffer#position() position} is set to zero;
+     * <li>Its {@linkplain ByteBuffer#capacity() capacity} and {@linkplain ByteBuffer#limit() limit}
+     * are both set to this segment' {@linkplain MemorySegment#byteSize() size}. For this reason, a byte buffer
+     * cannot be returned if this segment's size is greater than {@link Integer#MAX_VALUE};</li>
+     * <li>It is a {@linkplain ByteBuffer#isDirect() direct buffer}, if this is a native segment.</li>
+     * </ul>
```
JornVernee, line 945: Nice!

JornVernee (approved):

test/jdk/java/foreign/TestLayoutPaths.java
JornVernee: FWIW, it looks like this needs `@Test(expectedExceptions = IllegalArgumentException.class)` since range elements are not allowed for `byteOffsetHandle`

mcimadamore: Note sure - javadoc says:
```
* @throws IllegalArgumentException if the layout path is not <a href="#well-formedness">well-formed</a> for this layout.
     * @throws IllegalArgumentException if the layout path contains one or more <a href=#deref-path-elements>dereference path elements</a>.
```

Which seems correct. The handle version can accept indices, so that's ok?

mcimadamore: Note that the javadoc has been rectified here (as part of the PR) as I think it has always been wrong in the past (probably cut and paste issue from the non-MH version?)

JornVernee: But, the implementation of `byteOffsetHandle` rejects range elements: https://github.com/openjdk/jdk/pull/14098/files/265e0d5b082ad54d1cadea1d4a8bd844e72e3926#diff-90ecdacbee63a90f6e695a8bb4c1ad69b591602501613a774561c4032f01fb95R352

So this test fails (at least it does here...)

mcimadamore: I see, the issue is with the impl (I agree) not with the javadoc as I understood.

mcimadamore:
Closing, as the review is complete and I have to rebase against the forked 21 repo. A similar PR will pop up soon :-)

kevinrushforth:
@mcimadamore Is there a reason that this shouldn't be integrated into the mainline?? Absent a compelling reason, fixes should be integrated first into the mainline (for 22) and then backported to 21.

mcimadamore:
> @mcimadamore Is there a reason that this shouldn't be integrated into the mainline?? Absent a compelling reason, fixes should be integrated first into the mainline (for 22) and then backported to 21.

This fix was targeting 21 - but then the repo has forked. It is my understanding that fixes applied to the 21 fork will be forward-ported to 22 automatically.

mcimadamore:
Re-opening - as per new policies, changes will NOT be automatically forward ported (see [1]), so we need a separate PR for 22.

[1] - https://mail.openjdk.org/pipermail/jdk-dev/2023-June/007894.html

PR #14099, opened 2023-05-23
JDK-8264488: Improve warning for module names ending with digits

overheadhunter:
Since this is a rather trivial change, I didn't start a new discussion, as I deem this sufficiently discussed [back in early 2021](https://mail.openjdk.org/pipermail/discuss/2021-March/005779.html).

---------
### Progress

### Issue
 * [JDK-8264488](https://bugs.openjdk.org/browse/JDK-8264488): Improve warning for module names ending with digits (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14099/head:pull/14099` \
`$ git checkout pull/14099`

`$ git checkout pull/14099` \
`$ git pull https://git.openjdk.org/jdk.git pull/14099/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14099`

`$ git pr show -t 14099`

</details>

<a href="https://git.openjdk.org/jdk/pull/14099.diff">https://git.openjdk.org/jdk/pull/14099.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14099#issuecomment-1559639326)

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties
@@ -1861,7 +1861,7 @@ compiler.warn.finally.cannot.complete=\
 
 # 0: name
 compiler.warn.poor.choice.for.module.name=\
-    module name component {0} ends with a digit. Module names should not encode version information in module names should avoid terminal digits
+    module name component {0} ends with a digit. Module names should not encode version information in module names
 
 # 0: string
 compiler.warn.incubating.modules=\
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties
@@ -1861,7 +1861,7 @@ compiler.warn.finally.cannot.complete=\
 
 # 0: name
 compiler.warn.poor.choice.for.module.name=\
-    module name component {0} should avoid terminal digits
+    module name component {0} ends with a digit. Module names should not encode version information in module names should avoid terminal digits
 
 # 0: string
 compiler.warn.incubating.modules=\
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties
@@ -1861,7 +1861,7 @@ compiler.warn.finally.cannot.complete=\
 
 # 0: name
 compiler.warn.poor.choice.for.module.name=\
-    module name component {0} ends with a digit. Module names should not encode version information in module names
+    terminal digits in module name {0} look like a version number
 
 # 0: string
 compiler.warn.incubating.modules=\


overheadhunter:
> module name {0} should not end with a number, since it may be wrongly interpreted as a version number

Still confusing: The warning is not to avoid module consumers from _wrongly interpreting_ terminal digits such as in PKCS11 as a version number. It rather warns the module publisher to not append any version number to the module name.

How about: `module name {0} ends on terminal digits, which must not encode a version number`

src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties
```diff
@@ -1861,7 +1861,7 @@ compiler.warn.finally.cannot.complete=\
 
 # 0: name
 compiler.warn.poor.choice.for.module.name=\
-    module name component {0} should avoid terminal digits
+    module name component {0} ends with a digit. Module names should not encode version information in module names
```
jonathan-gibbons, line 1864: Duplicate text

_Module names_. should not ... in _module names_

overheadhunter:
> "terminal digits" means "the digits at the end", so there's no need to say "ends on" -- it would just be "module name {0} has terminal digits". In any case, "terminal digits" is a dense phrase that appeals to the technical construct of "Java digit" without spelling it out -- this is hard work for readers -- "number" is more straightforward.

I agree on the "terminal" and "ends on" being redundant.

However, digits don't necessarily need to be a number. This depends on whether or not they are used to denote a numeric value (see example below). Without context, they are just a set of characters. Since we strive for a concise error message, we should be very precise here.

> Also, the phrase "must not" is inappropriate for a warning message. "must not" is for error messages.

Agreed, that's why the warning says "should not" to the digits, while pointing out that it "must not" be a version number.

> 
> Here's a rewrite of your suggestion:
> 
> module name {0} ends with a number, which should not encode a version

This would be incorrect, though. While ["the specification was revised to allow digits at the end of module names"](https://openjdk.org/projects/jigsaw/spec/minutes/2017-05-18#VersionsInModuleNames) for non-version cases, we still don't want to allow adding version numbers (according to my understanding of the whole discussion about it). Hence, "should not" in conjunction with "encode a version" is too soft.

> The good thing about this text is that it starts by stating a fact, and only then (after the comma) presenting a judgement. That's how we phrase run-time warnings for illegal reflective access and for dynamic installation of a Security Manager (https://openjdk.org/jeps/411#Issue-warnings). However, javac warnings need to be brief, and tend to skip directly to the judgement that X is bad.

If the compiler can be sure about its judgement, that is fine. In this case, the compiler can not really make such a judgement. It requires human intelligence to judge whether one or more digits denote a number and whether this number denotes a version:

| module name        | digit? | number? | version? | judgement |
|--------------------|--------|---------|----------|-----------|
| com.bar.f00        | ✔︎      | ✘       | ✘        | allowed   |
| com.example.pkcs11 | ✔︎      | ✔︎       | ✘        | allowed   |
| org.openjdk.java17 | ✔︎      | ✔︎       | ✔︎        | forbidden |

Hence, directly skipping to the judgement without stating the fact might be a bit too hasty. But I believe we can put both, the observation and the cause of the problem in a very concise message without dropping any information.

> [...]
> module name {0} ends with a number that looks like a version

Quite good. But for the reasons stated above, let's stick to "terminal digits":

`terminal digits in module name {0} look like a version number`

Being semantically closer to the original warning also has the side effect that e.g. existing discussions on Stack Overflow [can easier be found](https://stackoverflow.com/search?q=terminal+digits+in+module+name) if devs look for help when when encountering the updated warning.

overheadhunter:
> Ideally, there would be a sibling warning in compiler.properties for
> when there's only one terminal digit:
> 
> terminal digit in module name {0} looks like a version number

That would be smoother, but since this warning is not end-user-facing and its meaning is perfectly clear regardless of cardinality, I would argue it doesn't justify the added complexity to count the number of digits.

Would you sponsor this PR? Then I would change the message as discussed (with digits in plural form).

overheadhunter:
> Isn't this warning emitted by javac to
> potentially any Java developer?

If this was a message popping up at runtime to my grandma, I would tend to agree. However, as you said yourself: We are addressing developers, not end users. Name one single developer who hasn't at least once written some message like `%d new message(s) in your inbox`.

But if you are proposing to split the message into two and add complexity to the compiler just for nicer formatting, this would be both out of scope of the issue and also something that needs to be decided by somebody who is more involved with long term maintenance than me. Therefore please CC whoever is responsible for making such decisions.

src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties
overheadhunter: fixed in eed69a0

overheadhunter:
Isn't there anyone willing to sponsor this? It is such a trivial change that has been discussed years ago already. Ignoring it for a couple of more years doesn't paint a good picture of how contributions are handled that are seemingly too insignificant. If this gets auto-closed due to idleness, I can't promise I find motivation to look into this any longer. 🤷‍♂️

PR #14105, opened 2023-05-23
8303215: Make thread stacks not use huge pages

poonamparhar:
When a system has Transparent Huge Pages (THP) enabled (/sys/kernel/mm/transparent_hugepage/enabled is set to 'always'), thread stacks can have significantly more resident set size (RSS) than they actually require. This occurs when the stack size is 2MB or larger, which makes the memory range of the stack more likely to be aligned on a Large Page Size boundary (2MB on most systems). This in turn makes the stack eligible to be backed by transparent huge pages resulting in more memory consumption than it would otherwise when standard small pages are used. This issue is more apparent on AArch64 platforms where the default stack size is 2MB. 

```
Example mapping from smaps illustrating this issue:
fffced200000-fffced204000 ---p 00000000 00:00 0 
Size:                16 kB     # guard pages
KernelPageSize:       4 kB
MMUPageSize:          4 kB
...
fffced204000-fffced400000 rw-p 00000000 00:00 0 
Size:              2032 kB    # stack space
KernelPageSize:       4 kB
MMUPageSize:          4 kB
Rss:               2032 kB    # entire stack resident in memory
```

This fix addresses this issue with the following two main changes:

1. Change the default stack size to 2040KB, which is 2 pages less than 2MB. This ensures that stacks don't get 2MB aligned. And why 2 pages less than 2MB, because for non-JavaThreads, glibc adds an additional guard page to the total stack size. To keep it simple and to keep the default stack size value for  all options - ThreadStackSize, CompilerThreadStackSize, and VMThreadStackSize same, we use the default value as 2040K.

Example mapping for a JavaThread:
```
ffff6e913000-ffff6e917000 ---p 00000000 00:00 0 
Size:                 16 kB
KernelPageSize:        4 kB
MMUPageSize:           4 kB
...
ffff6e917000-ffff6eb11000 rw-p 00000000 00:00 0 
Size:               2024 kB
KernelPageSize:        4 kB
MMUPageSize:           4 kB
Rss:                  92 kB
```
Example Mapping for a non-JavaThread (WatcherThread):
```
ffff6eb11000-ffff6eb12000 ---p 00000000 00:00 0 
Size:                  4 kB
KernelPageSize:        4 kB
MMUPageSize:           4 kB
...
ffff6eb12000-ffff6ed10000 rw-p 00000000 00:00 0 
Size:               2040 kB
KernelPageSize:        4 kB
MMUPageSize:           4 kB
Rss:                  12 kB
```

2. If the requested stack size is greater than or equal to the default large page size (2MB on most systems) and can be large-page aligned, then add an additional page to the stack size. This reduces the chances of stacks getting large page aligned.

---------
### Progress

### Issue
 * [JDK-8303215](https://bugs.openjdk.org/browse/JDK-8303215): Make thread stacks not use huge pages

### Reviewers
 * [Thomas Stuefe](https://openjdk.org/census#stuefe) (@tstuefe - **Reviewer**) ⚠️ Review applies to [a4698247](https://git.openjdk.org/jdk/pull/14105/files/a46982470c7c3c4f7e3d12aaa2a0b9d6dfb7e8a5)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14105/head:pull/14105` \
`$ git checkout pull/14105`

`$ git checkout pull/14105` \
`$ git pull https://git.openjdk.org/jdk.git pull/14105/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14105`

`$ git pr show -t 14105`

</details>

<a href="https://git.openjdk.org/jdk/pull/14105.diff">https://git.openjdk.org/jdk/pull/14105.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14105#issuecomment-1559921231)

### DIFF
--- a/src/hotspot/os/linux/os_linux.cpp
+++ b/src/hotspot/os/linux/os_linux.cpp
@@ -929,7 +929,7 @@ bool os::create_thread(Thread* thread, ThreadType thr_type,
 
   // Add an additional page to the stack size to reduce its chances of getting large page aligned
   // so that the stack does not get backed by a transparent huge page.
-  if (stack_size >= os::Linux::_default_large_page_size && is_aligned(stack_size, os::Linux::_default_large_page_size)) {
+  if (stack_size >= os::Linux::default_large_page_size() && is_aligned(stack_size, os::Linux::default_large_page_size())) {
     stack_size += os::vm_page_size();
   }
 
@@ -3625,7 +3625,7 @@ static void set_coredump_filter(CoredumpFilterBit bit) {
 
 static size_t _large_page_size = 0;
 
-size_t os::Linux::scan_default_large_page_size() {
+static size_t scan_default_large_page_size() {
   size_t default_large_page_size = 0;
 
   // large_page_size on Linux is used to round up heap size. x86 uses either
@@ -3752,7 +3752,7 @@ bool os::Linux::setup_large_page_type(size_t page_size) {
 
 void os::large_page_init() {
   // Scan default large page size
-  size_t default_large_page_size = os::Linux::scan_default_large_page_size();
+  size_t default_large_page_size = scan_default_large_page_size();
   os::Linux::_default_large_page_size = default_large_page_size;
 
   // 1) Handle the case where we do not want to use huge pages and hence
--- a/src/hotspot/os/linux/os_linux.cpp
+++ b/src/hotspot/os/linux/os_linux.cpp
@@ -930,10 +930,10 @@ bool os::create_thread(Thread* thread, ThreadType thr_type,
   // Add an additional page to the stack size to reduce its chances of getting large page aligned
   // so that the stack does not get backed by a transparent huge page.
   size_t default_large_page_size = os::Linux::default_large_page_size();
-  if (default_large_page_size != 0) {
-    if (stack_size >= default_large_page_size && is_aligned(stack_size, default_large_page_size)) {
-      stack_size += os::vm_page_size();
-    }
+  if (default_large_page_size != 0 &&
+      stack_size >= default_large_page_size &&
+      is_aligned(stack_size, default_large_page_size)) {
+    stack_size += os::vm_page_size();
   }
 
   int status = pthread_attr_setstacksize(&attr, stack_size);
--- a/src/hotspot/os/linux/os_linux.cpp
+++ b/src/hotspot/os/linux/os_linux.cpp
@@ -927,6 +927,14 @@ bool os::create_thread(Thread* thread, ThreadType thr_type,
   }
   assert(is_aligned(stack_size, os::vm_page_size()), "stack_size not aligned");
 
+  size_t default_large_page_size = os::Linux::scan_default_large_page_size();
+
+  // Add an additional page to the stack size to reduce its chances of getting large page aligned
+  // so that the stack does not get backed by a transparent huge page.
+  if (stack_size >= default_large_page_size && is_aligned(stack_size, default_large_page_size)) {
+    stack_size += os::vm_page_size();
+  }
+
   int status = pthread_attr_setstacksize(&attr, stack_size);
   if (status != 0) {
     // pthread_attr_setstacksize() function can fail
@@ -3619,7 +3627,7 @@ static void set_coredump_filter(CoredumpFilterBit bit) {
 
 static size_t _large_page_size = 0;
 
-static size_t scan_default_large_page_size() {
+size_t os::Linux::scan_default_large_page_size() {
   size_t default_large_page_size = 0;
 
   // large_page_size on Linux is used to round up heap size. x86 uses either
@@ -3765,7 +3773,7 @@ void os::large_page_init() {
   }
 
   // 2) Scan OS info
-  size_t default_large_page_size = scan_default_large_page_size();
+  size_t default_large_page_size = os::Linux::scan_default_large_page_size();
   os::Linux::_default_large_page_size = default_large_page_size;
   if (default_large_page_size == 0) {
     // No large pages configured, return.
--- a/src/hotspot/os/linux/os_linux.cpp
+++ b/src/hotspot/os/linux/os_linux.cpp
@@ -929,8 +929,11 @@ bool os::create_thread(Thread* thread, ThreadType thr_type,
 
   // Add an additional page to the stack size to reduce its chances of getting large page aligned
   // so that the stack does not get backed by a transparent huge page.
-  if (stack_size >= os::Linux::default_large_page_size() && is_aligned(stack_size, os::Linux::default_large_page_size())) {
-    stack_size += os::vm_page_size();
+  size_t default_large_page_size = os::Linux::default_large_page_size();
+  if (default_large_page_size != 0) {
+    if (stack_size >= default_large_page_size && is_aligned(stack_size, default_large_page_size)) {
+      stack_size += os::vm_page_size();
+    }
   }
 
   int status = pthread_attr_setstacksize(&attr, stack_size);
@@ -3751,12 +3754,11 @@ bool os::Linux::setup_large_page_type(size_t page_size) {
 }
 
 void os::large_page_init() {
-  // Scan default large page size
+  // Always initialize the default large page size even if large pages are not being used.
   size_t default_large_page_size = scan_default_large_page_size();
   os::Linux::_default_large_page_size = default_large_page_size;
 
-  // 1) Handle the case where we do not want to use huge pages and hence
-  //    there is no need to scan the OS for related info
+  // 1) Handle the case where we do not want to use huge pages
   if (!UseLargePages &&
       !UseTransparentHugePages &&
       !UseHugeTLBFS &&
--- a/src/hotspot/os/linux/os_linux.cpp
+++ b/src/hotspot/os/linux/os_linux.cpp
@@ -927,11 +927,9 @@ bool os::create_thread(Thread* thread, ThreadType thr_type,
   }
   assert(is_aligned(stack_size, os::vm_page_size()), "stack_size not aligned");
 
-  size_t default_large_page_size = os::Linux::scan_default_large_page_size();
-
   // Add an additional page to the stack size to reduce its chances of getting large page aligned
   // so that the stack does not get backed by a transparent huge page.
-  if (stack_size >= default_large_page_size && is_aligned(stack_size, default_large_page_size)) {
+  if (stack_size >= os::Linux::_default_large_page_size && is_aligned(stack_size, os::Linux::_default_large_page_size)) {
     stack_size += os::vm_page_size();
   }
 
@@ -3753,6 +3751,10 @@ bool os::Linux::setup_large_page_type(size_t page_size) {
 }
 
 void os::large_page_init() {
+  // Scan default large page size
+  size_t default_large_page_size = os::Linux::scan_default_large_page_size();
+  os::Linux::_default_large_page_size = default_large_page_size;
+
   // 1) Handle the case where we do not want to use huge pages and hence
   //    there is no need to scan the OS for related info
   if (!UseLargePages &&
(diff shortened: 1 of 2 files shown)


src/hotspot/os/linux/os_linux.cpp
```diff
@@ -927,6 +927,14 @@ bool os::create_thread(Thread* thread, ThreadType thr_type,
   }
   assert(is_aligned(stack_size, os::vm_page_size()), "stack_size not aligned");
```
tstuefe, line 929: Small misgiving here about having to scan /proc/meminfo anew for every thread start. The problem is os::large_page_init() bailing out early if UseLargePages=0, right, so we cannot rely on large page initialization?

tstuefe (commented):
Yes, this is a pragmatic workaround. I believe @aph recommended the same solution.

Did you test this on an aarch64 kernel with 64k pages? I suspect the 2040 default size only works for 4k or 8k pages, for larger page sizes pthread_create would just round up the stack size again to 2MB.

poonamparhar:
> Yes, this is a pragmatic workaround. I believe @theRealAph recommended the same solution.
> 
> Did you test this on an aarch64 kernel with 64k pages? I suspect the 2040 default size only works for 4k or 8k pages, for larger page sizes pthread_create would just round up the stack size again to 2MB.

This does not solve the problem when the default small page size is 64K. The stack size gets rounded up to 2048K.

JavaThread:
```
0000ffff18ff0000     192       0       0 -----   [ anon ]
0000ffff19020000    1856     192     192 rw---   [ anon ]
0000ffff191f0000     192       0       0 -----   [ anon ]
0000ffff19220000    1856     192     192 rw---   [ anon ]
0000ffff193f0000     192       0       0 -----   [ anon ]
0000ffff19420000    1856     192     192 rw---   [ anon ]
```
non-JavaThread:
```
0000ffff6c170000      64       0       0 -----   [ anon ]
0000ffff6c180000    2048      64      64 rw---   [ anon ]
0000ffff6c380000      64       0       0 -----   [ anon ]
0000ffff6c390000    2048      64      64 rw---   [ anon ]
```

For 64k page size, changing the stack size to 1920K (2048 - 2*64) would work:
```
0000fffdc5ba0000     192       0       0 -----   [ anon ]
0000fffdc5bd0000    1728     192     192 rw---   [ anon ]
0000fffdc5d80000     192       0       0 -----   [ anon ]
0000fffdc5db0000    1728     192     192 rw---   [ anon ]
0000fffdc5f60000     192       0       0 -----   [ anon ]
0000fffdc5f90000    1728     192     192 rw---   [ anon ]
```

poonamparhar:
For the 64K page size, the default huge page size is usually 512M. The aarch64 machine with 64K page size that I am testing on has:

```
bash-4.4$ cat /proc/meminfo | grep Hugepagesize
Hugepagesize:     524288 kB
```

And on that, the current set of changes behave correctly when I run with -Xss512m. Snippet of pmap output:

```
0000fff8b7ff0000     192       0       0 -----   [ anon ]        # guard pages
0000fff8b8020000  524160     192     192 rw---   [ anon ]  # usabel stack area
0000fff8d8000000   65472   65472   65472 rw---   [ anon ]
0000fff8dbff0000      64       0       0 -----   [ anon ]
0000fff8dfff0000     192       0       0 -----   [ anon ]
0000fff8e0020000  524160     192     192 rw---   [ anon ]
```
an additional 64k page gets added to the stack size to disturb the large page alignment.

iklam:
Does this PR affect only Linux/aarch64? If so, I think it's better to add that into the bug title.

tstuefe:
> Does this PR affect only Linux/aarch64? If so, I think it's better to add that into the bug title.

It affects all platforms, reproducible if you set stack size to 2M and THP to always. Aarch64 has 2M as default though.

src/hotspot/os/linux/os_linux.cpp
poonamparhar: Hi @tstuefe, I made this change. Please take a look. Thanks!

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -927,6 +927,12 @@ bool os::create_thread(Thread* thread, ThreadType thr_type,
...
 
+  // Add an additional page to the stack size to reduce its chances of getting large page aligned
+  // so that the stack does not get backed by a transparent huge page.
+  if (stack_size >= os::Linux::_default_large_page_size && is_aligned(stack_size, os::Linux::_default_large_page_size)) {
```
tstuefe, line 932: Could you use `os::Linux::default_large_page_size()` instead of accessing the private member directly?

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -3619,7 +3625,7 @@ static void set_coredump_filter(CoredumpFilterBit bit) {
 
 static size_t _large_page_size = 0;
 
-static size_t scan_default_large_page_size() {
+size_t os::Linux::scan_default_large_page_size() {
```
tstuefe, line 3628: I think this is not needed anymore, or? You can revert this back to a local static function, and call that in `os::large_page_init()`

src/hotspot/os_cpu/linux_aarch64/globals_linux_aarch64.hpp
```diff
@@ -30,10 +30,10 @@
 // (see globals.hpp)
 
 define_pd_global(bool, DontYieldALot,            false);
-define_pd_global(intx, ThreadStackSize,          2048); // 0 => use system default
-define_pd_global(intx, VMThreadStackSize,        2048);
+define_pd_global(intx, ThreadStackSize,          2040); // 0 => use system default
```
tstuefe, line 33: A comment would be good here to explain the odd size

tstuefe (changes requested):
Some small nits remain.

src/hotspot/os/linux/os_linux.cpp
poonamparhar: Resolved this.

src/hotspot/os/linux/os_linux.cpp
poonamparhar: Fixed this.

tstuefe (approved):
Looks good. Thanks for fixing.

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -3745,6 +3751,10 @@ bool os::Linux::setup_large_page_type(size_t page_size) {
 }
 
 void os::large_page_init() {
+  // Scan default large page size
+  size_t default_large_page_size = scan_default_large_page_size();
+  os::Linux::_default_large_page_size = default_large_page_size;
+
   // 1) Handle the case where we do not want to use huge pages and hence
   //    there is no need to scan the OS for related info
```
dholmes-ora, line 3759: The comment needs updating as now you do the scan first.

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -927,6 +927,12 @@ bool os::create_thread(Thread* thread, ThreadType thr_type,
   }
   assert(is_aligned(stack_size, os::vm_page_size()), "stack_size not aligned");
 
+  // Add an additional page to the stack size to reduce its chances of getting large page aligned
+  // so that the stack does not get backed by a transparent huge page.
+  if (stack_size >= os::Linux::default_large_page_size() && is_aligned(stack_size, os::Linux::default_large_page_size())) {
+    stack_size += os::vm_page_size();
+  }
```
dholmes-ora, line 934: Do we need to watch for a zero default size here so that we don't expand all stacks by a page?

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -3745,6 +3751,10 @@ bool os::Linux::setup_large_page_type(size_t page_size) {
 }
 
 void os::large_page_init() {
+  // Scan default large page size
```
dholmes-ora, line 3754: Suggestion:
```
// Always initialize the default large page size even if large pages are not being used.
```

src/hotspot/os_cpu/linux_aarch64/globals_linux_aarch64.hpp
```diff
@@ -30,10 +30,16 @@
 // (see globals.hpp)
 
 define_pd_global(bool, DontYieldALot,            false);
-define_pd_global(intx, ThreadStackSize,          2048); // 0 => use system default
-define_pd_global(intx, VMThreadStackSize,        2048);
 
-define_pd_global(intx, CompilerThreadStackSize,  2048);
+// set default stack sizes < 2MB so as to prevent stacks from getting
```
dholmes-ora, line 34: /set/Set/

src/hotspot/os_cpu/linux_aarch64/globals_linux_aarch64.hpp
```diff
@@ -30,10 +30,16 @@
...
+// large-page aligned and backed by THPs on systems where 2MB is the
+// default huge page size. For non-JavaThreads, glibc adds an additional
```
dholmes-ora, line 36: s/adds/may add/

src/hotspot/os_cpu/linux_aarch64/globals_linux_aarch64.hpp
```diff
@@ -30,10 +30,16 @@
...
+// guard page to the total stack size, so to keep the default sizes same
+// for all the following flags, we set them to 2 pages less than 2MB.
```
dholmes-ora, line 38: I still think we need to at least mention we assume 4K page here.

dholmes-ora (changes requested):

src/hotspot/os_cpu/linux_aarch64/globals_linux_aarch64.hpp
poonamparhar: @dholmes-ora I have updated the comment, please review.

src/hotspot/os/linux/os_linux.cpp
poonamparhar: Added a check for zero default page size. Please review.

src/hotspot/os/linux/os_linux.cpp
poonamparhar: Changed the comment.

src/hotspot/os_cpu/linux_aarch64/globals_linux_aarch64.hpp
poonamparhar: Done.

src/hotspot/os_cpu/linux_aarch64/globals_linux_aarch64.hpp
poonamparhar: Done.

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -929,8 +929,11 @@ bool os::create_thread(Thread* thread, ThreadType thr_type,
 
   // Add an additional page to the stack size to reduce its chances of getting large page aligned
   // so that the stack does not get backed by a transparent huge page.
-  if (stack_size >= os::Linux::default_large_page_size() && is_aligned(stack_size, os::Linux::default_large_page_size())) {
-    stack_size += os::vm_page_size();
+  size_t default_large_page_size = os::Linux::default_large_page_size();
+  if (default_large_page_size != 0) {
+    if (stack_size >= default_large_page_size && is_aligned(stack_size, default_large_page_size)) {
```
dholmes-ora, line 934: Nit: this could have been written as a single if-clause

dholmes-ora (approved):
One minor nit but otherwise nothing further from me.

Thanks.

src/hotspot/os/linux/os_linux.cpp
poonamparhar: Changed it. Thanks!

dholmes-ora (approved):

poonamparhar:
@tstuefe Hi Thomas, I have addressed the review comments from David Holmes. Could you please take a look at the updated changes again? Thanks!

tstuefe:
> @tstuefe Hi Thomas, I have addressed the review comments from David Holmes. Could you please take a look at the updated changes again? Thanks!

Still good. Thanks.

PR #14107, opened 2023-05-23
8299665: /proc/self/stat parsing in libmanagement broken by execname with spaces

kevinjwalls:
Code that parses /proc/self/stat may be broken if the executable name contains spaces.

We have code that does this correctly, and code that does not.
UnixOperatingSystem.c is Linux-specific OS code, and has vread_statdata which correctly uses strrchr to find the last ) and parses from there.

OperatingSystemImpl.c the generic, not Linux-specific code, is not OK, as Java_com_sun_management_internal_OperatingSystemImpl_getCommittedVirtualMemorySize0 has a problematic fscanf call. 
But the problematic fscanf call is Linux-specific: we use ifdef to create OS-specific implementations.

Need to change the fscanf of /proc/self/stat to do the same buffered read as read_ticks in UnixOperatingSystem.c (the Linux-specific code), with strrchr to read past the ) of the execname.

Moving the actually-Linux-specific getCommittedVirtualMemorySize0 to be with the other Linux-specific code means it can just use the existing method for accessing /proc/self/stat.

It does however call throw_internal_error, which is then not accessible.  So that function can be shared in the libmanagement_ext.h/.c for both uses.

The call to throw_internal_error is working: if I change the code to fail, it reports a java.lang.InternalError as it would before.

Existing tests still pass: 
i.e. test/jdk/com/sun/management/OperatingSystemMXBean/GetCommittedVirtualMemorySize.java

---------
### Progress

### Issue
 * [JDK-8299665](https://bugs.openjdk.org/browse/JDK-8299665): /proc/self/stat parsing in libmanagement broken by execname with spaces

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14107/head:pull/14107` \
`$ git checkout pull/14107`

`$ git checkout pull/14107` \
`$ git pull https://git.openjdk.org/jdk.git pull/14107/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14107`

`$ git pr show -t 14107`

</details>

<a href="https://git.openjdk.org/jdk/pull/14107.diff">https://git.openjdk.org/jdk/pull/14107.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14107#issuecomment-1560941244)

### DIFF
--- a/src/jdk.management/linux/native/libmanagement_ext/UnixOperatingSystem.c
+++ b/src/jdk.management/linux/native/libmanagement_ext/UnixOperatingSystem.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -22,6 +22,8 @@
  * or visit www.oracle.com if you need additional information or have any
  * questions.
  */
+
+#include "jni_util.h"
 
 #include <stdio.h>
 #include <stdint.h>
@@ -70,6 +72,13 @@ static int next_line(FILE *f) {
     } while (c != '\n' && c != EOF);
 
     return c;
+}
+
+static void throw_internal_error(JNIEnv* env, const char* msg) {
+    char errmsg[128];
+
+    snprintf(errmsg, sizeof(errmsg), "errno: %d error: %s\n", errno, msg);
+    JNU_ThrowInternalError(env, errmsg);
 }
 
 /**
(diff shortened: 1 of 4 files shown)


sspitsyn (commented):
How do you test the issue is fixed now?
Is there any test case which fails without your fix and passes with it?

kevinjwalls:
> How do you test the issue is fixed now? Is there any test case which fails without your fix and passes with it?

Hi - I tested manually, renaming the java executable to something with a space, then query the OperatingSystem MBean, and see something like: CommittedVirtualMemorySize = (java.lang.Long) 36202319872 (which is the same/similar to another run 
when not renamed).  The problem is seeing a completely wrong value, like: CommittedVirtualMemorySize = (java.lang.Long) 0 

These are lines like:

$ cat /proc/9496/stat                                                                 
9496 (java) S 625 9495 607 34823 9635 1077936192 12024 0 0 0 156 22 0 0 20 0 28 0 2172294643 36260237312 30463 18446744073709551615 93928094162944 93928094166632 140724289571696 0 0 0 4 0 16800975 0 0 0 -1 11 0 0 0 0 0 93928094174456 93928094175248 93928124657664 140724289577708 140724289577894 140724289577894 140724289581032 0

That's a different run, where 36260237312 was the value we wanted.  If "java" becomes "this is java"  (with 2 spaces), then we read field 21, not field 23 from the line of data and show zero.  

It's quite a niche thing to do so I didn't add a test that copied the java binary under test to a new name with a space in it...
The code that does the strrchr to scan from the last ) and then parses the remainder has been with us for some time. 8-)

sspitsyn:
Thank you for explanation! Looks good.

sspitsyn (approved):

alexmenkov (approved):

kevinjwalls:
Thanks @sspitsyn and @alexmenkov for the reviews!

PR #10933, opened 2022-11-01
8295724: VirtualMachineError: Out of space in CodeCache for method handle intrinsic

TheRealMDoerr:
This proposal prevents the VM from terminating unexpectedly in some rare cases (see JBS issue). It allows using NonNMethod code space for method handle intrinsics which are needed urgently if the other code cache spaces are full. There are other options (see JBS issue), but this one appears to be the simplest one.

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8295724](https://bugs.openjdk.org/browse/JDK-8295724): VirtualMachineError: Out of space in CodeCache for method handle intrinsic

### Reviewers
 * [Vladimir Kozlov](https://openjdk.org/census#kvn) (@vnkozlov - **Reviewer**)
 * [Dean Long](https://openjdk.org/census#dlong) (@dean-long - **Reviewer**) ⚠️ Review applies to [2c5d2839](https://git.openjdk.org/jdk/pull/10933/files/2c5d2839038bc2cb2574d5eda580dc82ee7e8403)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk pull/10933/head:pull/10933` \
`$ git checkout pull/10933`

Update a local copy of the PR: \
`$ git checkout pull/10933` \
`$ git pull https://git.openjdk.org/jdk pull/10933/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 10933`

View PR using the GUI difftool: \
`$ git pr show -t 10933`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jdk/pull/10933.diff">https://git.openjdk.org/jdk/pull/10933.diff</a>

</details>

PR #14108, opened 2023-05-23
8302351: "assert(!JavaThread::current()->is_interp_only_mode() || !nm->method()->is_continuation_enter_intrinsic() || ContinuationEntry::is_interpreted_call(return_pc)) failed: interp_only_mode but not in enterSpecial interpreted entry" in fixup_callers_callsite

pchilano:
Please review the following fix. Runtime methods called through the SharedRuntime::generate_resolve_blob() stub always return the value stored in _from_compiled_entry as the entry point to the callee method. This will either be the entry point to the compiled version of callee if there is one or the c2i adapter entry point. But this doesn't consider the case where an EnterInterpOnlyModeClosure handshake catches the JavaThread in the transition back to Java on those methods. In that case we should return the c2i adapter entry point even if there is a compiled entry point. Otherwise the JavaThread will continue calling the compiled versions of methods without noticing it's in interpreted only mode until it either calls a method that hasn't been compiled yet or it returns to the caller of that resolved callee where the change to interpreter only mode happened (since the EnterInterpOnlyModeClosure handshake marked all the frames on the stack for deoptimization). 

This is a long standing bug but has been made visible with the assert added as part of 8288949 where a related issue was fixed. There are more details in the bug comments about how this specific crash happens and its relation with 8288949. I also attached a reproducer.

These runtime methods are already using JRT_BLOCK_ENTRY/JRT_BLOCK so that the entry point to the callee is fetched only after the last possible safepoint in JRT_BLOCK_END. This guarantees that we will not return an entry point to compiled code that has already been removed. So the fix just adds a check to verify if the JavaThread entered into interpreted only mode in that transition back to Java and to return the c2i entry point instead.

I tested the patch in mach5 tiers 1-6. I also verified it with the reproducer I attached to the bug. I didn't include it as an additional test but I can do that if desired.

Thanks,
Patricio

---------
### Progress

### Issue
 * [JDK-8302351](https://bugs.openjdk.org/browse/JDK-8302351): "assert(!JavaThread::current()-&gt;is_interp_only_mode() || !nm-&gt;method()-&gt;is_continuation_enter_intrinsic() || ContinuationEntry::is_interpreted_call(return_pc)) failed: interp_only_mode but not in enterSpecial interpreted entry" in fixup_callers_callsite (**Bug** - P3)

### Reviewers
 * [Daniel D. Daugherty](https://openjdk.org/census#dcubed) (@dcubed-ojdk - **Reviewer**) ⚠️ Review applies to [54dee960](https://git.openjdk.org/jdk/pull/14108/files/54dee960e1ce8f23bb3c722e20369bc1c5fb5cdf)
 * [Dean Long](https://openjdk.org/census#dlong) (@dean-long - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14108/head:pull/14108` \
`$ git checkout pull/14108`

`$ git checkout pull/14108` \
`$ git pull https://git.openjdk.org/jdk.git pull/14108/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14108`

`$ git pr show -t 14108`

</details>

<a href="https://git.openjdk.org/jdk/pull/14108.diff">https://git.openjdk.org/jdk/pull/14108.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14108#issuecomment-1560204509)

### DIFF
--- a/src/hotspot/share/runtime/sharedRuntime.cpp
+++ b/src/hotspot/share/runtime/sharedRuntime.cpp
@@ -1500,7 +1500,9 @@ JRT_BLOCK_ENTRY(address, SharedRuntime::handle_wrong_method_ic_miss(JavaThread* 
     // Return Method* through TLS
     current->set_vm_result_2(callee_method());
   JRT_BLOCK_END
-  return callee_method->from_compiled_entry(current->is_interp_only_mode());
+  // return compiled code entry point after potential safepoints
+  assert(callee_method->verified_code_entry() != nullptr, " Jump to zero!");
+  return callee_method->verified_code_entry();
 JRT_END
 
 
@@ -1552,7 +1554,9 @@ JRT_BLOCK_ENTRY(address, SharedRuntime::handle_wrong_method(JavaThread* current)
     callee_method = SharedRuntime::reresolve_call_site(CHECK_NULL);
     current->set_vm_result_2(callee_method());
   JRT_BLOCK_END
-  return callee_method->from_compiled_entry(current->is_interp_only_mode());
+  // return compiled code entry point after potential safepoints
+  assert(callee_method->verified_code_entry() != nullptr, " Jump to zero!");
+  return callee_method->verified_code_entry();
 JRT_END
 
 // Handle abstract method call
@@ -1593,11 +1597,38 @@ JRT_END
 // resolve a static call and patch code
 JRT_BLOCK_ENTRY(address, SharedRuntime::resolve_static_call_C(JavaThread* current ))
   methodHandle callee_method;
+  bool enter_special = false;
   JRT_BLOCK
     callee_method = SharedRuntime::resolve_helper(false, false, CHECK_NULL);
     current->set_vm_result_2(callee_method());
+
+    if (current->is_interp_only_mode()) {
+      RegisterMap reg_map(current,
+                          RegisterMap::UpdateMap::skip,
+                          RegisterMap::ProcessFrames::include,
+                          RegisterMap::WalkContinuation::skip);
+      frame stub_frame = current->last_frame();
+      assert(stub_frame.is_runtime_frame(), "must be a runtimeStub");
+      frame caller = stub_frame.sender(&reg_map);
+      enter_special = caller.cb() != nullptr && caller.cb()->is_compiled()
+        && caller.cb()->as_compiled_method()->method()->is_continuation_enter_intrinsic();
+    }
   JRT_BLOCK_END
-  return callee_method->from_compiled_entry(current->is_interp_only_mode());
+
+  if (current->is_interp_only_mode() && enter_special) {
+    // enterSpecial is compiled and calls this method to resolve the call to Continuation::enter
+    // but in interp_only_mode we need to go to the interpreted entry
+    // The c2i won't patch in this mode -- see fixup_callers_callsite
+    //
+    // This should probably be done in all cases, not just enterSpecial (see JDK-8218403),
+    // but that's part of a larger fix, and the situation is worse for enterSpecial, as it has no
+    // interpreted version.
+    return callee_method->get_c2i_entry();
+  }
+
+  // return compiled code entry point after potential safepoints
+  assert(callee_method->verified_code_entry() != nullptr, " Jump to zero!");
+  return callee_method->verified_code_entry();
 JRT_END
 
 
@@ -1608,7 +1639,9 @@ JRT_BLOCK_ENTRY(address, SharedRuntime::resolve_virtual_call_C(JavaThread* curre
     callee_method = SharedRuntime::resolve_helper(true, false, CHECK_NULL);
     current->set_vm_result_2(callee_method());
   JRT_BLOCK_END
-  return callee_method->from_compiled_entry(current->is_interp_only_mode());
+  // return compiled code entry point after potential safepoints
+  assert(callee_method->verified_code_entry() != nullptr, " Jump to zero!");
+  return callee_method->verified_code_entry();
 JRT_END
 
 
@@ -1620,7 +1653,9 @@ JRT_BLOCK_ENTRY(address, SharedRuntime::resolve_opt_virtual_call_C(JavaThread* c
     callee_method = SharedRuntime::resolve_helper(true, true, CHECK_NULL);
     current->set_vm_result_2(callee_method());
   JRT_BLOCK_END
-  return callee_method->from_compiled_entry(current->is_interp_only_mode());
+  // return compiled code entry point after potential safepoints
+  assert(callee_method->verified_code_entry() != nullptr, " Jump to zero!");
+  return callee_method->verified_code_entry();
 JRT_END
 
 // The handle_ic_miss_helper_internal function returns false if it failed due
@@ -2035,9 +2070,6 @@ JRT_LEAF(void, SharedRuntime::fixup_callers_callsite(Method* method, address cal
   // Get the return PC for the passed caller PC.
   address return_pc = caller_pc + frame::pc_return_offset;
 
-  assert(!JavaThread::current()->is_interp_only_mode() || !nm->method()->is_continuation_enter_intrinsic()
-    || ContinuationEntry::is_interpreted_call(return_pc), "interp_only_mode but not in enterSpecial interpreted entry");
-
   // There is a benign race here. We could be attempting to patch to a compiled
   // entry point at the same time the callee is being deoptimized. If that is
   // the case then entry_point may in fact point to a c2i and we'd patch the
@@ -2075,8 +2107,6 @@ JRT_LEAF(void, SharedRuntime::fixup_callers_callsite(Method* method, address cal
         return;
       }
       if (nm->method()->is_continuation_enter_intrinsic()) {
-        assert(ContinuationEntry::is_interpreted_call(call->instruction_address()) == JavaThread::current()->is_interp_only_mode(),
-          "mode: %d", JavaThread::current()->is_interp_only_mode());
         if (ContinuationEntry::is_interpreted_call(call->instruction_address())) {
           return;
         }
--- a/src/hotspot/share/runtime/sharedRuntime.cpp
+++ b/src/hotspot/share/runtime/sharedRuntime.cpp
@@ -1500,9 +1500,11 @@ JRT_BLOCK_ENTRY(address, SharedRuntime::handle_wrong_method_ic_miss(JavaThread* 
     // Return Method* through TLS
     current->set_vm_result_2(callee_method());
   JRT_BLOCK_END
-  // return compiled code entry point after potential safepoints
-  assert(callee_method->verified_code_entry() != nullptr, " Jump to zero!");
-  return callee_method->verified_code_entry();
+  // Return compiled code entry point after potential safepoints.
+  // In case we are in interpreted only mode return c2i entry instead.
+  address target = current->is_interp_only_mode() ? callee_method->get_c2i_entry() : callee_method->verified_code_entry();
+  assert(target != nullptr, "Jump to zero!");
+  return target;
 JRT_END
 
 
@@ -1554,9 +1556,11 @@ JRT_BLOCK_ENTRY(address, SharedRuntime::handle_wrong_method(JavaThread* current)
     callee_method = SharedRuntime::reresolve_call_site(CHECK_NULL);
     current->set_vm_result_2(callee_method());
   JRT_BLOCK_END
-  // return compiled code entry point after potential safepoints
-  assert(callee_method->verified_code_entry() != nullptr, " Jump to zero!");
-  return callee_method->verified_code_entry();
+  // Return compiled code entry point after potential safepoints.
+  // In case we are in interpreted only mode return c2i entry instead.
+  address target = current->is_interp_only_mode() ? callee_method->get_c2i_entry() : callee_method->verified_code_entry();
+  assert(target != nullptr, "Jump to zero!");
+  return target;
 JRT_END
 
 // Handle abstract method call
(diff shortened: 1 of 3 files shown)


src/hotspot/share/runtime/sharedRuntime.cpp
```diff
@@ -1500,9 +1500,11 @@ JRT_BLOCK_ENTRY(address, SharedRuntime::handle_wrong_method_ic_miss(JavaThread*
...
     current->set_vm_result_2(callee_method());
   JRT_BLOCK_END
-  // return compiled code entry point after potential safepoints
-  assert(callee_method->verified_code_entry() != nullptr, " Jump to zero!");
-  return callee_method->verified_code_entry();
+  // Return compiled code entry point after potential safepoints.
+  // In case we are in interpreted only mode return c2i entry instead.
+  address target = current->is_interp_only_mode() ? callee_method->get_c2i_entry() : callee_method->verified_code_entry();
+  assert(target != nullptr, "Jump to zero!");
+  return target;
```
dcubed-ojdk, line 1507: There are now 5 copies of this exact code block including this one.

Seems like this is screaming to be refactored into some static function
with an appropriate name that can be called from all five places.

src/hotspot/share/runtime/sharedRuntime.cpp
```diff
@@ -1597,38 +1601,15 @@ JRT_END
...
-  if (current->is_interp_only_mode() && enter_special) {
-    // enterSpecial is compiled and calls this method to resolve the call to Continuation::enter
-    // but in interp_only_mode we need to go to the interpreted entry
-    // The c2i won't patch in this mode -- see fixup_callers_callsite
-    //
-    // This should probably be done in all cases, not just enterSpecial (see JDK-8218403),
-    // but that's part of a larger fix, and the situation is worse for enterSpecial, as it has no
-    // interpreted version.
-    return callee_method->get_c2i_entry();
-  }
```
dcubed-ojdk, line 1627: If I understand the analysis correctly, all this logic that we have to clearly
identify the calling context as ` jdk.internal.vm.Continuation.enterSpecial`
is no longer needed because we're switching to a more general fix of:
```
    if (current->is_interp_only_mode()) {
      target = callee_method->get_c2i_entry();
    } else {
      target = callee_method->verified_code_entry();
    }
    return target;
```
And that more general fix cover the special case code that we're deleting
(pun intended).

dcubed-ojdk (approved):
Thumbs up on the fix itself. I do think it could use some refactoring
to avoid the five copies of the same code block.

Thanks for documenting your testing. So far there have been 15 sightings
of this test failure in the JDK21 CI: 13 in tier4, 2 in tier7 and 1 in tier8. Since
you are touching 5 different SharedRuntime functions, I think you should
run this fix thru at least one Tier7 and one Tier8 job set.

pchilano:
> Thumbs up on the fix itself. I do think it could use some refactoring to avoid the five copies of the same code block.
> 
> Thanks for documenting your testing. So far there have been 15 sightings of this test failure in the JDK21 CI: 13 in tier4, 2 in tier7 and 1 in tier8. Since you are touching 5 different SharedRuntime functions, I think you should run this fix thru at least one Tier7 and one Tier8 job set.
>
Sounds good. I already lunched a tier7 job. I'll run tier8 after that.

src/hotspot/share/runtime/sharedRuntime.cpp
pchilano: I considered that initially but I preferred the readability part given that we are not saving much. But I can do that if you think it's better that way.

src/hotspot/share/runtime/sharedRuntime.cpp
pchilano: Exactly. This code was added to fix a related issue, where a JavaThread already noticed it was in interpreted only mode but because enterSpecial doesn't have an actual interpreted version we switched from interpreted to compiled and from there we incorrectly called the compiled version of enter() leading to asserts in JVMTI code about stack depth. This fix is more general so it already covers that case too.

pchilano:
Thanks for the review Dan! I'll update the PR with the extra testing once its done.

dean-long:
Do we ever call these resolve stubs on method handle intrinsics (if we could force the compilers not to inline them)?  I'm wondering what happens when we ask for the c2i adapter for intrinsics like _linkToNative, and also if these intrinsics do the right thing with is_interp_only_mode().  The fact that these intrinsics only have a compiled version came up before (JDK-8296336 and https://github.com/openjdk/jdk/pull/10933).  These intrinsics don't push a frame, but when they call the target do they check for is_interp_only_mode() and call the interpreted version instead of the compiled version?

pchilano:
> Do we ever call these resolve stubs on method handle intrinsics (if we could force the compilers not to inline them)? I'm wondering what happens when we ask for the c2i adapter for intrinsics like _linkToNative, and also if these intrinsics do the right thing with is_interp_only_mode(). The fact that these intrinsics only have a compiled version came up before (JDK-8296336 and #10933). These intrinsics don't push a frame, but when they call the target do they check for is_interp_only_mode() and call the interpreted version instead of the compiled version?
>
I don't know how method handle intrinsics work, so not sure if they use this stubs. If they do and there is no c2i adapter then this approach won't work, even if we check and ignore that case returning the normal compiled entry. Because we will remain in compile code and the issue is the same. Interpreted only mode is not checked while in compiled code because the code assumes that won't happen. The switch to interpreted only mode happens in a handshake, and we deoptimize all frames, so we should switch to interpreted mode when going back to Java after the handshake either from vm or native. But this case is special because we are returning to Java but not back to the deopted caller, we are returning to the runtime stub and them jumping to the callee entry.

src/hotspot/share/runtime/sharedRuntime.cpp
dcubed-ojdk: I'm not totally sold on the refactoring being a better idea. I would be okay if you
decided not to do that. Let's see if any other reviewer has an opinion one way or
the other.

dcubed-ojdk:
@matias9927 has been working with invoke dynamic data structures recently
and might be able to help shed some light here...

pchilano:
@dean-long I added extra asserts in all these methods just for testing and these stubs are used to resolve method handle intrinsics too, including _linkToNative which I found from tests in test/jdk/java/foreign/. For the _linkToNative case which doesn't have an interpreted version if we return the c2i adapter, since the _i2i_entry is set to the i2c adapter [1] we will actually crash in the i2c because we cannot go from c2i -> i2c. Now if this is the only such case I guess we could special case it, assuming also the carrier's path to enterSpecial will never need to resolve that method. But I'm not convinced on that approach. Otherwise the other option would be to keep the current behavior and just remove the assert added in 8288949 accepting that it's too strong, and keep this issue under the existing 8218403 bug. What do you think?

[1] https://github.com/openjdk/jdk/blob/master/src/hotspot/share/oops/method.cpp#L1488

src/hotspot/share/runtime/sharedRuntime.cpp
theRealAph: I wonder if something like

`callee_method->from_compiled_entry(current->is_interp_only_mode())`

might be readable and well factored.

pchilano: Yes, that works. I can do that but I'll wait since we might not go with this solution after all.

dean-long:
> @dean-long I added extra asserts in all these methods just for testing and these stubs are used to resolve method handle intrinsics too, including _linkToNative which I found from tests in test/jdk/java/foreign/. For the _linkToNative case which doesn't have an interpreted version if we return the c2i adapter, since the _i2i_entry is set to the i2c adapter [1] we will actually crash in the i2c because we cannot go from c2i -> i2c. Now if this is the only such case I guess we could special case it, assuming also the carrier's path to enterSpecial will never need to resolve that method. But I'm not convinced on that approach. Otherwise the other option would be to keep the current behavior and just remove the assert added in 8288949 accepting that it's too strong, and keep this issue under the existing 8218403 bug. What do you think?
> 
> [1] https://github.com/openjdk/jdk/blob/master/src/hotspot/share/oops/method.cpp#L1488

Sorry, are you saying we will crash in the i2c with your current patch, or only if you add extra asserts?
Which assert in 8288949 is too strong?
I'm OK with not leaving 8218403 issues out of this patch, but I'm not OK with it causing crashes :-)

src/hotspot/share/runtime/sharedRuntime.cpp
dean-long: > Yes, that works. I can do that but I'll wait since we might not go with this solution after all.

Is there an alternative solution you have in mind?

pchilano:
> > @dean-long I added extra asserts in all these methods just for testing and these stubs are used to resolve method handle intrinsics too, including _linkToNative which I found from tests in test/jdk/java/foreign/. For the _linkToNative case which doesn't have an interpreted version if we return the c2i adapter, since the _i2i_entry is set to the i2c adapter [1] we will actually crash in the i2c because we cannot go from c2i -> i2c. Now if this is the only such case I guess we could special case it, assuming also the carrier's path to enterSpecial will never need to resolve that method. But I'm not convinced on that approach. Otherwise the other option would be to keep the current behavior and just remove the assert added in 8288949 accepting that it's too strong, and keep this issue under the existing 8218403 bug. What do you think?
> > [1] https://github.com/openjdk/jdk/blob/master/src/hotspot/share/oops/method.cpp#L1488
> 
> Sorry, are you saying we will crash in the i2c with your current patch, or only if you add extra asserts? 
>
With the current patch. If a change to interpreter only mode happens while we are resolving _linkToNative, we will return the c2i entry. That will eventually call the i2c so we will crash right away (in debug mode at least). It's just that there is no current test where that scenario will happen. As a quick test if you change resolve_static_call_C() to always return the c2i entry point to simulate the change to interpreter mode and run test/jdk/java/foreign/ you will see it.

>Which assert in 8288949 is too strong? I'm OK with not fixing 8218403 issues in this patch, but I'm not OK with it causing crashes :-)
>
The one we are hitting in fixup_callers_callsite() (subject of PR).

src/hotspot/share/runtime/sharedRuntime.cpp
pchilano: So we can continue with this solution and just special case _linkToNative to always return the verified_code_entry(). The other option would be to just remove the assert in fixup_callers_callsite() added in 8288949 to fix this crash. It doesn't solve the underlying issue where after a change to interpreter only mode we still continue in compiled code, but this is not specific to Continuation.enterSpecial and Continuation.enter as the assert reads. With any caller-callee we face the same issue. This is a long standing bug that is described in 8218403 as you have filed already. 
Now if we continue with this solution the question would be, is _linkToNative the only method that doesn't have an interpreted version? Or are there more special cases to cover?

dean-long:
This patch seems like a step in the right direction, so let's keep the assert and special-case _linkToNative.  It is the only special case I know of.

pchilano:
> This patch seems like a step in the right direction, so let's keep the assert and special-case _linkToNative. It is the only special case I know of.
>
I have updated the PR with the suggested refactoring. Now, I have stress tested the patch by simulating always switching to interpreter only mode when going through this resolve methods in SharedRuntime and I found there is also an issue with the other method handle intrinsics, besides _linkToNative. If we return the c2i entry when resolving any of this methods we will end up jumping to the code generated by generate_method_handle_interpreter_entry(), but that code has an implicit null check for the MH receiver in generate_method_handle_dispatch(). If that is indeed null the signal handler will redirect us to generate_throw_exception() where we will call SharedRuntime::throw_NullPointerException_at_call() to create the exception. But the stack has already been changed in the c2i adapter to pass the arguments to the callee and also in that MH interpreter entry [1]. So when walking the stack we will crash when getting the sender of this compiled method that went through the c2i + MH interpreter entry. I found this issue with test compiler/jsr292/NullConstantReceiver.java. The bottom line issue is that we jumped to the interpreter target from the c2i entry but we never actually created an interpreter frame because we throw an exception before that, and since we modified the stack already we cannot walk it anymore. 

Today we never hit this issue because it seems the _from_compiled_entry for method handle intrinsics never points to the c2i, but always to the compiled entry generated by gen_special_dispatch() (there has been an interesting discussion about this in 8302320). One way to keep this behavior and still catch the switch to interpreted only mode when resolving this methods could be to still return the _from_compiled_entry for method handle intrinsics, but modify jump_from_method_handle() to check for interpreted only mode in both cases, not just when coming from interpreted code [2]. So if we came from compiled code and we are in interpreted only mode jump to the c2i instead. I have tested that and seems to work but I'm open to suggestions on this. I'm going away on vacations for a few days though so I'll resume work on this once I'm back.

[1] https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/x86/methodHandles_x86.cpp#L311
[2] https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/x86/methodHandles_x86.cpp#L138

sspitsyn (commented):
The fix looks pretty good to me.
I have one question on the comment:
```
1288 // TODO: _linkToNative doesn't have an interpreted version so we always
1289 // return the compiled code entry point.
```
I guess there is already a bug filed on this, is not it?

sspitsyn (approved):
Just removing the asserts looks okay to me.
Also, it will be safe to backport to 21.
Thanks,
Serguei

pchilano:
> Just removing the asserts looks okay to me. Also, it will be safe to backport to 21. Thanks, Serguei
>
Thanks for the review Serguei!

dean-long (approved):

pchilano:
Thanks for the review Dean!

PR #14109, opened 2023-05-23
8290499: new File(parent, "/") breaks normalization – creates File with slash at the end

bplb:
In `java.io.File`, change the constructors `File(File,String)` and `File(String,String)` so that they do not for typical cases return a `File` whose path has a trailing name separator.

---------
### Progress

### Issue
 * [JDK-8290499](https://bugs.openjdk.org/browse/JDK-8290499): new File(parent, "/") breaks normalization – creates File with slash at the end

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14109/head:pull/14109` \
`$ git checkout pull/14109`

`$ git checkout pull/14109` \
`$ git pull https://git.openjdk.org/jdk.git pull/14109/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14109`

`$ git pr show -t 14109`

</details>

<a href="https://git.openjdk.org/jdk/pull/14109.diff">https://git.openjdk.org/jdk/pull/14109.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14109#issuecomment-1560232453)

### DIFF
--- a/src/java.base/unix/classes/java/io/UnixFileSystem.java
+++ b/src/java.base/unix/classes/java/io/UnixFileSystem.java
@@ -101,15 +101,23 @@ final class UnixFileSystem extends FileSystem {
         return pathname.startsWith("/") ? 1 : 0;
     }
 
+    // Remove trailing '/' if present and there are at least two characters
+    private static String trimSeparator(String s) {
+        int len = s.length();
+        if (len > 1 && s.charAt(len - 1) == '/')
+            return s.substring(0, len - 1);
+        return s;
+    }
+
     @Override
     public String resolve(String parent, String child) {
         if (child.isEmpty()) return parent;
         if (child.charAt(0) == '/') {
             if (parent.equals("/")) return child;
-            return parent + child;
+            return trimSeparator(parent + child);
         }
-        if (parent.equals("/")) return parent + child;
-        return parent + '/' + child;
+        if (parent.equals("/")) return trimSeparator(parent + child);
+        return trimSeparator(parent + '/' + child);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


src/java.base/unix/classes/java/io/UnixFileSystem.java
```diff
@@ -101,15 +101,23 @@ public int prefixLength(String pathname) {
         return pathname.startsWith("/") ? 1 : 0;
     }
 
+    // Remove trailing '/' if present and there are at least two characters
+    private static String trimSeparator(String s) {
+        int len = s.length();
+        if (len > 1 && s.charAt(len - 1) == '/')
+            return s.substring(0, len - 1);
+        return s;
```
RogerRiggs, line 109: If there are multiple trailing `/`, is it sufficient to remove only a single `/`?
Does this have the same result as `normalize(pathname)`?

bplb: The main calling context is
```
            if (parent.path.isEmpty()) {
                this.path = FS.resolve(FS.getDefaultParent(),
                                       FS.normalize(child));
            } else {
                this.path = FS.resolve(parent.path,
                                       FS.normalize(child));
            }
```
where `parent` is non-`null` and already normalized. Thus both components should already be normalized and multiple trailing `/` impossible. I _think_ that it should give the same result as `normalize(pathname)`.

RogerRiggs (approved):
LGTM

PR #14110, opened 2023-05-23
8174722: Wrong behavior of DecimalFormat with RoundingMode.UP in special case

justin-curtis-lu:
Please review this PR, which addresses a case where Decimal Format would violate certain RoundingMode contracts given the right pattern and double. Note that this fix is intended only to address the edge case in the implementation, and the lack of specification regarding rounding will be addressed in [JDK-8313434](https://bugs.openjdk.org/browse/JDK-8313434).

For example,

```
DecimalFormat df = new DecimalFormat();
df.setRoundingMode(RoundingMode.UP);
double small = 0.0001;
double big = 1.0001;
df.applyPattern("0.00");
df.format(small); // returns 0.00, which violates UP
df.format(big); // returns 1.01, which does not violate UP
```

In this example `0.0001` becomes `0.00`, a decrease in magnitude. This violates the RoundingMode.UP contract as RoundingMode.UP states "Note that this rounding mode never decreases the magnitude of the calculated value."

The Decimal Format implementation makes an assumption that under-flowing to zero can be made if the pattern allows for it, skipping the consideration of RoundingMode. This is a result of when the pattern has less fractional digits than fractional leading zeroes in the numerical representation of the given double, and the double is less than .1.

---------
### Progress
- [ ] Change requires CSR request [JDK-8308798](https://bugs.openjdk.org/browse/JDK-8308798) to be approved

### Issues
 * [JDK-8174722](https://bugs.openjdk.org/browse/JDK-8174722): Wrong behavior of DecimalFormat with RoundingMode.UP in special case (**Bug** - P4)
 * [JDK-8308798](https://bugs.openjdk.org/browse/JDK-8308798): Wrong behavior of DecimalFormat with RoundingMode.UP in special case (**CSR**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14110/head:pull/14110` \
`$ git checkout pull/14110`

`$ git checkout pull/14110` \
`$ git pull https://git.openjdk.org/jdk.git pull/14110/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14110`

`$ git pr show -t 14110`

</details>

<a href="https://git.openjdk.org/jdk/pull/14110.diff">https://git.openjdk.org/jdk/pull/14110.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14110#issuecomment-1560250979)

### DIFF
--- a/src/java.base/share/classes/java/text/DigitList.java
+++ b/src/java.base/share/classes/java/text/DigitList.java
@@ -393,46 +393,26 @@ final class DigitList implements Cloneable {
      }
 
     /**
-     * Determine if a number should create a 1 in the least significant location
-     * if truncating the representation to the given number of digits would
-     * violate the current RoundingMode contract.
-     * @param maximumDigits The maximum number of digits to be shown.
+     * This method sets the Digit List to a state that is either
+     * equivalent to zero, or a state with a 1 in the least significant location
+     * if an underflow to zero would violate the current RoundingMode contract.
      *
-     * Upon return, count will either be one or zero.
+     * This method does not return any value, and instead adjusts
+     * the instance variables that the Digit List is composed of.
+     *
+     * @param maximumDigits The maximum number of digits to be shown.
      */
     private void underflowToZero(int maximumDigits) {
-        switch(roundingMode) {
-            case UP:
-                // RoundingMode.UP can not decrease the magnitude of the value
-                // whether negative or positive.
-                decimalAt = -maximumDigits + 1;
-                digits[0] = '1';
-                count = 1;
-                break;
-            case CEILING:
-                // RoundingMode.CEILING follows RoundingMode.UP behavior when
-                // the value is positive
-                if (!isNegative) {
-                    decimalAt = -maximumDigits + 1;
-                    digits[0] = '1';
-                    count = 1;
-                } else {
-                    count = 0;
-                }
-                break;
-            case FLOOR:
-                // RoundingMode.FLOOR can not increase the value
-                // when negative
-                if (isNegative) {
-                    decimalAt = -maximumDigits + 1;
-                    digits[0] = '1';
-                    count = 1;
-                } else {
-                    count = 0;
-                }
-                break;
-            default:
-                count = 0;
+        // These modes under the right conditions should not
+        // decrease the magnitude of the formatted value
+        if (roundingMode == RoundingMode.UP
+                || (roundingMode == RoundingMode.CEILING && !isNegative)
+                || (roundingMode == RoundingMode.FLOOR && isNegative)) {
+            decimalAt = -maximumDigits + 1;
+            digits[0] = '1';
+            count = 1;
+        } else {
+            count = 0;
         }
     }
 
--- a/src/java.base/share/classes/java/text/DigitList.java
+++ b/src/java.base/share/classes/java/text/DigitList.java
@@ -363,7 +363,7 @@ final class DigitList implements Cloneable {
             if (-decimalAt > maximumDigits) {
                 // Handle an underflow to zero when we round something like
                 // 0.0009 to 2 fractional digits.
-                count = 0;
+                underflowToZero(maximumDigits);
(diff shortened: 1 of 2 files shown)


jddarcy:
Note there are some hazards here with respect to the numerical values of double and the decimal string used to represent the double; in exact terms:

```
jshell> new BigDecimal(1.0001)
$2 ==> 1.0000999999999999889865875957184471189975738525390625

jshell> new BigDecimal(0.0001)
$3 ==> 0.000100000000000000004792173602385929598312941379845142364501953125


```

justin-curtis-lu:
> Note there are some hazards here with respect to the numerical values of double and the decimal string used to represent the double; in exact terms:
> 
> ```
> jshell> new BigDecimal(1.0001)
> $2 ==> 1.0000999999999999889865875957184471189975738525390625
> 
> jshell> new BigDecimal(0.0001)
> $3 ==> 0.000100000000000000004792173602385929598312941379845142364501953125
> ```

Thanks for bringing that to my attention Joe. I will re-evaluate how I test the changes then.

justin-curtis-lu:
> I just wonder if it is ok to reverse the rounding/truncation only in this special case. If the truncation always comes first, then 0.00 would be correct (then 1.01 may be something wrong). Looks to me that the original comment suggests it tries to truncate those cases. I suggest filing a CSR for further behavioral consideration.

The normal behavior of DecimalFormat (DigitList) "rounds" before truncation. For example, consider

`1.0000009` formatted to `0.0` with `RoundingMode.UP`

DigitList will search past the maximum digits allowed to decide if the formatted value should be rounded. '9' will be found, and as such the value will be "rounded", by incrementing the value by 1 at the max digit location -> `1.1000009`, then DigitList will truncate -> `1.1`

Thus I believe that this change is aligning this special case to be consistent with the existing behavior of rounding before truncation. I will follow-up in the CSR.

jddarcy:
> > I just wonder if it is ok to reverse the rounding/truncation only in this special case. If the truncation always comes first, then 0.00 would be correct (then 1.01 may be something wrong). Looks to me that the original comment suggests it tries to truncate those cases. I suggest filing a CSR for further behavioral consideration.
> 
> The normal behavior of DecimalFormat (DigitList) "rounds" before truncation. For example, consider
> 
> `1.0000009` formatted to `0.0` with `RoundingMode.UP`
> 
> DigitList will search past the maximum digits allowed to decide if the formatted value should be rounded. '9' will be found, and as such the value will be "rounded", by incrementing the value by 1 at the max digit location -> `1.1000009`, then DigitList will truncate -> `1.1`
> 
> Thus I believe that this change is aligning this special case to be consistent with the existing behavior of rounding before truncation. I will follow-up in the CSR.

A quick comment without diving into the particular issues here, for each finite floating-point value, there is a range of the real number line that will round to that floating-point value. Under the default rounding policy, some of those real numbers will be less than the floating-point value and other of those real numbers will be greater than the floating-point value.

So there are decimal strings corresponding to those numbers greater than and less than the floating-point value, all of which will get rounded to the same double value on string conversion.

If the operation is phrased in terms of *the floating-point value* and not *the real number value of what happened to be the input string*, then the actual real value of the floating-point number should be the reference point for rounding up/down on double -> String conversion.

This may give "surprising" results, but is numerically better. Without finding the bug ids, some previous changes were made to decimal format when rounding to nearest along those lines. HTH

naotoj:
Hi @jddarcy , are you referring to the cases like this:https://bugs.openjdk.org/browse/JDK-8062756 ? I understand the binary representation of real number could be below the real number, then the rounding may look strange, but in this particular case, I believe it won't affect the result, as the fix is to the following specific condition (excerpt of the comment there):
```
                // Handle an underflow to zero when we round something like
                // 0.0009 to 2 fractional digits.
```
So even if the number `0.0001` is represented by `0.00009999999something` it should round up to `0.01`.

PR #14111, opened 2023-05-24
8309271: A way to align already compiled methods with compiler directives

dchuyko:
Compiler Control (https://openjdk.org/jeps/165) provides method-context dependent control of the JVM compilers (C1 and C2). The active directive stack is built from the directive files passed with the `-XX:CompilerDirectivesFile` diagnostic command-line option and the Compiler.add_directives diagnostic command. It is also possible to clear all directives or remove the top from the stack.

A matching directive will be applied at method compilation time when such compilation is started. If directives are added or changed, but compilation does not start, then the state of compiled methods doesn't correspond to the rules. This is not an error, and it happens in long running applications when directives are added or removed after compilation of methods that could be matched. For example, the user decides that C2 compilation needs to be disabled for some method due to a compiler bug, issues such a directive but this does not affect the application behavior. In such case, the target application needs to be restarted, and such an operation can have high costs and risks. Another goal is testing/debugging compilers.

It would be convenient to optionally reconcile at least existing matching nmethods to the current stack of compiler directives (so bypass inlined methods).

Natural way to eliminate the discrepancy between the result of compilation and the broken rule is to discard the compilation result, i.e. deoptimization. Prior to that we can try to re-compile the method letting compile broker to perform it taking new directives stack into account. Re-compilation helps to prevent hot methods from execution in the interpreter.

A new flag `-r` has beed introduced for some directives related to compile commands: `Compiler.add_directives`, `Compiler.remove_directives`, `Compiler.clear_directives`. The default behavior has not changed (no flag). If the new flag is present, the command scans already compiled methods and puts methods that have any active non-default matching compiler directives to re-compilation if possible, otherwise marks them for deoptimization. There is currently no distinction which directives are found. In particular, this means that if there are rules for inlining into some method, it will be refreshed. On the other hand, if there are rules for a method and it was inlined, top-level methods won't be refreshed, but this can be achieved by having rules for them.

In addition, a new diagnostic command `Compiler.replace_directives`, has been added for convenience. It's like a combination of `Compiler.clear_directives` and `Compiler.add_directives`. It supports the same new optional `-r` flag that refreshes both cleared and added methods. t is more robust as we don't know when methods affected 'clear' are finally compiled.

The behavior of the `-r` flag is implemented in the new `CodeCache::mark_directives_matches`, `CodeCache::recompile_marked_directives_matches` and `DirectivesStack::hasMatchingDirectives` methods. The mark is stored as a method flag.

New `DirectivesRefreshTest` checks `-r` behavior (using `add` and `clear` commands). `CompilerDirectivesDCMDTest`  performs a basic check of `replace` command.

An alternative approach to the `-r` flag could be to have a special diagnostic command for re-compilation or de-optimization. It would get a list of method patterns and reuse the matcher, however this is not so trivial. Overall usage and effects will be similar but this is one more file format. The user will also need to monitor or query active directives in advance, e.g. to refresh all mentioned methods after clearing all directives.

An alternative approach for selection of refreshed methods could be to track down all inlining dependencies. This may be similar to searching references to old methods, but it requires scanning all code blobs, which looks too expensive.

Though it is a diagnostic feature, the performance impact on a running application matters. In benchmarks like Renaissance it takes ~<N*10ms to apply directives in the steady state to refresh ~M*1000 methods. If there are no specifically long compilation, re-compilation takes ~<M*sec. Refresh allows DCMD commands to change application behavior while running similar to having `-XX:CompilerDirectivesFile` from the start.

---------
### Progress

### Issue
 * [JDK-8309271](https://bugs.openjdk.org/browse/JDK-8309271): A way to align already compiled methods with compiler directives (**Enhancement** - P4)

### Reviewers
 * [Andrei Pangin](https://openjdk.org/census#apangin) (@apangin - no project role) ⚠️ Review applies to [a4578277](https://git.openjdk.org/jdk/pull/14111/files/a457827766ea0ea381bd492eb30c5640eeb59273)
 * [Tobias Holenstein](https://openjdk.org/census#tholenstein) (@tobiasholenstein - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14111/head:pull/14111` \
`$ git checkout pull/14111`

`$ git checkout pull/14111` \
`$ git pull https://git.openjdk.org/jdk.git pull/14111/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14111`

`$ git pr show -t 14111`

</details>

<a href="https://git.openjdk.org/jdk/pull/14111.diff">https://git.openjdk.org/jdk/pull/14111.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14111#issuecomment-1571686593)


phohensee:
"refresh" (-r) would be better than "deoptimize" (-d). The latter implies a specific implementation, the former is generic.

If the method is to be recompiled, perhaps rather than deopt and wait, add it to the compile queue immediately and deopt the old version when the new compilation is complete, similar to what happens when the c1 version of the method is replaced by the c2 version.

src/hotspot/share/code/codeCache.cpp
```diff
@@ -1368,6 +1369,23 @@ void CodeCache::mark_all_nmethods_for_deoptimization(DeoptimizationScope* deopt_
...
 }
 
+void CodeCache::mark_for_deoptimization_directives_matches(DeoptimizationScope *deopt_scope) {
+  MutexLocker mu(CodeCache_lock, Mutex::_no_safepoint_check_flag);
+  Thread *thread = Thread::current();
+
+  CompiledMethodIterator iter(CompiledMethodIterator::only_not_unloading);
+  while(iter.next()) {
+    CompiledMethod* nm = iter.method();
+    HandleMark hm(thread);
```
apangin, line 1379: Isn't it better to move `HandleMark` outside the loop?

src/hotspot/share/services/diagnosticCommand.cpp
```diff
@@ -884,21 +886,91 @@ void CompilerDirectivesPrintDCmd::execute(DCmdSource source, TRAPS) {
 
 CompilerDirectivesAddDCmd::CompilerDirectivesAddDCmd(outputStream* output, bool heap) :
                            DCmdWithParser(output, heap),
-  _filename("filename","Name of the directives file", "STRING",true) {
+  _filename("filename", "Name of the directives file", "STRING", true),
+  _force_deopt("-d", "Force deoptimization of affected methods.", "BOOLEAN", false, "false") {
```
apangin, line 890: I agree with Paul a generic alternative like `refresh` would be better.

src/hotspot/share/services/diagnosticCommand.cpp
```diff
@@ -884,21 +886,91 @@ void CompilerDirectivesPrintDCmd::execute(DCmdSource source, TRAPS) {
...
+  _dcmdparser.add_dcmd_option(&_force_deopt);
 }
 
 void CompilerDirectivesRemoveDCmd::execute(DCmdSource source, TRAPS) {
-  DirectivesStack::pop(1);
+  // Have to duplicate code to shut an assert in DTR of DeoptimizationScope
+  if (_force_deopt.value()) {
+    DeoptimizationScope deopt_scope;
+    CodeCache::mark_for_deoptimization_directives_matches(&deopt_scope);
+    DirectivesStack::pop(1);
```
apangin, line 946: Why deoptimizing methods from the whole stack if we change only the topmost list?

src/hotspot/share/services/diagnosticCommand.hpp
```diff
@@ -727,9 +731,36 @@ class CompilerDirectivesAddDCmd : public DCmdWithParser {
   virtual void execute(DCmdSource source, TRAPS);
 };
 
-class CompilerDirectivesClearDCmd : public DCmd {
+class CompilerDirectivesReplaceDCmd : public DCmdWithParser {
```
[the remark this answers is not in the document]
dchuyko: 'replace' is 'clear'+'add' (we don't have indexes and don't mention top element). Additional command is made to avoid double refreshing when '-r' is provided. It is more robust as we don't know when methods affected 'clear' are finally compiled.

src/hotspot/share/runtime/mutexLocker.cpp
```diff
@@ -271,7 +271,7 @@ void mutex_init() {
   MUTEX_DEFN(CompiledIC_lock                 , PaddedMutex  , nosafepoint);  // locks VtableStubs_lock, InlineCacheBuffer_lock
   MUTEX_DEFN(MethodCompileQueue_lock         , PaddedMonitor, safepoint);
   MUTEX_DEFN(CompileStatistics_lock          , PaddedMutex  , safepoint);
-  MUTEX_DEFN(DirectivesStack_lock            , PaddedMutex  , nosafepoint);
+  MUTEX_DEFN(DirectivesStack_lock            , PaddedMutex  , nosafepoint-3);
```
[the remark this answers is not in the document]
dchuyko: Changed the definition to use another macro and relative ranking.

src/hotspot/share/services/diagnosticCommand.cpp
dchuyko: Agree, now only top directive is involved for add/remove.

src/hotspot/share/services/diagnosticCommand.cpp
dchuyko: Changed the flag, description and names to 'refresh', 'r'.

src/hotspot/share/compiler/compilerDirectives.cpp
```diff
@@ -735,6 +735,29 @@ void DirectivesStack::release(CompilerDirectives* dir) {
...
+  {
+    MutexLocker locker(DirectivesStack_lock, Mutex::_no_safepoint_check_flag);
+
+    CompilerDirectives* dir = _top;
+    assert(dir != nullptr, "Must be initialized");
+
+    while (dir != nullptr) {
+      if (!dir->is_default_directive() && dir->match(method)) {
+        match_found = true;
+        break;
```
apangin, line 750: `match_found` is redundant: for better readability, you may just return true. Curly braces around MutexLocker won't be needed either.

src/hotspot/share/oops/method.hpp
```diff
@@ -815,6 +815,14 @@ class Method : public Metadata {
     return _method_counters;
   }
 
+  // Clear the flags related to compiler directives that were set by the compilerBroker,
+  // because the directives can be updated.
+  void clear_method_flags() {
```
apangin, line 820: The function name is a bit misleading - it clears only flags related to directives.

src/hotspot/share/oops/methodFlags.hpp
```diff
@@ -58,6 +58,7 @@ class MethodFlags {
    status(has_loops_flag              , 1 << 13) /* Method has loops */ \
    status(has_loops_flag_init         , 1 << 14) /* The loop flag has been initialized */ \
    status(on_stack_flag               , 1 << 15) /* RedefineClasses support to keep Metadata from being cleaned */ \
+   status(has_matching_directives     , 1 << 16) /* The method has matching directives */ \
```
apangin, line 61: It's worth noting that the flag is temporary and is valid only during DCmd execution.

src/hotspot/share/code/codeCache.cpp
```diff
@@ -1381,6 +1382,54 @@ void CodeCache::mark_all_nmethods_for_evol_deoptimization(DeoptimizationScope* d
...
+  HandleMark hm(thread);
+
+  RelaxedCompiledMethodIterator iter(RelaxedCompiledMethodIterator::only_not_unloading);
+  while(iter.next()) {
+    CompiledMethod* nm = iter.method();
+    methodHandle mh(thread, nm->method());
+    if (mh->has_matching_directives()) {
+      ResourceMark rm;
+      // Try the max level and let the directives be applied during the compilation.
+      int complevel = CompLevel::CompLevel_full_optimization;
```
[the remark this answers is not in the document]
dchuyko: Yes, changed to use `highest_compile_level()`.

src/hotspot/share/compiler/compilerDirectives.cpp
dchuyko: Thanks, that's indeed simpler.

src/hotspot/share/oops/method.hpp
dchuyko: Changed to `clear_directive_flags`.

src/hotspot/share/oops/methodFlags.hpp
dchuyko: Good point, updated the comment. This btw means that in another places this flag can be reused.

src/hotspot/share/code/codeCache.cpp
```diff
@@ -1381,6 +1382,54 @@ void CodeCache::mark_all_nmethods_for_evol_deoptimization(DeoptimizationScope* d
...
+}
+
+void CodeCache::recompile_marked_directives_matches() {
+  Thread *thread = Thread::current();
+  HandleMark hm(thread);
+
+  RelaxedCompiledMethodIterator iter(RelaxedCompiledMethodIterator::only_not_unloading);
+  while(iter.next()) {
+    CompiledMethod* nm = iter.method();
+    methodHandle mh(thread, nm->method());
```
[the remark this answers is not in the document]
dchuyko: If there are multiple method versions, it will be recompiled several times. The alternative is too keep some additional information which may complicate the code. OSRs is a good catch, I changed their handling to deopt.

apangin (approved):
The logic looks good to me now, thanks.

dchuyko:
@sspitsyn Can I please have someone from serviceability group take a look at this? Comments from Paul and Andrei have been addressed a while ago.

dchuyko:
/cc hotspot-compiler

As this is an extension of the existing compiler control mechanism.

src/hotspot/share/ci/ciEnv.cpp
```diff
@@ -1141,17 +1141,15 @@ void ciEnv::register_method(ciMethod* target,
 #endif
 
       if (entry_bci == InvocationEntryBci) {
-        if (TieredCompilation) {
```
sspitsyn, line 1144: Just a naive question. Why this check has been removed?

src/hotspot/share/services/diagnosticCommand.cpp
```diff
@@ -923,21 +924,81 @@ void CompilerDirectivesPrintDCmd::execute(DCmdSource source, TRAPS) {
 
 CompilerDirectivesAddDCmd::CompilerDirectivesAddDCmd(outputStream* output, bool heap) :
                            DCmdWithParser(output, heap),
-  _filename("filename","Name of the directives file", "STRING",true) {
+  _filename("filename", "Name of the directives file", "STRING", true),
+  _refresh("-r", "Refresh affected methods.", "BOOLEAN", false, "false") {
```
sspitsyn, line 928: Nit: The dot is not needed at the end, I think. The same applies to lines: 945, 970 and 987.

src/hotspot/share/ci/ciEnv.cpp
dchuyko: We want to let replacement of C2 method version by another C2 version of the same method in both tired and non-tired mode, which was not allowed

src/hotspot/share/services/diagnosticCommand.cpp
dchuyko: Thanks, the dots were removed.

sspitsyn (approved):
The fix looks good. But I do not have an expertise in the compiler-specific part.
So, a review from the Compiler team is still required.

tobiasholenstein (approved):
Looks good to me too (Compiler Team)

dchuyko:
Thank you, Seguei and Tobias.

PR #14112, opened 2023-05-24
8308339: AArch64: Remove extra `UseSVE` Predicate in ad file

fg1417:
If a match rule belongs to one of the following situations, we can remove extra `UseSVE` Predicate:

1. If any src operand type is `pReg`, which is SVE specific, we can remove `Predicate(UseSVE > 0)`. But if only dst operand type is `pReg`, we can't remove `Predicate(UseSVE > 0)`, since the DFA of matcher selects by src operands and instruction cost, not involving dst operand.

2. If matcher can use src operand type, i.e., `pReg` or `vReg`, to distinguish sve from neon, we can remove
`Predicate(UseSVE == 0)` for rules on neon.

3. When the condition in `Predicate()` is false on current platform, it's definitely impossible to generate the corresponding node pattern from C2. Then we can remove `Predicate()`, like removing `predicate(UseSVE > 0)` for all `PopulateIndex` rules.

After the patch, the code size of libjvm.so decreased from 25.42M to 25.39M, by 25.3K.

Testing:
  No new failures found on tier 1 - 3.
  No significant performance regression compared with master.

---------
### Progress

### Issue
 * [JDK-8308339](https://bugs.openjdk.org/browse/JDK-8308339): AArch64: Remove extra `UseSVE` Predicate in ad file

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14112/head:pull/14112` \
`$ git checkout pull/14112`

`$ git checkout pull/14112` \
`$ git pull https://git.openjdk.org/jdk.git pull/14112/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14112`

`$ git pr show -t 14112`

</details>

<a href="https://git.openjdk.org/jdk/pull/14112.diff">https://git.openjdk.org/jdk/pull/14112.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14112#issuecomment-1560368022)

### DIFF
--- a/src/hotspot/cpu/aarch64/aarch64_vector.ad
+++ b/src/hotspot/cpu/aarch64/aarch64_vector.ad
@@ -447,7 +447,6 @@ instruct storeV(vReg src, vmemA mem) %{
 // vector load/store - predicated
 
 instruct loadV_masked(vReg dst, vmemA mem, pRegGov pg) %{
-  predicate(UseSVE > 0);
   match(Set dst (LoadVectorMasked mem pg));
   format %{ "loadV_masked $dst, $pg, $mem" %}
   ins_encode %{
@@ -460,7 +459,6 @@ instruct loadV_masked(vReg dst, vmemA mem, pRegGov pg) %{
 %}
 
 instruct storeV_masked(vReg src, vmemA mem, pRegGov pg) %{
-  predicate(UseSVE > 0);
   match(Set mem (StoreVectorMasked mem (Binary src pg)));
   format %{ "storeV_masked $mem, $pg, $src" %}
   ins_encode %{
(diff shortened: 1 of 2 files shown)


theRealAph:
I appreciate the space saving, but this doesn't help maintainability.
A predicate `UseSVE` is also a flag for the reader that this is SVR-only.

src/hotspot/cpu/aarch64/aarch64_vector.ad
```diff
@@ -447,7 +447,6 @@ instruct storeV(vReg src, vmemA mem) %{
 // vector load/store - predicated
 
 instruct loadV_masked(vReg dst, vmemA mem, pRegGov pg) %{
-  predicate(UseSVE > 0);
```
theRealAph, line 450: How about

```
  // This predicate is unneeded because only SVE has pRegs.
  // predicate(UseSVE > 0);
```
I still don't like it much, but at least it's clear.

fg1417: > How about
> 
> ```
>   // This predicate is unneeded because only SVE has pRegs.
>   // predicate(UseSVE > 0);
> ```
> 
> I still don't like it much, but at least it's clear.

Hi @theRealAph, thanks for your kind suggestion. I agree on you said. Compared with current simple rules, the changed rules may be not easy to follow. Since code size is not a big problem now, maintaining the status quo seems better.

I'll close the PR. Thanks again for your review!

PR #14114, opened 2023-05-24
8308751: Create new switch to print error reporting output to both hs_err_pid file and stdout/stderr

masyano:
I think it makes sense to add the ErrorFileWithStdout and ErrorFileWithStderr for troubleshooting.
I would appriciate if someone could review it.

---------
### Progress
- [ ] Change requires CSR request [JDK-8308947](https://bugs.openjdk.org/browse/JDK-8308947) to be approved

### Issues
 * [JDK-8308751](https://bugs.openjdk.org/browse/JDK-8308751): Create new switch to print error reporting output to both hs_err_pid file and stdout/stderr (**Enhancement** - P4)
 * [JDK-8308947](https://bugs.openjdk.org/browse/JDK-8308947): Create new switch to print error reporting output to both hs_err_pid file and stdout/stderr (**CSR**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14114/head:pull/14114` \
`$ git checkout pull/14114`

`$ git checkout pull/14114` \
`$ git pull https://git.openjdk.org/jdk.git pull/14114/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14114`

`$ git pr show -t 14114`

</details>

<a href="https://git.openjdk.org/jdk/pull/14114.diff">https://git.openjdk.org/jdk/pull/14114.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14114#issuecomment-1560668271)

### DIFF
--- a/src/hotspot/share/utilities/vmError.cpp
+++ b/src/hotspot/share/utilities/vmError.cpp
@@ -1712,7 +1712,7 @@ void VMError::report_and_die(int id, const char* message, const char* detail_fmt
     }
   }
 
-  // Part 1: print to screen
+  // Part 1: print to stdout/stderr
   if (!out_done) {
     // print an abbreviated version (the '#' section) to stdout by default.
     // full error log is printed if the print option to stdout and stderr is specified
--- a/src/hotspot/share/utilities/vmError.cpp
+++ b/src/hotspot/share/utilities/vmError.cpp
@@ -1712,40 +1712,43 @@ void VMError::report_and_die(int id, const char* message, const char* detail_fmt
     }
   }
 
-  // Part 1: print an abbreviated version (the '#' section) to stdout.
+  // Part 1: print to screen
   if (!out_done) {
-    // Suppress this output if we plan to print Part 2 to stdout too.
-    // No need to have the "#" section twice.
-    if (!(ErrorFileToStdout && out.fd() == 1)) {
-      report(&out, false);
+    // print an abbreviated version (the '#' section) to stdout by default.
+    // full error log is printed if the print option to stdout and stderr is specified
+    bool verbose = false;
+    if (ErrorFileToStdout || ErrorFileToStderr || ErrorFileWithStdout || ErrorFileWithStderr) {
+      verbose = true;
+    }
+    if (ErrorFileToStderr || ErrorFileWithStderr) {
+      out.set_fd(2);
     }
 
+    report(&out, verbose);
+
     out_done = true;
+    out.set_fd(fd_out);
 
     _current_step = 0;
     _current_step_info = "";
   }
 
-  // Part 2: print a full error log file (optionally to stdout or stderr).
-  // print to error log file
-  if (!log_done) {
+  // Part 2: print a full error log file
+  if (!log_done && !ErrorFileToStdout && !ErrorFileToStderr) {
     // see if log file is already open
     if (!log.is_open()) {
       // open log file
-      if (ErrorFileToStdout) {
-        fd_log = 1;
-      } else if (ErrorFileToStderr) {
-        fd_log = 2;
-      } else {
-        fd_log = prepare_log_file(ErrorFile, "hs_err_pid%p.log", true,
-                 buffer, sizeof(buffer));
-        if (fd_log != -1) {
-          out.print_raw("# An error report file with more information is saved as:\n# ");
-          out.print_raw_cr(buffer);
-        } else {
-          out.print_raw_cr("# Can not save log file, dump to screen..");
-          fd_log = 1;
+      fd_log = prepare_log_file(ErrorFile, "hs_err_pid%p.log", true, buffer, sizeof(buffer));
+      if (fd_log != -1) {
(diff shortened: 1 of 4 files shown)


src/hotspot/share/utilities/vmError.cpp
```diff
@@ -1712,40 +1712,43 @@ void VMError::report_and_die(int id, const char* message, const char* detail_fmt
     }
   }
 
-  // Part 1: print an abbreviated version (the '#' section) to stdout.
+  // Part 1: print to screen
```
dholmes-ora, line 1715: What is a "screen"? We print to files/file-descriptors

dholmes-ora:
/csr needed

I have my doubts about the usefulness of this. The hs_err content can be very large (and only gets bigger over time) and can easily overwhelm any observed "screen". If stdout and stderr are being captured by another utility, such as jtreg when running tests, then again we can easily hit buffer limits and actually lose useful information in the captured log file for stdout/stderr.

While the implementation is relatively simple I do not like adding yet another pair of flags in this area - and a CSR request is needed to do this.

src/hotspot/share/utilities/vmError.cpp
masyano: Probably "screen" means the console on display. It is description before JDK-8220786 was fixed.
Is it more accurate to change it to "stdout/stderr"?

masyano:
@dholmes-ora I have created a CSR, so please confirm it.

As you say, if the error report is output only to stdout or stderr, information can be lost. Similarly, if the error report print to the hs-err log file only, there may not be any files left.
I would like to create an option to leave both in order to increase the possibility of a post-mortem investigation.

src/hotspot/share/utilities/vmError.cpp
dholmes-ora: Yes please

tstuefe:
> /csr needed
> 
> I have my doubts about the usefulness of this.

I agree. This is nothing a simple `tee` couldn't solve. 

We've long departed from the Unix line of "do one thing and do it well", but there is a limit to how much functionality we should heap onto the JVM.

masyano:
@tstuefe As you say, there are situations where we can solve with the tee command, but only if java is invoked as a shell command. In the case of an application server, it is difficult to redirect to the tee command because java is started as a subprocess of the administration service.

Another reason is that the tee command does not exist in every environment where java runs. In a windows environment, we cannot use features like the tee command without adding features like wsl or cygwin.

I'm not proposing complex features outside the Unix philosophy, I'm just suggesting a way to enable options that are currently mutually exclusive.

tstuefe:
> @tstuefe As you say, there are situations where we can solve with the tee command, but only if java is invoked as a shell command. In the case of an application server, it is difficult to redirect to the tee command because java is started as a subprocess of the administration service.
> 
> Another reason is that the tee command does not exist in every environment where java runs. In a windows environment, we cannot use features like the tee command without adding features like wsl or cygwin.
> 
> I'm not proposing complex features outside the Unix philosophy, I'm just suggesting a way to enable options that are currently mutually exclusive.

First off, your implementation is not valid, at least not for more complex cases: 

You call report() twice. `report(verbose=true)` should only be called once since it is a bit of a "one-shot" operation. Many of the error-reporting steps may be destructive. They may crash or hang. They even may change runtime context. Even if they don't, they will run at different times (since hs_er reporting can take a while), which could cause the second printout to be different and probably less truthful since it is further removed from the crash point.

So, calling report(verbose=true) twice is absolutely not an option and we should add asserts for this. The only time when calling it again is valid is in the context of secondary crash handling, and even then we never repeat an error reporting step but skip over the ones that already ran.

To implement what you want, there are two options:

A) either buffer the content of the first printout. 
B) add a new type of outputStream that forks and prints twice

(A) is not reasonable. hs_err files can be massive, in the multi-gb range, for very large heaps. This is not something you want to cache.

(B) adds more even complexity.

So it comes down to complexity vs. benefit. The costs are:

- obviously, review cost for CSR and PR. These costs can be a lot higher than the initial effort that the author spent on his PR.
- more complex user-facing interface: option must be documented and explained. Extending this interface causes us to be less flexible in the future, translating to cost-of-opportunity. Also, customer support: every new feature *will* be used by some enterprising customer, if things go wrong someone needs to help them.
- more technical debt, more complex code: the more complex this coding gets, the more expensive future maintenance gets.
- more regression tests for both the new option as well as for the new forking outputStream (B) (these are machine costs as well as costs to fix regressions)

All of this is normal and if the new feature adds enough benefit, we do this of course. But please remember, these costs have to be carried by us, the maintainers, almost exclusively. And carried in all perpetuity since it is an official option that is meant to stay. 

IMHO, in this case the benefits - while they exist - are too slim to carry the costs.

masyano:
I understand why you think this should not be added.
I would like to withdraw this PR and CSR in order to reconsider.

PR #14119, opened 2023-05-24
8240774: [REDO] G1DirtyCardQueue destructor has useless flush

kimbarrett:
Please review this change to VM initialization, to eliminate a workaround for
a bootstrapping problem.

A JavaThread must not touch oops and invoke barriers unless it is on the
threads list.  Barrier set operations on_thread_attach and on_thread_detach
bracket the time when a thread can do stuff with oops.

However, this restriction is currently violated by the main thread during VM
initialization.  It may (and indeed does) touch oops before it has been added
to the thread list, including executing post-barriers and enqueuing entries
in its DCQ.

That has led to two separate bootstrapping workarounds:

- ~G1DirtyCardQueue flushes the queue before proceeding to the base class
(PtrQueue) destructor. This is not normally needed, and is even
counter-productive, since it may hide a problem. But if an error is detected
during VM initialization that main thread gets deleted. Since it's not on the
threads list, there's no on_thread_detach to flush the queue. Without the
flush in the DCQ destructor that can lead to an assert in ~PtrQueue because
the queue isn't empty.

- on_thread_attach can't assert the queue is empty, even though normally it
must be.  But for the main thread, being added to the threads list (so calling
on_thread_attach) happens after it has enqueued entries in its DCQ.  So we
currently don't have those asserts.

To remove the need for these workarounds, this change moves the addition of
the main thread to the threads list earlier in the initialization sequence. It
needs to be before code that might execute barriers. After init_globals is too
late. But it can't be too early, before some relevant data structures are
initialized, so it can't be before init_globals. So this change splits
init_globals at a strategic point, and adds the main thread to the threads
list there. That point is just before the call to universe2_init, which calls
Universe::genesis.

With that change we can then remove the workarounds in ~G1DirtyCardQueue and
G1BarrierSet::on_thread_attach.

Testing:
mach5 tier1-5

Manually tested error handling for failures reported by init_globals and
init_globals2. This involved patching to force error reporting when
-XX:+UseNewCode. Handling errors from init_globals works as before, if a
failure happened in that part of the sequence. Handling errors from
init_globals2 runs afoul of pre-existing JDK-8308764.

---------
### Progress

### Issue
 * [JDK-8240774](https://bugs.openjdk.org/browse/JDK-8240774): [REDO] G1DirtyCardQueue destructor has useless flush

### Reviewers
 * [David Holmes](https://openjdk.org/census#dholmes) (@dholmes-ora - **Reviewer**) ⚠️ Review applies to [3b4b69e2](https://git.openjdk.org/jdk/pull/14119/files/3b4b69e25755d6ad8a400615ca7d95f900e29727)
 * [Albert Mingkun Yang](https://openjdk.org/census#ayang) (@albertnetymk - **Reviewer**) ⚠️ Review applies to [3b4b69e2](https://git.openjdk.org/jdk/pull/14119/files/3b4b69e25755d6ad8a400615ca7d95f900e29727)
 * [Thomas Schatzl](https://openjdk.org/census#tschatzl) (@tschatzl - **Reviewer**) ⚠️ Review applies to [3b4b69e2](https://git.openjdk.org/jdk/pull/14119/files/3b4b69e25755d6ad8a400615ca7d95f900e29727)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14119/head:pull/14119` \
`$ git checkout pull/14119`

`$ git checkout pull/14119` \
`$ git pull https://git.openjdk.org/jdk.git pull/14119/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14119`

`$ git pr show -t 14119`

</details>

<a href="https://git.openjdk.org/jdk/pull/14119.diff">https://git.openjdk.org/jdk/pull/14119.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14119#issuecomment-1560864047)


dholmes-ora (approved):
Can't really comment on the workaround removal, but the change to the init sequence looks okay.

Thanks.

albertnetymk (approved):

src/hotspot/share/gc/g1/g1BarrierSet.cpp
```diff
@@ -148,9 +148,9 @@ void G1BarrierSet::on_thread_attach(Thread* thread) {
   assert(!queue.is_active(), "SATB queue should not be active");
   assert(queue.buffer() == nullptr, "SATB queue should not have a buffer");
   assert(queue.index() == 0, "SATB queue index should be zero");
-  // Can't assert that the DCQ is empty.  There is early execution on
-  // the main thread, before it gets added to the threads list, which
-  // is where this is called.  That execution may enqueue dirty cards.
+  G1DirtyCardQueue& dq = G1ThreadLocalData::dirty_card_queue(thread);
```
tschatzl, line 151: Would prefer to use distinct names for the two kinds of queues here, i.e. `satb_queue` and `dirty_queue` for better distinction.

tschatzl (approved):
Lgtm

Feel free to ignore my bikeshedding as needed.

src/hotspot/share/runtime/threads.cpp
```diff
@@ -553,15 +553,26 @@ jint Threads::create_vm(JavaVMInitArgs* args, bool* canTryAgain) {
     return status;
   }
 
+  // Add main_thread to threads list to finish barrier setup with
+  // on_thread_attach.  Should be before starting to build Java objects in
+  // init_globals2, which invokes barriers.
+  { MutexLocker mu(Threads_lock);
+    Threads::add(main_thread);
+  }
```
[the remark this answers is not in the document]
kimbarrett: Changing.

src/hotspot/share/gc/g1/g1BarrierSet.cpp
kimbarrett: Changing to "satbq" and "dirtyq".

kimbarrett:
Thanks for reviews @albertnetymk , @tschatzl , and @dholmes-ora

PR #14121, opened 2023-05-24
8308766: TLAB initialization may cause div by zero

tschatzl:
Hi all,

  can I have reviews for this change that fixes an FP div by zero?

In `ThreadLocalAllocBuffer::initialize()` we initialize the TLAB using current available TLAB capacity for the thread. In G1, this can be zero in some situations, leading to that div by zero (see the CR for the crash when adding an assert).
The suggested fix is to just not sample at this point. TLAB resizing will fix TLAB sizing up.

Only G1 seems to be affected as it seems to be the only gc that uses a dynamic value for the capacity available for TLAB allocation. Other GCs seem to just use total heap capacity (Z, Shenandoah) or eden capacity (Serial, Parallel).
Not sure if that is actually better and I think won't result in the expected behavior (every thread should reload TLABs `target_refills()` times per mutator time); since even with G1 at TLAB resizing time eden is maximal, this hiccup at initialization does not seem too bad.

This may also be the cause for the behavior observed in https://bugs.openjdk.org/browse/JDK-8264798. 

Testing: gha

Thanks,
  Thomas

---------
### Progress

### Issue
 * [JDK-8308766](https://bugs.openjdk.org/browse/JDK-8308766): TLAB initialization may cause div by zero

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14121/head:pull/14121` \
`$ git checkout pull/14121`

`$ git checkout pull/14121` \
`$ git pull https://git.openjdk.org/jdk.git pull/14121/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14121`

`$ git pr show -t 14121`

</details>

<a href="https://git.openjdk.org/jdk/pull/14121.diff">https://git.openjdk.org/jdk/pull/14121.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14121#issuecomment-1560985628)

### DIFF
--- a/src/hotspot/share/gc/shared/threadLocalAllocBuffer.cpp
+++ b/src/hotspot/share/gc/shared/threadLocalAllocBuffer.cpp
@@ -213,9 +213,11 @@ void ThreadLocalAllocBuffer::initialize() {
   set_desired_size(initial_desired_size());
 
   size_t capacity = Universe::heap()->tlab_capacity(thread()) / HeapWordSize;
-  // Keep alloc_frac as float and not double to avoid the double to float conversion
-  float alloc_frac = desired_size() * target_refills() / (float) capacity;
-  _allocation_fraction.sample(alloc_frac);
+  if (capacity > 0) {
+    // Keep alloc_frac as float and not double to avoid the double to float conversion
+    float alloc_frac = desired_size() * target_refills() / (float)capacity;
+    _allocation_fraction.sample(alloc_frac);
+  }
 
   set_refill_waste_limit(initial_refill_waste_limit());
 


tschatzl:
> OK, so this does happen when a new thread comes at unfortunate time in VM lifecycle, like on shutdown? Anyway, the fix looks okay. I think many other versions are also affected, can you please add relevant Affected-Versions to the bug?

In this case, yes, a thread is attached on shutdown and you can get weird failures in other FP code (if you also enable FP exceptions, but it can leave something in a weird state apparently). I think (well I hope) it is also the cause for another similar bug that caused crashes in G1 extremely intermittently (that has been closed as CNR at that point after it stopped appearing).

That assert that tripped is something that I added for trying to reproduce [JDK-8264798](https://bugs.openjdk.org/browse/JDK-8264798), initially failed to do so, and then accidentally left in when testing another change....

I think it is worth cleaning up just in case.

tschatzl:
Added affects version back to JDK 8 since that code and the `tlab_capacity()` implementation are the same as they are now. Maybe other circumstance prevent this from happening.

albertnetymk (approved):
Thanks to Thomas' explanation, now I understand why it tracks the ratio instead of the actual alloc-amount. It's because (eden) capacity affects the distance btw two gc-pause (in STW GC), and alloc-amount is semi-proportional to gc-distance. Therefore, the ratio more or less reflects alloc-rate, which can be used to predict alloc-amount until the next gc-pause.

However, maintaining a constant number of refills btw gc-pauses seems an odd objective; preexisting issue.

tschatzl:
Thanks @albertnetymk @shipilev for your reviews.

/integrate

PR #14122, opened 2023-05-24
8308009: Generational ZGC: OOM before clearing all SoftReferences

fisk:
When a major GC in generational ZGC with a different cause that doesn’t pre-clean and doesn’t clear soft references, we ask if there are allocations stalled on old. And part of that condition is to check if we are not stalled on young. So if an allocation request comes in just before such a “weak” major GC, we will say we won’t clear soft references. But after that major collection we will satisfy all the constraints to throw OOM as both an YC and OC has passed since the allocation request was installed.
The solution is to let the driver remember if it cleared soft references or not, and only throw OOM if it cleared soft references.

---------
### Progress

### Issue
 * [JDK-8308009](https://bugs.openjdk.org/browse/JDK-8308009): Generational ZGC: OOM before clearing all SoftReferences

### Reviewers
 * [Stefan Karlsson](https://openjdk.org/census#stefank) (@stefank - **Reviewer**) ⚠️ Review applies to [45b2b269](https://git.openjdk.org/jdk/pull/14122/files/45b2b26964978c40b435094de173b8549f973d84)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14122/head:pull/14122` \
`$ git checkout pull/14122`

`$ git checkout pull/14122` \
`$ git pull https://git.openjdk.org/jdk.git pull/14122/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14122`

`$ git pr show -t 14122`

</details>

<a href="https://git.openjdk.org/jdk/pull/14122.diff">https://git.openjdk.org/jdk/pull/14122.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14122#issuecomment-1561028876)

### DIFF
--- a/src/hotspot/share/gc/z/zDriver.cpp
+++ b/src/hotspot/share/gc/z/zDriver.cpp
@@ -385,10 +385,6 @@ public:
       _gc_cause_setter(ZDriver::major(), _gc_cause),
       _stat_timer(ZPhaseCollectionMajor, gc_timer),
       _tracer(false /* minor */) {
-    // Set up soft reference policy
-    const bool clear = should_clear_soft_references(request.cause());
-    ZGeneration::old()->set_soft_reference_policy(clear);
-
     // Select number of worker threads to use
     ZGeneration::young()->set_active_workers(request.young_nworkers());
     ZGeneration::old()->set_active_workers(request.old_nworkers());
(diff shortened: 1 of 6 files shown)


stefank (approved):

fisk:
> I think it looks good.
> 
> Only one thing I can think of. There is an implicit assumption here that should_clear_soft_references implies should_preclean_young which is required to be spec compliant. Maybe `should_preclean_young` should start with
> 
> ```c++
> if (should_clear_soft_references(cause)) {
>   return true;
> }
> ```
> 
> to make this clearer. Or even short circuit the `if (should_preclean_young(...))` directly in `ZDriverMajor::collect_young`

Right. This is already guaranteed because

1) All the causes that trigger soft ref clearing GC, also trigger young precleaning
2) We hold the driver locker between the should_clear_soft_references and should_preclean_young, which means that if soft ref clearing was triggered due to there being a stall, the stall will be there again when we check if we should_preclean_young.

Do we want a comment explaining the above, or change the code structure to be more explicit? I personally don't mind.

xmas92 (approved):

fisk:
Thanks for the review @xmas92!

PR #14123, opened 2023-05-24
8308583: SIGSEGV in GraphKit::gen_checkcast

rwestrel:
At an `instanceof`, a node of type `bottom[int:>=0]` is checked to be of type `cc$Word` and on the success a `CheckCastPP` is inserted to change the node type. That `CheckCastPP` constant folds to `top` but the type check doesn't fold. The reason is that the type check loads the klass from the node with a `LoadNKlass` and the type of that node is `java/lang/Object`  when it should be `bottom[int:>=0]` but logic in `LoadNKlass::Value()` gets in the way.

---------
### Progress

### Issue
 * [JDK-8308583](https://bugs.openjdk.org/browse/JDK-8308583): SIGSEGV in GraphKit::gen_checkcast

### Reviewers
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - **Reviewer**) ⚠️ Review applies to [a886d516](https://git.openjdk.org/jdk/pull/14123/files/a886d5168c8a95d08c01e218f7763001fc9fbbed)
 * [Tom Rodriguez](https://openjdk.org/census#never) (@tkrodriguez - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14123/head:pull/14123` \
`$ git checkout pull/14123`

`$ git checkout pull/14123` \
`$ git pull https://git.openjdk.org/jdk.git pull/14123/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14123`

`$ git pr show -t 14123`

</details>

<a href="https://git.openjdk.org/jdk/pull/14123.diff">https://git.openjdk.org/jdk/pull/14123.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14123#issuecomment-1562369945)

### DIFF
--- a/src/hotspot/share/opto/graphKit.cpp
+++ b/src/hotspot/share/opto/graphKit.cpp
@@ -2888,6 +2888,7 @@ Node* GraphKit::type_check_receiver(Node* receiver, ciKlass* klass,
       // recv_xtype, since now we know what the type will be.
       Node* cast = new CheckCastPPNode(control(), receiver, recvx_type);
       (*casted_receiver) = _gvn.transform(cast);
+      assert(!(*casted_receiver)->is_top(), "that path should be unreachable");
       // (User must make the replace_in_map call.)
     }
   }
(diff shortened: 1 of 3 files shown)


TobiHartmann (approved):
Looks good to me.

tkrodriguez:
Thanks for the fix.  I did wonder why the type checks themselves weren't properly folding but didn't follow it through all the way.

I think it would be worth including an assert or guarantee in type_check_receiver that it never injects top into the graph.

src/hotspot/share/opto/graphKit.cpp
```diff
@@ -3522,7 +3522,7 @@ Node* GraphKit::get_layout_helper(Node* klass_node, jint& constant_value) {
   const TypeKlassPtr* inst_klass = _gvn.type(klass_node)->isa_klassptr();
   if (!StressReflectiveCode && inst_klass != nullptr) {
     bool    xklass = inst_klass->klass_is_exact();
-    if (xklass || inst_klass->isa_aryklassptr()) {
+    if (xklass || (inst_klass->isa_aryklassptr() && inst_klass->is_aryklassptr()->elem() != Type::BOTTOM)) {
```
vnkozlov, line 3525: Can you rename `inst_klass` to `klass_ptr` because it is general klass. It is weird to check like `inst_klass->isa_aryklassptr()`.

vnkozlov (commented):
In addition to normal mach5 testing we need to run our stress testing too for this changes to run with `StressReflectiveCode` on.

rwestrel:
> I think it would be worth including an assert or guarantee in type_check_receiver that it never injects top into the graph.

Right. I included one in the new commit I just pushed.

src/hotspot/share/opto/graphKit.cpp
rwestrel: I pushed a new commit that includes the rename.

vnkozlov (approved):
Good.

tkrodriguez (approved):

rwestrel:
@TobiHartmann @vnkozlov @tkrodriguez thanks for the reviews

PR #14125, opened 2023-05-24
8308780: Fix the Java Integer types on Windows

TheShermanTanker:
On Windows, the basic Java Integer types are defined as long and __int64 respectively. In particular, the former is rather problematic since it breaks compilation as the Visual C++ becomes stricter and more compliant with every release, which means the way Windows code treats long as a typedef for int is no longer correct, especially with -permissive- enabled. Instead of changing every piece of broken code to match the jint = long typedef, which is far too time consuming, we can instead change jint to an int (which is still the same 32 bit number type as long), as there are far fewer problems caused by this definition. It's better to get this over and done with sooner than later when a future version of Visual C++ finally starts to break on existing code

---------
### Progress

### Issue
 * [JDK-8308780](https://bugs.openjdk.org/browse/JDK-8308780): Fix the Java Integer types on Windows (**Bug** - P4)

### Reviewers
 * [David Holmes](https://openjdk.org/census#dholmes) (@dholmes-ora - **Reviewer**) ⚠️ Review applies to [80b6f787](https://git.openjdk.org/jdk/pull/14125/files/80b6f787982a4772d9359ebe09cba357bfad317c)
 * [Daniel Jeliński](https://openjdk.org/census#djelinski) (@djelinski - **Reviewer**) ⚠️ Review applies to [84f8e08c](https://git.openjdk.org/jdk/pull/14125/files/84f8e08c2ecc90ec50a13406fb99b8cd52f33b7c)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14125/head:pull/14125` \
`$ git checkout pull/14125`

`$ git checkout pull/14125` \
`$ git pull https://git.openjdk.org/jdk.git pull/14125/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14125`

`$ git pr show -t 14125`

</details>

<a href="https://git.openjdk.org/jdk/pull/14125.diff">https://git.openjdk.org/jdk/pull/14125.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14125#issuecomment-1561225340)

### DIFF
--- a/src/java.desktop/windows/native/libawt/java2d/windows/GDIRenderer.cpp
+++ b/src/java.desktop/windows/native/libawt/java2d/windows/GDIRenderer.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1999, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1999, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
--- a/src/java.desktop/windows/native/libawt/java2d/windows/GDIRenderer.cpp
+++ b/src/java.desktop/windows/native/libawt/java2d/windows/GDIRenderer.cpp
@@ -49,7 +49,7 @@ extern "C" {
 
 #define POLYTEMPSIZE    (512 / sizeof(POINT))
 
-static void AngleToCoord(jint angle, jint w, jint h, jint *x, jint *y)
+static void AngleToCoord(int angle, int w, int h, int *x, int *y)
 {
     const double pi = 3.1415926535;
     const double toRadians = 2 * pi / 360;
@@ -322,7 +322,7 @@ Java_sun_java2d_windows_GDIRenderer_doDrawArc
         return;
     }
 
-    jint sx, sy, ex, ey;
+    int sx, sy, ex, ey;
     if (angleExtent >= 360 || angleExtent <= -360) {
         sx = ex = x + w;
         sy = ey = y + h/2;
@@ -602,7 +602,7 @@ Java_sun_java2d_windows_GDIRenderer_doFillArc
     if (wsdo == NULL) {
         return;
     }
-    jint sx, sy, ex, ey;
+    int sx, sy, ex, ey;
     int angleEnd;
     if (angleExtent < 0) {
         angleEnd = angleStart;
--- a/src/java.desktop/windows/native/libawt/java2d/windows/GDIRenderer.cpp
+++ b/src/java.desktop/windows/native/libawt/java2d/windows/GDIRenderer.cpp
@@ -322,7 +322,7 @@ Java_sun_java2d_windows_GDIRenderer_doDrawArc
         return;
     }
 
-    long sx, sy, ex, ey;
+    jint sx, sy, ex, ey;
     if (angleExtent >= 360 || angleExtent <= -360) {
         sx = ex = x + w;
         sy = ey = y + h/2;
@@ -602,7 +602,7 @@ Java_sun_java2d_windows_GDIRenderer_doFillArc
     if (wsdo == NULL) {
         return;
     }
-    long sx, sy, ex, ey;
+    jint sx, sy, ex, ey;
     int angleEnd;
     if (angleExtent < 0) {
         angleEnd = angleStart;
--- a/src/java.desktop/windows/native/libawt/windows/ShellFolder2.cpp
+++ b/src/java.desktop/windows/native/libawt/windows/ShellFolder2.cpp
@@ -1080,8 +1080,8 @@ JNIEXPORT jintArray JNICALL Java_sun_awt_shell_Win32ShellFolder2_getIconBits
             // Extract the color bitmap
             int nBits = iconSize * iconSize;
 
-            jint *colorBits = nullptr;
-            int *maskBits = nullptr;
+            jint *colorBits = NULL;
+            int *maskBits = NULL;
 
             try {
                 entry_point();
--- a/src/java.desktop/windows/native/libawt/windows/ShellFolder2.cpp
+++ b/src/java.desktop/windows/native/libawt/windows/ShellFolder2.cpp
@@ -1086,7 +1086,7 @@ JNIEXPORT jintArray JNICALL Java_sun_awt_shell_Win32ShellFolder2_getIconBits
             try {
                 entry_point();
                 colorBits = (jint*)safe_Malloc(MAX_ICON_SIZE * MAX_ICON_SIZE * sizeof(jint));
-                GetDIBits(dc, iconInfo.hbmColor, 0, iconSize, colorBits, &bmi, DIB_RGB_COLORS);
+                GetDIBits(dc, iconInfo.hbmColor, 0, iconSize, (LPVOID) colorBits, &bmi, DIB_RGB_COLORS);
                 // XP supports alpha in some icons, and depending on device.
                 // This should take precedence over the icon mask bits.
                 BOOL hasAlpha = FALSE;
@@ -1101,7 +1101,7 @@ JNIEXPORT jintArray JNICALL Java_sun_awt_shell_Win32ShellFolder2_getIconBits
                 if (!hasAlpha) {
                     // Extract the mask bitmap
                     maskBits = (int*)safe_Malloc(MAX_ICON_SIZE * MAX_ICON_SIZE * sizeof(int));
-                    GetDIBits(dc, iconInfo.hbmMask, 0, iconSize, maskBits, &bmi, DIB_RGB_COLORS);
+                    GetDIBits(dc, iconInfo.hbmMask, 0, iconSize, (LPVOID) maskBits, &bmi, DIB_RGB_COLORS);
                     // Copy the mask alphas into the color bits
                     for (int i = 0; i < nBits; i++) {
                         if (maskBits[i] == 0) {
--- a/src/java.desktop/windows/native/libawt/windows/ShellFolder2.cpp
+++ b/src/java.desktop/windows/native/libawt/windows/ShellFolder2.cpp
@@ -1086,7 +1086,7 @@ JNIEXPORT jintArray JNICALL Java_sun_awt_shell_Win32ShellFolder2_getIconBits
             try {
                 entry_point();
                 colorBits = (jint*)safe_Malloc(MAX_ICON_SIZE * MAX_ICON_SIZE * sizeof(jint));
-                GetDIBits(dc, iconInfo.hbmColor, 0, iconSize, (LPVOID) colorBits, &bmi, DIB_RGB_COLORS);
+                GetDIBits(dc, iconInfo.hbmColor, 0, iconSize, colorBits, &bmi, DIB_RGB_COLORS);
                 // XP supports alpha in some icons, and depending on device.
                 // This should take precedence over the icon mask bits.
                 BOOL hasAlpha = FALSE;
@@ -1101,7 +1101,7 @@ JNIEXPORT jintArray JNICALL Java_sun_awt_shell_Win32ShellFolder2_getIconBits
                 if (!hasAlpha) {
                     // Extract the mask bitmap
                     maskBits = (int*)safe_Malloc(MAX_ICON_SIZE * MAX_ICON_SIZE * sizeof(int));
-                    GetDIBits(dc, iconInfo.hbmMask, 0, iconSize, (LPVOID) maskBits, &bmi, DIB_RGB_COLORS);
+                    GetDIBits(dc, iconInfo.hbmMask, 0, iconSize, maskBits, &bmi, DIB_RGB_COLORS);
                     // Copy the mask alphas into the color bits
                     for (int i = 0; i < nBits; i++) {
                         if (maskBits[i] == 0) {
--- a/src/java.desktop/windows/native/libawt/windows/ShellFolder2.cpp
+++ b/src/java.desktop/windows/native/libawt/windows/ShellFolder2.cpp
@@ -1080,12 +1080,12 @@ JNIEXPORT jintArray JNICALL Java_sun_awt_shell_Win32ShellFolder2_getIconBits
             // Extract the color bitmap
             int nBits = iconSize * iconSize;
 
-            long *colorBits = NULL;
-            long *maskBits = NULL;
+            jint *colorBits = nullptr;
+            int *maskBits = nullptr;
 
             try {
                 entry_point();
-                colorBits = (long*)safe_Malloc(MAX_ICON_SIZE * MAX_ICON_SIZE * sizeof(long));
+                colorBits = (jint*)safe_Malloc(MAX_ICON_SIZE * MAX_ICON_SIZE * sizeof(jint));
                 GetDIBits(dc, iconInfo.hbmColor, 0, iconSize, colorBits, &bmi, DIB_RGB_COLORS);
                 // XP supports alpha in some icons, and depending on device.
                 // This should take precedence over the icon mask bits.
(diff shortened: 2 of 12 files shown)


TheShermanTanker:
> I think the JNI type definition change is okay.
> 
> However many of the other changes appear to me to not involve Java variables and so don't need to be Java types i.e they should be `int` rather than `jint` - though as these are native Windows types there may not actually be any reason to change them from `long`. This is for the client-libs folk to decide.

All the changes from long were done since there was conversion from or to a jint somewhere down the line, and the compilation would fail if not done otherwise. I also changed them to jint rather than int so there wouldn't be a need to keep the variables in sync with the jni.h declarations, but I guess I'll wait for more reviews to see what to do here

TheShermanTanker:
Going to page for @aivanov-jdk for `client-libs` review

dholmes-ora:
> All the changes from long were done since there was conversion from or to a jint somewhere down the line,

Okay I see that now. It is a messy situation - at some point the incoming jint's need to be "converted" to a native type.

TheShermanTanker:
I'll see what I can do, I'll check the parameter type for the methods that are called in relevant code, though if they take an int as an argument I'm not really sure what to change

TheShermanTanker:
For reference as to what types the calls in affected code accepts as parameters, so any future reviews don't need to dig through the code
```
GDIRenderer.cpp:

- Java_sun_java2d_windows_GDIRenderer_doFillArc
  * AngleToCoord takes jints as arguments
  * ::Pie takes ints as arguments

- Java_sun_java2d_windows_GDIRenderer_doDrawArc
  * AngleToCoord takes jints as arguments
  * ::Arc takes ints as arguments

GDIWindowSurfaceData.cpp:

- GDIWinSD_GetRasInfo
  * SurfaceDataRasInfo's lutBase field is a jint*

awt_MenuBar.cpp (Encompasses the changes in awt_Menu.h and awt_MenuBar.h as well):

- AwtMenuBar::GetItem passes the only relevant jint (formerly long) into env->CallObjectMethod()

AccessInfo.cpp

- getAccessibleInfo
  * start and end (both formerly long) are both passed to GetAccessibleTextLineBounds and GetAccessibleTextRange which take jints
```

Only outlier is jaccesswalker, which I think I may have edited wrongly

TheShermanTanker:
Bumping

:(

TheShermanTanker:
While looking through the code for this I've come to realize that a staggering amount of code in the accessibility binaries specify longs where unsigned longs would be much more appropriate, wonder if this should also be fixed in the long term too

prrace:
I'm not sure I understand the logic here.
I would not want to move to using Java typedefs in places where the windows APIs specify the types they are expecting.
If something comes in from a JNI down-call we should convert it to the type expected by Windows using the name expected by Windows.

Also "compilation" isn't nearly good enough. How is this being tested ?

TheShermanTanker:
> I'm not sure I understand the logic here. I would not want to move to using Java typedefs in places where the windows APIs specify the types they are expecting. If something comes in from a JNI down-call we should convert it to the type expected by Windows using the name expected by Windows.

I can change the jints in this PR to regular ints if required. As listed above, the native Windows API routines that the java.desktop code calls are actually expecting ints, so our existing declarations of passing longs to them are also wrong regardless, even without the Java typedefs

Actually, now that I revisit this issue (shown in the list above), the only actual calls in this change that _don't_ take Java typedefs are the calls to ::Arc and ::Pie, so this is less of a problem than initially expected

> Also "compilation" isn't nearly good enough. How is this being tested?

`-permissive-`

TheShermanTanker:
Bumping

TheShermanTanker:
Anyone?

aivanov-jdk:
I'll take a look… hopefully next week.

TheShermanTanker:
Bumping

src/java.desktop/windows/native/libawt/windows/ShellFolder2.cpp
```diff
@@ -1080,12 +1080,12 @@ JNIEXPORT jintArray JNICALL Java_sun_awt_shell_Win32ShellFolder2_getIconBits
...
-            long *colorBits = NULL;
-            long *maskBits = NULL;
+            jint *colorBits = nullptr;
+            int *maskBits = nullptr;
 
             try {
                 entry_point();
-                colorBits = (long*)safe_Malloc(MAX_ICON_SIZE * MAX_ICON_SIZE * sizeof(long));
+                colorBits = (jint*)safe_Malloc(MAX_ICON_SIZE * MAX_ICON_SIZE * sizeof(jint));
                 GetDIBits(dc, iconInfo.hbmColor, 0, iconSize, colorBits, &bmi, DIB_RGB_COLORS);
```
prrace, line 1089: I just can't tell if the updates you are making as a result of the jni_md.h change are really the right ones and some of them don't look that way

You wrote
"As listed above, the native Windows API routines that the java.desktop code calls are actually expecting ints,"

Per the windows docs for GetDIBits()  it takes an LPVOID for the parameter these are used for.
For the cases above if takes an LPVOID which is really just a void*, and using jint in the malloc is just weird since jint doesn't mean anything to GDI

And  clearly this change requires running a whole load of client tests on windows [*]
but I have no idea what your reply to my question about that means. What is "-permissive" ?

[*] and I have no idea what VM etc tests need to be run just for the JNI change

TheShermanTanker: -permissive- is a compiler switch that forces the Microsoft Visual C compiler to be stricter in compiling C and C++ and makes it enforce the standard much more aggressively. It's becoming less permissive with every iteration of the Microsoft compiler and is stated to become enabled by default eventually by Microsoft. One of the consequences of this is that in the future our code cannot so loosely treat int and long as the same type on Windows (even though they are ultimately the same size in compiled code), as far as the compiler is concerned, they are semantically 2 entirely different types. That's the complication this Pull Request is trying to preempt by changing the jni.h typedef for Windows

I missed the LPVOID change in GetDIBits, but for the other changes I really only followed the existing declarations, the jint for colorBits is because one of the calls below takes a jint as a parameter. Let me know if I should change the declaration regardless, though

TheShermanTanker:
Bumping

TheShermanTanker:
Bumping

TheShermanTanker:
...

djelinski:
Compilation should be a good enough test for the `long` -> `jint` changes. These changes are supposed to address [this difference](https://learn.microsoft.com/en-us/cpp/overview/cpp-conformance-improvements-2019?view=msvc-170#overload-resolution-involving-integral-overloads-and-long-arguments) between MSVC behavior and the C++ standard. When compiled with `-permissive-` or with a compiler that puts more emphasis on standards conformance (like clang), the current code fails to compile.

I verified some of the generated binaries by comparing the results of `dumpbin /all` before and after the change. Most of the time the changes were limited to timestamp, UUID and mangled function names. `Jaccesswalker.exe` had a few more changes because of a changed format string. None of the changed function names in client libs area are externally visible, but there are some observable changes to `c2v` functions exported from jvm.dll.

I had to revert some of the `NULL`->`nullptr` changes to get this to compile; I assume this will be addressed before this PR is merged. Judging by the PR title, these changes don't belong here anyway.

src/jdk.accessibility/windows/native/jaccesswalker/jaccesswalker.cpp
```diff
@@ -543,8 +543,8 @@ void Jaccesswalker::addComponentNodes(long vmID, AccessibleContext context,
     } else {
         char s[LINE_BUFSIZE];
         snprintf( s, sizeof(s),
-            "ERROR calling GetAccessibleContextInfo; vmID = %X, context = %p",
-            vmID, (void*)context );
+            "ERROR calling GetAccessibleContextInfo; vmID = %lX, context = %p",
+            reinterpret_cast<unsigned long>(vmID), (void*)context );
```
djelinski, line 547: do you need this cast? I checked a few compilers and passing a signed long to "%lX" was fine with them.

TheShermanTanker: gcc will crash with a warning about a mismatched format specifier between signed and unsigned if this isn't done, unfortunately

djelinski: Which gcc? This code compiles without warnings:
```
#include <stdio.h>
int main() {
	unsigned long i = 1;
	long j = 2;
	printf("%ld %ld %lx %lx %lu %lu\n", i, j, i, j, i, j);
	return 0;
}
```
```
# gcc -Wall -Wextra -Wformat=2 test.c
# gcc --version
gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

TheShermanTanker: I'm currently running version 13.1, win32 threads

TheShermanTanker: I'll retry again, maybe the warning has changed now

TheShermanTanker: Seems like it doesn't trigger any longer, I'll revert the cast. Thanks for catching this

TheShermanTanker:
Alright, I've addressed the issues brought up by the reviews, including the GetDIBits cast. In general I left everything that passes into a Java call at least once as a jint, and everything else as a native type

src/java.desktop/windows/native/libawt/windows/ShellFolder2.cpp
```diff
@@ -1080,13 +1080,13 @@ JNIEXPORT jintArray JNICALL Java_sun_awt_shell_Win32ShellFolder2_getIconBits
...
-            long *maskBits = NULL;
+            jint *colorBits = nullptr;
+            int *maskBits = nullptr;
 
             try {
                 entry_point();
-                colorBits = (long*)safe_Malloc(MAX_ICON_SIZE * MAX_ICON_SIZE * sizeof(long));
-                GetDIBits(dc, iconInfo.hbmColor, 0, iconSize, colorBits, &bmi, DIB_RGB_COLORS);
+                colorBits = (jint*)safe_Malloc(MAX_ICON_SIZE * MAX_ICON_SIZE * sizeof(jint));
+                GetDIBits(dc, iconInfo.hbmColor, 0, iconSize, (LPVOID) colorBits, &bmi, DIB_RGB_COLORS);
```
djelinski, line 1089: I don't believe casting to LPVOID was what @prrace was asking here.
The function takes a void* parameter because any other pointer type is implicitly convertible to that. We are using long/int/jint because we're asking for 32 bits per pixel (see biBitCount above); if we asked for 24/16/8/any other number of bits per pixel, we would have used a different pointer type.
Please revert these casts. They don't add any value.

djelinski (approved):
/reviewers 2

Thanks. The code changes look good to me now. Some of the files have old copyright, please update them to 2023 before integrating.

TheShermanTanker:
Will do, thanks Daniel
@prrace @dholmes-ora Are both of you happy with the changes?

aivanov-jdk (commented):
Looks fine to me, except for a few comments.

The size of the types for `jint` and `jlong` remains the same after amending the typedef `jni_md.h`. Yet I'm still cautious about it. You should have an approval from hotspot.

Please also update the copyright year in the modified files.

src/java.desktop/windows/native/libawt/java2d/windows/GDIRenderer.cpp
```diff
@@ -322,7 +322,7 @@ Java_sun_java2d_windows_GDIRenderer_doDrawArc
         return;
     }
 
-    long sx, sy, ex, ey;
+    jint sx, sy, ex, ey;
```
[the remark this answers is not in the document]
TheShermanTanker: Resolved, will push soon

src/java.desktop/windows/native/libawt/java2d/windows/GDIRenderer.cpp
```diff
@@ -602,7 +602,7 @@ Java_sun_java2d_windows_GDIRenderer_doFillArc
     if (wsdo == NULL) {
         return;
     }
-    long sx, sy, ex, ey;
+    jint sx, sy, ex, ey;
```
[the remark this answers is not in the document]
TheShermanTanker: Resolve, pushing soon

src/java.desktop/windows/native/libawt/windows/ShellFolder2.cpp
```diff
@@ -1080,12 +1080,12 @@ JNIEXPORT jintArray JNICALL Java_sun_awt_shell_Win32ShellFolder2_getIconBits
             // Extract the color bitmap
             int nBits = iconSize * iconSize;
 
-            long *colorBits = NULL;
-            long *maskBits = NULL;
+            jint *colorBits = nullptr;
+            int *maskBits = nullptr;
```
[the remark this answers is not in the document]
TheShermanTanker: Yes, colorBits is a jint because of that

src/java.desktop/windows/native/libawt/windows/awt_MenuBar.cpp
```diff
@@ -145,7 +145,7 @@ int AwtMenuBar::CountItem(jobject menuBar)
     return nCount;
 }
 
-AwtMenuItem* AwtMenuBar::GetItem(jobject target, long index)
+AwtMenuItem* AwtMenuBar::GetItem(jobject target, jint index)
```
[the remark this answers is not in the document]
TheShermanTanker: I had it as a jint since it upcalls into Java

dholmes-ora (approved):
Hotspot changes still approved.

Other changes seem okay to me.

Thanks

TheShermanTanker:
Alright, I'll integrate once Alexsey approve, anyone else has further objections?

src/java.desktop/windows/native/libawt/windows/awt_Menu.h
```diff
@@ -73,7 +73,7 @@ class AwtMenu : public AwtMenuItem {
 
     /*for multifont menu */
     BOOL IsTopMenu();
-    virtual AwtMenuItem* GetItem(jobject target, long index);
+    virtual AwtMenuItem* GetItem(jobject target, int index);
```
djelinski, line 76: Hi @aivanov-jdk are you OK leaving this inconsistent with the definition?
https://github.com/openjdk/jdk/blob/16b5a91461db1765e2e7596ebaaf1299cec9b0c8/src/java.desktop/windows/native/libawt/windows/awt_Menu.cpp#L261

TheShermanTanker:
Wait a minute, I was right, it was a jint the whole time! Oh well, I'll wait for what @aivanov-jdk has to say

src/java.desktop/windows/native/libawt/windows/awt_Menu.h
aivanov-jdk: The declaration and implementation have to match.

aivanov-jdk: To minimise the number of changes, we can go for using `jint` in `AwtMenu::GetItem`.

What do you thing, @djelinski and @TheShermanTanker?

TheShermanTanker: Hmm, I lean towards jint as I feel it conveys the fact that it is a Java parameter clearer, intuitively to me it makes sense that a Java integer type would still work in a for loop

aivanov-jdk: You're right… it gives a hint it'll be an upcall into Java. Let's go for `jint` then.

I don't think there's a need to change the type of the for-loop variable to `jint`.

TheShermanTanker: Oh no, I didn't mean to change the loop variable, rather that leaving the jint as is should be fine in the for loop

TheShermanTanker:
Alright, waiting for you to do the honours :)

aivanov-jdk (changes requested):

TheShermanTanker:
Alright, done

TheShermanTanker:
@aivanov-jdk Is the final change ok with you?

aivanov-jdk (approved):

TheShermanTanker:
Haha, thanks Alexsey!

PR #14126, opened 2023-05-24
8300491: SymbolLookup::libraryLookup accepts strings with terminators

mcimadamore:
There is a difference in behavior between `System::loadLibrary` and `SymbolLookup::libraryLookup(String)`. While the former catches library names containing NULL chars (because, internally it uses Path/File logic, which reject those), `SymbolLookup` does not. As a result, it is possible to load libraries such as `libGL.so\0foobar`, because the string is passed directly to `dlopen`.

A similar issue arises when we consider `SymbolLookup::find`, as the name is again sent unmodified to `dlopen`. In reality, some filtering *does* happen in that case, as the native code calls `GetStringUTFChars`:

https://docs.oracle.com/en/java/javase/20/docs/specs/jni/functions.html#getstringutfchars

Which encodes the NULL chars in a different form:

https://docs.oracle.com/en/java/javase/20/docs/specs/jni/types.html#modified-utf-8-strings

So, in practice, even if the input is not validated, a call to `find` with one or more NULL chars is unlikely to succeed.

In this patch I have added several checks to make sure that:
* the name of library passed to `libraryLookup(String)` does not contain `\0` and, if so, throw `IllegalArgumentException` (which is already covered by current spec;
* the name of a symbol passed to `SymbolLookup::find` does not contain `\0` and, if so, return empty optional. This required changes on all the lookups we implement.

---------
### Progress

### Issue
 * [JDK-8300491](https://bugs.openjdk.org/browse/JDK-8300491): SymbolLookup::libraryLookup accepts strings with terminators

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14126/head:pull/14126` \
`$ git checkout pull/14126`

`$ git checkout pull/14126` \
`$ git pull https://git.openjdk.org/jdk.git pull/14126/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14126`

`$ git pr show -t 14126`

</details>

<a href="https://git.openjdk.org/jdk/pull/14126.diff">https://git.openjdk.org/jdk/pull/14126.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14126#issuecomment-1561382183)

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/SymbolLookup.java
+++ b/src/java.base/share/classes/java/lang/foreign/SymbolLookup.java
@@ -28,6 +28,7 @@ package java.lang.foreign;
 import jdk.internal.access.JavaLangAccess;
 import jdk.internal.access.SharedSecrets;
 import jdk.internal.foreign.MemorySessionImpl;
+import jdk.internal.foreign.Utils;
 import jdk.internal.javac.PreviewFeature;
 import jdk.internal.loader.BuiltinClassLoader;
 import jdk.internal.loader.NativeLibrary;
@@ -192,6 +193,7 @@ public interface SymbolLookup {
         }
         return name -> {
             Objects.requireNonNull(name);
+            if (Utils.containsNullChars(name)) return Optional.empty();
             JavaLangAccess javaLangAccess = SharedSecrets.getJavaLangAccess();
             // note: ClassLoader::findNative supports a null loader
             long addr = javaLangAccess.findNative(loader, name);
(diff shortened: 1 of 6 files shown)


mcimadamore:
> Do you also need to test on `SymbolLookup` returned from `Linker::defaultLookup`?

Yeah - some test would be better.

PaulSandoz (approved):

PR #14134, opened 2023-05-24
8308803: Improve java/util/UUID/UUIDTest.java

shipilev:
UUID is very important class that is used to track identities of objects in large scale systems. Yet, the coverage in JDK test is disappointing: it tests only 100 of UUID instances per test, which is way too small to detect collisions due to the bad randomness for example.

I have some pending work to improve UUID performance, and tests should be improved. 

The improved test still runs in less than 5 seconds on my laptop.

---------
### Progress

### Issue
 * [JDK-8308803](https://bugs.openjdk.org/browse/JDK-8308803): Improve java/util/UUID/UUIDTest.java

### Reviewers
 * [Jaikiran Pai](https://openjdk.org/census#jpai) (@jaikiran - **Reviewer**) ⚠️ Review applies to [48d6bc5c](https://git.openjdk.org/jdk/pull/14134/files/48d6bc5c531532921be819fa34bfb0bc70384ce2)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14134/head:pull/14134` \
`$ git checkout pull/14134`

`$ git checkout pull/14134` \
`$ git pull https://git.openjdk.org/jdk.git pull/14134/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14134`

`$ git pr show -t 14134`

</details>

<a href="https://git.openjdk.org/jdk/pull/14134.diff">https://git.openjdk.org/jdk/pull/14134.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14134#issuecomment-1561811751)

### DIFF
--- a/test/jdk/java/util/UUID/UUIDTest.java
+++ b/test/jdk/java/util/UUID/UUIDTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2003, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2003, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -22,22 +22,27 @@
  */
 
 /* @test
- * @bug 4173528 5068772 8148936 8196334
+ * @bug 4173528 5068772 8148936 8196334 8308803
  * @summary Unit tests for java.util.UUID
  * @key randomness
- * @run main/othervm -XX:+CompactStrings UUIDTest
- * @run main/othervm -XX:-CompactStrings UUIDTest
+ * @run main/othervm -Xmx1g -XX:+CompactStrings UUIDTest
+ * @run main/othervm -Xmx1g -XX:-CompactStrings UUIDTest
  */
 
 import java.util.*;
+import java.util.stream.IntStream;
 
 public class UUIDTest {
 
-    static Random generator = new Random();
+    // Single UUID instance is ~32 bytes, 1M instances take ~256M in the set
+    private static final int COUNT = 1_000_000;
+
+    static final Random generator = new Random();
 
     public static void main(String[] args) throws Exception {
-        containsTest();
+        negativeTest();
         randomUUIDTest();
+        randomUUIDTest_Multi();
         nameUUIDFromBytesTest();
         stringTest();
         versionTest();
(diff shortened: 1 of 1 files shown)


shipilev:
Any takers? Thanks!

test/jdk/java/util/UUID/UUIDTest.java
```diff
@@ -22,22 +22,27 @@
...
 import java.util.*;
+import java.util.stream.IntStream;
 
 public class UUIDTest {
 
-    static Random generator = new Random();
+    // Single UUID instance is ~32 bytes, 1M instances take ~256M in the set
+    private static final int COUNT = 1_000_000;
+
+    static final Random generator = new Random();
```
jaikiran, line 40: Hello Aleksey, I realize this is unrelated to the changes in this PR, but since we are updating this test, would using `jdk.test.lib.RandomFactory.getRandom()` from test library be a better idea here? That would then print the seed in the logs and could help debug any collisions more easily?

test/jdk/java/util/UUID/UUIDTest.java
```diff
@@ -228,32 +295,39 @@ private static void nodeTest() throws Exception {
...
     private static void hashCodeEqualsTest() throws Exception {
         // If two UUIDs are equal they must have the same hashCode
-        for (int i=0; i<100; i++) {
+        for (int i = 0; i < COUNT; i++) {
             UUID u1 = UUID.randomUUID();
             UUID u2 = UUID.fromString(u1.toString());
-            if (u1.hashCode() != u2.hashCode())
-                throw new Exception("Equal UUIDs with different hashcodes");
+            if (u1.hashCode() != u2.hashCode()) {
+                throw new Exception("Equal UUIDs with different hash codes: " + u1 + " and " + u2);
```
jaikiran, line 316: Perhaps we should even print the hash codes that weren't matching to provide assistance when debugging?

jaikiran:
Overall, this test-only change look OK to me. It updates the test to use a `Set` instead of `List` for "contains" checks, which detect random generated UUID collisions. The other part of this change is increasing the UUID generation count/repeatition and a new test method which generates the UUIDs in parallel and tries to detect any unexpected collisions.

I just have a couple of minor suggestions, which I have noted inline.

test/jdk/java/util/UUID/UUIDTest.java
shipilev: Yes, good idea, added.

test/jdk/java/util/UUID/UUIDTest.java
shipilev: Yes, added.

shipilev:
> Overall, this test-only change look OK to me. 

Thanks for looking at it! I believe I addressed the comments in the revision.

shipilev:
Thanks! Any other reviewers?

test/jdk/java/util/UUID/UUIDTest.java
```diff
@@ -49,56 +57,76 @@ public static void main(String[] args) throws Exception {
...
+        for (int i = 0; i < COUNT; i++) {
+            UUID u = UUID.randomUUID();
+            if (u.version() != 4) {
+                throw new Exception("Bad version: " + u);
+            }
+            if (u.variant() != 2) {
+                throw new Exception("Bad variant: " + u);
+            }
+            if (!set.add(u)) {
+                throw new Exception("UUID collision: " + u);
```
RogerRiggs, line 79: I would be concerned that if this failure was reported, it would be intermittent, hard to track down, and not reproducible.  Without a hook for the generator and the seed, its just going to be noise in the testing.

shipilev: This is a non-practical concern, IMO. By spec, `UUID.randomUUID` is generated from the cryptographically secure random, with >120 bits of randomness, so the collision is extremely unlikely. Collision math involves birthday paradox, but Wikipedia article on UUIDs fortunately gives us the approximated solutions already:  https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions 

Quote: "Thus, the probability to find a duplicate within 103 trillion version-4 UUIDs is one in a billion." 

In other words, finding a collision in this test with 1M UUIDs points to the implementation issue, not a test bug, with a very high probability. In yet another words, if a unit test with 1M UUIDs is able to find a collision, then this is a strong signal that many production systems that assume extremely low collision probability are up for subtle misbehavior.

RogerRiggs: My point was that its probably not practical to test (more than once).  
If it fails, it will be considered just as you propose and disregarded and in the meantime consumes test cycles in each of the test contexts. Either provide more information about the conditions under which it failed or remove it.

shipilev: Sorry, I have trouble following the argument here.

Let me re-iterate: the probability for bona-fide collision is so vanishingly low, the test failure here is a strong signal that something is wrong with the implementation. We can put more guidance in the test comments there, like "This is extremely unlikely to happen. If you see this failing, this highly likely points to the implementation bug, rather than the odd chance."

What I expect to happen when that test fails, is that it prompts the investigation with multiple stress tests to get a better estimate of the actual collision rate. Assuming we actually see a collision, it is likely to be caused by much higher probability error somewhere in the code. In fact, if this test is _actually noisy_ to the point it becomes a testing problem, this already gives us the signal that actual collision rate is many orders of magnitude higher than math predicts, and this becomes an even _stronger_ signal that random UUIDs are seriously broken for practical use.

RogerRiggs: Two thoughts here. 
The random number source (SecureRandom) should have its own tests, UUID has a simple dependency on a generator.  
The test for UUID is that it composes the bits into the UUID correctly.  The randomness of the generator should be factored out.

Second, to raise concern about collisions, then the test should not throw on the first detected collision but complete the cycle and provide the simple stats on the number of collisions per COUNT (1,000,000).

All those 5 second test runs add up.

shipilev: > Two thoughts here. The random number source (SecureRandom) should have its own tests, UUID has a simple dependency on a generator. The test for UUID is that it composes the bits into the UUID correctly. The randomness of the generator should be factored out.

This test verifies the way UUID uses the SecureRandom. Think about this as the less of a unit, and more of the integration test. This would be even more important once things like https://bugs.openjdk.org/browse/JDK-8308804 show up.

> Second, to raise concern about collisions, then the test should not throw on the first detected collision but complete the cycle and provide the simple stats on the number of collisions per COUNT (1,000,000).

All right, that we can do. See new commit.

> All those 5 second test runs add up.

Yes, and it would be sad to waste those 5 seconds on something irrelevant. I would argue that `UUID.randomUUID` breakage would be very unfortunate for the real world systems. Spending 5 seconds per test run on it is a good investment here.

RogerRiggs: We're into the area of diminishing returns.  
I'd expect the implementation bug mentioned above would be found by far fewer iterations. 
Thanks for the test change.

RogerRiggs (approved):
Looks good, thanks for the updates.

shipilev:
Thanks!

/integrate

PR #14135, opened 2023-05-24
8308804: Improve UUID.randomUUID performance with bulk/scalable PRNG access

shipilev:
UUID is the very important class that is used to track identities of objects in large scale systems. On some of our systems, `UUID.randomUUID` takes >1% of total CPU time, and is frequently a scalability bottleneck due to `SecureRandom` synchronization.

The major issue with UUID code itself is that it reads from the single `SecureRandom` instance by 16 bytes. So the heavily contended `SecureRandom` is bashed with very small requests. This also has a chilling effect on other users of `SecureRandom`, when there is a heavy UUID generation traffic.

We can improve this by doing the bulk reads from the backing SecureRandom and possibly striping the reads across many instances of it. 

```
Benchmark               Mode  Cnt  Score   Error   Units

### AArch64 (m6g.4xlarge, Graviton, 16 cores)

# Before
UUIDRandomBench.single  thrpt   15  3.545 ± 0.058  ops/us
UUIDRandomBench.max     thrpt   15  1.832 ± 0.059  ops/us ; negative scaling

# After
UUIDRandomBench.single  thrpt   15  4.823 ± 0.023  ops/us 
UUIDRandomBench.max     thrpt   15  6.561 ± 0.054  ops/us ; positive scaling, ~1.5x

### x86_64 (c6.8xlarge, Xeon, 18 cores)

# Before
UUIDRandomBench.single  thrpt   15  2.710 ± 0.038  ops/us
UUIDRandomBench.max     thrpt   15  1.880 ± 0.029  ops/us  ; negative scaling 

# After
Benchmark                Mode  Cnt  Score   Error   Units
UUIDRandomBench.single  thrpt   15  3.109 ± 0.026  ops/us
UUIDRandomBench.max     thrpt   15  3.561 ± 0.071  ops/us  ; positive scaling, ~1.2x
```

Note that there is still a scalability bottleneck in current default random (`NativePRNG`), because it synchronizes over a singleton instance for SHA1 mixer, then the engine itself, etc. -- it is quite a whack-a-mole to figure out the synchronization story there. The scalability fix in current default `SecureRandom` would be much more intrusive and risky, since it would change a core crypto class with unknown bug fanout.

Using the bulk reads even when the underlying PRNG is heavily synchronized is still a win. A more scalable PRNG would benefit from this as well. This PR adds a system property to select the PRNG implementation, and there we can clearly see the benefit with more scalable PRNG sources:

```
Benchmark               Mode  Cnt   Score   Error   Units

### x86_64 (c6.8xlarge, Xeon, 18 cores)

# Before, hacked `new SecureRandom()` to `SecureRandom.getInstance("SHA1PRNG")`
UUIDRandomBench.single  thrpt   15  3.661 ± 0.008  ops/us
UUIDRandomBench.max     thrpt   15  2.400 ± 0.031  ops/us  ; faster than NativePRNG, but still negative scalability

# After, -Djava.util.UUID.prngName=SHA1PRNG
UUIDRandomBench.single  thrpt   15   3.710 ± 0.079  ops/us
UUIDRandomBench.max     thrpt   15  60.768 ± 1.195  ops/us ; positive scaling, ~16x
```

Other scalable random number providers would improve the similar way. Note that just changing to`SHA1PRNG` right now would not help much, because it would still be very contended. This PR does not change the default PRNG provider, that would need a larger discussion. It only provides the means to select another one.

Since the buffers are allocated on-demand and stay permanently, there are allocation rate improvements: generating an UUID now takes 80 bytes per op instead of 120 bytes per op. The buffer cache also takes memory. Back-envelope: for large 192-core machine that takes UUIDs in all threads, the default settings add up to 768K of additional memory.

Additional testing:
 - [x] Updated tests from #14134 
 - [ ] Linux AArch64 fastdebug `tier1 tier2 tier3`

The new options are not strictly speaking required for this work to be useful, but it would be convenient to have them around for field tuning and diagnostics.

---------
### Progress

### Issue
 * [JDK-8308804](https://bugs.openjdk.org/browse/JDK-8308804): Improve UUID.randomUUID performance with bulk/scalable PRNG access (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14135/head:pull/14135` \
`$ git checkout pull/14135`

`$ git checkout pull/14135` \
`$ git pull https://git.openjdk.org/jdk.git pull/14135/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14135`

`$ git pr show -t 14135`

</details>

<a href="https://git.openjdk.org/jdk/pull/14135.diff">https://git.openjdk.org/jdk/pull/14135.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14135#issuecomment-1562754056)


src/java.base/share/classes/java/util/UUID.java
```diff
@@ -99,24 +104,190 @@ public final class UUID implements java.io.Serializable, Comparable<UUID> {
...
+        static final Buffer[] BUFS;
+
+        private static final int roundPowerOfTwo(int x) {
+            int n = -1 >>> Integer.numberOfLeadingZeros(x - 1);
+            return (n < 0) ? 1 : (n + 1);
+        }
+
+        static {
+            try {
+                PRNG_NAME = System.getProperty(PROP_NAME_PRNG_NAME, null);
```
AlanBateman, line 127: You'll have to use GetPropertyAction.privilegedGetProperty as this will otherwise fail if someone runs with a SM.

src/java.base/share/classes/java/util/UUID.java
```diff
@@ -99,24 +104,190 @@ public final class UUID implements java.io.Serializable, Comparable<UUID> {
...
+                throw new InternalError(e);
+            }
+        }
+
+        static SecureRandom newRandom() {
+            if (PRNG_NAME == null) {
+                return new SecureRandom();
+            } else {
+                try {
+                    return SecureRandom.getInstance(PRNG_NAME);
```
AlanBateman, line 149: Part of the change here is that the RNG algorithm is configurable via a system property. The naming of the naming (java.util.UUID.prngName) hints that you might want it to be a standard system property but maybe it's for testing?

shipilev: Yes, I used them for testing, but left them in, because it would be convenient to have them around for field tuning and diagnostics. This would require a CSR, right?

src/java.base/share/classes/java/util/UUID.java
```diff
@@ -99,24 +104,190 @@ public final class UUID implements java.io.Serializable, Comparable<UUID> {
...
+
+            public UUID next() {
+                long stamp = lock.tryOptimisticRead();
+                try {
+                    UUID uuid = null;
+
+                    // Optimistic path: optimistic locking succeeded.
+                    // Try to pull the UUID from the current buffer at current position.
+                    if (stamp != 0) {
+                        int p = (int)VH_POS.getAndAdd(this, UUID_CHUNK);
```
apangin, line 224: An atomic update together with an optimistic lock looks non-idiomatic use of StampedLock to me. Won't a simple CAS loop be simpler? E.g. in pseudocode:
```
while ((p = this.pos) + 16) < buf.length) {
    long msb = getLong(buf, p);
    long lsb = getLong(buf, p + 8);
    if (cas(this.pos, p, p + 16)) {
        return new UUID(msb, lsb);
    }
}

// refill buffer under lock

File: src/java.base/share/classes/java/util/UUID.java [hunk 4]
```diff
@@ -99,24 +104,190 @@ public final class UUID implements java.io.Serializable, Comparable<UUID> {
...
+                        if (p < BUF_SIZE) {
+                            uuid = new UUID(buf, p);
```
apangin, line 226: We can read msb/lsb from the buffer here and move object allocation outside the lock to reduce the length of the critical section.

src/java.base/share/classes/java/util/UUID.java
```diff
@@ -99,24 +104,190 @@ public final class UUID implements java.io.Serializable, Comparable<UUID> {
...
+                        if (p < BUF_SIZE) {
+                            return new UUID(buf, p);
+                        }
+                    }
+
+                    // Pessimistic path: buffer requires replenishment. Recreate it from the
+                    // provided random, and initialize all UUIDs at once to avoid further
+                    // initializations, and thus false sharing between reader threads.
+                    random.nextBytes(buf);
+                    for (int c = 0; c < BUF_SIZE; c += UUID_CHUNK) {
```
bokken, line 255: I think this could be faster by using a ByteBuffer (or VarHandle) to process as longs.

https://mail.openjdk.org/pipermail/core-libs-dev/2023-March/101249.html

src/java.base/share/classes/java/util/UUID.java
```diff
@@ -99,24 +104,190 @@ public final class UUID implements java.io.Serializable, Comparable<UUID> {
...
      * Private constructor which uses a byte array to construct the new UUID.
      */
-    private UUID(byte[] data) {
+    private UUID(byte[] data, int start) {
         long msb = 0;
         long lsb = 0;
-        assert data.length == 16 : "data must be 16 bytes in length";
-        for (int i=0; i<8; i++)
+        for (int i = start; i < start + 8; i++) {
             msb = (msb << 8) | (data[i] & 0xff);
```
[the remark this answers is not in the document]
SirYwell: `jdk.internal.util.ByteArray` has VarHandle-based methods ready for that

src/java.base/share/classes/java/util/UUID.java
shipilev: Nope, I don't think you cannot do this, because there is a race on concurrently-updating `buf`. The StampedLock (RW lock), protects the buffer-reading threads from seeing the `buf` undergoing the initialization by buffer-writing threads. Atomic pos only arbitrates the concurrent buffer-threads. That atomic does not help to resolve the buf races.

src/java.base/share/classes/java/util/UUID.java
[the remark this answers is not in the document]
shipilev: Yes, I have that optimization in the pipeline, and wanted to do so separately. I can still fold it here, let me see.

src/java.base/share/classes/java/util/UUID.java
```diff
@@ -99,24 +104,190 @@ public final class UUID implements java.io.Serializable, Comparable<UUID> {
...
+                    // Pessimistic path: buffer requires replenishment. Recreate it from the
+                    // provided random, and initialize all UUIDs at once to avoid further
+                    // initializations, and thus false sharing between reader threads.
+                    random.nextBytes(buf);
+                    for (int c = 0; c < BUF_SIZE; c += UUID_CHUNK) {
+                        buf[c + 6] &= 0x0f;  /* clear version        */
+                        buf[c + 6] |= 0x40;  /* set to version 4     */
+                        buf[c + 8] &= 0x3f;  /* clear variant        */
+                        buf[c + 8] |= (byte) 0x80;  /* set to IETF variant  */
+                    }
```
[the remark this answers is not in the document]
shipilev: Right, we can do things straight on locals, without ever touching the heap, let's see...

src/java.base/share/classes/java/util/UUID.java
shipilev: Yup, but that would probably fold into discussion here: https://github.com/openjdk/jdk/pull/14135#discussion_r1206097261

src/java.base/share/classes/java/util/UUID.java
shipilev: Yes, I am doing this.

src/java.base/share/classes/java/util/UUID.java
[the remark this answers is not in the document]
shipilev: Yup, that works beautifully.

src/java.base/share/classes/java/util/UUID.java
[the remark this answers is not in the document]
shipilev: Did so in new commit.

src/java.base/share/classes/java/util/UUID.java
shipilev: Good spot. Should be fixed now.

src/java.base/share/classes/java/util/UUID.java
```diff
@@ -99,9 +107,178 @@ public final class UUID implements java.io.Serializable, Comparable<UUID> {
...
+                }
+                int bufCount = GetIntegerAction.privilegedGetProperty(PROP_NAME_BUF_COUNT,
+                                    Runtime.getRuntime().availableProcessors());
+                if (bufCount < 1) {
+                    throw new IllegalArgumentException(PROP_NAME_BUF_COUNT + " is out of range");
+                }
+                BUFS_COUNT = roundPowerOfTwo(bufCount);
+                BUFS = new Buffer[BUFS_COUNT];
+            } catch (Exception e) {
+                throw new InternalError(e);
```
dfuch, line 144: Would it be better to throw `ExceptionInInitializerError` here?

shipilev: Sure, see new commit.

src/java.base/share/classes/java/util/UUID.java
```diff
@@ -99,9 +107,179 @@ public final class UUID implements java.io.Serializable, Comparable<UUID> {
...
+                this.buf = new byte[BUF_SIZE];
+                this.pos = BUF_SIZE; // trigger re-creation on first use
+            }
+
+            private static UUID fromRandom(long lsb, long msb) {
+                // set version to 3
+                lsb = (lsb & (0xFFFF_FFFF_FFFF_0FFFL)) | 0x0000_0000_0000_4000L;
+                // set variant to IETF
+                msb = (msb & (0x3FFF_FFFF_FFFF_FFFFL)) | 0x8000_0000_0000_0000L;
+                return new UUID(lsb, msb);
```
apangin, line 226: Doesn't `msb` [come first](https://github.com/openjdk/jdk/blob/deddaa6ea61f85e9822982a37dad51ff4310457b/src/java.base/share/classes/java/util/UUID.java#L314)?

src/java.base/share/classes/java/util/UUID.java
apangin: Right, my example suffers from the ABA problem. It could be likely solved by adding "generation" bits to the pos, but this will be a bit ugly. OK, let's stick with the StampedLock.

src/java.base/share/classes/java/util/UUID.java
shipilev: Yes. Swapped.

src/java.base/share/classes/java/util/UUID.java
RogerRiggs: I don't think there is a strong case for adding the system properties. They just add to the surface area that has to be maintained and add very little value except in very rare cases.  If someone needs very special treatment they can build it outside of UUID and use the new UUID(long, long) constructor.

shipilev: That sounds okay to me. It simplifies the whole thing and does not require me to implement more test and have a CSR discussion about the options too :) Removed in the new commit.

src/java.base/share/classes/java/util/UUID.java
```diff
@@ -99,9 +104,141 @@ public final class UUID implements java.io.Serializable, Comparable<UUID> {
...
+                    // Recreate it from the provided random.
+                    random.nextBytes(buf);
+
+                    // Take the UUID from new buffer. We are still under write lock,
+                    // so we know we are the only thread here.
+                    VH_POS.set(this, UUID_CHUNK);
+
+                    long msb = ByteArray.getLong(buf, 0);
+                    long lsb = ByteArray.getLong(buf, 8);
+                    return fromRandom(msb, lsb);
```
bokken, line 234: is there any value in moving this outside of the critical section?

schlosna:
I see this PR was closed out as stale and not integrated. I am curious if there is anything I could do to help push this over the line. As highlighted by @shipilev in the PR description, UUID generation is often a critical component of some systems and contention on SecureRandom can become a scalability bottleneck that I would love to reduce.

Thanks!

PR #14139, opened 2023-05-25
8308341: JNI_GetCreatedJavaVMs returns a partially initialized JVM

dholmes-ora:
We now track the in-progress and completed states of VM creation and only return a VM from JNI_GetCreatedJavaVMs when there is a fully initialized VM.

Testing: 
 - new regression test
 - tiers 1-3 (sanity)

Thanks

---------
### Progress
- [x] Change requires CSR request [JDK-8308816](https://bugs.openjdk.org/browse/JDK-8308816) to be approved

### Issues
 * [JDK-8308341](https://bugs.openjdk.org/browse/JDK-8308341): JNI_GetCreatedJavaVMs returns a partially initialized JVM
 * [JDK-8308816](https://bugs.openjdk.org/browse/JDK-8308816): JNI_GetCreatedJavaVMs returns a partially initialized JVM (**CSR**)

### Reviewers
 * [Johan Sjölen](https://openjdk.org/census#jsjolen) (@jdksjolen - **Reviewer**) ⚠️ Review applies to [8f6aa1f7](https://git.openjdk.org/jdk/pull/14139/files/8f6aa1f74dea4bd409c2cc2114d7ef965f3d1d62)
 * [Gerard Ziemski](https://openjdk.org/census#gziemski) (@gerard-ziemski - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14139/head:pull/14139` \
`$ git checkout pull/14139`

`$ git checkout pull/14139` \
`$ git pull https://git.openjdk.org/jdk.git pull/14139/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14139`

`$ git pr show -t 14139`

</details>

<a href="https://git.openjdk.org/jdk/pull/14139.diff">https://git.openjdk.org/jdk/pull/14139.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14139#issuecomment-1562277962)

### DIFF
--- a/src/hotspot/share/prims/jni.cpp
+++ b/src/hotspot/share/prims/jni.cpp
@@ -3472,7 +3472,14 @@ struct JNINativeInterface_* jni_functions_nocheck() {
 extern const struct JNIInvokeInterface_ jni_InvokeInterface;
 
 // Global invocation API vars
-volatile int vm_created = 0;
+enum VM_Creation_State {
+  NOT_CREATED = 0,
+  IN_PROGRESS,
+  COMPLETE
+};
+
+volatile VM_Creation_State vm_created = NOT_CREATED;
+
 // Indicate whether it is safe to recreate VM. Recreation is only
 // possible after a failed initial creation attempt in some cases.
 volatile int safe_to_recreate_vm = 1;
@@ -3541,7 +3548,7 @@ static jint JNI_CreateJavaVM_inner(JavaVM **vm, void **penv, void *args) {
   // We use Atomic::xchg rather than Atomic::add/dec since on some platforms
   // the add/dec implementations are dependent on whether we are running
   // on a multiprocessor Atomic::xchg does not have this problem.
-  if (Atomic::xchg(&vm_created, 1) == 1) {
+  if (Atomic::xchg(&vm_created, IN_PROGRESS) != NOT_CREATED) {
     return JNI_EEXIST;   // already created, or create attempt in progress
   }
 
@@ -3553,8 +3560,6 @@ static jint JNI_CreateJavaVM_inner(JavaVM **vm, void **penv, void *args) {
   if (Atomic::xchg(&safe_to_recreate_vm, 0) == 0) {
     return JNI_ERR;
   }
-
-  assert(vm_created == 1, "vm_created is true during the creation");
 
   /**
    * Certain errors during initialization are recoverable and do not
@@ -3572,9 +3577,11 @@ static jint JNI_CreateJavaVM_inner(JavaVM **vm, void **penv, void *args) {
   if (result == JNI_OK) {
     JavaThread *thread = JavaThread::current();
     assert(!thread->has_pending_exception(), "should have returned not OK");
-    /* thread is thread_in_vm here */
+    // thread is thread_in_vm here
     *vm = (JavaVM *)(&main_vm);
     *(JNIEnv**)penv = thread->jni_environment();
+    // mark creation complete for other JNI ops
+    Atomic::release_store(&vm_created, COMPLETE);
 
 #if INCLUDE_JVMCI
     if (EnableJVMCI) {
@@ -3637,7 +3644,8 @@ static jint JNI_CreateJavaVM_inner(JavaVM **vm, void **penv, void *args) {
     *(JNIEnv**)penv = 0;
     // reset vm_created last to avoid race condition. Use OrderAccess to
     // control both compiler and architectural-based reordering.
-    Atomic::release_store(&vm_created, 0);
+    assert(vm_created == IN_PROGRESS, "must be");
+    Atomic::release_store(&vm_created, NOT_CREATED);
   }
 
   // Flush stdout and stderr before exit.
@@ -3666,7 +3674,7 @@ _JNI_IMPORT_OR_EXPORT_ jint JNICALL JNI_CreateJavaVM(JavaVM **vm, void **penv, v
 _JNI_IMPORT_OR_EXPORT_ jint JNICALL JNI_GetCreatedJavaVMs(JavaVM **vm_buf, jsize bufLen, jsize *numVMs) {
   HOTSPOT_JNI_GETCREATEDJAVAVMS_ENTRY((void **) vm_buf, bufLen, (uintptr_t *) numVMs);
 
-  if (vm_created == 1) {
+  if (vm_created == COMPLETE) {
     if (numVMs != nullptr) *numVMs = 1;
     if (bufLen > 0)     *vm_buf = (JavaVM *)(&main_vm);
   } else {
@@ -3686,7 +3694,7 @@ static jint JNICALL jni_DestroyJavaVM_inner(JavaVM *vm) {
   jint res = JNI_ERR;
   DT_RETURN_MARK(DestroyJavaVM, jint, (const jint&)res);
 
-  if (vm_created == 0) {
+  if (vm_created != COMPLETE) {
     res = JNI_ERR;
     return res;
   }
@@ -3717,7 +3725,7 @@ static jint JNICALL jni_DestroyJavaVM_inner(JavaVM *vm) {
   ThreadStateTransition::transition_from_native(thread, _thread_in_vm);
   Threads::destroy_vm();
   // Don't bother restoring thread state, VM is gone.
-  vm_created = 0;
+  vm_created = NOT_CREATED;
   return JNI_OK;
 }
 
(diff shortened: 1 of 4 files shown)


test/hotspot/jtreg/runtime/jni/getCreatedJavaVMs/exeGetCreatedJavaVMs.c
```diff
@@ -0,0 +1,104 @@
+/*
```
jdksjolen, line 1: Can we make this test behave even worse than it already does and have you caused a crash with this test before applying your changes? If it has caused a crash, then I think we're OK with the current form of the test.

Two ideas to make it more likely to crash:

1. Have more threads racing
2. Let's say you spawn more threads, is it possible for `printf` to prevent data races as it's MT-Safe? More threads would increase likelihood of contention on the implicit lock, leading to some threads having been significantly slowed down by having to perform a more expensive locking procedure, is my thought process.

jdksjolen (commented):
Hi, over all looks good, but I have some questions regarding the tests and I wonder whether you missed a case.

src/hotspot/share/prims/jni.cpp
```diff
@@ -3927,7 +3937,10 @@ jint JNICALL jni_GetEnv(JavaVM *vm, void **penv, jint version) {
   jint ret = JNI_ERR;
   DT_RETURN_MARK(GetEnv, jint, (const jint&)ret);
 
-  if (vm_created == 0) {
+  // We can be called by native libraries in the JDK during VM
+  // initialization, so only bail-out if something seems very wrong.
+  // Though how would we get here in that case?
+  if (vm_created == NOT_CREATED) {
```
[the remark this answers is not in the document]
dholmes-ora: I guess my comment was not clear enough. :) It is normal to get here while IN_PROGRESS due to calls from the JDK native code during initialization. So we only consider it an error to call then when NOT_CREATED.

src/hotspot/share/prims/jni.cpp
```diff
@@ -3472,7 +3472,14 @@ struct JNINativeInterface_* jni_functions_nocheck() {
 extern const struct JNIInvokeInterface_ jni_InvokeInterface;
 
 // Global invocation API vars
-volatile int vm_created = 0;
+enum VM_Creation_State {
```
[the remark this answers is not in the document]
dholmes-ora: I prefer a basic enum for this (just named integer constants).

test/hotspot/jtreg/runtime/jni/getCreatedJavaVMs/exeGetCreatedJavaVMs.c
dholmes-ora: The test crashes an unpatched VM. We only need the two threads. One will initiate the creation of the VM and the other then tries to attach to it. We want the second thread to call GetCreatedJavaVMs very quickly to make it more likely the VM is not initialized, but not so quickly that GetCreatedJavaVMs reports zero VMs (regardless of the patch). The time it takes the main thread to create the second thread seems to handle that nicely. Of course it depends on number of CPUs etc but on my local machine the test, and the original version from the JBS issue, reliably crashes every time before the fix.

dholmes-ora:
Thanks for looking at this @jdksjolen !

src/hotspot/share/prims/jni.cpp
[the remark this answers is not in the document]
jdksjolen: Sloppy reading from me! That makes sense, and the code doing the work (L3960-L3965) seems to take all precautions necessary to ensure that nothing goes wrong if we're `IN_PROGRESS`.

Just a minor suggestion: "during VM initialization" could be "while VM initialization is in progress" to mirror the language in the enum, but that's up to you.

test/hotspot/jtreg/runtime/jni/getCreatedJavaVMs/exeGetCreatedJavaVMs.c
jdksjolen: Alright, then I'm very happy with this.

jdksjolen (approved):
Thank you for this patch, everything looks good to me )including the build changes).

gerard-ziemski (commented):
Looks fine to me.

I have one question though: why wouldn't it be enough to move `vm_created = 1` from where we had it before, down to where we now have `vm_created = COMPLETE` ?

I'm not 100% sure why we need the intermediate step `IN_PROGRESS`

test/hotspot/jtreg/runtime/jni/getCreatedJavaVMs/TestGetCreatedJavaVMs.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/*
+ * @test
+ * @library /test/lib
+ * @requires os.family == "Linux"
```
gerard-ziemski, line 27: Why is it Linux only?

Does `ProcessTools.createNativeTestProcessBuilder("GetCreatedJavaVMs");` only work on Linux?

dholmes-ora:
> I have one question though: why wouldn't it be enough to move `vm_created = 1` from where we had it before, down to where we now have `vm_created = COMPLETE` ?

Because we need to prevent two threads from concurrently loading and initializing a VM in the same process.

test/hotspot/jtreg/runtime/jni/getCreatedJavaVMs/TestGetCreatedJavaVMs.java
dholmes-ora: It was Linux only primarily because I copied another test. It should work on any platform that uses pthreads - so not Windows. I can expand the makefile logic to build the test code on all platforms.

dholmes-ora:
Thanks for the review @jdksjolen

dholmes-ora:
Thanks for looking at this @gerard-ziemski . The test now runs on all non-Windows platforms.

gerard-ziemski:
> > I have one question though: why wouldn't it be enough to move `vm_created = 1` from where we had it before, down to where we now have `vm_created = COMPLETE` ?
> 
> Because we need to prevent two threads from concurrently loading and initializing a VM in the same process.

Wouldn't it be less complex to use a lock or some "init once" mechanism (pthread_once) instead of a 3 stage atomic field? Do we really need to know whether the process is in the middle of initialization for any reason other than whether it's actually done?

Just to clarify - I'm OK with the fix you have proposed, I was just curious if you considered any other alternatives.

gerard-ziemski (approved):

dholmes-ora:
Thanks for the review @gerard-ziemski 

> Just to clarify - I'm OK with the fix you have proposed, I was just curious if you considered any other alternatives.

Atomics are the simplest cross-platform solution available at this stage of VM init (which is "nothing is initialized yet") - even Atomic use is limited to Atomic::xchg as we can't require any stubs. We have no VM locks available and anything external to the VM (like pthread_once) would require as OS-specific solution in shared code.

dholmes-ora:
I'm aware from https://github.com/openjdk/jdk/pull/14209 that this new test will fail on AIX but as it will need the same fix as being discussed there I'd rather patch it once we know we have a cross-platform solution. I will file a follow up bug for that (unless we can fix it in #14209.)

Thanks for the reviews.

/integrate

gerard-ziemski:
>> Thanks for the review @gerard-ziemski
>> 
>> > Just to clarify - I'm OK with the fix you have proposed, I was just curious if you considered any other alternatives.
>> 
>> Atomics are the simplest cross-platform solution available at this stage of VM init (which is "nothing is initialized yet") - even >Atomic use is limited to Atomic::xchg as we can't require any stubs. We have no VM locks available and anything external to the >VM (like pthread_once) would require as OS-specific solution in shared code.

I guess it's a compromise either way. You have chosen to use an OS independent mechanism, at the cost of exposing the implementation to the outside world, by introducing a new stage (needs CSR). I think, I would have prefer (with what I know at this time) with keeping the implementation details hidden from outside (no need for CSR), at the cost of using platform dependent locks here, or some sort of initialize once mechanism.

dholmes-ora:
> You have chosen to use an OS independent mechanism, at the cost of exposing the implementation to the outside world, by introducing a new stage (needs CSR).

That is not an accurate characterisation of this change. The CSR request is needed because of the change in behaviour it introduces (by only returning a fully initialized VM), and has nothing at all to do with the implementation. We still need to be able to distinguish when VM creation has started (to prevent concurrent attempt) and when it has completed (so GetCreatedJavaVMs can return  a valid VM) - the mechanism by which that is achieved is immaterial.

PR #14142, opened 2023-05-25
8308807: MulticastSocket cannot join IPv4 multicast group when IPv6 is enabled (aix)

deepa181:
DatagramSocket delegates to an inner DatagramSocket object. Irrespective of whether datagramSocket is IPv4 or IPv6, we create an IPv6 datagramChannel as its's delegate. So, This can cause problems with operations like joinGroup. 

On AIX, IPv6 datagramSocket can not join an IPv4 multicast group.

These failures can be fixed by making sure that the delegate created for a datagram socket has the same protocol family. 

Reported Issue : [JDK-8308807](https://bugs.openjdk.org/browse/JDK-8308807)

---------
### Progress

### Error
&nbsp;⚠️ 8308807 is used in problem lists: [test/jdk/ProblemList.txt]

### Issue
 * [JDK-8308807](https://bugs.openjdk.org/browse/JDK-8308807): MulticastSocket cannot join IPv4 multicast group when IPv6 is enabled (aix) (**Bug** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14142/head:pull/14142` \
`$ git checkout pull/14142`

`$ git checkout pull/14142` \
`$ git pull https://git.openjdk.org/jdk.git pull/14142/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14142`

`$ git pr show -t 14142`

</details>

<a href="https://git.openjdk.org/jdk/pull/14142.diff">https://git.openjdk.org/jdk/pull/14142.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14142#issuecomment-1562407930)

### DIFF
--- a/src/java.base/share/classes/sun/nio/ch/SelectorProviderImpl.java
+++ b/src/java.base/share/classes/sun/nio/ch/SelectorProviderImpl.java
@@ -54,6 +54,12 @@ public abstract class SelectorProviderImpl
         return new DatagramChannelImpl(this, /*interruptible*/false);
     }
 
+    public DatagramChannel openUninterruptibleDatagramChannel(ProtocolFamily family) throws IOException {
+    if(family == null)
+       return openUninterruptibleDatagramChannel();
+       return new DatagramChannelImpl(this, family, /*interruptible*/false);
+    }
+
     @Override
     public DatagramChannel openDatagramChannel(ProtocolFamily family) throws IOException {
         return new DatagramChannelImpl(this, family, /*interruptible*/true);
--- a/src/java.base/share/classes/sun/nio/ch/SelectorProviderImpl.java
+++ b/src/java.base/share/classes/sun/nio/ch/SelectorProviderImpl.java
@@ -55,7 +55,7 @@ public abstract class SelectorProviderImpl
     }
 
     public DatagramChannel openUninterruptibleDatagramChannel(ProtocolFamily family) throws IOException {
-    if(family == null)
+    if (family == null)
        return openUninterruptibleDatagramChannel();
        return new DatagramChannelImpl(this, family, /*interruptible*/false);
     }
--- a/src/java.base/share/classes/java/net/DatagramSocket.java
+++ b/src/java.base/share/classes/java/net/DatagramSocket.java
@@ -32,6 +32,7 @@ import java.nio.channels.MulticastChannel;
 import java.util.Objects;
 import java.util.Set;
 import sun.nio.ch.DefaultSelectorProvider;
+import java.net.InetSocketAddress;
 
 /**
  * This class represents a socket for sending and receiving datagram packets.
@@ -1394,6 +1395,10 @@ public class DatagramSocket implements java.io.Closeable {
         boolean multicast = (type == MulticastSocket.class);
         DatagramSocket delegate = null;
         boolean initialized = false;
+        ProtocolFamily family = null;
+        if (bindaddr instanceof InetSocketAddress inetSocketAddress) {
+            family = inetSocketAddress.getAddress() instanceof Inet6Address ? StandardProtocolFamily.INET6 : StandardProtocolFamily.INET;
+        }
         try {
             DatagramSocketImplFactory factory = DatagramSocket.factory;
             if (factory != null) {
@@ -1406,7 +1411,7 @@ public class DatagramSocket implements java.io.Closeable {
             } else {
                 // create NIO adaptor
                 delegate = DefaultSelectorProvider.get()
-                        .openUninterruptibleDatagramChannel()
+                        .openUninterruptibleDatagramChannel(family)
                         .socket();
             }
 


AlanBateman:
I've moved the JBS issue to the right place and fix the description.

The changes proposed here impact all platforms and will require discussion and cleanup. Also can you look at the isXXX methods defined by sun.nio.ch.Net to see how it is handled on other platforms.

msheppar:
at a cursory glance, and without a deep dive to follow any associated call flows into native code, these changes look like the are not aligned with the semantics espoused in the systems newtwork properties, especially  that of the first:

IPv4 / IPv6

    java.net.preferIPv4Stack (default: false)
    If IPv6 is available on the operating system the underlying native socket will be, by default, an IPv6 socket which lets applications connect to, and accept connections from, both IPv4 and IPv6 hosts. However, in the case an application would rather use IPv4 only sockets, then this property can be set to true. The implication is that it will not be possible for the application to communicate with IPv6 only hosts.

    java.net.preferIPv6Addresses (default: false)
    When dealing with a host which has both IPv4 and IPv6 addresses, and if IPv6 is available on the operating system, the default behavior is to prefer using IPv4 addresses over IPv6 ones. This is to ensure backward compatibility: for example, for applications that depend on the representation of an IPv4 address (e.g. 192.168.1.1). This property can be set to true to change that preference and use IPv6 addresses over IPv4 ones where possible, or system to preserve the order of the addresses as returned by the system-wide resolver

deepa181:
Yes, I agree with @AlanBateman  and @msheppar . Even when we try to have an IPv4 multicast socket join an IPv4 multicast group, we still fail because the delegate that is created for an IPv4 multicast socket is IPv4/IPv6 (dual stack) in nature. AIX does not allow dual stack (IPv4/IPv6) sockets to join IPv4 multicast groups.

AlanBateman:
> Yes, I agree with @AlanBateman and @msheppar . Even when we try to have an IPv4 multicast socket join an IPv4 multicast group, we still fail because the delegate that is created for an IPv4 multicast socket is IPv4/IPv6 (dual stack) in nature. AIX does not allow dual stack (IPv4/IPv6) sockets to join IPv4 multicast groups.

The "Platform dependencies" in java.nio.channels.MulticastChannel (implemented by DatagramChannel) makes it clear that it is implementation specific as to whether a channel to an IPv6 socket can join an IPv4 multicast group. The recommendation is to specify the protocol family when creating the channel.  Legacy DatagramSocket/MulticastSocket do not define a constructor that allows the protocol family be specified at xxxSocket create time. That issue has been there since JDK 1.4 when IPv6 support was added. I'm curious why this might be coming up with AIX now.

deepa181:
Here, We are looking at all of the test failures mentioned in the JBS issue right now and trying to fix them . So, @AlanBateman Do you have any suggestion to fix these tests? I would be happy to look into that.

Thanks.

AlanBateman:
@MBaesken @RealCLanger I wonder if you might have a bit of time to comment on this issue.  If I read Deepa's comment correctly, she is saying that IPv6 UDP sockets on AIX cannot join IPv4 multicast groups. The spec doesn't require this but it problematic for MulticastSocket (as it has been since Java 4 when IPv6 was added). The part that I'm wondering about is what this issue might be coming up now. As I understand, SAP have supporting/maintaining/testing on AIX for a long time. Maybe your systems don't have IPv6 enabled so you never run into this?

RealCLanger:
> @MBaesken @RealCLanger I wonder if you might have a bit of time to comment on this issue. If I read Deepa's comment correctly, she is saying that IPv6 UDP sockets on AIX cannot join IPv4 multicast groups. The spec doesn't require this but it problematic for MulticastSocket (as it has been since Java 4 when IPv6 was added). The part that I'm wondering about is why this issue might be coming up now. As I understand, SAP have supporting/maintaining/testing on AIX for a long time. Maybe your systems don't have IPv6 enabled so you never run into this?

Well, in that area we always had deviations on AIX, so definitely this rings a bell. However, since AIX was not the most important platform for us and that particular IPv6/IPv4 multicast feature didn't seem to be mission critical and wouldn't fail TCK tests, we rather resorted to excluding some tests internally, I believe. But it would be good to really resolve this compatibility/spec/documentation issue, if possible.

AlanBateman:
> Well, in that area we always had deviations on AIX, so definitely this rings a bell. However, since AIX was not the most important platform for us and that particular IPv6/IPv4 multicast feature didn't seem to be mission critical and wouldn't fail TCK tests, we rather resorted to excluding some tests internally, I believe. But it would be good to really resolve this compatibility/spec/documentation issue, if possible.

Thanks. So yes, if some/all of the tests for multicasting were excluded then it helps explain why it didn't come up. The spec doesn't require that a DatgaramChannel or DatagramChannel to an IPv6 socket be capable of joining an IPv4 multicast group. So the tests in several areas might need as I don't think we've had a platform to date where it wasn't possible.

deepa181:
> > Well, in that area we always had deviations on AIX, so definitely this rings a bell. However, since AIX was not the most important platform for us and that particular IPv6/IPv4 multicast feature didn't seem to be mission critical and wouldn't fail TCK tests, we rather resorted to excluding some tests internally, I believe. But it would be good to really resolve this compatibility/spec/documentation issue, if possible.
> 
> Thanks. So yes, if some/all of the tests for multicasting were excluded then it helps explain why it didn't come up. The spec doesn't require that a DatgaramSocket or DatagramChannel to an IPv6 socket be capable of joining an IPv4 multicast group. So the tests in several areas might need updates as I don't think we've had a platform to date where it wasn't possible.

@AlanBateman , is it possible if there is a central place where the change could me made for fixing all of the multicast  test cases?

AlanBateman:
> @AlanBateman , is it possible if there is a central place where the change could me made for fixing all of the multicast test cases?

Your starting point is probably to exclude all of the tests that are failing. This will be tests for DatagramSocket/MulticastSocket, DatagramChannel and the tests for the JMX agent that use the discovery protocol. Once they are excluded then you can start working through the issues.

For the JMX agent, then I would expect that if if the system property com.sun.management.jdp.address is configured then the protocol family is known so it should be possible to create the DatagramChannel to use that protocol family. It might so already, but you can check that.

For the networking and NIO tests then you probably should look at jdk.test.lib.NetworkConfiguration. That has a number of AIX specific comments that may be out of date. After that I think you will need to work through the tests one by one. Between NetworkConfiguration and jdk.test.lib.net.IPSupport, I would expect you have all the infrastructure needed for the tests to not attempt to join IPv4 multicast groups when IPSupport.hasIPv6() is true.

pushkarnk:
> > Yes, I agree with @AlanBateman and @msheppar . Even when we try to have an IPv4 multicast socket join an IPv4 multicast group, we still fail because the delegate that is created for an IPv4 multicast socket is IPv4/IPv6 (dual stack) in nature. AIX does not allow dual stack (IPv4/IPv6) sockets to join IPv4 multicast groups.
> 
> The "Platform dependencies" section in the java.nio.channels.MulticastChannel (implemented by DatagramChannel) makes it clear that it is implementation specific as to whether a channel to an IPv6 socket can join an IPv4 multicast group. The recommendation is to specify the protocol family when creating the channel. Legacy DatagramSocket/MulticastSocket do not define a constructor that allows the protocol family be specified at xxxSocket create time. That issue has been there since JDK 1.4 when IPv6 support was added. I'm curious why this might be coming up with AIX now.

IIUC what @deepa181  is saying, it looks like the "IPv4 socket being unable to join an IPv4 multicast group" problems surfaced when we moved away from `PlainDatagramSocketImpl` in JDK 17? Probably because the `delegate` created does not factor in the `ProtocolFamily` and though we started with attempting to have an "IPv4 multicast socket join an IPv4 group", we end up trying to have an "IPv4/IPv6 dual-stack socket (the delegate) join an IPv4 group" which AIX doesn't permit.

dfuch:
> IIUC what @deepa181 is saying, it looks like the "IPv4 socket being unable to join an IPv4 multicast group" problems surfaced when we moved away from PlainDatagramSocketImpl in JDK 17? Probably because the delegate created does not factor in the ProtocolFamily and though we started with attempting to have an "IPv4 multicast socket join an IPv4 group", we end up trying to have an "IPv4/IPv6 dual-stack socket (the delegate) join an IPv4 group" which AIX doesn't permit.

IIRC It has never been possible to create a non-dual `MulticastSocket` / `DatagramSocket` without setting the property `-Djava.net.preferIPv4Stack=true`. The possibility to choose a protocol family is only available with `DatagramChannel`.

AlanBateman:
> IIRC It has never been possible to create a non-dual `MulticastSocket` / `DatagramSocket` without setting the property `-Djava.net.preferIPv4Stack=true`. The possibility to select a protocol family independently of this property is only available with `DatagramChannel`.

Right now, Net.canIPv6SocketJoinIPv4Group() returns false on AIX so it will never attempt to join an IPv4 multicast group when the socket is IPv6.  The old DatagamSocketImpl didn't have an eager check, which makes me wonder what AIX supports and doesn't support, it's not clear.

deepa181:
Hi @AlanBateman , 

 I ran the Jtreg tests on **jdk17** and observed  the following tests passed unexpectedly when I set the **VM_OPTIONS to `-Djdk.net.usePlainDatagramSocketImpl=true`**

1. java/net/MulticastSocket/B6427403.java
2. java/net/MulticastSocket/NoLoopbackPackets.java
3. java/net/MulticastSocket/SetLoopbackMode.java
4. java/net/MulticastSocket/Test.java

I am curious about how setting the `jdk.net.usePlainDatagramSocketImpl` property impacts their behavior. I would like to understand why they pass with this configuration and whether it has any implications.

Could you please share your insights on this matter? I appreciate your assistance in understanding this behavior and ensuring the correctness of the tests.

Thank you.

AlanBateman:
> I am curious about how setting the `jdk.net.usePlainDatagramSocketImpl` property impacts their behavior. I would like to understand why they pass with this configuration and whether it has any implications.
> 
> Could you please share your insights on this matter? I appreciate your assistance in understanding this behavior and ensuring the correctness of the tests.

I think this issue needs an authoritative summary on what AIX supports and doesn't support. The observations reported conflict with some of the comments so I don't think I can say definitely if there are code or test changes required. 

For a long time, Net.canJoin6WithIPv4Group() has returned "false" on AIX so there is no attempt to join IPv4 multicast groups with a dual IPv4/IPv6 socket. DatagramChannel has supported multicasting since JDK 7 (2011) so if there were issues or test failures on AIX then surely they would have been noticed before now, or maybe all the test were excluded, or maybe the testing was done with IPv6 disabled.

The old DatagramSocketImpl implementation didn't have a setting/configuration to say if a dual IPv4/IPv6 socket could join an IPv4 multicast group, it just attempted it. Your results suggests that it works, which is surprising. Does AIX have the equivalent of strace to trace system calls? It might be useful to confirm that the socket is actually IPv6 and that it joining IPv4 multicast groups.

src/java.base/share/classes/sun/nio/ch/SelectorProviderImpl.java
```diff
@@ -54,6 +54,12 @@ public DatagramChannel openUninterruptibleDatagramChannel() throws IOException {
         return new DatagramChannelImpl(this, /*interruptible*/false);
     }
 
+    public DatagramChannel openUninterruptibleDatagramChannel(ProtocolFamily family) throws IOException {
+    if(family == null)
```
turbanoff, line 58:
```suggestion
    if (family == null)
```

TOatGithub:
> > @AlanBateman , is it possible if there is a central place where the change could me made for fixing all of the multicast test cases?
...
> For the networking and NIO tests then you probably should look at jdk.test.lib.NetworkConfiguration. That has a number of AIX specific comments that may be out of date. ...

Created https://bugs.openjdk.org/browse/JDK-8315651 to address jdk.test.lib.NetworkConfiguration.

src/java.base/share/classes/sun/nio/ch/SelectorProviderImpl.java
deepa181: Thank you @turbanoff for highlighting this.

MBaesken:
With this PR added to our build/test queue we noticed a number of additional test failures **cross platform** .
Examples
java/net/MulticastSocket/JoinLeave.java
java/net/MulticastSocket/MulticastAddresses.java

AlanBateman:
I see this PR has been re-based but it has the same changes as before. This PR lacks an authoritative statement on what AIX supports and doesn't supports, I don't think it's possible to make any progress without that.

dfuch:
I concur with Alan that the changes proposed here are not appropriate for DatagramSocket/MulticastSocket.

PR #11411, opened 2022-11-29
8294969: Convert jdk.jdeps javap to use the Classfile API

asotona:
javap uses proprietary com.sun.tools.classfile library to parse class files.

This patch converts javap to use Classfile API.

Please review.

Thanks,
Adam

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8294969](https://bugs.openjdk.org/browse/JDK-8294969): Convert jdk.jdeps javap to use the Classfile API (**Sub-task** - P4)

### Reviewers
 * [Vicente Romero](https://openjdk.org/census#vromero) (@vicente-romero-oracle - **Reviewer**) ⚠️ Review applies to [ae3b25bf](https://git.openjdk.org/jdk/pull/11411/files/ae3b25bf7f11bd2b41f4fd7e06e21e897bbb6006)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/11411/head:pull/11411` \
`$ git checkout pull/11411`

Update a local copy of the PR: \
`$ git checkout pull/11411` \
`$ git pull https://git.openjdk.org/jdk.git pull/11411/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 11411`

View PR using the GUI difftool: \
`$ git pr show -t 11411`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jdk/pull/11411.diff">https://git.openjdk.org/jdk/pull/11411.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/11411#issuecomment-1511385032)

PR #14143, opened 2023-05-25
8308842: Consolidate exceptions thrown from Class-File API

asotona:
Class-File API actually throws wide variety of exceptions based on the actual situation. Complete error handling code must cover `IndexOutOfBoundsException`, `IllegalStateException` and `IllegalArgumentException`. 

Based on previous discussions we decided to consolidate on `IllegalArgumentException` with possible sub-classes. 

It allows easy common error handling in majority of use case, however it allows also to distinguish source of the error when needed (for example `javap` needs to know if the exception comes from constant poll or not). 

Newly introduced `ConstantPoolException` extends `IllegalArgumentException` to indicate the source of the problem is in constant pool. 

Please review.

Thanks,
Adam

---------
### Progress

### Issue
 * [JDK-8308842](https://bugs.openjdk.org/browse/JDK-8308842): Consolidate exceptions thrown from Class-File API

### Reviewers
 * [Brian Goetz](https://openjdk.org/census#briangoetz) (@briangoetz - **Reviewer**) ⚠️ Review applies to [65a7e70a](https://git.openjdk.org/jdk/pull/14143/files/65a7e70ab5de736120605bb5572c6cae8a1f44ac)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14143/head:pull/14143` \
`$ git checkout pull/14143`

`$ git checkout pull/14143` \
`$ git pull https://git.openjdk.org/jdk.git pull/14143/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14143`

`$ git pr show -t 14143`

</details>

<a href="https://git.openjdk.org/jdk/pull/14143.diff">https://git.openjdk.org/jdk/pull/14143.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14143#issuecomment-1562507839)


liach:
What other types of exceptions does javap distinguish, and should Classfile API distinguish them as well?

asotona:
> What other types of exceptions does javap distinguish, and should Classfile API distinguish them as well?

Invalid CP index or content are cases where javap tries to recover, replace the value with '???' and print the rest.
This PR is extracted Classfile API - related part of #11411

liach:
Then this should suffice for now. We can create specific subtypes if we wish to expose specific exceptions in the future.

src/java.base/share/classes/jdk/internal/classfile/CodeBuilder.java
```diff
@@ -386,7 +386,7 @@ default CodeBuilder trying(Consumer<BlockCodeBuilder> tryHandler,
 
         // Check for empty try block
         if (tryBlock.isEmpty()) {
-            throw new IllegalStateException("The body of the try block is empty");
+            throw new IllegalArgumentException("The body of the try block is empty");
```
ExE-Boss, line 389: This exception should be documented in the method’s **Javadoc**.

asotona: Fixed, thanks.

src/java.base/share/classes/jdk/internal/classfile/constantpool/ConstantPoolException.java
```diff
@@ -0,0 +1,52 @@
...
+    @java.io.Serial
+    private static final long serialVersionUID = 7245472922409094120L;
+
+    /**
+     * Constructs a {@code ConstantPoolException} with no detail message.
+     */
+    public ConstantPoolException() {
+        super();
+    }
+
```
briangoetz, line 42: Exception classes in the JDK generally have four constructors: (), (Throwable), (String), and (String, Throwable).

briangoetz (approved):
Reviewed, only minor comment is the one about constructors in the new exception class.

src/java.base/share/classes/jdk/internal/classfile/constantpool/ConstantPoolException.java
asotona: I'll add missing constructors, thanks.

PR #14145, opened 2023-05-25
8232722: G1 archive region deallocation may shrink the heap below -Xms

tschatzl:
Hi all,

  can I have reviews for this change that properly deallocates/uncommits CDS archive regions when failing to load the CDS archive?

In particular this caused the nuisance mentioned in the CR where even if -Xms==-Xmx, g1 uncommitted the heap memory anyway.

Testing: gha, manual testing as below

There is no (existing) way to induce CDS load errors easily, so what I did was adding `-XX:+UseNewCode` in `filemap.cpp:2202` to simulate failures when enabled. Obviously I removed the flag in this change.

Here's the problematic case:

```
$java -XX:GCCardSizeInBytes=128 -Xmx128m -Xms128m -Xlog:gc+region=trace,gc+ergo+heap=debug -XX:+UseNewCode -version

[0.050s][trace][gc,region   ] G1HR ALLOC(OLD) [0x00000000ffe00000, 0x00000000fff00000, 0x00000000fff00000]
[0.050s][trace][gc,region   ] G1HR ALLOC(OLD) [0x00000000fff00000, 0x00000000fff06278, 0x0000000100000000]
[0.050s][trace][gc,region   ] G1HR INACTIVE(FREE) [0x00000000ffe00000, 0x00000000ffe00000, 0x00000000fff00000]
[0.050s][trace][gc,region   ] G1HR INACTIVE(FREE) [0x00000000fff00000, 0x00000000fff00000, 0x0000000100000000]
[0.050s][debug][gc,ergo,heap] Attempt heap shrinking (CDS archive regions). Total size: 2097152B
[0.050s][trace][gc,region   ] G1HR UNCOMMIT(FREE) [0x00000000ffe00000, 0x00000000ffe00000, 0x00000000fff00000]
[0.050s][trace][gc,region   ] G1HR UNCOMMIT(FREE) [0x00000000fff00000, 0x00000000fff00000, 0x0000000100000000]
[0.057s][trace][gc,region   ] G1HR ALLOC(EDEN) [0x00000000ffd00000, 0x00000000ffd00000, 0x00000000ffe00000]
[0.129s][trace][gc,region   ] G1HR ALLOC(EDEN) [0x00000000ffc00000, 0x00000000ffc00000, 0x00000000ffd00000]
openjdk version "21-internal" 2023-09-19
```

(The `GCCardSizeInBytes` option is there to decrease the minimum heap alignment to 512kb/1M so that setting `-Xms` to an odd value in a later test works)

I.e. the CDS regions are unconditionally uncommitted even through `-Xms == -Xmx`.

The next case just illustrates current (pre-existing) behavior with `-Xms != -Xmx`, showing that CDS regions are always committed, leading to higher than `-Xms` memory usage. I will file an enhancement here, as it is acceptable behavior (to me).

```
$ java -XX:GCCardSizeInBytes=128 -Xmx128m -Xms126m -Xlog:gc+region=trace,gc+ergo+heap=debug -XX:+UseNewCode -version

[0.048s][trace][gc,region   ] G1HR COMMIT(FREE) [0x00000000fff00000, 0x00000000fff00000, 0x0000000100000000]
[0.048s][trace][gc,region   ] G1HR ACTIVE(FREE) [0x00000000fff00000, 0x00000000fff00000, 0x0000000100000000]
[0.048s][debug][gc,ergo,heap] Allocate CDS archive regions. Allocated 2 Committed 2
[0.048s][trace][gc,region   ] G1HR ALLOC(OLD) [0x00000000ffe00000, 0x00000000fff00000, 0x00000000fff00000]
[0.048s][trace][gc,region   ] G1HR ALLOC(OLD) [0x00000000fff00000, 0x00000000fff05f50, 0x0000000100000000]
[0.063s][trace][gc,region   ] G1HR ALLOC(EDEN) [0x00000000ffd00000, 0x00000000ffd00000, 0x00000000ffe00000]
```

I.e. CDS allocation just commits the last few regions without caring about other existing mappings to not fragment the heap.

The next case is the new behavior:
```
java -XX:GCCardSizeInBytes=128 -Xmx128m -Xms128m -Xlog:gc+region=trace,gc+ergo+heap=debug -XX:+UseNewCode -version

[0.047s][trace][gc,region   ] G1HR ACTIVE(FREE) [0x00000000ffe00000, 0x00000000ffe00000, 0x00000000fff00000]
[0.047s][trace][gc,region   ] G1HR ACTIVE(FREE) [0x00000000fff00000, 0x00000000fff00000, 0x0000000100000000]
[0.048s][debug][gc,ergo,heap] Allocate CDS archive regions. Allocated 2 Committed 0
[0.048s][trace][gc,region   ] G1HR ALLOC(OLD) [0x00000000ffe00000, 0x00000000fff00000, 0x00000000fff00000]
[0.048s][trace][gc,region   ] G1HR ALLOC(OLD) [0x00000000fff00000, 0x00000000fff05f50, 0x0000000100000000]
[0.049s][trace][gc,region   ] G1HR CLEANUP(FREE) [0x00000000ffe00000, 0x00000000ffe00000, 0x00000000fff00000]
[0.050s][trace][gc,region   ] G1HR CLEANUP(FREE) [0x00000000fff00000, 0x00000000fff00000, 0x0000000100000000]
[0.050s][debug][gc,ergo,heap] Deallocate CDS archive regions. Freed 2 Uncommitted 0 regions.
[0.057s][trace][gc,region   ] G1HR ALLOC(EDEN) [0x00000000fff00000, 0x00000000fff00000, 0x0000000100000000]
[0.131s][trace][gc,region   ] G1HR ALLOC(EDEN) [0x00000000ffe00000, 0x00000000ffe00000, 0x00000000fff00000]
openjdk version "21-internal" 2023-09-19
```
G1 does not uncommit the region blindly any more.

The next test case shows that if `-Xms` is lower than `-Xmx`, the code still properly uncommits (and no only frees the region).
```
java -XX:GCCardSizeInBytes=128 -Xmx128m -Xms126m -Xlog:gc+region=trace,gc+ergo+heap=debug -XX:+UseNewCode -version

[0.048s][trace][gc,region   ] G1HR COMMIT(FREE) [0x00000000ffe00000, 0x00000000ffe00000, 0x00000000fff00000]
[0.048s][trace][gc,region   ] G1HR ACTIVE(FREE) [0x00000000ffe00000, 0x00000000ffe00000, 0x00000000fff00000]
[0.048s][trace][gc,region   ] G1HR COMMIT(FREE) [0x00000000fff00000, 0x00000000fff00000, 0x0000000100000000]
[0.048s][trace][gc,region   ] G1HR ACTIVE(FREE) [0x00000000fff00000, 0x00000000fff00000, 0x0000000100000000]
[0.048s][debug][gc,ergo,heap] Allocate CDS archive regions. Allocated 2 Committed 2
[0.048s][trace][gc,region   ] G1HR ALLOC(OLD) [0x00000000ffe00000, 0x00000000fff00000, 0x00000000fff00000]
[0.048s][trace][gc,region   ] G1HR ALLOC(OLD) [0x00000000fff00000, 0x00000000fff05f50, 0x0000000100000000]
[0.048s][trace][gc,region   ] G1HR INACTIVE(FREE) [0x00000000ffe00000, 0x00000000ffe00000, 0x00000000fff00000]
[0.049s][trace][gc,region   ] G1HR INACTIVE(FREE) [0x00000000fff00000, 0x00000000fff00000, 0x0000000100000000]
[0.049s][debug][gc,ergo,heap] Deallocate CDS archive regions. Freed 2 Uncommitted 2 regions.
[0.049s][trace][gc,region   ] G1HR UNCOMMIT(FREE) [0x00000000ffe00000, 0x00000000ffe00000, 0x00000000fff00000]
[0.049s][trace][gc,region   ] G1HR UNCOMMIT(FREE) [0x00000000fff00000, 0x00000000fff00000, 0x0000000100000000]
[0.056s][trace][gc,region   ] G1HR ALLOC(EDEN) [0x00000000ffd00000, 0x00000000ffd00000, 0x00000000ffe00000]
[0.124s][trace][gc,region   ] G1HR ALLOC(EDEN) [0x00000000ffc00000, 0x00000000ffc00000, 0x00000000ffd00000]
openjdk version "21-internal" 2023-09-19
```

And finally a "mixed" case where to allocate the CDS archive G1 uses a mix of already allocated regions and committed regions, and uncommits/gives back regions as expected.

```
java -XX:GCCardSizeInBytes=128 -Xmx128m -Xms127m -Xlog:gc+region=trace,gc+ergo+heap=debug -XX:+UseNewCode -version

[0.053s][trace][gc,region   ] G1HR COMMIT(FREE) [0x00000000fff00000, 0x00000000fff00000, 0x0000000100000000]
[0.053s][trace][gc,region   ] G1HR ACTIVE(FREE) [0x00000000fff00000, 0x00000000fff00000, 0x0000000100000000]
[0.053s][debug][gc,ergo,heap] Allocate CDS archive regions. Allocated 2 Committed 1
[0.053s][trace][gc,region   ] G1HR ALLOC(OLD) [0x00000000ffe00000, 0x00000000fff00000, 0x00000000fff00000]
[0.053s][trace][gc,region   ] G1HR ALLOC(OLD) [0x00000000fff00000, 0x00000000fff05f50, 0x0000000100000000]
[0.053s][trace][gc,region   ] G1HR INACTIVE(FREE) [0x00000000ffe00000, 0x00000000ffe00000, 0x00000000fff00000]
[0.054s][trace][gc,region   ] G1HR CLEANUP(FREE) [0x00000000fff00000, 0x00000000fff00000, 0x0000000100000000]
[0.054s][debug][gc,ergo,heap] Deallocate CDS archive regions. Freed 2 Uncommitted 1 regions.
[0.054s][trace][gc,region   ] G1HR UNCOMMIT(FREE) [0x00000000ffe00000, 0x00000000ffe00000, 0x00000000fff00000]
[0.061s][trace][gc,region   ] G1HR ALLOC(EDEN) [0x00000000fff00000, 0x00000000fff00000, 0x0000000100000000]
[0.138s][trace][gc,region   ] G1HR ALLOC(EDEN) [0x00000000ffd00000, 0x00000000ffd00000, 0x00000000ffe00000]
openjdk version "21-internal" 2023-09-19
```

Thanks,
  Thomas

---------
### Progress

### Issue
 * [JDK-8232722](https://bugs.openjdk.org/browse/JDK-8232722): G1 archive region deallocation may shrink the heap below -Xms

### Reviewers
 * [Ivan Walulya](https://openjdk.org/census#iwalulya) (@walulyai - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14145/head:pull/14145` \
`$ git checkout pull/14145`

`$ git checkout pull/14145` \
`$ git pull https://git.openjdk.org/jdk.git pull/14145/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14145`

`$ git pr show -t 14145`

</details>

<a href="https://git.openjdk.org/jdk/pull/14145.diff">https://git.openjdk.org/jdk/pull/14145.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14145#issuecomment-1562614745)

### DIFF
--- a/src/hotspot/share/gc/g1/g1CollectedHeap.cpp
+++ b/src/hotspot/share/gc/g1/g1CollectedHeap.cpp
@@ -532,7 +532,7 @@ void G1CollectedHeap::iterate_regions_in_range(MemRegion range, const Func& func
   }
 }
 
-bool G1CollectedHeap::alloc_archive_regions(MemRegion range) {
+bool G1CollectedHeap::alloc_archive_regions(MemRegion range, uint& regions_committed) {
   assert(!is_init_completed(), "Expect to be called at JVM init time");
   MutexLocker x(Heap_lock);
 
@@ -547,7 +547,6 @@ bool G1CollectedHeap::alloc_archive_regions(MemRegion range) {
   HeapWord* start_address = range.start();
   size_t word_size = range.word_size();
   HeapWord* last_address = range.last();
-  size_t commits = 0;
 
   guarantee(reserved.contains(start_address) && reserved.contains(last_address),
             "MemRegion outside of heap [" PTR_FORMAT ", " PTR_FORMAT "]",
@@ -555,15 +554,16 @@ bool G1CollectedHeap::alloc_archive_regions(MemRegion range) {
 
   // Perform the actual region allocation, exiting if it fails.
   // Then note how much new space we have allocated.
-  if (!_hrm.allocate_containing_regions(range, &commits, workers())) {
+  regions_committed = 0;
+  uint regions_allocated = 0;
+  if (!_hrm.allocate_containing_regions(range, &regions_committed, &regions_allocated, workers())) {
     return false;
   }
   increase_used(word_size * HeapWordSize);
-  if (commits != 0) {
-    log_debug(gc, ergo, heap)("Attempt heap expansion (allocate archive regions). Total size: " SIZE_FORMAT "B",
-                              HeapRegion::GrainWords * HeapWordSize * commits);
 
-  }
+  log_debug(gc, ergo, heap)("Allocate CDS archive regions. Allocated %u Committed %u",
+                            regions_allocated, regions_committed);
+
 
   // Mark each G1 region touched by the range as old, add it to
   // the old set, and set top.
@@ -591,11 +591,11 @@ void G1CollectedHeap::populate_archive_regions_bot_part(MemRegion range) {
                            });
 }
 
-void G1CollectedHeap::dealloc_archive_regions(MemRegion range) {
+void G1CollectedHeap::dealloc_archive_regions(MemRegion range, uint regions_committed) {
   assert(!is_init_completed(), "Expect to be called at JVM init time");
   MemRegion reserved = _hrm.reserved();
   size_t size_used = 0;
-  uint shrink_count = 0;
+  uint regions_freed = 0;
 
   // Free the G1 regions that are within the specified range.
   MutexLocker x(Heap_lock);
@@ -608,20 +608,26 @@ void G1CollectedHeap::dealloc_archive_regions(MemRegion range) {
   size_used += range.byte_size();
 
   // Free, empty and uncommit regions with CDS archive content.
+  uint shrink_count = 0;
   auto dealloc_archive_region = [&] (HeapRegion* r, bool is_last) {
     guarantee(r->is_old(), "Expected old region at index %u", r->hrm_index());
     _old_set.remove(r);
-    r->set_free();
-    r->set_top(r->bottom());
-    _hrm.shrink_at(r->hrm_index(), 1);
-    shrink_count++;
+    free_region(r, nullptr);
+    regions_freed++;
+    if (regions_committed > shrink_count) {
+      _hrm.shrink_at(r->hrm_index(), 1);
+      shrink_count++;
+    } else {
+      _hrm.insert_into_free_list(r);
+      hr_printer()->cleanup(r);
+    }
   };
 
   iterate_regions_in_range(range, dealloc_archive_region);
 
-  if (shrink_count != 0) {
-    log_debug(gc, ergo, heap)("Attempt heap shrinking (CDS archive regions). Total size: " SIZE_FORMAT "B",
-                              HeapRegion::GrainWords * HeapWordSize * shrink_count);
+  log_debug(gc, ergo, heap)("Deallocate CDS archive regions. Freed %u Uncommitted %u regions.",
+                            regions_freed, regions_committed);
+  if (regions_committed != 0) {
     // Explicit uncommit.
     uncommit_regions(shrink_count);
   }
--- a/src/hotspot/share/gc/g1/g1CollectedHeap.cpp
+++ b/src/hotspot/share/gc/g1/g1CollectedHeap.cpp
@@ -607,7 +607,7 @@ void G1CollectedHeap::dealloc_archive_regions(MemRegion range, uint regions_comm
          p2i(start_address), p2i(last_address));
   size_used += range.byte_size();
 
-  // Free, empty and uncommit regions with CDS archive content.
+  // Free, empty and optionally uncommit regions with CDS archive content.
   uint shrink_count = 0;
   auto dealloc_archive_region = [&] (HeapRegion* r, bool is_last) {
     guarantee(r->is_old(), "Expected old region at index %u", r->hrm_index());
--- a/src/hotspot/share/gc/g1/heapRegionManager.cpp
+++ b/src/hotspot/share/gc/g1/heapRegionManager.cpp
@@ -551,8 +551,9 @@ uint HeapRegionManager::find_highest_free(bool* expanded) {
   return G1_NO_HRM_INDEX;
 }
 
-bool HeapRegionManager::allocate_containing_regions(MemRegion range, size_t* commit_count, WorkerThreads* pretouch_workers) {
-  size_t commits = 0;
+bool HeapRegionManager::allocate_containing_regions(MemRegion range, uint* commit_count, uint* allocate_count, WorkerThreads* pretouch_workers) {
+  uint commits = 0;
+  uint allocated = 0;
   uint start_index = (uint)_regions.get_index_by_address(range.start());
   uint last_index = (uint)_regions.get_index_by_address(range.last());
 
@@ -567,10 +568,12 @@ bool HeapRegionManager::allocate_containing_regions(MemRegion range, size_t* com
     if (!curr_region->is_free()) {
       return false;
     }
+    allocated++;
   }
 
   allocate_free_regions_starting_at(start_index, (last_index - start_index) + 1);
   *commit_count = commits;
+  *allocate_count = allocated;
   return true;
 }
 
--- a/src/hotspot/share/cds/filemap.cpp
+++ b/src/hotspot/share/cds/filemap.cpp
@@ -2199,7 +2199,7 @@ bool FileMapInfo::map_heap_region_impl() {
   char* base = os::map_memory(_fd, _full_path, r->file_offset(),
                               addr, _mapped_heap_memregion.byte_size(), r->read_only(),
                               r->allow_exec());
-  if (UseNewCode || base == nullptr || base != addr) {
+  if (base == nullptr || base != addr) {
     dealloc_heap_region(regions_committed);
     log_info(cds)("UseSharedSpaces: Unable to map at required address in java heap. "
                   INTPTR_FORMAT ", size = " SIZE_FORMAT " bytes",
--- a/src/hotspot/share/cds/filemap.cpp
+++ b/src/hotspot/share/cds/filemap.cpp
@@ -2187,7 +2187,8 @@ bool FileMapInfo::map_heap_region_impl() {
   }
 
   // allocate from java heap
-  if (!G1CollectedHeap::heap()->alloc_archive_regions(_mapped_heap_memregion)) {
+  uint regions_committed;
+  if (!G1CollectedHeap::heap()->alloc_archive_regions(_mapped_heap_memregion, regions_committed)) {
     log_info(cds)("Unable to allocate region, java heap range is already in use.");
     return false;
   }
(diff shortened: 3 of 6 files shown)


src/hotspot/share/gc/g1/heapRegionManager.cpp
```diff
@@ -567,10 +568,12 @@ bool HeapRegionManager::allocate_containing_regions(MemRegion range, size_t* com
     if (!curr_region->is_free()) {
```
walulyai, line 568: What happens with regions committed before this false return?

walulyai (approved):
Lgtm!

src/hotspot/share/gc/g1/heapRegionManager.cpp
tschatzl: I filed https://bugs.openjdk.org/browse/JDK-8309117.

ashu-mehra:
@tschatzl what if the regions allocated for archive objects are not uncommitted but just added to the free list in  `dealloc_archive_regions`? That would also prevent heap going below Xms and would avoid tracking committed regions.
Do you see any issue if the regions are not uncommitted?

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -547,23 +547,23 @@ bool G1CollectedHeap::alloc_archive_regions(MemRegion range) {
...
 
   guarantee(reserved.contains(start_address) && reserved.contains(last_address),
             "MemRegion outside of heap [" PTR_FORMAT ", " PTR_FORMAT "]",
             p2i(start_address), p2i(last_address));
 
   // Perform the actual region allocation, exiting if it fails.
   // Then note how much new space we have allocated.
-  if (!_hrm.allocate_containing_regions(range, &commits, workers())) {
+  regions_committed = 0;
+  uint regions_allocated = 0;
```
albertnetymk, line 558: Why is alloc-count required? Isn't it sth like `ceil(range/region_size)`?

tschatzl: This adds implicit assumptions about how the range looks like, i.e. the start address; I think it is nicer if the one allocating the memory returns this value like any other information about the allocation.

tschatzl:
> @tschatzl what if the regions allocated for archive objects are not uncommitted but just added to the free list in `dealloc_archive_regions`? That would also prevent heap going below Xms and would avoid tracking committed regions. Do you see any issue if the regions are not uncommitted?

No real issue, but no real advantage either. In case of -Xms != -Xmx this would mean that the `main` method will not start with heap size of -Xms, but additional regions.
Given that CDS is enabled by default, this would effectively mean that unless disabled, the VM would always (effectively) start (at `main`) with an initial heap size a bit larger than requested.

This is allowed (initial heap size is somewhat of a hint), but unexpected I assume.

The question is, should G1 try to honor requested initial heap size or not? (Similar to https://bugs.openjdk.org/browse/JDK-8308854)

src/hotspot/share/gc/g1/heapRegionManager.cpp
tschatzl: Actually I checked/tested again, and the result is that these unnecessarily committed regions will stay around as free regions. No other error or something. I adapted the CR.

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
albertnetymk: I think the closure below, `set_region_to_old`, already assumes the start-addr is region-start. In fact, I believe that is an implicit assumption shared by `alloc_archive_regions`, `dealloc_archive_regions`, and `populate_archive_regions_bot_part`.

I'd prefer making it explicit, but YMMV.

albertnetymk (approved):

ashu-mehra:
> No real issue, but no real advantage either. In case of -Xms != -Xmx this would mean that the main method will not start with heap size of -Xms, but additional regions.

Right, that's the current behavior as well and not uncommitting the regions in `dealloc_archive_regions` doesn't change that. The difference appears only when the archive region mapping fails. If the regions are uncommitted it would bring back heap size to Xms. If not, heap size will remain slightly above Xms, which is same as the current behavior when mapping succeeds.

The advantage is avoiding the need to track committed regions which avoids leaks more G1 specific details to CDS code. Adding GC policy specific details would make it harder to implement https://bugs.openjdk.org/browse/JDK-8296263. I am trying to move G1 specific code out of CDS in https://github.com/openjdk/jdk/pull/14208.
So if not uncommitting simplifies the things then why not?

tschatzl:
> > No real issue, but no real advantage either. In case of -Xms != -Xmx this would mean that the main method will not start with heap size of -Xms, but additional regions.
> 
> Right, that's the current behavior as well and not uncommitting the regions in `dealloc_archive_regions` doesn't change that. The difference appears only when the archive region mapping fails. If the regions are uncommitted it would bring back heap size to Xms. If not, heap size will remain slightly above Xms, which is same as the current behavior when mapping succeeds.

I do not think the current behavior is intended or good. It is at most tolerable. At least it has not been specified as correct in any documentation I could find, and it seems to counter to how initial heap size behaves with other collectors afaik (they just load the data into wherever it lands in old gen and fixes up references).

Hence filing https://bugs.openjdk.org/browse/JDK-8308854.

> The advantage is avoiding the need to track committed regions which avoids leaks more G1 specific details to CDS code. Adding GC policy specific details would make it harder to implement https://bugs.openjdk.org/browse/JDK-8296263. I am trying to move G1 specific code out of CDS in #14208. So if not uncommitting simplifies the things then why not?

That is a strange argument given that it causes some (minor) observable difference in behavior between collectors too. Having collectors behave the same helps reasoning about it.

I will look at the PR later today, haven't noticed it before.

tschatzl:
Maybe it is better to make this change differently, and combine with https://bugs.openjdk.org/browse/JDK-8308854: archive allocation removes/uncommits regions from the heap first, then allocates the new memory. This keeps -Xms. Later deallocation simply frees that memory.

What do you think?

ashu-mehra:
> Maybe it is better to make this change differently, and combine with https://bugs.openjdk.org/browse/JDK-8308854: archive allocation removes/uncommits regions from the heap first, then allocates the new memory. This keeps -Xms. Later deallocation simply frees that memory.

+1 for this approach.

tschatzl:
> > Maybe it is better to make this change differently, and combine with https://bugs.openjdk.org/browse/JDK-8308854: archive allocation removes/uncommits regions from the heap first, then allocates the new memory. This keeps -Xms. Later deallocation simply frees that memory.
> 
> +1 for this approach.

Retracting this change to try the other one.

PR #14146, opened 2023-05-25
JDK-8308288: Fix xlc17 clang warnings in shared code

JoKern65:
When using the new xlc17 compiler (based on a recent clang) to build OpenJDk on AIX , we run into various "warnings as errors".
Some of those are in shared codebase and could be addressed by small adjustments.
A lot of those changes are in hotspot, some might be somewhere else in the OpenJDK C/C++ code.

---------
### Progress

### Issue
 * [JDK-8308288](https://bugs.openjdk.org/browse/JDK-8308288): Fix xlc17 clang warnings in shared code

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14146/head:pull/14146` \
`$ git checkout pull/14146`

`$ git checkout pull/14146` \
`$ git pull https://git.openjdk.org/jdk.git pull/14146/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14146`

`$ git pr show -t 14146`

</details>

<a href="https://git.openjdk.org/jdk/pull/14146.diff">https://git.openjdk.org/jdk/pull/14146.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14146#issuecomment-1562582887)

### DIFF
--- a/src/hotspot/share/runtime/javaThread.cpp
+++ b/src/hotspot/share/runtime/javaThread.cpp
@@ -109,6 +109,10 @@
 #include "jfr/jfr.hpp"
 #endif
 
+#ifdef _AIX
+#include <alloca.h>
+#endif
+
 // Set by os layer.
 size_t      JavaThread::_stack_size_at_create = 0;
 
--- a/src/hotspot/share/utilities/globalDefinitions_xlc.hpp
+++ b/src/hotspot/share/utilities/globalDefinitions_xlc.hpp
@@ -38,6 +38,14 @@
 #include <stddef.h>
 #include <stdio.h>
 #include <stdlib.h>
+// Until IBM modifies its stdlib.h header we repair the critical lines ourself
+// replacing the #define malloc vec_malloc by the modified malloc Function declaration.
+#if (defined(__VEC__) || defined(__AIXVEC)) && defined(AIX) \
+    && defined(__open_xl_version__) && __open_xl_version__ >= 17
+  #undef malloc
+  extern void	*malloc(size_t)  asm("vec_malloc");
+#endif
+
 #include <wchar.h>
 
 #include <math.h>
--- a/src/hotspot/share/utilities/globalDefinitions_xlc.hpp
+++ b/src/hotspot/share/utilities/globalDefinitions_xlc.hpp
@@ -43,7 +43,7 @@
 #if (defined(__VEC__) || defined(__AIXVEC)) && defined(AIX) \
     && defined(__open_xl_version__) && __open_xl_version__ >= 17
   #undef malloc
-  extern void	*malloc(size_t)  asm("vec_malloc");
+  extern void *malloc(size_t) asm("vec_malloc");
 #endif
 
 #include <wchar.h>
--- a/test/jdk/java/io/File/libGetXSpace.c
+++ b/test/jdk/java/io/File/libGetXSpace.c
@@ -125,7 +125,7 @@ Java_GetXSpace_getSpace0
     }
 #else
     struct statfs buf;
-    int result = statfs((const char*)chars, &buf);
+    int result = statfs((char*)chars, &buf);
     (*env)->ReleaseStringChars(env, root, chars);
     if (result < 0) {
         JNU_ThrowByNameWithLastError(env, "java/lang/RuntimeException",
--- a/make/modules/java.base/Lib.gmk
+++ b/make/modules/java.base/Lib.gmk
@@ -46,6 +46,8 @@ $(eval $(call SetupJdkLibrary, BUILD_LIBNET, \
     DISABLED_WARNINGS_gcc_net_util_md.c := format-nonliteral, \
     DISABLED_WARNINGS_gcc_NetworkInterface.c := unused-function, \
     DISABLED_WARNINGS_clang_net_util_md.c := format-nonliteral, \
+    DISABLED_WARNINGS_clang_aix_DefaultProxySelector.c := deprecated-non-prototype, \
+    DISABLED_WARNINGS_clang_aix_NetworkInterface.c := gnu-pointer-arith, \
     DISABLED_WARNINGS_microsoft_InetAddress.c := 4244, \
     DISABLED_WARNINGS_microsoft_ResolverConfigurationImpl.c := 4996, \
     LDFLAGS := $(LDFLAGS_JDKLIB) \
@@ -75,6 +77,7 @@ $(eval $(call SetupJdkLibrary, BUILD_LIBNIO, \
         libnio/ch \
         libnio/fs \
         libnet, \
+    DISABLED_WARNINGS_clang_aix_UnixNativeDispatcher.c := undef, \
     LDFLAGS := $(LDFLAGS_JDKLIB) \
         $(call SET_SHARED_LIBRARY_ORIGIN), \
     LIBS_unix := -ljava -lnet, \
--- a/make/modules/java.base/lib/CoreLibraries.gmk
+++ b/make/modules/java.base/lib/CoreLibraries.gmk
@@ -61,6 +61,7 @@ $(eval $(call SetupJdkLibrary, BUILD_LIBJAVA, \
     jdk_util.c_CFLAGS := $(VERSION_CFLAGS), \
     WARNINGS_AS_ERRORS_xlc := false, \
     DISABLED_WARNINGS_gcc_ProcessImpl_md.c := unused-result, \
+    DISABLED_WARNINGS_clang_aix_ProcessHandleImpl_unix.c := sign-compare, \
     LDFLAGS := $(LDFLAGS_JDKLIB) \
         $(call SET_SHARED_LIBRARY_ORIGIN), \
     LDFLAGS_macosx := -L$(SUPPORT_OUTPUTDIR)/native/$(MODULE)/, \
@@ -173,6 +174,7 @@ $(eval $(call SetupJdkLibrary, BUILD_LIBJLI, \
     CFLAGS := $(CFLAGS_JDKLIB) $(LIBJLI_CFLAGS), \
     DISABLED_WARNINGS_gcc := unused-function, \
     DISABLED_WARNINGS_clang := format-nonliteral deprecated-non-prototype, \
+    DISABLED_WARNINGS_clang_aix := gnu-pointer-arith, \
     LDFLAGS := $(LDFLAGS_JDKLIB) \
         $(call SET_SHARED_LIBRARY_ORIGIN), \
     LIBS_unix := $(LIBZ_LIBS), \
@@ -198,6 +200,7 @@ ifeq ($(call isTargetOs, aix), true)
       OPTIMIZATION := HIGH, \
       CFLAGS := $(STATIC_LIBRARY_FLAGS) $(CFLAGS_JDKLIB) $(LIBJLI_CFLAGS) \
           $(addprefix -I, $(LIBJLI_SRC_DIRS)), \
+      DISABLED_WARNINGS_clang := gnu-pointer-arith format-nonliteral deprecated-non-prototype, \
       ARFLAGS := $(ARFLAGS), \
       OBJECT_DIR := $(SUPPORT_OUTPUTDIR)/native/$(MODULE)/libjli_static))
 
--- a/make/modules/java.desktop/lib/Awt2dLibraries.gmk
+++ b/make/modules/java.desktop/lib/Awt2dLibraries.gmk
@@ -240,6 +240,13 @@ ifeq ($(call isTargetOs, windows macosx), false)
         DISABLED_WARNINGS_gcc_XRBackendNative.c := maybe-uninitialized, \
         DISABLED_WARNINGS_gcc_XToolkit.c := unused-result, \
         DISABLED_WARNINGS_gcc_XWindow.c := unused-function, \
+        DISABLED_WARNINGS_clang_aix := format-nonliteral deprecated-non-prototype, \
+        DISABLED_WARNINGS_clang_aix_awt_Taskbar.c := parentheses, \
+        DISABLED_WARNINGS_clang_aix_OGLPaints.c := format-nonliteral, \
+        DISABLED_WARNINGS_clang_aix_gtk2_interface.c := parentheses logical-op-parentheses, \
+        DISABLED_WARNINGS_clang_aix_gtk3_interface.c := parentheses logical-op-parentheses, \
+        DISABLED_WARNINGS_clang_aix_sun_awt_X11_GtkFileDialogPeer.c := parentheses, \
+        DISABLED_WARNINGS_clang_aix_awt_InputMethod.c := sign-compare, \
         LDFLAGS := $(LDFLAGS_JDKLIB) \
             $(call SET_SHARED_LIBRARY_ORIGIN) \
             -L$(INSTALL_LIBRARIES_HERE), \
@@ -446,7 +453,9 @@ else
    # Early re-canonizing has to be disabled to workaround an internal XlC compiler error
    # when building libharfbuzz
    ifeq ($(call isTargetOs, aix), true)
+    ifneq ($(TOOLCHAIN_TYPE), clang)
      HARFBUZZ_CFLAGS += -qdebug=necan
+    endif
    endif
 
    # hb-ft.cc is not presently needed, and requires freetype 2.4.2 or later.
@@ -689,6 +698,15 @@ ifeq ($(ENABLE_HEADLESS_ONLY), false)
 
   ifeq ($(call isTargetOs, linux), true)
     ifeq ($(call isTargetCpuArch, ppc), true)
+      LIBSPLASHSCREEN_CFLAGS += -DPNG_POWERPC_VSX_OPT=0
+    endif
+  endif
+  # temporarily disable PNG_POWERPC_VSX_OPT which would be set, because
+  # if defined(__PPC64__) && defined(__ALTIVEC__) && defined(__VSX__)
+  # is true, the then needed libpng function .png_init_filter_functions_vsx
+  # is not part of the libpng due the fact that the library was build with old compiler
+  ifeq ($(call isTargetOs, aix), true)
+    ifeq ($(TOOLCHAIN_TYPE), clang)
       LIBSPLASHSCREEN_CFLAGS += -DPNG_POWERPC_VSX_OPT=0
     endif
   endif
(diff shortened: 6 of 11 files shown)


src/hotspot/share/utilities/globalDefinitions_xlc.hpp
```diff
@@ -38,6 +38,14 @@
 #include <stddef.h>
 #include <stdio.h>
 #include <stdlib.h>
+// Until IBM modifies its stdlib.h header we repair the critical lines ourself
+// replacing the #define malloc vec_malloc by the modified malloc Function declaration.
+#if (defined(__VEC__) || defined(__AIXVEC)) && defined(AIX) \
+    && defined(__open_xl_version__) && __open_xl_version__ >= 17
+  #undef malloc
+  extern void *malloc(size_t) asm("vec_malloc");
+#endif
```
kimbarrett, line 47: Wow!  And I don't mean that in a good way.  I'm not questioning whether this is correct, just commenting
on how crazy it seems that such a thing might be needed.

test/jdk/java/io/File/libGetXSpace.c
```diff
@@ -125,7 +125,7 @@ Java_GetXSpace_getSpace0
     }
 #else
     struct statfs buf;
-    int result = statfs((const char*)chars, &buf);
+    int result = statfs((char*)chars, &buf);
```
kimbarrett, line 128: Is this working around a bug in IBM's declaration?

Also, pre-existing, the cast seems really suspicious.  The type of `chars` is `jchar*`, which is a
sequence of 16bit characters.  Does this actually work?  If so, how?

JoKern65: This is IBMs declaration of statfs
`extern int statfs(char *, struct statfs *);`
So the compiler will not accept a `const char*`
Indeed I do not know if this ever worked, but my change makes it not worse.

src/hotspot/share/utilities/globalDefinitions_xlc.hpp
TheRealMDoerr: The crazy thing is that `malloc` is defined! That means all places where we use the term malloc are getting replaced without such a workaround. (E.g. for log tags.)

test/jdk/java/io/File/libGetXSpace.c
MBaesken: Here is the documentation of statfs on AIX  https://www.ibm.com/docs/en/aix/7.2?topic=s-statfs-fstatfs-statfs64-fstatfs64-ustat-subroutine  (showing the IBM declaration Joachim told us) .

JoKern65:
Here are the reasons for the disabled warnings in make/modules/java.base/Lib.gmk
    ```
DISABLED_WARNINGS_clang_aix_DefaultProxySelector.c := deprecated-non-prototype, \
    DISABLED_WARNINGS_clang_aix_NetworkInterface.c := gnu-pointer-arith, \
    DISABLED_WARNINGS_clang_aix_UnixNativeDispatcher.c := undef, \
```

```
src/java.base/unix/native/libnet/DefaultProxySelector.c:378:41:22: error: passing arguments to a function without a prototype is deprecated in all versions of C and is not supported in C2x [-Werror,-Wdeprecated-non-prototype]
    proxies = (*g_proxy_resolver_lookup)(resolver, uri, NULL, &error);
                                        ^
```
```
src/java.base/unix/native/libnet/NetworkInterface.c:1612:24: error: arithmetic on a pointer to void is a GNU extension [-Werror,-Wgnu-pointer-arith]
    end = (void *)nddp + size;
          ~~~~~~~~~~~~ ^
```
```
src/java.base/unix/native/libnio/fs/UnixNativeDispatcher.c:1251:7: error: '_ALLBSD_SOURCE' is not defined, evaluates to 0 [-Werror,-Wundef]
#elif _ALLBSD_SOURCE
      ^

```

**Comment by JoKern65** [comment]
Here are the reasons for the disabled warnings in make/modules/java.base/lib/CoreLibraries.gmk
  DISABLED_WARNINGS_clang_aix_ProcessHandleImpl_unix.c := sign-compare, \
  DISABLED_WARNINGS_clang_aix := gnu-pointer-arith, \
  DISABLED_WARNINGS_clang := gnu-pointer-arith format-nonliteral deprecated-non-prototype, \

src/java.base/unix/native/libjava/ProcessHandleImpl_unix.c:638 comparison of integers of different signs: 'int' and 'unsigned long'
if (ret < sizeof(psinfo_t)) {
         ^

src/java.base/aix/native/libjli/java_md_aix.c:42:38: error: arithmetic on a pointer to void is a GNU extension [-Werror,-Wgnu-pointer-arith]
            addr < p->ldinfo_textorg + p->ldinfo_textsize) {
                   ~~~~~~~~~~~~~~~~~ ^

src/java.base/share/native/libzip/zlib/inffast.c:74:20: error: a function definition without a prototype is deprecated in all versions of C and is not supported in C2x [-Werror,-Wdeprecated-non-prototype]
void ZLIB_INTERNAL inflate_fast(strm, start)
                   ^
src/java.base/share/native/libjli/java.c:2311:22: error: format string is not a string literal [-Werror,-Wformat-nonliteral]
    vfprintf(stderr, fmt, vl);
                            ^~~

**Comment by MBaesken** [comment]
> src/java.base/unix/native/libnio/fs/UnixNativeDispatcher.c:1251:7: error: '_ALLBSD_SOURCE' is not defined, evaluates to 0 [-Werror,-Wundef]
> #elif _ALLBSD_SOURCE
> 

Should probably better be
`#elif defined(_ALLBSD_SOURCE)`

**Comment by MBaesken** [comment]
> src/java.base/share/native/libjli/java.c:2311:22: error: format string is not a string literal [-Werror,-Wformat-nonliteral] vfprintf(stderr, fmt, vl); ^~~

We disable this warning too for clang in BUILD_LIBJLI ( DISABLED_WARNINGS_clang := **format-nonlitera**l deprecated-non-prototype, \ ), so I think we should do it the same way for BUILD_LIBJLI_STATIC on AIX.

**Comment by MBaesken** [comment]
> Here are the reasons for the disabled warnings in make/modules/java.base/lib/CoreLibraries.gmk DISABLED_WARNINGS_clang_aix_ProcessHandleImpl_unix.c := sign-compare, DISABLED_WARNINGS_clang_aix := gnu-pointer-arith, DISABLED_WARNINGS_clang := gnu-pointer-arith format-nonliteral deprecated-non-prototype, \
> 
> src/java.base/unix/native/libjava/ProcessHandleImpl_unix.c:638 comparison of integers of different signs: 'int' and 'unsigned long' if (ret < sizeof(psinfo_t)) { ^
> 

`fread` returns a `size_t ` on Linux and AIX, could you please check Mac/BSD too ?
https://github.com/openjdk/jdk/blob/d3b9b364da8c11c9b4dd14a6451a7b24f41202e7/src/java.base/unix/native/libjava/ProcessHandleImpl_unix.c#L636

We should probably change to size_t  ret  (from type int), then the warning would not occur, correct ?

**Comment by JoKern65** [comment]
Here are the reasons for the disabled warnings in make/modules/java.desktop/lib/Awt2dLibraries.gmk

/data/d042520/pr/jdk/src/java.desktop/unix/native/common/awt/awt_GraphicsEnv.h:53:12: error: a function declaration without a prototype is deprecated in all versions of C and is treated as a zero-parameter prototype in C2x, conflicting with a previous declaration [-Werror,-Wdeprecated-non-prototype]
extern int XShmQueryExtension();
           ^
/usr/include/X11/extensions/XShm.h:91:6: note: conflicting prototype is here
Bool XShmQueryExtension(
     ^
solved by adding line (generic, because several source files are involved)
DISABLED_WARNINGS_clang_aix := deprecated-non-prototype, \


src/java.desktop/unix/native/libawt_xawt/xawt/awt_Taskbar.c:158:11: error: using the result of an assignment as a condition without parentheses [-Werror,-Wparentheses]
    if (m = fp_unity_launcher_entry_get_quicklist(entry)) {
        ~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
solved by adding line
DISABLED_WARNINGS_clang_aix_awt_Taskbar.c := parentheses, \

src/java.desktop/share/native/common/java2d/opengl/OGLPaints.c:581:48: error: format string is not a string literal [-Werror,-Wformat-nonliteral]
        snprintf(cycleCode, sizeof(cycleCode), noCycleCode, texCoordCalcCode);
                                               ^~~~~~~~~~~
solved by adding line
DISABLED_WARNINGS_clang_aix_OGLPaints.c := format-nonliteral, \

src/java.desktop/share/native/common/java2d/opengl/OGLBufImgOps.c:153:48: error: format string is not a string literal [-Werror,-Wformat-nonliteral]
    snprintf(finalSource, sizeof(finalSource), convolveShaderSource,
                                               ^~~~~~~~~~~~~~~~~~~~
solved by adding line
DISABLED_WARNINGS_clang_aix_OGLBufImgOps.c := format-nonliteral, \


src/java.desktop/unix/native/libawt_xawt/awt/gtk2_interface.c:1095:41: error: '&&' within '||' [-Werror,-Wlogical-op-parentheses]
    if ((synth_state & MOUSE_OVER) != 0 && (synth_state & PRESSED) == 0 ||
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ~~
src/java.desktop/unix/native/libawt_xawt/awt/gtk2_interface.c:1180:29: error: using the result of an assignment as a condition without parentheses [-Werror,-Wparentheses]
            if (init_result = (NULL == gtk2_widgets[_GTK_CHECK_MENU_ITEM_TYPE]))
                ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
solved by adding line
DISABLED_WARNINGS_clang_aix_gtk2_interface.c := parentheses logical-op-parentheses, \

src/java.desktop/unix/native/libawt_xawt/awt/gtk3_interface.c:903:29: error: using the result of an assignment as a condition without parentheses [-Werror,-Wparentheses]
            if (init_result = (NULL == gtk3_widgets[_GTK_BUTTON_TYPE]))
                ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
solved by adding line
DISABLED_WARNINGS_clang_aix_gtk3_interface.c := parentheses, \

src/java.desktop/unix/native/libawt_xawt/awt/sun_awt_X11_GtkFileDialogPeer.c:87:26: error: using the result of an assignment as a condition without parentheses [-Werror,-Wparentheses]
    if (pendingException = (*env)->ExceptionOccurred(env)) {
        ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
solved by adding line
DISABLED_WARNINGS_clang_aix_sun_awt_X11_GtkFileDialogPeer.c := parentheses, \


src/java.desktop/aix/native/libawt_xawt/awt/awt_InputMethod.c:1969:32: error: comparison of integers of different signs: 'Window' (aka 'unsigned long') and 'jlong' (aka 'long') [-Werror,-Wsign-compare]
        if (currentFocusWindow != w) {
            ~~~~~~~~~~~~~~~~~~ ^  ~
solved by adding line
DISABLED_WARNINGS_clang_aix_awt_InputMethod.c := sign-compare, \

**Comment by JoKern65** [comment]
Here is the reason for the disabled warning in make/modules/java.security.jgss/Lib.gmk

src/java.security.jgss/share/native/libj2gss/NativeUtil.h:30:
/data/d042520/pr/jdk/src/java.security.jgss/share/native/libj2gss/gssapi.h:48:5: error: 'TARGET_OS_MAC' is not defined, evaluates to 0 [-Werror,-Wundef]
#if TARGET_OS_MAC && (defined(__ppc__) || defined(__ppc64__) || defined(__i386__) || defined(__x86_64__))
    ^
make/modules/java.security.jgss/Lib.gmk add line 
DISABLED_WARNINGS_clang_aix := undef, \

**Comment by JoKern65** [comment]
Here is the reason for the disabled warning in make/modules/jdk.jdwp.agent/Lib.gmk

src/jdk.jdwp.agent/share/native/libdt_socket/socketTransport.c:718:33: error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
    struct in6_addr mappedAny = IN6ADDR_ANY_INIT;
                                ^~~~~~~~~~~~~~~~
/usr/include/netinet/in.h:454:32: note: expanded from macro 'IN6ADDR_ANY_INIT'
#define IN6ADDR_ANY_INIT      {0, 0, 0, 0}
solved by adding line
    DISABLED_WARNINGS_clang_aix := missing-braces, \

**Comment by TheRealMDoerr** [comment]
Just a side note: The warning elimination is new for AIX compilers. We had given it up at earlier times: https://bugs.openjdk.org/browse/JDK-8202325
I still appreciate if we can get warning free and it makes sense to review all of them. But, I wouldn't require it for JDK21.

File: src/hotspot/share/utilities/globalDefinitions_xlc.hpp (hunk 1 above)
Comment (tstuefe) [reply src/hotspot/share/utilities/globalDefinitions_xlc.hpp:47]: So, we do this only for malloc? Not for calloc, posix_memalign, realloc etc? What about free? 

As ugly as defining malloc is (and I remember QADRT), I hesitate about removing that define.

Removing that define and hard-coding it here assumes 1) our replacement is equivalent (ok, easy to check) 2) it will always be equivalent in future AIX versions 3) pointers it returns work with the unchanged free() and realloc() the system provides, and will always do so.

I don't know... I would not do this just to get rid of a warning.

Comment (TheRealMDoerr) [reply src/hotspot/share/utilities/globalDefinitions_xlc.hpp:47]: This one is not just to get rid of a warning. We get real test errors because malloc gets replaced by vec_malloc in log tags.

Comment (tstuefe) [reply src/hotspot/share/utilities/globalDefinitions_xlc.hpp:47]: > This one is not just to get rid of a warning. We get real test errors because malloc gets replaced by vec_malloc in log tags.

That does not invalidate my argument, nor does it answer my question. Those test errors could be also fixed by renaming the log tag. Maybe that would be the better way. 

Also, I'm curious, why does it not complain about "free", which is a logtag as well?

Comment (tstuefe) [reply src/hotspot/share/utilities/globalDefinitions_xlc.hpp:47]: I am basically worried that undefining malloc, even if it seems harmless now, exposes us to difficult-to-investigate problems down the road, since it depends on how the libc devs will reform those macros in the future. I would prefer a simple solution that does not depend on how the libc includes evolve.

Comment (kimbarrett) [reply src/hotspot/share/utilities/globalDefinitions_xlc.hpp:47]: Is it possible to see the stdlib.h source code that is being a problem?  Maybe more eyes can come up
with a better solution, or at least come to a better understanding of why we have to go this way.

Comment (kimbarrett) [reply src/hotspot/share/utilities/globalDefinitions_xlc.hpp:47]: Technically it's our fault.  The standard library is permitted to provide a
macro replacements for (nearly?) any function.  (C99 7.1.3/1 3rd bullet) But
it's really annoying.  Presumably AIX is only doing so for a subset of the
allocation functions, e.g. not "free" for example.

Having "malloc" defined as a macro seems like it should raise all kinds of
havoc, not just around log tags.  Consider our "os::malloc" function?  The
preprocessor knows nothing about C++ namespace qualifiers.

File: src/hotspot/share/runtime/javaThread.cpp [hunk 3]
```diff
@@ -109,6 +109,10 @@
 #include "jfr/jfr.hpp"
 #endif
 
+#ifdef _AIX
+#include <alloca.h>
+#endif
+
```
coleenp, line 115: Could these conditionals be included in globalDefinitions_xlc.hpp instead?

JoKern65: In principle the `#include <alloca.h>` could be included in globalDefinitions_xlc.hpp. But alloca.h implements it with a
`#undef  alloca`
`#define alloca(size)   __builtin_alloca (size)`
If this new global define does not introduce new trouble (even in future coding) when it is seen in every source, we can go this way.
Are there any obstacles from anyone else?

JoKern65: At least the whole jdk actually builds with `#include <alloca.h>` in globalDefinitions_xlc.hpp

prrace (approved):
I am Ok with the client-libs changes here. I did not look at anything else.
So you'll need more approvals before its ready to push.

JoKern65:
Hi,

As this PR is big and spans several components I split
off the java.base, java.desktop and the sercivability/security
issues  into extra JBS issues. 
https://bugs.openjdk.org/browse/JDK-8309219 Fix xlc17 clang 1.5 warnings in java.base
https://bugs.openjdk.org/browse/JDK-8309224 Fix xlc17 clang 1.5 warnings in java.desktop
https://bugs.openjdk.org/browse/JDK-8309225 Fix xlc17 clang 1.5 warnings in security and servicability
I’ll move the changes from this pull request into new pull requests.
I will incorporate the requested changes right in the new PRs.
I will reuse this issue 8308388 for the hotspot changes but come up
with a new, smaller PR.
@colleenp, I will move alloca.h to the globalDefinitions_xlc.hpp.
@prrace, I will come up with an identical PR for the client
files (java.desktop), but improve the comment as @kimbarrett proposed
@mbaesken, I will use AIX and take up some of the other fixes you proposed.

I guess we need to find a way to fix the issue with the malloc 
in globalDefinitions_xlc.hpp in the upcoming PR for hotspot.  

Thanks for your help so far!

JoKern65:
Hi,

As this PR is big and spans several components I split
off the java.base, java.desktop and the sercivability/security
issues into extra JBS issues.
https://bugs.openjdk.org/browse/JDK-8309219 Fix xlc17 clang 1.5 warnings in java.base
https://bugs.openjdk.org/browse/JDK-8309224 Fix xlc17 clang 1.5 warnings in java.desktop
https://bugs.openjdk.org/browse/JDK-8309225 Fix xlc17 clang 1.5 warnings in security and servicability
I’ll move the changes from this pull request into new pull requests.
I will incorporate the requested changes right in the new PRs.
I will reuse this issue 8308388 for the hotspot changes but come up
with a new, smaller PR.
@colleenp, I will move alloca.h to the globalDefinitions_xlc.hpp.
@prrace, I will come up with an identical PR for the client
files (java.desktop), but improve the comment as @kimbarrett proposed
@MBaesken, I will use AIX and take up some of the other fixes you proposed.

I guess we need to find a way to fix the issue with the malloc
in globalDefinitions_xlc.hpp in the upcoming PR for hotspot.

Thanks for your help so far!

PR #14150, opened 2023-05-25
8308589: gc/cslocker/TestCSLocker.java timed out

stefank:
We have found that this test is flawed and will cause deadlocks if allocations wait for a GC to complete. We tried to fix this issue by removing one source of allocations (see JDK-8308043), but that there are still more reasons why the JVM might allocate memory in the test. The test has a history of causing timeouts (likely caused by deadlocks), but we're currently only seeing hangs with Generational ZGC. I propose that we turn off this test for Generational ZGC, and if the test still cause problems in other configurations then we'll reevaluate if this should be handled some other way.

---------
### Progress

### Issue
 * [JDK-8308589](https://bugs.openjdk.org/browse/JDK-8308589): gc/cslocker/TestCSLocker.java timed out

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14150/head:pull/14150` \
`$ git checkout pull/14150`

`$ git checkout pull/14150` \
`$ git pull https://git.openjdk.org/jdk.git pull/14150/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14150`

`$ git pr show -t 14150`

</details>

<a href="https://git.openjdk.org/jdk/pull/14150.diff">https://git.openjdk.org/jdk/pull/14150.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14150#issuecomment-1562948118)

### DIFF
--- a/test/hotspot/jtreg/gc/cslocker/TestCSLocker.java
+++ b/test/hotspot/jtreg/gc/cslocker/TestCSLocker.java
@@ -32,6 +32,13 @@ import static gc.testlibrary.Allocation.blackHole;
  * @summary This short test check RFE 6186200 changes. One thread locked
  * @summary completely in JNI CS, while other is trying to allocate memory
  * @summary provoking GC. OOM means FAIL, deadlock means PASS.
+ *
+ * @comment This test assumes that no allocation happens during the sleep loop,   \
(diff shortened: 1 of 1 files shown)


fisk (approved):
Looks good.

PR #14152, opened 2023-05-25
8308657: ReplayInline is not availabe in production build

ashu-mehra:
DumpInline functionality is available in product build but ReplayInline is available in non product build only. This patch makes ReplayInline functionality available in product builds by moving it out of "#ifndef PRODUCT" directive.


vnkozlov (approved):
Good. We have already `InlineDataFile` as product flag.
And `DumpInline` compile command option is available in product too.
So we missed only this place.

rwestrel (approved):
Looks good to me.

TobiHartmann (approved):

PR #14154, opened 2023-05-25
8308881: Strong CLD oop handle roots are demoted to non-roots concurrently

fisk:
It is illegal to remove strong roots concurrently without clearing them first. A SATB collector with concurrent root scanning assumes that when strong roots disappear from the object graph, they are cleared first, which makes SATB notice the root. All global strong roots do this. Except CLD strong roots, which are turned into non-roots by decrementing the keep_alive counter to 0, when bootstrapping weak hidden class CLDs. This is not valid behaviour. This patch tries to treat these oops like we do any other global strong handles that are unlinked from the system: clear them when they stop being strong roots.

---------
### Progress

### Issue
 * [JDK-8308881](https://bugs.openjdk.org/browse/JDK-8308881): Strong CLD oop handle roots are demoted to non-roots concurrently

### Reviewers
 * [Coleen Phillimore](https://openjdk.org/census#coleenp) (@coleenp - **Reviewer**) ⚠️ Review applies to [b75fce32](https://git.openjdk.org/jdk/pull/14154/files/b75fce3261c5c49d3f1dee07daa66d8fd3766c5f)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14154/head:pull/14154` \
`$ git checkout pull/14154`

`$ git checkout pull/14154` \
`$ git pull https://git.openjdk.org/jdk.git pull/14154/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14154`

`$ git pr show -t 14154`

</details>

<a href="https://git.openjdk.org/jdk/pull/14154.diff">https://git.openjdk.org/jdk/pull/14154.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14154#issuecomment-1563135343)

### DIFF
--- a/src/hotspot/share/classfile/classLoaderData.cpp
+++ b/src/hotspot/share/classfile/classLoaderData.cpp
@@ -302,18 +302,34 @@ bool ClassLoaderData::try_claim(int claim) {
 
 void ClassLoaderData::demote_strong_roots() {
   // The oop handle area contains strong roots that the GC traces from. We are about
-  // to demote them to strong roots that the GC does *not* trace from. Conceptually,
-  // we are retiring a normal strong root, and creating a strong non-root handle, which
-  // happens to reuse the same address as the normal strong root had.
-  // The way we would retire a strong root, is by clearing it. Then its address can be
-  // reused for a new handle, that isn't a normal strong root. We do that dance below.
+  // to demote them to strong native oops that the GC does *not* trace from. Conceptually,
+  // we are retiring a rather normal strong root, and creating a strong non-root handle,
+  // which happens to reuse the same address as the normal strong root had.
+  // Unless we invoke the right barriers, the GC might not notice that a strong root
+  // has been pulled from the system, and is left unprocessed by the GC. There can be
+  // several consequences:
+  // 1. A concurrently marking snapshot-at-the-beginning GC might assume that the contents
+  //    of all strong roots get processed by the GC in order to keep them alive. Without
+  //    barriers, some objects might not be kept alive.
+  // 2. A concurrently relocating GC might assume that after moving an object, a subsequent
+  //    tracing from all roots can fix all the pointers in the system, which doesn't play
+  //    well with roots racingly being pulled.
+  // 3. A concurrent GC using colored pointers, might assume that tracing the object graph
+  //    from roots results in all pointers getting some particular color, which also doesn't
+  //    play well with roots being pulled out from the system concurrently.
+
   class TransitionRootsOopClosure : public OopClosure {
   public:
     virtual void do_oop(oop* p) {
+      // By loading the strong root with the access API, we can use the right barriers to
+      // store the oop as a strong non-root handle, that happens to reuse the same memory
+      // address as the strong root. The barriered store ensures that:
+      // 1. The concurrent SATB marking properties are satisfied as the store will keep
+      //    the oop alive.
+      // 2. The concurrent object movement properties are satisfied as we store the address
+      //    of the new location of the object, if any.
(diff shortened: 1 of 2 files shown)


stefank (commented):
Could this be using CollectedHeap::keep_alive instead of performing this clear/store dance?

fisk:
> Could this be using CollectedHeap::keep_alive instead of performing this clear/store dance?

No, because the problem isn't keeping the object alive. It's already kept alive using handles. The problem is that the snapshot of strong roots are not processed during marking, leading to ABA problems, causing things to go wrong later on. Therefore I wanted to model the solution conceptually as close as possible to other handles. Admittedly, it does look more like an exptic rain dance than ideal. Just got to clap our hands, spin around 360 degrees and then jump - then clap our hands again!

coleenp (approved):
This is strange but if it works, it's ok with me.

fisk:
Thanks for the review @coleenp!

src/hotspot/share/classfile/classLoaderData.cpp
```diff
@@ -300,6 +300,30 @@ bool ClassLoaderData::try_claim(int claim) {
...
+  // The way we would retire a strong root, is by clearing it. Then its address can be
+  // reused for a new handle, that isn't a normal strong root. We do that dance below.
+  class TransitionRootsOopClosure : public OopClosure {
+  public:
+    virtual void do_oop(oop* p) {
+      oop obj = NativeAccess<>::oop_load(p); // Load the strong root
+      NativeAccess<>::oop_store(p, nullptr); // Clear the strong root
+      // No-op free handle
+      // No-op allocate new handle using the same address
+      NativeAccess<>::oop_store(p, obj); // Store the strong non-root
```
xmas92, line 317: Is there an issue if someone with an `OopHandle(p)` racingly reads null while a root is transitioning?

fisk: Shouldn't be, but I'll remove the null store and write some more comments, to be on the safe side.

stefank (approved):

xmas92:
Is there any issue with memory order access here?
From my understanding this works because a GC which starts before `demote_strong_roots` has the property if it observers `_keep_alive == 0` then in must have observed `demote_strong_roots` (either root throughout or it observed the transition through barriers). And if a GC starts after `demote_strong_roots` it must never observe `_keep_alive != 0`.

fisk:
> Is there any issue with memory order access here?
> From my understanding this works because a GC which starts before `demote_strong_roots` has the property if it observers `_keep_alive == 0` then in must have observed `demote_strong_roots` (either root throughout or it observed the transition through barriers). And if a GC starts after `demote_strong_roots` it must never observe `_keep_alive != 0`.

If the GC races with these accesses, it doesn't matter if the race hits before or after the accesses. The accesses themselves will sync with the GC. To be precise:
1. If the object relocated, the load barrier will detect that and the new value stored will pointed to the to-space object.
2. The store barrier will load the previous value and enqueue for marking, so that we don't miss there was an oop in there that needed to be marked.
3. After the last store, the colors will be right with both ZGC collectors. GC will never perform non-monotonic transitions away from the best color, which the store ensures we will get.

If the GC processes the location before these accesses, it will ensure the above properties. Whoever gets there first doesn't matter as long as one of them will.

fisk:
And whether the GC observes the old or new value of the keep alive counter doesn't matter either, as at least one thread ensures the oop location gets processed. If the GC sees the counter as 0 in this race, the mutator will process the oop location as I described. And then at least one thread has processed it.

PR #14156, opened 2023-05-25
8307683: Loop Predication should not hoist range checks with trap on success projection by negating their condition

chhagedorn:
[JDK-4809552](https://bugs.openjdk.org/browse/JDK-4809552) allowed Loop Predication to be applied to `IfNodes` that have a positive value instead of a `LoadRangeNode`:
https://github.com/openjdk/jdk/blob/48d21bd089a3f344ee5407926f8ed2af3734d2b0/src/hotspot/share/opto/loopPredicate.cpp#L854-L862

This, however, is only correct if we have an actual `RangeCheckNode` for an array. The reason for that is that if we hoist a real range check and create a Hoisted Predicate for it, we only need to check the lower and upper bound of all array accesses (i.e. the array access of the first and the last loop iteration). All array accesses in between are implicitly covered and do not need to be checked again. 

But if we face an `IfNode` without a `LoadRangeNode`, we could be comparing anything. We do not have any guarantee that if the first and last loop iteration check succeed that the other loop iteration checks will also succeed. An example of this is shown in the test case `test()`. We wrongly create a Hoisted Range Check Predicate where the lower and upper bound are always true, but for some values of the loop induction variable, the hoisted check would actually fail. We then crash because an added Assertion Predicate exactly performs this failing check (crash with halt). Without any loop splitting (i.e. no Assertion Predicates), we have a wrong execution due to never executing the branch where we increment `iFld2` because we removed it together with the check.

Thanks,
Christian

---------
### Progress

### Issue
 * [JDK-8307683](https://bugs.openjdk.org/browse/JDK-8307683): Loop Predication should not hoist range checks with trap on success projection by negating their condition

### Reviewers
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - **Reviewer**) ⚠️ Review applies to [48ee1e40](https://git.openjdk.org/jdk/pull/14156/files/48ee1e40fe7e93cb07806ff2aae3b9025b5fdebf)
 * [Roland Westrelin](https://openjdk.org/census#roland) (@rwestrel - **Reviewer**) ⚠️ Review applies to [48ee1e40](https://git.openjdk.org/jdk/pull/14156/files/48ee1e40fe7e93cb07806ff2aae3b9025b5fdebf)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14156/head:pull/14156` \
`$ git checkout pull/14156`

`$ git checkout pull/14156` \
`$ git pull https://git.openjdk.org/jdk.git pull/14156/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14156`

`$ git pr show -t 14156`

</details>

<a href="https://git.openjdk.org/jdk/pull/14156.diff">https://git.openjdk.org/jdk/pull/14156.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14156#issuecomment-1563225687)

### DIFF
--- a/src/hotspot/share/opto/loopPredicate.cpp
+++ b/src/hotspot/share/opto/loopPredicate.cpp
@@ -867,7 +867,7 @@ bool IdealLoopTree::is_range_check_if(IfProjNode* if_success_proj, PhaseIdealLoo
     //
     // Example:
     // Loop: "for (int i = -1; i < 1000; i++)"
-    // init = "scale*iv + offset" in first loop iteration = 1*-1 + 0 = -1
+    // init = "scale*iv + offset" in the first loop iteration = 1*-1 + 0 = -1
     // last = "scale*iv + offset" in the last loop iteration = 1*999 + 0 = 999
     // limit = 100
     //


test/hotspot/jtreg/compiler/predicates/TestHoistedPredicateForNonRangeCheck.java
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+
+    public static void test() {
+        // (Inverted) Hoisted Predicate checks lower and upper bound: -1 < 0 && 1000 >= 100 -> always true and the predicate is removed
+        // while template assertion predicates are kept. When splitting this loop further, we insert an assertion predicate which
+        // fails for i = 0 and we halt. When not splitting this loop (with LoopMaxUnroll=0), we have a wrong execution due
+        // to never executing iFld2++ (we removed the check and the branch with the trap).
+        for (int i = -1; i < 1000; i++) {
+            if (Integer.compareUnsigned(i, 100) < 0) { // Loop Predication creates a Hoisted Range Check Predicate due to trap with Float.isNan().
```
TobiHartmann, line 82: Maybe add a comment explaining that this is equivalent to `i <u 100` <=> `i >= 0 && i < 100` and we add a predicate for the else branch, i.e. for `i < 0 || i >= 100`, and remove the if branch.

TobiHartmann (approved):
Looks good to me. Scary, that we didn't find this earlier.

src/hotspot/share/opto/loopPredicate.cpp
```diff
@@ -853,11 +853,13 @@ bool IdealLoopTree::is_range_check_if(IfNode *iff, PhaseIdealLoop *phase, BasicT
   range = cmp->in(2);
   if (range->Opcode() != Op_LoadRange) {
     const TypeInteger* tinteger = phase->_igvn.type(range)->isa_integer(bt);
-    if (tinteger == nullptr || tinteger->empty() || tinteger->lo_as_long() < 0) {
-      // Allow predication on positive values that aren't LoadRanges.
-      // This allows optimization of loops where the length of the
-      // array is a known value and doesn't need to be loaded back
-      // from the array.
+    if (!iff->is_RangeCheck() || tinteger == nullptr || tinteger->empty() || tinteger->lo_as_long() < 0) {
```
rwestrel, line 856: Can the problem not happen with a LoadRange as second input? Couldn't the LoadRange constant fold and then the predicates could constant fold as well?

rwestrel (approved):
Looks good to me. Thanks for the quick fix.
Do we want to file a bug to revisit this in a correct way?

rwestrel:
Do we want to file a bug to investigate whether this can be done in a correct way?

chhagedorn:
Thanks Tobias and Roland for the initial reviews.

After thinking about the fix again and discussing it with Roland, I updated the fix:

We disallow this pattern for an `IfNode`:
```
if (iv <u limit) {
  trap();
}
```
Because this is flipped to:
```
if (iv >=u limit) {

} else {
  trap();
}
```
to have the trap to the false projection. However, this does not match the range check pattern of a `RangeCheckNode`:
```
if (iv <u limit) {

} else {
  trap();
}
```
We should therefore bailout in these cases, regardless of whether `limit` is a positive constant or a `LoadRangeNode`. If we still created a Hoisted Range Check Predicate, it could succeed at runtime (i.e. true for the value of `iv` in the first loop iteration and true for the value of `iv` in the last loop iteration) while the check to be hoisted could fail in other loop iterations. An example for this is shown in `test()`:
```
// Hoisted Range Check Predicate is always true:
// iv_first_iteration >=u limit && iv_last_iteration >=u limit  <=>
// -1 >=u 100 && 999 >= u 100
for (int i = -1; i < 1000; i++) {
    if (Integer.compareUnsigned(i, 100) < 0) {
        iFld2++;
        Float.isNaN(34); // Float class is unloaded with -Xcomp -> inserts trap
    } else {
        iFld++;
    }
}
```
However, if `0 <= i < 100`, then the hoisted check `Integer.compareUnsigned(i, 100) < 0` would be false. We then wrongly skip the branch with the `Float.isNan(34)` trap (was removed when creating the Hoisted Range Check Predicate) and miss to execute `iFld2` leading to a wrong execution (or when splitting this loop, we halt because of an initialized Assertion Predicate which will fail).

src/hotspot/share/opto/loopPredicate.cpp
chhagedorn: You're right. It's also a problem with a `LoadRange`, even if it does not constant fold.

rwestrel (approved):
Looks good to me.

jerboaa:
This fails the `[compiler/loopopts/TestSkeletonPredicateNegation` test with:

```
#
#
#  Internal Error (/home/runner/work/jdk/jdk/src/hotspot/share/opto/loopPredicate.cpp:848), pid=32813, tid=32828
#  assert(!iff->is_RangeCheck()) failed: can only be IfNode because RangeCheckNodes always have trap on false projection
#
# JRE version: OpenJDK Runtime Environment (21.0) (fastdebug build 21-internal-chhagedorn-229583b7613867127e42baca158773bcf9c08c73)
# Java VM: OpenJDK 64-Bit Server VM (fastdebug 21-internal-chhagedorn-229583b7613867127e42baca158773bcf9c08c73, mixed mode, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
# Problematic frame:
# V  [libjvm.so+0x1298438]  IdealLoopTree::is_range_check_if(IfProjNode*, PhaseIdealLoop*, BasicType, Node*, Node*&, Node*&, long&) const+0x278
#
# CreateCoredumpOnCrash turned off, no core file dumped
#
# An error report file with more information is saved as:
# /home/runner/work/jdk/jdk/build/run-test-prebuilt/test-support/jtreg_test_hotspot_jtreg_tier1_compiler/scratch/hs_err_pid32813.log
#
# Compiler replay data is saved as:
# /home/runner/work/jdk/jdk/build/run-test-prebuilt/test-support/jtreg_test_hotspot_jtreg_tier1_compiler/scratch/replay_pid32813.log
#
# If you would like to submit a bug report, please visit:
#   https://bugreport.java.com/bugreport/crash.jsp
#
result: Error. Agent communication error: java.io.EOFException; check console log for any additional details
```

which seems related?

chhagedorn:
Thanks @jerboaa for reporting that. I've seen that in my testing over the weekend as well. I'm looking into it.

chhagedorn:
In the failing test case (see newly added `testRangeCheckNode()`), we have a `RangeCheckNode` with an unloaded trap on the true/success projection because we are always taking the false/exception path. Loop Predication wrongly detects this pattern as valid range check and tries to create a Hoisted Predicate for that `RangeCheckNode` by flipping the boolean condition. This is wrong due to the same reasons explained above for the `IfNode`. The failing assertion should have made sure that we will never have a `RangeCheckNode` there.

I'm therefore suggesting to remove the entire negation of range checks which seems to be wrong. We should bail out in `is_range_check_if()` if we have the false projection as success projection. I've pushed an update accordingly.

rwestrel (approved):
Looks good to me.

test/hotspot/jtreg/compiler/predicates/TestHoistedPredicateForNonRangeCheck.java
```diff
@@ -0,0 +1,182 @@
...
+        // where iFld = 0, we would take the true projection for the first time - but we removed that branch when
+        // creating the Hoisted Range Check Predicate. We therefore run into the same problem as with test(): We either
+        // halt due to Assertion Predicates catching this case or we have a wrong execution (iFld2 never updated).
+        for (int i = 0; i < 37; i++) {
+            try {
+                array[iFld] = 34; // Normal RangeCheckNode
+                iFld2++;
+                Math.ceil(34); // Never taken and unloaded -> trap
+            } catch (Exception e) {
+                // False Proj of RangeCheckNod
```
TobiHartmann, line 143:
```suggestion
                // False Proj of RangeCheckNode
```

src/hotspot/share/opto/loopPredicate.cpp
```diff
@@ -831,16 +831,53 @@ class Invariance : public StackObj {
...
+    //   trap();
+    // }
+    //
+    // If we create a Hoisted Range Check Predicate for this wrong pattern, it could succeed at runtime (i.e. true
+    // for the value of "scale*iv + offset" in the first loop iteration and true for the value of "scale*iv + offset"
+    // in the last loop iteration) while the check to be hoisted could fail in other loop iterations.
+    //
+    // Example:
+    // Loop: "for (int i = -1; i < 1000; i++)"
+    // init = "scale*iv + offset" in first loop iteration = 1*-1 + 0 = -1
```
TobiHartmann, line 870:
```suggestion
    // init = "scale*iv + offset" in the first loop iteration = 1*-1 + 0 = -1
```

TobiHartmann (approved):
Nice analysis. Looks good to me. As we discussed offline, let's run some sanity performance testing (we can already integrate this).

chhagedorn:
Thanks Tobias for reviewing the new fix again! That's a good idea. I've run some standard benchmarks and could not find any regressions.

hungk20:
May I know if the fix applied to jdk11 as well? 

We got the same issue with jdk11.0.19.

chhagedorn:
@hungk20 The negation of the condition for range checks was introduced in [JDK-7173584](https://bugs.openjdk.org/browse/JDK-7173584) which went into JDK 9. Therefore, JDK 11u is also affected. You could either backport this fix or backout [JDK-8297951](https://bugs.openjdk.org/browse/JDK-8297951) (see [JDK-8308884](https://bugs.openjdk.org/browse/JDK-8308884)).

jerboaa:
OpenJDK 11.0.20 in July will have [JDK-8297951](https://bugs.openjdk.org/browse/JDK-8297951) backed out. We'll revisit the situation for OpenJDK 11.0.21 (October) where the actual fix (i.e. this bug) will likely get in. See https://bugs.openjdk.org/browse/JDK-8309119

hungk20:
Thanks @chhagedorn @jerboaa, I will keep an eye on the the new versions.

KarlHakanNordgren:
@jerboaa && @chhagedorn : I'm not familiar with the term "backed out". Does "OpenJDK 11.0.20 in July will have [JDK-8297951](https://bugs.openjdk.org/browse/JDK-8297951) backed out." mean that JDK-8297951 has been reverted?

chhagedorn:
Hi @KarlHakanNordgren, yes, in OpenJDK 11.0.20, JDK-8297951 is backed out/reverted (see [JDK-8309121](https://bugs.openjdk.org/browse/JDK-8309121)).

There is a REDO task ([JDK-8309119](https://bugs.openjdk.org/browse/JDK-8309119)) which has not (yet?) made it into OpenJDK 11.0.21 but only in OpenJDK 17.0.9.

jerboaa:
@KarlHakanNordgren What @chhagedorn said. Current JDK 11 release (11.0.20) have [JDK-8297951](https://bugs.openjdk.org/browse/JDK-8297951) reverted.

PR #14157, opened 2023-05-25
8308678: (fs) UnixPath::toRealPath needs additional permissions when running with SM (macOS)

bplb:
When not resolving links and the case retention algorithm cannot be run due to a security manager being in place with no read permissions for the tree to be traversed, then fall back to the result obtained without case correction.

---------
### Progress

### Issue
 * [JDK-8308678](https://bugs.openjdk.org/browse/JDK-8308678): (fs) UnixPath::toRealPath needs additional permissions when running with SM (macOS)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14157/head:pull/14157` \
`$ git checkout pull/14157`

`$ git checkout pull/14157` \
`$ git pull https://git.openjdk.org/jdk.git pull/14157/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14157`

`$ git pr show -t 14157`

</details>

<a href="https://git.openjdk.org/jdk/pull/14157.diff">https://git.openjdk.org/jdk/pull/14157.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14157#issuecomment-1563250836)

### DIFF
--- a/src/java.base/unix/classes/sun/nio/fs/UnixPath.java
+++ b/src/java.base/unix/classes/sun/nio/fs/UnixPath.java
@@ -906,7 +906,7 @@ class UnixPath implements Path {
             final UnixFileKey elementKey = attrs.fileKey();
 
             // Obtain the directory stream pointer. It will be closed by
-            // UnixDirectoryStream::closeImpl.
+            // UnixDirectoryStream::close.
             long dp = -1;
             try {
                 dp = opendir(path);
--- a/src/java.base/unix/classes/sun/nio/fs/UnixPath.java
+++ b/src/java.base/unix/classes/sun/nio/fs/UnixPath.java
@@ -782,6 +782,20 @@ class UnixPath implements Path {
         return open(this, flags, 0);
     }
 
+    private static boolean canRead(UnixPath path) {
+        @SuppressWarnings("removal")
+        SecurityManager sm = System.getSecurityManager();
+        if (sm != null) {
+            try {
+                sm.checkRead(path.getPathForPermissionCheck());
+                return true;
+            } catch (SecurityException ignore) {
+                return false;
+            }
+        }
+        return true;
+    }
+
     void checkRead() {
         @SuppressWarnings("removal")
         SecurityManager sm = System.getSecurityManager();
@@ -879,7 +893,11 @@ class UnixPath implements Path {
         if (!fs.isCaseInsensitiveAndPreserving())
             return result;
 
+        // If SM and no read permission, then fall back to result derived so
+        // far with case possibly not retained
         UnixPath path = fs.rootDirectory();
+        if (!canRead(path))
+            return result;
 
         // Traverse the result obtained above from the root downward, leaving
         // any '..' elements intact, and replacing other elements with the
--- a/src/java.base/unix/classes/sun/nio/fs/UnixPath.java
+++ b/src/java.base/unix/classes/sun/nio/fs/UnixPath.java
@@ -782,12 +782,12 @@ class UnixPath implements Path {
         return open(this, flags, 0);
     }
 
-    private static boolean canRead(UnixPath path) {
+    private boolean canRead() {
         @SuppressWarnings("removal")
         SecurityManager sm = System.getSecurityManager();
         if (sm != null) {
             try {
-                sm.checkRead(path.getPathForPermissionCheck());
+                sm.checkRead(getPathForPermissionCheck());
                 return true;
             } catch (SecurityException ignore) {
                 return false;
@@ -896,7 +896,7 @@ class UnixPath implements Path {
         // If SM and no read permission, then fall back to result derived so
         // far with case possibly not retained
         UnixPath path = fs.rootDirectory();
-        if (!canRead(path))
+        if (!path.canRead())
             return result;
 
         // Traverse the result obtained above from the root downward, leaving
--- a/src/java.base/unix/classes/sun/nio/fs/UnixPath.java
+++ b/src/java.base/unix/classes/sun/nio/fs/UnixPath.java
@@ -782,20 +782,6 @@ class UnixPath implements Path {
         return open(this, flags, 0);
     }
 
-    private boolean canRead() {
-        @SuppressWarnings("removal")
-        SecurityManager sm = System.getSecurityManager();
-        if (sm != null) {
-            try {
-                sm.checkRead(getPathForPermissionCheck());
-                return true;
-            } catch (SecurityException ignore) {
-                return false;
-            }
-        }
-        return true;
-    }
-
     void checkRead() {
         @SuppressWarnings("removal")
         SecurityManager sm = System.getSecurityManager();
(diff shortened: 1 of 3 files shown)


test/jdk/java/nio/file/Path/MacToRealPathWithSM.java
```diff
@@ -0,0 +1,64 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/* @test
+ * @bug 8308678
+ * @requires (os.family == "mac")
+ * @summary Verify UnixPath::toRealPath falls back if no perms on macOS
+ * @run main/othervm -Djava.security.manager=allow MacToRealPathWithSM MacToRealPath.policy
+ */
```
LanceAndersen, line 29: Any thought given to specifying the policy on the command line and running the test with and without the SecurityManager enabled?

test/jdk/java/nio/file/Path/MacToRealPathWithSM.java
```diff
@@ -0,0 +1,64 @@
...
+public class MacToRealPathWithSM {
+    public static void main(String[] args) throws Exception {
+        String policyFile = args[0];
+        String testSrc = System.getProperty("test.src");
+        String tmpDir = System.getProperty("java.io.tmpdir");
+        if (testSrc == null || tmpDir == null)
+            throw new RuntimeException("This test must be run by jtreg");
+        System.out.println("testSrc: " + testSrc);
+        System.out.println("tmpDir: " + tmpDir);
+
```
LanceAndersen, line 43: Maybe use a printf here ?

test/jdk/java/nio/file/Path/MacToRealPathWithSM.java
```diff
@@ -0,0 +1,64 @@
...
+        // Install security manager with the given policy file
+        System.setProperty("java.security.policy",
+            src.resolve(policyFile).toString());
+        System.setSecurityManager(new SecurityManager());
+
+        // Derive real path
+        System.out.printf("real path: %s%n", path.toRealPath());
+        System.out.printf("real path no follow: %s%n",
+                          path.toRealPath(LinkOption.NOFOLLOW_LINKS));
+    }
```
LanceAndersen, line 63: Same comment regarding printf

Are you expecting an Exception to be thrown if there is an issue given there is no validation?  Either way, I would clarify with additional comments for future maintainers

test/jdk/java/nio/file/Path/MacToRealPathWithSM.java
bplb: I was following the pattern of how the SM is used in at least some of the other `java.nio` tests. That does not preclude however running with an without.

test/jdk/java/nio/file/Path/MacToRealPathWithSM.java
bplb: An `AccessControlException` is thrown by `path.toRealPath(LinkOption.NOFOLLOW_LINKS)` when run against JDK 20 and the current mainline. I concur that more comments would be better.

LanceAndersen: So here is a wild idea,  perhaps consider converting this to a junit test given it  is a new test and we can start to leverage some of the value brought by the junit framework.

test/jdk/java/nio/file/Path/MacToRealPathWithSM.java
bplb: Running with and without added in 8b45bc8980d44eef1d24c45348daea3bd5e80897.

test/jdk/java/nio/file/Path/MacToRealPathWithSM.java
bplb: Done in 8b45bc8980d44eef1d24c45348daea3bd5e80897.

test/jdk/java/nio/file/Path/MacToRealPathWithSM.java
bplb: I'll investigate.

src/java.base/unix/classes/sun/nio/fs/UnixPath.java
```diff
@@ -782,6 +782,20 @@ int openForAttributeAccess(boolean followLinks) throws UnixException {
         return open(this, flags, 0);
     }
 
+    private static boolean canRead(UnixPath path) {
```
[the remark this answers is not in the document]
bplb: So changed in c91d7189053a57fa813c9f89e1289335dc88da1e.

AlanBateman:
The SM permission issue isn't limited to the root directory as each call to provider.newDirectoryStream will do a permission check too.  I think it is wrong for the implementation to ignore the security exceptions as it creates a correctness issue: run it with and without a SM set and you might get different results on macOS.

I think the question for this issue is whether it should fail with a security exception (would require a spec change) or whether it should work like toRealPath() and only require SM permissions to read the file and read the user.dir property. Consistent with the the "follow link" case seems better, in which case, the code to traverse from the root down should move to a separate method and be called in a doPrivileged block. The SM permission checks that are specified by the method are already done at the top of the method.

bplb:
> * The code to traverse from the root down should move to a separate method and be called in a doPrivileged block. The SM permission checks that are specified by the method are already done at the top of the method.

I had thought of this approach initially but decided against it.

> * Replace provider.newDirectoryStream with direct use of UnixDirectoryStream. This would be my preferred option to avoid code in the implementation make use of the API that is implements.

Will investigate.

LanceAndersen (approved):

src/java.base/unix/classes/sun/nio/fs/UnixPath.java
```diff
@@ -905,12 +905,20 @@ public Path toRealPath(LinkOption... options) throws IOException {
             }
             final UnixFileKey elementKey = attrs.fileKey();
 
+            // Obtain the directory stream pointer. It will be closed by
+            // UnixDirectoryStream::closeImpl.
```
AlanBateman, line 909: I think you can replace "closeImpl" with "close" in the comment, that will be a bit clearer as the try-with-resources is just below it.

AlanBateman (approved):
Implementation changes are okay, just one comment. I don't have time to look at the time right now but I see Lance has been reviewing that.

AlanBateman (approved):
Implementation changes are okay, just one comment. I don't have time to look at the time right now but I see Lance has been reviewing that.

src/java.base/unix/classes/sun/nio/fs/UnixPath.java
bplb: Good point. So changed in ac0936c78b24ebe5132f89018dd905ea043ce92b.

LanceAndersen (approved):

PR #14158, opened 2023-05-25
8307488: Incorrect weight of the first ObjectAllocationSample JFR event

mgronlun:
Greetings,

this PR fixes the problem with incorrect weights attribution for the initial ObjectAllocationSample event.

Testing: jdk_jfr

Thanks
Markus

---------
### Progress

### Issue
 * [JDK-8307488](https://bugs.openjdk.org/browse/JDK-8307488): Incorrect weight of the first ObjectAllocationSample JFR event

### Reviewers
 * [Erik Gahlin](https://openjdk.org/census#egahlin) (@egahlin - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14158/head:pull/14158` \
`$ git checkout pull/14158`

`$ git checkout pull/14158` \
`$ git pull https://git.openjdk.org/jdk.git pull/14158/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14158`

`$ git pr show -t 14158`

</details>

<a href="https://git.openjdk.org/jdk/pull/14158.diff">https://git.openjdk.org/jdk/pull/14158.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14158#issuecomment-1563286469)

### DIFF
--- a/src/hotspot/share/jfr/recorder/service/jfrRecorderService.cpp
+++ b/src/hotspot/share/jfr/recorder/service/jfrRecorderService.cpp
@@ -132,18 +132,18 @@ class ClearObjectAllocationSampling : public ThreadClosure {
 };
 
 template <typename Iterator>
-static inline void iterate(Iterator& it, ClearObjectAllocationSampling& robs) {
+static inline void iterate(Iterator& it, ClearObjectAllocationSampling& coas) {
   while (it.has_next()) {
-    robs.do_thread(it.next());
+    coas.do_thread(it.next());
   }
 }
 
 static void clear_object_allocation_sampling() {
-  ClearObjectAllocationSampling robs;
+  ClearObjectAllocationSampling coas;
   JfrJavaThreadIterator jit;
-  iterate(jit, robs);
+  iterate(jit, coas);
   JfrNonJavaThreadIterator njit;
-  iterate(njit, robs);
+  iterate(njit, coas);
 }
 
 template <typename Instance, size_t(Instance::*func)()>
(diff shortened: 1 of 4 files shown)


egahlin (approved):

src/hotspot/share/jfr/recorder/service/jfrRecorderService.cpp
```diff
@@ -121,6 +122,30 @@ const Thread* JfrRotationLock::_owner_thread = nullptr;
...
+    coas.do_thread(it.next());
+  }
+}
+
+static void clear_object_allocation_sampling() {
+  ClearObjectAllocationSampling coas;
+  JfrJavaThreadIterator jit;
+  iterate(jit, coas);
+  JfrNonJavaThreadIterator njit;
+  iterate(njit, coas);
```
shipilev, line 146: Apologies for the post-integration review, but I have a question/suggestion. This code retains the JavaThread iterator while NonJavaThreadIterator is running, which feels unnecessary. Also, if we have thousands of threads, walking these threads and doing the unconditional `clear_last_allocated_bytes` there seems redundant and heavy-weight if the relevant JFR event is off.

mgronlun: At this point, no recording is running, and no event settings are applied. No state exists to know if the event was enabled when the previous last recording stopped.

shipilev: OK, that's unfortunate. 

I would still suggest to split the scopes for Java and non-Java iterators. And maybe check if `last_allocated_bytes` is not zero (which is a rare case, especially when event is disabled) before clearing, thus replacing the unconditional write with conditional read-write.

PR #14159, opened 2023-05-25
8308144: HttpClient - uncontrolled memory consumption in SSLFlowDelegate.Reader

zhurs:
When using HttpClient to make requests to HTTPS resources, there is an issue where the entire file is being downloaded into memory without the ability to limit the buffer size.
If the SSLEngine cannot decode the entire buffer due to the algorithm's blocking nature, it returns a decoded chunk of data and BUFFER_UNDERFLOW status, which leads to SSLFlowDelegate.Reader requesting more data despite the output queue being full.

---------
### Progress

### Integration blocker
&nbsp;⚠️ Title mismatch between PR and JBS for issue [JDK-8308144](https://bugs.openjdk.org/browse/JDK-8308144)

### Issue
 * [JDK-8308144](https://bugs.openjdk.org/browse/JDK-8308144): Uncontrolled memory consumption in SSLFlowDelegate.Reader when  -Djdk.internal.httpclient.debug=true (**Bug** - P4) ⚠️ Title mismatch between PR and JBS.

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14159/head:pull/14159` \
`$ git checkout pull/14159`

`$ git checkout pull/14159` \
`$ git pull https://git.openjdk.org/jdk.git pull/14159/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14159`

`$ git pr show -t 14159`

</details>

<a href="https://git.openjdk.org/jdk/pull/14159.diff">https://git.openjdk.org/jdk/pull/14159.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14159#issuecomment-1563484885)

### DIFF
--- a/src/java.net.http/share/classes/jdk/internal/net/http/common/SSLFlowDelegate.java
+++ b/src/java.net.http/share/classes/jdk/internal/net/http/common/SSLFlowDelegate.java
@@ -466,8 +466,8 @@ public class SSLFlowDelegate {
                                             + len + " bytes non decrypted.");
                                 }
                             }
-                            // request more data and return.
-                            requestMore();
+                            // request more data if needed and return.
+                            requestMoreDataIfNeeded();
                             return;
                         }
                         if (complete && result.status() == Status.CLOSED) {
--- a/test/jdk/java/net/httpclient/HttpsBackpressureTest.java
+++ b/test/jdk/java/net/httpclient/HttpsBackpressureTest.java
@@ -1,0 +1,188 @@
+/*
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+import jdk.test.lib.net.SimpleSSLContext;
+
+import javax.net.ssl.SSLContext;
+import java.io.*;
+import java.net.InetAddress;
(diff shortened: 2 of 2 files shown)


dfuch:
Hi, thanks a lot for the bug report and the fix.

The fix looks reasonable, however the test fails quite consistently in our CI on many platform:

```
java.lang.RuntimeException: Too large intermediate buffer, server sent 10x300000 bytes
	at HttpsBackpressureTest.main(HttpsBackpressureTest.java:87)
	at java.base/java.lang.reflect.Method.invoke(Method.java:578)
	at com.sun.javatest.regtest.agent.MainWrapper$MainTask.run(MainWrapper.java:138)
```

I'm not sure I understand the logic of the test either. Does it depend on some assumption about the size of the socket buffers? From where do the various constants in the test come from? Also could there be a better solution than `sleep(2000)` ?

zhurs:
@dfuch, thanks for your lightning-fast reply.

Indeed it relies on buffer sizes.
The test works for me on Linux and Mac, but real life seems to be more diverse.

Unfortunately, I don't know how else to reproduce the problem and test the fix - the tested logic is hidden deep.

It may be possible to commit the fix without unit-test?
The regression is not affected, and the behavior can be checked manually using the recipe from the issue.

dfuch:
The HttpClient supports property that can be used to tune the SEND and RCV buffer sizes. I would not recommend to use them in production unless you absolutely know what you do, because setting them can prevent underlying OD optimizations which in turn can affect performance negatively. However - using them in tests is fair game. If you could explain the relationship between the constants in the test and the assumed socket buffer sizes (on the client size I assume) then we may try to experiment forcing the test's client to use some specific value for SNDBUF and RCVBUF, and see if that works. It may not, but maybe we could explore this possibility before ditching the test. If we notice that there are some platforms where the test always pass reliably we could also experiment with using `@requires` to restrict the test to run on these platforms.

dfuch:
From the CI results it seems it only passes on macOS and fails consistently on all Linux or Windows flavors. But if it depends on socket buffer size it may actually depend on how these machines are configured by default.

zhurs:
Thank you, I will look at these options.

jaikiran:
Hello @zhurs, would you mind if one of us took over this PR and moved this forward? Daniel noted that the fix looks reasonable and it's the test which will need some work. We will add you as the co-author.

zhurs:
Hi, @jaikiran 
Unfortunately I didn't find the way to write good test for this PR.
I will be happy if someone will continue work.
Thank you!

BartXZX:
Hi @jaikiran I see the Pull Request is closed, but we were actually waiting for it to get merged. Is it still getting fixed? I see the JDK issue is still open.

dfuch:
Hello @BartXZX , I believe @jaikiran is still working on this.

jaikiran:
Hello Bart, like Daniel noted I'm still working on this. Sorry it has taken this long. I'll prioritize this work for the coming days.

pditommaso:
It turns out this is a serious issue in the Java HttpClient. Any chance to re-open this?

jaikiran:
Hello @pditommaso, I've the necessary test changes ready and I will be opening a PR this coming week.

pditommaso:
Thanks a lot! Very appreciated.

jaikiran:
This has now been addressed through https://github.com/openjdk/jdk/pull/16704

PR #14160, opened 2023-05-25
8307944: ClassFileDumper should only load java.nio.file.Path if enabled

mlchung:
Change `ClassFileDumper` constructor to take a `String` rather than `Path` to avoid the file system implementation classes loaded during startup.  In addition, `FilePermission` also causes the default file system to be initialized. 

This patch changes `ClassFileDumper` to validate the given path only if it's enabled and not to use limited `doPrivileged` to avoid the initialization of `FilePermission` class and the file system implementation classes by default.

This patch also changes `ClassFileDumper::write` to use `File` instead of `Path` because of JDK-8306678.  `OperatingSystem::<clinit>` causes `Enum::valueOf` to be called which calls core reflection which in turns invokes method handles.   If `-Djdk.invoke.MethodHandle.dumpMethodHandleInternals` is set, VM initialization would fail because `BootLoader::<clinit>` causes `ClassLoaderHelper` to initialize which in turn causes `OperatingSystem` be initialized.   If additional LFs are spinned and dumped, if `ClassFileDumper` uses `Path` for writing, `FileSystems::getDefault` would fail with `NullPointerException` because it calls `BootLoader.loadLibrary` which is not fully initialized. 

A separate issue will be filed to follow up the startup overhead of `OperatingSystem`.

---------
### Progress

### Issue
 * [JDK-8307944](https://bugs.openjdk.org/browse/JDK-8307944): ClassFileDumper should only load java.nio.file.Path if enabled

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14160/head:pull/14160` \
`$ git checkout pull/14160`

`$ git checkout pull/14160` \
`$ git pull https://git.openjdk.org/jdk.git pull/14160/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14160`

`$ git pr show -t 14160`

</details>

<a href="https://git.openjdk.org/jdk/pull/14160.diff">https://git.openjdk.org/jdk/pull/14160.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14160#issuecomment-1563486405)


src/java.base/share/classes/jdk/internal/util/ClassFileDumper.java
```diff
@@ -167,48 +129,62 @@ public void dumpFailedClass(String name, byte[] bytes) {
     }
 
     @SuppressWarnings("removal")
-    private void write(Path path, byte[] bytes) {
+    private void write(File file, byte[] bytes) {
+        // use java.io.File instead of java.nio.file.Path to avoid
+        // recursive initialization during early startup. e.g.
+        // Enum::valueOf causes method handle to be invoked which
+        // may cause classes to be spinned and dumped
```
RogerRiggs, line 136: This comment could be omitted; it won't age well and is out of context for the fix.

RogerRiggs (approved):
Looks good, delaying the checking of the file system until its is needed will save on startup.

mlchung:
#14181 decoupled `jdk.internal.util OperatingSystem` and `Version` classes such that the early startup would no longer load `OperatingSystem` class and avoid the cyclic initialization.  So `ClassFileDumper` is now reverted back to use `Path.

RogerRiggs (approved):
Path changes look fine.

PR #14169, opened 2023-05-26
8308931: Problemlist compiler/jvmci/TestUncaughtErrorInCompileMethod.java

dougxc:
In order to reduce the noise in the JDK 21 CI, I'm problem listing:

compiler/jvmci/TestUncaughtErrorInCompileMethod.java

which times out intermittently.


TobiHartmann (approved):
Looks good and trivial.

PR #14172, opened 2023-05-26
8312213: Remove unnecessary TEST instructions on x86 when flags reg will already be set

ichttt:
This patch adds peephole rules to remove TEST instructions that operate on the result and right after a AND, XOR or OR instruction.
This pattern can emerge if the result of the and is compared against two values where one of the values is zero. The matcher does not have the capability to know that the instruction mentioned above also set the flag register to the same value.
According to https://www.felixcloutier.com/x86/and, https://www.felixcloutier.com/x86/xor, https://www.felixcloutier.com/x86/or and https://www.felixcloutier.com/x86/test the flags are set to same values for TEST, AND, XOR and OR, so this should be safe.
By adding peephole rules to remove the TEST instructions, the resulting assembly code can be shortend and a small speedup can be observed:
Results on Intel Core i5-8250U CPU
Before this patch:
```
Benchmark                                              Mode  Cnt    Score    Error  Units
TestRemovalPeephole.benchmarkAndTestFusableInt         avgt    8  182.353 ±  1.751  ns/op
TestRemovalPeephole.benchmarkAndTestFusableIntSingle   avgt    8    1.110 ±  0.002  ns/op
TestRemovalPeephole.benchmarkAndTestFusableLong        avgt    8  212.836 ±  0.310  ns/op
TestRemovalPeephole.benchmarkAndTestFusableLongSingle  avgt    8    2.072 ±  0.002  ns/op
TestRemovalPeephole.benchmarkOrTestFusableInt          avgt    8   72.052 ±  0.215  ns/op
TestRemovalPeephole.benchmarkOrTestFusableIntSingle    avgt    8    1.406 ±  0.002  ns/op
TestRemovalPeephole.benchmarkOrTestFusableLong         avgt    8  113.396 ±  0.666  ns/op
TestRemovalPeephole.benchmarkOrTestFusableLongSingle   avgt    8    1.183 ±  0.001  ns/op
TestRemovalPeephole.benchmarkXorTestFusableInt         avgt    8   88.683 ±  2.034  ns/op
TestRemovalPeephole.benchmarkXorTestFusableIntSingle   avgt    8    1.406 ±  0.002  ns/op
TestRemovalPeephole.benchmarkXorTestFusableLong        avgt    8  113.271 ±  0.602  ns/op
TestRemovalPeephole.benchmarkXorTestFusableLongSingle  avgt    8    1.183 ±  0.001  ns/op
```
After this patch:
```
Benchmark                                              Mode  Cnt    Score   Error  Units  Change
TestRemovalPeephole.benchmarkAndTestFusableInt         avgt    8  141.615 ± 4.747  ns/op  ~29% faster
TestRemovalPeephole.benchmarkAndTestFusableIntSingle   avgt    8    1.110 ± 0.002  ns/op  (unchanged)
TestRemovalPeephole.benchmarkAndTestFusableLong        avgt    8  213.249 ± 1.094  ns/op  (unchanged)
TestRemovalPeephole.benchmarkAndTestFusableLongSingle  avgt    8    2.074 ± 0.011  ns/op  (unchanged)
TestRemovalPeephole.benchmarkOrTestFusableInt          avgt    8   91.500 ± 0.142  ns/op  ~27% slower
TestRemovalPeephole.benchmarkOrTestFusableIntSingle    avgt    8    0.814 ± 0.002  ns/op  ~45% faster
TestRemovalPeephole.benchmarkOrTestFusableLong         avgt    8  113.691 ± 0.839  ns/op  (unchanged)
TestRemovalPeephole.benchmarkOrTestFusableLongSingle   avgt    8    1.185 ± 0.001  ns/op  (unchanged)
TestRemovalPeephole.benchmarkXorTestFusableInt         avgt    8   84.953 ± 0.766  ns/op  ~4% faster
TestRemovalPeephole.benchmarkXorTestFusableIntSingle   avgt    8    0.814 ± 0.001  ns/op  ~72% faster
TestRemovalPeephole.benchmarkXorTestFusableLong        avgt    8  113.158 ± 0.629  ns/op  (unchanged)
TestRemovalPeephole.benchmarkXorTestFusableLongSingle  avgt    8    1.184 ± 0.002  ns/op  (unchanged)
```

Results from my i5 13600k:
Before this patch:
```
Benchmark                                              Mode  Cnt   Score    Error  Units
TestRemovalPeephole.benchmarkAndTestFusableInt         avgt    8  71,368 ±  0,857  ns/op
TestRemovalPeephole.benchmarkAndTestFusableIntSingle   avgt    8   0,430 ±  0,001  ns/op
TestRemovalPeephole.benchmarkAndTestFusableLong        avgt    8  66,734 ±  0,128  ns/op
TestRemovalPeephole.benchmarkAndTestFusableLongSingle  avgt    8   0,489 ±  0,001  ns/op
TestRemovalPeephole.benchmarkOrTestFusableInt          avgt    8  57,290 ±  0,200  ns/op
TestRemovalPeephole.benchmarkOrTestFusableIntSingle    avgt    8   0,390 ±  0,001  ns/op
TestRemovalPeephole.benchmarkOrTestFusableLong         avgt    8  70,606 ±  1,595  ns/op
TestRemovalPeephole.benchmarkOrTestFusableLongSingle   avgt    8   0,390 ±  0,001  ns/op
TestRemovalPeephole.benchmarkXorTestFusableInt         avgt    8  62,514 ±  0,735  ns/op
TestRemovalPeephole.benchmarkXorTestFusableIntSingle   avgt    8   0,391 ±  0,001  ns/op
TestRemovalPeephole.benchmarkXorTestFusableLong        avgt    8  70,886 ±  0,572  ns/op
TestRemovalPeephole.benchmarkXorTestFusableLongSingle  avgt    8   0,391 ±  0,001  ns/op
```
After this patch:
```
Benchmark                                              Mode  Cnt   Score   Error  Units  Change
TestRemovalPeephole.benchmarkAndTestFusableInt         avgt    8  65,083 ± 0,246  ns/op  ~10% faster
TestRemovalPeephole.benchmarkAndTestFusableIntSingle   avgt    8   0,429 ± 0,001  ns/op  (unchanged)
TestRemovalPeephole.benchmarkAndTestFusableLong        avgt    8  66,533 ± 0,096  ns/op  (unchanged)
TestRemovalPeephole.benchmarkAndTestFusableLongSingle  avgt    8   0,488 ± 0,001  ns/op  (unchanged)
TestRemovalPeephole.benchmarkOrTestFusableInt          avgt    8  74,042 ± 0,412  ns/op  ~10% slower
TestRemovalPeephole.benchmarkOrTestFusableIntSingle    avgt    8   0,391 ± 0,002  ns/op  (unchanged)
TestRemovalPeephole.benchmarkOrTestFusableLong         avgt    8  70,850 ± 0,342  ns/op  (unchanged)
TestRemovalPeephole.benchmarkOrTestFusableLongSingle   avgt    8   0,390 ± 0,002  ns/op  (unchanged)
TestRemovalPeephole.benchmarkXorTestFusableInt         avgt    8  56,626 ± 2,274  ns/op  ~10% faster
TestRemovalPeephole.benchmarkXorTestFusableIntSingle   avgt    8   0,392 ± 0,002  ns/op  (unchanged)
TestRemovalPeephole.benchmarkXorTestFusableLong        avgt    8  69,942 ± 1,380  ns/op  (unchanged)
TestRemovalPeephole.benchmarkXorTestFusableLongSingle  avgt    8   0,391 ± 0,001  ns/op  (unchanged)
```

As you can see, there are some cases where this PR the tests run slower, but especially on older architectures, the speed gains outrank the cases where some performance is lost by a lot. Also, the emitted instructuon sequence is still two bytes per removed test instruction shorter.
The long variants never see any speedup whatsoever, but as other architectures might observe a speedup here as well and the resulting assembly is at least 2 bytes shorter, I think the rules for long still makes sense.

---------
### Progress

### Issue
 * [JDK-8312213](https://bugs.openjdk.org/browse/JDK-8312213): Remove unnecessary TEST instructions on x86 when flags reg will already be set (**Enhancement** - P4)

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - **Reviewer**)
 * [Dean Long](https://openjdk.org/census#dlong) (@dean-long - **Reviewer**) ⚠️ Review applies to [af934150](https://git.openjdk.org/jdk/pull/14172/files/af934150b76a9da4ad1dbb7c2e0807287cfeb9ed)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14172/head:pull/14172` \
`$ git checkout pull/14172`

`$ git checkout pull/14172` \
`$ git pull https://git.openjdk.org/jdk.git pull/14172/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14172`

`$ git pr show -t 14172`

</details>

<a href="https://git.openjdk.org/jdk/pull/14172.diff">https://git.openjdk.org/jdk/pull/14172.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14172#issuecomment-1647940469)


merykitty:
Hi,

The `and` nodes have a `MachProj` output which is its flag output, the matcher does not have such capability but after peephole there is only local bundle scheduler. As a result, you can connect the outputs of the `test` directly to the `MachProj` output of the `and`.

Also, can this be also applied to `or`, `xor`, `not`, `andn`, `popcnt`, etc? I think it can also be applied to `add` and `sub` if the output of the `test` only reads ZF.

Thanks.

jaskarth:
Hmm, is it possible to add multiple `peepmatch` attributes to a single peephole currently? Since the `peepprocedure`s and `peepreplace`s in here are the same across data type, it may be nice to consolidate them in that way for maintainability, especially if more rules are added for the other operations as well.

ichttt:
Thanks for the feedback.
I will definitly look into other instructions such as or where this might be feasable.
Regarding the multiple peepmatchs: It looks like this is not possible with the current system, but I've just updated how the peepmatch block is interpreted in the test_may_remove peepprocedures. This way we can keep the rules consice

jaskarth (commented):
The changes to the ad file is nice! It's a lot more readable to me now. Is it possible to add some more entries to the JMH benchmark to show the results of the other `and` matches, such as the ones with memory? And do you need a JBS entry for this PR?

ichttt:
Thanks for the feedback!
I double checked, and it seems that other flags than the ZF are used as well. I've added some new cases to cover xor and or in addition to the existing and case, as they all set the same flags as TEST does.
@jaskarth Yeah I still need a JBS entry. If you can create on that would be nice, otherwise I'm going to use the webbug form again.

I benchmarked the new peephole rules and got the following results:
OLD:
```
Benchmark                                        Mode  Cnt   Score   Error  Units
TestRemovalPeephole.benchmarkAndTestFusableInt   avgt    8  71,351 ± 0,176  ns/op
TestRemovalPeephole.benchmarkAndTestFusableLong  avgt    8  66,763 ± 0,104  ns/op
TestRemovalPeephole.benchmarkOrTestFusableInt    avgt    8  57,555 ± 0,341  ns/op
TestRemovalPeephole.benchmarkOrTestFusableLong   avgt    8  72,158 ± 0,337  ns/op
TestRemovalPeephole.benchmarkXorTestFusableInt   avgt    8  61,551 ± 0,346  ns/op
TestRemovalPeephole.benchmarkXorTestFusableLong  avgt    8  70,554 ± 0,208  ns/op
```
NEW:
```
TestRemovalPeephole.benchmarkAndTestFusableInt   avgt    8  65,061 ± 0,096  ns/op  (10%)
TestRemovalPeephole.benchmarkAndTestFusableLong  avgt    8  66,855 ± 0,119  ns/op  (unchanged)
TestRemovalPeephole.benchmarkOrTestFusableInt    avgt    8  73,707 ± 0,385  ns/op  (-20%)
TestRemovalPeephole.benchmarkOrTestFusableLong   avgt    8  71,062 ± 1,190  ns/op  (unchanged)
TestRemovalPeephole.benchmarkXorTestFusableInt   avgt    8  56,420 ± 1,441  ns/op  (10%)
TestRemovalPeephole.benchmarkXorTestFusableLong  avgt    8  69,878 ± 1,318  ns/op  (unchanged)
```

I am currently not quite sure why there is a massive regression in the benchmarkOrTestFusableInt, but it seems to be reproducable at least on my machine. I plan to test another machine soon to see if this behaviour extends to other processors as well. The resulting assembly looks sane, though, and the XOR case also behaves as expected.

ichttt:
I've added a few more benchmarks that are basically the same as the old ones and reran the benchmark on my i5 13600k and my i5 8250U. Here are the results:
Results on Intel Core i5-8250U CPU
Before this patch:
```
Benchmark                                              Mode  Cnt    Score    Error  Units
TestRemovalPeephole.benchmarkAndTestFusableInt         avgt    8  182.353 ±  1.751  ns/op
TestRemovalPeephole.benchmarkAndTestFusableIntSingle   avgt    8    1.110 ±  0.002  ns/op
TestRemovalPeephole.benchmarkAndTestFusableLong        avgt    8  212.836 ±  0.310  ns/op
TestRemovalPeephole.benchmarkAndTestFusableLongSingle  avgt    8    2.072 ±  0.002  ns/op
TestRemovalPeephole.benchmarkOrTestFusableInt          avgt    8   72.052 ±  0.215  ns/op
TestRemovalPeephole.benchmarkOrTestFusableIntSingle    avgt    8    1.406 ±  0.002  ns/op
TestRemovalPeephole.benchmarkOrTestFusableLong         avgt    8  113.396 ±  0.666  ns/op
TestRemovalPeephole.benchmarkOrTestFusableLongSingle   avgt    8    1.183 ±  0.001  ns/op
TestRemovalPeephole.benchmarkXorTestFusableInt         avgt    8   88.683 ±  2.034  ns/op
TestRemovalPeephole.benchmarkXorTestFusableIntSingle   avgt    8    1.406 ±  0.002  ns/op
TestRemovalPeephole.benchmarkXorTestFusableLong        avgt    8  113.271 ±  0.602  ns/op
TestRemovalPeephole.benchmarkXorTestFusableLongSingle  avgt    8    1.183 ±  0.001  ns/op
```
After this patch:
```
Benchmark                                              Mode  Cnt    Score   Error  Units  Change
TestRemovalPeephole.benchmarkAndTestFusableInt         avgt    8  141.615 ± 4.747  ns/op  ~29% faster
TestRemovalPeephole.benchmarkAndTestFusableIntSingle   avgt    8    1.110 ± 0.002  ns/op  (unchanged)
TestRemovalPeephole.benchmarkAndTestFusableLong        avgt    8  213.249 ± 1.094  ns/op  (unchanged)
TestRemovalPeephole.benchmarkAndTestFusableLongSingle  avgt    8    2.074 ± 0.011  ns/op  (unchanged)
TestRemovalPeephole.benchmarkOrTestFusableInt          avgt    8   91.500 ± 0.142  ns/op  ~27% slower
TestRemovalPeephole.benchmarkOrTestFusableIntSingle    avgt    8    0.814 ± 0.002  ns/op  ~45% faster
TestRemovalPeephole.benchmarkOrTestFusableLong         avgt    8  113.691 ± 0.839  ns/op  (unchanged)
TestRemovalPeephole.benchmarkOrTestFusableLongSingle   avgt    8    1.185 ± 0.001  ns/op  (unchanged)
TestRemovalPeephole.benchmarkXorTestFusableInt         avgt    8   84.953 ± 0.766  ns/op  ~4% faster
TestRemovalPeephole.benchmarkXorTestFusableIntSingle   avgt    8    0.814 ± 0.001  ns/op  ~72% faster
TestRemovalPeephole.benchmarkXorTestFusableLong        avgt    8  113.158 ± 0.629  ns/op  (unchanged)
TestRemovalPeephole.benchmarkXorTestFusableLongSingle  avgt    8    1.184 ± 0.002  ns/op  (unchanged)
```

Results from my i5 13600k:
Before this patch:
```
Benchmark                                              Mode  Cnt   Score    Error  Units
TestRemovalPeephole.benchmarkAndTestFusableInt         avgt    8  71,368 ±  0,857  ns/op
TestRemovalPeephole.benchmarkAndTestFusableIntSingle   avgt    8   0,430 ±  0,001  ns/op
TestRemovalPeephole.benchmarkAndTestFusableLong        avgt    8  66,734 ±  0,128  ns/op
TestRemovalPeephole.benchmarkAndTestFusableLongSingle  avgt    8   0,489 ±  0,001  ns/op
TestRemovalPeephole.benchmarkOrTestFusableInt          avgt    8  57,290 ±  0,200  ns/op
TestRemovalPeephole.benchmarkOrTestFusableIntSingle    avgt    8   0,390 ±  0,001  ns/op
TestRemovalPeephole.benchmarkOrTestFusableLong         avgt    8  70,606 ±  1,595  ns/op
TestRemovalPeephole.benchmarkOrTestFusableLongSingle   avgt    8   0,390 ±  0,001  ns/op
TestRemovalPeephole.benchmarkXorTestFusableInt         avgt    8  62,514 ±  0,735  ns/op
TestRemovalPeephole.benchmarkXorTestFusableIntSingle   avgt    8   0,391 ±  0,001  ns/op
TestRemovalPeephole.benchmarkXorTestFusableLong        avgt    8  70,886 ±  0,572  ns/op
TestRemovalPeephole.benchmarkXorTestFusableLongSingle  avgt    8   0,391 ±  0,001  ns/op
```
After this patch:
```
Benchmark                                              Mode  Cnt   Score   Error  Units  Change
TestRemovalPeephole.benchmarkAndTestFusableInt         avgt    8  65,083 ± 0,246  ns/op  ~10% faster
TestRemovalPeephole.benchmarkAndTestFusableIntSingle   avgt    8   0,429 ± 0,001  ns/op  (unchanged)
TestRemovalPeephole.benchmarkAndTestFusableLong        avgt    8  66,533 ± 0,096  ns/op  (unchanged)
TestRemovalPeephole.benchmarkAndTestFusableLongSingle  avgt    8   0,488 ± 0,001  ns/op  (unchanged)
TestRemovalPeephole.benchmarkOrTestFusableInt          avgt    8  74,042 ± 0,412  ns/op  ~10% slower
TestRemovalPeephole.benchmarkOrTestFusableIntSingle    avgt    8   0,391 ± 0,002  ns/op  (unchanged)
TestRemovalPeephole.benchmarkOrTestFusableLong         avgt    8  70,850 ± 0,342  ns/op  (unchanged)
TestRemovalPeephole.benchmarkOrTestFusableLongSingle   avgt    8   0,390 ± 0,002  ns/op  (unchanged)
TestRemovalPeephole.benchmarkXorTestFusableInt         avgt    8  56,626 ± 2,274  ns/op  ~10% faster
TestRemovalPeephole.benchmarkXorTestFusableIntSingle   avgt    8   0,392 ± 0,002  ns/op  (unchanged)
TestRemovalPeephole.benchmarkXorTestFusableLong        avgt    8  69,942 ± 1,380  ns/op  (unchanged)
TestRemovalPeephole.benchmarkXorTestFusableLongSingle  avgt    8   0,391 ± 0,001  ns/op  (unchanged)
```

As you can see, there are still some cases where this PR the tests run slower, but especially on older architectures, the speed gains outrank the cases where some performance is lost by a lot. Also, the emitted instructuon sequence is still two bytes per removed test instruction shorter.

merykitty:
@JornVernee I saw you created [JDK-8311969](https://bugs.openjdk.org/projects/JDK/issues/JDK-8311969), you may want to take a look at this.

JornVernee:
@merykitty Thanks for the pointer!

src/hotspot/cpu/x86/x86_64.ad
```diff
@@ -13772,6 +13772,28 @@ instruct leaL_rReg_immI2_peep(rRegL dst, rRegL src, immI2 shift)
...
+%{
+  predicate(false);
+  match(Set cr (CmpL src zero));
+
+  format %{ "# noop testL - cr set by and" %}
+  ins_encode %{
+  %}
+  ins_pipe(ialu_reg);
+%}
+
```
JornVernee, line 13796: In `test_may_remove_5` the `new_root` function points to the constructor of one of these instructions (what's passed to `peepreplace`). Since the function pointer is unused, these are really not needed. (for `peepreplace` you can just pass `testI_reg` or `testL_reg`, it doesn't really matter).

JornVernee (commented):
> Also, can this be also applied to or, xor, not, andn, popcnt, etc? I think it can also be applied to add and sub if the output of the test only read ZF.

I agree there's potentially more that can be done here. I think one issue is that we don't know which of the flags are actually used by the instruction downstream of the `test` (e.g. `jl`, `jg`, whatever). So, I think we should match the peephole rule on the downstream instructions instead, basically any instruction that accepts a `cmpOp` operator (or a variant that).

We can track per instruction/rule which flags they set (for all instructions that `KILL cr`), and check whether the flags required by the downstream instruction match the flags set by the instruction feeding into the `test`.

src/hotspot/cpu/x86/x86_64.ad
JornVernee: Potentially we could modify ADLC to make `peepreplace` optional, and then pass `nullptr` to the peepprocedure.

ichttt:
Thanks for the feedback!
I agree that this is still far from a perfect solution and tracking the required flags would be ideal, but there many rules that use the flags register, and all of these would need to be tracked. If you are interested in that, feel free to take this PR as a starting point, but I'm currently not interested in pursuing that path. Otherwise we can still keep this PR in its current form IMO as a first step to removing redudant test instructions.

merykitty:
@ichttt Yes, I don't think tracking all the nodes that use flags is elegant, you can get the use from the test node, then iterate the operands of the use to find the `cmpOp` operand which you can know which flag it is using. For most nodes, this will be enough, and for `add`/`sub` the transformation can be applied if `cmpOp->ccode() == Assembler::zero || cmpOp->ccode() == Assembler::notZero`. I think try to do this for `add/sub` would be beneficial since they are more common and they can be macro-fuse with `jcc`.

chhagedorn:
JBS issue: https://bugs.openjdk.org/browse/JDK-8312213

ichttt:
I've updated the PR to follow the approach suggested by @JornVernee . The AD files now contains the information regarding which flag from the EFLAGS register is actually set or cleared.
The only downside of this I currently see is that it uses a lot of space from the flags register, which now only has space for three more flags in the future. Not all the flags that are set in the AD file are currently used, but I thoght it might be cleaner to list all of them instead of only those needed for the test removal.
I've tested this locally and it seems to work as intended. We also now have less overhead in the peephole phase as there are only two peepholes remaing (one of testI_reg and one for testL_reg) instead of 6.

merykitty:
In general I think this approach is very elegant. A point I want to address is the definition of setting a flag, a `test` can be elided after an `and` because the `and` sets the flag according to **the result**, instructions can set flags based on other factors, such as `add` sets OF based on the operation, `bsr` sets ZF based on the input, etc. As a result, I think for those that does not set the flags based on the result, it should be marked separately, and anything else should be marked as cloberring.

Thanks.

merykitty:
IR tests can match machine nodes so please do so, ideally for both branching and conditional moving. Thank a lot.

merykitty:
I see, personally I think OF should be marked as being clobbered for `add` because it is really specific to addition that it overflows and not on the result of that addition like `ZF` or `SF`. Thanks.

dean-long:
I don't see where the peephole rule is checking that the test instruction is testing the same register that was the destination of the earlier instruction that set the flags.  Also, is there an example of an instruction annotated with the new flag information that does NOT set the required flags?  If not, then I don't see why we need to track individual flags.

ichttt:
The peephole only applies the instruction that would be emitted immediately before the test is the input of the test. If that is the case, it must operate on the same register as the result of the above instruction. We then check the flags of the nodes.
We need all the information cause specific operation require specific flags. For example, if the instruction following the test checks if the value is greater (than zero), the sign and zero flag need to be set and the overflow flag needs to be cleared. Not all instructions (such as add) satisfy this requirement, so in this case we would need to emit the test, but we could omit it if we only check for zero, as that only requires the ZF, which the test instruction sets.


I also noticed another problem during the construction of the IR tests: Test instructions before conditional moves are currently not removed. This is due to the Matcher thinking that it needs to load a zero into one register when it emits a setb instruction. This is not the case, but the matcher does not know this and 1) emits a pointless register clear and 2) still has the loadConI0 in its graph, which causes the test peephole to bail out as its input is not the preceeding instruction.
I think the removal of the loadConI0 in this case is a topic for another PR though. Fixing this will result in the peephole also working for setb instructions, which would be a huge win.

idealgraphvisualizer64_c1r2gDJIi7

merykitty:
@ichttt The register clear via `xor` is mandatory since `setb` only sets the lowest byte. We can be more clever and do more complex transformations such as relaxing consecutiveness requirement, reordering nodes, converting a `xor` into a `mov`, etc but I think it can be achieved in another patch and this patch covers the most prominent appearances of redundant `test`. Thanks a lot.

dean-long:
> For example, if the instruction following the test checks if the value is greater (than zero), the sign and zero flag need to be set and the overflow flag needs to be cleared. Not all instructions (such as add) satisfy this requirement, so in this case we would need to emit the test, but we could omit it if we only check for zero, as that only requires the ZF, which the test instruction sets.

OK, I see your concern about ADD. Rather than reject it completely, you allow it depending on the kind of compare, which requires tracking individual flags.

dean-long (approved):

src/hotspot/share/adlc/adlparse.cpp
```diff
@@ -229,6 +229,7 @@ void ADLParser::instr_parse(void) {
     else if (!strcmp(ident, "opcode"))           instr->_opcode    = opcode_parse(instr);
     else if (!strcmp(ident, "size"))             instr->_size      = size_parse(instr);
     else if (!strcmp(ident, "effect"))           effect_parse(instr);
+    else if (!strcmp(ident, "flag"))             instr->_flag = flag_parse(instr);
```
JornVernee, line 232:
```suggestion
    else if (!strcmp(ident, "size"))             instr->_size      = size_parse(instr);
    else if (!strcmp(ident, "effect"))           effect_parse(instr);
    else if (!strcmp(ident, "flag"))             instr->_flag      = flag_parse(instr);
```

src/hotspot/share/adlc/output_c.cpp
```diff
@@ -4010,6 +4014,22 @@ void ArchDesc::buildMachNode(FILE *fp_cpp, InstructForm *inst, const char *inden
     fprintf(fp_cpp, " );\n");
     // #####
   }
+  if (inst->_flag != nullptr) {
+    Flag* node = inst->_flag;
+    const char* prefix = "Node::";
```
JornVernee, line 4019: You could potentially make the prefix here `Node::PD::`, then the extra `PD::` could be removed from the .ad file (I don't think it really adds much?).

src/hotspot/share/adlc/output_c.cpp
```diff
@@ -4010,6 +4014,22 @@ void ArchDesc::buildMachNode(FILE *fp_cpp, InstructForm *inst, const char *inden
...
+    bool node_flags_set = false;
+    do {
+      if (!node_flags_set) {
+        fprintf(fp_cpp, "%s node->add_flag(%s%s", indent, strncmp(node->_name, prefix, strlen(prefix)) != 0 ? prefix : "", node->_name);
```
JornVernee, line 4023: This seems to be guarding against a case where the flag is declared with the prefix already in the .ad file. Is this required for something?

(Otherwise I suggest just using `node->_name` here, as it forces the flag declarations in the .ad file to be consistent).

src/hotspot/cpu/x86/x86.ad
```diff
@@ -1254,8 +1254,18 @@ static inline bool is_clz_non_subword_predicate_evex(BasicType bt, int vlen_byte
...
+    Flag_sets_parity_flag     = Node::_last_flag << 3,
+    Flag_sets_zero_flag       = Node::_last_flag << 4,
+    Flag_sets_overflow_flag   = Node::_last_flag << 5,
+    Flag_sets_sign_flag       = Node::_last_flag << 6,
+    Flag_clears_carry_flag    = Node::_last_flag << 7,
+    Flag_clears_parity_flag   = Node::_last_flag << 8,
+    Flag_clears_zero_flag     = Node::_last_flag << 9,
+    Flag_clears_overflow_flag = Node::_last_flag << 10,
+    Flag_clears_sign_flag     = Node::_last_flag << 11,
+    _last_flag                = Flag_clears_sign_flag
```
JornVernee, line 1268: I think adding the flags here is good. If the number of flags becomes a problem, we could instead generate virtual methods on all the nodes to return the flag mask.

src/hotspot/cpu/x86/x86_64.ad
```diff
@@ -7673,6 +7675,8 @@ instruct addI_rReg_mem(rRegI dst, memory src, rFlagsReg cr)
 %{
   match(Set dst (AddI dst (LoadI src)));
   effect(KILL cr);
+  flag(PD::Flag_sets_overflow_flag, PD::Flag_sets_sign_flag, PD::Flag_sets_zero_flag, PD::Flag_sets_carry_flag, PD::Flag_sets_parity_flag);
+
```
JornVernee, line 7679: Please remove these extra blank lines. On `addI_rReg_mem`, `addI_mem_rReg`, and `blsrL_rReg_rReg`.

src/hotspot/cpu/x86/peephole_x86_64.cpp
```diff
@@ -132,6 +133,109 @@ bool lea_coalesce_helper(Block* block, int block_index, PhaseCFG* cfg_, PhaseReg
...
+    if (prevNode != nullptr) {
+      // Includes other flags as well, but that doesn't matter here
+      juint all_node_flags = prevNode->flags();
+      if (all_node_flags == 0) {
+        // We can return early - there is no way the test can be removed, the preceding node does not set any flags
+        return false;
+      }
+      juint required_flags = 0;
+      // Search for the uses of the node and compute which flags are required
+      for (DUIterator i = test_to_check->outs(); test_to_check->has_out(i); i++) {
```
JornVernee, line 197: AFAICS `fast_outs` can be used here, since you don't modify the `_opnds`.

test/hotspot/jtreg/compiler/c2/irTests/TestTestRemovalPeephole.java
```diff
@@ -0,0 +1,142 @@
...
+package compiler.c2.irTests;
+
+import jdk.test.lib.Asserts;
+import compiler.lib.ir_framework.*;
+import java.util.Random;
+import jdk.test.lib.Utils;
+
+/*
+ * @test
+ * @summary Test that patterns leading to Conv2B are correctly expanded.
```
JornVernee, line 33: Summary seems to be incorrect.

JornVernee (commented):
Thanks for the updated version. (Sorry that it took me a while to get back to this)

Left a few small inline comments. What testing did you run on this? (I'll also run tier 1-4 in our CI)

src/hotspot/share/adlc/output_c.cpp
ichttt: No, this is just a leftover from an earlier design, I'll remove the check and always add the prefix

src/hotspot/share/adlc/output_c.cpp
ichttt: Well I thought that it could also be used to add some non-arch specific flags in the future, and with this the keyword would be more generic and allow for this as well.

JornVernee:
The new test is failing in our CI. The failing cases are:

```
testIntAddtionEquals0(int,int)
testIntAddtionNotEquals0(int,int)
testIntOrEquals0(int,int)
testIntOrGreater0(int,int)
testIntOrNotEquals0(int,int)
testLongOrGreater0(long,long)
```

Looks like all of them are failing with interfering `loadConI0` or `loadConL0` nodes. e.g. for `testIntAddtionEquals0(int,int)`:

```
RDX     11  addI_rReg  === _ 15 16  [[ 12 10 ]] 
RFLAGS  12  MachProj  === 11  [[ ]] #1
RAX     13  loadConI0  === 1  [[ 14 9 ]] #0/0x00000000
RFLAGS  14  MachProj  === 13  [[ ]] #1
RFLAGS  10  testI_reg  === _ 11  [[ 9 ]] #0/0x00000000
RAX      9  cmovI_imm_01  === _ 10 13  [[ 2 ]] ne#1/0x00000001 !jvms: TestTestRemovalPeephole::testIntAddtionEquals0 @ bci:13 (line 50)
```

The failure occurs in the same way on several platforms.

ichttt:
Hmm yeah then it is emitting a cmov, which is currently not working due to reasons outlined above. I've changed the tests so they always produce a branch.
Regarding Testing on my own: I tested the previous commit using GHA and ran the tier 1 tests on my linux machine.

JornVernee:
Ok, thanks. I've submitted another CI run.

TobiHartmann:
@JornVernee's testing failed:

`-ea -esa -XX:CompileThreshold=100 -XX:+UnlockExperimentalVMOptions -server -XX:-TieredCompilation`


```
Failed IR Rules (6) of Methods (6)
----------------------------------
1) Method "public boolean compiler.c2.irTests.TestTestRemovalPeephole.testIntAddtionEquals0(int,int)" - [Failed IR rules: 1]:
   * @IR rule 1: "@compiler.lib.ir_framework.IR(applyIfCPUFeatureAnd={}, phase={FINAL_CODE}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={}, applyIfAnd={}, failOn={"_#X86_TESTI_REG#_", "_#X86_TESTL_REG#_"}, applyIfOr={}, applyIfNot={})"
     > Phase "Final Code":
       - failOn: Graph contains forbidden nodes:
         * Constraint 1: "(\\d+(\\s){2}(testI_reg.*)+(\\s){2}===.*)"
           - Matched forbidden node:
             * 10  testI_reg  === _ 11  [[ 9 ]] #0/0x00000000

2) Method "public boolean compiler.c2.irTests.TestTestRemovalPeephole.testIntAddtionNotEquals0(int,int)" - [Failed IR rules: 1]:
   * @IR rule 1: "@compiler.lib.ir_framework.IR(applyIfCPUFeatureAnd={}, phase={FINAL_CODE}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={}, applyIfAnd={}, failOn={"_#X86_TESTI_REG#_", "_#X86_TESTL_REG#_"}, applyIfOr={}, applyIfNot={})"
     > Phase "Final Code":
       - failOn: Graph contains forbidden nodes:
         * Constraint 1: "(\\d+(\\s){2}(testI_reg.*)+(\\s){2}===.*)"
           - Matched forbidden node:
             * 10  testI_reg  === _ 11  [[ 9 ]] #0/0x00000000

3) Method "public boolean compiler.c2.irTests.TestTestRemovalPeephole.testIntOrEquals0(int,int)" - [Failed IR rules: 1]:
   * @IR rule 1: "@compiler.lib.ir_framework.IR(applyIfCPUFeatureAnd={}, phase={FINAL_CODE}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={}, applyIfAnd={}, failOn={"_#X86_TESTI_REG#_", "_#X86_TESTL_REG#_"}, applyIfOr={}, applyIfNot={})"
     > Phase "Final Code":
       - failOn: Graph contains forbidden nodes:
         * Constraint 1: "(\\d+(\\s){2}(testI_reg.*)+(\\s){2}===.*)"
           - Matched forbidden node:
             * 10  testI_reg  === _ 11  [[ 9 ]] #0/0x00000000

4) Method "public boolean compiler.c2.irTests.TestTestRemovalPeephole.testIntOrGreater0(int,int)" - [Failed IR rules: 1]:
   * @IR rule 1: "@compiler.lib.ir_framework.IR(applyIfCPUFeatureAnd={}, phase={FINAL_CODE}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={}, applyIfAnd={}, failOn={"_#X86_TESTI_REG#_", "_#X86_TESTL_REG#_"}, applyIfOr={}, applyIfNot={})"
     > Phase "Final Code":
       - failOn: Graph contains forbidden nodes:
         * Constraint 1: "(\\d+(\\s){2}(testI_reg.*)+(\\s){2}===.*)"
           - Matched forbidden node:
             * 10  testI_reg  === _ 11  [[ 9 ]] #0/0x00000000

5) Method "public boolean compiler.c2.irTests.TestTestRemovalPeephole.testIntOrNotEquals0(int,int)" - [Failed IR rules: 1]:
   * @IR rule 1: "@compiler.lib.ir_framework.IR(applyIfCPUFeatureAnd={}, phase={FINAL_CODE}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={}, applyIfAnd={}, failOn={"_#X86_TESTI_REG#_", "_#X86_TESTL_REG#_"}, applyIfOr={}, applyIfNot={})"
     > Phase "Final Code":
       - failOn: Graph contains forbidden nodes:
         * Constraint 1: "(\\d+(\\s){2}(testI_reg.*)+(\\s){2}===.*)"
           - Matched forbidden node:
             * 10  testI_reg  === _ 11  [[ 9 ]] #0/0x00000000

6) Method "public boolean compiler.c2.irTests.TestTestRemovalPeephole.testLongOrGreater0(long,long)" - [Failed IR rules: 1]:
   * @IR rule 1: "@compiler.lib.ir_framework.IR(applyIfCPUFeatureAnd={}, phase={FINAL_CODE}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={}, applyIfAnd={}, failOn={"_#X86_TESTI_REG#_", "_#X86_TESTL_REG#_"}, applyIfOr={}, applyIfNot={})"
     > Phase "Final Code":
       - failOn: Graph contains forbidden nodes:
         * Constraint 2: "(\\d+(\\s){2}(testL_reg.*)+(\\s){2}===.*)"
           - Matched forbidden node:
             * 10  testL_reg  === _ 11  [[ 9 ]] #0/0x0000000000000000
```

JornVernee:
Hey @TobiHartmann that's the old test run for commit hash `af93415`, the newer one for commit `9872e71` came back clean (there are just 3 known failures).

TobiHartmann:
Ah, sorry for the confusion then :)

JornVernee:
I think this looks mostly good now from my perspective. It would be nice to know why there's a regression in the `benchmarkOrTestFusableInt` case though. I'll try to reproduce here as well.

src/hotspot/cpu/x86/peephole_x86_64.cpp
```diff
@@ -132,6 +133,109 @@ bool lea_coalesce_helper(Block* block, int block_index, PhaseCFG* cfg_, PhaseReg
...
+      ShouldNotReachHere();
+      return 0;
+  }
+}
+
+
+// This function removes the TEST instruction when it detected shapes likes AND r1, r2; TEST r1, r1
+// It checks the required EFLAGS for the downstream instructions of the TEST
+// and removes the TEST if the preceding instructions already sets all these flags
+bool Peephole::test_may_remove(Block* block, int block_index, PhaseCFG* cfg_, PhaseRegAlloc* ra_,
```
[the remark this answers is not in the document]
jatin-bhateja: I was curious to see impact of this patch on micro fusion.  Sapphire rapids has an explicit event for it and shows 3x drop in number of micro fused operations with attached test, not much impact on performance. 
[micro.txt](https://github.com/openjdk/jdk/files/12312909/micro.txt)
PR14172

ichttt:
Sorry for the long delay. I am not quite sure why there is so much deviation in the benchmark results and why release seems to be much slower than fastdebug.
But I've double-checked, and C++ compilers like clang also do not emit test instructions after instructions that already set the appropiate registers (See https://godbolt.org/z/n1K1v5YT9). That being said, it is mostly a size-releated optimization on newer machines, but as the benchmarks show a slight increase in performance in the average case. So IMO this is still a good optimization to implement.

JornVernee:
> Sorry for the long delay. I am not quite sure why there is so much deviation in the benchmark results and why release seems to be much slower than fastdebug. But I've double-checked, and C++ compilers like clang also do not emit test instructions after instructions that already set the appropiate registers (See https://godbolt.org/z/n1K1v5YT9). That being said, it is mostly a size-releated optimization on newer machines, but as the benchmarks show a slight increase in performance in the average case. So IMO this is still a good optimization to implement.

Okay, let's take the size win then.

JornVernee (approved):

JornVernee:
I'm running one last test job. I'll sponsor when that comes back green

PR #14177, opened 2023-05-26
8308336: Test java/net/HttpURLConnection/HttpURLConnectionExpectContinueTest.java failed: java.net.BindException: Address already in use

DarraghClarke:
`HttpURLConnectionExpectContinueTest` was throwing an error due to port being hardcoded, updated test to let the system decide which port to use.

---------
### Progress

### Issue
 * [JDK-8308336](https://bugs.openjdk.org/browse/JDK-8308336): Test java/net/HttpURLConnection/HttpURLConnectionExpectContinueTest.java failed: java.net.BindException: Address already in use (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14177/head:pull/14177` \
`$ git checkout pull/14177`

`$ git checkout pull/14177` \
`$ git pull https://git.openjdk.org/jdk.git pull/14177/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14177`

`$ git pr show -t 14177`

</details>

<a href="https://git.openjdk.org/jdk/pull/14177.diff">https://git.openjdk.org/jdk/pull/14177.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14177#issuecomment-1564365957)

### DIFF
--- a/test/jdk/java/net/HttpURLConnection/HttpURLConnectionExpectContinueTest.java
+++ b/test/jdk/java/net/HttpURLConnection/HttpURLConnectionExpectContinueTest.java
@@ -73,7 +73,7 @@ public class HttpURLConnectionExpectContinueTest {
 
         control.serverSocket = new ServerSocket();
         control.serverSocket.setReuseAddress(true);
-        control.serverSocket.bind(new InetSocketAddress("127.0.0.1", 54321));
+        control.serverSocket.bind(new InetSocketAddress("127.0.0.1", 0));
         Runnable runnable = () -> {
             while (!control.stop) {
                 try {
--- a/test/jdk/java/net/HttpURLConnection/HttpURLConnectionExpectContinueTest.java
+++ b/test/jdk/java/net/HttpURLConnection/HttpURLConnectionExpectContinueTest.java
@@ -425,7 +425,7 @@ public class HttpURLConnectionExpectContinueTest {
     private HttpURLConnection createConnection() throws Exception {
         URL url = URIBuilder.newBuilder()
                 .scheme("http")
-                .host(InetAddress.getLoopbackAddress())
+                .loopback()
                 .port(control.serverSocket.getLocalPort())
                 .toURL();
 
--- a/test/jdk/java/net/HttpURLConnection/HttpURLConnectionExpectContinueTest.java
+++ b/test/jdk/java/net/HttpURLConnection/HttpURLConnectionExpectContinueTest.java
@@ -27,8 +27,8 @@
  * @summary Verify that expect 100-continue doesn't hang
  * @library /test/lib
  * @run junit/othervm HttpURLConnectionExpectContinueTest
- * @run junit/othervm -Djava.net.preferIPv4Stack=True HttpURLConnectionExpectContinueTest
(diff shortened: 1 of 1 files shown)


test/jdk/java/net/HttpURLConnection/HttpURLConnectionExpectContinueTest.java
```diff
@@ -73,7 +73,7 @@ public void startServerSocket() throws Exception {
 
         control.serverSocket = new ServerSocket();
         control.serverSocket.setReuseAddress(true);
-        control.serverSocket.bind(new InetSocketAddress("127.0.0.1", 54321));
+        control.serverSocket.bind(new InetSocketAddress("127.0.0.1", 0));
```
dfuch, line 76: While you're at it can you change the address to use `InetAddress.getLoopbackAddress()`here, and the URIBuilder in createConnection() below?

msheppar: based on Daniel's last comment and requested change then I think it would be useful to add two additional run requests to the test which set the network properties  java.net.preferIPv4Stack=true and java.net.preferIPv6Addresses=true, for completeness.
You never know what might "pop out of the woodwork"

test/jdk/java/net/HttpURLConnection/HttpURLConnectionExpectContinueTest.java
```diff
@@ -419,8 +422,12 @@ public void testFixedLengthRequestWithDoubleExpect100ContinueResponse() throws E
     }
 
     // Creates a connection with all the common settings used in each test
-    private HttpURLConnection createConnection() throws IOException {
-        URL url = new URL("http://localhost:" + control.serverSocket.getLocalPort());
+    private HttpURLConnection createConnection() throws Exception {
+        URL url = URIBuilder.newBuilder()
+                .scheme("http")
+                .host(InetAddress.getLoopbackAddress())
```
msheppar, line 428: This will change the call flow of the test in a slightly subtle way. In the original the "host" is supplied i.e. localhost which should map to the loopback address, and in the change the loopback IP address is being supplied diectly. In terms of equivalence then supplying a host string might be more appropriate:
.host("localhost")

dfuch: No that's precisely what we want to avoid. Because how "localhost" maps to an InetAddress depends on the machine configuration, which is a recipe for intermittent failures.

I'd suggest:
```
        URL url = URIBuilder.newBuilder()
                .scheme("http")
                .localhost()
```
(which is actually the same)

DarraghClarke: At a glance there isn't a `.localhost()` only a `.loopback()` is that what you were referring to?

dfuch (approved):
LGTM.

DarraghClarke:
Spent some time double checking tests and everything seems to be stable

PR #14180, opened 2023-05-26
8308899: Introduce Classfile context and improve Classfile options

asotona:
Classfile context object and multi-state options have been discussed at https://mail.openjdk.org/pipermail/classfile-api-dev/2023-May/000321.html
This patch implements the proposed changes in Classfile API and fixes all affected code across JDK sources and tests.

Please review.

Thanks,
Adam

---------
### Progress

### Issues
 * [JDK-8308899](https://bugs.openjdk.org/browse/JDK-8308899): Introduce Classfile context and improve Classfile options (**Sub-task** - P4)
 * [JDK-8306650](https://bugs.openjdk.org/browse/JDK-8306650): Improve control of stack maps generation in Classfile API (**Enhancement** - P4)
 * [JDK-8308646](https://bugs.openjdk.org/browse/JDK-8308646): Typo in ConstantValueAttribute (**Bug** - P5)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - **Reviewer**) ⚠️ Review applies to [5723e66a](https://git.openjdk.org/jdk/pull/14180/files/5723e66ac9204671d942a6b5f49310112e4fd2e9)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14180/head:pull/14180` \
`$ git checkout pull/14180`

`$ git checkout pull/14180` \
`$ git pull https://git.openjdk.org/jdk.git pull/14180/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14180`

`$ git pr show -t 14180`

</details>

<a href="https://git.openjdk.org/jdk/pull/14180.diff">https://git.openjdk.org/jdk/pull/14180.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14180#issuecomment-1572201180)


liach (commented):
A side comment about the API changes:
I think we should recommend users to store Classfile instances in static final variables if sharing is reasonable, like in Module and BindingSpecializer; this way, we can cache class hierarchy resolution information across builds.

src/java.base/share/classes/jdk/internal/classfile/Classfile.java
```diff
@@ -57,133 +58,163 @@
...
      * Parse a classfile into a {@link ClassModel}.
-     * @param path the path to the classfile
-     * @param options the desired processing options
+     * @param bytes the bytes of the classfile
      * @return the class model
      */
-    public static ClassModel parse(Path path, Option... options) throws IOException {
-        return parse(Files.readAllBytes(path), options);
+    default ClassModel parse(byte[] bytes) {
+        return new ClassImpl(bytes, (ClassfileImpl)this);
```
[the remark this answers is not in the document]
asotona: yes, several methods should move to ClassfileImpl, thanks.

src/java.base/share/classes/jdk/internal/classfile/components/ClassRemapper.java
```diff
@@ -100,7 +100,7 @@ static ClassRemapper of(Function<ClassDesc, ClassDesc> mapFunction) {
      * @return re-mapped class file bytes
      */
     default byte[] remapClass(ClassModel clm) {
```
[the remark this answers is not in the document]
asotona: yes, I'll add the parameter, thanks.

src/java.base/share/classes/java/lang/Module.java
```diff
@@ -1590,8 +1590,8 @@ private Class<?> loadModuleInfoClass() {
      */
     private Class<?> loadModuleInfoClass(InputStream in) throws IOException {
         final String MODULE_INFO = "module-info";
-        byte[] bytes = Classfile.parse(in.readAllBytes(),
-                Classfile.Option.constantPoolSharing(false)).transform((clb, cle) -> {
+        var cc = Classfile.of(Classfile.ConstantPoolSharingOption.DO_NOT_SHARE_CONSTANT_POOL);
```
liach, line 1593: This `cc` can be stored in a static final field instead.

src/java.base/share/classes/jdk/internal/classfile/Classfile.java
```diff
@@ -28,175 +28,204 @@
...
 /**
  * Main entry points for parsing, transforming, and generating classfiles.
  */
-public class Classfile {
-    private Classfile() {
+public sealed interface Classfile
+        permits ClassfileImpl {
+
+    static Classfile of() {
+        return new ClassfileImpl();
```
liach, line 55: We can create a static final field in `ClassfileImpl` holding a default instance equivalent to that created by `new ClassfileImpl()` and have the `of()` factory return that instance instead.

src/java.base/share/classes/jdk/internal/classfile/impl/StackMapGenerator.java
```diff
@@ -832,22 +836,23 @@ private void generatorError(String msg, int offset) {
...
-                            b.writeU2(-1);//max locals
-                            b.writeInt(bytecode.limit());
-                            b.writeBytes(bytecode.array(), 0, bytecode.limit());
-                            b.writeU2(0);//exception handlers
-                            b.writeU2(0);//attributes
-                        }
-                    }));
-            ClassPrinter.toYaml(Classfile.parse(clb.build()).methods().get(0).code().get(), ClassPrinter.Verbosity.TRACE_ALL, sb::append);
+            //clone SplitConstantPool with alternate context
+            var cc = Classfile.of(Classfile.StackMapsOption.DO_NOT_GENERATE_STACK_MAPS);
```
liach, line 840: This should set the Class hierarchy resolver, patch dead code, and filter dead labels options from the original context.

liach:
In fact, I think we can probably mark Classfile as value-based. Then, whether it's created from a factory at every call or if it's cached (my first 2 review comments) no longer matters as much; the caching behavior will be handled by individual option instances (such as class hierarchy caching in the class hierarchy option's resolver)

src/java.base/share/classes/java/lang/Module.java
asotona: Yes, we can avoid repeated construction of default class hierarchy resolver here.

src/java.base/share/classes/jdk/internal/classfile/Classfile.java
asotona: Global static default context will hold global shared cached class hierarchy resolver and that we want to avoid.

src/java.base/share/classes/jdk/internal/classfile/impl/StackMapGenerator.java
asotona: Here the other options are irrelevant. The error report wraps already generated incomplete code attribute into a fake class to be able to print details to the error.
I think even the stack map generation suppression is irrelevant here, will double check.

src/java.base/share/classes/java/lang/Module.java
asotona: In this situation the ClassHierarchyResolver is never used and have no caching function.
Adding static field with initializer does not make much sense here.

src/java.base/share/classes/jdk/internal/classfile/constantpool/ConstantPoolBuilder.java
```diff
@@ -63,29 +63,22 @@
...
-     * will be empty and have the specified classfile processing options.
-     *
-     * @param options the processing options
+     * will be empty.
      */
-    static ConstantPoolBuilder of(Collection<Classfile.Option> options) {
-        return new SplitConstantPool(new Options(options));
+    static ConstantPoolBuilder of() {
+        return new SplitConstantPool();
     }
```
briangoetz, line 82: The old ctors here were (ClassModel) and (Options), where the ClassModel version had a trail of breadcrumbs back to options.  The new ctors are (ClassModel) and no-args, which is a slight departure.  What use did CPB make of options in the first place, if any?

briangoetz (commented):
Overall this looks pretty clean.  I like how simply it flows through the implementation classes.  There are a number of methods and classes, especially in Classfile, which need some more spec.

src/java.base/share/classes/jdk/internal/classfile/Classfile.java
```diff
@@ -28,175 +28,204 @@
...
 import java.lang.reflect.AccessFlag;
 import jdk.internal.classfile.attribute.CharacterRangeInfo;
 import jdk.internal.classfile.attribute.LocalVariableInfo;
 import jdk.internal.classfile.attribute.LocalVariableTypeInfo;
 import jdk.internal.classfile.instruction.ExceptionCatch;
-import java.lang.constant.PackageDesc;
 
 /**
  * Main entry points for parsing, transforming, and generating classfiles.
  */
```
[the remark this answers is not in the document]
asotona: Fixed, thanks.

src/java.base/share/classes/jdk/internal/classfile/Classfile.java
```diff
@@ -28,175 +28,204 @@
...
+    static Classfile of() {
+        return new ClassfileImpl();
+    }
+
+    static Classfile of(Option... options) {
+        return of().withOptions(options);
     }
 
+    Classfile withOptions(Option... options);
+
```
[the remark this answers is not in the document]
asotona: Fixed, thanks.

src/java.base/share/classes/jdk/internal/classfile/Classfile.java
```diff
@@ -28,175 +28,204 @@
...
     /**
      * An option that affects the writing of classfiles.
      */
```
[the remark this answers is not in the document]
asotona: Fixed, thanks.

src/java.base/share/classes/jdk/internal/classfile/ClassHierarchyResolver.java
```diff
@@ -48,13 +49,23 @@
...
-        return ClassHierarchyImpl.DEFAULT_RESOLVER;
+    static ClassHierarchyResolver ofSystem() {
+        var sysLoader = ClassLoader.getSystemClassLoader();
+        return ClassHierarchyResolver
+            .ofResourceParsing(sysLoader)
+            .orElse(ClassHierarchyResolver.ofClassLoading(sysLoader))
+            .cached(new Supplier<>() {
+                @Override
+                public Map<ClassDesc, ClassHierarchyResolver.ClassHierarchyInfo> get() {
+                    return new ConcurrentHashMap<>();
```
liach, line 66: Don't think we need to synchronize the cache if we have dedicated CHRs.

briangoetz:
Agreed.  The need for sync came from the shared static cache.

asotona:
Unfortunately thread-unsafe context makes sharing of it in tests executed in parallel a nightmare.
I can fix our Corpus tests and hope the race condition won't raise also somewhere else later.
However how to explain this limitation to users?

I suggest to make it always thread-safe (as the context is primary expected to be shared in multi-threaded environment) and users may make it faster in specific cases by providing non-synchronized map.

briangoetz:
Most entities in the system are immutable and therefore freely sharable 
without additional coordination.  Some have lazily-inflated single-field 
caches that conform to the "benign race" (there is only one possible 
non-default value) criteria.

The main hole is the CHA resolution cache.  There are two ways to 
address this: protect the cache, or don't share the context.

Unfortunately now that ClassModel refers back to the context, there's a 
risk of unexpected context sharing.  Here are a few options:

  - Make the CHA cache thread-safe using, say, CHM;
  - Make the CHA cache unshared, by using a ThreadLocal (this brings the 
context back to being immutable)
  - Detect when implicit use of a context happens on a thread other than 
the creating thread (currently the only vector for this is 
Classfile::transform), and inflate a new context with empty cache in 
that case;
  - Always create a fresh context with an empty cache in 
Classfile::transform

asotona:
>  Here are a few options: 
 - Make the CHA cache thread-safe using, say, CHM; 

This is current solution for the default system CH cache. 

We may also use a bit less strict custom semi-synchronization (just enough to be thread safe) instead of fully-synchronized `CHM::computeIfAbsent`, for example `CHM:get … (compute if absent without lock) ... CHM::put`.
CHM does not lock on get, so the only penalty would be synchronous put call. 

> - Make the CHA cache unshared, by using a ThreadLocal (this brings the context back to being immutable)

This is interesting option for small number of threads, however it makes big footprint for heavy-parallel systems.

> - Detect when implicit use of a context happens on a thread other than the creating thread (currently the only vector for this is Classfile::transform), and inflate a new context with empty cache in that case;

This would degrade the caching completely by triggering a new context for each transformed method in case there are two threads working synchronously.

> - Always create a fresh context with an empty cache in Classfile::transform

This would destroy the caching. Such cache would work only in scope of one method.

briangoetz:
On 6/8/2023 12:09 PM, Adam Sotona wrote:
>
>     Here are a few options:
>      - Make the CHA cache thread-safe using, say, CHM;
>
> This is current solution for the default system CH cache.
>
> We may also use a bit less strict custom semi-synchronization (just 
> enough to be thread safe) instead of fully-synchronized 
> |CHM::computeIfAbsent|, for example |CHM:get … (compute if absent 
> without lock) ... CHM::put|.
> CHM does not lock on get, so the only penalty would be synchronous put 
> call.
>
>      - Make the CHA cache unshared, by using a ThreadLocal (this
>     brings the context back to being immutable)
>
> This is interesting option for small number of threads, however it 
> makes big footprint for heavy-parallel systems.
>

I have a hard time imagining heavy parallel use here; concurrency 
against the same cache would most likely come from accidental sharing.  
So this is not necessarily a problem.  Are you imagining differnet use 
cases?

asotona:
> I have a hard time imagining heavy parallel use here; concurrency against the same cache would most likely come from accidental sharing.  So this is not necessarily a problem.  Are you imagining differnet use cases?

I think it is common practice for example in web services to pre-configure as much of the common parts as possible (and Classfile context logically falls into that category) and then register the service to the server, which use group of thread workers to handle individual requests.
With (non-reusable) virtual threads and structured concurrency on the table makes internal use of thread locals even less sense. I can imagine a new virtual thread launched (and finished) for each individual transformation. Our `CorpusTest` relying on Junit parallelization  is not so far from that scenario (I'm not sure if Junit already uses virtual threads or just a group of workers).

It is technically irrelevant if user will share the context directly or through a `ClassModel` (for example an instrumentor class used to transform all incoming code - this BTW an idea of faster `j.l.r.ProxyGenerator` I'm thinking about). I think we should be sure that `Classfile` context (and also expansion of `ClassModel`) does not fail in parallel environments.

asotona:
I'll run our standard benchmarks to compare following scenarios:
- static synchronized single instance cache (current master)
- context-based non-synchronized caches  `HM::computeIfAbsent`
- context-based fully synchronized caches `CHM::computeIfAbsent`
- context-based thread-safe lazy synchronized caches `CHM::get ... compute ... CHM::put`

And let's see the performance impact.

asotona:
I've performed a lot of benchmarking and the key result is that introduction of `Classfile` context cause a lot of regressions in our benchmarks.

Below are aggregated only relative results of benchmarks with significant regressions.
Current `master` with global static cache instance is a base for the relative regressions in the table.
- `ConcurrentHashMap::computeIfAbsent` represents the initial fully-synchronized cache implementation.
- `HashMap::computeIfAbsent` is the thread-unsafe option.
- `ConcurrentHashMap::get and put` is the lazy-synchronized solution
- **Always create a fresh context with an empty cache** is what Brian proposed as one of the options
- **Class loading with local cache only** is a step back to what ASM is using.


Interesting point is that **Always create a fresh context with an empty cache** works with similar performance as the others, except for `GenerateStackMaps.benchmark`, where it causes significant regression.

**Class loading with local cache only** assumes `ClassLoader` would do the caching for us. Even the class loading is initially slower and does not work for all use cases, it may be the best for the default CHR. And we could skip any `CHR::cached` in the API. 

Also interesting is that `ConcurrentHashMap::computeIfAbsent` cause 10% regression of **ParseOptions.transformNoStackmap** benchmark, where it should have no or minimal effect.

  | `ConcurrentHashMap::computeIfAbsent` | `HashMap::computeIfAbsent` | `ConcurrentHashMap::get and put` | Always create a fresh context with an empty cache | Class loading with local cache only
-- | -- | -- | -- | -- | --
AdHocAdapt.transform LIFT | -19% | -17% | -16% | -17% | -17%
AdHocAdapt.transform LIFT1 | -24% | -25% | -23% | -26% | -23%
AdHocAdapt.transform LIFT2 | -23% | -26% | -23% | -25% | -23%
AdaptInjectNoop.transform NOP_SHARED | -16% | -19% | -16% | -18% | -7%
AdaptNull.transform SHARED_3 | -18% | -21% | -19% | -17% | -18%
AdaptNull.transform SHARED_3_NO_DEBUG | -22% | -22% | -25% | -21% | -22%
GenerateStackMaps.benchmark |   |   |   | -27% |  
ParseOptions.transformNoDebug | -31% | -20% | -20% | -19% | -18%
ParseOptions.transformNoLineNumbers | -25% | -24% | -20% | -20% | -20%
ParseOptions.transformNoStackmap | -10% |   |   |   |

briangoetz:
FWIW: I have found AdaptNull.SHARED_3 to be among the most useful 
benchmarks for finding unexpected regressions.

I think there's something fishy going on.  Given that the baseline 
(static shared cache) has to protect against concurrent access already, 
and that there should be little or no other significant difference 
between the static version and the context version aside from caching, I 
can't see how those differences explain these regressions.  Are we 
making an apples-to-apples comparison of what CHR algorithm we're using? 
Is the creation of the CHR instances is perturbing the results?

Let's try to take CHR operation out of the equation: runt both baseline 
and context versions using a CHR that does almost nothing (e.g., says 
LUB(a,b) is always Object) or which uses a precomputed static cache.  
I'm guessing this shows no regression, and if so, this will likely tell 
us that the _context_ refactor is not the problem, but attempting to 
unshare the CHR cache.  Then we can drill into that problem.

src/java.base/share/classes/jdk/internal/classfile/impl/DirectCodeBuilder.java
```diff
@@ -318,52 +369,33 @@ public void writeBody(BufWriter b) {
...
-                if (!constantPool.options().generateStackmaps) {
-                    StackCounter cntr = StackCounter.of(DirectCodeBuilder.this, buf);
-                    maxStack = cntr.maxStack();
-                    maxLocals = cntr.maxLocals();
-                    stackMapAttr = null;
-                }
-                else if (canReuseStackmaps) {
-                    maxLocals = original.maxLocals();
-                    maxStack = original.maxStack();
-                    stackMapAttr = original.findAttribute(Attributes.STACK_MAP_TABLE).orElse(null);
```
liach, line 334: On the baseline, if a stack map table attribute can be reused but this findAttribute doesn't find an attribute (implying an implicit stack map, potentially for simple code?), we will not perform a stack map generation and simply reuse a trivial stack map (null), which should be a correct algorithm.

src/java.base/share/classes/jdk/internal/classfile/impl/DirectCodeBuilder.java
```diff
@@ -305,6 +308,54 @@ public void writeBody(BufWriter b) {
...
+                        } else {
+                            throw e;
+                        }
+                    }
+                } else {
+                    writeCounters(codeMatch, buf);
+                }
+            }
+
+            private boolean writeOriginalAttribute(BufWriterImpl buf) {
```
liach, line 349: As a result (see previous comment), we will run into a redundant stack map generation when the original stack map table is trivial (absent). Maybe we want to consider user-configured selective stack map regeneration to replace this faulty logiic.

liach (commented):
I have found the cause of the regression: there's redundant stack map generation in DirectCodeBuilder for methods with trivial stack map table (null).

liach:
Since "not generating stack maps" really means to use an implicit stack map, I suggest renaming the `STACK_MAPS_NEVER` to `STACK_MAPS_IMPLICIT` to indicate such a confusing feature of StackMapTable. (In fact, the enum element names of StackMapsOption doesn't fit in with other options, they should be ALWAYS_STACK_MAPS, NEVER_STACK_MAPS, etc.)

src/java.base/share/classes/jdk/internal/classfile/impl/DirectCodeBuilder.java
asotona: Yes, I caused this regression by "Revert of DirectCodeBuilder.needsStackMap pre-calculation" commit above.
The needStackMap pre-calculation on the other side caused regression in lambda performance, so one regression has been replaced with another.

src/java.base/share/classes/jdk/internal/classfile/impl/DirectCodeBuilder.java
asotona: I've fixed the logic in the commit below.

asotona:
All the performance regressions have been found and fixed.
Stackmaps generation logic is following:
- `STACK_MAPS_WHEN_REQUIRED` is the default option and performance-critical. It keeps original stack maps if code match and generate them only when required.
- `STACK_MAPS_ALWAYS` always (re)generate stack maps no matter if exist in the original
- `STACK_MAPS_NEVER` stack maps are not generated and dropped if already exist

Naming convention can be still adjusted.

Default `ClassHierarchyResolver` now use reflection from system class loader and it is not globally cached (so no dilemma about CHM vs HM). Any CHR is cached on method level if not already cached.

Benchmarks comparison of this PR vs master now shows even slight improvement:
```
Benchmark                                  (transform)    master  this PR Units
AdHocAdapt.transform                              LIFT     8.704    9.629 ops/s
AdHocAdapt.transform                             LIFT1     8.622    9.275 ops/s
AdHocAdapt.transform                             LIFT2     9.257    9.378 ops/s
AdaptInjectNoop.transform                   NOP_SHARED     4.709    4.978 ops/s
AdaptNull.transform                           SHARED_1    96.686   98.027 ops/s
AdaptNull.transform                           SHARED_2    48.820   49.257 ops/s
AdaptNull.transform                           SHARED_3     9.611    9.691 ops/s
AdaptNull.transform                  SHARED_3_NO_DEBUG    12.218   12.343 ops/s
AdaptNull.transform                         UNSHARED_3     3.614    3.603 ops/s
ParseOptions.transformNoDebug                      N/A    11.169   11.084 ops/s
ParseOptions.transformNoLineNumbers                N/A    10.590   10.523 ops/s
ParseOptions.transformNoStackmap                   N/A     6.220    9.646 ops/s
```

asotona:
>StackMapsOption doesn't fit in with other options, they should be ALWAYS_STACK_MAPS, NEVER_STACK_MAPS, etc.)

What about `STACK_MAPS_WHEN_REQUIRED`, `GENERATE_STACK_MAPS` and `DROP_STACK_MAPS`?

src/java.base/share/classes/jdk/internal/classfile/Classfile.java
asotona: As the default CHR stays static instance and it is no more cached, we can have also static holder for the default Classfile instance now.

asotona:
I think we've resolved controversy about global static CHR cache (by having no global static CHR cache ;) and turned all performance issues into minor improvements. A new `ClassfileBenchmark` has been added for more precise future continuous performance monitoring.

This PR is now waiting for review to unblock all follow-up conversion work.

Thank you,
Adam

src/java.base/share/classes/jdk/internal/classfile/impl/ClassfileImpl.java
```diff
@@ -0,0 +1,138 @@
...
+    public static final ClassfileImpl DEFAULT_CONTEXT = new ClassfileImpl(
+            StackMapsOption.STACK_MAPS_WHEN_REQUIRED,
+            DebugElementsOption.PASS_DEBUG,
+            LineNumbersOption.PASS_LINE_NUMBERS,
+            UnknownAttributesOption.PASS_UNKNOWN_ATTRIBUTES,
+            ConstantPoolSharingOption.SHARED_POOL,
+            ShortJumpsOption.FIX_SHORT_JUMPS,
+            DeadCodeOption.PATCH_DEAD_CODE,
+            DeadLabelsOption.FAIL_ON_DEAD_LABELS,
+            new ClassHierarchyResolverOptionImpl(ClassHierarchyResolver.defaultResolver()),
```
liach, line 63: Don't users of the default context still share a non-thread-safe-cached version of the default resolver? We might still need to convert DEFAULT_CONTEXT field to a static defaultContext() method that returns a `new ClassfileImpl(...);` (the attribute mapper option can be shared).

liach: Please disregard this comment: Now the distinction of cache of resolvers is done in `StackMapGenerator`, meaning each stack map generation task has its own cache. Didn't see it early on.

src/java.base/share/classes/jdk/internal/classfile/Classfile.java
```diff
@@ -284,332 +323,360 @@ public static void buildModuleTo(Path path,
...
+     * {@link ClassTransform}.  The transform will receive each element of
+     * this class, as well as a {@link ClassBuilder} for building the new class.
+     * The transform is free to preserve, remove, or replace elements as it
+     * sees fit.
+     *
+     * @implNote
+     * <p>This method behaves as if:
+     * {@snippet lang=java :
+     *     Classfile.of().build(thisClass(), ConstantPoolBuilder.of(this),
+     *                     b -> b.transform(this, transform));
```
liach, line 343:
```suggestion
     *     this.build(model.thisClass(), ConstantPoolBuilder.of(model),
     *                     b -> b.transform(model, transform));
```

src/java.base/share/classes/jdk/internal/classfile/Classfile.java
```diff
@@ -209,47 +251,44 @@ public static byte[] build(ClassDesc thisClass,
...
+    default void buildTo(Path path,
+                         ClassDesc thisClass,
+                         Consumer<ClassBuilder> handler) throws IOException {
+        Files.write(path, build(thisClass, handler));
     }
 
     /**
-     * Build a classfile into a file.
+     * Build a classfile into a file using the provided constant pool
+     * builder (which encapsulates classfile processing options.)
```
liach, line 272: The CP is now simply a data holder; it no longer holds the processing options. Same update is needed for the other few build methods taking a CP instance.
```suggestion
     * builder.
```

src/java.base/share/classes/jdk/internal/classfile/Classfile.java
```diff
@@ -28,175 +28,217 @@
...
  */
-public class Classfile {
-    private Classfile() {
-    }
+public sealed interface Classfile
+        permits ClassfileImpl {
 
     /**
-     * An option that affects the writing of classfiles.
+     * {@return a new context with default options}
```
liach, line 57:
```suggestion
     * {@return a context with default options}
```

liach:
@asotona I've sent a patch to your fork for a few API problems spotted a while ago and reported on the mailing list. Feel free to review my patch when you are back to work.

liach:
Since you fixed the string format typos, can you fix this typo as well? https://bugs.openjdk.org/browse/JDK-8308646 and remember to include that issue with `/issue add JDK-8308646` if you do.

liach:
I ran `make docs-classfile-api-javadoc` and the results look good in general. There are a few spec and API issues scattered across the whole API, which I think I will probably collect and list them in the mailing list later.

One thing of note is that the JDK itself has a `@sealedGraph` taglet to render the sealed class hierarchy of any class. We might want to add that tag to all our sealed public classes or interfaces. But it requires a somewhat more complicated build process (including generating dot files with graphviz and rendering them to svg) and you might tweak the `SetupExecute` to a full-on `SetupApiDocsGeneration`.

mcimadamore (approved):
The API changes look great. I like the instance usages of ClassFile, and how CF is now used in a centralized fashion to double down as a cache/option store (and entry point for transformation). Seems like a step in a good direction, and make the API feel a little more cohesive.

A small subjective comment - I'm not too fond of using "context" as a variable/accessor name for all the places where we use a classfile. E.g. if classfile is the context, perhaps these variable/accessor should just be named "classfile" ?

mcimadamore:
> The API changes look great. I like the instance usages of ClassFile, and how CF is now used in a centralized fashion to double down as a cache/option store (and entry point for transformation). Seems like a step in a good direction, and make the API feel a little more cohesive.
> 
> A small subjective comment - I'm not too fond of using "context" as a variable/accessor name for all the places where we use a classfile. E.g. if classfile is the context, perhaps these variable/accessor should just be named "classfile" ?

Also, another observation: the number of updates snippets in javadoc seems rather low for something that touches the main entry point by which client interacts with classfiles. This seems a sign of perhaps not having many end to end examples in the API javadoc?

asotona:
@mcimadamore thank you for review!
Description of `Classfile` instances as "context" is intentional compromise between discoverability (as originally proposed `Classfile.Context` didn't work very well) and clarity of the function (as `Classfile` instances may be simply confused with `ClassModel` instances).

asotona:
> I ran `make docs-classfile-api-javadoc` 

Thanks for pointing this out. The makefile patches accidentally fell into this PR.

> One thing of note is that the JDK itself has a `@sealedGraph` taglet to render the sealed class hierarchy of any class. 

I'll look at adding the tags, thanks.

liach:
Congrats, now I can update many of my patches that depend on this API update.

PR #14181, opened 2023-05-26
8308960: Decouple internal Version and OperatingSystem classes

RogerRiggs:
Decouple the jdk.internal.util OperatingSystem and Version classes to simplify class loading and avoid an indirect cyclic initialization.

Move the method to get the current OS version from OperatingSystem to the Version class and its initialization.
Revert to using String.toUpperCase() instead of custom code to uppercase.

Update the uses of OperatingSystem.version to be Version.current.

---------
### Progress

### Issue
 * [JDK-8308960](https://bugs.openjdk.org/browse/JDK-8308960): Decouple internal Version and OperatingSystem classes

### Reviewers
 * [Mandy Chung](https://openjdk.org/census#mchung) (@mlchung - **Reviewer**) ⚠️ Review applies to [f6332cd9](https://git.openjdk.org/jdk/pull/14181/files/f6332cd94ff5b7ce6e122e2c5bba0db7de46a88f)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14181/head:pull/14181` \
`$ git checkout pull/14181`

`$ git checkout pull/14181` \
`$ git pull https://git.openjdk.org/jdk.git pull/14181/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14181`

`$ git pr show -t 14181`

</details>

<a href="https://git.openjdk.org/jdk/pull/14181.diff">https://git.openjdk.org/jdk/pull/14181.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14181#issuecomment-1564531188)

### DIFF
--- a/src/java.base/macosx/classes/jdk/internal/loader/ClassLoaderHelper.java
+++ b/src/java.base/macosx/classes/jdk/internal/loader/ClassLoaderHelper.java
@@ -28,14 +28,13 @@ package jdk.internal.loader;
 import java.io.File;
 import java.util.ArrayList;
 
-import jdk.internal.util.OperatingSystem;
-import jdk.internal.util.Version;
+import jdk.internal.util.OSVersion;
 
 class ClassLoaderHelper {
 
     // SDK 10.15 and earlier always reports 10.16 instead of 11.x.x
-    private static final boolean hasDynamicLoaderCache = Version.current()
-            .compareTo(new Version(10, 16)) >= 0;
+    private static final boolean hasDynamicLoaderCache = OSVersion.current()
+            .compareTo(new OSVersion(10, 16)) >= 0;
 
     private ClassLoaderHelper() {}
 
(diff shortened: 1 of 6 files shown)


dfuch:
Should we take this opportunity to rename jdk.internal.util.Version into something like `OsVersion` to make it more clear what the `current()` method actually returns?

RogerRiggs:
> Should we take this opportunity to rename `jdk.internal.util.Version` into something like `OsVersion` to make it more clear what its `current()` method actually returns?

ok, it will give a bit more flavor; there are already multiple version classes with little to distinguish them.

mlchung (approved):
Looks good.  Thanks for fixing this.  It will allow ClassFileDumper to use `Path`.

mlchung:
> Should we take this opportunity to rename `jdk.internal.util.Version` into something like `OsVersion` to make it more clear what its `current()` method actually returns?

Alternatively, move `Version` record as a nested class in `OperatingSystem`.

dfuch:
> Alternatively, move `Version` record as a nested class in `OperatingSystem`.

That would work WRT naming, but I thought the point was to not have to load OperatingSystem in order to load Version?

mlchung:
I mean static member class of `OperatingSystem`.

RogerRiggs:
For the time being, I think they should be independent and clearly named.
The nesting creates a very unwieldy and long name.  (Not withstanding imports)

dfuch (commented):
Thanks for renaming the class. The renaming LGTM.

PR #14185, opened 2023-05-26
JDK-8307314: Implementation: Generational Shenandoah (Experimental)

kdnilsen:
OpenJDK Colleagues:

Please review this proposed integration of Generational mode for Shenandoah GC under https://bugs.openjdk.org/browse/JDK-8307314.

Generational mode of Shenandoah is enabled by adding `-XX:+UnlockExperimentalVMOptions -XX:ShenandoahGCMode=generational` to a command line that already specifies ` -XX:+UseShenandoahGC`.  The implementation automatically adjusts the sizes of old generation and young generation to efficiently utilize the entire heap capacity.  Generational mode of Shenandoah resembles G1 in the following regards:

1. Old-generation marking runs concurrently during the time that multiple young generation collections run to completion.
2. After old-generation marking completes, we perform a sequence of mixed collections.  Each mixed collection combines collection of young generation with evacuation of a portion of the old-generation regions identified for collection based on old-generation marking information.
3. Unlike G1, young-generation collections and evacuations are entirely concurrent, as with single-generation Shenandoah.
4. As with single-generation Shenandoah, there is no explicit notion of eden and survivor space within the young generation.  In practice, regions that were most recently allocated tend to have large amounts of garbage and these regions tend to be collected with very little effort.  Young-generation objects that survive garbage collection tend to accumulate in regions that hold survivor objects.  These regions tend to have smaller amounts of garbage, and are less likely to be collected.  If they survive a sufficient number of young-generation collections, the “survivor” regions are promoted into the old generation.

We expect to refine heuristics as we gain experience with more production workloads.  In the future, we plan to remove the “experimental” qualifier from generational mode, at which time we expect that generational mode will become the default mode for Shenandoah.

**Testing**: We continuously run jtreg tiers 1-4 + hotspot_gc_shenandoah, gcstress, jck compiler, jck runtime, Dacapo, SpecJBB, SpecVM, Extremem, HyperAlloc, and multiple AWS production workload simulators. We test on Linux x64 and aarch64, Alpine x64 and aarch64, macOS x64 and aarch64, and Windows x64.

---------
### Progress
- [x] Change requires CSR request [JDK-8307343](https://bugs.openjdk.org/browse/JDK-8307343) to be approved

### Issues
 * [JDK-8307314](https://bugs.openjdk.org/browse/JDK-8307314): Implementation: Generational Shenandoah (Experimental) (**Sub-task** - P4)
 * [JDK-8307343](https://bugs.openjdk.org/browse/JDK-8307343): Implementation: JEP 404: Generational Shenandoah (Experimental) (**CSR**)

### Reviewers
 * [Stefan Karlsson](https://openjdk.org/census#stefank) (@stefank - **Reviewer**) ⚠️ Review applies to [d4d2f1cf](https://git.openjdk.org/jdk/pull/14185/files/d4d2f1cf224200826593b67eabc2e0730f5a03f9)
 * [Y. Srinivas Ramakrishna](https://openjdk.org/census#ysr) (@ysramakrishna - **Reviewer**) ⚠️ Review applies to [d4d2f1cf](https://git.openjdk.org/jdk/pull/14185/files/d4d2f1cf224200826593b67eabc2e0730f5a03f9)
 * [Andrew Haley](https://openjdk.org/census#aph) (@theRealAph - **Reviewer**) ⚠️ Review applies to [8d80780a](https://git.openjdk.org/jdk/pull/14185/files/8d80780a64d7058bffa27b269157596d937a6b22)

### Contributors
 * Aleksey Shipilev `<shade@openjdk.org>`
 * Roman Kennke `<rkennke@openjdk.org>`
 * David Alvarez `<alvdavi@openjdk.org>`
 * Joshua Cao `<jcao@openjdk.org>`
 * Martin Doerr `<mdoerr@openjdk.org>`
 * Zhengyu Gu `<zgu@openjdk.org>`
 * William Kemper `<wkemper@openjdk.org>`
 * Bernd Mathiske `<bmathiske@openjdk.org>`
 * Kelvin Nilsen `<kdnilsen@openjdk.org>`
 * Y. Srinivas Ramakrishna `<ysr@openjdk.org>`
 * Wade Hennessey `<wadehenn@amazon.com>`
 * Celine Wu `<cfumiwu@amazon.com>`
 * Christian Young `<cyyoungm@amazon.com>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14185/head:pull/14185` \
`$ git checkout pull/14185`

`$ git checkout pull/14185` \
`$ git pull https://git.openjdk.org/jdk.git pull/14185/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14185`

`$ git pr show -t 14185`

</details>

<a href="https://git.openjdk.org/jdk/pull/14185.diff">https://git.openjdk.org/jdk/pull/14185.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14185#issuecomment-1568841389)


src/hotspot/share/gc/shared/gcConfiguration.cpp
```diff
@@ -75,7 +80,12 @@ GCName GCConfiguration::old_collector() const {
   }
 
   if (UseShenandoahGC) {
-    return Shenandoah;
+#if INCLUDE_SHENANDOAHGC
+    if (strcmp(ShenandoahGCMode, "generational") == 0) {
+      return Shenandoah;
+    }
+#endif
+    return NA;
```
stefank, line 88: You moved the order between Shenandoah and ZGC in `young_collector()`, so you should probably do the same here.

src/hotspot/cpu/x86/gc/shenandoah/shenandoahBarrierSetAssembler_x86.cpp
```diff
@@ -120,6 +122,29 @@ void ShenandoahBarrierSetAssembler::arraycopy_prologue(MacroAssembler* masm, Dec
...
+          // Save dst in r11 in the disjoint case
+          __ movq(r11, dst);
+        }
+      }
+#else
+      if (disjoint) {
+        __ mov(rdx, dst);          // save 'to'
+      }
+#endif
+    }
```
theRealAph, line 147: This logic is so gnarly that it's very hard to review and maintain, and IMO it's dangerous.

The problem is that its correctness depends on exactly how registers are allocated in its caller. This needs restructuring so that the register allocation is defined in a single place then passed down to everyone who needs it.

stefank (commented):
I've looked at the shared code and it's really nice that you've managed to keep them to a minimum. I have one tiny nit that would be nice to fix.

TheRealMDoerr:
Issues already reported to GenShen engineers:

gc/shenandoah/TestElasticTLAB.java#generational
#  Internal Error (src\hotspot\share\gc\shenandoah\shenandoahFreeSet.cpp:695), pid=23288, tid=23784
#  assert(size % CardTable::card_size_in_words() == 0) failed: size must be multiple of card table size, was 258

gc/stress/gcold/TestGCOldWithShenandoah.java#generational
#  Internal Error (src\hotspot\share\gc\shenandoah\heuristics\shenandoahOldHeuristics.cpp:82), pid=20828, tid=5836
#  assert(_old_generation->available() > old_evacuation_budget) failed: Cannot budget more than is available

gc/shenandoah/oom/TestAllocOutOfMemory.java#large
Execution failed: `main' threw exception: java.lang.RuntimeException: 'java.lang.OutOfMemoryError: Java heap space' missing from stdout/stderr
(Issue with 64k Pages)

gc/shenandoah/TestRetainObjects.java#no-tlab
gc/shenandoah/TestSieveObjects.java#no-tlab
Timeouts.

gc/shenandoah/TestAllocObjects.java#generational
gc/shenandoah/TestDynamicSoftMaxHeapSize.java#generational
#  Internal Error src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp:664), pid=18434, tid=29955
#  assert(is_global() || ShenandoahHeap::heap()->is_full_gc_in_progress() || (_used + _humongous_waste <= _affiliated_region_count * ShenandoahHeapRegion::region_size_bytes())) failed: used cannot exceed regions

test/hotspot/jtreg/gc/shenandoah/mxbeans/TestChurnNotifications.java
```diff
@@ -141,8 +165,8 @@ public void handleNotification(Notification n, Object o) {
                     Map<String, MemoryUsage> mapBefore = info.getGcInfo().getMemoryUsageBeforeGc();
                     Map<String, MemoryUsage> mapAfter = info.getGcInfo().getMemoryUsageAfterGc();
 
-                    MemoryUsage before = mapBefore.get("Shenandoah");
-                    MemoryUsage after = mapAfter.get("Shenandoah");
+                    MemoryUsage before = getUsage(mapBefore);
+                    MemoryUsage after = getUsage(mapAfter);
```
tstuefe, line 169: This also changes test logic for traditional Shenandoah, but its harmless. More correct would be to require "Young Gen" pool to only exist for -XX:ShenandoahGCMode=generational.

test/hotspot/jtreg/gc/shenandoah/oom/TestAllocOutOfMemory.java
```diff
@@ -0,0 +1,131 @@
...
+            allocate(size);
+            return;
+        }
+
+        // Called from jtreg, size is first argument
+        String size = args[0];
+        {
+            expectFailure("-Xmx16m",
+                          "-XX:+UnlockExperimentalVMOptions",
+                          "-XX:+UseShenandoahGC",
```
tstuefe, line 92: Nit: should not need UnlockExperimentalVMOptions anymore.

test/hotspot/jtreg/gc/shenandoah/oom/TestAllocOutOfMemory.java
```diff
@@ -0,0 +1,131 @@
...
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ *
+ */
```
tstuefe, line 23: Three tests folded into one, but it does not look like functionality changed for testing traditional Shenandoah. Okay.

tstuefe (commented):
I did a first read through the tests to check if any test changes affect traditional Shenandoah. To see if regression tests for non-generational are unchanged.

All good, I did not find anything noteworthy.

ysramakrishna:
> Issues already reported to GenShen engineers:
> 
> gc/shenandoah/TestElasticTLAB.java#generational #  Internal Error (src\hotspot\share\gc\shenandoah\shenandoahFreeSet.cpp:695), pid=23288, tid=23784 #  assert(size % CardTable::card_size_in_words() == 0) failed: size must be multiple of card table size, was 258
> 
> gc/stress/gcold/TestGCOldWithShenandoah.java#generational #  Internal Error (src\hotspot\share\gc\shenandoah\heuristics\shenandoahOldHeuristics.cpp:82), pid=20828, tid=5836 #  assert(_old_generation->available() > old_evacuation_budget) failed: Cannot budget more than is available
> 
> gc/shenandoah/oom/TestAllocOutOfMemory.java#large Execution failed: `main' threw exception: java.lang.RuntimeException: 'java.lang.OutOfMemoryError: Java heap space' missing from stdout/stderr (Issue with 64k Pages)
> 
> gc/shenandoah/TestRetainObjects.java#no-tlab gc/shenandoah/TestSieveObjects.java#no-tlab Timeouts.
> 
> gc/shenandoah/TestAllocObjects.java#generational gc/shenandoah/TestDynamicSoftMaxHeapSize.java#generational #  Internal Error src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp:664), pid=18434, tid=29955 #  assert(is_global() || ShenandoahHeap::heap()->is_full_gc_in_progress() || (_used + _humongous_waste <= _affiliated_region_count * ShenandoahHeapRegion::region_size_bytes())) failed: used cannot exceed regions

Thanks @TheRealMDoerr ; could you specify the platforms on which you see these failures, so we have a better chance at reproducing them? Thanks!

src/hotspot/share/gc/shared/gcConfiguration.cpp
kdnilsen: Thanks.  We've made your suggested change.

test/hotspot/jtreg/gc/shenandoah/oom/TestClassLoaderLeak.java
```diff
@@ -125,9 +126,10 @@ public static void main(String[] args) throws Exception {
         }
 
         String[][][] modeHeuristics = new String[][][] {
-             {{"satb"},    {"adaptive", "compact", "static", "aggressive"}},
-             {{"iu"},      {"adaptive", "aggressive"}},
-             {{"passive"}, {"passive"}}
+             {{"satb"},         {"adaptive", "compact", "static", "aggressive"}},
+             {{"iu"},           {"adaptive", "aggressive"}},
+             {{"passive"},      {"passive"}},
+             {{"generational"}, {"adaptive"}}
```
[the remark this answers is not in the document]
kdnilsen: Generational mode only works with adaptive heuristic.

stefank (approved):
I've reviewed the shared code and think that looks good.

feilongjiang:
Hi, I have built this pr based on aa85a90, Tier1 tests failed on `gc/TestAllocHumongousFragment.java#generational` with following output:
```
#
#
#  Internal Error (shenandoahVerifier.cpp:1244), pid=2951116, tid=2951124
#  Error: Verify init-mark remembered set violation; clean card should be dirty
#
# JRE version: OpenJDK Runtime Environment (21.0) (build 21-internal-adhoc.ubuntu.jdk)
# Java VM: OpenJDK 64-Bit Server VM (21-internal-adhoc.ubuntu.jdk, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, shenandoah gc, linux-riscv64)
```

Looks like Generational Shenandoah does not fully support RISC-V port, should we disable this test on RISC-V port for now?

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -1275,6 +1275,14 @@ HeapWord* ShenandoahHeap::allocate_new_gclab(size_t min_size,
...
                                             size_t word_size,
                                             size_t* actual_size) {
+  size_t unalignment = min_size % CardTable::card_size_in_words();
+  if (unalignment != 0) {
+    min_size = min_size - unalignment + CardTable::card_size_in_words();
+  }
+  unalignment = word_size % CardTable::card_size_in_words();
+  if (unalignment != 0) {
+    word_size = word_size - unalignment + CardTable::card_size_in_words();
+  }
```
ysramakrishna, line 1285: Probably not a big deal since this is only used when refilling a PLAB, which is an infrequent operation, but `mod` is an expensive operation, in general, and best to avoid in our code except in assertion checks (or even there given recent experiences with debug tests timing out). Since card size is a power of 2, may be we could use addition and masking instead. Something like defining the following inline in the CardTable class and using it everywhere where card alignment granularity is sought. There may even be a macro or method defined for this already perhaps:

```
(FOO + CardSize - 1) & ~((1 << LogCardSize) - 1)
```

One could even store the mask to avoid the arithmetic to produce the mask although it's pretty cheap.

This may turn out to be less expensive than mod, test, and branch, but as I said probably not a big deal here. We should make sure we don't overuse mods in our allocation paths much.

TheRealMDoerr:
> #  Internal Error (shenandoahVerifier.cpp:1244), pid=2951116, tid=2951124
> #  Error: Verify init-mark remembered set violation; clean card should be dirty

I've seen the same issue on PPC64: https://bugs.openjdk.org/browse/JDK-8309371

src/hotspot/cpu/riscv/gc/shenandoah/c1/shenandoahBarrierSetC1_riscv.cpp
```diff
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2018, 2019, Red Hat, Inc. All rights reserved.
  * Copyright (c) 2020, 2021, Huawei Technologies Co., Ltd. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 4: This should be backed out, since it seems that there is no (other) change to this fie.

TheRealMDoerr: Yes. And also from files which were changed by non-Amazon employees only, please.

ysramakrishna (approved):
Thanks for issue reports, for which we are filing JBS tickets as we receive reports, and will be resolving them as we go. The issues reported to us have been filed as issues linked to the JBS ticket for this PR. Reviewers and testers should please feel free to file tickets as they find other issues, given testing notes in the PR, and platforms on which we have inadequate or no testing or coverage. We appreciate the help!

We will be incrementally resolving as many of the reported issues as we are able to, including in follow up PRs as bug fixes as appropriate.

There were some issues reported with copyright headers, and those will be fixed soon in this PR.

Please follow the tickets for further updates.

Many thanks to all the reviewers so far.

I have been a reviewer at the project level for several of the commits that comprise this PR, so I am happy to review and approve this pull request as a reviewer, as well as a project participant and partial implementor for some fixes.

Onwards to the brave new generation of Shenandoah! ;-)

🛳       🚀

src/hotspot/cpu/riscv/gc/shenandoah/c1/shenandoahBarrierSetC1_riscv.cpp
ysramakrishna: Thanks, MArtin. Yes, we have noted that there were a few other files that were inadvertently caught in a opyright header dragnet. These will be reviewed and fixed in https://bugs.openjdk.org/browse/JDK-8309392 .

theRealAph:
This patch  was submitted 2023-05-26, two week before RDP1. It is large and complex.

My biggest concern it that it doesn't seem to be well-isolated from the existing Shenandoah code. The biggest risk is that it breaks Trad (i.e. non-generational) Shenandoah. A patch like this takes a few weeks to review properly. I don't believe we should hurry the review process.

While this looks interesting, and it has promise for the future, it's not worth breaking anything for.

chflood:
There are some applications where generational collectors are effective, but there are some applications where they aren't, like LRU caches.  I would like to see the changes made to Shenandoah to make it generational cleanly isolated from traditional Shenandoah so that both options remain available to our customers moving forward.

It seems unlikely that such changes can be completed before the deadline for RDP1.

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
```diff
@@ -657,11 +657,6 @@ void ShenandoahGeneration::establish_usage(size_t num_regions, size_t num_bytes,
 
 void ShenandoahGeneration::increase_used(size_t bytes) {
   Atomic::add(&_used, bytes);
-  // This detects arithmetic wraparound on _used.  Non-generational mode does not keep track of _affiliated_region_count
-  // TODO: REMOVE IS_GLOBAL() QUALIFIER AFTER WE FIX GLOBAL AFFILIATED REGION ACCOUNTING
-  assert(is_global() || ShenandoahHeap::heap()->is_full_gc_in_progress() ||
-         (_used + _humongous_waste <= _affiliated_region_count * ShenandoahHeapRegion::region_size_bytes()),
-         "used cannot exceed regions");
 }
```
kdnilsen, line 660: Note that C++ subexpression evaluation order is undefined.  Here is an example of what can go wrong with the removed assertion:

ThisThread:  fetches _affiliated_region_count * begins to multiply with region_size
OtherThread: increases _used and increases _affiliated_region_count appropriately (for a large allocation)
ThisThread: fetches _used and observes assert violation

kdnilsen: We have seen this race manifest in actual testing, which is what has motivated us to remove the assertions.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.inline.hpp
```diff
@@ -157,13 +157,6 @@ inline void ShenandoahHeapRegion::increase_live_data_gc_words(size_t s) {
 
 inline void ShenandoahHeapRegion::internal_increase_live_data(size_t s) {
   size_t new_live_data = Atomic::add(&_live_data, s, memory_order_relaxed);
-#ifdef ASSERT
```
kdnilsen, line 85: We have not observed violation of this assert during testing.  However, it appears unreliable in that _live_data increases monotonically under Atomic volatile math, whereas used() increases monotonically under heap lock, which we do not hold at the point of this assertion.

It is possible that another allocating thread increases both used() and _live_data.  We will see the increase in _live_data because _live_data changes are volatile Atomic.  However, we may not see the increase in used() because we did not acquire the heap lock and the value of _top() that contributes to calculation of used() is not volatile.  This situation can lead to an assertion failure.

src/hotspot/cpu/x86/gc/shenandoah/shenandoahBarrierSetAssembler_x86.cpp
```diff
@@ -631,7 +730,9 @@ void ShenandoahBarrierSetAssembler::store_at(MacroAssembler* masm, DecoratorSet
       BarrierSetAssembler::store_at(masm, decorators, type, Address(tmp1, 0), val, noreg, noreg, noreg);
     } else {
       iu_barrier(masm, val, tmp3);
+      // TODO: store_check missing in upstream
```
kdnilsen, line 733: Remove this comment with integration.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahPassiveHeuristics.cpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018, 2019, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
kdnilsen, line 3: Copyright overreach.   Revert.

tstuefe:
Hi @kdnilsen,

I see that following settings changed default values for all of Shenandoah:

- ShenandoahLearningSteps (was 10, now 5)
- ShenandoahImmediateThreshold (was 90, now 70)
- ShenandoahAdaptiveDecayFactor (was 0.5, now 0.1)
- ShenandoahFullGCThreshold (was 3, now 64)

Assuming that the behavior of legacy Shenandoah remains unchanged, I assume the switches are now handled differently to arrive at the same behavior.

I see that we now have ShenandoahOOMGCRetries. Does the changed default for ShenandoahFullGCThreshold and this new ShenandoahOOMGCRetries switch mean the degeneration behavior of legacy Shenandoah did change?

I think the general thrust of my questions is, you assured us that legacy Shenandoah will show the same behavior post-patch, but since the settings changed, I assume that the meaning of these settings did change. We will need to document these effects for users of legacy Shenandoah, in case they need to translate existing settings in their environment. A release note would be really helpful.

Cheers, Thomas

kdnilsen:
Thanks Thomas for the feedback:

These proposed changes represent improvements to both Generational and Non-generational modes of operation.  We can revert if that is desired, or we can specialize Generational versions of these parameters so that they can have different values in different modes, but here is a bit of background.  We've done considerable testing on a variety of synthetic workloads and some limited testing on production workloads.  As we move towards upstream integration, we expect this will help us gain exposure to more production workloads.  The following changes were based on results of this testing:

* Decrease ShenandoahLearningSteps to 5 (from 10): For some workloads, we observed that there were "way too many" learning cycles being triggered.  We also observed that the learning achieved during learning cycles was not as trustworthy as the learning achieved during actual operation, because these learning cycles typically trigger during initialization phases which are not representative of real-world operation and because they usually trigger so prematurely that there has not been enough time for allocated objects to die before we garbage collect.
* Change ShenandoahImmediateThreshold to 70 from 90: We discovered during experiments with settings on certain real production workloads that reducing the threshold for abbreviated cycles significantly improved throughput, reduced degenerated cycles, and reduced high percentile end-to-end latency on the relevant services.  These experiments were based on single-generation Shenandoah.  We saw no negative impact of making this change on our various workloads.
 * I'll let @earthling-amzn comment on the change to ShenandoahAdaptiveDecayFactor.  My recollection is that this change was also motivated by experience with single-generation Shenandoah on a real production workload.
 * The change of ShenandoahFullGCThreshold from 3 to 64 was motivated by some observations with specjbb performance as it ratchets up the workload to determine MaxJOPS.  We observed that for both single-generation Shenandoah and generational Shenandoah, the typical behavior was that a single Full GC trigger causes an "infinite" sequence of Full GC, even though we may have only lost the concurrent GC race by a small amount.  This is because (1) Full GC discards all the incremental work of the concurrent GC that was just interrupted, (2) STW Full GC creates a situation in which pent up demand for execution and allocation accumulates during the STW pause so there's a huge demand for allocation immediately following the end of Full GC, (3) The concurrent GC that triggers immediately after Full GC completes is "destined" to fail because no garbage has been introduced since Full GC finished and since SATB does not collect floating garbage that accumulates after the start of concurrent GC and since the allocation spike is so high immediately following the Full GC (e.g. 11GB/s instead of 3GB/s normally).  This change allows a sequence of degenerated GCs to manage slow evolution and sudden bursts of allocation rate much more effectively than the original code.  This is accompanied by a change in how we detect and throw OOM.  We wait for at least one Full GC but we don't force ShenandoahFullGCThreshold allocation failures before thowing OOM.

earthling-amzn:
Lowering `ShenandoahAdaptiveDecayFactor` allows the heuristic to give more weight to older samples of the allocation rate and cycle times. We found that with the original value (0.5), the heuristics would "forget" history too soon. With the original value, the heuristics were more likely to mistime their trigger because of a few recent, short cycles. This was particularly true after we lowered `ShenandoahImmediateThreshold`, which resulted in more cycles which could skip evacuation and updating refs.

chflood:
I wrote an LRU program back in 2017 which allocates trees and stores them in an array in a round robin fashion, freeing the last allocated.  At the time this was written it's purpose was to show how generational GCs can hit the wall and start performing very badly.  I ran this on a clean openjdk build, a genshen build in generational mode and a genshen  build in non-generational mode.   These results are repeatable for me. 

I would like to understand where the degradation is coming from before moving forward with this patch since it appears to penalize those who wish to just run traditional Shenandoah.

Clean
cflood@fedora java_programs]$ ~/genshen/cleanjdk/build/linux-x86_64-server-release/images/jdk/bin/java  -XX:+UnlockExperimentalVMOptions -XX:+UseShenandoahGC LRU 1000 1000
Took 341892ms to allocate 1000000 trees in a cache of 1000

Genshen generational (we expect this to be bad)
[cflood@fedora java_programs]$ ~/genshen/jdk/build/linux-x86_64-server-release/images/jdk/bin/java  -XX:+UnlockExperimentalVMOptions -XX:+UseShenandoahGC -XX:ShenandoahGCMode=generational LRU 1000 1000
Took 442012ms to allocate 1000000 trees in a cache of 1000

Genshen non-generational (shows what I feel is a significant degradation from the clean build)
[cflood@fedora java_programs]$ ~/genshen/jdk/build/linux-x86_64-server-release/images/jdk/bin/java  -XX:+UnlockExperimentalVMOptions -XX:+UseShenandoahGC LRU 1000 1000
Took 395679ms to allocate 1000000 trees in a cache of 1000

I think that generational Shenandoah can be a big win for some applications, but I want to fully understand the cost for all applications.

I can't attach a .java file so here it is inline in the post.

class TreeNode {
    public TreeNode left, right;
    public int val;
}


public class LRU {
    static int cache_size;
    static int reps;
    static int tree_height=16;

    private static TreeNode[] trees;

    private static int getIndex(int i) {return i % cache_size;}
    private static TreeNode makeTree(int h) {
        if (h == 0) { return null;}
        else {
	    TreeNode res = new TreeNode();
	    res.left = makeTree(h - 1);
	    res.right = makeTree(h - 1);
	    res.val = h;
	    return res;
	}
    }

        if (args.length != 2) {
            System.err.println("LRU requires args: cache_size reps");
            return;
        }
        cache_size = Integer.parseInt(args[0]);
        reps = Integer.parseInt(args[1]) * cache_size;
        trees = new TreeNode[cache_size];

        long start = System.currentTimeMillis();
        for (int i = 0; i < reps; i++)
            trees[getIndex(i)] = makeTree(tree_height);
        long end = System.currentTimeMillis();
        long ms = end - start;

        System.out.println("Took " + ms + "ms to allocate " + reps + " trees in a cache of " + cache_size);
    }
}

phohensee:
Thanks for finding the single-gen regression, we're very happy you took the time to run it and write up your results. We're very concerned about single-gen regressions too because we have single-gen Shen in production for several critical services.

We'd like to propose to push now, and tackle/fix the single-gen issue you identified during RDP1, as well as any other significant single-gen regressions that may come up. We have four Shen experts on board, Roman, Aleksey, Kelvin, and William, so believe it's doable before RDP2 in July. In the worst case that we fail, we'd emulate ZGC and move GenShen to it's own directory as an entirely separate collector before RDP2. Make sense?

theRealAph:
> We'd like to propose to push now, and tackle/fix the single-gen issue you identified during RDP1, as well as any other significant single-gen regressions that may come up. We have four Shen experts on board, Roman, Aleksey, Kelvin, and William, so believe it's doable before RDP2 in July. In the worst case that we fail, we'd emulate ZGC and move GenShen to it's own directory as an entirely separate collector before RDP2. Make sense?

That sounds great to me. I'll approve this PR now, but please wait for Christine's ack.

theRealAph (approved):

chflood:
This sounds good to me.

mbreinhold:
> We'd like to propose to push now, and tackle/fix the single-gen issue you identified during RDP1, as well as any other significant single-gen regressions that may come up. We have four Shen experts on board, Roman, Aleksey, Kelvin, and William, so believe it's doable before RDP2 in July. In the worst case that we fail, we'd emulate ZGC and move GenShen to it's own directory as an entirely separate collector before RDP2. Make sense?

If you’re planning for this amount of change during RDP 1 then I’d say that you’re not ready for RDP 1.

If this patch were less isolated from the rest of HotSpot then I’d be extremely nervous.

phohensee:
We understand, and would not have proposed the last chance split-directory alternative without the level of isolation Hotspot's GC interface enables. We've added single-gen performance enhancements on the way and would like to keep them!

test/hotspot/jtreg/gc/shenandoah/TestEvilSyncBug.java
```diff
@@ -23,13 +24,23 @@
...
 /*
- * @test
+ * @test id=default
  * @summary Tests for crash/assert when attaching init thread during shutdown
  * @requires vm.gc.Shenandoah
  * @library /test/lib
  * @modules java.base/jdk.internal.misc
  *          java.management
- * @run driver/timeout=480 TestEvilSyncBug
+ * @run driver/timeout=480 TestEvilSyncBug -XX:ShenandoahGCHeuristics=aggressive
```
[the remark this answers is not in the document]
kdnilsen: We moved this argument from the source code (original line 64) to here so that we can invoke the test differently in generational mode.

test/hotspot/jtreg/gc/shenandoah/oom/TestAllocOutOfMemory.java
kdnilsen: We actually do need to UnlockExperimentalVMOptions because generational mode is currently an experimental feature.

test/hotspot/jtreg/gc/shenandoah/TestEvilSyncBug.java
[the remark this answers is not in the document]
kdnilsen: See line 64 of the original source code.  We did not change the behavior.  We only changed how the behavior is realized.  This enables generalization of the test to generational mode.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
kdnilsen: Thanks for this suggestion.  I've modified the code.

src/hotspot/cpu/riscv/gc/shenandoah/c1/shenandoahBarrierSetC1_riscv.cpp
kdnilsen: I'm fixing this copyright notice and others.

ysramakrishna:
> Hi, I have built this pr based on [aa85a90](https://github.com/openjdk/jdk/commit/aa85a9073e2a71d6bf920409e739d555f9dcf302), Tier1 tests failed on `gc/TestAllocHumongousFragment.java#generational` on Linux/RISC-V with the following output:
> 
> ```
> #
> # A fatal error has been detected by the Java Runtime Environment:
> #
> #  Internal Error (shenandoahVerifier.cpp:1244), pid=2951116, tid=2951124
> #  Error: Verify init-mark remembered set violation; clean card should be dirty
> #
> # JRE version: OpenJDK Runtime Environment (21.0) (build 21-internal-adhoc.ubuntu.jdk)
> # Java VM: OpenJDK 64-Bit Server VM (21-internal-adhoc.ubuntu.jdk, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, shenandoah gc, linux-riscv64)
> ```
> 
> Looks like Generational Shenandoah does not fully support RISC-V port, should we disable this test on RISC-V port for now?

Fixed (platform disabled) by @kdnilsen in https://github.com/openjdk/jdk/pull/14185/commits/cc149904d76c78355fc994da171f0f21411e903f

src/hotspot/cpu/ppc/gc/shenandoah/shenandoahBarrierSetAssembler_ppc.cpp
```diff
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 2018, 2021, Red Hat, Inc. All rights reserved.
- * Copyright (c) 2012, 2021 SAP SE. All rights reserved.
+ * Copyright (c) 2012, 2022 SAP SE. All rights reserved.
  * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 4: I believe line 4 should deleted; the copyright header change here is unnecessary.

src/hotspot/cpu/riscv/gc/shenandoah/shenandoahBarrierSetAssembler_riscv.cpp
```diff
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2018, 2020, Red Hat, Inc. All rights reserved.
  * Copyright (c) 2020, 2021, Huawei Technologies Co., Ltd. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 4: Remove this line; extent of changes doesn't justify copyright header change.

src/hotspot/share/gc/shenandoah/c1/shenandoahBarrierSetC1.hpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018, 2021, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 3: Should probably be removed.

src/hotspot/share/gc/shenandoah/c2/shenandoahBarrierSetC2.hpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018, 2021, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 3: Check if this is necessary.

src/hotspot/share/gc/shenandoah/c2/shenandoahSupport.cpp
```diff
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2015, 2021, Red Hat, Inc. All rights reserved.
  * Copyright (C) 2022 THL A29 Limited, a Tencent company. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 4: Should be removed.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahCompactHeuristics.hpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018, 2019, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 3: Can be removed?

src/hotspot/share/gc/shenandoah/heuristics/shenandoahPassiveHeuristics.hpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018, 2019, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 3: Should be removed?

src/hotspot/share/gc/shenandoah/heuristics/shenandoahStaticHeuristics.hpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018, 2019, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 3: Should be removed?

src/hotspot/share/gc/shenandoah/mode/shenandoahPassiveMode.hpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2019, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 3: Should be removed?

src/hotspot/share/gc/shenandoah/shenandoahBarrierSetClone.inline.hpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2013, 2021, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 3: Unnecessary. Delete.

src/hotspot/share/gc/shenandoah/shenandoahConcurrentMark.hpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2013, 2021, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 3: Probably unnecessary.

tstuefe:
> Thanks Thomas for the feedback:
> 
> These proposed changes represent improvements to both Generational and Non-generational modes of operation. We can revert if that is desired, or we can specialize Generational versions of these parameters so that they can have different values in different modes, but here is a bit of background. We've done considerable testing on a variety of synthetic workloads and some limited testing on production workloads. As we move towards upstream integration, we expect this will help us gain exposure to more production workloads. The following changes were based on results of this testing:

> <snip>

Hi Kelvin,

thanks for the thorough explanations!

It is a pity that these valuable insights are buried in a GH discussion and these changes inside such a large patch. I also looked at the originating patch in openjdk/shenandoah, which I assume is your development repo for Shenandoah (?).

Could I convince you to adapt the JBS issue process in the shenandoah repo (so, opening an issue on JBS, with some clear explanation, then fixing the bug)? Roman convinced me of this for the Lilliput repository, and now I think the added work is well worth it. JBS is a treasure trove of insights, if filled with care, and can help us for many years.

Some more questions about `ShenandoahFullGCThreshold`:

I am looking at the nice ASCII art in `ShenandoahControlThread::service_concurrent_normal_cycle`. IIUC, the cycle goes:

```
Concurrent GC -> Alloc failure -> n x Degenerated GC -> Alloc Failure -> Full GC
```

right? So the change is now in how often we try a degenerated GC before falling back to a full GC?

With GenShen, does a degenerated GC still collect only the young regions? And only FullGC does collect all regions?

Are comment and ASCII-art still correct for GenShen? E.g. the comment says:

```
  // If second allocation failure happens during Degenerated GC cycle (for example, when GC
  // tries to evac something and no memory is available), cycle degrades to Full GC.
```

Is "second allocation failure" correct? Since even before this patch, we tried three times before falling back to a Full GC.

Thank you,

Thomas

src/hotspot/share/gc/shenandoah/shenandoahFullGC.hpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2014, 2021, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 3: Unnecessary

src/hotspot/share/gc/shenandoah/shenandoahGC.cpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2021, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 3: Unnecessary

src/hotspot/share/gc/shenandoah/shenandoahGC.hpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2021, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 3: Unnecessary

src/hotspot/share/gc/shenandoah/shenandoahNMethod.cpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2019, 2022, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 3: Unnecessary?

src/hotspot/share/gc/shenandoah/shenandoahNumberSeq.hpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018, 2019, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 3: Unnecessary?

src/hotspot/share/gc/shenandoah/shenandoahSTWMark.hpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2021, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 3: unnecessary.

src/hotspot/share/gc/shenandoah/shenandoahUnload.cpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2019, 2021, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 3: Delete

src/hotspot/share/gc/shenandoah/shenandoahWorkerPolicy.hpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2017, 2022, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 3: Unnecessary

test/hotspot/jtreg/gc/stress/gcold/TestGCOldWithShenandoah.java
```diff
@@ -1,5 +1,6 @@
 /*
 * Copyright (c) 2016, 2020, Oracle and/or its affiliates. All rights reserved.
+* Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
ysramakrishna, line 3: General comment: Looking at the history, I might have expected RedHat copyright headers for many of these tests, but that isn't a change that's happened with generational shenandoah. So, nothing for us to do in this PR.

src/hotspot/cpu/ppc/gc/shenandoah/shenandoahBarrierSetAssembler_ppc.cpp
kdnilsen: will remove.  I noticed that amazon had also contributed to this file, but changes were very minor.

src/hotspot/share/gc/shenandoah/shenandoahFullGC.hpp
kdnilsen: fixed.

src/hotspot/share/gc/shenandoah/shenandoahGC.cpp
kdnilsen: fixed.

src/hotspot/share/gc/shenandoah/shenandoahNumberSeq.hpp
kdnilsen: fixed.

src/hotspot/share/gc/shenandoah/shenandoahUnload.cpp
kdnilsen: fixed.

src/hotspot/share/gc/shenandoah/shenandoahSTWMark.hpp
kdnilsen: fixed.

src/hotspot/share/gc/shenandoah/shenandoahWorkerPolicy.cpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2017, 2019, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
kdnilsen: fixed.

src/hotspot/share/gc/shenandoah/shenandoahWorkerPolicy.hpp
kdnilsen: fixed.

src/hotspot/share/gc/shenandoah/shenandoahDegeneratedGC.hpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2021, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
kdnilsen: Thanks Ramki for sifting through these again.  Sorry I missed so many.  I'm making your suggested fixes.

kdnilsen:
Hi Thomas,

Thank you for your followup comments.  I am in total agreement that it is a shame the challenges we have faced and the progress we have made is not better documented in the history of JBS tickets.  I have been the worst offender.  I apologize.

One aspect of this problem is that our work has included a large degree of uncertainty and "research", and it is not always clear to us what needs to be addressed until after we finish and test certain fixes as integrated with a variety of other fixes.  We will commit to being more engaged with JBS from this point forward, both for any further work done on the Shenandoah branch, and definitely for work done on tip.

You are correct that the change is to N, the number of times in a row that we perform degenerated GC before we automatically upgrade to Full GC.  It is still possible that we will upgrade to Full GC before N is reached, because there are other situations, such as lack of progress by degenerated GC, that will cause us to upgrade to Full even before N is reached.

The comment is still valid as written.  During degenerated GC, the mutator threads are all blocked, so the ONLY kind of allocation failure that can occur during degenerated GC is a GC-worker-thread allocation for the purpose of evacuating memory.  If we experience an "evacuation failure" during degenerated GC. we will upgrade to Full GC.

src/hotspot/share/gc/shenandoah/c1/shenandoahBarrierSetC1.hpp
kdnilsen: removed.

src/hotspot/share/gc/shenandoah/c2/shenandoahBarrierSetC2.hpp
kdnilsen: ok.  i'll remove.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahPassiveHeuristics.hpp
kdnilsen: removed.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahStaticHeuristics.hpp
kdnilsen: removed.

src/hotspot/share/gc/shenandoah/mode/shenandoahPassiveMode.hpp
kdnilsen: removed.

src/hotspot/share/gc/shenandoah/mode/shenandoahSATBMode.cpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2019, 2021, Red Hat, Inc. All rights reserved.
+ * Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
kdnilsen: removed.

src/hotspot/share/gc/shenandoah/shenandoahBarrierSetClone.inline.hpp
kdnilsen: removed.

tstuefe:
> Hi Thomas,
> 
> Thank you for your followup comments. I am in total agreement that it is a shame the challenges we have faced and the progress we have made is not better documented in the history of JBS tickets. I have been the worst offender. I apologize.

Please, no need to apologize. I understand that during early development one needs to move quickly. I just thought that your team's experience with tuning Shenandoah is valuable, and it is regrettable when it is lost.

> You are correct that the change is to N, the number of times in a row that we perform degenerated GC before we automatically upgrade to Full GC. It is still possible that we will upgrade to Full GC before N is reached, because there are other situations, such as lack of progress by degenerated GC, that will cause us to upgrade to Full even before N is reached.
> 
> The comment is still valid as written. During degenerated GC, the mutator threads are all blocked, so the ONLY kind of allocation failure that can occur during degenerated GC is a GC-worker-thread allocation for the purpose of evacuating memory. If we experience an "evacuation failure" during degenerated GC. we will upgrade to Full GC.

Thank you for the thorough explanation.

tstuefe:
I won't be able to give reasonable input here in the short time left before RDP1. Nor am I the most qualified to do so. 

Just wanted to re-iterate that I see this rushed review with worry. Nothing I have not said already, and of course, it does not diminish the massive effort behind this JEP.

Process-wise, when talking about Lilliput integration we touched on the idea of moving RDP1 down to a sooner date. I still think this makes sense.

kdnilsen:
We would like to thank everyone who has taken time to review and provide feedback on our pull request.  Given the risks identified during the review process and the lack of time available to perform the thorough review that such a large contribution of code requires, we have decided to close this PR at the current time.  We will seek to target JDK 22.

theRealAph:
> We would like to thank everyone who has taken time to review and provide feedback on our pull request. Given the risks identified during the review process and the lack of time available to perform the thorough review that such a large contribution of code requires, we have decided to close this PR at the current time. We will seek to target JDK 22.

Thank you for this. It's the right decision. In hindsight, there never was a highly-likely prospect of getting such a substantial and interwoven patch successfully reviewed in such a short time, even with the most skilful and experienced team.

PR #14187, opened 2023-05-26
8306688: Support Windows serialized keystores (SST files)

macarte:
Added ability to load keystores from SST files on Windows.  Example usage:

KeyStore keyStore = KeyStore.getInstance("Windows-SST");
try (FileInputStream fis = new FileInputStream("mykeystore.sst")) {
   keyStore.load(fis, null);
}

Note that its not limited to file streams, it can be any stream.

The feature is behind a runtime flag ("sun.security.mscapi.keyStoreSSTSupport") as the KeyStore must have an input stream, but the JCK tests assume an input stream is optional

tier1 tests for linux/macos/Windows for x86_64

---------
### Progress
- [ ] Change requires a CSR request matching fixVersion 22 to be approved (needs to be created)

### Issue
 * [JDK-8306688](https://bugs.openjdk.org/browse/JDK-8306688): Support Windows serialized keystores (SST files) (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14187/head:pull/14187` \
`$ git checkout pull/14187`

`$ git checkout pull/14187` \
`$ git pull https://git.openjdk.org/jdk.git pull/14187/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14187`

`$ git pr show -t 14187`

</details>

<a href="https://git.openjdk.org/jdk/pull/14187.diff">https://git.openjdk.org/jdk/pull/14187.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14187#issuecomment-1564973973)

### DIFF
--- a/src/jdk.crypto.mscapi/windows/classes/sun/security/mscapi/CKeyStore.java
+++ b/src/jdk.crypto.mscapi/windows/classes/sun/security/mscapi/CKeyStore.java
@@ -56,6 +56,13 @@ abstract class CKeyStore extends KeyStoreSpi {
 
     private static final int LOCATION_CURRENTUSER = 0;
     private static final int LOCATION_LOCALMACHINE = 1;
+    private static final int LOCATION_SST = 2;
+
+    public static final class SST extends CKeyStore {
+        public SST() {
+            super("SST", LOCATION_SST);
+        }
+    }
 
     public static final class MY extends CKeyStore {
         public MY() {
@@ -702,8 +709,12 @@ abstract class CKeyStore extends KeyStoreSpi {
      */
     public void engineLoad(InputStream stream, char[] password)
             throws IOException, NoSuchAlgorithmException, CertificateException {
-        if (stream != null && !keyStoreCompatibilityMode) {
+        if (stream != null && !getRequiresKeyStoreFromStream() && !keyStoreCompatibilityMode) {
             throw new IOException("Keystore input stream must be null");
+        }
+        else if (stream == null && getRequiresKeyStoreFromStream())
+        {
+            throw new IOException("Keystore input stream cannot be null");
         }
 
         if (password != null && !keyStoreCompatibilityMode) {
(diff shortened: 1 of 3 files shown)


seanjmullan:
This Enhancement requires a CSR as you are introducing a new KeyStore type and system property. RDP 1 for JDK 21 is on June 8, so there is not much time left to review and approve the CSR and code for this issue. I recommend you retarget this to JDK 22.

macarte:
@wangweij - the set and delete will modify the in memory store instance.  However as of now, the store is a nop - historically as setting/removing from the MY/ROOT stores persists the changes (but not the case for in-memory stores using SST).

So I see three ways forward, either (1) leave as is, (2) open the memory store with CERT_STORE_READONLY_FLAG [exceptions will be thrown when trying to modify the store] or (3) support persisting the store to an output stream. 

(1) has value as it allows you to create in memory stores on Windows using windows APIs (vs PKS etc) 

(2) seems limited - but supports my customer needs

(3) expands beyond my current needs, but might offer the most flexibility

In cases (1) and (3): perhaps we should allow in-memory stores that don't need to be loaded from SST, ie create them empty.

macarte:
The SunMSCAPI CKeyStore which implements KeyStoreSpi does nothing in the engineStore, ie. its essential a NOP (No Operation), so right now the dev cannot overwrite original SST file

macarte:
Thanks for the explanation, I understand your point now. I'll make changes to support writing to the SST

ecki (commented):
Can a SST file actually contain keys? According to the Powershell page you linked it only exports certs.

macarte:
Removed

From: Weijun Wang ***@***.***>
Date: Monday, December 4, 2023 at 7:01 AM
To: openjdk/jdk ***@***.***>
Cc: Mat Carter ***@***.***>, Mention ***@***.***>
Subject: Re: [openjdk/jdk] 8306688: Support Windows serialized keystores (SST files) (PR #14187)

@macarte<https://github.com/macarte> Please remove the Fix Version/s field of this bug now that it will not go into jdk22.

PR #14191, opened 2023-05-27
8303451: Synchronization entry in C2 debug info is misleading

quadhier:
This should fix [JDK-8303451](https://bugs.openjdk.org/browse/JDK-8303451).

It is a trivial patch that fixes a misleading code comment at method entry printed by `-XX:+PrintAssembly`.

For exmple,
```asm
0x0000ffffa409da88:   stp x29, x30, [sp, #16]         ;*synchronization entry
```
will become
```asm
0x0000ffffa409da88:   stp x29, x30, [sp, #16]         ; * invocation entry (also synchronization entry if synchronized)
```


quadhier:
Sorry, some thing wrong with my commits. I'd close this and open a new one.

PR #14195, opened 2023-05-28
8308643: Incorrect value of 'used' jvmstat counter

kimbarrett:
Please review this change that fixes the space-used performance counters
provided by ParallelGC and SerialGC.

The prior fix for a problem in this area (JDK-8268265) wasn't correct.  It
used a static "last value" variable, for use when sampling is blocked by an
in-progress GC.  But there are multiple counters, so having one static "last
value" variable doesn't work.  (Not sure what I and my reviewers were thinking
at the time. Maybe we overlooked that there are multiple counters.)  The
solution is to associate a "last value" with each of the counters.

Also added a test of basic functionality (just accessibility) of the space
counters.  No regression test of the problem being fixed here, as its hard to
reliably set up.

Note: SpaceCounters and CSpaceCounters are very nearly identical.  It's likely
possible to refactor for more code sharing, which could be done as a followup.

Testing:
mach5 tier1

---------
### Progress

### Issue
 * [JDK-8308643](https://bugs.openjdk.org/browse/JDK-8308643): Incorrect value of 'used' jvmstat counter (**Bug** - P3)

### Reviewers
 * @giorgigagnidze16 (no known openjdk.org user name / role)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14195/head:pull/14195` \
`$ git checkout pull/14195`

`$ git checkout pull/14195` \
`$ git pull https://git.openjdk.org/jdk.git pull/14195/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14195`

`$ git pr show -t 14195`

</details>

<a href="https://git.openjdk.org/jdk/pull/14195.diff">https://git.openjdk.org/jdk/pull/14195.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14195#issuecomment-1565817466)


giorgigagnidze16 (approved):

tschatzl (approved):

src/hotspot/share/gc/parallel/spaceCounters.hpp
```diff
@@ -39,6 +39,7 @@ class SpaceCounters: public CHeapObj<mtGC> {
  private:
   PerfVariable*      _capacity;
   PerfVariable*      _used;
+  volatile size_t    _last_used_in_bytes;
```
albertnetymk, line 42: Why is an additional field required here? Can't `_used` fulfill the purpose?

kimbarrett: The _used counters only seem to be updated by GCs.  The take_sample function reports "current" data.

kimbarrett: Using the perf-counter to carry the periodic samples doesn't work. During
construction of the perf-counter, the sampler is called to provide the initial
value for the counter. Since the counter is not yet fully constructed and
accessible to the sampling function (_used is not yet initialized), this crashes.

A workaround for that problem would be to add constructors that take both a
sampler and an initial value. That's kind of messy, and doesn't seem better
than the proposed solution at hand.

albertnetymk: Thank you for the clarification.

albertnetymk (approved):

kimbarrett:
Thanks for reviews.

PR #14196, opened 2023-05-28
8308892: Bad graph detected in build_loop_late after JDK-8305635

chhagedorn:
The cleanup done in [JDK-8305635](https://bugs.openjdk.org/browse/JDK-8305635) wrongly identifies unrelated Parse Predicates which are not cleaned up, yet. It just walks from the entry of the loop up and tries to find each of the three Parse Predicates once but in no particular order. This order insensitive walk is wrong as seen in the following graph (from the attached replay file of this bug):

image

We first find `116 Parse Predicate` for Loop Predicates, then `84 Parse Predicate` for Profiled Loop Predicates and then stop when finding `71 Parse Predicate` for Loop Predicates because we've already found a Parse Predicate for Loop Predicates already. We then wrongly create Loop Predicates (above `116 Parse Predicate`) which are below newly created Profiled Loop Predicates (above `84 Parse Predicate`). This could lead to a bad graph because of data dependencies that rely on the fact that Loop Predicates are above Profiled Loop Predicates:
https://github.com/openjdk/jdk/blob/547a8b40b324917e66c71409b31421feacce79d7/src/hotspot/share/opto/loopPredicate.cpp#L1529-L1543

The fix is straight forward to make the assignment of Parse Predicate projections in `ParsePredicates` aware of the relative ordering constraint. Note that this class will be refactored again in [JDK-8305636](https://bugs.openjdk.org/browse/JDK-8305636). But I think properly fixing this first is better than waiting for JDK-8305636 to go in.

Testing: tier1-4, hs-precheckin-comp, hs-stress-comp

Thanks,
Christian

---------
### Progress

### Issue
 * [JDK-8308892](https://bugs.openjdk.org/browse/JDK-8308892): Bad graph detected in build_loop_late after JDK-8305635

### Reviewers
 * [Roberto Castañeda Lozano](https://openjdk.org/census#rcastanedalo) (@robcasloz - **Reviewer**) ⚠️ Review applies to [806bdba5](https://git.openjdk.org/jdk/pull/14196/files/806bdba5c47622c1e060f041e0ea3983161770ed)
 * [Roland Westrelin](https://openjdk.org/census#roland) (@rwestrel - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14196/head:pull/14196` \
`$ git checkout pull/14196`

`$ git checkout pull/14196` \
`$ git pull https://git.openjdk.org/jdk.git pull/14196/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14196`

`$ git pr show -t 14196`

</details>

<a href="https://git.openjdk.org/jdk/pull/14196.diff">https://git.openjdk.org/jdk/pull/14196.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14196#issuecomment-1566282531)

### DIFF
--- a/src/hotspot/share/opto/loopPredicate.cpp
+++ b/src/hotspot/share/opto/loopPredicate.cpp
@@ -1799,6 +1799,9 @@ ParsePredicateNode* ParsePredicates::get_parse_predicate_or_null(Node* parse_pre
 }
 
 // Initialize the Parse Predicate projection field that matches the kind of the parent of `parse_predicate_proj`.
+// Only initialize if Parse Predicate projection itself or any of the Parse Predicate projections coming further up
+// in the graph are not already initialized (this would be a sign of repeated Parse Predicates which are not cleaned up,
+// yet).
 bool ParsePredicates::assign_predicate_proj(ParsePredicateSuccessProj* parse_predicate_proj) {
   ParsePredicateNode* parse_predicate = get_parse_predicate_or_null(parse_predicate_proj);
   assert(parse_predicate != nullptr, "must exist");
@@ -1811,13 +1814,16 @@ bool ParsePredicates::assign_predicate_proj(ParsePredicateSuccessProj* parse_pre
       _loop_predicate_proj = parse_predicate_proj;
       break;
     case Deoptimization::DeoptReason::Reason_profile_predicate:
-      if (_profiled_loop_predicate_proj != nullptr) {
+      if (_profiled_loop_predicate_proj != nullptr ||
+          _loop_predicate_proj != nullptr) {
         return false;
       }
       _profiled_loop_predicate_proj = parse_predicate_proj;
       break;
     case Deoptimization::DeoptReason::Reason_loop_limit_check:
-      if (_loop_limit_check_predicate_proj != nullptr) {
+      if (_loop_limit_check_predicate_proj != nullptr ||
+          _loop_predicate_proj != nullptr ||
+          _profiled_loop_predicate_proj != nullptr) {
         return false;
       }
       _loop_limit_check_predicate_proj = parse_predicate_proj;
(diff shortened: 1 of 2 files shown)


robcasloz (approved):
Looks good!

Windows build failure in GHA testing is unrelated.

chhagedorn:
Thanks Roberto for your review!

rwestrel:
> The cleanup done in [JDK-8305635](https://bugs.openjdk.org/browse/JDK-8305635) wrongly identifies unrelated Parse Predicates which are not cleaned up, yet.

So which of the 3 predicates will get cleaned up? Are some of the 3 predicates for another loop that doesn't exist anymore?

rwestrel:
> The useless predicates were added normally for the loop while the additional `116 Parse Predicate` was added with `maybe_add_predicate_after_if()` (the profiled loop predicate was skipped due to `too_may_trap(reason)` being true).

Do I read that correctly that parsing inserts useless/duplicate predicates?

chhagedorn:
After the first IGVN round, after parsing, they become useless because the `If` that triggered the insertion of the predicates with `maybe_add_predicate_after_if()` is folded away. I think it's unfortunate, though, that we eliminate `71 Parse Predicate` and `84 Parse Predicate` instead of `116 Parse Predicate` (so, we cannot add profiled loop predicates anymore). But that might be an edge case.

rwestrel:
Thanks for the explanation. Can a test case be added?

chhagedorn:
Thanks Roland for your review!

I was able to come up with a test that fails with the same graph pattern. I've pushed an update.

TobiHartmann (approved):
Looks good to me.

chhagedorn:
Thanks Tobias for your review! I've simplified the test some more but the fix remains the same.

TobiHartmann (approved):
Still looks good.

PR #14202, opened 2023-05-29
8308961: Automatically find configuration matching name of checked of branch

edvbld:
Hi all,

please review this smaller patch to the build system. This patch changes what the build system will do when there are multiple configurations available and none has been selected (neither via `CONF` nor via `SPEC`). Instead of printing an error message (current behavior) the build system will instead check if the name of any configuration exactly matches the name of the checked out Git branch. If so, then that configuration will be built.

The rationale for this change is that many developers (including me) use branches to work on multiple things concurrently. Having (at least) one configuration per branch, named after the branch, is a very natural model then for setting up one's build configurations. Having the build configuration corresponding to the checked out Git branch built automatically is then very convenient (instead of typing `make CONF=$(git branch --show-current)` every time).

A couple of design considerations:
- why use `$$(shell command -v git 2>/dev/null)` instead of `$$(GIT)` from autoconf? Because we don't have a `spec.gmk` available because we haven't chosen a configuration yet, so we don't have `$$(GIT)`.
- why use `git rev-parse --abbrev-ref HEAD` instead of `git branch --show-current`? Because `git rev-parse --abbrev-ref` has been around since [2009](https://github.com/git/git/commit/a45d34691ea624e93863e95706eeb1b1909304f3) and `git branch --show-current` was introduced in Git 2.22 in 2019 (plus that `rev-parse --abbrev-ref` works with a detached `HEAD`). `git rev-parse --is-inside-work-tree` is from [2007](https://github.com/git/git/commit/892c41b98ae2e6baf3aa13901cb10db9ac67d2f3), but I think requiring a Git installation more recent than 2009 should be ok.
- why match the branch name exactly instead of a looser matching? This could be beneficial if branches are named e.g. `$(git branch --show-current)-fastdebug`, but I wanted to start out with something strict and then the matching can be made looser if needed.
- is this backwards compatible? Yes, at least I think so. Previously the build system would return an error but now it might build a configuration, that seems compatible. The patch does not interfere at all with the code paths where `CONF` or `SPEC` has been set.

### Testing
- [x] Tested locally on macOS/aarch64 and Linux/x64 with and without configurations matching the named Git branch.
- [x] Tested locally on macOS/aarch64 and Linux/x64 without a Git client.
- [x] Tested locally on macOS/aarch64 and Linux/x64 with a source directory without a `.git` directory (i.e. not a repository).

Thanks,
Erik

---------
### Progress

### Issue
 * [JDK-8308961](https://bugs.openjdk.org/browse/JDK-8308961): Automatically find configuration matching name of checked of branch (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14202/head:pull/14202` \
`$ git checkout pull/14202`

`$ git checkout pull/14202` \
`$ git pull https://git.openjdk.org/jdk.git pull/14202/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14202`

`$ git pr show -t 14202`

</details>

<a href="https://git.openjdk.org/jdk/pull/14202.diff">https://git.openjdk.org/jdk/pull/14202.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14202#issuecomment-1567142469)

### DIFF
--- a/make/InitSupport.gmk
+++ b/make/InitSupport.gmk
@@ -243,15 +243,16 @@ ifeq ($(HAS_SPEC),)
           GIT_IS_INSIDE_WORK_TREE := $$(shell $$(GIT) rev-parse --is-inside-work-tree 2>/dev/null)
           ifeq ($$(GIT_IS_INSIDE_WORK_TREE), true)
             GIT_CURRENT_BRANCH := $$(shell $$(GIT) rev-parse --abbrev-ref HEAD)
-            GIT_BRANCH_CONF := $$(strip $$(filter $$(GIT_CURRENT_BRANCH), $$(all_confs)))
+            GIT_BRANCH_CONFS := $$(strip $$(foreach var, $$(all_confs), \
+                $$(if $$(findstring $$(GIT_CURRENT_BRANCH), $$(var)), $$(var))))
           endif
         endif
         ifeq ($$(words $$(all_spec_files)), 1)
           # We found exactly one configuration, use it
           SPECS := $$(strip $$(all_spec_files))
-        else ifeq ($$(words $$(GIT_BRANCH_CONF)), 1)
-          # We found exactly one configuration with the name of the checked out Git branch, use it
-          SPECS := $$(build_dir)/$$(GIT_BRANCH_CONF)/spec.gmk
+        else ifeq ($$(words $$(GIT_BRANCH_CONFS)), 1)
+          # We found exactly one configuration containing the name of the checked out Git branch, use it
+          SPECS := $$(build_dir)/$$(GIT_BRANCH_CONFS)/spec.gmk
         else
           $$(info Error: No CONF given, but more than one configuration found.)
           $$(info Available configurations in $$(build_dir):)
--- a/make/InitSupport.gmk
+++ b/make/InitSupport.gmk
@@ -238,7 +238,18 @@ ifeq ($(HAS_SPEC),)
         SPECS := $$(addsuffix /spec.gmk, $$(addprefix $$(build_dir)/, $$(matching_confs)))
       else
         # No CONF or SPEC given, check the available configurations
-        ifneq ($$(words $$(all_spec_files)), 1)
+        GIT := $$(shell command -v git 2>/dev/null)
+        ifneq ($$(GIT),)
+          GIT_CURRENT_BRANCH := $$(shell $$(GIT) rev-parse --abbrev-ref HEAD)
+          GIT_BRANCH_CONF := $$(strip $$(filter $$(GIT_CURRENT_BRANCH), $$(all_confs)))
+        endif
+        ifeq ($$(words $$(all_spec_files)), 1)
+          # We found exactly one configuration, use it
+          SPECS := $$(strip $$(all_spec_files))
+        else ifeq ($$(words $$(GIT_BRANCH_CONF)), 1)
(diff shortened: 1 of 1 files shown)


make/InitSupport.gmk
```diff
@@ -238,17 +238,28 @@ ifeq ($(HAS_SPEC),)
...
+        ifneq ($$(GIT),)
+          GIT_IS_INSIDE_WORK_TREE := $$(shell $$(GIT) rev-parse --is-inside-work-tree 2>/dev/null)
+          ifeq ($$(GIT_IS_INSIDE_WORK_TREE), true)
+            GIT_CURRENT_BRANCH := $$(shell $$(GIT) rev-parse --abbrev-ref HEAD)
+            GIT_BRANCH_CONF := $$(strip $$(filter $$(GIT_CURRENT_BRANCH), $$(all_confs)))
+          endif
+        endif
+        ifeq ($$(words $$(all_spec_files)), 1)
+          # We found exactly one configuration, use it
+          SPECS := $$(strip $$(all_spec_files))
```
dholmes-ora, line 251: It may be better to keep this as the first action and only do the git checks if necessary. That would minimise potential interference with single branch repos.

dholmes-ora (commented):
I'm not sure how useful this is in general as it very much depends on your own style of working. I use branch-based configurations as well but I use a wrapper script to manage that - in part because the wrapper will also run configure if no configuration currently exists. The main limitation I see with the proposal (and I may be misreading it) is that the exact name match seems to preclude working when your configurations have the form `mybranch` and `mybranch-debug` etc.

make/InitSupport.gmk
edvbld: That is a good idea, and I thought about it, but it is hard to structure the logic that way and not end up with duplicated code (early return doesn't seem to be a thing in GNU Make macros). What I means that the following is (at least to me) hard to express in a GNU Make macro: 

```
ifeq ($$(words $$(all_spec_files)), 1)
   # Somehow set SPECS and return, do not continue to execute
endif
GIT := ...
ifneq ($$(GIT),)
   ...
endif
...
```
The two Git commands potentially being executed are `O(1)`, they do not scale with repository size and should have minimal overhead. Since both flags to `git rev-parse` being used have been around since 2009 I think the risk of encountering a Git installation where they do not exist or do not work is very low. The `rev-parse` command is a so called "plumbing" command in Git which means that the Git project considers the CLI interface an API and keep it backwards compatible (so we shouldn't run in to future Git releases renaming and/or removing these flags). I therefore think it is worth keeping the logic a bit nicer and instead run these two extra `git rev-parse` commands and one extra `command -v` invocation (sometimes a shell built-in) for the single branch repo case.

erikj79: Only evaluating the external commands when needed is a hard requirement here. Running a few extra external commands on every make invocation may not be noticeable on Linux, but on Windows it quickly adds up, and we have done several passes through the makefiles to eliminate those over the years. You should be able to avoid repeated code by nesting if blocks.

It may also be a good idea be explicit about running `git` inside `$(topdir)` (defined in `Makefile`), but that can be debated. That variable is always pointing to the top of the OpenJDK repository. Your current implementation is based on CWD which may be in some other repository depending on the workspace setup. If going with `$(topdir)` the first git invocation can be replaced with checking for `$(wildcard $(topdir)/.git)`.

PR #14207, opened 2023-05-29
8307648: java/net/httpclient/ExpectContinueTest.java timed out

dfuch:
Please find here a fix for the java/net/httpclient/ExpectContinueTest.java which was failing intermittently.

It turns out there was two bugs here:

- on on the server side, which added the END_STREAM flag to the HeaderFrame containing the 100 response.
  This can obviously never be right since 100 is an intermediary response, which should be followed by a final response.
  
- on the client side, because this was interpreted as "there will be no body" - but it didn't prevent the client from waiting for the final response.

With this fix a 100 response carrying the END_STREAM flag will cause the exchange to be cancelled. I manually verified that this worked before fixing the server side.

The logic on server side is fixed to ignore whatever length is passed to sendResponseHeaders() if 100 is supplied, and to not add the END_STREAM flag to the HeaderFrame that carries the 100 response (which is a side effect of setting the expected length parameter to 0 in that case).

---------
### Progress

### Issue
 * [JDK-8307648](https://bugs.openjdk.org/browse/JDK-8307648): java/net/httpclient/ExpectContinueTest.java timed out

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14207/head:pull/14207` \
`$ git checkout pull/14207`

`$ git checkout pull/14207` \
`$ git pull https://git.openjdk.org/jdk.git pull/14207/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14207`

`$ git pr show -t 14207`

</details>

<a href="https://git.openjdk.org/jdk/pull/14207.diff">https://git.openjdk.org/jdk/pull/14207.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14207#issuecomment-1567364243)

### DIFF
--- a/src/java.net.http/share/classes/jdk/internal/net/http/Stream.java
+++ b/src/java.net.http/share/classes/jdk/internal/net/http/Stream.java
@@ -519,16 +519,20 @@ class Stream<T> extends ExchangeImpl<T> {
                     .firstValueAsLong(":status")
                     .orElseThrow(() -> new IOException("no statuscode in response"));
             // If informational code, response is partially complete
-            if (responseCode < 100 || responseCode > 199)
+            if (responseCode < 100 || responseCode > 199) {
                 this.finalResponseCodeReceived = true;
-            if (responseCode == 100 && hf.getFlag(HeaderFrame.END_STREAM)
-                    && !finalResponseCodeReceived) {
+            } else if (hf.getFlag(HeaderFrame.END_STREAM)) {
+                // see RFC 9113 section 8.1:
+                // A HEADERS frame with the END_STREAM flag set that carries an
+                // informational status code is malformed
+                String msg = ("Stream %s PROTOCOL_ERROR: " +
+                        "HEADERS frame with status %s has END_STREAM flag set")
+                        .formatted(streamid, responseCode);
                 if (debug.on()) {
-                    debug.log("Received 100 statusCode, but FIN bit is set");
+                    debug.log(msg);
                 }
-                cancelImpl(new IOException("stream closed after 100, no other response expected"));
+                cancelImpl(new IOException(msg), ResetFrame.PROTOCOL_ERROR);
             }
-
 
             response = new Response(
                     request, exchange, responseHeaders, connection(),
--- a/src/java.net.http/share/classes/jdk/internal/net/http/Stream.java
+++ b/src/java.net.http/share/classes/jdk/internal/net/http/Stream.java
@@ -592,7 +592,7 @@ class Stream<T> extends ExchangeImpl<T> {
                     requestBodyCF.complete(null);
                 }
             }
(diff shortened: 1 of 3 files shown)


src/java.net.http/share/classes/jdk/internal/net/http/Stream.java
```diff
@@ -521,6 +521,14 @@ protected void handleResponse() throws IOException {
             // If informational code, response is partially complete
             if (responseCode < 100 || responseCode > 199)
                 this.finalResponseCodeReceived = true;
+            if (responseCode == 100 && hf.getFlag(HeaderFrame.END_STREAM)
+                    && !finalResponseCodeReceived) {
+                if (debug.on()) {
+                    debug.log("Received 100 statusCode, but FIN bit is set");
+                }
+                cancelImpl(new IOException("stream closed after 100, no other response expected"));
```
djelinski, line 529: This will produce the wrong error code; see https://datatracker.ietf.org/doc/html/rfc9113#name-http-message-framing :
> An interim response consists of a [HEADERS](https://datatracker.ietf.org/doc/html/rfc9113#HEADERS) frame (which might be followed by zero or more [CONTINUATION](https://datatracker.ietf.org/doc/html/rfc9113#CONTINUATION) frames) containing the control data and header section of an interim (1xx) HTTP response (see [Section 15](https://www.rfc-editor.org/rfc/rfc9110#section-15) of [[HTTP](https://datatracker.ietf.org/doc/html/rfc9110)]). A [HEADERS](https://datatracker.ietf.org/doc/html/rfc9113#HEADERS) frame with the END_STREAM flag set that carries an informational status code is [malformed](https://datatracker.ietf.org/doc/html/rfc9113#malformed) ([Section 8.1.1](https://datatracker.ietf.org/doc/html/rfc9113#malformed)).

> Malformed requests or responses that are detected MUST be treated as a [stream error](https://datatracker.ietf.org/doc/html/rfc9113#StreamErrorHandler) ([Section 5.4.2](https://datatracker.ietf.org/doc/html/rfc9113#StreamErrorHandler)) of type [PROTOCOL_ERROR](https://datatracker.ietf.org/doc/html/rfc9113#PROTOCOL_ERROR).

as far as I could tell, `cancelImpl` sends `CANCEL` error code.

djelinski (approved):
LGTM.
Given that this alters the client behavior, please add a test or a noreg label.

dfuch:
I have added 8307648 to the `@bug` tag in ExpectContinueTest, and also logged https://bugs.openjdk.org/browse/JDK-8309118 to add more deterministic tests for various scenarios.

djelinski (approved):

PR #14208, opened 2023-05-29
8309065: Move the logic to determine archive heap location from CDS to G1 GC

ashu-mehra:
This patch is the first step towards having a single set of GC APIs for allocating heap space for the archived objects (See https://bugs.openjdk.org/browse/JDK-8296263).
It moves some of the G1 specific logic from CDS to G1 gc without changing the functionality.

Changes that add/update GC APIs for handling archive heap would be introduced in upcoming patches.

---------
### Progress

### Issue
 * [JDK-8309065](https://bugs.openjdk.org/browse/JDK-8309065): Move the logic to determine archive heap location from CDS to G1 GC (**Sub-task** - P4)

### Reviewers
 * [Ioi Lam](https://openjdk.org/census#iklam) (@iklam - **Reviewer**) ⚠️ Review applies to [caccaa12](https://git.openjdk.org/jdk/pull/14208/files/caccaa1244ff908c339e97149abe263b53cd1244)
 * [Thomas Stuefe](https://openjdk.org/census#stuefe) (@tstuefe - **Reviewer**) ⚠️ Review applies to [82ddb97c](https://git.openjdk.org/jdk/pull/14208/files/82ddb97cf121ac669237398eaf59f556157a1be7)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14208/head:pull/14208` \
`$ git checkout pull/14208`

`$ git checkout pull/14208` \
`$ git pull https://git.openjdk.org/jdk.git pull/14208/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14208`

`$ git pr show -t 14208`

</details>

<a href="https://git.openjdk.org/jdk/pull/14208.diff">https://git.openjdk.org/jdk/pull/14208.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14208#issuecomment-1567527249)

### DIFF
--- a/src/hotspot/share/cds/filemap.cpp
+++ b/src/hotspot/share/cds/filemap.cpp
@@ -2094,17 +2094,6 @@ address FileMapInfo::heap_region_mapped_address() {
 }
 
 bool FileMapInfo::map_heap_region() {
-  init_heap_region_relocation();
-
-  if (_heap_pointers_need_patching) {
-    char* bitmap_base = map_bitmap_region();
-    if (bitmap_base == nullptr) {
-      log_info(cds)("CDS heap cannot be used because bitmap region cannot be mapped");
-      _heap_pointers_need_patching = false;
-      return false;
-    }
-  }
-
   if (map_heap_region_impl()) {
 #ifdef ASSERT
     // The "old" regions must be parsable -- we cannot have any unused space
@@ -2130,66 +2119,28 @@ bool FileMapInfo::map_heap_region() {
   }
 }
 
-void FileMapInfo::init_heap_region_relocation() {
-  assert(UseG1GC, "the following code assumes G1");
-  _heap_pointers_need_patching = false;
-
-  MemRegion heap_range = G1CollectedHeap::heap()->reserved();
-  MemRegion archive_range = get_heap_region_requested_range();
-
-  address requested_bottom = (address)archive_range.start();
-  address heap_end = (address)heap_range.end();
-  assert(is_aligned(heap_end, HeapRegion::GrainBytes), "must be");
-
-  // We map the archive heap region at the very top of the heap to avoid fragmentation.
-  // To do that, we make sure that the bottom of the archived region is at the same
-  // address as the bottom of the highest possible G1 region.
-  address mapped_bottom = heap_end - align_up(archive_range.byte_size(), HeapRegion::GrainBytes);
-
-  if (UseCompressedOops &&
-      (narrow_oop_mode() != CompressedOops::mode() ||
-       narrow_oop_shift() != CompressedOops::shift())) {
-    log_info(cds)("CDS heap data needs to be relocated because the archive was created with an incompatible oop encoding mode.");
-    _heap_pointers_need_patching = true;
-  } else if (requested_bottom != mapped_bottom) {
-    log_info(cds)("CDS heap data needs to be relocated because it is mapped at a different address @ " INTPTR_FORMAT,
-                  p2i(mapped_bottom));
-    _heap_pointers_need_patching = true;
-  }
-
-  ptrdiff_t delta = 0;
-  if (_heap_pointers_need_patching) {
-    delta = mapped_bottom - requested_bottom;
-  }
-
-  log_info(cds)("CDS heap data relocation delta = " INTX_FORMAT " bytes", delta);
-  ArchiveHeapLoader::init_mapped_heap_relocation(delta, narrow_oop_shift());
-}
-
 bool FileMapInfo::map_heap_region_impl() {
+  assert(UseG1GC, "the following code assumes G1");
+
   FileMapRegion* r = region_at(MetaspaceShared::hp);
   size_t size = r->used();
-
-  if (size > 0) {
-    HeapWord* start = (HeapWord*)heap_region_mapped_address();
-    _mapped_heap_memregion = MemRegion(start, size / HeapWordSize);
-    log_info(cds)("Trying to map heap data at " INTPTR_FORMAT ", size = " SIZE_FORMAT_W(8) " bytes",
-                  p2i(start), size);
-  } else {
+  if (size == 0) {
     return false; // no archived java heap data
   }
 
-  // Check that the region is within the java heap
-  if (!G1CollectedHeap::heap()->check_archive_addresses(_mapped_heap_memregion)) {
-    log_info(cds)("Unable to allocate region, range is not within java heap.");
+  size_t word_size = size / HeapWordSize;
+  address requested_start = heap_region_requested_address();
+
+  log_info(cds)("Preferred address to map heap data (to avoid relocation) is " INTPTR_FORMAT, p2i(requested_start));
+
+  // allocate from java heap
+  HeapWord* start = G1CollectedHeap::heap()->alloc_archive_regions(word_size);
+  if (start == nullptr) {
+    log_info(cds)("UseSharedSpaces: Unable to allocate java heap region for archive heap.");
     return false;
   }
 
-  // allocate from java heap
-  if (!G1CollectedHeap::heap()->alloc_archive_regions(_mapped_heap_memregion)) {
-    log_info(cds)("Unable to allocate region, java heap range is already in use.");
-    return false;
-  }
+  _mapped_heap_memregion = MemRegion(start, word_size);
 
   // Map the archived heap data. No need to call MemTracker::record_virtual_memory_type()
   // for mapped region as it is part of the reserved java heap, which is already recorded.
@@ -2208,10 +2159,37 @@ bool FileMapInfo::map_heap_region_impl() {
   r->set_mapped_base(base);
   if (VerifySharedSpaces && !r->check_region_crc()) {
     dealloc_heap_region();
-    log_info(cds)("mapped heap region is corrupt");
+    log_info(cds)("UseSharedSpaces: mapped heap region is corrupt");
     return false;
   }
 
+  // If the requested range is different from the range allocated by GC, then
+  // the pointers need to be patched.
+  address mapped_start = (address) _mapped_heap_memregion.start();
+  ptrdiff_t delta = mapped_start - requested_start;
+  if (UseCompressedOops &&
+      (narrow_oop_mode() != CompressedOops::mode() ||
+       narrow_oop_shift() != CompressedOops::shift())) {
+    _heap_pointers_need_patching = true;
+  }
+  if (delta != 0) {
+    _heap_pointers_need_patching = true;
+  }
+  ArchiveHeapLoader::init_mapped_heap_relocation(delta, narrow_oop_shift());
+
+  if (_heap_pointers_need_patching) {
+    char* bitmap_base = map_bitmap_region();
+    if (bitmap_base == NULL) {
+      log_info(cds)("CDS heap cannot be used because bitmap region cannot be mapped");
+      dealloc_heap_region();
+      unmap_region(MetaspaceShared::hp);
+      _heap_pointers_need_patching = false;
+      return false;
+    }
+  }
+  log_info(cds)("Heap data mapped at " INTPTR_FORMAT ", size = " SIZE_FORMAT_W(8) " bytes",
+                p2i(mapped_start), _mapped_heap_memregion.byte_size());
+  log_info(cds)("CDS heap data relocation delta = " INTX_FORMAT " bytes", delta);
   return true;
 }
 
@@ -2265,7 +2243,6 @@ void FileMapInfo::unmap_regions(int regions[], int num_regions) {
 // Unmap a memory region in the address space.
 
 void FileMapInfo::unmap_region(int i) {
-  assert(!HeapShared::is_heap_region(i), "sanity");
   FileMapRegion* r = region_at(i);
   char* mapped_base = r->mapped_base();
   size_t size = r->used_aligned();
--- a/src/hotspot/share/cds/filemap.cpp
+++ b/src/hotspot/share/cds/filemap.cpp
@@ -2126,7 +2126,7 @@ bool FileMapInfo::map_heap_region_impl() {
   log_info(cds)("Preferred address to map heap data (to avoid relocation) is " INTPTR_FORMAT, p2i(requested_start));
 
   // allocate from java heap
-  HeapWord* start = G1CollectedHeap::heap()->alloc_archive_regions(word_size);
+  HeapWord* start = G1CollectedHeap::heap()->alloc_archive_region(word_size, (HeapWord*)requested_start);
   if (start == nullptr) {
     log_info(cds)("UseSharedSpaces: Unable to allocate java heap region for archive heap.");
     return false;
--- a/src/hotspot/share/cds/filemap.cpp
+++ b/src/hotspot/share/cds/filemap.cpp
@@ -2085,14 +2085,6 @@ address FileMapInfo::heap_region_requested_address() {
   }
 }
 
-// The address where this shared heap region is actually mapped at runtime. This function
-// can be called only after we have determined the value for ArchiveHeapLoader::mapped_heap_delta().
-address FileMapInfo::heap_region_mapped_address() {
-  assert(UseSharedSpaces, "runtime only");
-  assert(ArchiveHeapLoader::can_map(), "cannot be used by ArchiveHeapLoader::can_load() mode");
-  return heap_region_requested_address() + ArchiveHeapLoader::mapped_heap_delta();
-}
-
 bool FileMapInfo::map_heap_region() {
   if (map_heap_region_impl()) {
 #ifdef ASSERT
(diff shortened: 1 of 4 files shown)


iklam:
Hi Ashutosh, I've started to test your patch and so far it looks good. I tried dumping with -Xmx128m and running with -Xmx{128, 256, 512m} and no heap relocation is needed. This is the same as before, so I think the logic is unchanged.

With larger heaps (e.g. -Xmx8g), we always relocate, but that's the same as before your change. I've filed a bug for it [JDK-8309170](https://bugs.openjdk.org/browse/JDK-8309170)

`FileMapInfo::heap_region_mapped_address()` was a round-about way of calculating the bottom address of the mapped region. You new code has simplified the logic so this function is no longer needed. It can be removed.

ashu-mehra:
@iklam thanks for testing the patch. I have remove the unused method `FileMapInfo::heap_region_mapped_address()`.

iklam (approved):
The changes look good to me. It passed tiers 1-4 in our CI.

src/hotspot/share/cds/filemap.cpp
```diff
@@ -2130,67 +2111,29 @@ bool FileMapInfo::map_heap_region() {
...
-  }
+  size_t word_size = size / HeapWordSize;
+  address requested_start = heap_region_requested_address();
+
+  log_info(cds)("Preferred address to map heap data (to avoid relocation) is " INTPTR_FORMAT, p2i(requested_start));
 
   // allocate from java heap
-  if (!G1CollectedHeap::heap()->alloc_archive_regions(_mapped_heap_memregion)) {
-    log_info(cds)("Unable to allocate region, java heap range is already in use.");
+  HeapWord* start = G1CollectedHeap::heap()->alloc_archive_regions(word_size);
```
tschatzl, line 2129: I'm not convinced not giving a preferred location is a good idea. That seems to reduce the opportunity to directly map archives significantly. Previously, with only heap size changes, the archive could be mapped in still.

How that reduces the API to the GC seems unclear, this call is still embedded in G1 specific code.

Since this change is an intermediate step, could you provide an overview of the final API/change too? It is hard to comment on this without knowing where you are going with that.

Without knowing where this ends up, I would prefer if this method passed in the full suggested memory range (or like memory reservation works, pass in a `requested_addr`/`requested_start` that is null in case it's not used if you prefer) instead of just the size.
The method may still return the actual location. Other collectors may simply ignore that hint (which is better than the caller not even bothering to give the hint).

ashu-mehra:
@tschatzl 

> I'm not convinced not giving a preferred location is a good idea. That seems to reduce the opportunity to directly map archives significantly. Previously, with only heap size changes, the archive could be mapped in still.

I am not sure I get this. The patch does not change the ability to map the archive. It just moved the calculation to map the archive region from CDS to G1. Before this patch CDS code would determine the address for mapping the archive space towards the top of the heap and pass that address to G1. This patch just moves that calculation to G1. So it should be at-par with the current state. If it is not, please point out and I can work on fixing that.

> Since this change is an intermediate step, could you provide an overview of the final API/change too? It is hard to comment on this without knowing where you are going with that.

Ok, I have updated the description of the main issue [JDK-8296263](https://bugs.openjdk.org/browse/JDK-8296263) with some details on the expected changes in the future patches.
My main aim with this work is mainly code reorganization to avoid using different GC APIs in CDS depending on the GC policy in use. When this is completed I expect it to provide same functionality as today. Any enhancement, like passing preferred location to map archive heap, can be built on top of this.

Hope this helps.

iklam:
> @tschatzl
> 
> > I'm not convinced not giving a preferred location is a good idea. That seems to reduce the opportunity to directly map archives significantly. Previously, with only heap size changes, the archive could be mapped in still.
> 
> I am not sure I get this. The patch does not change the ability to map the archive. It just moved the calculation to map the archive region from CDS to G1. Before this patch CDS code would determine the address for mapping the archive space towards the top of the heap and pass that address to G1. This patch just moves that calculation to G1. So it should be at-par with the current state. If it is not, please point out and I can work on fixing that.
> 
> > Since this change is an intermediate step, could you provide an overview of the final API/change too? It is hard to comment on this without knowing where you are going with that.
> 
> Ok, I have updated the description of the main issue [JDK-8296263](https://bugs.openjdk.org/browse/JDK-8296263) with some details on the expected changes in the future patches. My main aim with this work is mainly code reorganization to avoid using different GC APIs in CDS depending on the GC policy in use. When this is completed I expect it to provide same functionality as today. Any enhancement, like passing preferred location to map archive heap, can be built on top of this.
> 
> Hope this helps.

Hi Ashutosh,

You are right that in the existing code, although filemap.cpp finds out where the requested range is, it doesn't actually pass that to G1. It just requests G1 to reserve a range at the end of the runtime heap. So your PR preserves this behavior.

I think Thomas's point is, the requested range should be passed to the `alloc_archive_regions()` API, even though the collector may simply ignore it.

In the past, the archive G1 regions were not movable, so it was preferable to put them at the end of the heap, even though that might cause relocation. Now that the archive regions are just regular "old" regions, which can move, it may be preferable to reserve them at the requested range.

BTW, perhaps `alloc_archive_regions()` should be renamed to `alloc_archive_range()` going forward.  The plural form of "regions" sounds odd for non-region based collectors.

tschatzl:
> Hi Ashutosh,
> 
> You are right that in the existing code, although filemap.cpp finds out where the requested range is, it doesn't actually pass that to G1. It just requests G1 to reserve a range at the end of the runtime heap. So your PR preserves this behavior.
> 
> I think Thomas's point is, the requested range should be passed to the `alloc_archive_regions()` API, even though the collector may simply ignore it.

Exactly, sorry if I wasn't clear enough.

> 
> In the past, the archive G1 regions were not movable, so it was preferable to put them at the end of the heap, even though that might cause relocation. Now that the archive regions are just regular "old" regions, which can move, it may be preferable to reserve them at the requested range.
> 
> BTW, perhaps `alloc_archive_regions()` should be renamed to `alloc_archive_range()` going forward. The plural form of "regions" sounds odd for non-region based collectors.

+1

Thanks,
  Thomas

ashu-mehra:
> I think Thomas's point is, the requested range should be passed to the alloc_archive_regions() API, even though the collector may simply ignore it.

Ok. I have initially planned to do that after completing this work. But lets add it in this patch.

> BTW, perhaps alloc_archive_regions() should be renamed to alloc_archive_range() going forward.

That's a typo. It should be alloc_archive_space(). I hope that is acceptable. Feel free to suggest changes to the APIs if it doesn't sound correct.

ashu-mehra:
> That's a typo. It should be alloc_archive_space().

Oh, I realized its not a typo but the change to alloc_archive_space() would happen in the next patch. So I didn't bother changing the name of existing G1 api. I will update that as well.

ashu-mehra:
> I think Thomas's point is, the requested range should be passed to the alloc_archive_regions() API, even though the collector may simply ignore it.

@tschatzl @iklam Other collectors may ignore the requested addr, but G1 can just allocate the regions at the requested addr, right? Do we also need the fallback to allocating at the top of heap if for some reason allocation at requested addr fails?

iklam:
> > I think Thomas's point is, the requested range should be passed to the alloc_archive_regions() API, even though the collector may simply ignore it.
> 
> @tschatzl @iklam Other collectors may ignore the requested addr, but G1 can just allocate the regions at the requested addr, right? Do we also need the fallback to allocating at the top of heap if for some reason allocation at requested addr fails?

Yes, I think we need the fallback. However, I think we can keep the behavior exactly the same in this PR (only allocate at the heap end), and implement the more optimal allocation in a subsequent PR.

ashu-mehra:
Pushed a commit with suggested changes.

src/hotspot/share/gc/g1/g1CollectedHeap.hpp
```diff
@@ -700,21 +700,16 @@ class G1CollectedHeap : public CollectedHeap {
...
   void iterate_regions_in_range(MemRegion range, const Func& func);
 
-  // Commit the appropriate G1 region(s) containing the specified range
-  // and mark them as 'old' region(s).
-  bool alloc_archive_regions(MemRegion range);
+  // Commit the required number of G1 region(s) according to the size requested
+  // and mark them as 'old' region(s). Preferred address is treated as a hint for
+  // the location of the archive space in the heap. The returned address may or may
+  // not be same as the preferred address.
+  HeapWord* alloc_archive_region(size_t word_size, HeapWord* preferred_addr);
```
iklam, line 712: Sorry to be picky, but I think `region` implies a single region, but G1 could allocate one or more regions to satisfy the request. I think it's better to use `allocate_archive_range(MemRegion requested_range)` to be more neutral. Passing the range in a `MemRegion` will also look similar to the API right above this one.

Maybe we should also change `populate_archive_regions_bot_part` and `dealloc_archive_regions` to use `_range` as well. What do you think, @tschatzl

ashu-mehra: How about replacing `allocate_archive_range` with  `allocate_archive_space` which is actually planned for the next patch?
I am more inclined to keep size and address as separate parameters because the address is just a hint for the collectors, while size is not. With MemRegion as the type this information is not conveyed to the reader without reading the comments/code. This is also the reason why I prefer using "preferred_addr" rather than "requested_addr".
But if you feel otherwise I will update the API to use MemRegion.

> Maybe we should also change populate_archive_regions_bot_part and dealloc_archive_regions to use _range as well

I am not sure if it is necessary to update these APIs, as I would anyway be replacing them with fixup_archive_space() and handle_archive_space_failure() as mentioned in the description of main issue [JDK-8296263](https://bugs.openjdk.org/browse/JDK-8296263). IMO these names are more generic. If you want I can update the code to use these new names in this patch.

iklam: If you plan to change these soon I think it's OK to leave the names as you have for this PR.

ashu-mehra: > If you plan to change these soon I think it's OK to leave the names as you have for this PR.

Okay, then lets do the renaming in the next PR.

ashu-mehra:
@tschatzl let me know if there are any other concerns to address in this patch.

tschatzl (changes requested):
(I can't just request changes without a comment, although I commented before, so here the same again)

Given soon JDK21 fork, and the risk of having this API change for a long time in JDK 21, I am not good with leaving the change as is. Please either fix up the naming or leave integration until after the fork.

ashu-mehra:
> Given the API is in flux, and the follow-up not certain to be reviewed by Thursday I would like to have this patch at least moved to after the JDK 21 fork. Otherwise maintainers need to deal with this awkwardness for a long time unnecessarily.

@tschatzl thanks for clarifying. I agree with your reasoning and I am fine with not having this patch in JDK 21. Having partial changes in JDK 21 doesn't serve any purpose anyway.

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -532,37 +528,34 @@ void G1CollectedHeap::iterate_regions_in_range(MemRegion range, const Func& func
...
-  guarantee(reserved.contains(start_address) && reserved.contains(last_address),
-            "MemRegion outside of heap [" PTR_FORMAT ", " PTR_FORMAT "]",
-            p2i(start_address), p2i(last_address));
-
-  // Perform the actual region allocation, exiting if it fails.
-  // Then note how much new space we have allocated.
+  // Attempt to allocate towards the end of the heap.
+  HeapWord* start_addr = reserved.end() - align_up(word_size, HeapRegion::GrainWords);
+  MemRegion range = MemRegion(start_addr, word_size);
+  HeapWord* last_address = range.last();
```
tstuefe, line 551: Do I understand this - and the old code - correctly: we map the archive region at the end of the heap, therefore, if heap size changed between dump time and runtime, we will always have to relocate? I didn't look but I assume the dumptime heap size is the default heap size?

Another, possibly stupid, question: why don't we use the Heap bottom - we'd still avoid fragmentation but have a much higher chance of not relocating?

tstuefe (approved):
This looks okay, it is a cleaner separation than before.

Small nits and questions only. Approved (not for jdk21 and pending @tschatzl approvement)

One note, I don't see a clear usage of INCLUDE_CDS and INCLUDE_G1GC, and it predates this patch. Don't we do this anymore? We still want to be buildable without CDS but with G1, and vice versa, right?

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
ashu-mehra: > we map the archive region at the end of the heap, therefore, if heap size changed between dump time and runtime, we will always have to relocate

Not always. For small heap sizes under 2G with compressedoops the heap is mapped such that top of the heap is at 4G boundary. So by keeping the archive regions towards the end, we can get same offset even if the heap size changes. If we map towards the bottom, then the offsets would change as the heap size changes.

For example, the default archive is created with 128m. So executing:
```
$ java -Xlog:cds -Xmx128m -version
[0.003s][info][cds] The current max heap size = 128M, HeapRegion::GrainBytes = 1048576
[0.003s][info][cds]     narrow_klass_base = 0x0000000800000000, narrow_klass_shift = 0
[0.003s][info][cds]     narrow_oop_mode = 0, narrow_oop_base = 0x0000000000000000, narrow_oop_shift = 0
[0.003s][info][cds]     heap range = [0x00000000f8000000 - 0x0000000100000000]
[0.003s][info][cds] Preferred address to map heap data (to avoid relocation) is 0x00000000ffe00000
[0.003s][info][cds] Heap data mapped at 0x00000000ffe00000, size =  1071752 bytes
[0.003s][info][cds] CDS heap data relocation delta = 0 bytes
```

Changing the heap size to 256m:
```
$ java -Xlog:cds -Xmx256m -version
[0.003s][info][cds] The current max heap size = 256M, HeapRegion::GrainBytes = 1048576
[0.003s][info][cds]     narrow_klass_base = 0x0000000800000000, narrow_klass_shift = 0
[0.003s][info][cds]     narrow_oop_mode = 0, narrow_oop_base = 0x0000000000000000, narrow_oop_shift = 0
[0.003s][info][cds]     heap range = [0x00000000f0000000 - 0x0000000100000000]
[0.003s][info][cds] Preferred address to map heap data (to avoid relocation) is 0x00000000ffe00000
[0.003s][info][cds] Heap data mapped at 0x00000000ffe00000, size =  1071752 bytes
[0.003s][info][cds] CDS heap data relocation delta = 0 bytes
```

Notice that relocation delta is 0 in both cases and the heap data is mapped at same address.

src/hotspot/share/cds/filemap.cpp
```diff
@@ -2130,67 +2111,29 @@ bool FileMapInfo::map_heap_region() {
...
     return false; // no archived java heap data
   }
 
-  // Check that the region is within the java heap
-  if (!G1CollectedHeap::heap()->check_archive_addresses(_mapped_heap_memregion)) {
-    log_info(cds)("Unable to allocate region, range is not within java heap.");
-    return false;
-  }
+  size_t word_size = size / HeapWordSize;
+  address requested_start = heap_region_requested_address();
```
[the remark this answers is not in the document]
ashu-mehra: Not sure I get this. Are you suggesting replacing `heap_region_requested_address` with `r->mapping_offset()`. But they are not the same, right?

src/hotspot/share/gc/g1/g1CollectedHeap.hpp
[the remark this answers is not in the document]
ashu-mehra: > Could we add this limitation to the comment for the casual reader?

Yes, good point. I will update the comment to add this limitation.

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
ashu-mehra: > I didn't look but I assume the dumptime heap size is the default heap size?

I am wondering why do you think dumptime heap size is the default heap size?  When creating an archive the heap size can be anything. For default archive it is 128m.

iklam: > > I didn't look but I assume the dumptime heap size is the default heap size?
> 
> 
> 
> I am wondering why do you think dumptime heap size is the default heap size?  When creating an archive the heap size can be anything. For default archive it is 128m.

The dump time heap size is set to 128mb for the default archive so that it can be mapped without relocation for small workloads with heaps that can fit under the 4gb boundary.

ashu-mehra:
> One note, I don't see a clear usage of INCLUDE_CDS and INCLUDE_G1GC, and it predates this patch. Don't we do this anymore? We still want to be buildable without CDS but with G1, and vice versa, right?

Yes, that should be possible. There is already an issue for adding missing INCLUDE_CDS: [JDK-8302989](https://bugs.openjdk.org/browse/JDK-8302989)

ashu-mehra:
Thanks @iklam @tschatzl @tstuefe for reviewing the changes.

iklam:
I merged with the latest mainline and started a final test run on our CI. I will sponsor this PR after the test passes.

PR #14209, opened 2023-05-30
JDK-8303549: [AIX] TestNativeStack.java is failing with exit value 1

varada1110:
Stackoverflow exception has caused an exit value 1 when native thread is attaching as daemon. 
On AIX the default stack for posix thread is 192 KB. For this particular test we run thread in pthread_create() which is defined in pthread library thus not following JVM thread creation routines. Hence we have to explicitly provide stack size. 

Reported Issue :[JDK-8303549]( https://bugs.openjdk.org/browse/JDK-8303549)

### DIFF
--- a/test/hotspot/jtreg/runtime/jni/nativeStack/libnativeStack.c
+++ b/test/hotspot/jtreg/runtime/jni/nativeStack/libnativeStack.c
@@ -109,7 +109,17 @@ Java_TestNativeStack_triggerJNIStackTrace
 
   warning = warn;
 
-  if ((res = pthread_create(&thread, NULL, thread_start, NULL)) != 0) {
+#ifdef AIX
+  pthread_attr_t attr;
+  pthread_attr_init(&attr);
+  size_t stack_size = 0x100000;
+  pthread_attr_setstacksize(&attr, stack_size);
+  res = pthread_create(&thread, &attr, thread_start, NULL);
+#else
+  res = pthread_create(&thread, NULL, thread_start, NULL);
+#endif //AIX
+
+  if (res != 0) {
     fprintf(stderr, "TEST ERROR: pthread_create failed: %s (%d)\n", strerror(res), res);
     exit(1);
   }


dholmes-ora (commented):
I'm still a little confused about this issue because we have:
```
./os_cpu/aix_ppc/os_aix_ppc.cpp:size_t os::_java_thread_min_stack_allowed = 64 * K;
```
but the current issues suggests the pthread default of 192K is too small, so 64K is definitely too small!

I would expect we have the same problem with other tests ... though some of them may be (artificially?) restricted to run on Linux. I have a test in https://github.com/openjdk/jdk/pull/14139 that will likely need the same adjustment.

test/hotspot/jtreg/runtime/jni/nativeStack/libnativeStack.c
```diff
@@ -109,7 +109,17 @@ Java_TestNativeStack_triggerJNIStackTrace
 
   warning = warn;
 
-  if ((res = pthread_create(&thread, NULL, thread_start, NULL)) != 0) {
+#ifdef AIX
```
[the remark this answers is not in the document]
varada1110: Yes you are right, then I would change it. 

> I'm still a little confused about this issue because we have:
> 
> ```
> ./os_cpu/aix_ppc/os_aix_ppc.cpp:size_t os::_java_thread_min_stack_allowed = 64 * K;
> ```
> 
> but the current issues suggests the pthread default of 192K is too small, so 64K is definitely too small!
> 
> I would expect we have the same problem with other tests ... though some of them may be (artificially?) restricted to run on Linux. I have a test in #14139 that will likely need the same adjustment.

For this test , the default thread stack size is defined here  : 
`./src/hotspot/os_cpu/aix_ppc/globals_aix_ppc.hpp : define_pd_global(intx, ThreadStackSize,          2048); // 0 => use system default`

`test/hotspot/jtreg/runtime/jni/getCreatedJavaVMs/TestGetCreatedJavaVMs.java `failed with an exit value 134. The test is passing when stack size is explicitly provided.

test/hotspot/jtreg/runtime/jni/nativeStack/libnativeStack.c
```diff
@@ -109,7 +109,17 @@ Java_TestNativeStack_triggerJNIStackTrace
...
+  pthread_attr_t attr;
+  pthread_attr_init(&attr);
+  size_t stack_size = 0x100000;
```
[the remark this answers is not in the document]
varada1110: I got this error when given 1 * M `error: use of undeclared identifier 'M'`
By giving value as `1024 * 1024` also works.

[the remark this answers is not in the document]
dholmes-ora: Sorry I was not thinking straight - this is test code not VM code so no `M`.

test/hotspot/jtreg/runtime/jni/nativeStack/libnativeStack.c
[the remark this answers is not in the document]
dholmes-ora: My point is that according to the definition you should be able to specify `-Xss64k` and it will work fine. But obviously that can't be the case if the default 192K is not sufficient. So `_java_thread_min_stack_allowed` seems to be set wrong for this platform.

dholmes-ora:
@varada1110  I have integrated the change and test for #14139. Once this fix is finalized in form could you apply it to that new test too? Thanks.

varada1110:
> @varada1110 I have integrated the change and test for #14139. Once this fix is finalized in form could you apply it to that new test too? Thanks.

I have applied changes to` test/hotspot/jtreg/runtime/jni/getCreatedJavaVMs/TestGetCreatedJavaVMs.java` as well. 

Thank you

test/hotspot/jtreg/runtime/jni/getCreatedJavaVMs/exeGetCreatedJavaVMs.c
```diff
@@ -91,7 +91,11 @@ int main (int argc, char* argv[]) {
   pthread_t threads[NUM_THREADS];
   for (int i = 0; i < NUM_THREADS; i++ ) {
     printf("[*] Creating thread %d\n", i);
-    int status = pthread_create(&threads[i], NULL, thread_runner, (void *)(intptr_t)i);
+    pthread_attr_t attr;
+    pthread_attr_init(&attr);
+    size_t stack_size = 0x100000;
+    pthread_attr_setstacksize(&attr, stack_size);
+    int status = pthread_create(&threads[i], &attr, thread_runner, (void *)(intptr_t)i);
```
gerard-ziemski, line 98: In real world app, we should use `pthread_attr_destroy(&attr)` after we no longer need it, but in this test it's no big deal (hopefully anyone reading this code will not stop here and learn to leak that resource).

gerard-ziemski (approved):

test/hotspot/jtreg/runtime/jni/getCreatedJavaVMs/exeGetCreatedJavaVMs.c
dholmes-ora: Yes including the destroy would be good code hygiene.

dholmes-ora (commented):
Thanks for the additional update. I will sanity test this in our CI.

dholmes-ora (approved):
Testing was fine.

test/hotspot/jtreg/runtime/jni/getCreatedJavaVMs/exeGetCreatedJavaVMs.c
varada1110: Thanks @gerard-ziemski and @dholmes-ora for the suggestion. I have included `pthread_attr_destroy(&attr)`

test/hotspot/jtreg/runtime/jni/getCreatedJavaVMs/exeGetCreatedJavaVMs.c
```diff
@@ -91,11 +91,22 @@ int main (int argc, char* argv[]) {
...
     if (status != 0) {
       printf("[*] Error creating thread %d - %d\n", i, status);
       exit(-1);
     }
+
+    status = pthread_attr_destroy(&attr);
+    if (status != 0) {
+      printf("pthread_attr_destroy failed: %d - %d\n", i, status);
+      exit(-1);
+    }
```
dholmes-ora, line 109: Nit: `pthread_attr_destroy` has no specified failure modes. We never check its return value elsewhere in the OpenJDK code.

varada1110: I have removed the error check.

Thank you

test/hotspot/jtreg/runtime/jni/getCreatedJavaVMs/exeGetCreatedJavaVMs.c
```diff
@@ -91,11 +91,16 @@ int main (int argc, char* argv[]) {
   pthread_t threads[NUM_THREADS];
   for (int i = 0; i < NUM_THREADS; i++ ) {
     printf("[*] Creating thread %d\n", i);
-    int status = pthread_create(&threads[i], NULL, thread_runner, (void *)(intptr_t)i);
+    pthread_attr_t attr;
+    pthread_attr_init(&attr);
+    size_t stack_size = 0x100000;
+    pthread_attr_setstacksize(&attr, stack_size);
```
dholmes-ora, line 97: Sorry I hate to do this but Just realized that the `pthread-attr` stuff doesn't need to be done on each loop iteration - one attr object can be used for all the threads - and so can be moved outside the loop. Thanks.

dholmes-ora (approved):
Looks good. Thanks for your patience here.

varada1110:
Thank you @dholmes-ora 
/integrate

backwaterred:
Thanks for you work on this @varada1110. I appreciate the time it took to get to the bottom of this one. Even if the code change (by lines) was small, it was a big change :smile:

Seeing that the requirements are met, I'll go ahead and sponsor.

/sponsor

varada1110:
> Thanks for you work on this @varada1110. I appreciate the time it took to get to the bottom of this one. Even if the code change (by lines) was small, it was a big change 😄
> 
> Seeing that the requirements are met, I'll go ahead and sponsor.
> 
> /sponsor

Thank you @backwaterred :sparkles:

PR #14211, opened 2023-05-30
8289220: Locale.forLanguageTag throws NPE due to soft ref used in locale cache being cleared

sunny868:
command: make test CONF=fastdebug JTREG="VM_OPTIONS=-Xcomp" TEST=gc/TestAllocHumongousFragment.java
error info: 

Caused by: java.lang.NullPointerException: Cannot invoke "sun.util.locale.BaseLocale.getVariant()" because "base" is null
at java.base/java.util.Locale.forLanguageTag(Locale.java:1802)
at java.base/sun.util.cldr.CLDRBaseLocaleDataMetaInfo.<clinit>(CLDRBaseLocaleDataMetaInfo.java:41)
... 24 more

Note that the test runs with -XX:ShenandoahGCHeuristics=aggressive -XX:+ShenandoahOOMDuringEvacALot and SoftReferences are involved (LocaleObjectCache uses SoftReferences, used by printf method called in getRandomInstance(Utils.java:511)).

Maybe we have to deal with the case where the getBaseLocale() return value is null. the call stack is:

	at java.base/sun.util.locale.LocaleObjectCache.get(LocaleObjectCache.java:64)
	at java.base/sun.util.locale.BaseLocale.getInstance(BaseLocale.java:169)
	at java.base/sun.util.locale.InternalLocaleBuilder.getBaseLocale(InternalLocaleBuilder.java:524)
	at java.base/java.util.Locale.forLanguageTag(Locale.java:1874)

in LocaleObjectCache.java:64

	 62             if (key == null || newVal == null) {                                
	 63                 // subclass must return non-null key/value object               
	 64                 return null; // run here
	 65             }

---------
### Progress

### Issue
 * [JDK-8289220](https://bugs.openjdk.org/browse/JDK-8289220): Locale.forLanguageTag throws NPE due to soft ref used in locale cache being cleared (**Bug** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14211/head:pull/14211` \
`$ git checkout pull/14211`

`$ git checkout pull/14211` \
`$ git pull https://git.openjdk.org/jdk.git pull/14211/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14211`

`$ git pr show -t 14211`

</details>

<a href="https://git.openjdk.org/jdk/pull/14211.diff">https://git.openjdk.org/jdk/pull/14211.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14211#issuecomment-1568010173)

### DIFF
--- a/src/java.base/share/classes/sun/util/locale/BaseLocale.java
+++ b/src/java.base/share/classes/sun/util/locale/BaseLocale.java
@@ -36,6 +36,7 @@ import jdk.internal.misc.CDS;
 import jdk.internal.util.StaticProperty;
 import jdk.internal.vm.annotation.Stable;
 
+import java.lang.ref.Reference;
 import java.lang.ref.SoftReference;
 import java.util.StringJoiner;
 
@@ -366,7 +367,9 @@ public final class BaseLocale {
             assert (key.holder != null && key.holderRef == null);
             BaseLocale l = key.holder;
             BaseLocale locale = new BaseLocale(l.getLanguage(), l.getScript(), l.getRegion(), l.getVariant(), true);
-            return (new Key(locale)).getBaseLocale();
+            BaseLocale value = (new Key(locale)).getBaseLocale();
+            Reference.reachabilityFence(locale);
+            return value;
         }
     }
 }
--- a/src/java.base/share/classes/sun/util/locale/BaseLocale.java
+++ b/src/java.base/share/classes/sun/util/locale/BaseLocale.java
@@ -272,6 +272,13 @@ public final class BaseLocale {
             this.hash = hashCode(locale);
         }
 
+        private Key(BaseLocale locale) {
+            this.normalized = true;
+            this.holderRef = new SoftReference<>(locale);
+            this.holder = null;
+            this.hash = hashCode(locale);
+        }
+
         public int hashCode() {
             return hash;
         }
@@ -352,7 +359,14 @@ public final class BaseLocale {
 
         @Override
         protected BaseLocale createObject(Key key) {
-            return Key.normalize(key).getBaseLocale();
+            if (key.normalized) {
+                return key.getBaseLocale();
+            }
+
+            assert (key.holder != null && key.holderRef == null);
+            BaseLocale l = key.holder;
+            BaseLocale locale = new BaseLocale(l.getLanguage(), l.getScript(), l.getRegion(), l.getVariant(), true);
+            return (new Key(locale)).getBaseLocale();
         }
     }
 }
--- a/src/java.base/share/classes/sun/util/locale/BaseLocale.java
+++ b/src/java.base/share/classes/sun/util/locale/BaseLocale.java
@@ -367,9 +367,12 @@ public final class BaseLocale {
             assert (key.holder != null && key.holderRef == null);
             BaseLocale l = key.holder;
             BaseLocale locale = new BaseLocale(l.getLanguage(), l.getScript(), l.getRegion(), l.getVariant(), true);
-            BaseLocale value = (new Key(locale)).getBaseLocale();
-            Reference.reachabilityFence(locale);
-            return value;
+            try {
+                BaseLocale value = (new Key(locale)).getBaseLocale();
+                return value;
+            } finally {
+               Reference.reachabilityFence(locale);
+            }
         }
     }
 }
--- a/src/java.base/share/classes/sun/util/locale/LocaleObjectCache.java
+++ b/src/java.base/share/classes/sun/util/locale/LocaleObjectCache.java
@@ -57,9 +57,8 @@ public abstract class LocaleObjectCache<K, V> {
             value = entry.get();
         }
         if (value == null) {
-            key = normalizeKey(key);
             V newVal = createObject(key);
-            if (key == null || newVal == null) {
+            if (newVal == null) {
                 // subclass must return non-null key/value object
                 return null;
             }
--- a/src/java.base/share/classes/sun/util/locale/LocaleObjectCache.java
+++ b/src/java.base/share/classes/sun/util/locale/LocaleObjectCache.java
@@ -60,7 +60,7 @@ public abstract class LocaleObjectCache<K, V> {
             V newVal = createObject(key);
             if (newVal == null) {
                 // subclass must return non-null key/value object
-                return null;
+                throw new NullPointerException();
             }
 
             CacheEntry<K, V> newEntry = new CacheEntry<>(key, newVal, queue);
--- a/src/java.base/share/classes/java/util/Locale.java
+++ b/src/java.base/share/classes/java/util/Locale.java
@@ -1872,6 +1872,9 @@ public final class Locale implements Cloneable, Serializable {
         InternalLocaleBuilder bldr = new InternalLocaleBuilder();
         bldr.setLanguageTag(tag);
         BaseLocale base = bldr.getBaseLocale();
+        if (base == null) {
+            return null;
+        }
         LocaleExtensions exts = bldr.getLocaleExtensions();
         if (exts == null && !base.getVariant().isEmpty()) {
             exts = getCompatibilityExtensions(base.getLanguage(), base.getScript(),
--- a/src/java.base/share/classes/java/util/Locale.java
+++ b/src/java.base/share/classes/java/util/Locale.java
@@ -1872,9 +1872,6 @@ public final class Locale implements Cloneable, Serializable {
         InternalLocaleBuilder bldr = new InternalLocaleBuilder();
         bldr.setLanguageTag(tag);
         BaseLocale base = bldr.getBaseLocale();
-        if (base == null) {
-            return null;
-        }
         LocaleExtensions exts = bldr.getLocaleExtensions();
         if (exts == null && !base.getVariant().isEmpty()) {
             exts = getCompatibilityExtensions(base.getLanguage(), base.getScript(),
--- a/test/micro/org/openjdk/bench/java/util/LocaleDefaults.java
+++ b/test/micro/org/openjdk/bench/java/util/LocaleDefaults.java
@@ -60,5 +60,10 @@ public class LocaleDefaults {
     public Locale getDefaultFormat() {
         return Locale.getDefault(Locale.Category.FORMAT);
     }
+
+    @Benchmark
+    public Locale forLanguageTag() {
+        return Locale.forLanguageTag("en-US-x-lvariant-POSIX");
+    }
 }
 


shipilev:
This also likely has little to do with Shenandoah itself, and it "only" reproduces because Shenandoah has the aggressively cleaning mode. So the synopsis should reflect the exact issue in `Locale`, once found.

sunny868:
as @AlanBateman said, this patch may not be the radical solution.  But once executed here, the ` return null` should be handled.

 62             if (key == null || newVal == null) {                                
 63                 // subclass must return non-null key/value object               
 64                 return null; // then what we can do?
 65             }

-XX:ShenandoahGCHeuristics=aggressive -XX:+ShenandoahOOMDuringEvacALot can trigger frequent GC without caring if the memory is really low. So the problem of soft references can be exposed. I suspect that this code is not executed under normal circumstances, so there is no exposure problem.

AlanBateman:
> as @AlanBateman said, this patch may not be the radical solution. But once executed here, the ` return null` should be handled.

No, we can't return null if the API isn't already specified to return null. I think you'll need to dig into soft ref usages in this code as I assume there is clearing happening at a time that the code doesn't expect. It's unlikely to be specific to ShenandoahGC.

sunny868:
Jtreg tier1 can trigger the same error with vmoptions:"-Xcomp -XX:+UseShenandoahGC -XX:ShenandoahGCHeuristics=aggressive -XX:+ShenandoahOOMDuringEvacALot
I found the GC occurs between when the soft reference is assigned and when it is used.

         private BaseLocale getBaseLocale() {
-            return (holder == null) ? holderRef.get() : holder;
+//            return (holder == null) ? holderRef.get() : holder;
+            if (holder == null) {
+              System.out.println("getBaseLocale this=" + this + "  SoftReference=" + holderRef.get()); // null
+              return holderRef.get();
+            } else {
+              System.out.println("getBaseLocale return holder");
+              return holder;
+            }
         }

The following modification verifies this.
--- a/src/java.base/share/classes/sun/util/locale/BaseLocale.java
+++ b/src/java.base/share/classes/sun/util/locale/BaseLocale.java

@@ -257,19 +261,21 @@ public final class BaseLocale {

         private final boolean normalized;
         private final int hash;
-
+        private BaseLocale locale;    // make locale to a member variable
         private Key(String language, String script, String region,
                     String variant, boolean normalize) {
-            BaseLocale locale = new BaseLocale(language, script, region, variant, normalize);
+            locale = new BaseLocale(language, script, region, variant, normalize);
             this.normalized = normalize;

But this should not be a reasonable solution. I think I should to find a better solution.

AlanBateman:
> Jtreg tier1 can trigger the same error with vmoptions:"-Xcomp -XX:+UseShenandoahGC -XX:ShenandoahGCHeuristics=aggressive -XX:+ShenandoahOOMDuringEvacALot I found the GC occurs between when the soft reference is assigned and when it is used.

Yes, this seems to be a bug here so I think move the issue to core-libs/java.util:i18n as it looks like the caching done by BaseLocale needs to be re-examined.

AlanBateman:
I've moved the issue to core-libs/java.util:i18n and change the title of the JBS issue to make it clearer what this is about. Can you adjust the PR description too as this is not ShenandoahGC issue.

sunny868:
I've done basic testing jtreg tier1-3 and partial shenandoah testing with vmoptions `-Xcomp -XX:+UseShenandoahGC -XX:ShenandoahGCHeuristics=aggressive -XX:+ShenandoahOOMDuringEvacALot`, the results are all OK.

naotoj:
Those SoftReference caches are introduced with this change: https://bugs.openjdk.org/browse/JDK-8196869, and there is a stress test added for it: test/jdk/java/util/Locale/SoftKeys.java
@sunny868 Would you be able to add a JMH test to make sure that your change would not affect the startup time?
@cl4es would you take a look at this change?
Thanks!

naotoj:
As to the patch, would you please elaborate on your changes more? To me, it is simply copying `Key.normalize(key)` content into `Cache.createObject()`, and not sure how it prevents the issue in which the referent got GC'ed during the reference creation and use.

src/java.base/share/classes/sun/util/locale/BaseLocale.java
```diff
@@ -352,7 +359,14 @@ protected Key normalizeKey(Key key) {
...
         protected BaseLocale createObject(Key key) {
-            return Key.normalize(key).getBaseLocale();
+            if (key.normalized) {
+                return key.getBaseLocale();
+            }
+
+            assert (key.holder != null && key.holderRef == null);
+            BaseLocale l = key.holder;
+            BaseLocale locale = new BaseLocale(l.getLanguage(), l.getScript(), l.getRegion(), l.getVariant(), true);
+            return (new Key(locale)).getBaseLocale();
```
sunny868, line 369: Perhaps a more rigorous approach would look like this：
<pre><code class="java">
BaseLocale locale = new BaseLocale(l.getLanguage(), l.getScript(), l.getRegion(), l.getVariant(), true);
BaseLocal value = (new Key(locale)).getBaseLocale();
Reference.reachabilityFence(locale);
return value;
</code></pre>
But the current patch has passed the tests with `-XX:ShenandoahGCHeuristics=aggressive -XX:+ShenandoahOOMDuringEvacALot`, so I think the current patch is OK.

sunny868:
>  @sunny868 Would you be able to add a JMH test to make sure that your change would not affect the startup time? 

All JMH tests or some of them? I had trigger JMH test, waiting for run results.

sunny868:
`make test CONF=fastdebug JTREG="VM_OPTIONS=-Xcomp -XX:+UseShenandoahGC -XX:ShenandoahGCHeuristics=aggressive -XX:+ShenandoahOOMDuringEvacALot" TEST=java/util/Locale/SoftKeys.java` can trigger  same error:
<pre><code class="shell">
Caused by: java.lang.NullPointerException: Cannot invoke "sun.util.locale.BaseLocale.getVariant()" because "base" is null
	at java.base/java.util.Locale.forLanguageTag(Locale.java:1876)
	at java.base/sun.util.cldr.CLDRBaseLocaleDataMetaInfo.<clinit>(CLDRBaseLocaleDataMetaInfo.java:45)
	... 19 more
</code></pre>
this patch can fix it.

sunny868:
Existing JMH Local test results be tested on LOONGARCH64 show no performance fallback.
<pre><code class="shell">
// parent
Benchmark                         Mode  Cnt   Score   Error  Units
LocaleDefaults.getDefault         avgt    3  35.731 ± 0.805  ns/op
LocaleDefaults.getDefaultDisplay  avgt    3  36.099 ± 1.037  ns/op
LocaleDefaults.getDefaultFormat   avgt    3  36.123 ± 1.718  ns/op

// current
Benchmark                         Mode  Cnt   Score   Error  Units
LocaleDefaults.getDefault         avgt    3  35.649 ± 0.264  ns/op
LocaleDefaults.getDefaultDisplay  avgt    3  36.141 ± 1.283  ns/op
LocaleDefaults.getDefaultFormat   avgt    3  36.092 ± 1.371  ns/op

</code></pre>

src/java.base/share/classes/sun/util/locale/BaseLocale.java
naotoj: I still don't see a clear explanation of how the proposed patch fixes the problem. Also, I would appreciate the reasoning in the comments.

sunny868: For hotspot, when GC occurs, it causes all threads to run to the nearest safepoint and then freeze. Generally, safepoints are generated at branch jumps, method ends(ret instructions), loops instructions, and so on. Therefore, the purpose of this patch is to make the creation and use of a softReferences in the same method without branch, jumps and loops in between, that is ensure that GC will not occur in the process of the sofeReferences be created and used.

That's why I didn't use `Reference.reachabilityFence(locale)`.

fisk: This reasoning seems invalid. There are method calls in there, and you rely on inlining heuristics for this not to break. Please use reachabilityFence instead.

rose00: It appears you are assuming that some combination of bytecodes constitutes a  critical section that excludes the GC.  But the JVMS makes no guarantees about GC exclusion across bytecodes.

A thread can be pre-empted at any bytecode (even in the middle of a bytecode).  Another thread can trigger a GC.  Maybe the first thread will be rolled forward to a place convenient to the JVM, but you cannot predict what that will be, because the JVMS does not give you any contract about that matter.

For example, some JITs deoptimize (branch to the interpreter) at unpredictable points for reasons no Java programmer should ever think about (because it’s not in the JVMS contract).  Deoptimizing can allocate storage (for example, to materialize objects whose allocation was deferred by escape analysis).  Thus, it is not safe to assume that any particular bytecode is immune from GC.

Also, some JITs (like C2) inject synthetic safepoints injected as part of arbitrarily complex loop transformations.  A GC at such a safepoint might possibly appear to be tied to a bytecode which is simply a fall-through from a previous bytecode.  This can happen if the loop is rotated, and a fallthrough point begins to function as a back-branch in the IR.

The rule of thumb for non-JIT engineers is, if you find yourself trying to predict what how “the JIT must work”, stop.

The net of this is that, if you need to preserve an object across a critical section, don’t try to read the JIT’s mind or expect it to read yours.  Put in a reachability fence that spans that critical section.

stuart-marks: I don't know what the right fix for this is. If it involves reachabilityFence, then a try-finally statement should be used, with the "critical section" being within the try-clause and the reachabilityFence within the finally-clause.

sunny868: Thanks @fisk @rose00 for the explanation. This patch is indeed based on the idea that small functions are inlined, cause `-Xcomp` be used, and I have seen with parameter `-XX:+PrintCompilation -XX:+PrintInlining` that functions `Key::<init>` and `getBaseLocale` are inlined. 
But as @rose00  say, it's not safe, so reachabilityFence should be used. please review again.

src/java.base/share/classes/sun/util/locale/BaseLocale.java
```diff
@@ -352,7 +360,19 @@ protected Key normalizeKey(Key key) {
...
-            return Key.normalize(key).getBaseLocale();
+            if (key.normalized) {
+                return key.getBaseLocale();
+            }
+
+            assert (key.holder != null && key.holderRef == null);
+            BaseLocale l = key.holder;
+            BaseLocale locale = new BaseLocale(l.getLanguage(), l.getScript(), l.getRegion(), l.getVariant(), true);
+            try {
+                BaseLocale value = (new Key(locale)).getBaseLocale();
```
turbanoff, line 371: I don't understand. Why do we need to create `new Key` here?

djelinski:
In its current form the `createObject` function can return `null`; the object returned by `Key.normalizeKey(key)` holds a soft reference to locale, and by the time `getBaseLocale` is called, the reference may be cleared.

`Key` only uses `SoftReferences` because it must not hold a reference to the object returned by `createObject`. We could refactor `createObject` to return a clone of the `BaseLocale` referenced by `Key`, and then `Key` would not need to use `SoftReference`s.

That being said, the number of long-lived `BaseLocale` objects is very limited; we only keep them in Locale, LocaleKey, and Locale.CONSTANT_LOCALES. Unless I'm missing something, we could solve this problem by removing `BaseLocale` caching, and improve performance at the same time.

src/java.base/share/classes/sun/util/locale/LocaleObjectCache.java
```diff
@@ -57,11 +57,10 @@ public V get(K key) {
             value = entry.get();
         }
         if (value == null) {
-            key = normalizeKey(key);
             V newVal = createObject(key);
-            if (key == null || newVal == null) {
+            if (newVal == null) {
                 // subclass must return non-null key/value object
-                return null;
+                throw new NullPointerException();
```
AlanBateman, line 63: Is this caught somewhere, it seem strange to throw NPE here.

AlanBateman:
> That being said, the number of long-lived `BaseLocale` objects is very limited; we only keep them in Locale, LocaleKey, and Locale.CONSTANT_LOCALES. Unless I'm missing something, we could solve this problem by removing `BaseLocale` caching, and improve performance at the same time.

The motivation seems to be startup ([JDK-8196869](https://bugs.openjdk.org/browse/JDK-8196869)). It looks like it needs to be re-evaluated and if we continue with caching in this area then I think we'll need some stress tests to go with it.

naotoj:
Good point Daniel. I created an issue for re-examining the cache mechanism in `BaseLocale`: https://bugs.openjdk.org/browse/JDK-8309622

PR #14220, opened 2023-05-30
8140326: G1: Consider putting regions where evacuation failed into next collection set

tschatzl:
This change adds management of retained regions, i.e. trying to evacuate evacuation failed regions asap.

The advantage is that evacuation failed regions do not need to wait until the next marking to be cleaned out; as they are often very sparsely occupied (often being eden regions), this occupies a lot of space, potentially causing additional evacuation failures later on.
Another use of this change will be region pinning, which are basically evacuation failed regions that can not be reclaimed as long as they are pinned - however as soon as they are unpinned, they should be reclaimed for the same reasons as well.

It consists of several behavioral changes:

During garbage collection:

... in the Evacuation phase:
* always collect dirty cards into evacuation failed regions proactively. In tests, the amount of cards/live objects per evacuation failed region is typically very small. Dirty cards are always put into the global refinement buffer immediately, assuming that we will keep most if not all evacuation failed regions.

... during Post Evacuation 2/Free Collection Set phase:
* determine whether the region will be retained (kept for "immediate" evacuation) or not. Highly occupied regions are assumed to stay (mostly) live at least until the next marking, so do not bother with them.

These "retained" regions are collected in a new "from retained" set in the collection set candidates and managed separately from "from marking" regions. Having the "from retained" and "from marking" sets separate in the collection set candidates is easier to manage than to use a single list and the picking stuff from it. Particularly wrt to making sure that mixed gcs preferentially pick from the "from marking" list first, then second from the "from retained" list.

... determining the collection set during the pause:
  * during gc, the collection set is preferentially (first) populated with regions from the "from marking" candidates (these are the important regions to get cleaned out), second from the "from retained" list.
  * g1 reserves up to 20% of max gc pause time for retained regions as optional candidates (this is a random number) to make sure that these are cleared out asap to free memory. There is also a minimum number of regions to take from the retained regions list.

During marking

... changes to marking proper
* retained regions will not be marked through during concurrent mark, i.e. they are considered outside of the snapshot. So they are added to the "root regions" during the concurrent start pause. This may be a performance issue (we can't do a gc until all root regions have been marked through), but so far since evacuation failure regions are typically very sparsely populated, this is very fast.

... changes to scrubbing
* during scrubbing, regions may now be reclaimed. That means scrubbing needs to be aware of (more) regions being reclaimed while working on them.

During mutator time:

... try to accomodate retained candidate regions in the predictions, giving them at most 20% of pause time (random value)

Testing: multiple tier1-5 runs, with forced verification on and/or induced evacuation failure

---------
### Progress

### Issue
 * [JDK-8140326](https://bugs.openjdk.org/browse/JDK-8140326): G1: Consider putting regions where evacuation failed into next collection set (**Enhancement** - P4)

### Reviewers
 * [Ivan Walulya](https://openjdk.org/census#iwalulya) (@walulyai - **Reviewer**) ⚠️ Review applies to [364fa823](https://git.openjdk.org/jdk/pull/14220/files/364fa8239f490e84eb678216173cff8a78cfa887)

### Contributors
 * Albert Mingkun Yang `<ayang@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14220/head:pull/14220` \
`$ git checkout pull/14220`

`$ git checkout pull/14220` \
`$ git pull https://git.openjdk.org/jdk.git pull/14220/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14220`

`$ git pr show -t 14220`

</details>

<a href="https://git.openjdk.org/jdk/pull/14220.diff">https://git.openjdk.org/jdk/pull/14220.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14220#issuecomment-1570267170)


tschatzl:
Because I've been asked about why the strict separation of from-marking and retained regions in the policy: to keep the current gc cycle policy fairly intact.
I.e. do the same as before, with some almost optional additional reclamation of known sparsely populated regions (compared to existing code).

The current heuristic to do young gcs, then a fixed amount of mixed gcs that clean out the old gen asap is, as ugly as it is, surprisingly good in the general case. Treating the retained regions the same as from-marking regions would make it necessary to rethink that: Retained regions are often/most of the time prime targets for evacuation (high efficiency), which means that g1 would start concentrating on these regions first even during the mixed phase (which is generally fine...), but due to how mixed gc works (use "smallest young gen", conceptually fixed amount of gcs) it ultimately would not clean out old gen *fast* enough (or completely, depends) as tuned right now. All the low efficiency regions would need to be cleaned out later. However the prediction isn't good enough to cope well with them. They are typically predicted worse than high efficicency ones, that means failing to be exact for them is worse than for high efficiency regions, so it will not take enough of them.

Now one could extend the mixed phase, but in corner cases g1 would then potentially stay in mixed phase (if evacuation failure/and later pinned regions were commonly encountered but still few) forever, which prohibits marking (leading to full gcs), and degrades performance (as it will be using a small young gen).

In some way, evacuation failed regions, can be seen as kind of extra regions/work due to a mismatch between application and VM configuration (compared to current master). Concentrating on that extra work in the phase that's about keeping the gc cycle going without full gc isn't the best thing to do (the current policy just has fairly strong provisions to take low efficiency regions and avoid full gc), particularly it has less overall impact to stuff high efficiency region collection into the existing young gcs that can go almost if not at full speed (i.e. it has less impact to be wrong about the prediction of a high efficiency region vs. low efficiency one).

Basically I think g1 ultimately needs to get away from what mixed gc is now, and how g1 determines when to start/stop that reclamation phase and how to determine the "right" amount of things to evacuate at what speed. That will certainly have to do something with making sure that the old gen allocation rate is countered, and at the same time being efficient overall (i.e. doing more than necessary to do as many young collections as possible within allowed time goal) without degrading into full gcs.

Fwiw: https://bugs.openjdk.org/browse/JDK-8159697.

I tried and failed to do that in this change (to be better than the current heuristic). Apart from being a different topic, this change in itself has its own merit: it improves resiliency vs. evacuation failure. In the past, what you often had is like having an evacuation failure because you ran out of memory. Since this produced lots of garbage, there has been a very high risk of getting into another evacuation failure because the even more decreased available free space causes gcs more often (smaller young gen available), which causes more surviving objects, resulting in more serious evacuation failures (with more live objects). Ultimately you end up with a full gc very quickly because there is not enough time to mark through the old gen.
With this change at least in the next young gc g1 will compact the evacuation failed regions fairly quickly (next gc), allow better recovery and/or allowing G1 to more easily not loose the race with the mutator until the next marking cycle.

Obviously it is also an important stepping stone for handling pinned regions reasonably.

src/hotspot/share/gc/g1/g1CollectionSetChooser.cpp
```diff
@@ -153,11 +152,13 @@ class G1BuildCandidateRegionsTask : public WorkerTask {
       // before we fill them up).
       if (should_add(r) && !G1CollectedHeap::heap()->is_old_gc_alloc_region(r)) {
         add_region(r);
-      } else if (r->is_old()) {
-        // Keep remembered sets for humongous regions, otherwise clean them out.
+        assert(r->rem_set()->is_complete(), "must be %u", r->hrm_index());
```
walulyai, line 155: Maybe the assert comes before the `add_region(r);`

src/hotspot/share/gc/g1/g1CollectionSetChooser.cpp
```diff
@@ -153,11 +152,13 @@ class G1BuildCandidateRegionsTask : public WorkerTask {
       // before we fill them up).
       if (should_add(r) && !G1CollectedHeap::heap()->is_old_gc_alloc_region(r)) {
```
walulyai, line 153: Pre-existing: why not move the `!G1CollectedHeap::heap()->is_old_gc_alloc_region(r)` to `should_add` helper?

src/hotspot/share/gc/g1/g1ParScanThreadState.hpp
```diff
@@ -109,6 +109,16 @@ class G1ParScanThreadState : public CHeapObj<mtGC> {
   PreservedMarks* _preserved_marks;
   EvacuationFailedInfo _evacuation_failed_info;
   G1EvacFailureRegions* _evac_failure_regions;
+  // Number of additional cards into evacuation failed regions enqueued into
+  // the local DCQS. This is an approximation, as cards that would be added later
+  // outside of evacuation failure will not be subtracted again.
+  size_t _evac_failure_enqeued_cards;
```
walulyai, line 115:
```suggestion
  size_t _evac_failure_enqueued_cards;
```

test/hotspot/jtreg/gc/g1/TestVerifyGCType.java
```diff
@@ -144,6 +144,8 @@ private static void testBadVerificationType() throws Exception {
         output = testWithVerificationType(new String[] {"old"});
         output.shouldHaveExitValue(0);
 
+System.out.println(output.getStdout());
```
walulyai, line 147: seems like leftover from debugging

src/hotspot/share/gc/g1/g1CollectionSetChooser.cpp
tschatzl: I already discussed this with @albertnetymk with the result of moving this to a separate commit.

src/hotspot/share/gc/g1/g1Policy.cpp
```diff
@@ -1494,6 +1529,75 @@ double G1Policy::select_candidates_from_marking(G1CollectionCandidateList* marki
   return time_remaining_ms;
 }
 
+void G1Policy::select_candidates_from_retained(G1CollectionCandidateList* retained_list,
+                                               double time_remaining_ms,
+                                               G1CollectionCandidateRegionList* initial_old_regions,
```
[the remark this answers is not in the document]
tschatzl: I would prefer to keep things as is. The overhead isn't big: after some initialization, the main loop will not be entered anyway, but we'd need to duplicate the log messages (or add a special one) in both cases.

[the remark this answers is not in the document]
walulyai: Noted!

walulyai (approved):
Lgtm!

src/hotspot/share/gc/g1/g1CollectionSetCandidates.hpp
```diff
@@ -129,13 +134,17 @@ class G1CollectionCandidateList : public CHeapObj<mtGC> {
   }
 };
 
-// Iterator for G1CollectionSetCandidates.
+// Iterator for G1CollectionSetCandidates. Multiplexes across the marking/retained
+// region lists based on gc efficiency.
```
albertnetymk, line 138: Why does the iteration require ordering? Going through its use sites, I find they just need to perform SIMD like op on each element regardless of the order.

src/hotspot/share/gc/g1/g1Policy.cpp
```diff
@@ -513,6 +513,27 @@ double G1Policy::predict_survivor_regions_evac_time() const {
   return survivor_regions_evac_time;
 }
 
+double G1Policy::predict_retained_regions_evac_time() const {
```
albertnetymk, line 516: Seems that the result is only for log-print in its sole caller, `predict_base_time_ms`. However, it's also odd that this method is called there at all.

src/hotspot/share/gc/g1/g1CollectionSetCandidates.hpp
tschatzl: API consistency. All iterators return them sorted in efficiency order, and if not doing this for that as well will give unnecessary surprises.
I can change this if you insist, but I can already see problems with that in the future.

src/hotspot/share/gc/g1/g1Policy.cpp
tschatzl: This is a bug. It's supposed to be used for calculating the total_time in that method too. Fixed.

src/hotspot/share/gc/g1/g1YoungGCPostEvacuateTasks.cpp
```diff
@@ -342,9 +342,25 @@ class G1PostEvacuateCollectionSetCleanupTask2::ClearRetainedRegionBitmaps : publ
...
+      // Concurrent mark does not mark through regions that we retain (they are root
+      // regions wrt to marking), so we can drop their mark data. The marking information
+      // is also only required in a Concurrent Start pause.
+      bool drop_mark_data = !g1h->collector_state()->in_concurrent_start_gc() ||
+                            g1h->policy()->retain_evac_failed_region(r);
+
+      if (drop_mark_data) {
+        g1h->clear_bitmap_for_region(r);
+        r->reset_top_at_mark_start();
+        cm->clear_statistics(r);
```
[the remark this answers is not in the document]
tschatzl: > I don't get why tams and CM-related data-structure need to be cleared here; shouldn't they be handled inside G1ClearBitmapClosure already?

First, evacuation failure already updated region statistics during concurrent start (as well as g1 updated TAMS during concurrent start).

It is not that evacuation failure and marking are linked, it is that bitmaps and TAMS and statistics are linked. Previous code also assumed that all old regions were marked through, so (as optimizations):

1) Bitmap: g1 keeps the marks on the bitmap of evac-fail regions
2) TAMS: could be kept to top() for all evac-failed regions during concurrent start
3) statistics: evacuation failure did already update them; clearing them in some cases simply been avoided by conditionally updating statistics (maybe remove that `if` now in evacuation failure handling now, what do you think?).

Generally, if you update one, you need to keep the other in sync for marking to work. However none of the marking data cleanup has been necessary before because of that assumption that all old regions are to be marked through.

`G1ClearBitmapClosure` doesn't interact with marking proper (as in: marking live objects on the bitmap), the regions that are cleared in that phase are only those that may not have their bitmap cleared yet due to the "clear bitmaps concurrently" optimization. If that were not the case, g1 would not need that phase.
At that time TAMS and statistics have already been reset by marking, so it does not need to (and both are marking data structures anyway).

That is, the reason to have `G1ClearBitmapClosure` is different compared to `ClearRetainedRegionBitmaps`.

Now that not all old regions may be marked through at concurrent start any more, g1 needs to undo the eagerly updated statistics and TAMS to not confuse other code (mainly marking, but there are asserts wrt to TAMS all over the place).

I do not think there is a way to avoid evacuation failure to interact with concurrent marking (during the concurrent start pause at least) if we want to clear out pinned regions efficiently and quickly.

Maybe making this explicit with something like (didn't test):

```
bool drop_mark_data = g1h->collector_state()->in_concurrent_start_gc() && 
                            g1h->policy()->retain_evac_failed_region(r);
if (drop_mark_data) {
  // clear statistics and tams
} else {
  // tams and statistics must be "set"
}
bool drop_bitmap = !g1h->collector_state()->in_concurrent_start_gc() ||
                            g1h->policy()->retain_evac_failed_region(r);
if (drop_bitmap) {
  // clear bitmap
} else {
  // bitmap must contain marks
}
```
would help?

I thought this would be too complicated for not much gain (resetting TAMS and statistics is little work compared to bitmaps), so I did not. The change I'm going to post is trying to improve the comment.

> (On a higher reasoning, evac-fail processing shouldn't interact with CM; I am assuming the bitmap is not related to CM.)

Evacuation failure already did and still does by a) using mark bitmaps and keeping the marks during concurrent start and b) needing to update marking statistics/tams if necessary (because it keeps marks on the bitmap). This particular change is only required extension to support the new functionality.

I assume we will keep the optimization to keep marks on the bitmap created by evacuation failure (we did in the past, the reason did not change), so evacuation failure be required to keep messing with CM.

I renamed `ClearRetainedRegionBitmaps` to `ClearRetainedRegionData` to make it more clear that retained regions need more processing than just bitmaps now.

tschatzl:
During review, @albertnetymk found that there is a bug in base time calculation when finalizing the young generation: it added retained regions time, which is wrong. Evacuating retained regions is optional and should not count there.

There has also been the question why the policy to take retained regions is different from old regions: as described earlier, retained regions are considered fairly optional to take, and any retained region that one takes is better than the current handling. It is also very hard to come up with a good policy that subsumes both from-marking regions and retained regions; this change does not even attempt that. The current policy for retained regions uses magic constants at this time based on experience that while we want these regions be fairly optional, we still want to account for them when determining eden size somewhat.

Then there has been the question about `ClearRetainedRegionData` again and what marking expects at mark start for particular regions:
* regions that are not going to be marked through need `tams` at `bottom()`, no marks on the bitmap and their `liveness` value zero.
* regions that are going to be marked through need `tams` at `top()`, and either no marks on the bitmap and their `liveness` value zero OR marks on the bitmap and their `liveness` value correspond to the total size of the marked objects. This is the case for evacuation failed regions that are not retained (or, _before_ this change, all evacuation failed regions). There is now some code to verify that during `VerifyAfterGC`.

This verification passed tier 1-5 without any special flag and with all verification (before/during/after including this new verification) on plus `G1EvacuationFailureALot` enabled.

src/hotspot/share/gc/g1/g1YoungGCPostEvacuateTasks.cpp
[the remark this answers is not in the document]
tschatzl: It can, see a prototype at https://github.com/openjdk/jdk/compare/pr/14220...tschatzl:jdk:8140326-evacuate-retained-regions-at-any-time-work?expand=1 ; would that change be what you had in mind?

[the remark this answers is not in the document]
albertnetymk: Yes; the change in `note_start_of_marking` is surprising though. Why do regions in cset need tams? I thought marking doesn't touch cset.

tschatzl:
After some discussion with @albertnetymk we agreed that updating stats (tams, liveness) as implemented in that other branch would be the best way forward in this matter.
There has been a minor optimization to `note_start_of_marking()`: if the calls to it were executed before selecting the collection sets, the condition to set the TAMS could be simplified greatly, because all regions in the collection set must have been candidates before.
This is implemented in the latest commit.
Currently testing this change more thoroughly, with initial local testing having passed.

tschatzl:
I also did another merge with master.

src/hotspot/share/gc/g1/g1CollectionSetCandidates.hpp
albertnetymk: > I can already see problems with that in the future

Can you elaborate? In my mind, those two sources of candidates (marking vs retained) should never be placed in a single ranking.

src/hotspot/share/gc/g1/g1Policy.cpp
```diff
@@ -513,6 +513,27 @@ double G1Policy::predict_survivor_regions_evac_time() const {
...
+  double result = 0.0;
+
+  G1CollectionCandidateList& list = candidates()->retained_regions();
+
+  for (HeapRegion* r : list) {
+    double predicted_time_ms = predict_region_total_time_ms(r, collector_state()->in_young_only_phase());
+
+    if (result + predicted_time_ms >= max_time_for_retaining()) {
+      // Over limit. Exit.
+      break;
```
albertnetymk, line 527: Using a contrived example, if the first region exceeds the limit, this will say retain-time is zero. However, inside young-gc pause, the min num of retained regions is 1, decided in `G1Policy::select_candidates_from_retained`.

This discrepancy seems a bit unsettling. Can one use the min-value here, like its counterpart `calculate_desired_eden_length_before_mixed` above?

src/hotspot/share/gc/g1/g1HeapVerifier.cpp
```diff
@@ -437,6 +437,78 @@ void G1HeapVerifier::verify_region_sets() {
...
+
+  // Verify TAMSes, bitmaps and liveness statistics.
+  //
+  // - if part of marking: TAMS != bottom, liveness == 0, bitmap clear
+  // - if evacuation failed + part of marking: TAMS != bottom, liveness != 0, bitmap has at least on object set (corresponding to liveness)
+  // - if not part of marking: TAMS == bottom, liveness == 0, bitmap clear; must be in root region
+
+  // To compare liveness recorded in G1ConcurrentMark and actual we need to flush the
+  // cache.
+  G1CollectedHeap::heap()->concurrent_mark()->flush_all_task_caches();
```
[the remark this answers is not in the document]
tschatzl: This verification checks consistency after the concurrent start pause. It has been very important during development of this change to verify expected with actual liveness to make sure that the state after that GC is correct/consistent.
I would like to keep this as is.

Now one could also verify after marking is complete (if we do not do already), but the earlier issues are caught, the better to me.

src/hotspot/share/gc/g1/g1Policy.cpp
```diff
@@ -267,7 +267,7 @@ uint G1Policy::calculate_young_desired_length(size_t pending_cards, size_t rs_le
   if (use_adaptive_young_list_length()) {
     desired_eden_length_by_mmu = calculate_desired_eden_length_by_mmu();
 
-    double base_time_ms = predict_base_time_ms(pending_cards, rs_length);
+    double base_time_ms = predict_base_time_ms(pending_cards, rs_length, true /* include_retained */);
```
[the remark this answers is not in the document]
tschatzl: I know, the main reason for this somewhat complicated use is that I wanted to keep a single log message containing all relevant information. I'll look into this again, but not seeing an obviously better option.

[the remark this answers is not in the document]
tschatzl: I made the method non-context sensitive as I do not care too much, unfortunately making debugging more onerous.

tschatzl:
Pushed a new change, testing more thoroughly.

tschatzl:
After some discussion with @albertnetymk we changed the following:
* Iterator for G1CollectionSetCandidates does not need ordering, so removed that code
* there has been a bug in `G1ConcurrentMark::set_live_bytes()` where the code did not divide the given live bytes by HeapWordSize to correctly set live words
* Added some missing includes
* Retained regions evac time prediction for young gen length only considers minimum amount of regions similar to marking regions now.

albertnetymk (approved):

walulyai:
Still LGTM!

tschatzl:
Thanks @albertnetymk @walulyai for your reviews

/integrate

tschatzl:
tier1-5 passed

PR #14221, opened 2023-05-30
8304135: Introduce unified overridden methods query

pavelrappo:
Please review this change.

It turns out that JavaDoc's model for overriding and inheritance for methods is simplistic. As a result, it fails to accommodate upcoming changes in the "method comments algorithm" and the new feature called "directed documentation inheritance".

This change aligns JavaDoc's model for overriding and inheritance for methods with that of the Java Language Specification. The most important part of the change pertains to overridden methods. The model now recognizes that "overrides" is not a transitive relation: from the fact that A overrides B and B overrides C, it does not necessarily follow that A overrides C. The model also recognizes that "overrides" depends on three parameters, not two: the method that overrides, the method that is being overridden, and the class where the override takes place. For details, see the test that this change adds: `TestMethodMembers.java`.

Additionally, the change provides a unified query for method overrides: be it defined in a class or interface, be it abstract or concrete, an overridden method can be queried through a single new API method in `VisibleMemberTable`: 

    public OverrideSequence overrideAt(ExecutableElement method)

That query accepts a method and returns a sequence of methods that the accepted method overrides from the class or interface described by the instance of `VisibleMemberTable` that the query has been called on. That sequence is totally ordered and can be traversed in either direction.

It's quite remarkable that there are only 2 existing tests that required some adjustment after the change. However, there are approximately 250 changed files in the generated JDK API Documentation (i.e. the result of `make docs`). Most of them are benign or beneficial (e.g. they fix a bug). However, there is at least one that will need to be fixed after the upcoming but separate "directed documentation inheritance" update: `java.util.concurrent.LinkedBlockingDeque`.

With this change, it takes the same time (as measured with `TIME(1)`) to run the tests and produce JDK API documentation. Proper benchmarks will be done later on.

---

_Meanwhile, I'll try to figure out a convenient way to attach diffs for JDK API Documentation to this PR._

---------
### Progress

### Issue
 * [JDK-8304135](https://bugs.openjdk.org/browse/JDK-8304135): Introduce unified overridden methods query (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14221/head:pull/14221` \
`$ git checkout pull/14221`

`$ git checkout pull/14221` \
`$ git pull https://git.openjdk.org/jdk.git pull/14221/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14221`

`$ git pr show -t 14221`

</details>

<a href="https://git.openjdk.org/jdk/pull/14221.diff">https://git.openjdk.org/jdk/pull/14221.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14221#issuecomment-1568648184)


pavelrappo:
This comment describes changes to approximately 250 files in the generated JDK API Documentation which are mentioned in this PR description. Since the publication of this PR, I've been able to shave off a few changes that I previously categorized as non-benign.

As a sidenote, while analysing the changes to the JDK API Documentation, I once again saw how brittle and sometimes non-intuitive the automatic documentation inheritance algorithm is. No mater the search order that the algorithm imposes, some results are surprising. The published documentation for the current JDK likely contains many unnoticed bugs of the type "Gee, why does it inherit this text instead of that?". I've seen a few of those, but won't bring them into this discussion not to muddy the waters further. Hopefully, soon we'll have directed documentation inheritance, which will enable fixing some of those immediately and the rest of those in principle.

Back to the PR. Here are the changed files:

 * https://cr.openjdk.org/~prappo/8304135/before/api/
 * https://cr.openjdk.org/~prappo/8304135/after/api/

The majority of those changes can be described as an improvement to the following consistency aspect: a method should appear either in the "Method Summary" and "Method Details" sections, or in one or more "Methods declared in ..." lists. This aspect corresponds to the fact that a method can either be declared or inherited, but never both. Declared methods are summarised and detailed, while inherited methods are merely mentioned.

If a method is mentioned in N > 1 "Methods declared in ..." lists, these are actually N different methods. The Java Language Specification allows a class or interface to inherit multiple abstract methods whose signatures are override-equivalent. In JDK API it rarely happens: I can see fewer than 10 cases like that. One case is AbstractList which inherits two abstract `size` methods: [one] from AbstractCollection (which in turn inherits it from Collection) and the [other] from List.

Here are representative examples for some kinds of changes in those files:

1. DataOutputStream.write(byte[] b): a previously unrepresentable simple override via inheritance. It now only appears in "Methods declared in class java.io.FilterOutputStream". In particular, it does not appear in "Methods declared in interface java.io.DataOutput" or "Method Details".
2. PushbackReader.close overrides FilterReader.close, which is a non-simple override of Reader.close because Reader is abstract and FilterReader isn't. That was a bug.
3. "Specified by" and "Overrides" related to generic methods in java.lang.Class, java.time.chrono.ChronoLocalDateTime and java.time.chrono.ChronoZonedDateTime now correctly refer to wildcard `?` instead of the generic parameter.
4.  StringBuffer's chars() and codePoints() are no longer mentioned in "Methods declared in interface java.lang.CharSequence" as they are really taken from AbstractStringBuilder, an "undocumented enclosure", where they override corresponding methods declared in CharSequence.
   That mechanics also applies to java.time.chrono.HijrahDate's methods `toString`, `until` and their counterparts in java.time.chrono.ChronoLocalDateImpl.
5. Inline links to `#length()` in StringBuffer now correctly refer to the local method, not the `CharSequence.html#length()` supermethod.
6. StringBuffer.toString now correctly mentions that it overrides Object.toString. Previously it didn't mention any override because its immediate superclass, AbstractStringBuilder, is an "undocumented enclosure".
7. Finally, there are changes to class-use and index files, mostly related to the above changes in documentation.

[one]: https://cr.openjdk.org/~prappo/8304135/after/api/java.base/java/util/AbstractList.html#methods-inherited-from-class-java.util.Collection
[other]: https://cr.openjdk.org/~prappo/8304135/after/api/java.base/java/util/AbstractList.html#methods-inherited-from-class-java.util.List

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/TagletWriterImpl.java
```diff
@@ -612,6 +607,53 @@ private Content linkSeeReferenceOutput(Element holder,
...
+     * empty optional if that method is not documented anywhere and thus cannot
+     * be hyperlinked to.
+     *
+     * Why is this needed? Depending on configuration, a declared method does
+     * not always result in a corresponding "Method Details" entry, which we
+     * could link to. To not produce broken links (i.e. links that lead
+     * nowhere), we need to find the method that is a substitute for the
+     * given method.
+     */
+    public Optional<ExecutableElement> getActualMethod(ExecutableElement method) {
```
jonathan-gibbons, line 622: This doesn't feel like it belongs on `TagletWriterImpl`

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/VisibleMemberTable.java
```diff
@@ -775,8 +1362,15 @@ private boolean overridingSignatureChanged(ExecutableElement method, ExecutableE
         if (!method.getThrownTypes().equals(overriddenMethod.getThrownTypes())) {
             return true;
         }
-        // Documented annotations added anywhere in the method signature
-        return !getDocumentedAnnotations(method).equals(getDocumentedAnnotations(overriddenMethod));
+        // Documented annotations, other than java.lang.Override, added anywhere in the method signature
+        var JAVA_LANG_OVERRIDE = elements().getTypeElement("java.lang.Override");
```
jonathan-gibbons, line 1366: Interesting observation...  (not wrong)

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/VisibleMemberTable.java
```diff
@@ -286,45 +291,769 @@ public List<Element> getMembers(Kind kind) {
...
+                }
+            result = findOverriddenBy(m, s, result, foundOverriddenSoFar);
+        }
+        return result;
+    }
+
+    /*
+     * Translates the provided class or interface to a type given its subtype.
+     *
+     * The correctness relies of no more than 1 parameterization from JLS. (TODO link the correct section)
```
jonathan-gibbons, line 1026: typo? of -> on ?

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/VisibleMemberTable.java
```diff
@@ -286,45 +291,769 @@ public List<Element> getMembers(Kind kind) {
...
+        }
+        // Collect the super-interfaces of the supertype.
+        if (superType != null)
+            addSuperInterfaces(superType, results, visited);
+    }
+
+
+
+    private com.sun.tools.javac.util.List<OverrideData> findOverriddenBy(ExecutableElement m) {
+        return findOverriddenBy(m, (DeclaredType) te.asType(), com.sun.tools.javac.util.List.nil(), new HashSet<>());
```
jonathan-gibbons, line 1001: What is pushing you into having to use `javac`'s internal `List` class?

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/HtmlDocletWriter.java
```diff
@@ -290,30 +289,44 @@ public String replaceDocRootDir(String htmlstr) {
      * @param dl the content to which the method information will be added
      */
     private void addMethodInfo(ExecutableElement method, Content dl) {
```
jonathan-gibbons, line 291: It's not anything you've changed, and not part of the problem you're fixing, but this doesn't feel like it belongs on `HtmlDocletWriter`.  But that's another story of different architectural Technical Debt.

jonathan-gibbons:
It's clear that a lot of work has gone into this (well done) and that in pursuing the original seemingly-simple goal of directed inheritDoc, you have uncovered and at least partially addressed a **lot** of Technical Debt in this area.

It's also clear that you are battling less-than-ideal support in other APIs, like `javax.lang.model`, which expose some of the "quirks" in the way that `javac` models JLS.  

Various comments come to mind:

1. It's clear that it is good to have (and start with) an accurate model of the API _as specified by JLS_, especially as it relates to all the terms like _inheritance_, _overriding_, _hiding_ etc.  I also not the complications caused by override-equivalent signatures.

2. It's also clear that a JLS-only model is insufficient, and that we need to augment the model as needed with additional javadoc information and requirements. For example,
    * the existence of "undocumented enclosures" (to use your term) such that members may be treated as being "declared" in the first visible documented enclosure(s).
    * the presence of multiple comments that may all be relevant to a single method signature, and the difficulty of choosing the "best" one
    * the concept of _simple overrides_ such that even if a method is declared in a type element, we choose to "ignore" the declaration if it appears to be just an implementation-specific override.

3. While directed inheritDoc will be a valuable tool to assist authors to explicitly select the super type from which to inherit documentation, it does not help in determining the presence (or absence) of methods in the "summary lists" below the main "summary table", which is one of the areas where you have uncovered unexpected Technical Debt.

4. While I believe the work here will lead to more accurate documentation, I think it is too late to go in this release. I appreciate the work you have done to compare docs before and after this change, and for the analysis you describe in your [earlier comment](https://github.com/openjdk/jdk/pull/14221#issuecomment-1573670298) but I suggest the conservative approach would to push this early in the next release, when we have more time to address any unexpected issues that may arise.

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/VisibleMemberTable.java
```diff
@@ -286,45 +291,769 @@ public List<Element> getMembers(Kind kind) {
...
+                                                                         com.sun.tools.javac.util.List<OverrideData> result,
+                                                                         Set<ExecutableElement> foundOverriddenSoFar) {
+        for (var s : directSupertypes(declaredType)) {
+            var supertypeElement = (TypeElement) s.asElement();
+            for (var m1 : mcache.getVisibleMemberTable(supertypeElement).getMethodMembers())
+                if (utils.elementUtils.overrides(m, m1, te) && foundOverriddenSoFar.add(m1)) {
+                    // use any value for `simpleOverride` for now: the actual value
+                    // will be computed once the complete sequence is available
+                    var simpleOverride = false;
+                    result = result.append(new OverrideData(toDeclaringType(s,
```
jonathan-gibbons, line 1015: `append` is *horrible* on `javac.List`

Unless there is a string reason to be using it, you should probably be using plain old collection Lists, like `ArrayList` or `List.of`. 

If you're using `javac.List` because you're calling javac internal functionality that requires that kind of List, then your usage should be in methods in `WorkArounds`, which is intended to the the only doclet class that accesses javac internal methods and types.

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/TagletWriterImpl.java
pavelrappo: I'm okay with making it private, but this functionality is only currently needed in this class to replace code 50-ish lines above.

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/VisibleMemberTable.java
pavelrappo: IIRC, it's a combination of vestigial reasons and the fact that there's the `reverse()` method, whose cousin SequencedCollection.reversed() is sadly unavailable in langtools in JDK 21 because of "JDK N-1".

I'll try to get rid of javac's list here.

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/VisibleMemberTable.java
pavelrappo: Replied here https://github.com/openjdk/jdk/pull/14221#discussion_r1224059741

pavelrappo:
One thing that bothers me after extensive analysis of JDK API documentation differences caused by this PR is loss of information. Before this PR, JavaDoc documented more methods on a class or interface page than there were member methods in that class or interface.

For example, consider `AbstractCollection`. It does not declare `equals`, but it extends `Object` and implements `Collection`, each of which declare and document `equals()`. According to Java Language Specification (JLS), `AbtractCollection` inherits `Object.equals`, but doesn't inherit `Collections.equals`. So, as far as method members go, there's no place for `Collection.equals` among members of `AbstractCollection`. Yet, VMT for AbstractCollection stored both of them. 

In the new VMT there's no place for non-members. That observation alone suggests that JavaDoc view is not the same as JLS view. While JavaDoc shouldn't contradict JLS or mislead the reader, it should also help the reader to discover and access relevant information beyond that immediately available from the language model. Before we figure out how to do that, **I'm parking this PR**.

I note that all the tests are green: https://cr.openjdk.org/~prappo/8304135/tests.txt

PR #14225, opened 2023-05-30
8308031: Linkers should reject unpromoted variadic parameters

JornVernee:
In C, arguments smaller than `int` are promoted to (`unsigned`) `int`, and `float` is promoted to `double`, when being passed as variadic argument (see e.g. https://en.cppreference.com/w/c/language/conversion#Default_argument_promotions). This patch adds the automatic argument promotion for variadic arguments, to align the implementation with the C specification.

The fallback linker is also updated to use to correct function to link variadic calls (not doing this turned out not to be a problem so far, but it is problematic for instance on Mac/AArch64 when using the fallback linker). Adding the restriction on layouts for all linkers is also partly motivated by the fallback linker rejecting such unsupported variadic layouts already.

I've added a small paragraph to the Linker javadoc as well that explains this.

TestVarArgs needed to be updated in order to account for the difference in the promoted `float` values.

Testing:
- local testing on Windows/x64
- local testing on Linux/x64 using the fallback linker
- tier1-3 + jdk-tier5
- manual test run on mac/aarch64 with the fallback linker to verify the correctness of the fallback linker changes.

---------
### Progress
- [x] Change requires CSR request [JDK-8309205](https://bugs.openjdk.org/browse/JDK-8309205) to be approved

### Issues
 * [JDK-8308031](https://bugs.openjdk.org/browse/JDK-8308031): Linkers should reject unpromoted variadic parameters (**Enhancement** - `"4"`)
 * [JDK-8309205](https://bugs.openjdk.org/browse/JDK-8309205): Linkers should reject unpromoted variadic parameters (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14225/head:pull/14225` \
`$ git checkout pull/14225`

`$ git checkout pull/14225` \
`$ git pull https://git.openjdk.org/jdk.git pull/14225/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14225`

`$ git pr show -t 14225`

</details>

<a href="https://git.openjdk.org/jdk/pull/14225.diff">https://git.openjdk.org/jdk/pull/14225.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14225#issuecomment-1570314494)


test/jdk/java/foreign/StdLibTest.java
```diff
@@ -383,7 +383,7 @@ enum PrintfArg {
         STRING(MemorySegment.class, C_POINTER, "%s", arena -> {
             return arena.allocateUtf8String("str");
         }, "str"),
-        CHAR(byte.class, C_CHAR, "%c", arena -> (byte) 'h', 'h'),
+        CHAR(int.class, C_INT, "%c", arena -> (int) 'h', (int) 'h'), // promote to int, per C spec
```
mcimadamore, line 386: is it important to retain this, given that there's already an INT case?

mcimadamore (approved):
Looks good

test/jdk/java/foreign/StdLibTest.java
JornVernee: Yes, good idea.

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -409,6 +409,12 @@
  * }
  * }
  *
+ * <p>
+ * It should be noted that values passed as variadic arguments undergo default argument promotion in C. Each value of
+ * type {@code float} is converted to {@code double}, and each integral type smaller than {@code int} is converted to
+ * {@code int}. As such, the native linker will reject attempts to link a function with variadic parameters which have
```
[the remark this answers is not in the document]
JornVernee: I did another pass over this. I went back to the first paragraph and defined the term "variadic argument layout" which I then use to explain which layouts can not be used as variadic argument layouts.

I've also moved the note the be directly after the first paragraph so that it appears closed to where the term is defined.

Hopefully there should be no gaps now.

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -366,10 +366,18 @@
  * <h3 id="variadic-funcs">Variadic functions</h3>
  *
  * Variadic functions (e.g. a C function declared with a trailing ellipses {@code ...} at the end of the formal parameter
```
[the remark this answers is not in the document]
mcimadamore: Also, should we say somewhere that, for prototype-less functions, `firstVariadicArg` should always have an index of 0 (e.g. any other value is illegal, trivially). It's not super important, just one of those little things that can reinforce understanding.

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -366,10 +366,18 @@
...
+ * directly by the native linker. However, it is still possible to link a variadic function by using a <em>specialized</em>
+ * function descriptor, together with {@linkplain Linker.Option#firstVariadicArg(int) a linker option} which indicates the
+ * position of the first variadic argument in that specialized descriptor. The corresponding argument layout, and all following
+ * argument layouts in the descriptor, are considered to be <em>variadic argument layouts</em>. These layouts correspond
+ * to the arguments passed in place of the ellipsis {@code ...}, or to the arguments passed to a prototype-less function.
+ * <p>
+ * It should be noted that values passed as variadic arguments undergo default argument promotion in C. Each value of
+ * type {@code float} is converted to {@code double}, and each integral type smaller than {@code int} is converted to
+ * {@code int}. As such, the native linker will reject attempts to link function descriptors with certain variadic argument
+ * layouts. Namely, {@linkplain ValueLayout value layouts} that have a carrier type of {@code boolean}, {@code byte},
```
mcimadamore, line 379: Is there any reason as to why you decided to say which layouts are **not** allowed as variadic layouts? I'm wondering whether, if we add more carriers in the future, this list will probably need to be updated? Or do the restriction only involve these "small" types, and stuff like `long double` is always allowed?

src/java.base/share/classes/java/lang/foreign/Linker.java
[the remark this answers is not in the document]
JornVernee: > Perhaps if we used a bullet-list to define the two ways in which variadic function can be declared, the definition could stand out even more?

Yes, I think this is a good idea. I'll do another pass.

> Also, should we say somewhere that, for prototype-less functions, firstVariadicArg should always have an index of 0

I think this is also a good idea. It help hammer the point home.

src/java.base/share/classes/java/lang/foreign/Linker.java
JornVernee: No real reason for the current polarity. The specification explicitly calls out the float -> double conversion, and then for the integral types it refers to 'integer promotions' [1], which is a somewhat complex rule set for assigning ranks to integer types.

So, I think if we add more carriers in the future, we would have to worry about integral types that have a rank less than (unsigned) int when considering this list. This is currently only the types I've listed (though, `char` is a bit of a special case), and this seems relatively stable to me. e.g. `long double` would not need to be added here.

[1]: [Integer_promotions](https://en.cppreference.com/w/c/language/conversion#Integer_promotions)

src/java.base/share/classes/java/lang/foreign/Linker.java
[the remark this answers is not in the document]
JornVernee: Did another pass: https://github.com/openjdk/jdk/pull/14225/commits/e806fae5dc295726846d197dad59e6b0ba715ceb

There are now 3 paragraphs in the javadoc: One that explains what a variadic function is (in C), one  that explains how they are linked using specialized function descriptors, and finally one which explains the limitation on variadic argument layouts.

Additionally I've switched the polarity of the check to be the same as it is in the doc, and aligned the exception message with the terminology in the doc.

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -365,11 +365,28 @@
...
- * list or with an empty formal parameter list) are not supported directly by the native linker. However, it is still possible
- * to link a variadic function by using a <em>specialized</em> function descriptor, together with a
- * {@linkplain Linker.Option#firstVariadicArg(int) a linker option} which indicates the position of the first variadic argument
- * in that specialized descriptor.
+ * Variadic functions are C functions which can accept a variable number and type of arguments. They are declared:
+ * <ol>
+ * <li>With a trailing ellipsis ({@code ...}) at the end of the formal parameter list, such as: {@code void foo(int x, ...);}</li>
+ * <li>With an empty formal parameter list, such as: {@code void foo();}</li>
+ * </ol>
+ * The latter is often called a <em>prototype-less</em> function as well. The arguments passed in place of the ellipsis,
```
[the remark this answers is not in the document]
JornVernee: Yes, will do.

JornVernee:
Thanks for the review! I've updated the CSR with the finalized version of the javadoc. Could you take a look at that as well? https://bugs.openjdk.org/browse/JDK-8309205

JornVernee:
After some offline discussion with Maurizio, we concluded that the linker should automatically apply the needed argument promotions, rather than reject the 'incorrect' layouts. This is more convenient to the user because they don't have to figure out how to do the promotions, but also because they don't have to figure out the platform dependent promotion process.

src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
```diff
@@ -260,4 +245,90 @@ private void checkByteOrder(ValueLayout vl) {
...
+
+    // For now all of these are the same for every implementation
+    // in the future we might need to make one of them abstract
+    // and implement on a per-linker basis.
+    private static boolean requiresVariadicIntPromotion(ValueLayout vl) {
+        return vl.carrier() == boolean.class || vl.carrier() == byte.class
+                    || vl.carrier() == char.class || vl.carrier() == short.class;
+    }
+
+    private static boolean requiresVariadicFloatPromotion(ValueLayout vl) {
```
mcimadamore, line 327: Should this be called `requiresVariadicDoublePromotion` ? (In the other method we say `int`, which is the target of the conversion).

src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
```diff
@@ -260,4 +245,90 @@ private void checkByteOrder(ValueLayout vl) {
...
+                return fva.index();
+            }
+        }
+        return -1;
+    }
+
+    // C spec mandates that variadic arguments smaller than int are promoted to int,
+    // and float is promoted to double
+    // See: https://en.cppreference.com/w/c/language/conversion#Default_argument_promotions
+    private static FunctionDescriptor promoteVariadicArgs(FunctionDescriptor function, int firstVariadicArgIndex) {
```
mcimadamore, line 261: Maybe it's unneeded optimization, but we're scanning the list of layouts once, and then once again to determine the resulting method type. Should we try to do both in one pass?

mcimadamore (approved):
Looks good. I left minor stylistic comments.

src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
JornVernee: Yes, I can do that. I was going for 'int' meaning 'integral' and 'float' meaning floating point. But I think your suggestion is clearer.

src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
JornVernee: I considered keeping track of the promoted args in a boolean[], but I figured it was simpler to just scan the layouts again. And this is not really performance critical code.

mcimadamore (approved):

JornVernee:
After more offline discussion we concluded that by applying argument promotions automatically we create an issue for unsigned types. We currently allow an unsigned value to be represented with the signed equivalent. For example, `JAVA_SHORT` can represent both the C `short` and `unsigned short` types. The issue is that if we were to promote `short` values to `int`, it is ambiguous whether sign extension should take place.

As a result, we decided to go back to the approach of rejecting variadic layouts representing a C type that would undergo default argument promotion. If/when we add better support for unsigned types, we can revisit.

mcimadamore (approved):

PR #8441, opened 2022-04-28
JDK-8255439: System Tray icons get corrupted when windows scaling changes

honkar-jdk:
In Windows, when desktop scaling is changed the tray icons was distorted/blurred a bit each time scaling changes.

With the proposed fix, the tray icon scales according to on-the-fly DPI scale settings. A test case has been added which adds a MRI icon to system tray, to observe the icon scaling when DPI is changed. Since the scale cannot be programmatically changed (for dynamic on-the-fly scale changes), I have used a manual test case to test this scenario.

When DPI changes usually two messages are sent by windows -

- [WM_DPICHANGED](https://docs.microsoft.com/en-us/windows/win32/hidpi/wm-dpichanged)
- [WMPOSCHANGING](https://docs.microsoft.com/en-us/windows/win32/winmsg/wm-windowposchanging)

I'm triggering an update on tray icons on receiving WMPOSCHANGING msg through the Tray icon's Window Procedure. Triggering an update on WM_DPICHANGED was still causing the icons to be distorted, hence WMPOSCHANGING is being used as the message to trigger the update.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed (1 review required, with at least 1 reviewer)

### Issue
 * [JDK-8255439](https://bugs.openjdk.java.net/browse/JDK-8255439): System Tray icons get corrupted when windows scaling changes

### Reviewers
 * [Kevin Rushforth](https://openjdk.java.net/census#kcr) (@kevinrushforth - Author)
 * [Phil Race](https://openjdk.java.net/census#prr) (@prrace - **Reviewer**)
 * [Alexey Ivanov](https://openjdk.java.net/census#aivanov) (@aivanov-jdk - **Reviewer**)

### Contributors
 * Alexey Ivanov `<aivanov@openjdk.org>`

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk pull/8441/head:pull/8441` \
`$ git checkout pull/8441`

Update a local copy of the PR: \
`$ git checkout pull/8441` \
`$ git pull https://git.openjdk.java.net/jdk pull/8441/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 8441`

View PR using the GUI difftool: \
`$ git pr show -t 8441`

</details>
<details><summary>Using diff file</summary>

[...]

PR #14226, opened 2023-05-30
8296972: [macos13] java/awt/Frame/MaximizedToIconified/MaximizedToIconified.java: getExtendedState() != 6 as expected.

alisenchung:
added displayChanged call to CPlatformWindow when frame first needs to deiconify or unmaximize
All client tests passed after change

---------
### Progress

### Issue
 * [JDK-8296972](https://bugs.openjdk.org/browse/JDK-8296972): [macos13] java/awt/Frame/MaximizedToIconified/MaximizedToIconified.java: getExtendedState() != 6 as expected. (**Bug** - P4)

### Reviewers
 * [Dmitry Markov](https://openjdk.org/census#dmarkov) (@dmarkov20 - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14226/head:pull/14226` \
`$ git checkout pull/14226`

`$ git checkout pull/14226` \
`$ git pull https://git.openjdk.org/jdk.git pull/14226/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14226`

`$ git pr show -t 14226`

</details>

<a href="https://git.openjdk.org/jdk/pull/14226.diff">https://git.openjdk.org/jdk/pull/14226.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14226#issuecomment-1568912817)

### DIFF
--- a/src/java.desktop/macosx/classes/sun/lwawt/macosx/CPlatformWindow.java
+++ b/src/java.desktop/macosx/classes/sun/lwawt/macosx/CPlatformWindow.java
@@ -986,7 +986,7 @@ public class CPlatformWindow extends CFRetainedResource implements PlatformWindo
                     // let's return into the normal states first
                     // the zoom call toggles between the normal and the max states
                     unmaximize();
-                    peer.displayChanged();
+                    peer.getLWToolkit().realSync();
                 }
                 execute(CWrapper.NSWindow::miniaturize);
                 break;
@@ -994,7 +994,8 @@ public class CPlatformWindow extends CFRetainedResource implements PlatformWindo
                 if (prevWindowState == Frame.ICONIFIED) {
                     // let's return into the normal states first
                     execute(CWrapper.NSWindow::deminiaturize);
-                    peer.displayChanged();
+                    peer.getLWToolkit().realSync();
+
                 }
                 maximize();
                 break;
--- a/src/java.desktop/macosx/classes/sun/lwawt/macosx/CPlatformWindow.java
+++ b/src/java.desktop/macosx/classes/sun/lwawt/macosx/CPlatformWindow.java
@@ -967,7 +967,7 @@ public class CPlatformWindow extends CFRetainedResource implements PlatformWindo
     }
 
     private void waitForWindowState(int state) {
-        if(peer.getState() == state) {
+        if (peer.getState() == state) {
             return;
         }
 
@@ -990,7 +990,6 @@ public class CPlatformWindow extends CFRetainedResource implements PlatformWindo
                 } catch (InterruptedException ie) {}
             }
         }
-
         target.removeWindowStateListener(wsl);
     }
 
--- a/src/java.desktop/macosx/classes/sun/lwawt/macosx/CPlatformWindow.java
+++ b/src/java.desktop/macosx/classes/sun/lwawt/macosx/CPlatformWindow.java
@@ -42,6 +42,7 @@ import java.awt.Toolkit;
 import java.awt.Window;
 import java.awt.event.FocusEvent;
 import java.awt.event.WindowEvent;
+import java.awt.event.WindowStateListener;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 import java.lang.reflect.InvocationTargetException;
@@ -965,6 +966,26 @@ public class CPlatformWindow extends CFRetainedResource implements PlatformWindo
         return isFullScreenMode;
     }
 
+    private void waitForWindowState(int state) {
+        Object lock = new Object();
+        target.addWindowStateListener(new WindowStateListener() {
+            public void windowStateChanged(WindowEvent e) {
+                synchronized (lock) {
+                    if (e.getNewState() == state) {
+                        lock.notifyAll();
+                    }
+                }
+            }
+        });
+        if (peer.getState() != state) {
+            synchronized (lock) {
+                try {
+                    lock.wait();
+                } catch (InterruptedException ie) {}
+            }
+        }
+    }
+
     @Override
     public void setWindowState(int windowState) {
         if (peer == null || !peer.isVisible()) {
@@ -986,7 +1007,7 @@ public class CPlatformWindow extends CFRetainedResource implements PlatformWindo
                     // let's return into the normal states first
                     // the zoom call toggles between the normal and the max states
                     unmaximize();
-                    peer.getLWToolkit().realSync();
+                    waitForWindowState(Frame.NORMAL);
                 }
                 execute(CWrapper.NSWindow::miniaturize);
                 break;
@@ -994,7 +1015,7 @@ public class CPlatformWindow extends CFRetainedResource implements PlatformWindo
                 if (prevWindowState == Frame.ICONIFIED) {
                     // let's return into the normal states first
                     execute(CWrapper.NSWindow::deminiaturize);
-                    peer.getLWToolkit().realSync();
+                    waitForWindowState(Frame.NORMAL);
 
                 }
                 maximize();
--- a/src/java.desktop/macosx/classes/sun/lwawt/macosx/CPlatformWindow.java
+++ b/src/java.desktop/macosx/classes/sun/lwawt/macosx/CPlatformWindow.java
@@ -986,6 +986,7 @@ public class CPlatformWindow extends CFRetainedResource implements PlatformWindo
                     // let's return into the normal states first
                     // the zoom call toggles between the normal and the max states
                     unmaximize();
+                    peer.displayChanged();
                 }
                 execute(CWrapper.NSWindow::miniaturize);
                 break;
@@ -993,6 +994,7 @@ public class CPlatformWindow extends CFRetainedResource implements PlatformWindo
                 if (prevWindowState == Frame.ICONIFIED) {
                     // let's return into the normal states first
                     execute(CWrapper.NSWindow::deminiaturize);
+                    peer.displayChanged();
                 }
                 maximize();
                 break;
--- a/src/java.desktop/macosx/classes/sun/lwawt/macosx/CPlatformWindow.java
+++ b/src/java.desktop/macosx/classes/sun/lwawt/macosx/CPlatformWindow.java
@@ -967,8 +967,12 @@ public class CPlatformWindow extends CFRetainedResource implements PlatformWindo
     }
 
     private void waitForWindowState(int state) {
+        if(peer.getState() == state) {
+            return;
+        }
+
         Object lock = new Object();
-        target.addWindowStateListener(new WindowStateListener() {
+        WindowStateListener wsl = new WindowStateListener() {
             public void windowStateChanged(WindowEvent e) {
                 synchronized (lock) {
                     if (e.getNewState() == state) {
@@ -976,7 +980,9 @@ public class CPlatformWindow extends CFRetainedResource implements PlatformWindo
                     }
                 }
             }
-        });
+        };
+
+        target.addWindowStateListener(wsl);
         if (peer.getState() != state) {
             synchronized (lock) {
                 try {
@@ -984,6 +990,8 @@ public class CPlatformWindow extends CFRetainedResource implements PlatformWindo
                 } catch (InterruptedException ie) {}
             }
         }
+
+        target.removeWindowStateListener(wsl);
     }
 
     @Override
--- a/test/jdk/java/awt/Frame/MaximizedToIconified/MaximizedToIconified.java
+++ b/test/jdk/java/awt/Frame/MaximizedToIconified/MaximizedToIconified.java
@@ -45,7 +45,7 @@ import java.awt.event.WindowStateListener;
 
 public class MaximizedToIconified
 {
-    static volatile int lastFrameState = Frame.NORMAL;
+    static volatile int lastFrameState;
     static volatile boolean failed = false;
     static volatile Toolkit myKit;
     private static Robot robot;
@@ -76,6 +76,8 @@ public class MaximizedToIconified
         Frame frame = new Frame("test");
         frame.setSize(200, 200);
         frame.setVisible(true);
+
+        lastFrameState = Frame.NORMAL;
 
         robot.waitForIdle();
 
@@ -114,6 +116,7 @@ public class MaximizedToIconified
         //    because Toolkit.isFrameStateSupported() method reports these states
         //    as not supported. And such states will simply be skipped.
         examineStates(new int[] {Frame.MAXIMIZED_BOTH, Frame.ICONIFIED, Frame.NORMAL});
+        System.out.println("------");
         examineStates(new int[] {Frame.ICONIFIED, Frame.MAXIMIZED_BOTH, Frame.NORMAL});
 
     }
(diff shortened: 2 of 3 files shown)


src/java.desktop/macosx/classes/sun/lwawt/macosx/CPlatformWindow.java
```diff
@@ -986,13 +986,15 @@ public void setWindowState(int windowState) {
                     // let's return into the normal states first
                     // the zoom call toggles between the normal and the max states
                     unmaximize();
+                    peer.displayChanged();
```
dmarkov20, line 989: Why do you call `peer.displayChanged()` only when the previous windows state is `MAXIMIZED_BOTH`? What happens if the previous windows state is `NORMAL`?

src/java.desktop/macosx/classes/sun/lwawt/macosx/CPlatformWindow.java
```diff
@@ -986,13 +986,15 @@ public void setWindowState(int windowState) {
...
                 }
                 execute(CWrapper.NSWindow::miniaturize);
                 break;
             case Frame.MAXIMIZED_BOTH:
                 if (prevWindowState == Frame.ICONIFIED) {
                     // let's return into the normal states first
                     execute(CWrapper.NSWindow::deminiaturize);
+                    peer.displayChanged();
```
dmarkov20, line 997: Same as above. What happens if we move from `NORMAL` to `MAXIMIZED_BOTH`?
 
Shall we call `peer.displayChanged()` when a frame is moved to `NORMAL` state? I know that cases are not covered by the test but it seems the issue takes place there. Can you investigate, please?

dmarkov20 (changes requested):

mrserb:
Why this method is not called automatically by our native code? what events are changed since macOS 13.0?

mrserb:
BTW it is unclear what the call to "displayChanged" will fix. Will it set the correct "graphicsDevice", if yes then what gD was used before the call, if not then what property is updated by the "displayChanged"?

src/java.desktop/macosx/classes/sun/lwawt/macosx/CPlatformWindow.java
alisenchung: I believe the main issue is that there is a native event that changes the window state, but when two state changes are done back to back sometimes that native event isn't delivered quickly enough before the second state change occurs, which causes this test to fail. When a frame is moved to a normal state, there are no back to back state changes, since all we do is change the state back to normal.

alisenchung: If I set a delay between the state changes, the problem will also disappear and the test passes.

alisenchung:
> BTW it is unclear what the call to "displayChanged" will fix. Will it set the correct "graphicsDevice", if yes then what gD was used before the call, if not then what property is updated by the "displayChanged"?

I think the problem is that the first state change (uniconify) causes a native event that changes the frame's state, but that event isn't delivered until after the second state change (maximize) occurs, which is causing problems on the final state of the frame. The displayChanged call between these state changes was intended to flush the native event queue and cause the event to be delivered between the state changes.

honkar-jdk:
> Why this method is not called automatically by our native code? what events are changed since macOS 13.0?

@mrserb There might be a new window state in macOS 13 due to a the new [Stage Manager](https://support.apple.com/en-us/HT213315) feature but I believe this shouldn't affect any existing states.

alisenchung:
> If I got it right the switching to `ICONIFIED`/`MAXIMIZED_BOTH` state takes more time on macOS 13 than on previous macOS version. So the test failure is triggered by the lack of synchronisation. When you added displayChanged() call you introduced some kind of delay and the problem stopped happening but it is still present. I do not think any changes are required in JDK but the test definitely needs some modifications. In particular I believe the test must check the state of the frame only when the corresponding `window state changed` event is received.

I did some testing and I think that without the delay, when the frame moves from iconified to maximized states, the second windowStateChanged event never comes (first being iconified -> normal, second being normal -> maximized), so the test will not pass even if we only check state of the frame on a windowStateChanged event. 

Why do you think no changes are required in the JDK? Shouldn't there be some synchronization needed when calling the native deminiaturize and maximize functions in CPlatformWindow?

mrserb:
Possibly some synchronization is needed in the jdk, it needs to be checked why the message is not posted. But it should not be fixed by the "displayChanged" used as a "delay"

honkar-jdk:
@alisenchung @mrserb 

After reading through the discussion, I think the fix on native side rather than on Java might work better here. I was revisiting a Tray Icon issue which had similar sync problem when I was trying to get updated scale value on Java side by using a DisplayChangedListener [Details here - [PR#8441](https://github.com/openjdk/jdk/pull/8441#issuecomment-1121345834)] and there were multiple events occurring back to back. When I had a hard-coded delay set on Java side, the update to tray icon worked fine. I think this is a similar case where introducing a delay solves the issue. But since hard-coded delay is not the ideal solution probably triggering/ calling the required piece of code after everything is updated on native side is the way to go (event/message-driven).

Handling the code on Window's native code with the help of windows messaging event solved the Trayicon problem.

honkar-jdk:
Also noticed this comment in setWindowState() of [CPlatformWindow.java#L1011 ](https://github.com/openjdk/jdk/blob/ea41907396a70933c4f6e6b0f247f7ba589a3a23/src/java.desktop/macosx/classes/sun/lwawt/macosx/CPlatformWindow.java#L1011)

_// NOTE: the SWP.windowState field gets updated to the newWindowState
//       value when the native notification comes to us_

May be waiting untill the required native notification using an event-driven wait might work here?

honkar-jdk:
@alisenchung Before digging deeper on Cocoa Event Handling, please check if fixing on native side by waiting for an approriate NSEvent is a viable solution in this case. 

A gist of Cocoa Event Handling mechanism - 

1. NSRunLoop in the event loop [NSApplicationAWT.m#L303](https://github.com/openjdk/jdk/blob/ea41907396a70933c4f6e6b0f247f7ba589a3a23/src/java.desktop/macosx/native/libosxapp/NSApplicationAWT.m#L303) handled by NSApp. 
2. NSApp converts this event into NSEvent object and dispatches using sendEvent to the appropriate NSView object.

src/java.desktop/macosx/classes/sun/lwawt/macosx/CPlatformWindow.java
```diff
@@ -986,13 +986,16 @@ public void setWindowState(int windowState) {
                     // let's return into the normal states first
                     // the zoom call toggles between the normal and the max states
                     unmaximize();
+                    peer.getLWToolkit().realSync();
```
dmarkov20, line 989: Well.. `realSync()` works here but I am not sure if that invocation really fixes the issue or just hides it by introducing a delay.
I believe we need some function which ensures that the frame is in correct (`NORMAL`) state right after `unmaximize()` call and before `execute(CWrapper.NSWindow::miniaturize)` invocation.

Possible implementation of that method below:

```
private void waitForWindowState(int state) {
        Object lock = new Object();
        target.addWindowStateListener(new WindowStateListener() {
            public void windowStateChanged(WindowEvent e) {
                synchronized(lock) {
                    if (e.getNewState() == state) {
                        lock.notifyAll();
                    }
                }
            }
        });
        if (peer.getState() != state) {
            synchronized(lock) {
                try {
                    lock.wait();
                } catch(InterruptedException ie) {}
            }
        }
    }
```

So in code you will call `waitForWindowState(Frame.NORMAL)` instead of `realSync()`

dmarkov20 (changes requested):

honkar-jdk:
@alisenchung 

The current fix works for the cases listed in the test but fails for the following scenarios. To make the test more stable it is better to reassign the static variable `lastFrameState = Frame.NORMAL` in examineStates after creating a new Frame [here ](https://github.com/openjdk/jdk/blob/57b8251241e2044d5039ce162bf4637a9b2e5466/test/jdk/java/awt/Frame/MaximizedToIconified/MaximizedToIconified.java#L78), so that each time a new case is run it sets lastFrameState to default state.

```
// CASE 1:
examineStates(new int[] {Frame.NORMAL, Frame.MAXIMIZED_BOTH, Frame.ICONIFIED});
// CASE 2:
examineStates(new int[] {Frame.NORMAL, Frame.ICONIFIED, Frame.MAXIMIZED_BOTH});
```

honkar-jdk (changes requested):
@alisenchung 

The current fix works for the cases listed in the test but fails for the following scenarios. To make the test more stable it is better to reassign the static variable `lastFrameState = Frame.NORMAL` in examineStates after creating a new Frame [here ](https://github.com/openjdk/jdk/blob/57b8251241e2044d5039ce162bf4637a9b2e5466/test/jdk/java/awt/Frame/MaximizedToIconified/MaximizedToIconified.java#L78), so that each time a new case is run it sets lastFrameState to default state.

```
// CASE 1:
examineStates(new int[] {Frame.NORMAL, Frame.MAXIMIZED_BOTH, Frame.ICONIFIED});
// CASE 2:
examineStates(new int[] {Frame.NORMAL, Frame.ICONIFIED, Frame.MAXIMIZED_BOTH});
```

prrace:
I don't like the idea of a delay in either of the ways discussed here.
And although I do think that its much better to wait for delivery of an event,  I'm not sure I understand the windowState listener proposal.
How do you know in this code that the transition of
iconified-> normal will then followed by normal->maximised ?
Perhaps you'll wait for ever ?
But I don't know what the macOS native model is for all of this.
This sequence is kind of odd and I'm curious what happens on older macOS releases - Sergey touched on this but I don't see direct followup.
There seems to be assumptions it was like this previously but now it is slower ?

I suggest explicit verification of what happens on macOS 12.6.
Are we now getting two native events when there used to be only one ?

alisenchung:
> I don't like the idea of a delay in either of the ways discussed here. And although I do think that its much better to wait for delivery of an event, I'm not sure I understand the windowState listener proposal. How do you know in this code that the transition of iconified-> normal will then followed by normal->maximised ? 

The code here is trying to do a transition from iconified -> maximized, but splits it up into iconified -> normal and normal -> maximized.

dmarkov20:
> I don't like the idea of a delay in either of the ways discussed here. And although I do think that its much better to wait for delivery of an event, I'm not sure I understand the windowState listener proposal. How do you know in this code that the transition of iconified-> normal will then followed by normal->maximised ? Perhaps you'll wait for ever ? But I don't know what the macOS native model is for all of this. This sequence is kind of odd and I'm curious what happens on older macOS releases - Sergey touched on this but I don't see direct followup. There seems to be assumptions it was like this previously but now it is slower ?
> 
> I suggest explicit verification of what happens on macOS 12.6. Are we now getting two native events when there used to be only one ?

We cannot move a frame directly from `ICONIFIED` to `MAXIMIZED_BOTH` and vice versa. We need to go through the `NORMAL` state, (i.e. `ICONIFIED` -> `NORMAL` -> `MAXIMIZED_BOTH`). If my memory does not fail me we use that approach from the very beginning.

We perform two state-transitions subsequently and don't care of real state of the frame between them. It used to work till macOS 13 where the transition from one state to another takes a bit longer. 
In other words we do ICONIFIED` -> `NORMAL` and then immediately `NORMAL` -> `MAXIMIZED_BOTH` but the frame is NOT in `NORMAL` state yet.

dmarkov20 (changes requested):

test/jdk/java/awt/Frame/MaximizedToIconified/MaximizedToIconified.java
```diff
@@ -45,7 +45,7 @@
 
 public class MaximizedToIconified
 {
-    static volatile int lastFrameState = Frame.NORMAL;
+    static volatile int lastFrameState;
```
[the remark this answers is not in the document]
honkar-jdk: @dmarkov20 Although the test does not explicitly test the cases mentioned below, it fails for the following two cases because the lastFrameState is changed as we progress through the states and should be reassigned when testing a new set of states. @alisenchung  is assigning this var within examineStates now.

> @alisenchung
> 
> The current fix works for the cases listed in the test but fails for the following scenarios. To make the test more stable it is better to reassign the static variable `lastFrameState = Frame.NORMAL` in examineStates after creating a new Frame [here ](https://github.com/openjdk/jdk/blob/57b8251241e2044d5039ce162bf4637a9b2e5466/test/jdk/java/awt/Frame/MaximizedToIconified/MaximizedToIconified.java#L78), so that each time a new case is run it sets lastFrameState to default state.
> 
> ```
> // CASE 1:
> examineStates(new int[] {Frame.NORMAL, Frame.MAXIMIZED_BOTH, Frame.ICONIFIED});
> // CASE 2:
> examineStates(new int[] {Frame.NORMAL, Frame.ICONIFIED, Frame.MAXIMIZED_BOTH});
> ```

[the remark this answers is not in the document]
dmarkov20: Ah, I see now. I would suggest including that cases into the test.

test/jdk/java/awt/Frame/MaximizedToIconified/MaximizedToIconified.java
```diff
@@ -114,6 +116,7 @@ private static void doTest() {
         //    because Toolkit.isFrameStateSupported() method reports these states
         //    as not supported. And such states will simply be skipped.
         examineStates(new int[] {Frame.MAXIMIZED_BOTH, Frame.ICONIFIED, Frame.NORMAL});
+        System.out.println("------");
         examineStates(new int[] {Frame.ICONIFIED, Frame.MAXIMIZED_BOTH, Frame.NORMAL});
```
honkar-jdk, line 121: @alisenchung Now that the `lastFrameState` is fixed, I would suggest adding the cases mentioned [here](https://github.com/openjdk/jdk/pull/14226/files#r1237259547) too, to make the test comprehensive.

honkar-jdk (commented):
The update fix - with the current state check before acquiring the lock and removing the WindowStateListener suggested by @dmarkov20 avoids unnecessary acquiring of lock  and looks to be a better solution that before.

honkar-jdk (approved):

dmarkov20 (commented):
@alisenchung The fix looks OK. Did you test it on macOS 12? If not, can you ensure that previous macOS version(s) works fine with the proposed changes?

alisenchung:
> @alisenchung The fix looks OK. Did you test it on macOS 12? If not, can you ensure that previous macOS version(s) works fine with the proposed changes?

I've tested on macOS12 and it works fine with the changes.

dmarkov20 (approved):

src/java.desktop/macosx/classes/sun/lwawt/macosx/CPlatformWindow.java
```diff
@@ -965,6 +966,33 @@ public boolean isFullScreenMode() {
...
+                    }
+                }
+            }
+        };
+
+        target.addWindowStateListener(wsl);
+        if (peer.getState() != state) {
+            synchronized (lock) {
+                try {
+                    lock.wait();
```
mrserb, line 989: Are we sure that waiting forever is a good thing? Can this request be ignored by the OS for any reason?

dmarkov20: It is very unlikely that we are waiting forever here unless there are some serious problems with corresponding platform API. 
I understand your concern and agree that waiting for some reasonable time instead of forever is more suitable here.
@alisenchung Can you open a bug to replace forever waiting with timeout.

PR #14231, opened 2023-05-30
8309136: [JVMCI] add -XX:+UseGraalJIT flag

dougxc:
Use of the Graal-based JIT in OpenJDK currently requires the following flag: `-XX:+EnableJVMCIProduct`

This has no direct association with Graal. If the JDK image happens to include a non-Graal JVMCI implementation, it will be automatically selected. This would come as a surprise to users who equate JVMCI with Graal.

This PR introduces a new flag, `-XX:+UseGraalJIT` to address these shortcomings. It is an alias for `-XX:+EnableJVMCIProduct -Djvmci.Compiler=graal`. 

When `-XX:+UseGraalJIT` is specified, the VM fails fast at startup if there is a non-Graal JVMCI implementation or no JVMCI implementation in the JDK image.

---------
### Progress
- [x] Change requires CSR request [JDK-8309137](https://bugs.openjdk.org/browse/JDK-8309137) to be approved

### Issues
 * [JDK-8309136](https://bugs.openjdk.org/browse/JDK-8309136): [JVMCI] add -XX:+UseGraalJIT flag
 * [JDK-8309137](https://bugs.openjdk.org/browse/JDK-8309137): [JVMCI] add -XX:+UseGraalJIT flag (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14231/head:pull/14231` \
`$ git checkout pull/14231`

`$ git checkout pull/14231` \
`$ git pull https://git.openjdk.org/jdk.git pull/14231/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14231`

`$ git pr show -t 14231`

</details>

<a href="https://git.openjdk.org/jdk/pull/14231.diff">https://git.openjdk.org/jdk/pull/14231.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14231#issuecomment-1569220539)


src/hotspot/share/jvmci/jvmci_globals.hpp
```diff
@@ -48,6 +48,11 @@ class fileStream;
   product(bool, EnableJVMCI, false, EXPERIMENTAL,                           \
           "Enable JVMCI")                                                   \
                                                                             \
+  product(bool, UseGraalJIT, false,                                         \
```
dholmes-ora, line 51: This should be EXPERIMENTAL

src/hotspot/share/jvmci/jvmci_globals.hpp
```diff
@@ -48,6 +48,11 @@ class fileStream;
...
+          "Select the Graal JVMCI compiler. This is a macro for: "          \
```
dholmes-ora, line 52: s/a macro/an alias/ please

dougxc:
> though not what was linked in the CSR request

Would you mind expanding on the discrepancy? I'd like to make it consistent.

dholmes-ora:
In the CSR request when asked about the response when the run-time image does not include the Graal JIT you pointed to this Java code:
https://github.com/openjdk/jdk/blob/927a9ed68371597eba1609f97ac03dd1de812e26/src/jdk.internal.vm.ci/share/classes/jdk/vm/ci/hotspot/HotSpotJVMCICompilerConfig.java#L112
but it seems to me that the logic in arguments.cpp handles the case where Graal is not available. Or are there different levels of "availability" here?

dougxc:
The logic in arguments.cpp only ensures the value of `jvmci.Compiler` is consistent with `-XX:+UseGraalJIT`. Actually checking the availability of Graal can only be done in the code I linked to and it's only done on the first Graal JIT compilation (JVMCI compiler initialization is lazy to reduce impact to VM startup time).

vnkozlov (commented):
What about `UseJVMCICompiler` flag? A lot of code in HotSpot is guarded by this flag.

dougxc:
> What about `UseJVMCICompiler` flag? A lot of code in HotSpot is guarded by this flag.

`UseGraalJIT` is an alias for `EnableJVMCIProduct` and the latter sets `UseJVMCICompiler` to true.

src/hotspot/share/runtime/arguments.cpp
```diff
@@ -2827,14 +2827,27 @@ jint Arguments::parse_each_vm_init_arg(const JavaVMInitArgs* args, bool* patch_m
         return JNI_ERR;
 #endif // INCLUDE_MANAGEMENT
 #if INCLUDE_JVMCI
-    } else if (match_option(option, "-XX:-EnableJVMCIProduct")) {
+    } else if (match_option(option, "-XX:-EnableJVMCIProduct") || match_option(option, "-XX:-UseGraalJIT")) {
       if (EnableJVMCIProduct) {
         jio_fprintf(defaultStream::error_stream(),
-                  "-XX:-EnableJVMCIProduct cannot come after -XX:+EnableJVMCIProduct\n");
+                  "-XX:-EnableJVMCIProduct or -XX:-UseGraalJIT cannot come after -XX:+EnableJVMCIProduct or -XX:+UseGraalJIT\n");
         return JNI_EINVAL;
```
vnkozlov, line 2834: Why this restriction? Usually latest specified flag wins. May be comment with explanation.
Also you did not check for `if (UseGraalJIT)`. 
What about the case `-XX:+EnableJVMCIProduct -XX:-UseGraalJIT`?
I would expect this kind of checks done in `JVMCIGlobals::check_jvmci_flags_are_consistent()`

test/hotspot/jtreg/compiler/jvmci/TestEnableJVMCIProduct.java
```diff
@@ -67,24 +75,33 @@ public static void main(String[] args) throws Exception {
             new Expectation("EnableJVMCI", "false", "command line"),
             new Expectation("UseJVMCICompiler", "false", "default"));
         test("-XX:+EnableJVMCIProduct",
-            new Expectation("EnableJVMCIProduct", "true", "command line"),
+            new Expectation("EnableJVMCIProduct", "true", "(?:command line|jimage)"),
```
vnkozlov, line 78: What "(?:command line|jimage)" means? Also `|` hard to see.

vnkozlov (commented):
Update Copyright year in test.

dholmes-ora:
>  Actually checking the availability of Graal can only be done in the code I linked to and it's only done on the first Graal JIT compilation (JVMCI compiler initialization is lazy to reduce impact to VM startup time).

Are you saying that the application will start executing quite happily in interpreted mode and then only abort when the JIT first kicks in? That would be a very bad thing to do unless you can guarantee the jit will kick in extremely early during VM init sequence.

src/hotspot/share/runtime/arguments.cpp
dougxc: > Why this restriction? Usually latest specified flag wins. May be comment with explanation.

This is a pre-existing restriction on `EnableJVMCIProduct` and since `UseGraalJIT` is an alias for `EnableJVMCIProduct`, it must abide by the same restrictions. For whatever reason, the effect of `+EnableJVMCIProduct` cannot be undone and so the same must hold for `+UseGraalJIT`.

> Also you did not check for `if (UseGraalJIT)`.

This check is not necessary as `UseGraalJIT` is an alias. When `EnableJVMCIProduct` is set to true, `UseGraalJIT` will also be set to true. Stating this now, I see that I missed it: https://github.com/openjdk/jdk/pull/14231/commits/430b7a58975f49f02de0f0eeb6de5267ee461558

> What about the case -XX:+EnableJVMCIProduct -XX:-UseGraalJIT?

That results in:
```
-XX:-EnableJVMCIProduct or -XX:-UseGraalJIT cannot come after -XX:+EnableJVMCIProduct or -XX:+UseGraalJIT
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```
which is what is expected.

test/hotspot/jtreg/compiler/jvmci/TestEnableJVMCIProduct.java
dougxc: That's a standard Java regular expression and it means the value is expected have been set on the command line or by the VM options read from the jimage (see [JDK-8232080](https://bugs.openjdk.org/browse/JDK-8232080)).

vnkozlov:
> > Actually checking the availability of Graal can only be done in the code I linked to and it's only done on the first Graal JIT compilation (JVMCI compiler initialization is lazy to reduce impact to VM startup time).
> 
> Are you saying that the application will start executing quite happily in interpreted mode and then only abort when the JIT first kicks in? That would be a very bad thing to do unless you can guarantee the jit will kick in extremely early during VM init sequence.

I assume this is required only for Java Graal. For libgraal you can check library presence at the VM startup. Or I am wrong?

dougxc:
> > > Actually checking the availability of Graal can only be done in the code I linked to and it's only done on the first Graal JIT compilation (JVMCI compiler initialization is lazy to reduce impact to VM startup time).
> > 
> > 
> > Are you saying that the application will start executing quite happily in interpreted mode and then only abort when the JIT first kicks in? That would be a very bad thing to do unless you can guarantee the jit will kick in extremely early during VM init sequence.
> 
> I assume this is required only for Java Graal. For libgraal you can check library presence at the VM startup. Or I am wrong?

JVMCI compiler initialization is always lazy. It has been this way since the initial JVMCI implementation. Even libgraal initialization will impact VM startup. The first top tier compilation usually happens within the first 100 ms of VM execution.

vnkozlov (approved):
Looks good now.  You need second approval.

dholmes-ora (approved):
The changes here seem fine, but I remain concerned about the compiler detection logic that presently exists. I can run trivial "programs" using `-XX:+UnlockExperimentalVMOptions -XX:+EnableJVMCIProduct -Djvmci.Compiler=graal` and there is no error. That means that JIT compilation is not kicking in during VM init and that application code is being executed prior to the VM aborting! That is very wrong behaviour IMO, but it is pre-existing so would need to be addressed in separate bug.

dougxc:
Thanks for input David. I agree that it's best to open a new JBS issue to discuss concerns about lazy JVMCI compiler initialization.

ysramakrishna:
> Thanks for input David. I agree that it's best to open a new JBS issue to discuss concerns about lazy JVMCI compiler initialization.

@dougxc : was the "new JBS issue" opened? If so, can it be linked here and in https://bugs.openjdk.org/browse/JDK-8309136.

dougxc:
@ysramakrishna I'm not aware of an issue being opened. I'd prefer to avoid opening an issue until we better understand where the concern results in a real problem in practice. If you have something along these lines, feel free to open the issue.

PR #14232, opened 2023-05-30
8308593: Add Keepalive Extended Socket Options Support for Windows

tkyc:
The PR adds support for the keepalive extended socket options on Windows. For TCP_KEEPIDLE and TCP_KEEPINTVL, these options are supported starting from Windows 10 version 1709. TCP_KEEPCNT is supported starting from Windows 10 version 1703. Information on these socket options can be found [here](https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-tcp-socket-options).

I've also corrected the `handleError()` function. On Windows, the error needs to be retrieved using `WSAGetLastError()` and error codes are prefixed with "WSA". Information on this can be found [here](https://learn.microsoft.com/en-us/windows/win32/winsock/error-codes-errno-h-errno-and-wsagetlasterror-2).

>The error codes returned by Windows Sockets are similar to UNIX socket error code constants, but the constants are all prefixed with WSA.

>Error codes set by Windows Sockets are not made available through the errno variable.

No new tests were added as the existing tests should cover this.

---------
### Progress

### Issue
 * [JDK-8308593](https://bugs.openjdk.org/browse/JDK-8308593): Add KEEPALIVE Extended Socket Options Support for Windows (**Enhancement** - P4)

### Reviewers
 * [Michael McMahon](https://openjdk.org/census#michaelm) (@Michael-Mc-Mahon - **Reviewer**) ⚠️ Review applies to [2ada2207](https://git.openjdk.org/jdk/pull/14232/files/2ada2207c1f9fa16da3b1ca2977229403c712852)
 * [Vyom Tewari](https://openjdk.org/census#vtewari) (@vyommani - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14232/head:pull/14232` \
`$ git checkout pull/14232`

`$ git checkout pull/14232` \
`$ git pull https://git.openjdk.org/jdk.git pull/14232/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14232`

`$ git pr show -t 14232`

</details>

<a href="https://git.openjdk.org/jdk/pull/14232.diff">https://git.openjdk.org/jdk/pull/14232.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14232#issuecomment-1570719929)

### DIFF
--- a/src/jdk.net/windows/classes/jdk/net/WindowsSocketOptions.java
+++ b/src/jdk.net/windows/classes/jdk/net/WindowsSocketOptions.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2022, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
--- a/src/jdk.net/windows/classes/jdk/net/WindowsSocketOptions.java
+++ b/src/jdk.net/windows/classes/jdk/net/WindowsSocketOptions.java
@@ -42,6 +42,11 @@ class WindowsSocketOptions extends PlatformSocketOptions {
     }
 
     @Override
+    boolean keepAliveOptionsSupported() {
+        return keepAliveOptionsSupported0();
+    }
+
+    @Override
     void setIpDontFragment(int fd, final boolean value, boolean isIPv6) throws SocketException {
         setIpDontFragment0(fd, value, isIPv6);
     }
@@ -51,8 +56,45 @@ class WindowsSocketOptions extends PlatformSocketOptions {
         return getIpDontFragment0(fd, isIPv6);
     }
 
+    @Override
+    void setTcpkeepAliveProbes(int fd, final int value) throws SocketException {
+        setTcpkeepAliveProbes0(fd, value);
+    }
+
+    @Override
+    int getTcpkeepAliveProbes(int fd) throws SocketException {
+        return getTcpkeepAliveProbes0(fd);
+    }
+
+    @Override
+    void setTcpKeepAliveTime(int fd, final int value) throws SocketException {
+        setTcpKeepAliveTime0(fd, value);
+    }
+
+    @Override
+    int getTcpKeepAliveTime(int fd) throws SocketException {
+        return getTcpKeepAliveTime0(fd);
+    }
+
+    @Override
+    void setTcpKeepAliveIntvl(int fd, final int value) throws SocketException {
+        setTcpKeepAliveIntvl0(fd, value);
+    }
+
+    @Override
+    int getTcpKeepAliveIntvl(int fd) throws SocketException {
+        return getTcpKeepAliveIntvl0(fd);
+    }
+
+    private static native boolean keepAliveOptionsSupported0();
     private static native void setIpDontFragment0(int fd, boolean value, boolean isIPv6) throws SocketException;
     private static native boolean getIpDontFragment0(int fd, boolean isIPv6) throws SocketException;
+    private static native void setTcpkeepAliveProbes0(int fd, int value) throws SocketException;
+    private static native int getTcpkeepAliveProbes0(int fd) throws SocketException;
+    private static native void setTcpKeepAliveTime0(int fd, int value) throws SocketException;
+    private static native int getTcpKeepAliveTime0(int fd) throws SocketException;
+    private static native void setTcpKeepAliveIntvl0(int fd, int value) throws SocketException;
+    private static native int getTcpKeepAliveIntvl0(int fd) throws SocketException;
 
     static {
         if (System.getSecurityManager() == null) {
--- a/src/jdk.net/windows/classes/jdk/net/WindowsSocketOptions.java
+++ b/src/jdk.net/windows/classes/jdk/net/WindowsSocketOptions.java
@@ -57,13 +57,13 @@ class WindowsSocketOptions extends PlatformSocketOptions {
     }
 
     @Override
-    void setTcpkeepAliveProbes(int fd, final int value) throws SocketException {
-        setTcpkeepAliveProbes0(fd, value);
+    void setTcpKeepAliveProbes(int fd, final int value) throws SocketException {
+        setTcpKeepAliveProbes0(fd, value);
     }
 
     @Override
-    int getTcpkeepAliveProbes(int fd) throws SocketException {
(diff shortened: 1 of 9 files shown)


AlanBateman:
I think you may have missed my comment in the JBS issue where I asked about the semantics of SIO_KEEPALIVE_VALS and whether we could implement TCP_KEEPIDLE and TCP_KEEPINTERVAL on Windows by knowing what the defaults are. The proposal that you have here is a new Windows specific and JDK-specific socket option that requires a lot of thinking about before going that route.

AlanBateman:
In the discussion on net-dev/nio-dev, Daniel Jeliński pointed out that Windows does have TCP_KEEPIDLE and TCP_KEEPINTVL since Windows 10 version 1709. So I assume this PR can be returned to closed or re-returned to draft while the implementation changes are developed to make use of these socket options.

tkyc:
HI, just wondering if there's anything needed from me further (the related threads have been inactive for some time).

AlanBateman:
> HI, just wondering if there's anything needed from me further (the related threads have been inactive for some time).

The original to use TCP_SIO_KEEPALIVE was problematic for several reasons, changing it to use TCP_KEEPCNT, TCP_KEEPIDLE, and TCP_KEEPINTVL after finding out that these socket options are supported on Windows is good. So I think we should review and include this.

jaikiran:
The changes to `WindowsSocketOptions.java` which are merely wiring changes, look OK to me. I don't have experience with Windows nor much JNI code, so I'll let other reviewers take a look.

I see that the JNI code is updated to use `WSAGetLastError()` instead of `errno` to get the error code. You already provided a link to the documentation of that API and reading through it, it appears to be the right thing to do and it seems to be supported on all relevant Windows versions of interest. However, that change does impact other existing code in that file. Do you think, we could do that change in a separate PR?

In the meantime, could you please update the copyright year on both these changed files to `2022, 2023, ` from the current `2022, `?

jaikiran:
Hello Terry,

> No new tests were added as the existing tests should cover this.

The existing `test/jdk/java/net/SocketOption/TcpKeepAliveTest.java` test should indeed cover these changes. It checks the `supportedOptions()` and if it finds these `TCP_KEEPCOUNT`, `TCP_KEEPIDLE` and `TCP_KEEPINTERVAL` to be supported then it runs the relevant tests. I'll run some tests with the current state of this PR to make sure these tests are running as expected.

Michael-Mc-Mahon:
Webrev 01 seems to be missing a number of files.

tkyc:
Thanks all for the review so far. I'll make the copyright year changes. So, the consensus for the `errno` to `WSAGetLastError` change is that it's the right thing to do. But, I should make this change in another PR? Are we all in agreement with that?

jaikiran:
I think Michael means that you can keep the `WSAGetLastError` change in this current PR. I don't have any objection to that.

Michael-Mc-Mahon:
> I think Michael means that you can keep the `WSAGetLastError` change in this current PR. I don't have any objection to that.

Right. That's what I meant. The change looks good to me, but I might run the patch through the test system here to see how it works with the existing tests in this area and the change above doesn't cause any issues.

AlanBateman:
> I think it is right to change the `errno` to `WSAGetLastError`. There is an errno.h in VC++ which must be getting included indirectly. It is used in hotspot. But, we shouldn't be using it in the networking code.

Right, WSAGetLastError should be used with Windows Sockets, not errno.

Michael-Mc-Mahon (approved):
Looks good to me, subject to copyright update.

tkyc:
Awesome, thanks all for the review.

djelinski (approved):
LGTM. Verified that `test/jdk/java/net/SocketOption/TcpKeepAliveTest.java` is able to set and get the options on Windows 11 as expected.

On a side note, the test could be more verbose - currently it is not clear if the test passes because the options worked or because they were unsupported. But that's preexisting and not much of a problem anyway.

src/jdk.net/windows/classes/jdk/net/WindowsSocketOptions.java
```diff
@@ -51,8 +56,45 @@ boolean getIpDontFragment(int fd, boolean isIPv6) throws SocketException {
...
+    @Override
+    int getTcpKeepAliveIntvl(int fd) throws SocketException {
+        return getTcpKeepAliveIntvl0(fd);
+    }
+
+    private static native boolean keepAliveOptionsSupported0();
     private static native void setIpDontFragment0(int fd, boolean value, boolean isIPv6) throws SocketException;
     private static native boolean getIpDontFragment0(int fd, boolean isIPv6) throws SocketException;
+    private static native void setTcpkeepAliveProbes0(int fd, int value) throws SocketException;
+    private static native int getTcpkeepAliveProbes0(int fd) throws SocketException;
```
vyommani, line 93: there is inconsistency in function naming some place you use 'keep' and other places you use 'Keep', can you please make it consistence ?.

tkyc: Fixed inconsistency.

vyommani (approved):
thanks for doing changes for other platforms as well. Latest changes looks OK to me.

Michael-Mc-Mahon:
It looks like you need to type the "integrate" command again, after the last change. We can sponsor it then.

jaikiran:
> On a side note, the test could be more verbose - currently it is not clear if the test passes because the options worked or because they were unsupported. But that's preexisting and not much of a problem anyway.

Agreed. When I tested this PR in our CI, I had to locally update `TcpKeepAliveTest.java` to make sure the options were indeed being considered. I'll open a separate JBS issue and propose a PR with these log messages changes in that test.

djelinski (approved):

tkyc:
Thanks all. /integrate

dfuch:
Hi, this issue is marked as needing a release note but none has been produced.
Could you please add a release note to [JDK-8308593](https://bugs.openjdk.org/browse/JDK-8308593) for this?
Instructions are available here: https://openjdk.org/guide/#release-notes

tkyc:
@dfuch Sorry I missed that step. I'll add a release note according to your linked instructions.

tkyc:
Just a heads up, incase things move slow. Since I don't have a JBS account, I'm just waiting on some help from my msft liaison to help add the release note to the ticket.

tkyc:
Hello, a release note was added to the ticket awhile ago. Let me know if there's anything else I need to do or correct. Thanks again.

dfuch:
Thanks Terry. I have made some updates and marked it DELIVERED. Can you double check that everything still looks good?

tkyc:
@dfuch LGTM. Thanks for the quick update Daniel.

PR #14238, opened 2023-05-31
8308764: Reporting errors from create_vm may crash

dholmes-ora:
There are more details in the bug report but the basic fix is to not delete the main_thread so that it is available to report the exception that caused VM initialization to fail. Added some additional comments too.

Testing:
 - basic fault injection to test non-exception exit paths both before and after Universe:;is_fully_initialized() which exercises the failing use of JavaThread::current()
 -  tiers 1-3 sanity testing

Thanks.

---------
### Progress

### Issue
 * [JDK-8308764](https://bugs.openjdk.org/browse/JDK-8308764): Reporting errors from create_vm may crash (**Bug** - `"4"`)

### Reviewers
 * [Coleen Phillimore](https://openjdk.org/census#coleenp) (@coleenp - **Reviewer**) ⚠️ Review applies to [1697ed7e](https://git.openjdk.org/jdk/pull/14238/files/1697ed7e7735f0fe9f59c9b5c5889b70326049f5)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14238/head:pull/14238` \
`$ git checkout pull/14238`

`$ git checkout pull/14238` \
`$ git pull https://git.openjdk.org/jdk.git pull/14238/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14238`

`$ git pr show -t 14238`

</details>

<a href="https://git.openjdk.org/jdk/pull/14238.diff">https://git.openjdk.org/jdk/pull/14238.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14238#issuecomment-1569423869)


dholmes-ora:
Redirected emails to hotspot-runtime-dev.

src/hotspot/share/runtime/threads.cpp
```diff
@@ -548,7 +548,7 @@ jint Threads::create_vm(JavaVMInitArgs* args, bool* canTryAgain) {
   // Initialize global modules
   jint status = init_globals();
   if (status != JNI_OK) {
-    main_thread->smr_delete();
+    // Don't delete main_thread here as we may need to report exceptions.
     *canTryAgain = false; // don't let caller call JNI_CreateJavaVM again
```
coleenp, line 552: So could the jni code try to create a vm, fail and then go on to do other stuff?  Would this then be a memory leak?

coleenp: maybe this should be a if (!HAS_PENDING_EXCEPTION) main_thread->smr_delete(); and then in the caller will exit with vm_exit_during_initialization() which does exit() so not necessary to delete the main thread.

dholmes-ora: > Would this then be a memory leak?

Yes but at this point we are "leaking everything that VM initialization created but does not clean up. So adding a JavaThread instance to that seems insignificant.

> maybe this should be a if (!HAS_PENDING_EXCEPTION) main_thread->smr_delete(); 

That would work too. I can make that switch.

tstuefe (approved):
Looks good.

dholmes-ora:
Thanks for the review @tstuefe !

src/hotspot/share/runtime/threads.cpp
```diff
@@ -548,7 +548,10 @@ jint Threads::create_vm(JavaVMInitArgs* args, bool* canTryAgain) {
   // Initialize global modules
   jint status = init_globals();
   if (status != JNI_OK) {
-    main_thread->smr_delete();
+    if (!main_thread->has_pending_exception()) {
+      main_thread->smr_delete();
+    }
+    // Else don't delete main_thread as we need to report exceptions in the caller.
```
kimbarrett, line 554: Is it really okay to just leak main_thread?  If the process is going to die anyway then sure, but I'm not sure
that's true for all potential callers.  Of course, there's probably lots of other stuff getting leaked, so maybe
it doesn't matter.

src/hotspot/share/runtime/threads.cpp
```diff
@@ -548,7 +548,10 @@ jint Threads::create_vm(JavaVMInitArgs* args, bool* canTryAgain) {
   // Initialize global modules
   jint status = init_globals();
   if (status != JNI_OK) {
-    main_thread->smr_delete();
+    if (!main_thread->has_pending_exception()) {
+      main_thread->smr_delete();
+    }
```
kimbarrett, line 553: We have the exact same problem below after the call to init_globals2 (very recently introduced by JDK-8240774).

kimbarrett (changes requested):

src/hotspot/share/runtime/threads.cpp
kimbarrett: I see this was already discussed, with the same rationale for not caring about the leak.

dholmes-ora: Thanks for looking at this Kim. With everything else we "leak" when we don't abort the process during VM init, adding the main thread to it would not make any difference IMO. But in any case with Coleen's suggestion we no longer "leak" the main thread as we will abort.

src/hotspot/share/runtime/threads.cpp
dholmes-ora: Thanks @kimbarrett I had obviously missed that.

dholmes-ora: Fixed.

coleenp: Is it possible to have a pending exception after init_globals() now since Java initializations are done in init_globals2?

src/hotspot/share/prims/jni.cpp
```diff
@@ -3633,7 +3633,13 @@ static jint JNI_CreateJavaVM_inner(JavaVM **vm, void **penv, void *args) {
         HandleMark hm(THREAD);
         vm_exit_during_initialization(Handle(THREAD, PENDING_EXCEPTION));
       }
+      // Unclear if we can actually reach here as all error returns should have
+      // set an exception pending. But it is okay to fall-through to the general
+      // termination code below.
```
kimbarrett, line 3638: If it were true that all error returns here had an exception pending then we wouldn't need the conditional
smr_deletes in create_vm.  And if we return without an exception then we'll also not have a current thread
and the JavaThread::current call will return null (or assert in a debug build), leading to failures.  That is,
this change doesn't seem to have addressed the issue described in the bug at all.

kimbarrett (changes requested):

coleenp (approved):
I just had one question, but it looks fine either with or without the extra pending exception check.

src/hotspot/share/prims/jni.cpp
dholmes-ora: I think you are right - I tried to make too many changes. The original code just didn't delete the main-thread - nice and simple but "leaks". With Coleen's suggested change we only keep the main thread if there is an exception pending, but that then opens up the possibility that we reach this section of code with no current thread again.

dholmes-ora:
Take 3.

After the split of init_globals into two we only have the pending exception issue for the second one. As we only keep the main thread when there is a pending exception then we have to watch for a null thread before checking for exceptions. Removed additional commentary that could just cause confusion.

Please re-review @tstuefe , @coleenp , @kimbarrett - thanks.

src/hotspot/share/runtime/threads.cpp
```diff
@@ -564,7 +564,12 @@ jint Threads::create_vm(JavaVMInitArgs* args, bool* canTryAgain) {
   status = init_globals2();
   if (status != JNI_OK) {
     Threads::remove(main_thread, false);
-    main_thread->smr_delete();
+    // It is possible that we managed to fully initialize Universe but have then
+    // failed by throwing an exception. In that case our caller JNI_CreateJavaVM
+    // we will want to report it, so we can't delete the main thread.
```
kimbarrett, line 569: s/we//

kimbarrett (approved):
One minor comment nit.  Looks good otherwise.

src/hotspot/share/runtime/threads.cpp
dholmes-ora: Fixed - thanks Kim!

src/hotspot/share/runtime/threads.cpp
```diff
@@ -564,7 +564,12 @@ jint Threads::create_vm(JavaVMInitArgs* args, bool* canTryAgain) {
   status = init_globals2();
   if (status != JNI_OK) {
     Threads::remove(main_thread, false);
-    main_thread->smr_delete();
+    // It is possible that we managed to fully initialize Universe but have then
+    // failed by throwing an exception. In that case our caller JNI_CreateJavaVM
+    // will want to report it, so we can't delete the main thread.
+    if (!main_thread->has_pending_exception()) {
+      main_thread->smr_delete();
+    }
```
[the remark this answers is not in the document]
coleenp: Also init_globals2 should have a better name, like init_objects_and_globals() or something like that.  But that's not this issue.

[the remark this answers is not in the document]
kimbarrett: I was later wondering the same thing. Here's my rationale for putting it here
instead of in the JNI code (i.e. agreeing with Coleen's suggestion).

I think neither place is really correct. Failure to create and initialize the
VM ought to always propogate out to the JNI client to decide what to do about
the failure. But we don't do that, not only here but all over the place. These
(and many other) failures are treated as grounds for terminating the process
containing the VM, regardless of what else that process might be doing. This
limits the utility of the VM in ways similar to not being able to have
multiple VMs (at the same time or serially) in one process. Oh well.

Given that, putting the vm_exit_during_initialization() in create_vm seems
simpler.  It's not even clear create_vm should ever return failure with
canTryAgain false; maybe it should always terminate if there's a non-retryable
failure.  But perhaps that's for a later day.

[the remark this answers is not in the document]
kimbarrett: I agree that init_globals2 isn't a great name. I failed to come up with
something better, either now or when I added it.  Anyone with something better
should go for it.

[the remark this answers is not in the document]
dholmes-ora: Sorry but this is getting a bit out of hand. Rearranging the deckchairs really doesn't change anything. You could argue that everything should happen in create_vm so JNI_CreateJavaVM is just a minimal wrapper, but that is not what fixing this issue is about. All I'm trying to do here is avoid the crash when there is no longer a current thread.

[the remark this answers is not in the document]
dholmes-ora: New version committed.

dholmes-ora:
Take 4

dholmes-ora:
Scratch take 4. You can't just move the exception handling into create_vm at that location because there are other places after that which can throw exceptions. Trying to revert to version 3 but I think I messed up my repo. <sigh>

src/hotspot/share/runtime/threads.cpp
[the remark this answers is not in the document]
dholmes-ora: New version reverted. You can't just do the exception check here because there is other code below that can thrown exceptions.

kimbarrett:
> Scratch take 4. You can't just move the exception handling into create_vm at that location because there are other places after that which can throw exceptions.

OK, that's convincing.  Looks good.

dholmes-ora:
Thanks @kimbarrett . can you hit the approve button again so it records which version you approved. Thanks.

kimbarrett (approved):

tstuefe (approved):
Okay.

I preferred the old solution, since if an error happens at initialization, I like the JVM to stop as fast as possible. Going through smr_delete seems superfluous, and the leak not really important, since we leak a lot of other stuff anyway.

dholmes-ora:
Thanks @tstuefe ! This solution is kind of minimalistic in that it only preserves the main-thread when we know there is an exception pending.

dholmes-ora:
As time is pressing I can't wait for Coleen's final thumbs up - sorry.

Thanks for all the reviews and time spent on this.

/integrate

PR #14241, opened 2023-05-31
8308387: CLD created and unloading list sharing _next node pointer leads to concurrent YC missing CLD roots

xmas92:
[JDK-8307106](https://bugs.openjdk.org/browse/JDK-8307106) introduced the ability to walk the created CLDs list in the CLDG without a lock. This was primarily introduced to allow lockless concurrent CLD roots scanning for young collections in generational ZGC. However because the CLD _next node pointer is shared between the two list this can lead to a concurrent iteration of the created CLDs missing list entries. 

This change introduces a second _unloading_next node pointer which is used for the unloading CLDs list. The set_next is now maintains the invariant that it only ever unlinks is_unloading() CLDs and maintains a consistent view of the tail list for anyone reading the list concurrently.  

Testing: tier1-3 and tier1-7 with Generational ZGC

---------
### Progress

### Issue
 * [JDK-8308387](https://bugs.openjdk.org/browse/JDK-8308387): CLD created and unloading list sharing _next node pointer leads to concurrent YC missing CLD roots

### Reviewers
 * [David Holmes](https://openjdk.org/census#dholmes) (@dholmes-ora - **Reviewer**) ⚠️ Review applies to [320be537](https://git.openjdk.org/jdk/pull/14241/files/320be5372ad4071c601ac3f1ab2ff510944df5df)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14241/head:pull/14241` \
`$ git checkout pull/14241`

`$ git checkout pull/14241` \
`$ git pull https://git.openjdk.org/jdk.git pull/14241/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14241`

`$ git pr show -t 14241`

</details>

<a href="https://git.openjdk.org/jdk/pull/14241.diff">https://git.openjdk.org/jdk/pull/14241.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14241#issuecomment-1569744670)

### DIFF
--- a/src/hotspot/share/classfile/classLoaderData.hpp
+++ b/src/hotspot/share/classfile/classLoaderData.hpp
@@ -153,7 +153,27 @@ class ClassLoaderData : public CHeapObj<mtClass> {
   GrowableArray<Metadata*>*      _deallocate_list;
 
   // Support for walking class loader data objects
-  ClassLoaderData* _next; /// Next loader_datas created
+  // The ClassLoaderDataGraph maintains two lists of CLDs, one list which
+  // contains the created CLDs, and another which contains the unloading
+  // CLDs. To enable the lock free concurrent iteration of the created
+  // CLDs list used by the GC for root scanning, it is important that unlinking
+  // CLDs from the created list upholds the invariant that all unlinked CLDs are
+  // is_unloading() and that the new tail list inserted is the same as the tail
+  // list before unlinking.
+  // That is given a list A -> B -> C -> D -> E where B and C are is_unloading()
+  // the only valid calls to set_next(ClassLoaderData*) are:
+  //   A.set_next(B),A.set_next(C),A.set_next(D)
+  //   B.set_next(C),B.set_next(D)
+  //   C.set_next(D)
+  //   D.set_next(E)
+  //   E.set_next(nullptr)
+  // Any insertions to the created CLDs list is done at the head and keeps
+  // the list tail invariant.
+  // CLDs are unlinked in ClassLoaderDataGraph::do_unloading() and released in
+  // ClassLoaderDataGraph::purge(), so some syncronization is required between
+  // the two to ensure that no unlinked CLD remains in the system leading
+  // to a use after free.
+  ClassLoaderData* _next;
   ClassLoaderData* _unloading_next;
 
   Klass*  _class_loader_klass;
--- a/src/hotspot/share/classfile/classLoaderData.hpp
+++ b/src/hotspot/share/classfile/classLoaderData.hpp
@@ -157,18 +157,20 @@ class ClassLoaderData : public CHeapObj<mtClass> {
   // The ClassLoaderDataGraph maintains two lists to keep track of CLDs.
   //
   // The first list [_head, _next] is where new CLDs are registered. The CLDs
-  // are only inserted at the _head, and the _next pointers are never rewritten.
+  // are only inserted at the _head, and the _next pointers are only rewritten
+  // from unlink_next() which unlinks one unloading CLD by setting _next to
+  // _next->_next.
   // This allows GCs to concurrently walk the list while the CLDs are being
   // concurrently unlinked.
   //
   // The second list [_unloading_head, _unloading_next] is where dead CLDs get
   // moved to during class unloading. See: ClassLoaderDataGraph::do_unloading().
-  // This list is never modified while other threads are iterating over it. 
+  // This list is never modified while other threads are iterating over it.
   //
   // After all dead CLDs have been moved to the unloading list, there's a
   // synchronziation point (handshake) to ensure that all threads reading these
   // CLDs finish their work. This ensures that we don't have a user-after-free
-  // when we later delete the CLDs. 
+  // when we later delete the CLDs.
   //
   // And finally, when no threads are using the unloading CLDs anymore, we
   // remove them from the class unloading list and delete them. See:
--- a/src/hotspot/share/classfile/classLoaderData.hpp
+++ b/src/hotspot/share/classfile/classLoaderData.hpp
@@ -168,7 +168,7 @@ class ClassLoaderData : public CHeapObj<mtClass> {
   //
   // After all dead CLDs have been moved to the unloading list, there's a
   // synchronization point (handshake) to ensure that all threads reading these
-  // CLDs finish their work. This ensures that we don't have a user-after-free
+  // CLDs finish their work. This ensures that we don't have a use-after-free
   // when we later delete the CLDs.
   //
   // And finally, when no threads are using the unloading CLDs anymore, we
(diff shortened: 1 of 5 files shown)


src/hotspot/share/classfile/classLoaderData.hpp
```diff
@@ -153,15 +153,32 @@ class ClassLoaderData : public CHeapObj<mtClass> {
...
+  // CLDs. To enable the lock free concurrent iteration of the created
+  // CLDs list used by the GC for root scanning, it is important that unlinking
+  // CLDs from the created list upholds the invariant that all unlinked CLDs are
+  // is_unloading().
+  // Any insertions to the created CLDs list is done at the head and keeps
+  // the list tail invariant.
+  // CLDs are unlinked in ClassLoaderDataGraph::do_unloading() and released in
+  // ClassLoaderDataGraph::purge(), so some syncronization is required between
+  // the two to ensure that no unlinked CLD remains in the system leading
+  // to a use after free.
```
stefank, line 167: I talked to Axel about the wording around what the two list contained, and how there's a slight risk that someone who's not up-to-speed in this area might misinterpret parts of the comment.

What do you think about something along the lines of this suggestion?:

```suggestion
  //
  // The ClassLoaderDataGraph maintains two lists to keep track of CLDs.
  //
  // The first list [_head, _next] is where new CLDs are registered. The CLDs
  // are only inserted at the _head, and the _next pointers are never rewritten.
  // This allows GCs to concurrently walk the list while the CLDs are being
  // concurrently unlinked.
  //
  // The second list [_unloading_head, _unloading_next] is where dead CLDs get
  // moved to during class unloading. See: ClassLoaderDataGraph::do_unloading().
  // This list is never modified while other threads are iterating over it. 
  //
  // After all dead CLDs have been moved to the unloading list, there's a
  // synchronziation point (handshake) to ensure that all threads reading these
  // CLDs finish their work. This ensures that we don't have a user-after-free
  // when we later delete the CLDs. 
  //
  // And finally, when no threads are using the unloading CLDs anymore, we
  // remove them from the class unloading list and delete them. See:
  // ClassLoaderDataGraph::purge();
```

stefank (approved):
I think the patch looks good. I've added a suggestion for a rewording of the comment about the two list. Feel free to skip it if you liked your comment better.

src/hotspot/share/classfile/classLoaderData.hpp
xmas92: Your comment seems more structured, I will use it. Will only update the `and the _next pointers are never rewritten.` part to reflect the fact that `unlink_next()` will update it by removing one is_unloading CLD.

stefank (approved):

stefank (approved):

src/hotspot/share/classfile/classLoaderDataGraph.cpp
```diff
@@ -501,37 +494,34 @@ bool ClassLoaderDataGraph::is_valid(ClassLoaderData* loader_data) {
...
-  ClassLoaderData* data = _head;
   ClassLoaderData* prev = nullptr;
   bool seen_dead_loader = false;
   uint loaders_processed = 0;
   uint loaders_removed = 0;
 
-  data = _head;
-  while (data != nullptr) {
+  for (ClassLoaderData* data = _head; data != nullptr; data = data->next()) {
     if (data->is_alive()) {
```
coleenp, line 503: There used to be more code for the case where data->is_alive() so this code had this continue, but you could simplify this by having an 'else' clause and remove the continue. That would clean this up a little.

coleenp (approved):
Looks good.  Is there a way to add a test?

src/hotspot/share/classfile/classLoaderDataGraph.cpp
```diff
@@ -268,14 +269,6 @@ inline void assert_is_safepoint_or_gc() {
          "Must be called by safepoint or GC");
 }
 
-void ClassLoaderDataGraph::cld_unloading_do(CLDClosure* cl) {
```
coleenp, line 271: Thanks for noticing that we're not using this anymore and cleaning it up.  JFR used to use this.

src/hotspot/share/classfile/classLoaderData.hpp
```diff
@@ -153,15 +153,41 @@ class ClassLoaderData : public CHeapObj<mtClass> {
...
+  // _next->_next. This allows GCs to concurrently walk the list while the CLDs
+  // are being concurrently unlinked.
+  //
+  // The second list [_unloading_head, _unloading_next] is where dead CLDs get
+  // moved to during class unloading. See: ClassLoaderDataGraph::do_unloading().
+  // This list is never modified while other threads are iterating over it.
+  //
+  // After all dead CLDs have been moved to the unloading list, there's a
+  // synchronization point (handshake) to ensure that all threads reading these
+  // CLDs finish their work. This ensures that we don't have a user-after-free
```
dholmes-ora, line 171: use-after-free ?

dholmes-ora (approved):
This logic is much clearer. Thanks for fixing the bug and cleaning it up.

dholmes-ora:
Is the YC in the synopsis meant to be GC? If so please fix in JBS and the PR. Thanks

src/hotspot/share/classfile/classLoaderDataGraph.cpp
xmas92: Fixed

src/hotspot/share/classfile/classLoaderData.hpp
xmas92: Fixed

xmas92:
> Looks good. Is there a way to add a test?

The timing window here is really narrow. It requires the young root scan iteration to reach a dead CLD before it is unlinked by the old collection and hold on to it until after it is linked onto the unloading list, the young root scan filters out dead CLDs so this is just a couple of instructions. It is possible to make the the race more reproducible with a couple of sleeps in the VM and running frequent concurrent young collections and class unloading.

There might be a possibility to abstract the linked list implementation and make it mockable in gtests, and then have gtests which stresses concurrent reads and unlinks and asserts invariants. However that would require a rewrite that does not feel worth it.

> Is the YC in the synopsis meant to be GC? If so please fix in JBS and the PR. Thanks

YC is suppose to abbreviate Young Collection in a generational collector. But maybe YC/OC are not common enough terms to be used in a general bug report. 
I will at least extend the description on JBS to reflect the bug, not just the symptom (the crash).

stefank (approved):

coleenp (approved):

coleenp:
> Looks good. Is there a way to add a test?

Better not to add a test if it's sensitive to timing.  thanks.

xmas92:
Thanks for all the reviews.
/integrate

PR #14246, opened 2023-05-31
8006421: GraphicsConfiguration of a frame is changed when the frame is moved to another screen

TejeshR13:
When a frame is dragged from one screen to another screen, the `GraphicsConfigurations` may vary depending on screen/monitor. However, transparency has to be maintained since it is expected as it is set. So the fix addresses in maintaining the translucency capability when frame is moved from one screen to another screen. For main screen, translucency is set when `setBackground` is called, but for second screen the same is not taken care. Hence setting of Translucency Capable `GC` is taken care when screen switch happens. The fix handles only for non-opaque windows alone retaining the behavior for opaque windows. 
The fix is verified in CI for regression check, which is Green.

---------
### Progress

### Issue
 * [JDK-8006421](https://bugs.openjdk.org/browse/JDK-8006421): GraphicsConfiguration of a frame is changed when the frame is moved to another screen (**Bug** - P3)

### Reviewers
 * [Alexander Zvegintsev](https://openjdk.org/census#azvegint) (@azvegint - **Reviewer**) ⚠️ Review applies to [7b2b2b01](https://git.openjdk.org/jdk/pull/14246/files/7b2b2b01bb262960e42a96fe30af30c27d1d7d1e)
 * [Sergey Bylokhov](https://openjdk.org/census#serb) (@mrserb - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14246/head:pull/14246` \
`$ git checkout pull/14246`

`$ git checkout pull/14246` \
`$ git pull https://git.openjdk.org/jdk.git pull/14246/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14246`

`$ git pr show -t 14246`

</details>

<a href="https://git.openjdk.org/jdk/pull/14246.diff">https://git.openjdk.org/jdk/pull/14246.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14246#issuecomment-1571315042)


src/java.desktop/unix/classes/sun/awt/X11/XWindowPeer.java
```diff
@@ -694,12 +694,34 @@ public void checkIfOnNewScreen(Rectangle newBounds) {
...
+                                && target.getBackground() != SystemColor.window
+                                && !newGC.isTranslucencyCapable()) {
+                            GraphicsConfiguration[] configs = gds[i].getConfigurations();
+                            for (int j = 0; j < configs.length; ++j) {
+                                if (configs[j].isTranslucencyCapable()) {
+                                    newGC = configs[j];
+                                    break;
+                                }
+                            }
+                        }
```
azvegint, line 707: Looks like this code is also repeated on lines 714-724 and can be extracted to a function.

mrserb:
We should check why that code does not work, note the code in that patch also support both Window and Canvas.

TejeshR13:
> We should check why that code does not work, note the code in that patch also support both Window and Canvas.

We can use `getAppropriateGraphicsConfiguration()` of `XCanvasPeer` class, which solves the problem. We can use it whenever new screen is set/new screen GC is set (Not sure of why it wasn't used before). The old code gets defaultGC whenever screen switch happens. We can use the existing implementation, which is actually better than selecting TransparencyCapable GC alone. Their exist a regression for this re-use solution i.e., test [WindowGCChangeTest.java](https://github.com/openjdk/jdk/blob/master/test/jdk/java/awt/Multiscreen/WindowGCChangeTest/WindowGCChangeTest.java), which is basically windows specific test.

src/java.desktop/unix/classes/sun/awt/X11/XWindowPeer.java
```diff
@@ -694,12 +694,34 @@ public void checkIfOnNewScreen(Rectangle newBounds) {
                         // Completely on this screen - done!
                         newScreenNum = i;
                         newGC = gds[i].getDefaultConfiguration();
+                        if (target != null
+                                && target.getBackground() != SystemColor.window
```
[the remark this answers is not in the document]
DamonGuy: I agree. I recently worked on a bug involving window backgrounds and translucency is only one aspect of it. The color itself may be different and this conditional looks like it would definitely trigger in the case where the window changes color between screens.

test/jdk/java/awt/Window/MultiscreenTransparencyTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+import javax.swing.JFrame;
+import javax.swing.JPanel;
+import javax.swing.SwingUtilities;
+
+/*
+ * @test
+ * @bug 8006421
+ * @key headful
+ * @requires (os.family == "linux")
```
DamonGuy, line 43: Does this have to be linux only? Can't this be a generic test?

TejeshR13: The fix is specifically for linux ( XWindowPeer).

TejeshR13: Not sure if it can be made generic, because as explained it causes a regression for test - [WindowGCChangeTest.java](https://github.com/openjdk/jdk/blob/master/test/jdk/java/awt/Multiscreen/WindowGCChangeTest/WindowGCChangeTest.java).

DamonGuy: I see the difference. Also, I looked back on other fixes that I recall were expanded to be generic, and I see my source of confusion. LAF's were expanded, not tested OS's.

TejeshR13: Yeah, so as @azvegint suggested I will update the fix. And also I will update the regression test to windows alone.

azvegint: I just did a quick test on Windows and macos, this test seems to work.
(However, line 136 should be removed to show the translucent window on macos).

Therefore, we may consider doing a test not only for Linux.

BTW, It may be better to put this test in test/jdk/java/awt/Multiscreen/, where most of multiscreen tests are located.

test/jdk/java/awt/Window/MultiscreenTransparencyTest.java
```diff
@@ -0,0 +1,141 @@
...
+    static Point dragStart = new Point();
+    static JFrame frame;
+    static JPanel panel;
+    static PassFailJFrame passFailJFrame;
+
+    public static void main(String[] args) throws Exception {
+        GraphicsEnvironment ge = GraphicsEnvironment.getLocalGraphicsEnvironment();
+        GraphicsDevice[] gds = ge.getScreenDevices();
+
+        if (gds.length != 2) {
```
azvegint, line 61: Shouldn't it be `> 1`? 
To perform the test when 2 or more screens are present.

mrserb:
> > We should check why that code does not work, note the code in that patch also support both Window and Canvas.
> 
> We can use `getAppropriateGraphicsConfiguration()` of `XCanvasPeer` class, which solves the problem. We can use it whenever new screen is set/new screen GC is set (Not sure of why it wasn't used before). The old code gets defaultGC whenever screen switch happens. We can use the existing implementation, which is actually better than selecting TransparencyCapable GC alone. Their exist a regression for this re-use solution i.e., test [WindowGCChangeTest.java](https://github.com/openjdk/jdk/blob/master/test/jdk/java/awt/Multiscreen/WindowGCChangeTest/WindowGCChangeTest.java), which is basically windows specific test.

If think that method should work even before the fix. In the peer we set the GC to some value we found, then call Component.setGraphicsConfiguration() which for canvas and window should call getAppropriateGraphicsConfiguration. Why it does not work?

test/jdk/java/awt/Multiscreen/WindowGCChangeTest/WindowGCChangeTest.java
```diff
@@ -27,6 +27,7 @@
   @bug 4868278
   @summary Tests that GraphicsConfig for invisible (peerless) windows is
            updated after showing the window
+  @requires os.family == "windows"
```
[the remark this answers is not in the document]
TejeshR13: Strange, it passes for me too with recent fix. Maybe I messed up with previous fix while testing in linux machine. Will revert it.

test/jdk/java/awt/Window/MultiscreenTransparencyTest.java
TejeshR13: Updated.

TejeshR13:
> > > We should check why that code does not work, note the code in that patch also support both Window and Canvas.
> > 
> > 
> > We can use `getAppropriateGraphicsConfiguration()` of `XCanvasPeer` class, which solves the problem. We can use it whenever new screen is set/new screen GC is set (Not sure of why it wasn't used before). The old code gets defaultGC whenever screen switch happens. We can use the existing implementation, which is actually better than selecting TransparencyCapable GC alone. Their exist a regression for this re-use solution i.e., test [WindowGCChangeTest.java](https://github.com/openjdk/jdk/blob/master/test/jdk/java/awt/Multiscreen/WindowGCChangeTest/WindowGCChangeTest.java), which is basically windows specific test.
> 
> If think that method should work even before the fix. In the peer we set the GC to some value we found, then call Component.setGraphicsConfiguration() which for canvas and window should call getAppropriateGraphicsConfiguration. Why it does not work?

In Window, GC is set when during initialization, i.e., `initGC()` is called, also there is separate `setGraphicsConfiguration() ` method. But unlike `Canvas.setGraphicsConfiguration()`, in window `getAppropriateGraphicsConfiguration()` is not used. So with Window, unlike Canvas setting a GC is quite different and doesn't involve  `getAppropriateGraphicsConfiguration()` I hope.

azvegint (approved):
I assume that the pre-integration testing is green.

mrserb:
> > > > We should check why that code does not work, note the code in that patch also support both Window and Canvas.
> > > 
> > > 
> > > We can use `getAppropriateGraphicsConfiguration()` of `XCanvasPeer` class, which solves the problem. We can use it whenever new screen is set/new screen GC is set (Not sure of why it wasn't used before). The old code gets defaultGC whenever screen switch happens. We can use the existing implementation, which is actually better than selecting TransparencyCapable GC alone. Their exist a regression for this re-use solution i.e., test [WindowGCChangeTest.java](https://github.com/openjdk/jdk/blob/master/test/jdk/java/awt/Multiscreen/WindowGCChangeTest/WindowGCChangeTest.java), which is basically windows specific test.
> > 
> > 
> > If think that method should work even before the fix. In the peer we set the GC to some value we found, then call Component.setGraphicsConfiguration() which for canvas and window should call getAppropriateGraphicsConfiguration. Why it does not work?
> 
> In Window, GC is set when during initialization, i.e., `initGC()` is called, also there is separate `setGraphicsConfiguration() ` method. But unlike `Canvas.setGraphicsConfiguration()`, in window `getAppropriateGraphicsConfiguration()` is not used. So with Window, unlike Canvas setting a GC is quite different and doesn't involve `getAppropriateGraphicsConfiguration()` I hope.

So this bug is not reproduced for Canvas? Then probably we should use the code similar to Canvas.setGraphicsConfiguration in the WIndow class?

TejeshR13:
> > > > > We should check why that code does not work, note the code in that patch also support both Window and Canvas.
> > > > 
> > > > 
> > > > We can use `getAppropriateGraphicsConfiguration()` of `XCanvasPeer` class, which solves the problem. We can use it whenever new screen is set/new screen GC is set (Not sure of why it wasn't used before). The old code gets defaultGC whenever screen switch happens. We can use the existing implementation, which is actually better than selecting TransparencyCapable GC alone. Their exist a regression for this re-use solution i.e., test [WindowGCChangeTest.java](https://github.com/openjdk/jdk/blob/master/test/jdk/java/awt/Multiscreen/WindowGCChangeTest/WindowGCChangeTest.java), which is basically windows specific test.
> > > 
> > > 
> > > If think that method should work even before the fix. In the peer we set the GC to some value we found, then call Component.setGraphicsConfiguration() which for canvas and window should call getAppropriateGraphicsConfiguration. Why it does not work?
> > 
> > 
> > In Window, GC is set when during initialization, i.e., `initGC()` is called, also there is separate `setGraphicsConfiguration() ` method. But unlike `Canvas.setGraphicsConfiguration()`, in window `getAppropriateGraphicsConfiguration()` is not used. So with Window, unlike Canvas setting a GC is quite different and doesn't involve `getAppropriateGraphicsConfiguration()` I hope.
> 
> So this bug is not reproduced for Canvas? Then probably we should use the code similar to Canvas.setGraphicsConfiguration in the WIndow class?

Can't be sure of both. Because `getAppropriateGraphicsConfiguration()` is used only if initialization of GC during `initGC` fails, i.e., `GraphicsEnvironment.getLocalGraphicsEnvironment().getDefaultScreenDevice().getDefaultConfiguration();` fails to load GC. Moreover in this bug we are trying to address the switch over between screen, hence my fix is in Peer component. So adding a similar code in Window class might not be a fix, rather we **might** have to change the way GC is loading from initialization too.

TejeshR13:
> The patch which added the getAppropriateGraphicsConfiguration was supposed to fix the same bug but looks like it was implemented for the Canvas only, I think we can do the same thing for the Window class as well. Note that getAppropriateGraphicsConfiguration on all other platforms is noop so it should not affected.
Hmm, ok then let me work on fixing this. In windows, there is an issue too when we drag a window to another screen like the ellipse becomes an arc. So if  we re using getAppropriateGraphicsConfiguration at setGraphicsConfiguration, that one might as well get solved.

src/java.desktop/share/classes/java/awt/Window.java
```diff
@@ -3215,13 +3215,11 @@ public AccessibleStateSet getAccessibleStateSet() {
...
     @Override
     void setGraphicsConfiguration(GraphicsConfiguration gc) {
-        if (gc == null) {
-            gc = GraphicsEnvironment.
-                    getLocalGraphicsEnvironment().
-                    getDefaultScreenDevice().
-                    getDefaultConfiguration();
-        }
-        synchronized (getTreeLock()) {
+        synchronized(getTreeLock()) {
```
turbanoff, line 3218:
```suggestion
        synchronized (getTreeLock()) {
```

TejeshR13:
@azvegint Can you please re-review the PR and let me know if the updated fix is fine.

src/java.desktop/unix/classes/sun/awt/X11/XWindowPeer.java
```diff
@@ -206,6 +207,39 @@ void preInit(XCreateWindowParams params) {
...
+        XToolkit.awtLock();
+        try {
+            final int screenNum = ((X11GraphicsDevice) gc.getDevice()).getScreen();
+            return (X11GraphicsDevice) GraphicsEnvironment.
+                    getLocalGraphicsEnvironment().
+                    getScreenDevices()[screenNum];
+        } finally {
+            XToolkit.awtUnlock();
+        }
+    }
```
azvegint, line 240: The same code is already in XCanvasPeer.

XWindowPeer is a descendant of XCanvasPeer, there is no need to repeat the code here.

src/java.desktop/windows/classes/sun/awt/windows/WWindowPeer.java
```diff
@@ -662,6 +662,13 @@ boolean isTargetUndecorated() {
      @Override
      public native void repositionSecurityWarning();
 
+    @Override
+    public GraphicsConfiguration getAppropriateGraphicsConfiguration(
+            GraphicsConfiguration gc)
+    {
+        return gc;
+    }
+
```
azvegint, line 671: Same here, WWindowPeer is a WCanvasPeer descendant.

src/java.desktop/share/classes/java/awt/Window.java
```diff
@@ -3215,13 +3215,11 @@ public AccessibleStateSet getAccessibleStateSet() {
 
     @Override
     void setGraphicsConfiguration(GraphicsConfiguration gc) {
-        if (gc == null) {
-            gc = GraphicsEnvironment.
-                    getLocalGraphicsEnvironment().
-                    getDefaultScreenDevice().
-                    getDefaultConfiguration();
-        }
```
azvegint, line 3223: It looks like `gc` was always not null before the fix, which is no longer true after the fix.
We need to double check that we didn't break anything in this regard.

src/java.desktop/macosx/classes/sun/lwawt/macosx/CFileDialog.java
```diff
@@ -199,6 +199,13 @@ public void setTitle(String title) {
     public void repositionSecurityWarning() {
     }
 
+    @Override
+    public GraphicsConfiguration getAppropriateGraphicsConfiguration(
+            GraphicsConfiguration gc)
+    {
```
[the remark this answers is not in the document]
TejeshR13: Updated.

src/java.desktop/unix/classes/sun/awt/X11/XWindowPeer.java
TejeshR13: Updated.

src/java.desktop/windows/classes/sun/awt/windows/WWindowPeer.java
TejeshR13: Updated.

src/java.desktop/share/classes/java/awt/Window.java
TejeshR13: Yeah, have tested in CI multiple time and also few manual dual screen related test. No regression found with fix.

src/java.desktop/unix/classes/sun/awt/X11/XWindowPeer.java
```diff
@@ -61,6 +61,7 @@
 import sun.awt.IconInfo;
 import sun.awt.SunToolkit;
 import sun.awt.X11GraphicsDevice;
+import sun.awt.X11GraphicsConfig;
```
azvegint, line 64: Looks like we don't need this anymore.

azvegint (approved):

TejeshR13:
@mrserb I have updated the PR with Canvas Patch with @azvegint as reviewer. Please let me know if you have anything else about this PR? (CI testing is green).

src/java.desktop/share/classes/java/awt/Window.java
mrserb: We have a couple of places where we call "setGraphicsConfiguration(null)", previously that sets the default GC, now it will set the null, can we try to trigger that codepath and check how it will work?

TejeshR13: Removing a Component will call `Component.setGraphicsConfiguration()`, which in turn triggers `Component.updateGraphicsData()`. So `setGraphicsConfiguration(null)` never invoked `Window.setGraphicsConfiguration()`.

TejeshR13: Anyhow I have tested the updated code in CI system multiple time, never encountered a failure due to null. As part of precautionary measures, let me retain the old code too, along with updated ones.

TejeshR13: @mrserb I have addressed your review comments, Please let me know if anything else is there w.r.t PR? Let me know if the fix is fine.

mrserb (approved):

mrserb:
Looks fine.

PR #14251, opened 2023-05-31
8309200: java/net/httpclient/ExecutorShutdown fails intermittently, if connection closed during upgrade

dfuch:
The ExecutorShutdown test has been observed failing intermittently, notably if by misfortune the shutdown sequence causes a connection to get aborted while upgrading. The issue is that the `ConnectionAborter` class that allows to mark the connection as being scheduled for closing before a handle to the connection is actually available isn't forwarding the original exception for which closing the connection was requested. When the connection is eventually closed, a generic `IOException: connection closed locally` is raised at the `SocketTube` level, which unfortunately can race with the original cause. 

The fix makes it possible to relay the original cause to the place where the IOException is raised, in order to set it as the cause of the new exception.

---------
### Progress

### Issue
 * [JDK-8309200](https://bugs.openjdk.org/browse/JDK-8309200): java/net/httpclient/ExecutorShutdown fails intermittently, if connection closed during upgrade

### Reviewers
 * [Daniel Jeliński](https://openjdk.org/census#djelinski) (@djelinski - **Reviewer**) ⚠️ Review applies to [5fb0731e](https://git.openjdk.org/jdk/pull/14251/files/5fb0731ea45dcedb63319d52e9b55a7a01b6e0f2)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14251/head:pull/14251` \
`$ git checkout pull/14251`

`$ git checkout pull/14251` \
`$ git pull https://git.openjdk.org/jdk.git pull/14251/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14251`

`$ git pr show -t 14251`

</details>

<a href="https://git.openjdk.org/jdk/pull/14251.diff">https://git.openjdk.org/jdk/pull/14251.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14251#issuecomment-1570592812)

### DIFF
--- a/src/java.net.http/share/classes/jdk/internal/net/http/Exchange.java
+++ b/src/java.net.http/share/classes/jdk/internal/net/http/Exchange.java
@@ -164,7 +164,8 @@ final class Exchange<T> {
             HttpConnection connection;
             Throwable cause;
             synchronized (this) {
-                if ((cause = this.cause) == null) {
+                cause = this.cause;
+                if (cause == null) {
                     cause = this.cause = error;
                 }
                 connection = this.connection;
--- a/src/java.net.http/share/classes/jdk/internal/net/http/Exchange.java
+++ b/src/java.net.http/share/classes/jdk/internal/net/http/Exchange.java
@@ -142,6 +142,7 @@ final class Exchange<T> {
     static final class ConnectionAborter {
         private volatile HttpConnection connection;
         private volatile boolean closeRequested;
+        private volatile Throwable cause;
 
         void connection(HttpConnection connection) {
             boolean closeRequested;
@@ -156,20 +157,25 @@ final class Exchange<T> {
                     this.closeRequested = false;
                 }
             }
-            if (closeRequested) closeConnection(connection);
+            if (closeRequested) closeConnection(connection, cause);
         }
 
-        void closeConnection() {
+        void closeConnection(Throwable error) {
             HttpConnection connection;
+            Throwable cause;
             synchronized (this) {
+                if ((cause = this.cause) == null) {
+                    cause = this.cause = error;
+                }
                 connection = this.connection;
                 if (connection == null) {
                     closeRequested = true;
                 } else {
                     this.connection = null;
+                    this.cause = null;
                 }
             }
-            closeConnection(connection);
+            closeConnection(connection, cause);
         }
 
         HttpConnection disable() {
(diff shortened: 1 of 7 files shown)


src/java.net.http/share/classes/jdk/internal/net/http/Exchange.java
```diff
@@ -156,20 +157,25 @@ void connection(HttpConnection connection) {
...
-            if (closeRequested) closeConnection(connection);
+            if (closeRequested) closeConnection(connection, cause);
         }
 
-        void closeConnection() {
+        void closeConnection(Throwable error) {
             HttpConnection connection;
+            Throwable cause;
             synchronized (this) {
+                if ((cause = this.cause) == null) {
```
turbanoff, line 167: let's extract the assignment to a separate statement. Code is a bit hard to read.

src/java.net.http/share/classes/jdk/internal/net/http/Exchange.java
```diff
@@ -156,20 +157,26 @@ void connection(HttpConnection connection) {
...
+                cause = this.cause;
+                if (cause == null) {
+                    cause = this.cause = error;
+                }
                 connection = this.connection;
                 if (connection == null) {
                     closeRequested = true;
                 } else {
                     this.connection = null;
+                    this.cause = null;
```
jaikiran, line 176: Hello Daniel, I had to read this a few times to understand why we are resetting `this.cause` which we potentially set a few lines above to `null`. From what I understand, we only want to set `this.cause` when the `connection` is not `null`. Do you think this code could instead of written as:

```java
Throwable cause;
synchronized (this) {
 cause = this.cause;
 if (cause == null) {
     cause = error;
 }
 connection = this.connection;
 if (connection == null) {
     closeRequested = true;
     this.cause = cause;
 } else {
     this.connection = null;
     this.cause = null;
 }
}
```

jaikiran (approved):
The changes look fine to me. I have a minor comment about the code in the ConnectionAborter, which I've added inline. It's OK with me if you prefer to maintain it in its current form in the PR.

djelinski (approved):

src/java.net.http/share/classes/jdk/internal/net/http/Exchange.java
dfuch: Much better yes. I'll steal it!

jaikiran (approved):
Thank you Daniel for the update. Looks good to me.

PR #14255, opened 2023-06-01
8307887: (fs) Files.createSymbolicLink throws less specific exception when in developer mode and file already exists

bplb:
If the second attempt to create the link fails, throw the exception referring to the last error if it is anything other than due to lack of link creation privilege or failure to recognize the flag for unprivileged link creation.

### DIFF
--- a/src/java.base/windows/classes/sun/nio/fs/WindowsNativeDispatcher.java
+++ b/src/java.base/windows/classes/sun/nio/fs/WindowsNativeDispatcher.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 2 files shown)


src/java.base/windows/classes/sun/nio/fs/WindowsNativeDispatcher.java
```diff
@@ -976,15 +976,22 @@ static void CreateSymbolicLink(String link, String target, int flags)
...
-                    // builds older than 14972.
+                } catch (WindowsException y) {
+                    // Throw an exception if and only if it is not due to symbolic link creation
+                    // privilege not being held (ERROR_PRIVILEGE_NOT_HELD) nor the
+                    // SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE flag not being recognized
+                    // (ERROR_INVALID_PARAMETER). The latter will occur for Windows builds
+                    // older than 14972.
+                    int lastError = y.lastError();
+                    if (lastError != ERROR_PRIVILEGE_NOT_HELD && lastError != ERROR_INVALID_PARAMETER)
+                        throw y;
```
AlanBateman, line 994: The methods that WindowsNativeDispatcher defines are meant to be 1-1 mapping to win32 calls. It looks like we put the changes for JDK-8221852 in the wrong place, they need to be the usage rather than here.

bplb: Restored `WindowsNativeDispatcher` 1:1 mapping for `CreateSymbolcLink`: logic moved to a new static method in `WindowsLinkSupport` (255f5c7f8d86a3d653fb648ed993bb4589792881).

test/jdk/java/nio/file/Files/Links.java
```diff
@@ -48,6 +48,19 @@ static void assertTrue(boolean okay) {
...
+        Path pathTarget = Files.createTempFile("target", null);
+        try {
+            Files.createSymbolicLink(pathLink, pathTarget);
+            throw new RuntimeException("Symbolic link created");
+        } catch (FileAlreadyExistsException expected) {
+        } finally {
+            Files.deleteIfExists(pathLink);
+            Files.deleteIfExists(pathTarget);
+        }
+
```
AlanBateman, line 63: I think it's confusing to have a test for sym links before testing if sym links are supported. I think it's okay to not include tests that target very specific cases like Windows "developer mode" but a test that the optional specific exception is thrown is okay, it just feels like it should be just one of the tests, rather than done before knowing the test can run, e.g. try this on a non-NTFS system and I assume createSymbolicLink will fail and cause the test to fail.

bplb: Logically I don't like the ordering either but without this happening first the error was not reproduced on the pre-PR code.

bplb: Perhaps it would be better to revert the test change altogether and label the issue `noreg-hard`?

AlanBateman (approved):

PR #14259, opened 2023-06-01
8281455: Change JVM options with small ranges from 64 to 32 bits, for gc_globals.hpp

afshin-zafari:
The `uintx/intx` options whose ranges are small enough are changed to `uint/int`, otherwise gcc complains 
when `-Wconversion` is used in build.
Their uses in printf formats are changed accordingly.

### Tests
Locally hotspot:tier1 tested on linux-x64
mach5 tiers 1-4 on Linux and Windows 64bits platforms passed.

---------
### Progress

### Issue
 * [JDK-8281455](https://bugs.openjdk.org/browse/JDK-8281455): Change JVM options with small ranges from 64 to 32 bits, for gc_globals.hpp (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14259/head:pull/14259` \
`$ git checkout pull/14259`

`$ git checkout pull/14259` \
`$ git pull https://git.openjdk.org/jdk.git pull/14259/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14259`

`$ git pr show -t 14259`

</details>

<a href="https://git.openjdk.org/jdk/pull/14259.diff">https://git.openjdk.org/jdk/pull/14259.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14259#issuecomment-1580237728)

### DIFF
--- a/src/hotspot/share/gc/g1/g1Arguments.cpp
+++ b/src/hotspot/share/gc/g1/g1Arguments.cpp
@@ -229,7 +229,7 @@ void G1Arguments::initialize() {
 
   // By default do not let the target stack size to be more than 1/4 of the entries
   if (FLAG_IS_DEFAULT(GCDrainStackTargetSize)) {
-    FLAG_SET_ERGO(GCDrainStackTargetSize, MIN2(GCDrainStackTargetSize, (uint)TASKQUEUE_SIZE / 4));
+    FLAG_SET_ERGO(GCDrainStackTargetSize, MIN2(GCDrainStackTargetSize, TASKQUEUE_SIZE / 4));
   }
 
 #ifdef COMPILER2
--- a/src/hotspot/share/gc/g1/g1Arguments.cpp
+++ b/src/hotspot/share/gc/g1/g1Arguments.cpp
@@ -229,7 +229,7 @@ void G1Arguments::initialize() {
 
   // By default do not let the target stack size to be more than 1/4 of the entries
   if (FLAG_IS_DEFAULT(GCDrainStackTargetSize)) {
-    FLAG_SET_ERGO(GCDrainStackTargetSize, MIN2(GCDrainStackTargetSize, (uintx)TASKQUEUE_SIZE / 4));
+    FLAG_SET_ERGO(GCDrainStackTargetSize, MIN2(GCDrainStackTargetSize, (uint)TASKQUEUE_SIZE / 4));
   }
 
 #ifdef COMPILER2
--- a/src/hotspot/share/gc/parallel/jvmFlagConstraintsParallel.cpp
+++ b/src/hotspot/share/gc/parallel/jvmFlagConstraintsParallel.cpp
@@ -42,11 +42,11 @@ JVMFlag::Error ParallelGCThreadsConstraintFuncParallel(uint value, bool verbose)
   return JVMFlag::SUCCESS;
 }
 
-JVMFlag::Error InitialTenuringThresholdConstraintFuncParallel(uintx value, bool verbose) {
+JVMFlag::Error InitialTenuringThresholdConstraintFuncParallel(uint value, bool verbose) {
   // InitialTenuringThreshold is only used for ParallelGC.
   if (UseParallelGC && (value > MaxTenuringThreshold)) {
       JVMFlag::printError(verbose,
-                          "InitialTenuringThreshold (" UINTX_FORMAT ") must be "
+                          "InitialTenuringThreshold (%u) must be "
                           "less than or equal to MaxTenuringThreshold (%u)\n",
                           value, MaxTenuringThreshold);
       return JVMFlag::VIOLATES_CONSTRAINT;
@@ -54,12 +54,12 @@ JVMFlag::Error InitialTenuringThresholdConstraintFuncParallel(uintx value, bool 
   return JVMFlag::SUCCESS;
 }
 
-JVMFlag::Error MaxTenuringThresholdConstraintFuncParallel(uintx value, bool verbose) {
+JVMFlag::Error MaxTenuringThresholdConstraintFuncParallel(uint value, bool verbose) {
   // As only ParallelGC uses InitialTenuringThreshold,
   // we don't need to compare InitialTenuringThreshold with MaxTenuringThreshold.
   if (UseParallelGC && (value < InitialTenuringThreshold)) {
     JVMFlag::printError(verbose,
-                        "MaxTenuringThreshold (" UINTX_FORMAT ") must be "
+                        "MaxTenuringThreshold (%u) must be "
                         "greater than or equal to InitialTenuringThreshold (%u)\n",
                         value, InitialTenuringThreshold);
     return JVMFlag::VIOLATES_CONSTRAINT;
--- a/src/hotspot/share/gc/parallel/jvmFlagConstraintsParallel.cpp
+++ b/src/hotspot/share/gc/parallel/jvmFlagConstraintsParallel.cpp
@@ -47,7 +47,7 @@ JVMFlag::Error InitialTenuringThresholdConstraintFuncParallel(uintx value, bool 
   if (UseParallelGC && (value > MaxTenuringThreshold)) {
       JVMFlag::printError(verbose,
                           "InitialTenuringThreshold (" UINTX_FORMAT ") must be "
-                          "less than or equal to MaxTenuringThreshold (" UINTX_FORMAT ")\n",
+                          "less than or equal to MaxTenuringThreshold (%u)\n",
                           value, MaxTenuringThreshold);
       return JVMFlag::VIOLATES_CONSTRAINT;
   }
@@ -60,7 +60,7 @@ JVMFlag::Error MaxTenuringThresholdConstraintFuncParallel(uintx value, bool verb
   if (UseParallelGC && (value < InitialTenuringThreshold)) {
     JVMFlag::printError(verbose,
                         "MaxTenuringThreshold (" UINTX_FORMAT ") must be "
-                        "greater than or equal to InitialTenuringThreshold (" UINTX_FORMAT ")\n",
+                        "greater than or equal to InitialTenuringThreshold (%u)\n",
                         value, InitialTenuringThreshold);
     return JVMFlag::VIOLATES_CONSTRAINT;
   }
--- a/src/hotspot/share/gc/parallel/psAdaptiveSizePolicy.cpp
+++ b/src/hotspot/share/gc/parallel/psAdaptiveSizePolicy.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2002, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2002, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
--- a/src/hotspot/share/gc/parallel/psAdaptiveSizePolicy.cpp
+++ b/src/hotspot/share/gc/parallel/psAdaptiveSizePolicy.cpp
@@ -349,7 +349,7 @@ void PSAdaptiveSizePolicy::compute_eden_space_size(
     log_debug(gc, ergo)(
           "PSAdaptiveSizePolicy::compute_eden_space_size: gc time limit"
           " gc_cost: %f "
-          " GCTimeLimit: " UINTX_FORMAT,
+          " GCTimeLimit: " UINT32_FORMAT,
           gc_cost(), GCTimeLimit);
   }
 
@@ -526,7 +526,7 @@ void PSAdaptiveSizePolicy::compute_old_gen_free_space(
     log_debug(gc, ergo)(
           "PSAdaptiveSizePolicy::compute_old_gen_free_space: gc time limit"
           " gc_cost: %f "
-          " GCTimeLimit: " UINTX_FORMAT,
+          " GCTimeLimit: " UINT32_FORMAT,
           gc_cost(), GCTimeLimit);
   }
 
--- a/src/hotspot/share/gc/parallel/psAdaptiveSizePolicy.cpp
+++ b/src/hotspot/share/gc/parallel/psAdaptiveSizePolicy.cpp
@@ -526,7 +526,7 @@ void PSAdaptiveSizePolicy::compute_old_gen_free_space(
     log_debug(gc, ergo)(
           "PSAdaptiveSizePolicy::compute_old_gen_free_space: gc time limit"
           " gc_cost: %f "
-          " GCTimeLimit: " UINT32_FORMAT,
+          " GCTimeLimit: %u",
           gc_cost(), GCTimeLimit);
   }
 
--- a/src/hotspot/share/gc/parallel/psAdaptiveSizePolicy.cpp
+++ b/src/hotspot/share/gc/parallel/psAdaptiveSizePolicy.cpp
@@ -349,7 +349,7 @@ void PSAdaptiveSizePolicy::compute_eden_space_size(
     log_debug(gc, ergo)(
           "PSAdaptiveSizePolicy::compute_eden_space_size: gc time limit"
           " gc_cost: %f "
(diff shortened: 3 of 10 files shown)


tschatzl:
Although it's a bit early to review, when I looked at this briefly, I believe also the `uintx` that have a `juint_max` upper range can be replaced with `uint` (and use the equivalent `UINT_MAX` as upper range, `j`-types should only be used with JNI code).

afshin-zafari:
> Although it's a bit early to review, when I looked at this briefly, I believe also the `uintx` that have a `juint_max` upper range can be replaced with `uint` (and use the equivalent `UINT_MAX` as upper range, `j`-types should only be used with JNI code).

Done.

offamitkumar:
Wouldn't that be good to keep the alignment for `\` symbol in `gc_global.hpp` file ?

src/hotspot/share/gc/parallel/psAdaptiveSizePolicy.cpp
```diff
@@ -349,7 +349,7 @@ void PSAdaptiveSizePolicy::compute_eden_space_size(
     log_debug(gc, ergo)(
           "PSAdaptiveSizePolicy::compute_eden_space_size: gc time limit"
           " gc_cost: %f "
-          " GCTimeLimit: " UINTX_FORMAT,
+          " GCTimeLimit: " UINT32_FORMAT,
```
tschatzl, line 352: We use `%u` as format specifiers for uint.

src/hotspot/share/gc/shared/gcOverheadChecker.cpp
```diff
@@ -91,11 +91,11 @@ void GCOverheadChecker::check_gc_overhead_limit(GCOverheadTester* time_overhead,
 
   if (UseGCOverheadLimit) {
     if (gc_overhead_limit_exceeded()) {
-      log_trace(gc, ergo)("GC is exceeding overhead limit of " UINTX_FORMAT "%%", GCTimeLimit);
+      log_trace(gc, ergo)("GC is exceeding overhead limit of %d%%", GCTimeLimit);
```
tschatzl, line 94: Not sure why this place uses `%d` as format specifier compared to the `UINT32_FORMAT` the previous places did. Use `%u` ;)

src/hotspot/share/gc/shared/gc_globals.hpp
```diff
@@ -166,7 +166,7 @@
           "A System.gc() request invokes a concurrent collection; "         \
           "(effective only when using concurrent collectors)")              \
                                                                             \
-  product(uintx, GCLockerEdenExpansionPercent, 5,                           \
+  product(uint, GCLockerEdenExpansionPercent, 5,                           \
```
tschatzl, line 169: The `\` needs to be aligned back.

src/hotspot/share/utilities/globalDefinitions.hpp
```diff
@@ -431,13 +431,15 @@ const intx  min_intx  = (intx)1 << (sizeof(intx)*BitsPerByte-1);
...
 
+
 // Table of values:
 //      sizeof intx         4               8
 // min_intx             0x80000000      0x8000000000000000
 // max_intx             0x7FFFFFFF      0x7FFFFFFFFFFFFFFF
 // max_uintx            0xFFFFFFFF      0xFFFFFFFFFFFFFFFF
 
 typedef unsigned int uint;   NEEDS_CLEANUP
+const uint  max_uint  = (uint)-1;
```
tschatzl, line 442: The change should use the existing `UINT_MAX` instead of adding a new constant. Instead of a cast of `-1`, one should also use `~0` to get all-ones.

tschatzl (changes requested):
More candidates:

ParGCArrayScanChunk
MarkStackSize(Max) (these should probably be a `size_t`)
VerifyArchivedFields
ScavengeALotInterval
FullGCALotInterval
FullGCALotStart
FullGCALotDummies
VerifyGCLevel
MaxTenuringThreshold
InitialTenuringThreshold
GCDrainStackTargetSize

src/hotspot/share/gc/parallel/psAdaptiveSizePolicy.cpp
afshin-zafari: `%u` is used.

src/hotspot/share/utilities/globalDefinitions.hpp
afshin-zafari: `UINT_MAX` is used. New definition of max_uint is removed.

afshin-zafari (commented):
Suggested changes in Comments are applied including the 'more candidates'.

src/hotspot/share/gc/g1/g1Arguments.cpp
```diff
@@ -229,7 +229,7 @@ void G1Arguments::initialize() {
 
   // By default do not let the target stack size to be more than 1/4 of the entries
   if (FLAG_IS_DEFAULT(GCDrainStackTargetSize)) {
-    FLAG_SET_ERGO(GCDrainStackTargetSize, MIN2(GCDrainStackTargetSize, (uintx)TASKQUEUE_SIZE / 4));
+    FLAG_SET_ERGO(GCDrainStackTargetSize, MIN2(GCDrainStackTargetSize, (uint)TASKQUEUE_SIZE / 4));
```
kimbarrett, line 232: TASKQUEUE_SIZE should be a constant of type size_t rather than a macro.  GCDrainStackTargetSize should be of type size_t.  Then we don't need any casts here.  But some uses may need cleaning up, either to remove casts or to adjust surrounding types.  I suggest this be split out as a separate RFE to avoid cluttering the simplicity of this PR.

src/hotspot/share/gc/parallel/psScavenge.cpp
```diff
@@ -553,7 +553,7 @@ bool PSScavenge::invoke_no_policy() {
                                                            _tenuring_threshold,
                                                            survivor_limit);
 
-       log_debug(gc, age)("Desired survivor size " SIZE_FORMAT " bytes, new threshold %u (max threshold " UINTX_FORMAT ")",
+       log_debug(gc, age)("Desired survivor size " SIZE_FORMAT " bytes, new threshold %u (max threshold %u)",
```
kimbarrett, line 556: Consider also, instead of SIZE_FORMAT we can use "%zu".

src/hotspot/share/gc/shared/ageTable.cpp
```diff
@@ -99,7 +99,7 @@ uint AgeTable::compute_tenuring_threshold(size_t desired_survivor_size) {
   }
 
 
-  log_debug(gc, age)("Desired survivor size " SIZE_FORMAT " bytes, new threshold " UINTX_FORMAT " (max threshold " UINTX_FORMAT ")",
+  log_debug(gc, age)("Desired survivor size " SIZE_FORMAT " bytes, new threshold " UINTX_FORMAT " (max threshold %u)",
```
kimbarrett, line 102: Consider also, instead of SIZE_FORMAT we can use "%zu".

src/hotspot/share/gc/g1/g1Arguments.cpp
afshin-zafari: The change is applied and some cleanup made. Maybe no need to a new RFE, if you agree.

src/hotspot/share/gc/parallel/jvmFlagConstraintsParallel.cpp
```diff
@@ -47,7 +47,7 @@ JVMFlag::Error InitialTenuringThresholdConstraintFuncParallel(uintx value, bool
   if (UseParallelGC && (value > MaxTenuringThreshold)) {
       JVMFlag::printError(verbose,
                           "InitialTenuringThreshold (" UINTX_FORMAT ") must be "
-                          "less than or equal to MaxTenuringThreshold (" UINTX_FORMAT ")\n",
+                          "less than or equal to MaxTenuringThreshold (%u)\n",
```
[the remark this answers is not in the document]
afshin-zafari: Fixed.

src/hotspot/share/gc/parallel/psScavenge.cpp
afshin-zafari: `%zu` is used instead.

src/hotspot/share/gc/shared/ageTable.cpp
afshin-zafari: `%zu` is used.

src/hotspot/share/utilities/globalDefinitions.hpp
```diff
@@ -431,6 +431,7 @@ const intx  min_intx  = (intx)1 << (sizeof(intx)*BitsPerByte-1);
 const intx  max_intx  = (uintx)min_intx - 1;
 const uintx max_uintx = (uintx)-1;
 
+
```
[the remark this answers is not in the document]
afshin-zafari: Fixed.

src/hotspot/share/utilities/globalDefinitions.hpp
```diff
@@ -439,7 +440,6 @@ const uintx max_uintx = (uintx)-1;
 
 typedef unsigned int uint;   NEEDS_CLEANUP
```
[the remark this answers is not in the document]
afshin-zafari: Maybe, it is because the `uint` is already defined within the guarded lines 149-154: 
```C++
#ifdef __USE_MISC
/* Old compatibility names for C types.  */
typedef unsigned long int ulong;
typedef unsigned short int ushort;
typedef unsigned int uint;
#endif
```

coleenp (commented):
This looks like a good change to me to avoid Wconversion warnings for mis-sized argument values but I'll leave it up to the GC experts.

src/hotspot/share/gc/shared/gc_globals.hpp
```diff
@@ -212,12 +212,12 @@
   /* where does the range max value of (max_jint - 1) come from? */         \
   product(size_t, MarkStackSizeMax, NOT_LP64(4*M) LP64_ONLY(512*M),         \
           "Maximum size of marking stack")                                  \
-          range(1, (max_jint - 1))                                          \
+          range(1, (SIZE_MAX - 1))                                          \
```
kimbarrett, line 215: This is a large increase in the max size.  Not at all sure it's appropriate.  Similarly for MarkStackSize below.

src/hotspot/share/gc/shared/gc_globals.hpp
```diff
@@ -677,15 +677,15 @@
...
                                                                             \
   develop(uintx, GCExpandToAllocateDelayMillis, 0,                          \
           "Delay between expansion and allocation (in milliseconds)")       \
                                                                             \
-  product(uintx, GCDrainStackTargetSize, 64,                                \
+  product(size_t, GCDrainStackTargetSize, 64,                               \
           "Number of entries we will try to leave on the stack "            \
           "during parallel gc")                                             \
-          range(0, max_juint)                                               \
+          range(0, SIZE_MAX)                                                \
```
kimbarrett, line 688: This is a large increase in the max on 64bit platforms.

kimbarrett (changes requested):

kimbarrett:
> > There are a lot of uses of the options that have casts, many of which ought to be removable with this change. Some of those might pull more tangled strings, but it would be nice if there was a plan to look at those too.
> 
> During this change, I changed all the cases that I found them to be changed. Would you please, give some examples or some 'search criteria' that I can use to find the cases. Then, I will do them here in this PR or create a new RFE for them. 

I spotted some when looking earlier, but haven't re-found any of them.

src/hotspot/share/gc/shared/gc_globals.hpp
```diff
@@ -382,52 +382,52 @@
...
-  product(uintx, AdaptiveTimeWeight,       25,                              \
+  product(uint, AdaptiveTimeWeight,       25,                               \
           "Weight given to time in adaptive policy, between 0 and 100")     \
           range(0, 100)                                                     \
                                                                             \
-  product(uintx, PausePadding, 1,                                           \
+  product(uint, PausePadding, 1,                                            \
           "How much buffer to keep for pause time")                         \
-          range(0, max_juint)                                               \
+          range(0, UINT_MAX)                                                \
```
coleenp, line 395: Why did you change from max_juint -> UINT_MAX ?  max_juint should have compile without integer size mismatch.

src/hotspot/share/gc/shared/gc_globals.hpp
coleenp: Yes, you should revert the change from max_juint to SIZE_MAX.  Is the change to size_t a cleanup to remove a cast?

src/hotspot/share/utilities/globalDefinitions.hpp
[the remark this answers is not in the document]
coleenp: I was curious and found it in /usr/include/x86_64-linux-gnu/sys/types.h on ubuntu. I guess it needs cleanup because it might not be defined with USE_MISC?  There are too many int type names to choose from for int.  Why does it say "Old compatibility names for C types."? Should we even be using uint ? I don't see it in https://en.cppreference.com/w/cpp/language/types - should we prefer "unsigned"?

src/hotspot/share/gc/shared/gc_globals.hpp
coleenp: Or, not in this change, but should we prefer UINT_MAX (found in /usr/include/limits.h) to our globalDefinitions.hpp definitions?

coleenp (commented):
I have a few comments.  Narrowing the types of these arguments was to prevent Wconversion warnings where they're used and then later narrowed to 32 bit ints.  From my sampling of some of the arguments changed, there were no casts, which was the problem.  Some of these (sample GCTimeLimit) are promoted back up to uintx but having the field be the narrowest representation allows the code to use it without an implicit conversion to a narrower int type.  So I think this is a good step for cleaning up the int types.

src/hotspot/share/gc/shared/gc_globals.hpp
coleenp: Now I see Thomas's comments about this.  I guess changing a couple at a time would be ok.

src/hotspot/share/gc/shared/gcOverheadChecker.cpp
```diff
@@ -91,11 +91,11 @@ void GCOverheadChecker::check_gc_overhead_limit(GCOverheadTester* time_overhead,
 
   if (UseGCOverheadLimit) {
     if (gc_overhead_limit_exceeded()) {
-      log_trace(gc, ergo)("GC is exceeding overhead limit of " UINTX_FORMAT "%%", GCTimeLimit);
+      log_trace(gc, ergo)("GC is exceeding overhead limit of %u%%", GCTimeLimit);
       reset_gc_overhead_limit_count();
     } else if (print_gc_overhead_limit_would_be_exceeded) {
       assert(_gc_overhead_limit_count > 0, "Should not be printing");
-      log_trace(gc, ergo)("GC would exceed overhead limit of " UINTX_FORMAT "%% %d consecutive time(s)",
+      log_trace(gc, ergo)("GC would exceed overhead limit of %u %% %d consecutive time(s)",
```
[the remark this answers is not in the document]
afshin-zafari: Fixed.

src/hotspot/share/gc/shared/gc_globals.hpp
afshin-zafari: Reverted back to original.

src/hotspot/share/utilities/globalDefinitions.hpp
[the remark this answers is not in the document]
coleenp: This NEEDS_CLEANUP block shouldn't be part of this change.

tschatzl (approved):
Lgtm, it's a step forward to cleaning up the gc_globals.hpp types/ranges.

coleenp (approved):
This looks good and definitely a step forward.

afshin-zafari:
Thank you Coleen, Kim, Thomas and Amit for your comments and reviewing this PR.
/integrate

PR #14262, opened 2023-06-01
6212244: StringTokenizer should implement Enumeration<String>

amarkevich:
Enumeration implemented using existing hasMoreTokens()/nextToken() and have no additional logic [1]. Existing implementations can safely remove overridden methods [2].

1. https://github.com/apache/commons-collections/blob/COLLECTIONS_3_2_X/src/java/org/apache/commons/collections/ExtendedProperties.java#L403
2. https://github.com/eclipse/jetty.project/commit/33976b30eb749dc075a8cc5f52fdb1f03b0276d5

---------
### Progress

### Error
&nbsp;⚠️ OCA signatory status must be verified

### Issue
 * [JDK-6212244](https://bugs.openjdk.org/browse/JDK-6212244): StringTokenizer should implement Enumeration&lt;String&gt; (**Bug** - P4) ⚠️ Issue is not open.

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14262/head:pull/14262` \
`$ git checkout pull/14262`

`$ git checkout pull/14262` \
`$ git pull https://git.openjdk.org/jdk.git pull/14262/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14262`

`$ git pr show -t 14262`

</details>

<a href="https://git.openjdk.org/jdk/pull/14262.diff">https://git.openjdk.org/jdk/pull/14262.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/util/StringTokenizer.java
+++ b/src/java.base/share/classes/java/util/StringTokenizer.java
@@ -98,7 +98,7 @@ import java.lang.*;
  * @see     java.io.StreamTokenizer
  * @since   1.0
  */
-public class StringTokenizer implements Enumeration<Object> {
+public class StringTokenizer implements Enumeration<String> {
     private int currentPosition;
     private int newPosition;
     private int maxPosition;
@@ -387,15 +387,13 @@ public class StringTokenizer implements Enumeration<Object> {
      * @see     java.util.Enumeration
      * @see     java.util.StringTokenizer#hasMoreTokens()
      */
-    public boolean hasMoreElements() {
+    public final boolean hasMoreElements() {
         return hasMoreTokens();
     }
 
     /**
-     * Returns the same value as the {@code nextToken} method,
-     * except that its declared return value is {@code Object} rather than
-     * {@code String}. It exists so that this class can implement the
-     * {@code Enumeration} interface.
+     * Returns the same value as the {@code nextToken} method. It exists so
+     * that this class can implement the {@code Enumeration} interface.
      *
      * @return     the next token in the string.
      * @throws     NoSuchElementException  if there are no more tokens in this
@@ -403,7 +401,7 @@ public class StringTokenizer implements Enumeration<Object> {
      * @see        java.util.Enumeration
      * @see        java.util.StringTokenizer#nextToken()
      */
-    public Object nextElement() {
+    public final String nextElement() {
         return nextToken();
     }
 


SirYwell:
As the issue was marked as Won't Fix in 2005 as it would break existing code, I doubt this will be accepted.

amarkevich:
> added a final

'final' added because of binary incompatibility after this change. Changing type parameter from Object to String give compilation error anyway; so 'final' make sense to inform implementers about existing default implementation

amarkevich:
> Won't Fix in 2005
just highlight this bad design (IMHO), because got a error 
`Type mismatch: cannot convert from ArrayList<Object> to List<String>`
for
`List<String> result = Collections.list(new StringTokenizer(commaSeparatedString, ", ");`

liach:
> > Won't Fix in 2005
> 
> just highlight this bad design (IMHO), because got a error `Type mismatch: cannot convert from ArrayList<Object> to List<String>` for `List<String> result = Collections.list(new StringTokenizer(commaSeparatedString, ", ");`

From the API specification of StringTokenizer:
```
{@code StringTokenizer} is a legacy class that is retained for
compatibility reasons although its use is discouraged in new code. It is
recommended that anyone seeking this functionality use the {@code split}
method of {@code String} or the java.util.regex package instead.
```
You should use `List<String> result = Arrays.asList(commaSeparatedString.split(", "));` instead.

amarkevich:
> List<String> result = Arrays.asList(commaSeparatedString.split(", "));

Result is not the same for `String commaSeparatedString = " 1 , 2 , 3";`: `[1, 2, 3]` vs `[ 1 , 2 , 3]`

liach:
> > List result = Arrays.asList(commaSeparatedString.split(", "));
> 
> Result is not the same for `String commaSeparatedString = " 1 , 2 , 3";`: `[1, 2, 3]` vs `[ 1 , 2 , 3]`

You should be using this then:
```java
List<String> result = Arrays.asList(commaSeparatedString.split(","));
result.replaceAll(String::strip);
```
If you want to remove spaces.

Or a more regex-heavy version that matches comma with any number of spaces around it and initial/final spaces:
```java
List<String> result = new ArrayList<>(Arrays.asList(commaSeparatedString.split("^( )*|( )*,( )*|( )*$")));
if (!result.isEmpty() && result.get(0).isBlank()) result.remove(0);
if (!result.isEmpty() && result.get(result.size() - 1).isBlank()) result.remove(result.size() - 1);
```
Which removes the empty parts generated between the string beginning/ending and initial/final spaces.

amarkevich:
> regex-heavy version

produce wrong result `[, 1, 2, 3]`

For split + strip I created a small benchmark https://gist.github.com/amarkevich/01f6e9aa6df874a353cafe67125b70f8

```
StringTokenizer: 3 MB 5 ms
split + strip: 4 MB 11 ms
```
StringTokenizer runs faster, consumes less memory and can be written in one line

liach:
> produce wrong result [, 1, 2, 3]

Because you ignored the 2 lines below.

StringTokenizer is naturally faster due to its lack or regex, but it's not a reason to adapt it to a legacy interface that slows it down. You should use an array instead of any list if you really care that much about performance.

Your benchmark should be written in JMH, which takes care of unintended hotspot optimizations that may skew the results.

I recommend you to withdraw this patch. This change is backward incompatible and adds little functionality, as Enumeration is rarely used nowadays. Even a refit to implement iterator may makes more sense: they can be used by enhanced for loops.

PR #14266, opened 2023-06-01
8280994: [XWayland] Drag and Drop does not work in java -> wayland app direction

azvegint:
This change resolves the issue that drag and drop from a Java application to a native Wayland application (such as gedit) is not possible on Linux.

Our code it quite picky for the drop target, and requires it to be a top level window with the WM_STATE property set, whereas the XWayland server provides dummy windows without this property.
So now we make an exclusion for this case.

After the fix it successfully passes previously failed related manual jtreg/conformance tests on Wayland.
Other tests are also passed.

---------
### Progress

### Issue
 * [JDK-8280994](https://bugs.openjdk.org/browse/JDK-8280994): [XWayland] Drag and Drop does not work in java -&gt; wayland app direction

### Reviewers
 * [Phil Race](https://openjdk.org/census#prr) (@prrace - **Reviewer**) ⚠️ Review applies to [ec4493f5](https://git.openjdk.org/jdk/pull/14266/files/ec4493f5273746fdbc2a9c9451c15050d04730d2)
 * [Prasanta Sadhukhan](https://openjdk.org/census#psadhukhan) (@prsadhuk - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14266/head:pull/14266` \
`$ git checkout pull/14266`

`$ git checkout pull/14266` \
`$ git pull https://git.openjdk.org/jdk.git pull/14266/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14266`

`$ git pr show -t 14266`

</details>

<a href="https://git.openjdk.org/jdk/pull/14266.diff">https://git.openjdk.org/jdk/pull/14266.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14266#issuecomment-1572266650)

### DIFF
--- a/src/java.desktop/unix/classes/sun/awt/X11/XDragSourceContextPeer.java
+++ b/src/java.desktop/unix/classes/sun/awt/X11/XDragSourceContextPeer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2003, 2015, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2003, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -27,6 +27,7 @@ package sun.awt.X11;
 
 import java.awt.Component;
 import java.awt.Cursor;
+import java.awt.Toolkit;
 import java.awt.Window;
 
 import java.awt.datatransfer.DataFlavor;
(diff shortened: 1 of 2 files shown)


src/java.desktop/unix/classes/sun/awt/X11/XDragSourceContextPeer.java
```diff
@@ -392,6 +393,41 @@ private boolean updateSourceAction(int state) {
...
+     * This is not true when performing drag and drop from XWayland
+     * to a native Wayland application.
+     * In this case XWayland creates a dummy window with only one property,
+     * XdndAware.
+     *
+     * @param window to test
+     * @return true if window has XdndAware property when running under Wayland
+     */
+    private static boolean isXWaylandDndAwareWindow(long window) {
+        if (Toolkit.getDefaultToolkit() instanceof SunToolkit sunToolkit
```
prrace, line 408: So if it is NOT an instance of sunToolkit .. then you carry on to the code that gets the property.
Is that what you meant to happen ?

azvegint: Fixed.

src/java.desktop/unix/classes/sun/awt/X11/XDragSourceContextPeer.java
```diff
@@ -392,6 +393,40 @@ private boolean updateSourceAction(int state) {
...
+    private static boolean isXWaylandDndAwareWindow(long window) {
+        Toolkit toolkit = Toolkit.getDefaultToolkit();
+        if (!(toolkit instanceof SunToolkit)
+                || !((SunToolkit) toolkit).isRunningOnWayland()) {
+            return false;
+        }
+
+        WindowPropertyGetter wpg =
+            new WindowPropertyGetter(window, XDnDConstants.XA_XdndAware, 0, 1,
+                    false, XConstants.AnyPropertyType);
```
prsadhuk, line 416: Minor nit: I guess "false" should be aligned below "window"

src/java.desktop/unix/classes/sun/awt/X11/XDragSourceContextPeer.java
```diff
@@ -408,6 +445,10 @@ private static long findClientWindow(long window) {
             }
         }
 
+        if (isXWaylandDndAwareWindow(window)) {
```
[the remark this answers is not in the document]
prsadhuk: Yes, since it's a recursive function, I was just trying to understand if `isXWaylandDndAwareWindow `is called at right place
I tried a simple recursive program similar to `findClientWindow `which gives StackOverflowError so I was thinking if `findClientWindow  `can also give SOF and whether we should call `isXWaylandDndAwareWindow ` at l446 or l439

```
public class Recursion {
        long value = 22;
        System.out.println(test(value));
    }

    private static long test(long value) {
        long arr[] = {1L,2L,3L};
        for (long i : arr) {
            long val = test(i);
            if (val != 0) {
                System.out.println("val " + val);
                return val;
            }
        }
        if (getValue(value) == 22) {
            return value;
        }
        return 0;
    }
    private static long getValue(long value) {
        return value;
    }
}
```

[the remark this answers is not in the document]
azvegint: Yes, this could happen if the implementation changes, currently the dummy window has no children.

Updated.

prsadhuk (approved):

PR #14267, opened 2023-06-01
8307840: SequencedMap view method specification and implementation adjustments

stuart-marks:


src/java.base/share/classes/java/util/SequencedMap.java
```diff
@@ -257,45 +257,56 @@ default V putLast(K k, V v) {
...
-     * implementation that delegates all operations either to this map or to this map's
-     * {@link #keySet}, except for its {@link SequencedSet#reversed reversed} method,
-     * which instead returns the result of calling {@code sequencedKeySet} on this map's
-     * reverse-ordered view.
+     * The implementation in this interface returns a {@code SequencedSet} instance
+     * that behaves as follows. Its {@link SequencedSet#add add} and {@link
+     * SequencedSet#addAll addAll} methods throw {@link UnsupportedOperationException}.
+     * Its {@link SequencedSet#reversed reversed} method returns the {@link
+     * #sequencedKeySet sequencedKeySet} view of the {@link #reversed reversed} view of
+     * this map. Each of its other methods calls the corresponding method of the {@link
```
jaikiran, line 268: Hello Stuart, should this (and other similar places in this PR) instead be:
```
Each of its other methods call
```
call instead of calls?

src/java.base/share/classes/java/util/AbstractMap.java
```diff
@@ -893,35 +893,35 @@ public String toString() {
...
-        public boolean removeAll(Collection<?> c) { return view.removeAll(c); }
-        public boolean removeIf(Predicate<? super E> filter) { return view.removeIf(filter); }
-        public boolean retainAll(Collection<?> c) { return view.retainAll(c); }
-        public int size() { return view.size(); }
-        public Spliterator<E> spliterator() { return view.spliterator(); }
-        public Stream<E> stream() { return view.stream(); }
-        public Object[] toArray() { return view.toArray(); }
-        public <T> T[] toArray(IntFunction<T[]> generator) { return view.toArray(generator); }
-        public <T> T[] toArray(T[] a) { return view.toArray(a); }
+        public void clear() { view().clear(); }
```
jaikiran, line 908: Each of these methods should perhaps have the `@Override` to make it clear that these correspond to the interface this class is implementing. But I see that the `AbstractMap` class itself doesn't use `@Override` anywhere, so I think it's OK to keep it consistent in its current form in this PR.

jddarcy (approved):

AlanBateman (approved):

src/java.base/share/classes/java/util/AbstractMap.java
stuart-marks: Yeah the collections classes are inconsistent about using `@Override`. Some maintainers (Martin Buchholz, Doug Lea) prefer not to use it at all. However, some usage has crept in from time to time. Personally I tend not to use it, as I think it adds a lot of clutter and doesn't provide much value. The main value this annotation provides detecting an incorrectly declared override (e.g., declaring an equals(Foo) method instead of equals(Object) method). In practice this doesn't seem to happen. These skeletons of these internal collection classes are usually auto-generated by an IDE, which won't make this mistake. The collections are usually pretty well tested so if something isn't overridden properly it will usually be caught.

I'm considering ripping out all occurrences of `@Override` from collections and declaring the prevailing style not to use it.

src/java.base/share/classes/java/util/SequencedMap.java
stuart-marks: "Each" is usually considered singular so "Each ... calls" is proper. It looks weird though since the plural "methods" is right next to "calls."

PR #14271, opened 2023-06-01
8308591: JLine as the default Console provider

naotoj:
This is to bring the JLine-based Console implementation as the default, which was the initial intention in [JDK-8295803](https://bugs.openjdk.org/browse/JDK-8295803). A corresponding CSR has also been drafted.

---------
### Progress
- [x] Change requires CSR request [JDK-8309141](https://bugs.openjdk.org/browse/JDK-8309141) to be approved

### Issues
 * [JDK-8308591](https://bugs.openjdk.org/browse/JDK-8308591): JLine as the default Console provider (**Enhancement** - P4)
 * [JDK-8309141](https://bugs.openjdk.org/browse/JDK-8309141): JLine as the default Console provider (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14271/head:pull/14271` \
`$ git checkout pull/14271`

`$ git checkout pull/14271` \
`$ git pull https://git.openjdk.org/jdk.git pull/14271/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14271`

`$ git pr show -t 14271`

</details>

<a href="https://git.openjdk.org/jdk/pull/14271.diff">https://git.openjdk.org/jdk/pull/14271.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14271#issuecomment-1636151939)


AlanBateman:
This looks quite good. I think Console::isTerminal works as a method name but we might have to iterate a bit on the javadoc to more clearly define it.

For the CSR I think the behavior change to call out is that System::console will return a Console for cases where it didn't previously (at least not by default) and that will be observed by code that assumes non-null means isatty, cue the new method to provide a standard way to test this.

naotoj:
Thanks, Alan.

> This looks quite good. I think Console::isTerminal works as a method name but we might have to iterate a bit on the javadoc to more clearly define it.
> 

Do you think adding an `implNote` that describes the platform behavior, e.g., on Unix it is equivalent to calling `isatty(stdin/out)` and stdin/out file descriptor are valid character devices on Windows, would help?

src/java.base/share/classes/java/io/Console.java
```diff
@@ -344,16 +344,35 @@ public Charset charset() {
...
     }
 
+    /**
+     * {@return {@code true} if the {@code Console} instance is a terminal}
+     * <p>
+     * If it is {@code true}, the {@code Console} instance is attached to a terminal
+     * provided by the underlying platform, typically its input and output
+     * character devices are available to the {@code Console}. Otherwise it could
+     * mean that the implementation of the {@code Console} may simulate those
+     * devices within.
```
AlanBateman, line 354: The implNote helps as this is a difficult method to specify in a platform independent way.

The class description uses "interactive command line" and  "will typically be connected to the keyboard and display" and maybe we could re-use that here. Here's a suggestion for the second paragraph:

"This method returns true if the console device, associated with the current Java virtual machine, is a terminal, typically an interactive command line connected to a keyboard and display."

I don't have any other comments on the implementation changes, they look fine.

naotoj: Thanks. Replaced the second paragraph with your suggested sentence.

AlanBateman (approved):

PR #14274, opened 2023-06-01
JDK-8272147

lgxbslgx:
Hi all,

This patch removes the class `markSweep.hpp::PreservedMark` and uses the shared `PreservedMarksSet` instead.

In order to keep the meaning of the field `_preserved_marks` in `markSweep`,
I need to use `OopAndMarkWord` explicitly. So I move `OopAndMarkWord` out of `PreservedMarks`.

Tests:
`hotspot:tier1` passed locally (x86 & linux).

Thanks for the review.

Best Regards,
-- Guoxiong

### DIFF
--- a/src/hotspot/share/gc/serial/genMarkSweep.cpp
+++ b/src/hotspot/share/gc/serial/genMarkSweep.cpp
@@ -48,6 +48,7 @@
 #include "gc/shared/space.hpp"
 #include "gc/shared/strongRootsScope.hpp"
 #include "gc/shared/weakProcessor.hpp"
+#include "gc/shared/preservedMarks.inline.hpp"
 #include "memory/universe.hpp"
 #include "oops/instanceRefKlass.hpp"
 #include "oops/oop.inline.hpp"
@@ -133,13 +134,15 @@ void GenMarkSweep::allocate_stacks() {
   // revert to malloc.
   if (scratch != nullptr) {
     _preserved_count_max =
-      scratch->num_words * HeapWordSize / sizeof(PreservedMark);
+      scratch->num_words * HeapWordSize / sizeof(OopAndMarkWord);
   } else {
     _preserved_count_max = 0;
   }
 
-  _preserved_marks = (PreservedMark*)scratch;
+  _preserved_marks = (OopAndMarkWord*)scratch;
   _preserved_count = 0;
+
+  _preserved_overflow_stack_set.init(1);
 }
 
 
@@ -147,7 +150,7 @@ void GenMarkSweep::deallocate_stacks() {
   GenCollectedHeap* gch = GenCollectedHeap::heap();
   gch->release_scratch();
 
-  _preserved_overflow_stack.clear(true);
+  _preserved_overflow_stack_set.reclaim();
   _marking_stack.clear();
   _objarray_stack.clear(true);
 }
--- a/src/hotspot/share/gc/serial/markSweep.cpp
+++ b/src/hotspot/share/gc/serial/markSweep.cpp
@@ -44,10 +44,10 @@ uint                    MarkSweep::_total_invocations = 0;
 Stack<oop, mtGC>              MarkSweep::_marking_stack;
 Stack<ObjArrayTask, mtGC>     MarkSweep::_objarray_stack;
 
-Stack<PreservedMark, mtGC>    MarkSweep::_preserved_overflow_stack;
+PreservedMarksSet       MarkSweep::_preserved_overflow_stack_set(false /* in_c_heap */);
 size_t                  MarkSweep::_preserved_count = 0;
 size_t                  MarkSweep::_preserved_count_max = 0;
-PreservedMark*          MarkSweep::_preserved_marks = nullptr;
+OopAndMarkWord*         MarkSweep::_preserved_marks = nullptr;
 STWGCTimer*             MarkSweep::_gc_timer        = nullptr;
 SerialOldTracer*        MarkSweep::_gc_tracer       = nullptr;
 
@@ -142,14 +142,6 @@ template <class T> void MarkSweep::follow_root(T* p) {
 void MarkSweep::FollowRootClosure::do_oop(oop* p)       { follow_root(p); }
 void MarkSweep::FollowRootClosure::do_oop(narrowOop* p) { follow_root(p); }
 
-void PreservedMark::adjust_pointer() {
-  MarkSweep::adjust_pointer(&_obj);
-}
-
-void PreservedMark::restore() {
-  _obj->set_mark(_mark);
-}
-
 // We preserve the mark which should be replaced at the end and the location
 // that it will go.  Note that the object that this markWord belongs to isn't
 // currently at that address but it will be after phase4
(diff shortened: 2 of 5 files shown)


src/hotspot/share/gc/shared/preservedMarks.hpp
```diff
@@ -34,27 +34,30 @@ class WorkerTask;
...
-    void set_oop(oop obj) { _o = obj; }
-  };
   typedef Stack<OopAndMarkWord, mtGC> OopAndMarkWordStack;
 
   OopAndMarkWordStack _stack;
 
   inline bool should_preserve_mark(oop obj, markWord m) const;
 
 public:
+  OopAndMarkWordStack& get_stack() { return _stack; }
```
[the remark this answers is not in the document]
lgxbslgx: The methods `MarkSweep::adjust_pointer` and `PreservedMarks::adjust_during_full_gc` have different implementations. The method `MarkSweep::adjust_pointer` seems doing more thing. So in `MarkSweep::adjust_marks`, I use `get_stack` to get the stack and invoke `MarkSweep::adjust_pointer` explicitly. If you can confirm `PreservedMarks::adjust_during_full_gc` is good to replace `MarkSweep::adjust_pointer`, I will remove `get_stack` and use `PreservedMarks::adjust_during_full_gc` instead of `MarkSweep::adjust_pointer`.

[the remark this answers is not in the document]
tschatzl: You can replace that code with `PreservedMarks::adjust_during_full_gc` from what I can see; the additional code in `MarkSweep::adjust_pointer` only deals with the generic case if the passed `oop` is null, which can't be the case for oops in the preserved marks set.

tschatzl (changes requested):

src/hotspot/share/gc/serial/genMarkSweep.cpp
```diff
@@ -48,6 +48,7 @@
 #include "gc/shared/space.hpp"
 #include "gc/shared/strongRootsScope.hpp"
 #include "gc/shared/weakProcessor.hpp"
+#include "gc/shared/preservedMarks.inline.hpp"
```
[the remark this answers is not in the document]
lgxbslgx: Fixed.

src/hotspot/share/gc/serial/markSweep.cpp
```diff
@@ -159,9 +151,9 @@ void MarkSweep::preserve_mark(oop obj, markWord mark) {
   // sufficient space for the marks we need to preserve but if it isn't we fall
   // back to using Stacks to keep track of the overflow.
   if (_preserved_count < _preserved_count_max) {
-    _preserved_marks[_preserved_count++] = PreservedMark(obj, mark);
+    _preserved_marks[_preserved_count++] = OopAndMarkWord(obj, mark);
   } else {
-    _preserved_overflow_stack.push(PreservedMark(obj, mark));
+    _preserved_overflow_stack_set.get()->push_if_necessary(obj, mark);
```
albertnetymk, line 156: Why not `push_always`? I'd expect this mark must be preserved.

src/hotspot/share/gc/serial/markSweep.cpp
```diff
@@ -205,30 +197,23 @@ AdjustPointerClosure MarkSweep::adjust_pointer_closure;
 void MarkSweep::adjust_marks() {
   // adjust the oops we saved earlier
   for (size_t i = 0; i < _preserved_count; i++) {
-    _preserved_marks[i].adjust_pointer();
+    MarkSweep::adjust_pointer(_preserved_marks[i].get_oop_pointer());
```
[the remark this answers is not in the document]
lgxbslgx: > I wonder if one can add new API in PreservedMarks to adjust oops.

Should I add it in this patch? And, if such method is added, the method `PreservedMarks::adjust_during_full_gc` can use it too.

src/hotspot/share/gc/serial/markSweep.cpp
lgxbslgx: Fixed.

src/hotspot/share/gc/serial/markSweep.cpp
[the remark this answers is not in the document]
albertnetymk: Yes, please.

src/hotspot/share/gc/shared/preservedMarks.hpp
```diff
@@ -34,20 +34,21 @@ class WorkerTask;
 class PreservedMarksSet;
 class WorkerThreads;
 
+class OopAndMarkWord {
```
albertnetymk, line 37: Now that this class is in public domain, I tend to think it should be renamed to `PreservedMark` to be consistent with other classes in this file. (Ofc, this is quite subjective.)

albertnetymk (approved):

src/hotspot/share/gc/shared/preservedMarks.hpp
lgxbslgx: Renamed.

lgxbslgx:
Waiting for another review.

src/hotspot/share/gc/shared/preservedMarks.hpp
```diff
@@ -34,21 +34,22 @@ class WorkerTask;
...
+};
 
-    oop get_oop() { return _o; }
-    inline void set_mark() const;
-    void set_oop(oop obj) { _o = obj; }
-  };
-  typedef Stack<OopAndMarkWord, mtGC> OopAndMarkWordStack;
+class PreservedMarks {
+private:
+  typedef Stack<PreservedMark, mtGC> OopAndMarkWordStack;
```
rkennke, line 52: You renamed OopAndMarkWord to PreservedMark, which I like. Maybe you also want to rename OopAndMarkWordStack to PreservedMarkStack? Otherwise it looks confusing I think.

src/hotspot/share/gc/shared/preservedMarks.cpp
```diff
@@ -34,21 +34,24 @@
...
-    const OopAndMarkWord elem = _stack.pop();
+    const PreservedMark elem = _stack.pop();
     elem.set_mark();
   }
   assert_empty();
 }
 
+void PreservedMarks::adjust_preserved_mark(PreservedMark* elem) {
+  oop obj = elem->get_oop();
+  if (obj->is_forwarded()) {
```
tschatzl, line 45: Pre-existing: we probably only need to check this because G1 full gc preserves marks during marking, unlike the other collectors which (attempt to) preserve marks for actually forwarded objects.
I'll file an RFE for this.

tschatzl (approved):
Lgtm.
Please fix the naming suggested by @rkennke though before pushing.

src/hotspot/share/gc/shared/preservedMarks.hpp
lgxbslgx: Fixed.

lgxbslgx:
@albertnetymk @rkennke @tschatzl Thanks for the reviews. I will integrate this patch after `Pre-submit tests` passing.

PR #14276, opened 2023-06-02
8309216: Cast from jchar* to char* in test java/io/GetXSpace.java

bplb:
For Unix, copy the `jchar`s into an allocated `char` array.

---------
### Progress

### Issue
 * [JDK-8309216](https://bugs.openjdk.org/browse/JDK-8309216): Cast from jchar* to char* in test java/io/GetXSpace.java

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14276/head:pull/14276` \
`$ git checkout pull/14276`

`$ git checkout pull/14276` \
`$ git pull https://git.openjdk.org/jdk.git pull/14276/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14276`

`$ git pr show -t 14276`

</details>

<a href="https://git.openjdk.org/jdk/pull/14276.diff">https://git.openjdk.org/jdk/pull/14276.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14276#issuecomment-1572943099)

### DIFF
--- a/test/jdk/java/io/File/libGetXSpace.c
+++ b/test/jdk/java/io/File/libGetXSpace.c
@@ -129,7 +129,7 @@ Java_GetXSpace_getSpace0
     }
 #else
     int len = (int)(*env)->GetStringLength(env, root);
-    char* chars = (char*)malloc(len*sizeof(char) + 1);
+    char* chars = (char*)malloc((len + 1)*sizeof(char));
     if (chars == NULL) {
         (*env)->ReleaseStringChars(env, root, strchars);
         JNU_ThrowByNameWithLastError(env, "java/lang/RuntimeException",
--- a/test/jdk/java/io/File/libGetXSpace.c
+++ b/test/jdk/java/io/File/libGetXSpace.c
@@ -83,13 +83,12 @@ Java_GetXSpace_getSpace0
         // use GetDiskSpaceInformationW
         DISK_SPACE_INFORMATION diskSpaceInfo;
         BOOL hres = pfnGetDiskSpaceInformation(path, &diskSpaceInfo);
+        (*env)->ReleaseStringChars(env, root, strchars);
         if (FAILED(hres)) {
-            (*env)->ReleaseStringChars(env, root, strchars);
             JNU_ThrowByNameWithLastError(env, "java/lang/RuntimeException",
                                          "GetDiskSpaceInformationW");
             return totalSpaceIsEstimated;
         }
-        (*env)->ReleaseStringChars(env, root, strchars);
 
         ULONGLONG bytesPerAllocationUnit =
             diskSpaceInfo.SectorsPerAllocationUnit*diskSpaceInfo.BytesPerSector;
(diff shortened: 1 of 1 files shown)


bplb:
Preliminary test run passes.

test/jdk/java/io/File/libGetXSpace.c
```diff
@@ -124,9 +128,24 @@ Java_GetXSpace_getSpace0
...
+    if (chars == NULL) {
+        (*env)->ReleaseStringChars(env, root, strchars);
+        JNU_ThrowByNameWithLastError(env, "java/lang/RuntimeException",
+                                     "malloc");
+        return JNI_FALSE;
+    }
+
+    for (int i = 0; i < len; i++) {
+        chars[i] = (char)strchars[i];
+    }
```
RogerRiggs, line 142: Is this `memcpy(chars, strchars, len);`

test/jdk/java/io/File/libGetXSpace.c
```diff
@@ -77,17 +77,19 @@ Java_GetXSpace_getSpace0
         }
     }
 
-    LPCWSTR path = (LPCWSTR)chars;
+    LPCWSTR path = (LPCWSTR)strchars;
 
     if (pfnGetDiskSpaceInformation != NULL) {
         // use GetDiskSpaceInformationW
         DISK_SPACE_INFORMATION diskSpaceInfo;
         BOOL hres = pfnGetDiskSpaceInformation(path, &diskSpaceInfo);
```
RogerRiggs, line 85: The strchars/path could be released before checking the return; keeping the cleanup in the common flow.
Here and below after calling GetDiskFreeSpaceExW.
Your call.

test/jdk/java/io/File/libGetXSpace.c
bplb: I don't think so if `chars` is `char *` and `strchars` is `jchar*` as I think `char` is 8 bits and `jchar` is 16 bits.

test/jdk/java/io/File/libGetXSpace.c
bplb: Good idea; will update.

bplb: So changed in 0ecf89b1d060bb8c1a35402ec108cbe9f3f35bb9.

test/jdk/java/io/File/libGetXSpace.c
naotoj: I just wonder if jchar is UTF-16 and char is platform encoding, would it work by simply casting the low 8bit to char?

bplb: Probably not. This is only intended to cover the constrained case of this test. Something like the macro `WITH_PLATFORM_STRING` in `libjava/io_util.h` would be needed for the general case, but I thought that that might be overdoing it for here. I could of course be mistaken.

naotoj: OK, if the usage is limited and covers the required test cases, that sounds reasonable.

naotoj (approved):

test/jdk/java/io/File/libGetXSpace.c
bplb: Only file system roots are passed as the paths handed to the `getSpace0()` method so hopefully it should be all right.

PR #14281, opened 2023-06-02
JDK-8309219: Fix xlc17 clang 15 warnings in java.base

JoKern65:
This pr is a split off from JDK-8308288: Fix xlc17 clang warnings in shared code https://github.com/openjdk/jdk/pull/14146
It handles the part in java.base.

Compiling on AIX with xlc17 which contains the new clang 15 frontend
shows the following warnings.

built by make/modules/java.base/Lib.gmk

src/java.base/unix/native/libnet/DefaultProxySelector.c:378:41:22: error: passing arguments to a function without a prototype is deprecated in all versions of C and is not supported in C2x [-Werror,-Wdeprecated-non-prototype]
proxies = (*g_proxy_resolver_lookup)(resolver, uri, NULL, &error);
^

src/java.base/unix/native/libnet/NetworkInterface.c:1612:24: error: arithmetic on a pointer to void is a GNU extension [-Werror,-Wgnu-pointer-arith]
end = (void *)nddp + size;

src/java.base/unix/native/libnio/fs/UnixNativeDispatcher.c:1251:7: error: '_ALLBSD_SOURCE' is not defined, evaluates to 0 [-Werror,-Wundef]
#elif _ALLBSD_SOURCE
The whole file checks whether _ALLBSD_SOURCE is defined, only where #elif is used the call to defined() was forgotten.

built by make/modules/java.base/lib/CoreLibraries.gmk

src/java.base/unix/native/libjava/ProcessHandleImpl_unix.c:638 comparison of integers of different signs: 'int' and 'unsigned long'
if (ret < sizeof(psinfo_t)) {
ProcessHandleImpl_unix.c:  Just using the proper type `size_t` instead of `int`.

src/java.base/aix/native/libjli/java_md_aix.c:42:38: error: arithmetic on a pointer to void is a GNU extension [-Werror,-Wgnu-pointer-arith]
addr < p->ldinfo_textorg + p->ldinfo_textsize) {
we can fix this with temporary casting the `void*` pointer to `char*` or by disabling the warning in CoreLibraries.gmk

src/java.base/share/native/libzip/zlib/inffast.c:74:20: error: a function definition without a prototype is deprecated in all versions of C and is not supported in C2x [-Werror,-Wdeprecated-non-prototype]
void ZLIB_INTERNAL inflate_fast(strm, start)
^
src/java.base/share/native/libjli/java.c:2311:22: error: format string is not a string literal [-Werror,-Wformat-nonliteral]
vfprintf(stderr, fmt, vl);
^~~

The test library libGetXSpace.c also shows a warning.
In addition, the cast in that lib is wrong, we opened "[JDK-8309216](https://bugs.openjdk.org/browse/JDK-8309216) cast from jchar* to char* in test GetXSpace.java" for that.

---------
### Progress

### Issue
 * [JDK-8309219](https://bugs.openjdk.org/browse/JDK-8309219): Fix xlc17 clang 15 warnings in java.base (**Bug** - `"4"`)

### Reviewers
 * [Goetz Lindenmaier](https://openjdk.org/census#goetz) (@GoeLin - **Reviewer**) ⚠️ Review applies to [f3e92804](https://git.openjdk.org/jdk/pull/14281/files/f3e92804bf64fe06922b53ff4317b5a603e020ea)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14281/head:pull/14281` \
`$ git checkout pull/14281`

`$ git checkout pull/14281` \
`$ git pull https://git.openjdk.org/jdk.git pull/14281/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14281`

`$ git pr show -t 14281`

</details>

<a href="https://git.openjdk.org/jdk/pull/14281.diff">https://git.openjdk.org/jdk/pull/14281.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14281#issuecomment-1573492255)


GoeLin (approved):
Lib.gmk / DefaultProxySelector.c
Fixing this in the file would require a larger rework. Good.
Lib.gmk / NetworkInterface.c
This looks as if it could also be fixed in the code.
But previously, other warnings have not been fixed either.
So I'm fine with this.


CoreLibraries / deprecated-non-prototype
Fixes warning in zlib, we should not touch the code
there as it is from external. Good.

CoreLibraries / format-nonliteral
In hotspot, these warnings have been fixed, or
an attribute is used to disable them locally.
These places pop up in code scanners, too.
See compilerWarnings.hpp / ATTRIBUTE_PRINTF().
The code here can not be fixed, and
I don't see any usage of the attribute or similar
in java.base, so disabling the warning seems fine to me.
Also, it's treated similar in the other
compile block.

java_md_aix.c: good.

ProcessHandleImpl_unix.c: good

UnixNativeDispatcher.c: definitely an improvement to the code!

libGetXSpace.c: good. A test library

TheRealMDoerr (approved):
LGTM.

TheRealMDoerr (approved):

PR #14283, opened 2023-06-02
JDK-8308288: Fix xlc17 clang warnings and build errors in hotspot

JoKern65:
This pr is a split off from JDK-8308288 : Fix xlc17 clang warnings in shared code https://github.com/openjdk/jdk/pull/14146
It handles the part in hotspot.

It handles the error introduced by a redefine of malloc in stdlib.h resulting in the following build error:

/data/d042520/pr/jdk/src/hotspot/share/runtime/os.cpp:616:5: error: no member named '_vec_malloc' in 'LogTag'; did you mean 'vec_malloc'?
    log_warning(malloc, free)("ptr caught: " PTR_FORMAT, p2i(ptr));
    ^~~~~~~~~~~~~~~~~~~~~~~~~
/data/d042520/pr/jdk/src/hotspot/share/logging/log.hpp:46:28: note: expanded from macro 'log_warning'
#define log_warning(...) (!log_is_enabled(Warning, __VA_ARGS__)) ? (void)0 : LogImpl<LOG_TAGS(__VA_ARGS__)>::write<LogLevel::Warning>
                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/data/d042520/pr/jdk/src/hotspot/share/logging/log.hpp:68:45: note: expanded from macro 'log_is_enabled'
#define log_is_enabled(level, ...) (LogImpl<LOG_TAGS(__VA_ARGS__)>::is_level(LogLevel::level))
                                            ^~~~~~~~~~~~~~~~~~~~~
/data/d042520/pr/jdk/src/hotspot/share/logging/logTag.hpp:221:38: note: expanded from macro 'LOG_TAGS'
#define LOG_TAGS(...) EXPAND_VARARGS(LOG_TAGS_EXPANDED(__VA_ARGS__, _NO_TAG, _NO_TAG, _NO_TAG, _NO_TAG, _NO_TAG, _NO_TAG))
                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/data/d042520/pr/jdk/src/hotspot/share/logging/logTag.hpp:217:57: note: expanded from macro 'LOG_TAGS_EXPANDED'
#define LOG_TAGS_EXPANDED(T0, T1, T2, T3, T4, T5, ...)  PREFIX_LOG_TAG(T0), PREFIX_LOG_TAG(T1), PREFIX_LOG_TAG(T2), \
                                                        ^~~~~~~~~~~~~~~~~~
   ... (rest of output omitted)

Additionally it solves the need for an #include <alloca.h> on AIX for any usage of the alloca function, by adding the include to globalDefinitions_xlc.hpp

### DIFF
--- a/src/hotspot/share/runtime/os.cpp
+++ b/src/hotspot/share/runtime/os.cpp
@@ -162,7 +162,7 @@ char* os::iso8601_time(jlong milliseconds_since_19700101, char* buffer, size_t b
   // No offset when dealing with UTC
   time_t UTC_to_local = 0;
   if (!utc) {
-#if defined(_ALLBSD_SOURCE) || defined(_GNU_SOURCE)
+#if (defined(_ALLBSD_SOURCE) || defined(_GNU_SOURCE)) && !defined(AIX)
     UTC_to_local = -(time_struct.tm_gmtoff);
 #elif defined(_WINDOWS)
     long zone;
@@ -878,8 +878,8 @@ bool os::print_function_and_library_name(outputStream* st,
   // this as a function descriptor for the reader (see below).
   if (!have_function_name && os::is_readable_pointer(addr)) {
     address addr2 = (address)os::resolve_function_descriptor(addr);
-    if (have_function_name = is_function_descriptor =
-        dll_address_to_function_name(addr2, p, buflen, &offset, demangle)) {
+    if ((have_function_name = is_function_descriptor =
+        dll_address_to_function_name(addr2, p, buflen, &offset, demangle))) {
       addr = addr2;
     }
   }
--- a/src/hotspot/share/utilities/globalDefinitions_xlc.hpp
+++ b/src/hotspot/share/utilities/globalDefinitions_xlc.hpp
@@ -32,12 +32,22 @@
 // globally used constants & types, class (forward)
 // declarations and a few frequently used utility functions.
 
+#include <alloca.h>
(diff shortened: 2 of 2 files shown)


GoeLin (approved):
os.cpp / timezone
I leave it to you that it's the correct behaviour
for AIX, for the other platforms it looks good.

os.cpp / print_function_and_library_name()
I don't like assignements in if(), but the fix here
should not change this. Good.

globalDefinitions_xlc.hpp / alloca.h
Good.

globalDefinitions_xlc.hpp / malloc
Having looked at the discussion about this in #14146,
I think this is the solution we want to have here.
I don't think we should harden the code wrt. the fact
that malloc and others can be macros. At least as long
as none of the compilers of the main platforms detects
this (e.g. running in GHA).
Also, I know that Joachim asked IBM to remove the
macro from the header. After all, it worked with xlc16.
So let's fix this here in the _xlc file.

JoKern65:
@kimbarret, @tstuefe, are you fine with pushing the fix for the malloc
macro as is?

MBaesken (approved):
The malloc change does not look very nice, but I think for now we can use that on AIX with xlc17.

PR #14291, opened 2023-06-02
8309390: [JVMCI] improve copying system properties into libgraal

dougxc:
This PR reduces the amount of code executed during libgraal initialization. This not only improves VM startup overall but all fixes a number of JDK test failures that are caused by Java code executing "too early". For example, `java/util/Locale/CompatWarning.java` can fail if `sun.util.locale.provider.LocaleProviderAdapter` is initialized before the `CheckWarning` handler is registered.

Instead of serializing `VM.savedProps` via `VMSupport.serializeSavedPropertiesToByteArray`, the `jdk.vm.ci.services.Services` class now directly reads `Arguments::system_properties()` using `Unsafe`. Furthermore, the value of a system property is lazily converted to a `String` from a C string pointer.

## Times

The basic benchmarking below shows that this change brings the time for a nop Java app with eager libgraal initialization (2) down to almost the same time as lazy libgraal initialization (1). The latter typically means no libgraal initialization happens as a top tier JIT compilation is never scheduled in such a short running app.

```
public class Nop {
    public static void main(String[] args) {}
}
```

(1) Baseline (no options):
```
> for i in (seq 10); java Nop; end
        0.05 real         0.04 user         0.01 sys
        0.04 real         0.03 user         0.01 sys
        0.04 real         0.03 user         0.01 sys
        0.04 real         0.03 user         0.01 sys
        0.03 real         0.03 user         0.00 sys
        0.04 real         0.03 user         0.01 sys
        0.04 real         0.03 user         0.00 sys
        0.03 real         0.03 user         0.00 sys
        0.04 real         0.03 user         0.01 sys
        0.03 real         0.03 user         0.00 sys
```

(2) Eagerly initialize libgraal (with PR):
```
> for i in (seq 10); /usr/bin/time java -XX:+EagerJVMCI Nop; end
        0.06 real         0.04 user         0.01 sys
        0.05 real         0.03 user         0.01 sys
        0.05 real         0.03 user         0.01 sys
        0.05 real         0.03 user         0.01 sys
        0.05 real         0.03 user         0.01 sys
        0.05 real         0.03 user         0.01 sys
        0.05 real         0.03 user         0.01 sys
        0.05 real         0.03 user         0.01 sys
        0.05 real         0.03 user         0.01 sys
        0.05 real         0.03 user         0.01 sys
```

(3) Eagerly initialize libgraal (without PR):
```
> for i in (seq 10); /usr/bin/time java -XX:+EagerJVMCI Nop; end
        0.11 real         0.08 user         0.02 sys
        0.08 real         0.06 user         0.01 sys
        0.08 real         0.07 user         0.01 sys
        0.10 real         0.07 user         0.01 sys
        0.08 real         0.06 user         0.01 sys
        0.10 real         0.07 user         0.01 sys
        0.08 real         0.07 user         0.01 sys
        0.08 real         0.07 user         0.01 sys
        0.08 real         0.06 user         0.01 sys
        0.08 real         0.06 user         0.01 sys
```

---------
### Progress

### Issue
 * [JDK-8309390](https://bugs.openjdk.org/browse/JDK-8309390): [JVMCI] improve copying system properties into libgraal (**Enhancement** - P4)

### Reviewers
 * [Tom Rodriguez](https://openjdk.org/census#never) (@tkrodriguez - **Reviewer**) ⚠️ Review applies to [6c346d35](https://git.openjdk.org/jdk/pull/14291/files/6c346d35923402fb5d6ae0ef83214f8f0ca07b69)
 * [Vladimir Kozlov](https://openjdk.org/census#kvn) (@vnkozlov - **Reviewer**) ⚠️ Review applies to [6c346d35](https://git.openjdk.org/jdk/pull/14291/files/6c346d35923402fb5d6ae0ef83214f8f0ca07b69)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14291/head:pull/14291` \
`$ git checkout pull/14291`

`$ git checkout pull/14291` \
`$ git pull https://git.openjdk.org/jdk.git pull/14291/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14291`

`$ git pr show -t 14291`

</details>

<a href="https://git.openjdk.org/jdk/pull/14291.diff">https://git.openjdk.org/jdk/pull/14291.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14291#issuecomment-1574884455)


src/java.base/share/classes/jdk/internal/util/SystemProps.java
```diff
@@ -59,7 +59,7 @@ public static boolean isBadIoTmpdir() {
      * Note:  Build-defined properties such as versions and vendor information
      * are initialized by VersionProps.java-template.
      *
-     * @return a Properties instance initialized with all of the properties
```
dougxc, line 62: I took the liberty of correcting this javadoc to reflect the code.

tkrodriguez:
I don't really love the hard code parsing of the HashMap.  What properties are actually required for JVMCI?  It seems to me that the contents of Arguments::system_properties() should contain all the properties we want to advertise to JVMCI.  That would have avoid having to decode them after they've been converted into Java objects.

dougxc:
> I don't really love the hard code parsing of the HashMap. What properties are actually required for JVMCI? It seems to me that the contents of Arguments::system_properties() should contain all the properties we want to advertise to JVMCI. That would have avoid having to decode them after they've been converted into Java objects.

I tired this but unfortunately, Graal relies on some properties that are only initialized in Java:
```
Caused by: java.lang.NullPointerException: Cannot invoke "String.compareTo(String)" because "this.javaSpecVersion" is null
	at jdk.internal.vm.compiler/org.graalvm.compiler.hotspot.JVMCIVersionCheck.run(JVMCIVersionCheck.java:181)
	at jdk.internal.vm.compiler/org.graalvm.compiler.hotspot.JVMCIVersionCheck.check(JVMCIVersionCheck.java:166)
	at jdk.internal.vm.compiler/org.graalvm.compiler.hotspot.HotSpotGraalCompilerFactory.initialize(HotSpotGraalCompilerFactory.java:117)
	at jdk.internal.vm.compiler/org.graalvm.compiler.hotspot.HotSpotGraalCompilerFactory.ensureInitialized(HotSpotGraalCompilerFactory.java:90)
	at jdk.internal.vm.compiler/org.graalvm.compiler.hotspot.HotSpotGraalCompilerFactory.createCompiler(HotSpotGraalCompilerFactory.java:180)
	at jdk.internal.vm.compiler/org.graalvm.compiler.hotspot.HotSpotGraalCompilerFactory.createCompiler(HotSpotGraalCompilerFactory.java:53)
	at jdk.internal.vm.ci/jdk.vm.ci.hotspot.HotSpotJVMCIRuntime.getCompiler(HotSpotJVMCIRuntime.java:810)
	at org.graalvm.nativeimage.pointsto/com.oracle.graal.pointsto.util.GraalAccess.<clinit>(GraalAccess.java:50)
```

That code is reading the "java.specification.version" property which is initialized in `java.lang.VersionProps#init`.

tkrodriguez:
HotSpot sets `java.vm.specification.version` which has the same value so we could change Graal to read that instead.

dougxc:
Ok, I've pushed a change with your suggestion Tom and it seems to work. It assumes `Arguments::system_properties()` is fully initialized and effectively read-only by the time `JVMCIEnv::init_saved_properties` is called.

src/hotspot/share/jvmci/jvmciJavaClasses.hpp
```diff
@@ -57,7 +57,7 @@
                          jvmci_method, \
                          jvmci_constructor) \
   start_class(Services, jdk_vm_ci_services_Services)                                                          \
-    jvmci_method(CallStaticVoidMethod, GetStaticMethodID, call_static, void, Services, initializeSavedProperties, byte_array_void_signature, (JVMCIObject serializedProperties)) \
```
dougxc, line 60: The final arg for the `jvmci_method` macro is never used so I removed it everywhere.

src/jdk.internal.vm.ci/share/classes/jdk/vm/ci/services/Services.java
```diff
@@ -261,22 +252,128 @@ protected void threadTerminated(T value) {
...
+                // Some internal properties (e.g. jdk.boot.class.path.append) can have a null
+                // value and should just be ignored. Note that null is different than the empty string.
+            }
+        }
+        Map<String, SystemProperties.Value> props = new HashMap<>(count + 1);
+        int i = 0;
+        for (long prop = systemProperties; prop != 0; prop = unsafe.getLong(prop + nextOffset)) {
+            String key = toJavaString(unsafe, unsafe.getLong(prop + keyOffset));
+            long valueAddress = unsafe.getLong(prop + valueOffset);
+            if (valueAddress != 0) {
```
tkrodriguez, line 314: Is the lazy value construction really necessary?  It seems like it requires quite a bit of extra machinery.

tkrodriguez (approved):

dougxc:
Thanks for the reviews!

/integrate

PR #14296, opened 2023-06-03
8309403: Serial: Remove the useless adaptive size policy in GenCollectedHeap

lgxbslgx:
Hi all,

This patch removes the useless code related to the `AdaptiveSizePolicy` in `GenCollectedHeap`.
As a result, the `SoftRefGenPolicy` is unnecessary too, use `SoftRefPolicy` instead.

Thanks for the review.

Best Regards,
-- Guoxiong

---------
### Progress

### Issue
 * [JDK-8309403](https://bugs.openjdk.org/browse/JDK-8309403): Serial: Remove the useless adaptive size policy in GenCollectedHeap (**Enhancement** - P5)

### Reviewers
 * [Albert Mingkun Yang](https://openjdk.org/census#ayang) (@albertnetymk - **Reviewer**) ⚠️ Review applies to [6d233e90](https://git.openjdk.org/jdk/pull/14296/files/6d233e90416ee7bcd908851309e482a5f759ccf7)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14296/head:pull/14296` \
`$ git checkout pull/14296`

`$ git checkout pull/14296` \
`$ git pull https://git.openjdk.org/jdk.git pull/14296/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14296`

`$ git pr show -t 14296`

</details>

<a href="https://git.openjdk.org/jdk/pull/14296.diff">https://git.openjdk.org/jdk/pull/14296.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14296#issuecomment-1574749013)


src/hotspot/share/gc/shared/softRefGenPolicy.cpp
```diff
@@ -1,40 +0,0 @@
...
-#include "gc/shared/genCollectedHeap.hpp"
-#include "gc/shared/softRefGenPolicy.hpp"
-
-void SoftRefGenPolicy::cleared_all_soft_refs() {
-  // If near gc overhear limit, continue to clear SoftRefs.  SoftRefs may
-  // have been cleared in the last collection but if the gc overhear
-  // limit continues to be near, SoftRefs should still be cleared.
-  AdaptiveSizePolicy* size_policy = GenCollectedHeap::heap()->size_policy();
-  if (size_policy != nullptr) {
-    set_should_clear_all_soft_refs(size_policy->gc_overhead_limit_near());
```
albertnetymk, line 36: Why is this call removed? I'd expect `set_should_clear_all_soft_refs(false);` here.

lgxbslgx: The default value of `_should_clear_all_soft_refs` is `false` and I don't see other places to set it to `true` in Serial GC. And from the comment, I think it is a leftover of CMS.

>  // If near gc overhear limit, continue to clear SoftRefs.  SoftRefs may
>  // have been cleared in the last collection but if the gc overhear
>  // limit continues to be near, SoftRefs should still be cleared.

lgxbslgx: The `WB_FullGC` is just test code. I have several problems. If the test code wants the GC to clear all soft references (`set_should_clear_all_soft_refs(true)` in `WB_FullGC`), why do we set it to `false` again? If we have a reason to set it to `false` again, why do other GCs not set it (Why is `SoftRefGenPolicy` only used at Serial GC?).

> // If near gc overhear limit, continue to clear SoftRefs. SoftRefs may
> // have been cleared in the last collection but if the gc overhear
> // limit continues to be near, SoftRefs should still be cleared.

From the comment, the intent of the method `SoftRefGenPolicy::cleared_all_soft_refs` is letting the next GC re-clear the soft reference if **near gc overhear limit**. But the Serial GC has no gc overhear limit check now. So I still think it is just a leftover of CMS.

albertnetymk: > why do we set it to false again

`WB_FullGC` should restore the global env. Otherwise, the following non-whitebox-full-gc cycles can unnecessarily clear soft-refs.

> If we have a reason to set it to false again, why do other GCs not set it

G1 does that inside `WB_FullGC`. ZGC doesn't rely on `SoftRefPolicy::should_clear_all_soft_refs`. Parallel doesn't clear the flag, which I believe is a bug.

This inconsistency is probably accidental; keeping the global env intact (essentially what ZGC does) seems the cleanest approach IMO. I wonder if other collectors can also adopt it; in another PR(s) ofc.

lgxbslgx: > > why do we set it to false again
> 
> `WB_FullGC` should restore the global env. Otherwise, the following non-whitebox-full-gc cycles can unnecessarily clear soft-refs.
>
> > If we have a reason to set it to false again, why do other GCs not set it
> 
> G1 does that inside `WB_FullGC`. ZGC doesn't rely on `SoftRefPolicy::should_clear_all_soft_refs`. Parallel doesn't clear the flag, which I believe is a bug.
> 
> This inconsistency is probably accidental; keeping the global env intact (essentially what ZGC does) seems the cleanest approach IMO. I wonder if other collectors can also adopt it; in another PR(s) ofc.

Thanks for your answers. I propose the following changes.

In this patch:
- Serial GC should use `SoftRefPolicy` instead of `SoftRefGenPolicy`, which is done in this patch now.
- Serial GC should restore the global env inside `WB_FullGC`, just like the G1. Because both of them don't implement overheap limit check now.
- Don't remove the class `SoftRefGenPolicy` because it will be used by Parallel GC in the future. 

In another patch:
- Rename `SoftRefGenPolicy` in order to avoid misunderstanding, such as `GCOverhearLimitSoftRefPolicy`.
- Use `SoftRefGenPolicy` (may be renamed at the first step) in Parallel GC.

lgxbslgx: >  I will file new tickets to follow up.

Filed https://bugs.openjdk.org/browse/JDK-8309616 and https://bugs.openjdk.org/browse/JDK-8309619 to follow up.

src/hotspot/share/prims/whitebox.cpp
```diff
@@ -1445,12 +1445,12 @@ WB_END
 WB_ENTRY(void, WB_FullGC(JNIEnv* env, jobject o))
   Universe::heap()->soft_ref_policy()->set_should_clear_all_soft_refs(true);
   Universe::heap()->collect(GCCause::_wb_full_gc);
-#if INCLUDE_G1GC
-  if (UseG1GC) {
-    // Needs to be cleared explicitly for G1
+#if defined(INCLUDE_G1GC) || defined(INCLUDE_PARALLELGC) || defined(INCLUDE_SERIALGC)
+  if (UseG1GC || UseParallelGC || UseSerialGC) {
```
albertnetymk, line 1449: I don't think you can do the same for Parallel due to the presence of `GCOverheadChecker`, which also updates the global var.

lgxbslgx: Fixed.

src/hotspot/share/prims/whitebox.cpp
```diff
@@ -1445,12 +1445,12 @@ WB_END
 WB_ENTRY(void, WB_FullGC(JNIEnv* env, jobject o))
   Universe::heap()->soft_ref_policy()->set_should_clear_all_soft_refs(true);
   Universe::heap()->collect(GCCause::_wb_full_gc);
-#if INCLUDE_G1GC
-  if (UseG1GC) {
-    // Needs to be cleared explicitly for G1
+#if defined(INCLUDE_G1GC) || defined(INCLUDE_SERIALGC)
```
albertnetymk, line 1448: `INCLUDE_G1GC` can be defined as zero; it's best to treat them as bool vars. The rest look good.

albertnetymk (approved):

lgxbslgx:
Waiting for another review.

tschatzl (approved):

lgxbslgx:
@albertnetymk @tschatzl Thanks for the reviews. Integrating.

/integrate

PR #14297, opened 2023-06-03
8309404: Parallel: Process class loader data graph in parallel in young gc

lgxbslgx:
Hi all,

This patch parallelizes the process of the class loader data graph in young gc.

The class `ClassLoaderData` has a field `_claim` to avoid applying oop closure more than once 
and the method `ClassLoaderData::oops_do` can check if the CLD had been claimed.
The parallel full gc has already used them in `MarkFromRootsTask::work` and `PSAdjustTask::work`.

But I don't have experience to test/verify the performance improvement of the GC.
If this patch needs such test data before integrating, please guide and help me here.

Thanks for the review and guidance.

Best Regards,
-- Guoxiong

---------
### Progress

### Issue
 * [JDK-8309404](https://bugs.openjdk.org/browse/JDK-8309404): Parallel: Process class loader data graph in parallel in young gc (**Enhancement** - P5)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14297/head:pull/14297` \
`$ git checkout pull/14297`

`$ git checkout pull/14297` \
`$ git pull https://git.openjdk.org/jdk.git pull/14297/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14297`

`$ git pr show -t 14297`

</details>

<a href="https://git.openjdk.org/jdk/pull/14297.diff">https://git.openjdk.org/jdk/pull/14297.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14297#issuecomment-1574828584)

### DIFF
--- a/src/hotspot/share/gc/parallel/psClosure.inline.hpp
+++ b/src/hotspot/share/gc/parallel/psClosure.inline.hpp
@@ -131,7 +131,7 @@ public:
       _oop_closure.set_scanned_cld(cld);
 
       // Clean the cld since we're going to scavenge all the metadata.
-      cld->oops_do(&_oop_closure, ClassLoaderData::_claim_none, /*clear_modified_oops*/true);
+      cld->oops_do(&_oop_closure, ClassLoaderData::_claim_strong, /*clear_modified_oops*/true);
 
       _oop_closure.set_scanned_cld(nullptr);
     }
--- a/src/hotspot/share/gc/parallel/psRootType.hpp
+++ b/src/hotspot/share/gc/parallel/psRootType.hpp
@@ -34,7 +34,6 @@ public:
   // The order reflects the order these roots are to be processed,
   // We do not want any holes in the enum as we enumerate these values by incrementing them.
   enum Value {
-    class_loader_data,
     code_cache,
     //"threads" are handled in parallel as a special case
     sentinel
--- a/src/hotspot/share/gc/parallel/psScavenge.cpp
+++ b/src/hotspot/share/gc/parallel/psScavenge.cpp
@@ -87,17 +87,9 @@ static void scavenge_roots_work(ParallelRootType::Value root_type, uint worker_i
   assert(ParallelScavengeHeap::heap()->is_gc_active(), "called outside gc");
 
   PSPromotionManager* pm = PSPromotionManager::gc_thread_promotion_manager(worker_id);
-  PSScavengeRootsClosure roots_closure(pm);
   PSPromoteRootsClosure  roots_to_old_closure(pm);
 
   switch (root_type) {
-    case ParallelRootType::class_loader_data:
-      {
-        PSScavengeCLDClosure cld_closure(pm);
-        ClassLoaderDataGraph::cld_do(&cld_closure);
-      }
-      break;
-
     case ParallelRootType::code_cache:
       {
         MarkingCodeBlobClosure code_closure(&roots_to_old_closure, CodeBlobToOopClosure::FixRelocations, false /* keepalive nmethods */);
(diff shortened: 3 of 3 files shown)


tschatzl (changes requested):
The reason why this is not parallelized in stw pauses is because the CLD data structure is not amenable to parallelization at all. This is basically a linked list, and when parallelizing it this way, 
* every thread visits every CLD anyway (doing the pointer chasing)
* threads are massively choking themselves on obtaining the claim value
* the code adds another pass through the CLD linked list clearing the claim marks
In my experience you *will* get significant negative scaling (i.e. that phase taking significant multiples of the original time) with this simple approach.

I.e. this has been analyzed before, see https://bugs.openjdk.org/browse/JDK-8030144; only Shenandoah does a parallel walk, but with limited number of threads (e.g. https://bugs.openjdk.org/browse/JDK-8246097)

The latter CR also provides some applications with apparently many CLDG entries (Spring Boot and CLion), which may be used for this investigation.
Fwiw, it may be easier to start this investigation with G1 as it has the timing logging already implmented (but it is no problem to add this to Parallel temporarily(?)).

lgxbslgx:
> The reason why this is not parallelized in stw pauses is because the CLD data structure is not amenable to parallelization at all. This is basically a linked list, and when parallelizing it this way,
> 
> * every thread visits every CLD anyway (doing the pointer chasing)
> * threads are massively choking themselves on obtaining the claim value
> * the code adds another pass through the CLD linked list clearing the claim marks
>   In my experience you _will_ get significant negative scaling (i.e. that phase taking significant multiples of the original time) with this simple approach.
> 
> I.e. this has been analyzed before, see https://bugs.openjdk.org/browse/JDK-8030144; only Shenandoah does a parallel walk, but with limited number of threads (e.g. https://bugs.openjdk.org/browse/JDK-8246097)
> 
> The latter CR also provides some applications with apparently many CLDG entries (Spring Boot and CLion), which may be used for this investigation. Fwiw, it may be easier to start this investigation with G1 as it has the timing logging already implmented (but it is no problem to add this to Parallel temporarily(?)).

Thanks for the guidance. The full GC in `Parallel` had parallelized it, so previously, I thought it is good to do the same thing in young GC. One of young GC or full GC should be adjusted. I will investigate more information in related issues (mainly test the performance).

lgxbslgx:
@tschatzl I have a problem. During searching the roots in parallel, the claim values of the threads also need to be changed atomically by all the worker threads. Does it also have the same problem? Do you know/do any research related to it?

```java
// file threads.cpp
void Threads::possibly_parallel_threads_do(bool is_par, ThreadClosure* tc) {
  assert_at_safepoint();

  uintx claim_token = Threads::thread_claim_token();
  ALL_JAVA_THREADS(p) {
    if (p->claim_threads_do(is_par, claim_token)) {
      tc->do_thread(p);
    }
  }
  for (NonJavaThread::Iterator njti; !njti.end(); njti.step()) {
    Thread* current = njti.current();
    if (current->claim_threads_do(is_par, claim_token)) {
      tc->do_thread(current);
    }
  }
}

// file thread.cpp
bool Thread::claim_par_threads_do(uintx claim_token) {
  uintx token = _threads_do_token;
  if (token != claim_token) {
    uintx res = Atomic::cmpxchg(&_threads_do_token, token, claim_token);
    if (res == token) {
      return true;
    }
    guarantee(res == claim_token, "invariant");
  }
  return false;
}
```

tschatzl:
> @tschatzl I have a question. During searching the roots in parallel, the claim values of the threads also need to be changed atomically by all the worker threads. Does it also have the same problem? Do you know/do any research related to it?
> 

It generally has the same problem, but is mitigated because the work per Thread is much larger (a Thread has typically many stacks levels to look through), so the relative impact smaller. The amount of contention on that is much lower too as the probability that multiple threads try to claim the same claim value is lower due to longer and more variable time to process the task.

There is also a newer API available that allows less contention claiming, see e.g. `G1JavaThreadsListClaimer` and the use in `G1PreEvacuateCollectionSetBatchTask`, however at this point it is only used in a few places. Replacement work isn't trivial.

PR #14298, opened 2023-06-03
8307153: JVMTI GetThreadState on carrier should return STATE_WAITING

sspitsyn:
When a virtual thread is mounted, the carrier thread should be reported as "waiting" until the virtual thread unmounts. Right now, GetThreadState reports a state based the JavaThread status when it should return JVMTI_THREAD_STATE_WAITING | JVMTI_THREAD_STATE_WAITING_INDEFINITELY.
The fix adds:
 - a special case for passive carrier threads
 - necessary test coverage to the existing JVMTI test: `serviceability/jvmti/vthread/ThreadStateTest`.

Testing:
   - tested with the updated test: `serviceability/jvmti/vthread/ThreadStateTest`
   - submitted mach5 tiers 1-5
   - TBD: to submit mach5 tier 6

---------
### Progress

### Issue
 * [JDK-8307153](https://bugs.openjdk.org/browse/JDK-8307153): JVMTI GetThreadState on carrier should return STATE_WAITING (**Bug** - `"3"`)

### Reviewers
 * [Alex Menkov](https://openjdk.org/census#amenkov) (@alexmenkov - **Reviewer**) ⚠️ Review applies to [3e7618c4](https://git.openjdk.org/jdk/pull/14298/files/3e7618c4497849ece7c524da00133503baf3114e)
 * [Chris Plummer](https://openjdk.org/census#cjplummer) (@plummercj - **Reviewer**) ⚠️ Review applies to [3e7618c4](https://git.openjdk.org/jdk/pull/14298/files/3e7618c4497849ece7c524da00133503baf3114e)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14298/head:pull/14298` \
`$ git checkout pull/14298`

`$ git checkout pull/14298` \
`$ git pull https://git.openjdk.org/jdk.git pull/14298/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14298`

`$ git pr show -t 14298`

</details>

<a href="https://git.openjdk.org/jdk/pull/14298.diff">https://git.openjdk.org/jdk/pull/14298.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14298#issuecomment-1574876720)


src/hotspot/share/prims/jvmtiEnvBase.cpp
```diff
@@ -756,6 +756,16 @@ JvmtiEnvBase::get_thread_state(oop thread_oop, JavaThread* jt) {
   return state;
 }
 
+jint
+JvmtiEnvBase::get_thread_state(oop thread_oop, JavaThread* jt) {
+  jint state = get_thread_state_base(thread_oop, jt);
+
+  if (is_passive_carrier_thread(jt, thread_oop)) {
+    state |= (JVMTI_THREAD_STATE_WAITING | JVMTI_THREAD_STATE_WAITING_INDEFINITELY);
```
AlanBateman, line 764: This is testing if the jt is carrying thread_oop and it's okay for the JVMTI state to reported as WAITING when waiting for something other than Object.wait.

One thing that is a bit confusing is the function name "is_passive_carrier_thread". A platform thread is either a carrier or not. Maybe for a different PR but I think is_passive_carrier_thread should be renamed to avoid the use of the word "passive".

sspitsyn: The lines 763-764 are to correct the state exactly for passive carrier thread, a carrier thread which can't progress until the execution control has not been returned from a virtual thread executed on the top. It is never for a platform thread which is not a carrier thread. "Passive" is the best word I was able to find for this meaning. Do you have any other word in mind?

AlanBateman: > The lines 763-764 are to correct the state exactly for passive carrier thread, a carrier thread which can't progress until the execution control has not been returned from a virtual thread executed on the top. It is never for a platform thread which is not a carrier thread. "Passive" is the best word I was able to find for this meaning. Do you have any other word/suggestion in mind?

It's just a carrier. A platform thread becomes a carrier when a virtual thread is mounted, it ceases to be a carrier once the virtual thread is unmounted. The mental model is that the carrier is blocked so reporting its state as waiting indefinitely is correct. Maybe you don't want to rename it in this PR but renaming this function to something like is_carrying would convey that it's asking the question if a given JavaThread is carrying the given virtual thread oop.

sspitsyn:
>

src/hotspot/share/prims/jvmtiEnvBase.cpp
sspitsyn: Okay, I see you point. Unfortunately, I've always referred the platform thread with an executed FJP schedular as a carrier thread. The term 'carrier' with this meaning is everywhere in the JVMTI code. It looks very confusing to call a thread to be a carrier thread only during some phases of its execution.

AlanBateman: > Okay, I see you point. Unfortunately, I've always referred the platform thread with an executed FJP schedular as a carrier thread. The term 'carrier' with this meaning is everywhere in the JVMTI code. It looks very confusing to call a thread to be a carrier thread only during some phases of its execution.

Okay, I'm just pointing out that is_passive_carrier_thread is confusing looks a bit strange here as the is testing if a JavaThread is carrying a virtual thread oop -  it's not testing if the thread is owned by the virtual thread scheduler.

alexmenkov, line 764: Not sure I understand this.
I'd expect 
`JVMTI_THREAD_STATE_ALIVE | JVMTI_THREAD_STATE_WAITING | JVMTI_THREAD_STATE_WAITING_INDEFINITELY` to be returned in the case.
How can a thread be JVMTI_THREAD_STATE_RUNNABLE and JVMTI_THREAD_STATE_WAITING at the same time?

sspitsyn: Good catch, thanks. Fixed now.

sspitsyn: I'm still thinking what identifier to use instead of `is_passive_carrier_thread`.
Just `is_carrier_thread` is going to be confusing as well.
What about `is_carrier_thread_waiting_for_virtual`?

src/hotspot/share/prims/jvmtiEnvBase.cpp
```diff
@@ -756,6 +756,17 @@ JvmtiEnvBase::get_thread_state(oop thread_oop, JavaThread* jt) {
...
+jint
+JvmtiEnvBase::get_thread_state(oop thread_oop, JavaThread* jt) {
+  jint state = get_thread_state_base(thread_oop, jt);
+
+  if (is_passive_carrier_thread(jt, thread_oop)) {
+    state = JVMTI_THREAD_STATE_ALIVE | JVMTI_THREAD_STATE_WAITING |
+            JVMTI_THREAD_STATE_WAITING_INDEFINITELY;
+  }
+  return state;
+}
```
alexmenkov, line 768: You don't need to call get_thread_state_base in case "passive carrier thread":

    if (is_passive_carrier_thread(jt, thread_oop)) {
      return JVMTI_THREAD_STATE_ALIVE | JVMTI_THREAD_STATE_WAITING
             | JVMTI_THREAD_STATE_WAITING_INDEFINITELY;
    }
    return get_thread_state_base(thread_oop, jt);

sspitsyn: Thanks. Yes, noticed it. :) Fixed now.

src/hotspot/share/prims/jvmtiEnvBase.cpp
alexmenkov: `is_carrying_carrier_thread`? a bit artificial, but it's a carrier thread and it's carrying a virtual thread

src/hotspot/share/prims/jvmtiEnvBase.hpp
```diff
@@ -382,7 +382,8 @@ class JvmtiEnvBase : public CHeapObj<mtInternal> {
   // get carrier thread last java vframe
   static javaVFrame* get_cthread_last_java_vframe(JavaThread* jt, RegisterMap* reg_map);
 
-  // get ordinary thread thread state
+  // get platform thread state
+  static jint get_thread_state_base(oop thread_oop, JavaThread* jt);
```
alexmenkov, line 386: maybe rename it to `get_platform_thread_state`?

alexmenkov (approved):

src/hotspot/share/prims/jvmtiEnvBase.cpp
sspitsyn: > is_carrying_carrier_thread? a bit artificial, but it's a carrier thread and it's carrying a virtual thread

I guess, your suggestion is `is_carrying_virtual_thread`. Is it right?
If so, I like this suggestion.

src/hotspot/share/prims/jvmtiEnvBase.hpp
sspitsyn: I was thinking about it. It will be inconsistent with`get_vthread_state`.
Ideally, then the `get_vthread_state` needs to be replaced with the `get_virtual_thread_state`.

src/hotspot/share/prims/jvmtiEnvBase.cpp
alexmenkov: > > is_carrying_carrier_thread? a bit artificial, but it's a carrier thread and it's carrying a virtual thread
> 
> I guess, your suggestion is `is_carrying_virtual_thread`. Is it right? If so, I like this suggestion.

Up to you. I think any of this names is better than is_passive_carrier_thread.

plummercj (approved):

src/hotspot/share/prims/jvmtiEnvBase.cpp
AlanBateman: > I guess, your suggestion is `is_carrying_virtual_thread`. Is it right? If so, I like this suggestion.

Good, I think will be easy to understand at the use sites.

sspitsyn: Thank you, Alan and Alex.

src/hotspot/share/prims/jvmtiEnvBase.cpp
```diff
@@ -1728,7 +1741,7 @@ JvmtiEnvBase::suspend_thread(oop thread_oop, JavaThread* java_thread, bool singl
   // An attempt to handshake-suspend a passive carrier thread will result in
```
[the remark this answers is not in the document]
sspitsyn: Thanks. Alan. Will do this cleanup when there is a chance.

PR #8490, opened 2022-05-01
8277090 :  jsr166 refresh for jdk19

DougLea:
This is the jsr166 refresh for jdk19. See https://bugs.openjdk.java.net/browse/JDK-8285450 and https://bugs.openjdk.java.net/browse/JDK-8277090

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed (1 reviews required, with at least 1 reviewer)
- [x] Change requires a CSR request to be approved

### Issues
 * [JDK-8277090](https://bugs.openjdk.java.net/browse/JDK-8277090): jsr166 refresh for jdk19
 * [JDK-8285450](https://bugs.openjdk.java.net/browse/JDK-8285450): jsr166 refresh for jdk19 (**CSR**)

### Reviewers
 * [Alan Bateman](https://openjdk.java.net/census#alanb) (@AlanBateman - **Reviewer**) ⚠️ Review applies to [c3901b45](https://git.openjdk.java.net/jdk/pull/8490/files/c3901b45d11a4bfbdbd31fe678059c4dae21f40c)
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - **Reviewer**) ⚠️ Review applies to [c3901b45](https://git.openjdk.java.net/jdk/pull/8490/files/c3901b45d11a4bfbdbd31fe678059c4dae21f40c)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk pull/8490/head:pull/8490` \
`$ git checkout pull/8490`

Update a local copy of the PR: \
`$ git checkout pull/8490` \
`$ git pull https://git.openjdk.java.net/jdk pull/8490/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 8490`

View PR using the GUI difftool: \
`$ git pr show -t 8490`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.java.net/jdk/pull/8490.diff">https://git.openjdk.java.net/jdk/pull/8490.diff</a>

</details>

PR #14301, opened 2023-06-03
8288899: java/util/concurrent/ExecutorService/CloseTest.java failed with "InterruptedException: sleep interrupted"

DougLea:
Addresses Jdk 8288899 : java/util/concurrent/ExecutorService/CloseTest.java failed with "InterruptedException: sleep interrupted" and related issues.

This is a major ForkJoin update (and hard to review -- sorry) that finally addresses incompatibilities between ExecutorService and ForkJoinPool (which claims to implement it), with the goal of avoiding continuing bug reports and incompatibilities. Doing this required reworking internal control to use phaser/seqlock-style versioning schemes (affecting nearly every method) that ensure consistent data structures and actions without requiring global synchronization or locking on every task execution that would massively degrade performance. The previous lack of a solution to this was the main reason for these incompatibilities.

---------
### Progress
- [x] Change requires CSR request [JDK-8309407](https://bugs.openjdk.org/browse/JDK-8309407) to be approved

### Issues
 * [JDK-8288899](https://bugs.openjdk.org/browse/JDK-8288899): java/util/concurrent/ExecutorService/CloseTest.java failed with "InterruptedException: sleep interrupted" (**Bug** - P3)
 * [JDK-8309407](https://bugs.openjdk.org/browse/JDK-8309407): java/util/concurrent/ExecutorService/CloseTest.java failed with "InterruptedException: sleep interrupted" (**CSR**)

### Reviewers
 * [Alan Bateman](https://openjdk.org/census#alanb) (@AlanBateman - **Reviewer**) ⚠️ Review applies to [005ea626](https://git.openjdk.org/jdk/pull/14301/files/005ea626532c6a9551f4a35c55f5f4495df19b31)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14301/head:pull/14301` \
`$ git checkout pull/14301`

`$ git checkout pull/14301` \
`$ git pull https://git.openjdk.org/jdk.git pull/14301/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14301`

`$ git pr show -t 14301`

</details>

<a href="https://git.openjdk.org/jdk/pull/14301.diff">https://git.openjdk.org/jdk/pull/14301.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14301#issuecomment-1576848724)


src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java
```diff
@@ -2482,59 +2773,61 @@ static int getSurplusQueuedTaskCount() {
...
-            for (int i = 0; i < n; ++i) {
-                WorkQueue q; Thread thread;
-                if ((q = qs[(r + i) & (n - 1)]) != null &&
-                    (thread = q.owner) != current && q.access != STOP) {
-                    for (ForkJoinTask<?> t; (t = q.poll(null)) != null; )
-                        ForkJoinTask.cancelIgnoringExceptions(t);
-                    if (thread != null && !thread.isInterrupted()) {
-                        q.forcePhaseActive();      // for awaitWork
+            for (int l = n; l > 0; --l, ++r) {
+                int j; WorkQueue q; Thread o; // cancel tasks; interrupt workers
```
turbanoff, line 2795: `int j` is unused

src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java
```diff
@@ -2708,19 +3001,17 @@ public ForkJoinPool(int parallelism,
...
         if (factory == null || unit == null)
             throw new NullPointerException();
+        int size = 1 << (33 - Integer.numberOfLeadingZeros(p - 1));
         this.parallelism = p;
         this.factory = factory;
         this.ueh = handler;
         this.saturate = saturate;
-        this.config = asyncMode ? FIFO : 0;
         this.keepAlive = Math.max(unit.toMillis(keepAliveTime), TIMEOUT_SLOP);
-        int corep = Math.clamp(corePoolSize, p, MAX_CAP);
```
turbanoff, line 2717: it it expected that `int corePoolSize` is now unused?

DougLea:
Just a keep-alive comment

src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java
```diff
@@ -395,27 +400,45 @@ public class ForkJoinPool extends AbstractExecutorService {
...
+     * in scans (see below) are always prefaced by a volatile read of
+     * runState (or equivalent constructions), ensuring that its state is
+     * current at the point it is used (which is all we require). To
      * simplify index-based operations, the array size is always a
      * power of two, and all readers must tolerate null slots.  Worker
-     * queues are at odd indices. Worker ids masked with SMASK match
-     * their index. Shared (submission) queues are at even
+     * queues are at odd indices. Worker phase ids masked with SMASK
+     * match their index. Shared (submission) queues are at even
      * indices. Grouping them together in this way simplifies and
```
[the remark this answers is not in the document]
DougLea: More-or-less. At top-level, both kinds of queues should be sampled with approx the same probability, which is simpler if they are all in the same array. But we also need to identify what kind they are without looking at them, leading to odd/even scheme. One disadvantage is that there are usually many fewer submission queues, so there can be a lot of wasted probes (null slots). But still cheaper than alternatives because of cache locality etc.

src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java
```diff
@@ -610,79 +678,111 @@ public class ForkJoinPool extends AbstractExecutorService {
...
+     * they are ignored, and 0 is an OK default.) The scan in method
      * helpJoin uses these markers to try to find a worker to help
-     * (i.e., steal back a task from and execute it) that could hasten
-     * completion of the actively joined task.  Thus, the joiner
-     * executes a task that would be on its own local deque if the
-     * to-be-joined task had not been stolen. This is a conservative
-     * variant of the approach described in Wagner & Calder
-     * "Leapfrogging: a portable technique for implementing efficient
-     * futures" SIGPLAN Notices, 1993
+     * (i.e., steal back a task from and execute it) that could makje
```
viktorklang-ora, line 721:
```suggestion
     * (i.e., steal back a task from and execute it) that could make
```

src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java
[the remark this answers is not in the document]
DougLea: (I also added some similar wording to the internal docs.)

src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java
DougLea: Thanks. Done.

src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java
```diff
@@ -1594,49 +1806,48 @@ final String nextWorkerThreadName() {
...
-                WorkQueue[] qs; int n;                  // find queue index
+                WorkQueue[] qs; int n;
                 if ((qs = queues) != null && (n = qs.length) > 0) {
-                    int k = n, m = n - 1;
-                    for (; qs[id &= m] != null && k > 0; id -= 2, k -= 2);
-                    if (k == 0)
-                        id = n | 1;                     // resize below
-                    w.phase = w.config = id | cfg;      // now publishable
-
+                    for (int k = n, m = n - 1;  ; id += 2) {
```
viktorklang-ora, line 1830: @DougLea Might make sense to create a constant each for worker-queue stride and submission-queue stride so it becomes easier to reason about the `2` here.

DougLea: There seem to be too many variants of queues array traversal loops to do this uniformly, but I did add some explanation:

     * alternatives. Other loops over the queues array vary in origin
     * and stride depending on whether they cover only submission
     * (even) or worker (odd) queues or both, and whether they require
     * randomness (in which case cyclically exhaustive strides may be
     * used).

test/jdk/java/util/concurrent/tck/JSR166TestCase.java
```diff
@@ -1670,10 +1680,12 @@ Thread newStartedThread(Action action) {
...
     void awaitTermination(Thread thread, long timeoutMillis) {
-        try {
-            thread.join(timeoutMillis);
-        } catch (InterruptedException fail) {
-            threadUnexpectedException(fail);
+        for (;;) { // ignore stray interrupts by test harness
+            try {
+                thread.join(timeoutMillis);
+                break;
+            } catch (InterruptedException ignore) {
```
viktorklang-ora, line 1687: @DougLea Shouldn't this deduct the wait-time for the next join if it gets interrupted? 🤔

DougLea: Ordinarily yes, but here it shouldn't matter: the timeout is just an arbitrary value to catch stuck tests, and sometimes waiting twice as long because of an interrupt is OK.

src/java.base/share/classes/java/util/concurrent/ForkJoinTask.java
```diff
@@ -899,8 +899,15 @@ public State state() {
...
-            throw new IllegalStateException();
+        int s = status;
+        if ((s & DONE) == 0)
+             throw new IllegalStateException("Task has not completed");
+        if ((s & ABNORMAL) != 0) {
+            if ((s & THROWN) != 0)
+                throw new IllegalStateException("Task completed with exception");
+            else
+                throw new IllegalStateException("Task was cancelled");
+        }
```
viktorklang-ora, line 910: 👍

DougLea:
Backport instructions:
1. change all methods with new Since tags to nonpublic
2. Comment/uncomment ForkJonPool.invokeAll according to instructions there (the new version fixes an ExecutorService signature incompatibility, at the price of a backwards incompatibility in throws clause.)
3. It is possible that some (incorrect) java/util/concurrent/tck tests will fail, in which case replace with new versions

AlanBateman (approved):
This is a large update to fix a really hard problem. I've been testing and running with these changes on and off for several months and they are stable now. The code is very dense and hard to read but the description at a top is very good/helpful.

The API changes are okay.  The source compatibility issue with dropping the override of the invokeAll combinator is unfortunate but combined with everything else it means that FJP now implements the ES contract and can work as a replacement for say TPE.

So I think this changes are good to go.

src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java
```diff
@@ -2874,7 +3149,12 @@ public <T> ForkJoinTask<T> submit(ForkJoinTask<T> task) {
      */
     @Override
     public <T> ForkJoinTask<T> submit(Callable<T> task) {
-        return poolSubmit(true, new ForkJoinTask.AdaptedCallable<T>(task));
+        ForkJoinTask<T> t =
+            (Thread.currentThread() instanceof ForkJoinWorkerThread) ?
```
viktorklang-ora, line 3153: @DougLea Won't we need to check if the submitting thread belongs to the same pool here?

src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java
```diff
@@ -2998,167 +3303,75 @@ public <T> List<Future<T>> invokeAll(Collection<? extends Callable<T>> tasks) {
...
-                if (!f.isDone()) {
-                    if (!timedOut)
-                        timedOut = !f.quietlyJoin(ns, TimeUnit.NANOSECONDS);
-                    if (timedOut)
-                        ForkJoinTask.cancelIgnoringExceptions(f);
-                    else
-                        ns = nanos - (System.nanoTime() - startTime);
-                }
-            }
+            for (int i = futures.size() - 1; i >= 0; --i)
```
viktorklang-ora, line 3324: @DougLea Might not be necessary for performance, but using an ArrayDeque and using `pollLast()` might be cleaner?

viktorklang-ora:
I gave the PR a final review just now, there was a few questions which might not warrant any changes.

I concur with the assessment that these changes should be good to go.

src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java
DougLea: No, but I can imagine this being the subject of a Puzzler someday, because either choice might be unexpected by someone! To maximize compatibility wrt incompatible specs (initial FJ and ES), calls to submit from workers are by default noninterruptible, even if they are client calls into another pool. The opposite choice might be more defensible but could break existing code.

src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java
DougLea: This was a good enough idea that I started doing it until realizing/remembering that ArrayDeque doesn't implement List so can't be used here (without making a copy). Oh well.

viktorklang-ora: @DougLea Ouch—I completely missed that it is also returned. Apologies :(

src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java
viktorklang-ora: @DougLea Ok, that's good to know. :+1:

src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java
```diff
@@ -2980,16 +3282,34 @@ public int setParallelism(int size) {
     }
 
     /**
-     * @throws NullPointerException       {@inheritDoc}
-     * @throws RejectedExecutionException {@inheritDoc}
+     * Uninterrupible version of {@code InvokeAll}. Executes the given
```
jddarcy, line 3285: Typo: "InvokeAll" vs "invokeAll".

src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java
techsy730: I was curious about this too, so I dug into the history of this.

So it seems that `corePoolSize` became "unused" and stopped affecting anything in OpenJDK 19, during the changes for the JSR166 refresh (Project Loom stuff I think).

It happened in
00e6c63cd12e3f92d0c1d007aab4f74915616ffb
Pull request: https://github.com/openjdk/jdk/pull/8536
  * Or even more specifically, the original commit in the original pull request: https://github.com/openjdk/jdk/pull/8490/commits/a2521708771059fb1537209284564d9d181a1be9

The changes were
* `ForkJoinPool.bounds` was changed from an `int` to a `long`
* `corePoolSize` got changed into the upper half of the now bigger `bounds`, previously being stuffed in the total thread count bits of `ForkJoinPool.ctl` (or more accurately, the negative of it) (See `TC_SHIFT`/`TC_MASK`, etc)
  * As a fun side note. Due to this hack of tracking being below the core pool count as having "negative total threads", this meant that prior to OpenJDK 19, `ForkJoinPool.getPoolSize()` could return a _negative_ number, when the number of threads was below the core pool thread count _and_ corePoolSize > parallelism.
* The only method that used `ForkJoinPool.bounds` field, `ForkJoinPool.tryCompensate`, was not updated to read the new bits core pool size bits of `bounds` to factor into its "should I make more threads?" logic

So in short, `corePoolSize` was moved to a new location, but logic wasn't updated to use the new location


In this pull request, the `bounds` variable was removed, and the `maxSpares` (maximumPoolSize  - parallelism), and the release count (see `RC_SHIFT`) bits were now being put into the `config` field instead.
  (Oh and this also means in OpenJDK 19 and above, `getPoolSize` should now always return a number >= 0)

And maybe there weren't enough "free to use" bits left over for `corePoolSize`? Or maybe @DougLea noticed that the corePoolThreads logic was effectively dead and didn't bother trying to find a new place to store it?

In any case, this was just formally removing the use of a feature that had been dead for a while (since OpenJDK 19).

PR #14303, opened 2023-06-04
8309408: Thread.sleep cleanup

AlanBateman:
Thread.sleep has had quite a bit of churn recently to support virtual threads, add sleep(Duration), a JFR event, and the change the underlying implementation to support sub-millis precision. I think the changes have settled down now so we can do some small cleanups that came up in PR discussions. The cleanups were kicked down the road as it requires tracking down faraway tests that depend on the stack depth and the names of internal methods. The two cleanups proposed here are:

1.  Add a private sleepNanos method that creates/commits the JFR event around the sleep, this avoids duplicate code in the 3 sleep methods.
2.  Rename JVM_Sleep to JVM_SleepNanos to make it clear that it takes the sleep time in nanoseconds, esp. when Thread.sleep's parameter is milliseconds.

---------
### Progress

### Issue
 * [JDK-8309408](https://bugs.openjdk.org/browse/JDK-8309408): Thread.sleep cleanup (**Enhancement** - P4)

### Reviewers
 * [David Holmes](https://openjdk.org/census#dholmes) (@dholmes-ora - **Reviewer**) ⚠️ Review applies to [ae6b531f](https://git.openjdk.org/jdk/pull/14303/files/ae6b531f66afb3d8e060d2125af5cdba1eb5396b)
 * [Stefan Karlsson](https://openjdk.org/census#stefank) (@stefank - **Reviewer**) ⚠️ Review applies to [ae6b531f](https://git.openjdk.org/jdk/pull/14303/files/ae6b531f66afb3d8e060d2125af5cdba1eb5396b)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14303/head:pull/14303` \
`$ git checkout pull/14303`

`$ git checkout pull/14303` \
`$ git pull https://git.openjdk.org/jdk.git pull/14303/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14303`

`$ git pr show -t 14303`

</details>

<a href="https://git.openjdk.org/jdk/pull/14303.diff">https://git.openjdk.org/jdk/pull/14303.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14303#issuecomment-1576113982)


test/hotspot/jtreg/vmTestbase/nsk/monitoring/share/ThreadController.java
```diff
@@ -656,7 +656,8 @@ public SleepingThread(ThreadController controller, String name, Log log, Threads
         expectedLength += 4;
 
         expectedMethods.add(Thread.class.getName() + ".sleep");
-        expectedMethods.add(Thread.class.getName() + ".sleep0");
+        expectedMethods.add(Thread.class.getName() + ".sleepNanos");
+        expectedMethods.add(Thread.class.getName() + ".sleepNanos0");
```
dholmes-ora, line 660: I'm surprised this test doesn't list `beforeSleep` and `afterSleep`.

dholmes-ora (approved):
Looks good!

Keeping these tests up-to-date is painful, but as you note this is hopefully stabilized now.

There is one potential, pre-existing, test omission noted below.

Thanks.

test/hotspot/jtreg/vmTestbase/nsk/monitoring/share/ThreadController.java
AlanBateman: > There is one potential, pre-existing, test omission noted below.
> I'm surprised this test doesn't list `beforeSleep` and `afterSleep`.

The monitoring/stress/thread tests will fail if they observe an unexpected method name in the stack trace. I don't think it can happen because the tests poll the thread state and for SleepingThread, it will sample the stack trace when the thread state is timed-wait. The beforeSleep/afterSleep methods won't in the stack trace when sleeping. It would be harmless to add them in that they aren't going to cause these tests to fail but might help with any further changes.

shipilev (commented):
I think we need to delay this until [JDK-8309361](https://bugs.openjdk.org/browse/JDK-8309361) is resolved, in case we would like to revert [JDK-8305092](https://bugs.openjdk.org/browse/JDK-8305092).

test/hotspot/jtreg/vmTestbase/nsk/monitoring/share/ThreadController.java
AlanBateman: > The following commit in loom heavily modified this file with a lot of added expected methods. There are other related tests with similar changes. I'm not so sure I understand the need for so many additions, and also why expectedLength is so out of sync with the number of added method. I don't believe this commit was reviewed individually, but was just part of the overall loom review when merge into jdk. Perhaps it should be revisited.

These tests aren't easy to read or maintain, it would be good to re-visit them. In some cases, the tests capture the stack trace asynchronously so the test needs to know about all code paths.

As regards ThreadController, used by the nsk/monitoring/stress/thread/straceXXX tests, the main thread waits at a barrier (a CountDownLatch) until all sleeping threads are ready to sleep. Once the main thread is released, it checks all the sleepers are in TIMED_WAITING state and samples their stack traces with the ThreadMXBean and related APIs. The test fails if there are frames corresponding to methods that the test doesn't know about. If a thread is sleeping then we shouldn't see frames for beforeSleep/afterSlee. My reading of these tests is that the main thread could poll a SleepingThread after it counts down and before it parks in sleep. It's doing an expensive ThreadMXBean::getAllThreadIds once released and that may explain why it hasn't been seen.

test/hotspot/jtreg/vmTestbase/nsk/jdwp/ThreadReference/OwnedMonitorsStackDepthInfo/ownedMonitorsStackDepthInfo001/ownedMonitorsStackDepthInfo001a.java
```diff
@@ -104,13 +104,13 @@ protected void init(String args[]) {
...
     static int adjustStackDepth(int depth) {
-        // The stack depth does not take into account the extra frame below Thread.sleep() (sleep0),
+        // The stack depth does not take into account the extra frames below Thread.sleep()
         // so we need to add it to the stack depth. See LockingThread.expectedDepth(), which is where the
         // depth is calculated.
         if (depth == -1) {
             return -1;
         } else {
-            return depth + 1;
+            return depth + 2;
```
stsypanov, line 113: Doesn't it change the logic?

AlanBateman:
> > I think we need to delay this until [JDK-8309361](https://bugs.openjdk.org/browse/JDK-8309361) is resolved, in case we would like to revert [JDK-8305092](https://bugs.openjdk.org/browse/JDK-8305092).
> 
> Okay, I won't integrate this until we see what the issue is. That said, if the sub-mills support needs to be reverted then I think we should keep the interface as nanos on the boundary between the VM and the libraries. The reason is that it's mostly nanos at the Java level now, meaning 2 out of the 3 sleep methods support sub-mills, and the virtual thread sleep and the JFR event are in nanos too.

JDK-8309361 doesn't have a conclusion yet but doesn't look like it's a JDK implementation issue or heading toward removing the support for sub-mills precision.  So I think we should integrate this cleanup and get it out of the way. Even if there are changes then I think it's better for the precision at the interface to be nanos.

PR #14304, opened 2023-06-04
JDK-8309410: [aarch64] Use pre-computed tables for logical immediates

tstuefe:
We compute the lookup tables for logical immediates on startup. We should use pre-computed tables instead.

If we do that, we don't even need the encoding->immediate lookup table, since that is only used during generation of the reverse lookup table. Since we hardcode the latter, we don't need to store the former.

I kept the old generator code around to test the hard-coded table (see gtest). To keep reviewing simple, the generator code itself lives still in the same place, mostly unchanged.

Note that we could shave off some more space for the reverse lookup table by storing immediate and encoding in separate arrays, since we now pay 16 bytes per entry due to alignment. We could also reduce the encoding from 32 bits to 16 bits. I did not do this to not overload this RFE.

Tests: manually ran tests on linux aarch64, OsX aarch64

---------
### Progress

### Issue
 * [JDK-8309410](https://bugs.openjdk.org/browse/JDK-8309410): [aarch64] Use pre-computed tables for logical immediates

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14304/head:pull/14304` \
`$ git checkout pull/14304`

`$ git checkout pull/14304` \
`$ git pull https://git.openjdk.org/jdk.git pull/14304/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14304`

`$ git pr show -t 14304`

</details>

<a href="https://git.openjdk.org/jdk/pull/14304.diff">https://git.openjdk.org/jdk/pull/14304.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14304#issuecomment-1575616532)


jdksjolen:
Hi Thomas,

The idea makes sense, but can't we do this with `constexpr` only? Looking at the intiailization code, this seems doable. Only issue would be sorting, as that'd have to be reimplemented as a contexpr function. No big table, no test that ensures that the table hasn't drifted away from code that's actually dead, seems like a big win!

Here's a sketch in constexpr which I think serves as a PoC that this could work: https://godbolt.org/z/o3PTbanYY

And for completeness here's the full source code replicated in this comment:

```c++
#include <cstdio>
struct LITableStruct {
    // couldn't bother with imports for sized (u)ints
    static const unsigned int tbl_sz = 1 << 13;
    int  LITable[tbl_sz];
    unsigned long InvLITable[tbl_sz];

    // We can dereference pointers, compare and set them in constexpr
    // (in other words, we can write a sorting routine)
    constexpr void comp_and_swap(int* a, int* b) {
        if (*a < *b) {
            *a = *b;
        }
    }
    // Magic incantation!
    constexpr LITableStruct()
    : LITable(), InvLITable() {
        // comp and swap compiles
        int a[2] = {0, 1};
        comp_and_swap(&a[0], &a[1]);
        comp_and_swap(&LITable[0], &LITable[255]);

        // loop and assign no problem
        for (unsigned int idx = 0; idx < tbl_sz; idx++) {
            LITable[idx] = LITable[0] + 255;
            // just an if statement depending on some other state
            // bit twiddling and other nonsense
            if (LITable[0] == 1337 || LITable[67] & 5 | 37) {
                LITable[2500] = 73;
            }
        }
    }
};

static struct LITableStruct my_table{};
int main() {
    printf("Wow, an int: %d\n", my_table.LITable[0]);
}
```

tstuefe:
Hi @jdksjolen,

> The idea makes sense, but can't we do this with constexpr only?

Maybe , but
- I don't have much more time to spend on this, this was a sunday afternoon thing and is useful in its current form, but shouldn't spiral out into a C++ research project (since I would have to babysit every compiler to make sure it generates what I assume it generates)
- I don't see why: we'd try to cajole the C++ compiler into generating a table for us that we just as well may write out plainly. Easier to understand, faster to compile, less reliance on the abilities of the C++ compiler. The end result is identical - a hard coded table in the text segment.
- My variant is easy to review, since I did not change the old generator code. We could throw the testing code out at some point in the future.

Cheers, Thomas

jdksjolen:
> Hi @jdksjolen,
> 
> > The idea makes sense, but can't we do this with constexpr only?
> 
> Maybe , but I don't see why: we'd try to cajole the C++ compiler into generating a table that we just as well may write out plainly. Easier to understand, faster to compile, less reliance on the abilities of the C++ compiler. The result is identical - a hard-coded table in the text segment.
> 
> Cheers, Thomas

Hi Thomas,

Especially good point on ensuring that all compilers do what we expect them to do, I didn't consider that. I'm just Sunday reviewing today, I'll give you a complete review  tomorrow.

Cheers,
Johan

theRealAph:
We don't need a lookup table, because the boolean function is invertible. LLVM does it, for example.

I never bothered to do it because it doesn't seem worth the effort.

theRealAph:
Whether this is a win depends on a bunch of things that are hard to predict, such as the time take to page in the table from mass storage. If we're really going to revisit this, and I'm not at all sure we should, we should be thorough.

Can we compute the immediates directly? Yes.
Can we replace the binary search with a hash table? Yes.
If we have a precomputed table, must we include the program used to generate it? Yes.
Is the extra code complexity justified? Don't know.

tstuefe:
> Whether this is a win depends on a bunch of things that are hard to predict, such as the time take to page in the table from mass storage. If we're really going to revisit this, and I'm not at all sure we should, we should be thorough.

The table is ~64K, and we only use it very sparsely, since we only look up a small subset of possible encodings. Computing the table at runtime involves paging in the whole space for both tables to populate them, apart from the bit fiddling. I assumed that would be more expensive. Measuring this would be an effort though.

> If we're really going to revisit this, and I'm not at all sure we should, we should be thorough.

My proposal was an incremental improvement and it deliberately does not touch the table format and the decoding. If you think it is not worth it, I'll close it - this was just a small improvement and there is no time for more.

theRealAph:
> My proposal was an incremental improvement and it deliberately does not touch the table format and the decoding. If you think it is not worth it, I'll close it - this was just a small improvement and there is no time for more.

I really don't know one way or another. I am sure, however, that for this patch to be complete it _must_ include the code used to generate the table, and that's a significant bump in complexity.

tstuefe:
> > My proposal was an incremental improvement and it deliberately does not touch the table format and the decoding. If you think it is not worth it, I'll close it - this was just a small improvement and there is no time for more.
> 
> I really don't know one way or another. I am sure, however, that for this patch to be complete it _must_ include the code used to generate the table, and that's a significant bump in complexity.

True. We can add this code separately from hotspot though, e.g. as python script.

But since I don't have any more time to spend, I'll shelve it for now. Also, maybe there is a way to compute the encoding from the immediate bit pattern; it cannot be that hard. Maybe that's the better way to go.

Anyway, thanks for looking at this!

PR #14305, opened 2023-06-05
JDK-8295071: Spec Clarification : ClassFileFormatVersion: System property java.class.version | Java class format version number

jddarcy:
Document existing behavior.

Please also review the CSR https://bugs.openjdk.org/browse/JDK-8309415

---------
### Progress
- [x] Change requires CSR request [JDK-8309415](https://bugs.openjdk.org/browse/JDK-8309415) to be approved

### Issues
 * [JDK-8295071](https://bugs.openjdk.org/browse/JDK-8295071): Spec Clarification : ClassFileFormatVersion: System property java.class.version | Java class format version number (**Bug** - `"4"`)
 * [JDK-8309415](https://bugs.openjdk.org/browse/JDK-8309415): Spec Clarification : ClassFileFormatVersion: System property java.class.version | Java class format version number (**CSR**)

### Reviewers
 * [Iris Clark](https://openjdk.org/census#iris) (@irisclark - **Reviewer**) ⚠️ Review applies to [33bb646b](https://git.openjdk.org/jdk/pull/14305/files/33bb646b4ea15e29fb698c69838837583c21dfe4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14305/head:pull/14305` \
`$ git checkout pull/14305`

`$ git checkout pull/14305` \
`$ git pull https://git.openjdk.org/jdk.git pull/14305/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14305`

`$ git pr show -t 14305`

</details>

<a href="https://git.openjdk.org/jdk/pull/14305.diff">https://git.openjdk.org/jdk/pull/14305.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14305#issuecomment-1575900807)


src/java.base/share/classes/java/lang/System.java
```diff
@@ -753,7 +753,9 @@ public static native void arraycopy(Object src,  int  srcPos,
      * <tr><th scope="row">{@systemProperty java.specification.name}</th>
      *     <td>Java Runtime Environment specification  name</td></tr>
      * <tr><th scope="row">{@systemProperty java.class.version}</th>
-     *     <td>Java class format version number</td></tr>
+     *     <td>Greatest Java class file format version recognized by the runtime as {@code "MAJOR.MINOR"}
+     *     where {@link java.lang.reflect.ClassFileFormatVersion#major() MAJOR} and {@code MINOR}
+     *     are both formatted as decimal integers</td></tr>
```
AlanBateman, line 758: I'm not sure about using the adjective "Greatest" here, only because "Greatest ... recognized" reads a bit like recognition of greatness when it really wants to convey that it's the highest version recognized.

Does the wording need to take account of running with preview features enabled? 

The table uses "Java runtime" rather than "runtime" in the description of some of the other properties.

In passing, should the descriptions of RELEASE_xxx in ClassFileFormatVersion say "The highest version recognized" rather than the "The version recognized"?

jddarcy: Update PR and CSR to use the wording:

"Latest Java class file format version recognized by the Java runtime as "MAJOR.MINOR" where MAJOR and MINOR are both formatted as decimal integers"

While it wouldn't be unreasonable to assume that when running under --enable-preview that the minor version was 65535, it is printed out as 0 with or without preview being enabled.

I didn't want to broach preview-ness as part of this update to either promise or document the current behavior.

mlchung: Related to Alan's comment on RELEASE_xxx in ClassFileFormatVersion, perhaps the class description can clarify that a class file version has major version number `M` and minor version number `m` and a Java Virtual Machine implementation can support a class file format of version v if and only if v lies in some contiguous range Mi.0 ≤ v ≤ Mj.m.   RELEASE_xxx represents the major version number. 

That way, the description of RELEASE_xxx can remain to say "The version recognized...".

irisclark: The decimal integer should be unsigned and without leading zeros.  A regex similar to that provided by Runtime.Version would make things clear.  I think that Mandy's suggestion would be a reasonable alternative as the Java class file version number isn't truly arbitrary.

irisclark (approved):

src/java.base/share/classes/java/lang/System.java
jddarcy: > The decimal integer should be unsigned and without leading zeros. A regex similar to that provided by Runtime.Version would make things clear. I think that Mandy's suggestion would be a reasonable alternative as the Java class file version number isn't truly arbitrary.

A quick note on API philosophy, in its current version the PR proposes to go from very imprecise information:

"Java class format version number"

to to moderate level of precision:

"Latest Java class file format version recognized by the Java runtime as "MAJOR.MINOR" where MAJOR and MINOR are both formatted as decimal integers"

Especially given the terse description of most of the property keys, I don't think grammar-level detailing is necessary in this context.

For example, to me the specification of, say, Method.toString is unhelpful to the reader since the gist of the semantics are overwhelmed with details as if the primary consumer was a parser rather than a human reader.

AlanBateman: > "Latest Java class file format version recognized by the Java runtime as "MAJOR.MINOR" where MAJOR and MINOR are both formatted as decimal integers"

Okay, and linking to ClassFileFormatVersion.latest() is good.

Again, just to lobby for putting an adjective into the ClassFileFormatVersion.RELEASE_XX descriptions to make it clearer than the value is the latest version recognized rather than single version that the wording currently suggests.

> While it wouldn't be unreasonable to assume that when running under --enable-preview that the minor version was 65535, it is printed out as 0 with or without preview being enabled.
> 
> I didn't want to broach preview-ness as part of this update to either promise or document the current behavior.

You're right, I shouldn't have brought this up here.

jddarcy: > > "Latest Java class file format version recognized by the Java runtime as "MAJOR.MINOR" where MAJOR and MINOR are both formatted as decimal integers"
> 
> Okay, and linking to ClassFileFormatVersion.latest() is good.
> 
> Again, just to lobby for putting an adjective into the ClassFileFormatVersion.RELEASE_XX descriptions to make it clearer than the value is the latest version recognized rather than single version that the wording currently suggests.

Okay, for the constants for 1.2 and later, I switched to "The version introduced by...".

The 1.0.2 and 1.1 versions of the platform use the same class file format version.

Explanatory apiNote for ClassFileFormatVersion also added.

I don't think those are spec changes per se, so I didn't update the CSR to include them.

PR #14310, opened 2023-06-05
8304878: ConcurrentModificationException in javadoc tool

pavelrappo:
Although its name might suggest otherwise, java.util.ConcurrentModificationException (CME) is not necessarily thrown in multithreaded context. For example, a map might throw CME if it detects modification in a midst of an operation which is supposed to be exclusive:

    jshell> Map<Integer, Integer> m = new HashMap<>();
       ...> m.computeIfAbsent(0, key -> m.put(1, 2));
    m ==> {}
    |  Exception java.util.ConcurrentModificationException
    |        at HashMap.computeIfAbsent (HashMap.java:1229)
    |        at (#2:1)

Somewhat similar happens in the reproducer: a computation of a tree path for a package element eventually modifies the very map that the result of that computation is to be put in. The map detects that and throws CME.

Here are more details. The tree path is to be found for the `bar` package element. The tree path is computed and put into the cache in one go (HashMap.computeIfAbsent). Usually that works fine, but not this time. Because the `bar` package resides in an implicitly referred module (test.bar), that package is discovered late. The computation triggers the `bar` package completion/parsing, which eventually calls back to JavadocEnter.visitTopLevel, which in turn puts an alternatively computed tree path for the `bar` package into that same cache. When the call stack returns to computeIfAbsent, that cache modification is detected and CME is thrown.

I can see three obvious ways of fixing this bug:

  1. Replace computeIfAbsent with an equivalent combination of simpler methods.
  2. Use a map whose computeIfAbsent can withstand non-exclusive modifications.
  3. Restructure the code so it avoids such modifications altogether.

(3) Is quite risky/complicated and not really needed at this time. (2) While it's unlikely to have noticeable performance overhead, ConcurrentHashMap feels like symptomatic treatment that is also out of place given that JavaDoc is single-threaded.

(1) seems the most reasonable. One potential problem with (1) though is that future refactoring may reintroduce the issue again, but in a way that the regression test won't catch.

---------
### Progress

### Issue
 * [JDK-8304878](https://bugs.openjdk.org/browse/JDK-8304878): ConcurrentModificationException in javadoc tool

### Reviewers
 * [Jonathan Gibbons](https://openjdk.org/census#jjg) (@jonathan-gibbons - **Reviewer**) ⚠️ Review applies to [5fa14506](https://git.openjdk.org/jdk/pull/14310/files/5fa1450657f2fd8bba54dd9a40a3183e29868ab8)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14310/head:pull/14310` \
`$ git checkout pull/14310`

`$ git checkout pull/14310` \
`$ git pull https://git.openjdk.org/jdk.git pull/14310/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14310`

`$ git pr show -t 14310`

</details>

<a href="https://git.openjdk.org/jdk/pull/14310.diff">https://git.openjdk.org/jdk/pull/14310.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14310#issuecomment-1576567498)


liach:
Such CMEs do happen often with structures where one map entry may delegate its value to be computed from another entry.

For the 3 approaches, I think 1 is the best for now, that we should never use the functional computation methods if one map entry may query and potentially request another map entry. 2 appears meaningless, for such a map will have performance penalties and slower than regular put. For 3, I think a way to approach is to encapsulate such vulnerable maps (with recursive value computation dependencies) and group their modification (including lazy computation) methods altogether: for example, we might prohibit modifications to the returned map of getElementToTreePath.

test/langtools/jdk/javadoc/doclet/testLatePackageDiscovery/TestLatePackageDiscovery.java
```diff
@@ -0,0 +1,67 @@
...
+        tb.writeJavaFiles(src.resolve("test.foo"),
+                "module test.foo { exports foo; requires test.bar; }",
+                "package foo;",
+                "package foo; public class Foo extends bar.Bar { }");
+
+        javadoc("-d", base.resolve("out").toString(),
+                "--module-source-path", src.toString(),
+                "--module", "test.foo",
+                "-nodeprecated");
+        checkExit(Exit.OK);
```
jonathan-gibbons, line 65: (Optional)
Since this is about ensuring we don't get CME again,
either explicitly turn on exception checking (even though it is on by default)
or insert a comment noting than exception checking is on by default

jonathan-gibbons (approved):
Nice fix. Thanks for the detailed explanation and suggested solutions

test/langtools/jdk/javadoc/doclet/testLatePackageDiscovery/TestLatePackageDiscovery.java
pavelrappo: Done in db1509b389f.

pavelrappo:
While the tests are running one last time before integration, I'll take a moment to note what might not have been obvious from the diff: computeIfAbsent in the original code was overkill.

If we reached computeIfAbsent, the map was guaranteed NOT to contain the key. Using computeIfAbsent was either a typo/remnant from something else or stemmed from a misconception of how it works. The conditionality of that compound operation does not generally apply to checking whether the key it is about to enter has already somehow appeared as the result of call to the mapping function.

pavelrappo:
/integrate

Test are green.

PR #14312, opened 2023-06-05
JDK-8309471: Limit key characters in static index pages

hns:
Please review a simple change to avoid adding section index items (added in JDK-8286470) from appearing in the static index pages where the tend to cause trouble. 

The change is implemented by making `IndexItem.isTagItem()` behave true to its specification again and only return `true` if it actually represents a JavaDoc tag, and add a new `IndexItem.isSectionItem()` method for the new section index items. The new method is used to make sure section links are not listed in the index pages.

---------
### Progress

### Issue
 * [JDK-8309471](https://bugs.openjdk.org/browse/JDK-8309471): Limit key characters in static index pages (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14312/head:pull/14312` \
`$ git checkout pull/14312`

`$ git checkout pull/14312` \
`$ git pull https://git.openjdk.org/jdk.git pull/14312/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14312`

`$ git pr show -t 14312`

</details>

<a href="https://git.openjdk.org/jdk/pull/14312.diff">https://git.openjdk.org/jdk/pull/14312.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14312#issuecomment-1576857895)

### DIFF
--- a/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/IndexBuilder.java
+++ b/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/IndexBuilder.java
@@ -300,7 +300,14 @@ public class IndexBuilder {
     }
 
     private static Character keyCharacter(String s) {
-        return s.isEmpty() ? '*' : Character.toUpperCase(s.charAt(0));
+        // Use first valid java identifier start character as key,
+        // or '*' for strings that do not contain one.
+        for (int i = 0; i < s.length(); i++) {
+            if (Character.isJavaIdentifierStart(s.charAt(i))) {
+                return Character.toUpperCase(s.charAt(i));
+            }
+        }
+        return '*';
     }
 
 
--- a/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/IndexWriter.java
+++ b/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/IndexWriter.java
@@ -356,7 +356,7 @@ public class IndexWriter extends HtmlDocletWriter {
         content.add(new HtmlTree(TagName.BR));
         List<Content> pageLinks = Stream.of(IndexItem.Category.values())
                 .flatMap(c -> mainIndex.getItems(c).stream())
-                .filter(i -> !(i.isElementItem() || i.isTagItem()))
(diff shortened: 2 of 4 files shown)


jonathan-gibbons:
I think it is (still) a good default to be able to say that the static and interactive index have the same content.

I agree that headings that begin with a perceived section number are somewhat problematic.   While it might be possible to detect/skip a leading dotted-number "word", I think a simpler algorithm to index under the first letter is simpler to explain and mostly good enough.    (It would only be less than ideal if the first word is an alphanumeric sequence like `802.11az`

hns:
I updated the pull request to limit key characters to valid Java identifier start characters in the static index pages. The asterisk character `*` is used for index items that do not contain such a character. This is in line with the previous behaviour where asterisk was used for empty strings. The test was updated to check the exact HTML for the index page.

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/IndexBuilder.java
```diff
@@ -300,7 +300,14 @@ private boolean shouldIndex(Element element) {
     }
 
     private static Character keyCharacter(String s) {
-        return s.isEmpty() ? '*' : Character.toUpperCase(s.charAt(0));
+        // Use first valid java identifier start character as key,
+        // or '*' for strings that do not contain one.
+        for (int i = 0; i < s.length(); i++) {
+            if (Character.isJavaIdentifierStart(s.charAt(i))) {
+                return Character.toUpperCase(s.charAt(i));
```
jonathan-gibbons, line 307: maybe use the overload that takes a locale, and use `Locale.ROOT`

jonathan-gibbons (approved):
Nice solution

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/IndexBuilder.java
hns: I just learned that `Character.toUpperCase` does not have a locale-sensitive overload, we'd have to use `String.toUpperCase` for that. I don't think it's that important in this case, so I'm leaving the code as it is.

PR #14316, opened 2023-06-05
8301379: Verify TLS_ECDH_* cipher suites cannot be negotiated

mpdonova:
This PR implements a test to verify that Java will not negotiate TLS connections when one side of the connection requests disabled cipher suites.

---------
### Progress

### Issue
 * [JDK-8301379](https://bugs.openjdk.org/browse/JDK-8301379): Verify TLS_ECDH_* cipher suites cannot be negotiated (**Enhancement** - P4)

### Reviewers
 * [Sean Mullan](https://openjdk.org/census#mullan) (@seanjmullan - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14316/head:pull/14316` \
`$ git checkout pull/14316`

`$ git checkout pull/14316` \
`$ git pull https://git.openjdk.org/jdk.git pull/14316/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14316`

`$ git pr show -t 14316`

</details>

<a href="https://git.openjdk.org/jdk/pull/14316.diff">https://git.openjdk.org/jdk/pull/14316.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14316#issuecomment-1577280883)


test/jdk/javax/net/ssl/TLS/TLSWontNegotiateDisabledCipherAlgos.java
```diff
@@ -0,0 +1,166 @@
...
+
+public class TLSWontNegotiateDisabledCipherAlgos {
+    public static void main(String [] args) throws Exception {
+        boolean useDisabledAlgo = Boolean.parseBoolean(args[1]);
+        if (useDisabledAlgo) {
+            Security.setProperty("jdk.tls.disabledAlgorithms", "");
+        }
+
+        if (args[0].equals("server")) {
+            try(TLSServer server = new TLSServer(useDisabledAlgo, args[2])) {
```
seanjmullan, line 83: Nit, space after `try` (also on line 98 and 152).

test/jdk/javax/net/ssl/TLS/TLSWontNegotiateDisabledCipherAlgos.java
```diff
@@ -0,0 +1,166 @@
...
+            if (useDisableAlgo) {
+                socket.setEnabledCipherSuites(new String[]{cipherSuite});
+            }
+        }
+
+        public void run() throws IOException {
+            try {
+                socket.getOutputStream().write("SECRET MESSAGE".getBytes(StandardCharsets.UTF_8));
+            } catch (SSLHandshakeException exc) {
+                // handshake failures are expected
```
seanjmullan, line 121: Can you inspect the handshake failure message to make sure it is failing for the right reason as you do for the server case?

test/jdk/javax/net/ssl/TLS/TLSWontNegotiateDisabledCipherAlgos.java
```diff
@@ -0,0 +1,166 @@
...
+            SSLContext context = createClientSSLContext();
+            socket = (SSLSocket)context.getSocketFactory().createSocket("localhost", portNumber);
+            if (useDisableAlgo) {
+                socket.setEnabledCipherSuites(new String[]{cipherSuite});
+            }
+        }
+
+        public void run() throws IOException {
+            try {
+                socket.getOutputStream().write("SECRET MESSAGE".getBytes(StandardCharsets.UTF_8));
```
seanjmullan, line 119: Should this fail if this handshake passes?

test/jdk/javax/net/ssl/TLS/TLSWontNegotiateDisabledCipherAlgos.java
```diff
@@ -0,0 +1,166 @@
...
+import java.util.List;
+
+/*
+ * @test id=Server
+ * @bug 8301379
+ * @summary Verify that Java will not negotiate disabled cipher suites when the
+ * other side of the connection requests them.
+ *
+ * @library /javax/net/ssl/templates
+ * @run main/othervm TLSWontNegotiateDisabledCipherAlgos server true TLS_ECDH_ECDSA_WITH_AES_256_GCM_SHA384
```
seanjmullan, line 39: Consider collapsing all of these run lines into one run line which sets all of the disabled cipher suites together in one handshake. I think you still are testing that none of the suites are negotiated.

test/jdk/javax/net/ssl/TLS/TLSWontNegotiateDisabledCipherAlgos.java
```diff
@@ -0,0 +1,166 @@
...
+import java.net.Socket;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+import java.security.Security;
+import java.util.List;
+
+/*
+ * @test id=Server
+ * @bug 8301379
+ * @summary Verify that Java will not negotiate disabled cipher suites when the
```
seanjmullan, line 35: This test could be enhanced to test all of the other suites that are disabled. I think that might be a good idea, to centralize that. Although you could do it as a follow-on RFE.

test/jdk/javax/net/ssl/TLS/TLSWontNegotiateDisabledCipherAlgos.java
mpdonova: I updated the code to throw an exception here. It's useful for logging but if the handshake is successful and data is transferred the server will report the error.

test/jdk/javax/net/ssl/TLS/TLSWontNegotiateDisabledCipherAlgos.java
mpdonova: I added a check for the failure message.

test/jdk/javax/net/ssl/TLS/TLSWontNegotiateDisabledCipherAlgos.java
mpdonova: I added all of the disable ciphers that are listed in DisabledAlgorithms. Let me know if this isn't what you were thinking. Thanks!

seanjmullan: Yes, this is what I was thinking. There is another test `test/jdk/javax/net/ssl/ciphersuites/DisabledAlgorithms.java` that is very similar to this test. Can you compare these two tests and consider removing that test, or combining them if there are other things that test is testing that this one isn't?

mpdonova: The tests look similar but are definitely testing different code paths. I didn't combine the tests because they're different enough in how they work that making them one test would be messier than having two tests. I did, however, relocate my new test to the same directory as `DisabledAlgorithms` and changed it to use the same list of disabled ciphers so when we need to add to the list, it's covered by both tests. Also, while there, I updated DisabledAlgorithms to use the SSL test template classes instead of the binary {key,trust}store files.

seanjmullan: Ok, I suggest adding a better comments to the `DisabledAlgorithms` test to note that they are different. The one difference I see is that `DisabledAlgorithms` is also testing that the disabled suites can still be used if the security properties are empty. Other than that, they look somewhat similar but using different templates for setting things up.

seanjmullan (approved):

PR #14317, opened 2023-06-05
8301341: LinkedTransferQueue does not respect timeout for poll()

DougLea:
This update addresses performance issues across both LinkedTransferQueue and SynchronousQueue by creating a common basis for implementation across them (mainly in LinkedTransferQueue).  Pasting from internal doc summary of changes:
     * * Class DualNode replaces Qnode, with fields and methods
     *   that apply to any match-based dual data structure, and now
     *   usable in other j.u.c classes. in particular, SynchronousQueue.
     * * Blocking control (in class DualNode) accommodates
     *   VirtualThreads and (perhaps virtualized) uniprocessors.
     * * All fields of this class (LinkedTransferQueue) are
     *   default-initializable (to null), allowing further extension
     *   (in particular, SynchronousQueue.Transferer)
     * * Head and tail fields are lazily initialized rather than set
     *   to a dummy node, while also reducing retries under heavy
     *   contention and misorderings, and relaxing some accesses,
     *   requiring accommodation in many places (as well as
     *   adjustments in WhiteBox tests).

---------
### Progress

### Issues
 * [JDK-8301341](https://bugs.openjdk.org/browse/JDK-8301341): LinkedTransferQueue does not respect timeout for poll() (**Bug** - P4)
 * [JDK-8300663](https://bugs.openjdk.org/browse/JDK-8300663): java/util/concurrent/SynchronousQueue/Fairness.java failed with "Error: fair=true i=0 j=1" (**Bug** - P3)
 * [JDK-8267502](https://bugs.openjdk.org/browse/JDK-8267502): JDK-8246677 caused 16x performance regression in SynchronousQueue (**Bug** - P3)

### Reviewers
 * [Alan Bateman](https://openjdk.org/census#alanb) (@AlanBateman - **Reviewer**) ⚠️ Review applies to [eec7e491](https://git.openjdk.org/jdk/pull/14317/files/eec7e4915a70d4b8803374794f37abeeeb2e7360)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14317/head:pull/14317` \
`$ git checkout pull/14317`

`$ git checkout pull/14317` \
`$ git pull https://git.openjdk.org/jdk.git pull/14317/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14317`

`$ git pr show -t 14317`

</details>

<a href="https://git.openjdk.org/jdk/pull/14317.diff">https://git.openjdk.org/jdk/pull/14317.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14317#issuecomment-1577305984)


AlanBateman (approved):
This looks okay to me, it was a bit easier than I expected.

The only comment I have is an interrupt while spinning won't be noticed until count goes to 0, that should be rare enough that it doesn't matter. We might have to tune SPIN_FOR_TIMEOUT_THRESHOLD at some point for virtual threads.

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
```diff
@@ -372,6 +398,15 @@
...
     static final long SPIN_FOR_TIMEOUT_THRESHOLD = 1023L;
 
+    /**
+     * The number of times to spin for a non-virtual thread before
+     * blocking when a node is apparently the first waiter in the
+     * queue.  See above for explanation. Must be a power of two. The
+     * value is empirically derived -- it works pretty well across a
+     * variety of processors, numbers of CPUs, and OSes.
+     */
+    private static final int FRONT_SPINS   = 1 << 7;
```
turbanoff, line 408: nit
```suggestion
    private static final int FRONT_SPINS = 1 << 7;
```

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
```diff
@@ -632,55 +672,55 @@ private E xfer(E e, boolean haveData, int how, long nanos) {
...
+            } else if (w.isInterrupted() || (timed && nanos <= 0L)) {
+                if (s.casItem(e, s.isData ? null : s)) {
+                    unsplice(pred, s);          // cancel and unlink
                     return e;
+                }                               // else retry
+            } else if (spins < 0) {             // initialize spins
+                if (w.isVirtual() || pred == null)
+                    spins = 0;                  // don't spin
+                else if (pred.isData != isData || pred.isMatched())
+                    spins =  FRONT_SPINS;
```
turbanoff, line 701: nit
```suggestion
                    spins = FRONT_SPINS;
```

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
```diff
@@ -107,7 +108,11 @@ public class LinkedTransferQueue<E> extends AbstractQueue<E>
      * http://www.cs.rochester.edu/u/scott/papers/2009_Scherer_CACM_SSQ.pdf)
      * additionally arrange that threads enqueuing unmatched data also
      * block.  Dual Transfer Queues support all of these modes, as
-     * dictated by callers.
+     * dictated by callers. All enqueue/dequeue operations can be
+     * handled by a single method (here, "xfer") with parameters
+     * indicating whether to act as some form of offer, put, poll,
+     * take, or transfer (each possibly with timeout), as described
+     * bwlow.
```
viktorklang-ora, line 115: @DougLea Small typo—s/bwlow/below

DougLea: ... which I just did, so hopefully easier to read through

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
```diff
@@ -174,16 +174,17 @@ public class LinkedTransferQueue<E> extends AbstractQueue<E>
...
-     * usually implemented by checking equality of traversal pointers.
-     * Because CASes updating fields may fail and threads attempting
-     * to do so may stall, the actual slack may exceed targeted
-     * slack. To reduce the consequent staleness impact, threads may
-     * help update (method unslacken) when traversal lengths exceed an
-     * imposed limit of MAX_SLACK.
+     * traversal chains and out-of-order updates, while smaller values
+     * increase CAS contention and overhead. Using the smallest
+     * non-zero value of one is both simple and empirically a good
+     * choice in most applicatkions.  The slack value is hard-wired: a
```
viktorklang-ora, line 180: s/applicatkions/applications

viktorklang-ora (commented):
Need to digest this a bit. Will give more concrete feedback ASAP :)

src/java.base/share/classes/java/util/concurrent/SynchronousQueue.java
```diff
@@ -98,718 +101,144 @@ public class SynchronousQueue<E> extends AbstractQueue<E>
...
-                            LockSupport.unpark(waiter);
-                        return (x != null) ? (E)x : e;
-                    }
-                }
-            }
-        }
+    /**
+     * The transferer. (See below about serialization.)
+     */
+    private transient final Transferer<E> transferer;
```
turbanoff, line 232: let's use blessed modifiers order
```suggestion
    private final transient Transferer<E> transferer;
```

src/java.base/share/classes/java/util/concurrent/SynchronousQueue.java
```diff
@@ -98,718 +101,144 @@ public class SynchronousQueue<E> extends AbstractQueue<E>
...
-                         dn != d &&                //   that is on list
-                         dp.casNext(d, dn)))       // d unspliced
-                        casCleanMe(dp, null);
-                    if (dp == pred)
-                        return;      // s is already saved node
-                } else if (casCleanMe(null, pred))
-                    return;          // Postpone cleaning s
-            }
-        }
+    private transient final boolean fair;
```
turbanoff, line 234: nit
```suggestion
    private final transient boolean fair;
```

src/java.base/share/classes/java/util/concurrent/SynchronousQueue.java
DougLea: Thanks. Done

src/java.base/share/classes/java/util/concurrent/SynchronousQueue.java
DougLea: Thanks. Done.

DougLea:
A backport note: This set of changes may be applied jdk19+ by removing only the lines (currently L423-424) in LinkedTransferQueue:
            if w.isVirtual())             // don't spin
                spin = false;

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
```diff
@@ -465,7 +481,7 @@ final Object await(Object e, long nanos, Object blocker, boolean spin) {
      * to reduce retries due to stalls in updating head and tail.  Must
      * be at least 4.
      */
-    private static final int MAX_SLACK = 1 << 6;
+    private static final int MAX_SLACK = 1 << 7;
```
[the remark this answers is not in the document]
DougLea: I removed this after finding ways to reduce slack accumulation so it isn't needed

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
```diff
@@ -307,151 +368,114 @@ final DualNode cmpExNext(DualNode cmp, DualNode val) {
...
+        /**
+         * The number of nanoseconds for which it is faster to spin
+         * rather than to use timed park. A rough estimate suffices.
+         */
+        private static final long SPIN_FOR_TIMEOUT_THRESHOLD = 1L << 10;
+
+        /**
+         * True if system is a uniprocessor, occasionally rechecked.
+         */
+        private static boolean isUniprocessor =
```
viktorklang-ora, line 397: @DougLea  If this is intentionally non-volatile, I think it is worth documenting the rationale.

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
```diff
@@ -307,151 +368,114 @@ final DualNode cmpExNext(DualNode cmp, DualNode val) {
...
+                } else if (w.isInterrupted() ||
+                           (timed &&       // try to cancel with impossible match
+                            ((ns = deadline - System.nanoTime()) <= 0L))) {
+                    m = cmpExItem(e, (e == null) ? this : null);
+                    break;
+                } else if (timed) {
+                    if (ns < SPIN_FOR_TIMEOUT_THRESHOLD)
+                        Thread.onSpinWait();
+                    else
+                        LockSupport.parkNanos(ns);
```
viktorklang-ora, line 446: @DougLea If `ns` is sufficiently large, would it not make sense to use managed blocking in case the current thread is a FJWT as below? 🤔

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
```diff
@@ -528,73 +557,59 @@ final int addSweepVote() {
...
-            unsplice(pred, s);               // cancelled
-        return match;
+        if (s == null || ns <= 0L)
+            m = e;                          // don't wait
+        else if ((m = s.await(e, ns, this,  // spin if at or near head
+                              p == null || p.waiter == null)) == e)
+            unsplice(p, s);                 // cancelled
+        else if (m != null)
+            s.selfLinkItem();
+        return m;
```
viktorklang-ora, line 612: @DougLea I'd probably add a newline before the return statement to visually distinguish that it isn't intended to be read as a part of the if-else branches.

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
DougLea: Added: (L348)
     * than unnecessarily requiring volatile accesses elsewhere. This
     * fence also separates accesses to field isUniprocessor.

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
DougLea: It's a good point, but we don't normally do this. Added (L336):
     * returns just barely too soon. As is the case in most j.u.c
     * blocking support, untimed waits use ManagedBlockers when
     * callers are ForkJoin threads, but timed waits use plain
     * parkNanos, under the rationale that known-to-be transient
     * blocking doesn't require compensation. (This decision should be
     * revisited here and elsewhere to deal with very long timeouts.)

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
DougLea: Thanks. Done.

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
DougLea: I should have mentioned that the tradeoffs include that using ManagedBlocking for timed waits here adds overhead: -- constructing of another object just to track the time, and duplicate checks of other parts of await(. Which requires choosing a threshold where it would be worthwhile. Which is  a hard-to-win decision.

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
```diff
@@ -604,45 +619,54 @@ final Object xfer(Object e, long nanos) {
...
-        // assert p != null && !c.isLive() && c != p;
-        return ((pred != null ?
-                 pred.cmpExNext(c, p) :
-                 tryAdvanceHead(c, p))) == c;
+        // assert p != null && c.matched() && c != p;
+        if (pred != null)
+            return pred.cmpExNext(c, p) == c;
+        else if (cmpExHead(c, p) != c)
+            return false;
+        if (c != null)
```
viktorklang-ora, line 674: @DougLea If this `if` is not related to the above if-tree, I'd put a newline between them. (Also makes sense to add a newline before the last `return` as well (to indicate that it is standalone from the if)

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
```diff
@@ -697,10 +721,14 @@ else if ((q = p.next) == null)    // end of list
                 break;
             else if (p == q)                  // self-link; restart
                 p = h = head;
-            else if (p != h)                  // collapse
-                p = h = ((u = tryAdvanceHead(h, q)) == h) ? q : u;
-            else                              // traverse past header
+            else if (p == h)                  // traverse past header
                 p = q;
+            else if ((u = cmpExHead(h, q)) != h)
```
viktorklang-ora, line 733: @DougLea Always nice to see `compareAndExchange` being used :)

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
```diff
@@ -865,15 +893,16 @@ final class Itr implements Iterator<E> {
         private void advance(DualNode pred) {
             for (DualNode p = (pred == null) ? head : pred.next, c = p;
                  p != null; ) {
-                final Object item;
-                if ((item = p.item) != null && p.isData) {
+                boolean isData = p.isData;
```
viktorklang-ora, line 903: @DougLea Are you finding that manual hoisting of reads to final members has a perf edge? 🤔

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
```diff
@@ -1542,18 +1575,19 @@ else if (!p.isData && item == null)
...
+            boolean isData = p.isData;
+            Object item = p.item;
             DualNode q = p.next;
-            final Object item;
-            if ((item = p.item) != null) {
-                if (p.isData) {
+            if (item != null) {
+                if (isData) {
                     action.accept((E) item);
                     pred = p; p = q; continue;
```
viktorklang-ora, line 1591: @DougLea These three statements deserve their own lines, I'd say. :)

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
viktorklang-ora: @DougLea I suspect that a good "governing principle" here is: is the timed park expected to be prematurely unparked, and is the delay sufficiently large to warrant compensation—use ManagedBlocker. Ideally (perhaps) the call to ManagedBlocker takes an "expected conservative estimate of blocking"-nanos, and then that behavior could be completely enclosed within the ManagedBlocker-construct itself (and thus be uniformly applied). It could also use heuristics about expected lifetime of compensating threads to see if it is justified to add one... Just thoughts off of the top of my head. 🤔

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
```diff
@@ -604,45 +619,54 @@ final Object xfer(Object e, long nanos) {
...
+                matched = seen = true;
+            else
+                matched = p.matched();
+            if ((f = p.next) == p)
+                p = h = head;
+            else if (f != null && matched)
+                p = f;
+            else {
+                if (p != h && cmpExHead(h, p) == h)
+                    h.next = h; // h.selfLink();
```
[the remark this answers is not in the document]
DougLea: Not really; just a undeleted remnant of a previous factoring. Fixed.

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
DougLea: Thanks. Done.

src/java.base/share/classes/java/util/concurrent/LinkedTransferQueue.java
DougLea: Collection-support code from method countOfMode on down was minimally touched from its previous version. This shows up as a style clash into very branchy self-link-coping code. Changing might require more changes to WhiteBox tests that are finicky about node counts and self-links expected in traversals etc.  (These tests are useful but are themselves hard to change to accommodate small implementation differences.)

viktorklang-ora: @DougLea Alright, good to know :)

wborn:
Thanks for making the fixes Doug!
Would it also be possible to backport thes fixes to Java 17?

It seems to be a very common issue for openHAB users now that they upgrade to openHAB 4 which requires Java 17.
See: https://community.openhab.org/t/consistent-100-cpu-use-of-safecall-queue-thread/143792

DougLea:
@wborn I think 17 should also be OK modulo deleting 2 lines for pre-21 mentioned above. I only checked with 19 though..

suryag10:
@DougLea is there any timeline where we can expect the backport of this fix into jdk17? or any other work around?

DougLea:
@suryag10 Sorry I'm not the right person to ask about backports.

jaikiran:
Like Doug notes, the JDK backports are managed as a separate project. Details are available here https://openjdk.org/projects/jdk-updates/. For JDK 17 the corresponding project is https://wiki.openjdk.org/display/JDKUpdates/JDK+17u which has the necessary details about the process as well as the mailing list details where you can bring up the backporting question.

suryag10:
Thanks for the info @DougLea and @jaikiran

viktorklang-ora:
Just my 2 cents—given the size of this change, I'd be hesitant to have it backported. Also, more than this PR would need to be backported, for instance: https://github.com/openjdk/jdk/pull/19271

PR #14322, opened 2023-06-05
8309505

plummercj:
The test fails because it tries to determine the main debuggee thread by allowing it run until the debuggee class is loaded (it waits for the ClassPrepareEvent). Normally this would be done on the main debuggee thread. However, when using virtual threads, the main thread has yet to spawn the virtual thread to run the test on. The test is fixed by instead just waiting until the debuggee main method is entered.


alexmenkov (approved):

sspitsyn (approved):
Looks good.
Thanks,
Serguei

plummercj:
Thanks for the reviews Alex and Serguei!

/integrate

PR #14323, opened 2023-06-05
8309506

plummercj:
The test fails when the main debuggee thread is a virtual thread, because virtual threads are always daemon threads. Because of this all the test threads that the debuggee creates are also daemon threads. The main deuggee thread immediately exits after creating the test threads, and at that point there are no non-daemon threads left running to keep the debuggee process alive, so the test threads never get a chance to do all the work that is expected of them. 

The fix is to use Thread.join() to prevent the main test thread from exiting until all the created tests threads have exited. I also updated the test so when run using the virtual test thread factory, the debuggee test threads will now be virtual threads. Previously just the main debuggee thread would be a virtual thread. This is not a bug fix, but does provide better virtual thread testing.


alexmenkov (approved):

sspitsyn (approved):
Looks good.
Thanks,
Serguei

PR #14324, opened 2023-06-06
8309509

plummercj:
The test fails with the virtual test thread factory because it tries to find the "main" thread in the list of threads returned by JDI, but "main" is a virtual thread and will only be returned by JDI if the debug agent is launched with includevirtualthreads=y. As a result the thread is not found and the test asserts:

java.lang.RuntimeException: assertTrue: expected true, was false
	at jdk.test.lib.Asserts.fail(Asserts.java:594)
	at jdk.test.lib.Asserts.assertTrue(Asserts.java:486)
	at jdk.test.lib.Asserts.assertTrue(Asserts.java:472)
	at TestNestmateAttr.checkGoodTransforms(TestNestmateAttr.java:511)
	at TestNestmateAttr.methodEntered(TestNestmateAttr.java:320)
	at TestScaffold$EventHandler.notifyEvent(TestScaffold.java:205)
	at TestScaffold$EventHandler.run(TestScaffold.java:279)

The fix is to always run the debug agent with includevirtualthreads=y.

Tested by running all com/sun/jdi tests locally with and without the virtual test thread factory. Also ran tier1 and tier5 svc test tasks.

### DIFF
--- a/test/jdk/ProblemList-Virtual.txt
+++ b/test/jdk/ProblemList-Virtual.txt
@@ -32,7 +32,6 @@ com/sun/jdi/JdbMethodExitTest.java 8285422 generic-all
 com/sun/jdi/MethodEntryExitEvents.java 8285422 generic-all
 com/sun/jdi/MultiBreakpointsTest.java 8285422 generic-all
 com/sun/jdi/RedefineCrossStart.java 8278470 generic-all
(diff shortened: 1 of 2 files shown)


sspitsyn (approved):
Looks good.
Thanks,
Serguei

alexmenkov (approved):

PR #14330, opened 2023-06-06
JDK-8309462: [AIX] vmTestbase/nsk/jvmti/RunAgentThread/agentthr001/TestDescription.java crashing due to empty while loop

JoKern65:
The sys_thread_3() function contains an empty while loop, which by the standard can be optimized away. Please refer to discussion in https://github.com/llvm/llvm-project/issues/60622
The xlc17 compiler is doing so, and IBM claims that they are following the standard and will not fix this on compiler side.
So we have (at least) 3 ways to circumvent this behavior.

1. we can introduce the call of a nop library function, which will hinder the optimizer to throw away the loop (This is our proposed solution, but instead of a heavy looping thread, the result is a more or less idle thread):
`#include <unistd.h>`
`static void`
`sys_thread_3(jvmtiEnv* jvmti, JNIEnv* jni, void *p)`
`{`
`                while (1) {`
`                  sleep(1);`
`                }`
`}`

2. We can make use of a volatile variable in the loop body which also hinders the optimizer to throw away the loop:
`static void`
`sys_thread_3(jvmtiEnv* jvmti, JNIEnv* jni, void *p)`
`{`
`                volatile int i = 1;`
`                while (i) {`
`                  i += 2;`
`                }`
`}`

3. we can use the __attribute__ ((optnone)) modifier in the function declaration to suppress the optimization at all:
`static void`
`sys_thread_3(jvmtiEnv* jvmti, JNIEnv* jni, void *p) __attribute__ ((optnone))`
`{`
`                while (1) {`
`                }`
`}`

To make the third approach platform independent, we can implement it in the following way:
In globalDefinitions.hpp
`#ifndef OPTNONE`
`#define OPTNONE`
`#endif`

In globalDefinitions_xlc.hpp
`// optnone support`
`//`
`// To use if a function should not be optimized`
`// Usage:`
`// void* func(size_t size) OPTNONE {...}`
`#define OPTNONE __attribute__(( optnone))`

With this we can change libagentthr001.cpp in a platform independent way to
`static void`
`sys_thread_3(jvmtiEnv* jvmti, JNIEnv* jni, void *p) OPTNONE`
`{`
`     while (1) {`
`     }`
`}`

---------
### Progress

### Issue
 * [JDK-8309462](https://bugs.openjdk.org/browse/JDK-8309462): [AIX] vmTestbase/nsk/jvmti/RunAgentThread/agentthr001/TestDescription.java crashing due to empty while loop (**Bug** - P3)

### Reviewers
 * [Matthias Baesken](https://openjdk.org/census#mbaesken) (@MBaesken - **Reviewer**) ⚠️ Review applies to [ae0d67ae](https://git.openjdk.org/jdk/pull/14330/files/ae0d67aef19d2316da3cab1ccb964a73626058f1)
 * [Christoph Langer](https://openjdk.org/census#clanger) (@RealCLanger - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14330/head:pull/14330` \
`$ git checkout pull/14330`

`$ git checkout pull/14330` \
`$ git pull https://git.openjdk.org/jdk.git pull/14330/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14330`

`$ git pr show -t 14330`

</details>

<a href="https://git.openjdk.org/jdk/pull/14330.diff">https://git.openjdk.org/jdk/pull/14330.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14330#issuecomment-1578331940)


MBaesken (approved):
looks okay to me

TheRealMDoerr (approved):
LGTM. `sleep(1)` is my preferred solution. Thanks for fixing it!

test/hotspot/jtreg/vmTestbase/nsk/jvmti/RunAgentThread/agentthr001/agentthr001.cpp
```diff
@@ -23,6 +23,7 @@
 
 #include <stdio.h>
 #include <string.h>
+#include <unistd.h>
```
TheRealMDoerr, line 26: Breaks Windows build: agentthr001.cpp(26): fatal error C1083: Cannot open include file: 'unistd.h': No such file or directory

TheRealMDoerr (changes requested):

test/hotspot/jtreg/vmTestbase/nsk/jvmti/RunAgentThread/agentthr001/agentthr001.cpp
RealCLanger: On Windows I think you need windows.h. I guess you can do it like here: https://github.com/openjdk/jdk/blob/4a75fd462c002a209201d8bfc8d6c9eb286a7444/src/java.base/share/native/libjli/wildcard.c#L99

test/hotspot/jtreg/vmTestbase/nsk/jvmti/RunAgentThread/agentthr001/agentthr001.cpp
```diff
@@ -137,6 +138,7 @@ sys_thread_2(jvmtiEnv* jvmti, JNIEnv* jni, void *p) {
 static void JNICALL
 sys_thread_3(jvmtiEnv* jvmti, JNIEnv* jni, void *p) {
     while (1) {
+      sleep(1);
```
RealCLanger, line 141: Maybe you could add a comment here that the call to sleep is necessary to avoid the compiler optimization to elide the loop.

RealCLanger (changes requested):
The solution to use sleep looks fine. Please address the comments and also update the copyright year.

TheRealMDoerr (approved):
I don't like empty endless loops in general, but I think your solution is ok for a test.

RealCLanger (approved):

PR #14331, opened 2023-06-06
8308504: C2: "malformed control flow" after JDK-8303466

eme64:
This is another case where imprecise type computation leads to corrupted control flow.

The loop incr `AddI` does never overflow at runtime, this is guaranteed by the loop limit check (we ensure that adding the incr value to the limit would not overflow). However, the type so far did often overflow, and returns type `int`. The loop  tripcount `Phi` does not have such type overflow. So we had cases where the Phi had type `minint...7`, but the AddI had type `int`, since it adds `-3`, which makes the lower bound underflow (even though the runtime value never would). If we had known that an underflow is impossible because of the limit check, then we could have had type `minint...4` for the incr AddI.

Since JDK-8303466, the type of the limit can now not overflow and is more precise. In the example, the limit is `8...maxint`. The CastII after the zero-trip-guard thus concludes that the type must be `9...maxint` (we have a decrement loop).

Since the main-loop knows that the trip-count Phi has type `minint...7`, we get an empty range (intersection of `9...maxint` and `minint...7`), and the main-loop (in this case one of the assertion predicates) start to corrode away because of TOP inputs. This creates malformed control flow, we have an if with only one projection.

**Why did this arise with JDK-8303466?** We made the loop-limit type more precise, forbidding overflow. So before that change I think the loop limit just had a type would have overfown, and returned `int`. That would have not allowed the main-loop internals to detect that it could never be reached. The main-loop would not have died at all, but it would still never have been entered.

**Solution** Since the AddI (incr) of the tripcount Phi cannot overflow/underflow, **we should also prevent the type of the incr from overflowing/underflowing**. That means that it basically has the same type as the Phi, if not even a bit better. And this means that the type of the incr used to compare against the loop-limit is consistent with the type information in the main-loop.

I also had to improve the notification in IGVN, I encountered a case where a incr node was replaced via Identity, so the new incr node needed its type to be tightened (no overflow).

**Risk** Just as JDK-8303466, this fix here makes types more precise. That can always mean that somewhere else we also need more precise types than we currently have. It is difficult to know what such bugs are still lurking for us.

**Testing** Attached one regression test (takes less than `1.5sec`). Tested up to tier6 and stress testing. **Running**

--------

**Details**

example

```
621 CountedLoop (pre-loop)
622 Phi (tripcount, minint...7)
612 AddI (incr, int)    ---> with patch type minint...4

650 CmpI / 702 Bool (cannot detect that tripcount < limit)  ---> with patch it can detect it!

871 ConvL2I (loop limit, 8...maxint)

703 If (zero-trip-guard)
704 IfFalse (projection towards main-loop)

654 CastII (value of 612 AddI, if zero-trip-guard goes to main-loop, so value > limit, hence type 9...maxint)
809 CastII (remembers that tripcount cannot overflow, so has type minint...7)
(when the Opaque nodes vanish, then the two types create an empty range, and TOP starts corrupting the CFG)
```

---------
### Progress

### Issue
 * [JDK-8308504](https://bugs.openjdk.org/browse/JDK-8308504): C2: "malformed control flow" after JDK-8303466 (**Bug** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14331/head:pull/14331` \
`$ git checkout pull/14331`

`$ git checkout pull/14331` \
`$ git pull https://git.openjdk.org/jdk.git pull/14331/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14331`

`$ git pr show -t 14331`

</details>

<a href="https://git.openjdk.org/jdk/pull/14331.diff">https://git.openjdk.org/jdk/pull/14331.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14331#issuecomment-1598158859)


test/hotspot/jtreg/compiler/loopopts/TestLoopIncrNoOverflow.java
```diff
@@ -0,0 +1,61 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/*
+ * @test
+ * @bug 8308504
+ * @summary With JDK-8303466 the unroll-limit became more precise, and does not overflow.
+ *          We must ensure that the type of the loop incr also does not overflow.
+ * @run main/othervm -Xcomp -XX:-TieredCompilation
```
vnkozlov, line 29: Should we require C2 enabled for this test? `StressIGVN` is C2 flag.

src/hotspot/share/opto/addnode.cpp
```diff
@@ -419,30 +419,64 @@ Node* AddINode::Identity(PhaseGVN* phase) {
...
+  // go outside the int range.
+  jlong lo = java_add(r0->lo_as_long(), r1->lo_as_long());
+  jlong hi = java_add(r0->hi_as_long(), r1->hi_as_long());
+  int widen = MAX2(r0->_widen,r1->_widen);
+
+  if (overflow_impossible) {
+    // Overflow / Underflow impossible
+    if (hi < (jlong)min_jint || lo > (jlong)max_jint) {
+      // [lo, hi] is outside of int range -> never valid
+      assert(false, "is there any such case?"); // TODO remove
```
vnkozlov, line 439: Please, dump nodes/types to get more information when we hit this assert.

src/hotspot/share/opto/addnode.cpp
```diff
@@ -419,30 +419,64 @@ Node* AddINode::Identity(PhaseGVN* phase) {
...
+      jint add_con = java_add(r0->get_con(), r1->get_con());
+      return TypeInt::make(add_con);
+    } else {
+      if (lo < (jlong)min_jint || hi > (jlong)max_jint) {
+        // Overflow / Underflow -> return int
+        return TypeInt::INT;
+      } else {
+        assert((jlong)min_jint <= lo &&
+               lo <= hi &&
+               hi <= (jlong)max_jint, "no overflow");
```
vnkozlov, line 462: May be print `lo`, `hi` values in assert.

rwestrel:
If in `PhaseIdealLoop::is_counted_loop`, when the loop limit check is added, we added a `CastII` to cast the limit to a narrower type wouldn't that have the same effect? `PhiNode::value()` would compute a narrower type for the iv Phi. Would the loop incr still overflow? Assuming that does work wouldn't that be a simpler change?

rwestrel:
It doesn't seem to be true that the loop incr never overflows in the general case. See this example:
```
public class TestOverflowCountedLoopIncr {
        for (int i = 0; i < 20_000; i++) {
            test(Integer.MAX_VALUE);
        }
    }

    private static float test(int start) {
        float v = 1;
        int i = start;
        do {
            synchronized (new Object()) {}
            v *= 2;
            i++;
        } while (i < Integer.MIN_VALUE + 100);
        return v;
    }
}
```
That one also hangs with -XX:LoopMaxUnroll=0. A bug in LSM it seems. I will file a bug for that.

rwestrel:
> @rwestrel thanks for the idea. So you mean I should make the `Phi` type narrow enough, such that adding the `stride` to it should never lead to a type overflow, right? Then there would be no type overflow in the `incr`.

That's what I was thinking indeed.

> You want to `CastII` the limit for the exit check, I assume? And we would have to add a `CastII` again if the limit is ever changed during the unrolling, right?

Would that be required? I was hoping that the CastII on the limit before unrolling would cause the phi type to be narrow 
enough.

rwestrel:
> > It doesn't seem to be true that the loop incr never overflows in the general case. See this example:
> 
> But don't we check that the limit is small enough at runtime, such that there cannot be an overflow? We do that with `check_stride_overflow` and `insert_loop_limit_check_predicate`. And if it does overflow, we do not go into the counted-loop, but we uncommon trap, I think. Or are you sure that we actually enter the counted-loop with your example? Or just some peeled version?

Ran with:
```
 -XX:-BackgroundCompilation -XX:-UseOnStackReplacement -XX:-TieredCompilation -XX:CompileOnly=TestOverflowCountedLoopIncr::test -XX:CompileCommand=quiet -XX:LoopMaxUnroll=0 -XX:+UseCountedLoopSafepoints 
```
I see a single counted loop, no uncommon trap in the IR.

eme64:
@rwestrel so you think the `incr` can indeed overflow, and that is ok? Or would that be a bug? Why do we even have the loop limit check in the first place, if overflow is allowed?

rwestrel:
> @rwestrel so you think the `incr` can indeed overflow, and that is ok? Or would that be a bug? Why do we even have the loop limit check in the first place, if overflow is allowed?

To guarantee no overflow requires init < limit (for a loop going up). Nothing guarantees that when c2 pattern matches a counted loop. Whether overflow is a problem or not would require taking a closer look at individual optimizations.

PR #14338, opened 2023-06-06
8297923: java.awt.ScrollPane broken after multiple scroll up/down

aivanov-jdk:
**Problem description**

If you grab the thumb of the scroll bar of `ScrollPane` and drag it slowly and continuously up and down, you'll notice the UI stops rendering correctly: the child component of the scroll pane will render on the left of the frame itself, the inside of the scroll pane will be filled with the background color of its child component.

**Root cause**

AWT calls the [`::SetScrollInfo`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setscrollinfo) function on EDT when AWT processes scroll bar tracking events. This Windows API is not thread-safe, calling this function on an incorrect thread leads to leaking GDI objects.

When the process reaches the limit on the number of GDI objects, it cannot create new GDI objects, which results in rendering issues.

**Fix**

To resolve the problem, I modified the code so that `::SetScrollInfo` and [`::GetScrollInfo`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getscrollinfo) are always called on the toolkit thread, all AWT components are created on the toolkit thread.

An automatic test is provided. The test scrolls the vertical scroll bar up and down. Then the test takes a screenshot. When the bug is reproduced, the robot cannot create new GDI objects to capture the screenshot and it throws `OutOfMemoryError`.

---------
### Progress

### Issue
 * [JDK-8297923](https://bugs.openjdk.org/browse/JDK-8297923): java.awt.ScrollPane broken after multiple scroll up/down (**Bug** - P3)

### Reviewers
 * [Sergey Bylokhov](https://openjdk.org/census#serb) (@mrserb - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14338/head:pull/14338` \
`$ git checkout pull/14338`

`$ git checkout pull/14338` \
`$ git pull https://git.openjdk.org/jdk.git pull/14338/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14338`

`$ git pr show -t 14338`

</details>

<a href="https://git.openjdk.org/jdk/pull/14338.diff">https://git.openjdk.org/jdk/pull/14338.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14338#issuecomment-1579070135)

### DIFF
--- a/src/java.desktop/windows/native/libawt/windows/awt_ScrollPane.cpp
+++ b/src/java.desktop/windows/native/libawt/windows/awt_ScrollPane.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1996, 2014, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1996, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -701,7 +701,7 @@ Java_sun_awt_windows_WScrollPanePeer_getOffset(JNIEnv *env, jobject self,
     gos->scrollpane = env->NewGlobalRef(self);
     gos->orient = orient;
 
-    return static_cast<jint>(reinterpret_cast<INT_PTR>(AwtToolkit::GetInstance().SyncCall(
+    return static_cast<jint>(reinterpret_cast<INT_PTR>(AwtToolkit::GetInstance().InvokeFunction(
         (void *(*)(void *))AwtScrollPane::_GetOffset, gos)));
     // global ref and gos are deleted in _GetOffset()
 
@@ -742,7 +742,7 @@ Java_sun_awt_windows_WScrollPanePeer_setScrollPosition(JNIEnv *env,
     ssps->x = x;
     ssps->y = y;
 
-    AwtToolkit::GetInstance().SyncCall(AwtScrollPane::_SetScrollPos, ssps);
+    AwtToolkit::GetInstance().InvokeFunctionLater(AwtScrollPane::_SetScrollPos, ssps);
     // global ref and ssps are deleted in _SetScrollPos()
 
     CATCH_BAD_ALLOC;
@@ -803,7 +803,7 @@ Java_sun_awt_windows_WScrollPanePeer_setSpans(JNIEnv *env, jobject self,
     sss->childWidth = childWidth;
     sss->childHeight = childHeight;
 
-    AwtToolkit::GetInstance().SyncCall(AwtScrollPane::_SetSpans, sss);
+    AwtToolkit::GetInstance().InvokeFunctionLater(AwtScrollPane::_SetSpans, sss);
     // global ref and sss are deleted in _SetSpans
 
     CATCH_BAD_ALLOC;
--- a/test/jdk/java/awt/ScrollPane/ScrollPaneLeakTest.java
+++ b/test/jdk/java/awt/ScrollPane/ScrollPaneLeakTest.java
@@ -1,0 +1,189 @@
+/*
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+import java.awt.AWTException;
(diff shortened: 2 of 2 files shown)


aivanov-jdk:
The screenshot of the test case after the problem is reproduced:

ScrollPaneTest

If you scroll up or down, the left side of the frame gets filled with the light-green color of the canvas which is the child component inside the scroll pane.

src/java.desktop/windows/native/libawt/windows/awt_ScrollPane.cpp
```diff
@@ -701,7 +701,7 @@ Java_sun_awt_windows_WScrollPanePeer_getOffset(JNIEnv *env, jobject self,
     gos->scrollpane = env->NewGlobalRef(self);
     gos->orient = orient;
 
-    return static_cast<jint>(reinterpret_cast<INT_PTR>(AwtToolkit::GetInstance().SyncCall(
+    return static_cast<jint>(reinterpret_cast<INT_PTR>(AwtToolkit::GetInstance().InvokeFunction(
```
prrace, line 704: So this needs to block until the toolkit thread can process this and return, but since its
directly called from Java (ie we are in a JNI method) I think this is likely fine.

src/java.desktop/windows/native/libawt/windows/awt_ScrollPane.cpp
```diff
@@ -742,7 +742,7 @@ Java_sun_awt_windows_WScrollPanePeer_setScrollPosition(JNIEnv *env,
     ssps->x = x;
     ssps->y = y;
 
-    AwtToolkit::GetInstance().SyncCall(AwtScrollPane::_SetScrollPos, ssps);
+    AwtToolkit::GetInstance().InvokeFunctionLater(AwtScrollPane::_SetScrollPos, ssps);
```
prrace, line 745: Whereas in the _GetOffset case above, you clearly need to wait until the result is returned - I guess
you didn't see a need to block here ? Or the case below ?

src/java.desktop/windows/native/libawt/windows/awt_ScrollPane.cpp
aivanov-jdk: That's right.

In addition to that, `WScrollPanePeer.getScrollOffset` is unused, so it's never called.

src/java.desktop/windows/native/libawt/windows/awt_ScrollPane.cpp
honkar-jdk: I had a same question as to why it was `InvokeFunction` for `_getOffset` and `InvokeFunctionLater` for `_SetScrollPos` and `_SetSpans`. The above discussion makes it clear.

honkar-jdk (commented):
> The screenshot of the test case after the problem is reproduced:
> 
> ScrollPaneTest
> 
> If you scroll up or down, the left side of the frame gets filled with the light-green color of the canvas which is the child component inside the scroll pane.

Test works as expected with the fix. Without the fix OutOfMemory exception is thrown as described.
When running the test without the fix, I had to introduce a small delay after the frame is maximized again to see the light-green block outside the canvas.

src/java.desktop/windows/native/libawt/windows/awt_ScrollPane.cpp
aivanov-jdk: No, there's no need to wait, and no result is returned.

Java has updated the position of the child component based on the new position of the scroll bar. Now the position of the non-client scroll bars of the scroll pane window need to be updated.

The `_SetSpans` recalculates the position and range of the scroll bars if the size of the scroll pane or its child component changes. The changes are already handled on the Java side, these changes need to applied to the native windows.

I have found no scenario where we need to block EDT.

aivanov-jdk:
> When running the test without the fix, I had to introduce a small delay after the frame is maximized again to see the light-green block outside the canvas.

You can use the test case attached to [JDK-8297923](https://bugs.openjdk.org/browse/JDK-8297923), `ScrollPaneTest.java`, to play around. Alternatively, you can comment out disposing of the frame in the automatic test.

prrace (approved):

honkar-jdk (approved):

src/java.desktop/windows/native/libawt/windows/awt_ScrollPane.cpp
mrserb: If it is unused can we delete it?

mrserb:
Probably we should update the Java_sun_awt_windows_WScrollbarPeer_setValues as well? it calls the SetScrollInfo as well

aivanov-jdk:
> Probably we should update the Java_sun_awt_windows_WScrollbarPeer_setValues as well? it calls the SetScrollInfo

You're right. It needs to be updated too. However, I'd rather do it separately.

src/java.desktop/windows/native/libawt/windows/awt_ScrollPane.cpp
aivanov-jdk: > If it is unused can we delete it?

It's possible. Yet I'd rather leave it for another time.

There's also a specific `WM_AWT_SET_SCROLL_INFO` message to call `::SetScrollInfo`. This message is never used, it can also be removed.

mrserb:
> > Probably we should update the Java_sun_awt_windows_WScrollbarPeer_setValues as well? it calls the SetScrollInfo
> 
> You're right. It needs to be updated too. However, I'd rather do it separately.

That is up2you, but it looks strange when the patch updates the code which is never executed and skips the code which is executed and has the bug

mrserb (approved):

aivanov-jdk:
> > > Probably we should update the Java_sun_awt_windows_WScrollbarPeer_setValues as well? it calls the SetScrollInfo
> > 
> > 
> > You're right. It needs to be updated too. However, I'd rather do it separately.
> 
> That is up2you, but it looks strange when the patch updates the code which is never executed and skips the code which is executed and has the bug

It has taken a very long time till the root cause for this bug was identified. I admit I was too focused on the code for `ScrollPane` only and I overlooked the fact that `::SetScrollInfo` is also used in `ScrollBar`.

As such, the updated code is limited to the native implementation of `ScrollPane`. Let's leave it this way.

Thank you for finding a similar problem in `ScrollBar`, it needs fixing too.

PR #14341, opened 2023-06-06
JDK-8309550: jdk.jfr.internal.Utils::formatDataAmount method should gracefully handle amounts equal to Long.MIN_VALUE

fthevenet:
Please review this simple fix to JDK-8309550.

NB: The problem originally reported only concerned `src/jdk.jfr/share/classes/jdk/jfr/internal/Utils.java`, but since this code was duplicated in `src/jdk.jfr/share/classes/jdk/jfr/internal/util/ValueFormatter.java` with, as far as I understood from the bug description, with an intent to refactor it a a later time, I opted to change both occurrences of the method.

---------
### Progress

### Issue
 * [JDK-8309550](https://bugs.openjdk.org/browse/JDK-8309550): jdk.jfr.internal.Utils::formatDataAmount method should gracefully handle amounts equal to Long.MIN_VALUE (**Bug** - `"4"`)

### Reviewers
 * [Thomas Stuefe](https://openjdk.org/census#stuefe) (@tstuefe - **Reviewer**) ⚠️ Review applies to [dca5b72c](https://git.openjdk.org/jdk/pull/14341/files/dca5b72c929672a7a99a7f3a678c268048645986)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14341/head:pull/14341` \
`$ git checkout pull/14341`

`$ git checkout pull/14341` \
`$ git pull https://git.openjdk.org/jdk.git pull/14341/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14341`

`$ git pr show -t 14341`

</details>

<a href="https://git.openjdk.org/jdk/pull/14341.diff">https://git.openjdk.org/jdk/pull/14341.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14341#issuecomment-1579319456)

### DIFF
--- a/src/jdk.jfr/share/classes/jdk/jfr/internal/Utils.java
+++ b/src/jdk.jfr/share/classes/jdk/jfr/internal/Utils.java
@@ -123,7 +123,7 @@ public final class Utils {
 
     // handle Long.MIN_VALUE as a special case since its absolute value is negative
     private static String formatDataAmount(String formatter, long amount) {
-        int exp = amount == Long.MIN_VALUE ? 6 : (int) (Math.log(Math.abs(amount)) / Math.log(1024));
+        int exp = (amount == Long.MIN_VALUE) ? 6 : (int) (Math.log(Math.abs(amount)) / Math.log(1024));
         char unitPrefix = "kMGTPE".charAt(exp - 1);
         return String.format(formatter, amount / Math.pow(1024, exp), unitPrefix);
     }
--- a/src/jdk.jfr/share/classes/jdk/jfr/internal/util/ValueFormatter.java
+++ b/src/jdk.jfr/share/classes/jdk/jfr/internal/util/ValueFormatter.java
@@ -148,7 +148,7 @@ public final class ValueFormatter {
 
     // handle Long.MIN_VALUE as a special case since its absolute value is negative
     private static String formatDataAmount(String formatter, long amount) {
-        int exp = amount == Long.MIN_VALUE ? 6 : (int) (Math.log(Math.abs(amount)) / Math.log(1024));
+        int exp = (amount == Long.MIN_VALUE) ? 6 : (int) (Math.log(Math.abs(amount)) / Math.log(1024));
         char unitPrefix = "kMGTPE".charAt(exp - 1);
         return String.format(formatter, amount / Math.pow(1024, exp), unitPrefix);
     }


tstuefe (approved):
Small nit, otherwise fine.

tstuefe:
Also, please enable Github actions in your repository (somewhere in settings). That way we see GHA tests here.

fthevenet:
Thanks Thomas.

tstuefe:
And I'm not sure if JFR changes need one or two reviewers. @mgronlun?

mgronlun (approved):

mgronlun:
Normally we are ok wiyh only one Reviewer. But I approved as well. Cheers.

egahlin:
JFR uses Long.MIN_VALUE is used to indicate that a value is missing. I think a correct fix should be to check for that before calling the method, or doing it in the method.

fthevenet:
Thanks for the clarification, @egahlin 

IMHO, putting the responsibility of checking for Long.MIN_VALUE on the caller is taking the risk to run in the same issue in the future, so having `formatDataAmount` return the string `"N/A"` in such a case seems the better solution. 

If we deem it preferable to let callers handle this verification, then I believe we should still consider keeping this patch in - which is at least *technically correct* and avoid a seemingly unrelated exception being thrown - or at least throw a more helpful IllegalArgumentException to avoid further confusion.

What do you think?

egahlin:
> Thanks for the clarification, @egahlin
> 
> IMHO, putting the responsibility of checking for Long.MIN_VALUE on the caller is taking the risk to run in the same issue in the future, so having `formatDataAmount` return the string `"N/A"` in such a case seems the better solution.
> 
> If we deem it preferable to let callers handle this verification, then I believe we should still consider keeping this patch in - which is at least _technically correct_ and avoid a seemingly unrelated exception being thrown - or at least throw a more helpful IllegalArgumentException to avoid further confusion.
> 
> What do you think?

The change is an improvement to java.lang.StringIndexOutOfBoundsException. Having the N/A check in the formatDataAmount is probably best, but I haven't looked at all call sites. Maybe some Long.MIN_VALUE checked can be removed.

fthevenet:
I actually took a closer look at one those call site, namely the one that shows up in the stack you can reproduce from the elements in the JBS bug:
``` java
java.lang.StringIndexOutOfBoundsException: Index -1 out of bounds for length 6
        at java.base/jdk.internal.util.Preconditions$1.apply(Preconditions.java:55)
        at java.base/jdk.internal.util.Preconditions$1.apply(Preconditions.java:52)
        at java.base/jdk.internal.util.Preconditions$4.apply(Preconditions.java:213)
        at java.base/jdk.internal.util.Preconditions$4.apply(Preconditions.java:210)
        at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:98)
        at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
        at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
        at java.base/java.lang.String.checkIndex(String.java:4828)
        at java.base/java.lang.StringLatin1.charAt(StringLatin1.java:46)
        at java.base/java.lang.String.charAt(String.java:1555)
        at jdk.jfr/jdk.jfr.internal.Utils.formatDataAmount(Utils.java:127)
        at jdk.jfr/jdk.jfr.internal.Utils.formatBytes(Utils.java:155)
        at jdk.jfr/jdk.jfr.internal.tool.PrettyWriter.printFormatted(PrettyWriter.java:595)
        at jdk.jfr/jdk.jfr.internal.tool.PrettyWriter.printValue(PrettyWriter.java:357)
        at jdk.jfr/jdk.jfr.internal.tool.PrettyWriter.printFieldValue(PrettyWriter.java:276)
        at jdk.jfr/jdk.jfr.internal.tool.PrettyWriter.print(PrettyWriter.java:214)
        at jdk.jfr/jdk.jfr.internal.tool.PrettyWriter.print(PrettyWriter.java:75)
        at jdk.jfr/jdk.jfr.internal.tool.EventPrintWriter.print(EventPrintWriter.java:82)
        at jdk.jfr/jdk.jfr.internal.tool.Print.execute(Print.java:165)
        at jdk.jfr/jdk.jfr.internal.tool.Main.main(Main.java:92) 
```
That initially got me very confused as I couldn't understand why `PrettyWriter.printFormatted` from `PrettyWritter.printValue` would get called from `PrettyWritter.printValue` as it should have been cut short by the [check line 343](https://github.com/openjdk/jdk/blob/master/src/jdk.jfr/share/classes/jdk/jfr/internal/tool/PrettyWriter.java#L343-L348).

It turns out that this is because in that specific case at least, the value passed to printValue is actually a `BigDecimal` instance, which jumps over the check. 
Latter on in `printFormatted`, that same argument is cast as a `Number` instance, and a `long` value derived from it, triggering the issue: https://github.com/openjdk/jdk/blob/b5b5b7ce7220df650f6142c40d6e89c0462877ce/src/jdk.jfr/share/classes/jdk/jfr/internal/tool/PrettyWriter.java#L582

All this to say that a follow-up is indeed needed to more it more robust, but that for now the check directly in the `formatDataAmount ` method is useful as an immediate from of protection. 
However, because this may involve changes in several places, it might be wise to only undertake that after the large refactor initiated by [JDK-8306703](https://bugs.openjdk.org/browse/JDK-8306703) is finished? Wdyt?

PR #14342, opened 2023-06-06
8308995: Update Network IO JFR events to be static mirror events

tprinzing:
The socket read/write JFR events currently use instrumentation of java.base code using templates in the jdk.jfr modules. This results in some java.base code residing in the jdk.jfr module which is undesirable.

JDK19 added static support for event classes. The old instrumentor classes should be replaced with mirror events using the static support.

In the java.base module:
Added two new events, jdk.internal.event.SocketReadEvent and jdk.internal.event.SocketWriteEvent.
java.net.Socket and sun.nio.ch.SocketChannelImpl were changed to make use of the new events.

In the jdk.jfr module:
jdk.jfr.events.SocketReadEvent and jdk.jfr.events.SocketWriteEvent were changed to be mirror events.
In the package jdk.jfr.internal.instrument, the classes SocketChannelImplInstrumentor, SocketInputStreamInstrumentor, and SocketOutputStreamInstrumentor were removed. The JDKEvents class was updated to reflect all of those changes.

The existing tests in test/jdk/jdk/jfr/event/io continue to pass with the new implementation:
Passed: jdk/jfr/event/io/TestSocketChannelEvents.java
Passed: jdk/jfr/event/io/TestSocketEvents.java

I added a micro benchmark which measures the overhead of handling the jfr socket events.
test/micro/org/openjdk/bench/java/net/SocketEventOverhead.java.
It needs access the jdk.internal.event package, which is done at runtime with annotations that add the extra arguments.
At compile time the build arguments had to be augmented in make/test/BuildMicrobenchmark.gmk

---------
### Progress

### Issue
 * [JDK-8308995](https://bugs.openjdk.org/browse/JDK-8308995): Update Network IO JFR events to be static mirror events (**Enhancement** - P4)

### Reviewers
 * [Erik Gahlin](https://openjdk.org/census#egahlin) (@egahlin - **Reviewer**) ⚠️ Review applies to [d6f7df72](https://git.openjdk.org/jdk/pull/14342/files/d6f7df72bbdcdf6666eb5e541be663f489a94ebb)
 * [Alan Bateman](https://openjdk.org/census#alanb) (@AlanBateman - **Reviewer**) ⚠️ Review applies to [d6f7df72](https://git.openjdk.org/jdk/pull/14342/files/d6f7df72bbdcdf6666eb5e541be663f489a94ebb)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14342/head:pull/14342` \
`$ git checkout pull/14342`

`$ git checkout pull/14342` \
`$ git pull https://git.openjdk.org/jdk.git pull/14342/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14342`

`$ git pr show -t 14342`

</details>

<a href="https://git.openjdk.org/jdk/pull/14342.diff">https://git.openjdk.org/jdk/pull/14342.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14342#issuecomment-1601786863)


src/java.base/share/classes/java/net/Socket.java
```diff
@@ -1095,6 +1098,20 @@ public int read() throws IOException {
...
+            long start = SocketReadEvent.timestamp();
+            try {
+                nbytes = read0(b, off, len);
+            } finally {
+                SocketReadEvent.checkForCommit(start, nbytes, parent.getRemoteSocketAddress(), parent.getSoTimeout());
+            }
+            return nbytes;
+        }
+
+        private int read0(byte[] b, int off, int len) throws IOException {
```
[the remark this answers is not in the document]
AlanBateman: > Are we using a convention of `implRead` or `readImpl`? Either is ok with me, but I think we had been using `readImpl` and similar elsewhere.

This code is already using implXXX so it's just be consistent.

egahlin:
I believe the fields SOCKET_READ and SOCKET_WRITE in the jdk.jfr.events.EventConfigurations class can be removed.

src/java.base/share/classes/java/net/Socket.java
```diff
@@ -1095,6 +1098,20 @@ public int read() throws IOException {
         }
         @Override
         public int read(byte[] b, int off, int len) throws IOException {
+            if (! SocketReadEvent.enabled()) {
```
AlanBateman, line 1101: Drop the space in "! SocketReadEvent" as it is inconsistent with the existing code.

src/java.base/share/classes/java/net/Socket.java
```diff
@@ -1095,6 +1098,20 @@ public int read() throws IOException {
...
+                return read0(b, off, len);
+            }
+            int nbytes = 0;
+            long start = SocketReadEvent.timestamp();
+            try {
+                nbytes = read0(b, off, len);
+            } finally {
+                SocketReadEvent.checkForCommit(start, nbytes, parent.getRemoteSocketAddress(), parent.getSoTimeout());
```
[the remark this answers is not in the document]
AlanBateman: The other issue to think about here is where the Socket is asynchronously closed. In t hat case, implRead will throw but we'll end up with a confusing suppressed exception due to the call to getSoTimeout. I think this will have to be replaced with a call to a helper method that returns the timeout or 0.

src/java.base/share/classes/jdk/internal/event/SocketReadEvent.java
```diff
@@ -0,0 +1,131 @@
...
+     * commit method.
+     *
+     * @param start  the start time
+     * @param nbytes  how many bytes were transferred
+     * @param remote  the address of the remote socket
+     * @param timeout  maximum time to wait
+     */
+    public static void checkForCommit(long start, long nbytes, SocketAddress remote, long timeout) {
+        long duration = timestamp() - start;
+        if (shouldCommit(duration)) {
```
AlanBateman, line 119: I think you know this, but this will need to be re-examined for SocketChannel configured non-blocking.

src/java.base/share/classes/sun/nio/ch/SocketChannelImpl.java
```diff
@@ -403,6 +405,20 @@ private void throwConnectionReset() throws SocketException {
...
     public int read(ByteBuffer buf) throws IOException {
+        if (!SocketReadEvent.enabled()) {
+            return implRead(buf);
+        }
+        int nbytes = 0;
+        long start = SocketReadEvent.timestamp();
+        try {
+            nbytes = implRead(buf);
+        } finally {
+            SocketReadEvent.checkForCommit(start, nbytes, getRemoteAddress(), 0);
```
AlanBateman, line 416: This will need to be changed to use remoteAddress(), can't use getRemoteAddress() as it might throw.

src/java.base/share/classes/sun/nio/ch/SocketChannelImpl.java
```diff
@@ -446,6 +462,22 @@ public int read(ByteBuffer buf) throws IOException {
     @Override
     public long read(ByteBuffer[] dsts, int offset, int length)
         throws IOException
+    {
+        if (! SocketReadEvent.enabled()) {
```
AlanBateman, line 466: Spurious space here too, several other places.

src/java.base/share/classes/sun/nio/ch/SocketChannelImpl.java
```diff
@@ -446,6 +462,22 @@ public int read(ByteBuffer buf) throws IOException {
...
+            return implRead(dsts, offset, length);
+        }
+        long nbytes = 0;
+        long start = SocketReadEvent.timestamp();
+        try {
+            nbytes = implRead(dsts, offset, length);
+        } finally {
+            SocketReadEvent.checkForCommit(start, nbytes, getRemoteAddress(), 0);
```
AlanBateman, line 474: This has to be remoteAddress() too.

src/java.base/share/classes/sun/nio/ch/SocketChannelImpl.java
```diff
@@ -403,6 +405,20 @@ private void throwConnectionReset() throws SocketException {
 
     @Override
     public int read(ByteBuffer buf) throws IOException {
+        if (!SocketReadEvent.enabled()) {
```
AlanBateman, line 408: The read/write with sun.nio.ch.SocketInputStream and SocketOutputStream does not go through SC.read/write so I think SocketAdaptor read/write will need attention, maybe a future PR as there are other code paths that aren't covered in this PR.

dfuch:
The new code seems to accurately correspond to what the various `*Instrumentor` classes were doing, so that is good. I agree with Alan that potential exception that may arise when generating the event are an issue (e.g. call to getRemoteAddress() or other getter that may make a check on the socket state) - though that was already present in the original code. Maybe that could be fixed here though. 
In cases where the implRead/implWrite call throws an exception, shouldn't the event contain that exception, or at least exception message? If it doesn't should it be emitted at all, or should another event be emitted instead?

egahlin:
> The new code seems to accurately correspond to what the various `*Instrumentor` classes were doing, so that is good. I agree with Alan that potential exception that may arise when generating the event are an issue (e.g. call to getRemoteAddress() or other getter that may make a check on the socket state) - though that was already present in the original code. Maybe that could be fixed here though.
> In cases where the implRead/implWrite call throws an exception, shouldn't the event contain that exception, or at least exception message? If it doesn't should it be emitted at all, or should another event be emitted instead?

An exception event will be emitted. The event is disabled by default, but there is ongoing work on a throttling mechanism, so it can be always-on.

src/java.base/share/classes/sun/nio/ch/SocketChannelImpl.java
tprinzing: I've created https://bugs.openjdk.org/browse/JDK-8310978 to drive the future PR to support the missing code paths

src/java.base/share/classes/java/net/Socket.java
```diff
@@ -1108,6 +1125,16 @@ public int read(byte[] b, int off, int len) throws IOException {
                 throw e;
             }
         }
+
+        private int getSoTimeout() {
+            try {
+                return parent.getSoTimeout();
+            } catch (Throwable t) {
+                // ignored - avoiding exceptions in jfr event data gathering
```
AlanBateman, line 1133: This should be SocketException, not Throwable. That said, I think it would be useful to know why the SocketReadEvent includes the timeout. Is this used to see If read durations are close to the timeout? I assume once this code is fixed to deal with the exceptional case that the need to include the timeout for the success case will mostly go away.

tprinzing:
I believe this has all requested changes or has separate bug reports to address changes yet needing to be made.
https://bugs.openjdk.org/browse/JDK-8310979  - better exception handling
https://bugs.openjdk.org/browse/JDK-8310978 - missing code paths for event generation
https://bugs.openjdk.org/browse/JDK-8310994 - non-blocking, event for select ops

This request has been patiently waiting for approval for a long time!

AlanBateman:
> I believe this has all requested changes or has separate bug reports to address changes yet needing to be made. https://bugs.openjdk.org/browse/JDK-8310979 - better exception handling https://bugs.openjdk.org/browse/JDK-8310978 - missing code paths for event generation https://bugs.openjdk.org/browse/JDK-8310994 - non-blocking, event for select ops
> 
> This request has been patiently waiting for approval for a long time!

I plan to come back to this soon, been busy with other things.  It's good to have JBS issues created for the next steps. I assume you can start to prepare the changes for the next steps so that all the pieces go into the same release.

egahlin (approved):
Looks good as far as I can tell. I would perhaps name the checkForCommit(...) method commitEvent(...) to make it more clear that it writes the event, but it's not a big thing.

src/java.base/share/classes/java/net/Socket.java
```diff
@@ -39,6 +41,7 @@
 import java.util.Set;
 import java.util.Collections;
 
+
```
AlanBateman, line 44: Minor nit, I assume this additional blank line is left over from a previous iteration.

AlanBateman:
> https://bugs.openjdk.org/browse/JDK-8310979 - better exception handling https://bugs.openjdk.org/browse/JDK-8310978 - missing code paths for event generation https://bugs.openjdk.org/browse/JDK-8310994 - non-blocking, event for select ops

Do you have a sense yet on events when the channel is configured non-blocking? I realise the threshold is 20ms so they are probably not recorded today but I wonder if these code paths will eventually include `|| !blocking` or if it useful to record data on all socket read/write ops.

AlanBateman (approved):
Overall, it's good to move away from the "faraway instrumentation". Looks forward to the next steps to get to a more complete solution.

egahlin:
The stack trace will start in the incorrect method, i.e checkForCommit, when a utility method is used. This is probably something we will have to fix anyway. I filed an issue for that: https://bugs.openjdk.org/browse/JDK-8314745

tprinzing:
> > https://bugs.openjdk.org/browse/JDK-8310979 - better exception handling https://bugs.openjdk.org/browse/JDK-8310978 - missing code paths for event generation https://bugs.openjdk.org/browse/JDK-8310994 - non-blocking, event for select ops
> 
> Do you have a sense yet on events when the channel is configured non-blocking? I realise the threshold is 20ms so they are probably not recorded today but I wonder if these code paths will eventually include `|| !blocking` or if it useful to record data on all socket read/write ops.

I think it's useful if trying to trace the calls (i.e. set to 0ms).  Apparently the security manager was being used for that by some.

AlanBateman:
> I think it's useful if trying to trace the calls (i.e. set to 0ms). Apparently the security manager was being used for that by some.

The SM isn't called once connected so I don't think anyone could have every done that. Yes, you could set the threshold to 0ms to emit event for every read/write but I wonder how useful that would be, maybe I/O stats would be more interesting.

src/java.base/share/classes/java/net/Socket.java
tprinzing: No.  I think it's a relic from the distant past though.  I think the timeout field should be removed.  It's not used on SocketChannel at all, and it doesn't seem useful on Socket.

AlanBateman:
> @tprinzing 
> Your change (at version [9fa2745](https://github.com/openjdk/jdk/commit/9fa2745960aea0bc45642081bac89fb5ef65809e)) is now ready to be sponsored by a Committer.

@tprinzing I don't mind sponsoring but I think it would help if you could sync up the branch and provide a summary on the testing was done.  The jdk_net and jdk_nio test groups are in tier2. The jdk_jfr test group is in tier3.

src/java.base/share/classes/java/net/Socket.java
dfuch: Should we log an RFE to that effect?

dfuch (commented):
LGTM. Are there existing that will help verify that the read events and write events are still emitted as expected? If not shouldn't we write some?

tprinzing:
I sync'd master and updated the bug report with the successful test results.

Created [JDK-8316558] to track potential timeout field removal.  

The existing JFR tests TestSocketChannelEvents and TestSocketEvents in jdk.jfr.event.io verify the events are still emitted as expected.

dfuch:
> The existing JFR tests TestSocketChannelEvents and TestSocketEvents in jdk.jfr.event.io verify the events are still emitted as expected.

Thanks Tim. Should 8308995 be listed in the `@bug` clause of these two tests?

AlanBateman:
> Thanks Tim. Should 8308995 be listed in the `@bug` clause of these two tests?

I don't think so as these tests are just used to check that changes haven't broken anything.

PR #14349, opened 2023-06-07
8309463: IGV: Dynamic graph layout algorithm

emmyyin:
### Purpose

IGV currently uses a static layout algorithm to visualize graphs. However, this is problematic due to the use cases of IGV. Most often, the graphs that are visualized are dynamic, meaning the graphs change over time. A dynamic graph can be thought of as a sequence of graphs where a given graph in the sequence is the state of the dynamic graph at that point in time. Static layout algorithms do not account for the rest of the sequence when visualizing a given graph. On one hand, it makes each layout more readable. But on the other hand, the layout for two consecutive graphs in the sequence can be vastly different even though the difference between the graphs is small. This makes it difficult to identify the changes that has occurred to the graph and can damage the internal understanding of the graph that the viewer has obtained. A dynamic layout algorithm takes the changes into account when visualizing a graph. To enhance IGV, such an algorithm has been implemented in this PR.

The layout drawn by the static layout algorithm is called "sea of nodes", while the layout drawn by the dynamic algorithm is called "stable sea of nodes".

The difference between the algorithms is illustrated in the following video:

https://github.com/openjdk/jdk/assets/52547536/35023362-a191-425e-b066-c7474db631f1

This work is the result of my Master's thesis which can be found [here](https://kth.diva-portal.org/smash/get/diva2:1770643/FULLTEXT01.pdf).

### Implementation

The algorithm is based on update actions that are applied incrementally to a graph layout in order to obtain the layout of the next graph in the sequence. By doing so, the nodes that appears in both graphs remain in their relative positions. A new layout manager called `HierarchicalStableLayoutManager` has been added which holds the core algorithm. The corresponding layout manager with the static layout algorithm is called `HierarchicalLayoutManager`.

If no layouts have been drawn yet, the `HierarchicalLayoutManager` is used. This is because the dynamic algorithm needs an initial layout to apply the update actions on.

The whole graph is represented by `LayoutNode` and `LayoutEdge` objects, that holds the positions of the nodes and edges along with other relevant information such as ID, name and size. These are updated, added and removed in accordance with the update actions.

Since `HierarchicalStableLayoutManager` tries to preserve the node positions, the layouts might get unreadable after a few iterations. In case that happens, one can "reset" the layout using the `HierarchicalLayoutManager` by clicking the icon for the sea of nodes layout and then go back to the stable sea of nodes layout.

### Testing

The testing has mostly been through manual testing. A program that randomly tests graphs has also been utilized, which can be [found on this branch](https://github.com/emmyyin/jdk/tree/JDK-8309463-test). 

### Miscellaneous

Some things that are helpful to be aware of:

* The code used to test the layout quality and stability for the thesis still remains in the class.
* The dynamic algorithm works best for small changes and is recommended for exploration (expanding and collapsing nodes).
* Every time a change occurs to the displayed graph, the obtained graph is treated as a completely new graph. It is therefore necessary to go through all the `LayoutNode` objects in both graphs to see if they appear in both graphs, and if so update the reference from the old to the new node. Similarly for the edges.

To further enhance the algorithm, one could explore:

* Automatic resetting with `HierarchicalLayoutManager`, e.g. when the layout has too many edge crossings or bent edges
* The order of which the update actions are applied
* How the layers and positions are chosen for nodes that are to be inserted into the graph

### Known issues
* The difference graph functionality (layout highlighting the difference between two chosen graphs) is not working for the `HierarchicalStableLayoutManager`. It has currently been disabled for the stable sea of nodes layout entirely.

---------
### Progress

### Issue
 * [JDK-8309463](https://bugs.openjdk.org/browse/JDK-8309463): IGV: Dynamic graph layout algorithm (**Enhancement** - P4)

### Reviewers
 * [Tobias Holenstein](https://openjdk.org/census#tholenstein) (@tobiasholenstein - **Reviewer**)
 * [Roberto Castañeda Lozano](https://openjdk.org/census#rcastanedalo) (@robcasloz - **Reviewer**) ⚠️ Review applies to [3418f070](https://git.openjdk.org/jdk/pull/14349/files/3418f070f8e8c3279605f9fbd6436b536006ca4d)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14349/head:pull/14349` \
`$ git checkout pull/14349`

`$ git checkout pull/14349` \
`$ git pull https://git.openjdk.org/jdk.git pull/14349/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14349`

`$ git pr show -t 14349`

</details>

<a href="https://git.openjdk.org/jdk/pull/14349.diff">https://git.openjdk.org/jdk/pull/14349.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14349#issuecomment-1682000544)


src/utils/IdealGraphVisualizer/ControlFlow/src/main/java/com/sun/hotspot/igv/controlflow/HierarchicalGraphLayout.java
```diff
@@ -161,7 +161,7 @@ protected void performGraphLayout(UniversalGraph<N, E> graph) {
         HierarchicalLayoutManager m = new HierarchicalLayoutManager(HierarchicalLayoutManager.Combine.NONE);
 
         LayoutGraph layoutGraph = new LayoutGraph(links, vertices);
-        m.doLayout(layoutGraph);
```
tobiasholenstein, line 164: should be uncommented

tobiasholenstein:
A few tips and comments:
- Remove TODOs, unused code (commented out) and personal comments
- Adjust copyright year of all touched files to 2023
- Add comments to important and/or complex parts of your code
- Can we somehow separate the evaluation from the layout algorithm?
   - Either in separate files or at least disable evaluation by default with a flag
- Do we want to keep the shortcuts?
   - Perhaps keep `LayoutAction1` for relayout with static layout algorithm
   - rename `LayoutAction1`  and give it a better shortcut
- ideally leave `HierarchicalLayoutManager.java` untouched or at least make sure other parts of IGV work as before
- rename `new_layout.png` , `NewLayoutManager.java` and `EnableNewLayoutAction.java`
- make sure to document in the PR description what changed other than the new layout algorithm   
   - e.g. what changed in the settings (`ANIMATION_LIMIT`) 
   - what changed in `ServerCompilerScheduler.java`

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalLayoutManager.java
```diff
@@ -1458,9 +1420,8 @@ protected void run() {
 
                 // Whether the node has non-self reversed edges going downwards.
                 // If so, reversed edges going upwards are drawn to the left.
-                boolean hasReversedDown =
-                    reversedDown.size() > 0 &&
-                    !(reversedDown.size() == 1 && hasSelfEdge);
```
tobiasholenstein, line 1463: Please avoid style changes to code that you did not touch otherwise (sometimes the IDE does this automatically)

src/utils/IdealGraphVisualizer/View/src/main/java/com/sun/hotspot/igv/view/DiagramScene.java
```diff
@@ -1178,26 +1210,31 @@ public void animatorPostTick(AnimatorEvent animatorEvent) {
...
 
         updateVisibleFigureWidgets();
         updateNodeHull();
         updateVisibleBlockWidgets();
 
         HashSet<Figure> visibleFigures = getVisibleFigures();
         HashSet<Connection> visibleConnections = getVisibleConnections();
-        if (getModel().getShowSea()) {
+
+        String key = getModel().getGraph().getGroup().getName() + "::" + getModel().getGraph().getName();
```
tobiasholenstein, line 1221: What is the purpose of this? Is it needed?

robcasloz:
Thanks Emmy for creating this pull request!
The changes related to improving animations (merged from https://github.com/openjdk/jdk/compare/master...robcasloz:jdk:igv-stable) are orthogonal to the addition of a dynamic layout algorithm and should be extracted, I think, into a separate RFE. That would simplify the task of reviewing and testing your core changes.

emmyyin:
Some TODOs:

- [ ] Fix relayouting. When the layout drawn by the stable layout manager is no longer readable, redraw using regular layout manager. When user goes back and forth from stable -> regular -> stable it should be "resetted"
- [ ] Ensure filters work properly
- [ ] Fix CF for graphs drawn by stable layout manager (on RHS window panel)

emmyyin:
OBS: does not handle self-edges atm, is it something that should be considered?

robcasloz:
> OBS: does not handle self-edges atm, is it something that should be considered?

Self-edges are only relevant in the CFG view, so we don't need to consider them in the scope of this work.

robcasloz:
I just tried out the PR a little bit and it is starting to look good and robust, good job Emmy!

Sometimes, the layout in "Stable sea of nodes" changes when going from one graph to the next in the group even though there are no graph changes at all. Is there any simple solution for that? E.g. not re-computing the layout at all if no graph changes are detected.

src/utils/IdealGraphVisualizer/View/src/main/java/com/sun/hotspot/igv/view/EditorTopComponent.java
```diff
@@ -190,6 +195,7 @@ public void mouseMoved(MouseEvent e) {}
         cfgLayoutButton.setSelected(diagramViewModel.getShowCFG());
         layoutButtons.add(cfgLayoutButton);
         toolBar.add(cfgLayoutButton);
+
```
tobiasholenstein, line 198: I suggest to deactivate `HierarchicalStableLayoutManager` for difference graphs regarding
> The difference graph functionality (layout highlighting the difference between two chosen graphs) is not working for the HierarchicalStableLayoutManager.

```suggestion
        
        diagramViewModel.getGraphChangedEvent().addListener(model -> {
            // HierarchicalStableLayoutManager is not stable for difference graphs
            boolean isDiffGraph = model.getGraph().isDiffGraph();
            // deactivate HierarchicalStableLayoutManager for difference graphs
            stableSeaLayoutButton.setEnabled(!isDiffGraph);
            if (stableSeaLayoutButton.isSelected() && isDiffGraph) {
                // fallback to HierarchicalLayoutManager for difference graphs
                seaLayoutButton.setSelected(true);
            }
        });
```

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
```diff
@@ -0,0 +1,1913 @@
...
+         * https://www.geeksforgeeks.org/check-if-two-given-line-segments-intersect/
+         *
+         * @param p
+         * @param q
+         * @param r
+         * @return true if point q lies on line segment 'pr'
+         */
+        private boolean onSegment(Point p, Point q, Point r) {
+            if (q.x <= Math.max(p.x, r.x) && q.x >= Math.min(p.x, r.x) &&
+                    q.y <= Math.max(p.y, r.y) && q.y >= Math.min(p.y, r.y))
```
tobiasholenstein, line 1543: you can directly `return (q.x <= Math.max(p.x, r.x) && q.x >= Math.min(p.x, r.x) &&
                    q.y <= Math.max(p.y, r.y) && q.y >= Math.min(p.y, r.y))`

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
```diff
@@ -0,0 +1,1913 @@
...
+         *
+         * @return 0 --> p, q and r are collinear, 1 --> Clockwise, 2 -->
+         *         Counterclockwise
+         */
+        private int orientation(Point p, Point q, Point r) {
+            int val = (q.y - p.y) * (r.x - q.x) -
+                    (q.x - p.x) * (r.y - q.y);
+
+            if (val == 0)
+                return 0; // collinear
```
tobiasholenstein, line 1561: We usually use {} even for one line blocks

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
```diff
@@ -0,0 +1,1913 @@
...
+            // Find the four orientations needed for general and
+            // special cases
+            int o1 = orientation(p1, q1, p2);
+            int o2 = orientation(p1, q1, q2);
+            int o3 = orientation(p2, q2, p1);
+            int o4 = orientation(p2, q2, q1);
+
+            // General case
+            if (o1 != o2 && o3 != o4)
+                return true;
```
tobiasholenstein, line 1586: We usually use {} even for one line blocks

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
```diff
@@ -0,0 +1,1913 @@
...
+            } else {
+                return 0;
+            }
+        }
+
+        /**
+         * Computes how much the bends in an edge deviates from being straight, on
+         * average
+         *
+         * @return
```
tobiasholenstein, line 1711: I would leave away the `@return` here

tobiasholenstein (commented):
Look good to me besides a few minor comments

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
```diff
@@ -0,0 +1,1913 @@
...
+                    n.succs.remove(edgeToRemove);
+                    break;
+                } else {
+                    // Must remove edges between dummy nodes
+                    boolean found = true;
+                    LayoutNode prev = toNode;
+                    while (n.vertex == null && found) {
+                        found = false;
+
+                        if (n.vertex == null && n.succs.size() <= 1 && n.preds.size() <= 1) {
```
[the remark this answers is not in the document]
emmyyin: This is to make the loop break once we hit the non-dummy node where the edge goes from. I.e. for the edge (u,v) with lots of dummy nodes in between nodes u and v, we only want to remove the dummy nodes and then break the loop as soon as we are at node u.

tobiasholenstein (approved):
looks good to me. Thanks for the work! :)

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
```diff
@@ -0,0 +1,1909 @@
...
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+                // Edge crossings across current layer and layer below
+                if (layers.keySet().contains(layer + 1)) {
+                    List<LayoutNode> succsNodes = layers.get(layer + 1);
+                    // For each link with an end point in vertex, check how many edges crosses it
```
robcasloz, line 706:
```suggestion
                    // For each link with an end point in vertex, check how many edges cross it
```

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
```diff
@@ -0,0 +1,1909 @@
...
+            }
+
+            sanityCheckEdges();
+            sanityCheckNodesAndLayerNodes();
+        }
+
+        /**
+         * Calculate which layer the given vertex should be inserted at to minimize
+         * reversed edges and edge lengths
+         * If there are multiple options, choose the bottom most layer
```
robcasloz, line 1104:
```suggestion
         * If there are multiple options, choose the bottom-most layer
```

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
```diff
@@ -0,0 +1,1909 @@
...
+            } else if (layers.keySet().size() == 0) {
+                return 0;
+            }
+
+            int reversedEdges = Integer.MAX_VALUE;
+            int totalEdgeLength = Integer.MAX_VALUE;
+            int neighborsOnSameLayer = Integer.MAX_VALUE;
+            int layer = -1;
+            for (int i = 0; i < layers.keySet().size(); i++) {
+                // System.out.println("Testing layer " + i);
```
robcasloz, line 1122: Please remove this line.

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
```diff
@@ -0,0 +1,1909 @@
...
+
+                curY += maxHeight + baseLine + bottomBaseLine;
+                curY += LAYER_OFFSET + ((int) (Math.sqrt(maxXOffset) * 1.5));
+            }
+        }
+    }
+
+    private class ComputeLayoutScore {
+        /**
+         * https://www.geeksforgeeks.org/check-if-two-given-line-segments-intersect/
```
robcasloz, line 1533: Please replace this reference with a standalone comment describing the method.

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
```diff
@@ -0,0 +1,1909 @@
...
+         * @return true if point q lies on line segment 'pr'
+         */
+        private boolean onSegment(Point p, Point q, Point r) {
+            return q.x <= Math.max(p.x, r.x) && q.x >= Math.min(p.x, r.x) &&
+                    q.y <= Math.max(p.y, r.y) && q.y >= Math.min(p.y, r.y);
+        }
+
+        /**
+         * To find orientation of ordered triplet (p, q, r).
+         * https://www.geeksforgeeks.org/check-if-two-given-line-segments-intersect/
```
robcasloz, line 1547: Please remove this reference.

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
```diff
@@ -0,0 +1,1909 @@
...
+
+            if (val == 0) {
+                return 0; // collinear
+            }
+
+            return (val > 0) ? 1 : 2; // clock or counterclock wise
+        }
+
+        /**
+         * https://www.geeksforgeeks.org/check-if-two-given-line-segments-intersect/
```
robcasloz, line 1564: Please remove this reference.

robcasloz (changes requested):
Looks good! I have tested the patch and checked that no major performance regression is introduced in the existing sea-of-nodes view. I only have a few minor comments about the code.

chhagedorn (commented):
Great work Emmy! I've played around with some graphs and it works quite well for small sets of nodes.

While clicking around in IGV, I've noticed that sometimes it takes a long time to open a new graph. 

Example:
1. Open IGV
2. Run: `java -XX:CompileOnly=*::putUTF8 -XX:+PrintIdealGraph -XX:PrintIdealGraphLevel=3 HelloWorld.java`
3. Open the graph `PhaseCCP1` of the `putUTF8` compilation.
4. Switch to `stable sea of nodes` layout.
5. Open the graph `Before RemoveUseless` -> hangs for ~1min

I've had a closer look at that example with a profiler and it seems that over 90% of the time is spent in `sanityCheckNodesAndLayerNodes()` (~66%) and `sanityCheckEdges()` (~25%) which seems quite a lot for just sanity checks. @robcasloz double checked the example and then ran the same steps from above with the two sanity check methods disabled. On his machine, the time went down from 64s to 2s which is a huge improvement. 

We therefore suggest to either completely disable sanity checking with these two methods or limit it to a few places (not sure how easy that is). If we disable it, we could still think about sanity checking with them in IGV unit tests only.

I'm not very familiar with the IGV code base, so I only have some more general code comments.

Thanks,
Christian

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
```diff
@@ -0,0 +1,1909 @@
...
+        private LayoutEdge addBetween(LayoutEdge e, int layer) {
+            LayoutNode n = new LayoutNode();
+            n.width = DUMMY_WIDTH;
+            n.height = DUMMY_HEIGHT;
+            n.succs.add(e);
+            LayoutEdge result = new LayoutEdge();
+            result.vip = e.vip;
+            n.preds.add(result);
+            result.to = n;
+            result.relativeTo = n.width / 2;
```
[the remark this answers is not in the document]
emmyyin: Yes since `relativeTo` and `relativeFrom` refers to the ports on the node, which is irrelevant for the dummy nodes. Could definitely be done differently, but this is how it is in `HierarchicalLayoutManager` and I thought it would be better to be consistent across the layout managers

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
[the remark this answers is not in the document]
emmyyin: Exactly, if `n.vertex == null` we want to break the loop. There are two cases we need to consider when removing the dummy nodes: 1) there is a long chain of dummy nodes between node u and v, and 2) the edge is part of an edge concentration of multiple edges. In case 1) we just traverse the edge with all the dummy nodes and remove them as we go until we hit node u (`n.vertex != null`). In case 2) we traverse along the edge until we find the anchor node (which is a dummy node with one ingoing edge and multiple outgoing edges), and break the loop when we reach the anchor node.

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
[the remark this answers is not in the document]
chhagedorn: Okay, thanks for the explanation. The code in `HierarchicalLayoutManager` looks very similar. So, you could also change `relativeTo` and `relativeFrom` to zero there. Or even better share the code somehow (if I see that correctly, the only difference is how to insert the node - `nodes.add(n)` vs. `insertNode(n, layer)`).

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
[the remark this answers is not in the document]
chhagedorn: Thanks for explaining it in more details. We probably had a misunderstanding here, though. Toby and I were referring to this `n.vertex == null` here:
```
 if (n.vertex == null && n.succs.size() <= 1 && n.preds.size() <= 1)
```
and not the one in
```
while (n.vertex == null && found) 
```
From your explanation it makes sense to keep the one in the `while` but the other one in the `if` always seems to be true and could thus be removed.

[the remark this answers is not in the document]
emmyyin: Oh okay thanks for clarifying that makes sense!

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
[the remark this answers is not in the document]
emmyyin: @robcasloz has created a suggestion list of how to improve `HierarchicalLayoutManager`. Could we perhaps add this to that list instead? I think there must have been some reason to set `n.width/2` (maybe just in case you want to make the dummy nodes visible with a larger width) so it could be worth to work it through some more

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
```diff
@@ -0,0 +1,1909 @@
...
+    private void sanityCheckEdges() {
+        for (LayoutNode n : nodes) {
+            for (LayoutEdge e : List.copyOf(n.succs)) {
+                assert e.from.equals(n);
+                if (nodes.contains(e.to)) {
+                    assert e.to.preds.contains(e);
+                    assert layers.get(n.layer + 1).contains(e.to);
+                    assert e.to.layer == n.layer + 1;
+                } else {
+                    n.succs.remove(e);
```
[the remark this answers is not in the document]
emmyyin: Yes, it breaks often if removed. It's mostly sanity checking but also ensuring the state of the graph is correct by fixing weird edge connections

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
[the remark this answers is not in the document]
chhagedorn: I'm fine with moving this and my other suggested clean-ups to this list. However, even after removing `sanityCheckNodesAndLayerNodes()` and additionally removing the `assert` checks (which are currently still in the code), the example graph mentioned earlier still needs around 18s to load. The bottleneck is now `sanityCheckEdges()`.

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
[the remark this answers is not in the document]
chhagedorn: As mentioned below, this method is still a bottleneck and we spend quite some time inside it. Overall, it attributes to around 90% of the overall time to open the graph in the example mentioned earlier. Is there another way to fix these breakages instead of looping over all nodes and edges here?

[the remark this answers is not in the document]
emmyyin: I am not sure, I've tried to find other solutions but have not managed. The function can be called less, but not removed entirely. A suggestion is to put it up as a future enhancement?

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
```diff
@@ -0,0 +1,1909 @@
...
+            this.space = space;
+        }
+
+        public int offset(LayoutNode n1, LayoutNode n2) {
+            int v1 = space.get(n1.pos) + n1.width;
+            int v2 = space.get(n2.pos);
+            return v2 - v1;
+        }
+
+        public void insert(LayoutNode n, int pos) {
```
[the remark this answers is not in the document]
emmyyin: I have made the shared parts public in `HierarchicalLayoutManager` and reused in `HierarchicalStableLayoutManager`. Is this the way to go or should we move them out to a  `LayoutManagerUtils` class instead?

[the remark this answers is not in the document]
robcasloz: Reusing directly from `HierarchicalLayoutManager` looks good to me, since `HierarchicalStableLayoutManager` already depends on it.

robcasloz (changes requested):
As far as I understand, the only issue left to be addressed in this PR is the performance overhead caused by `sanityCheckEdges()`. I (together with @tobiasholenstein and @chhagedorn) suggest to replace it with a simpler method that only contains the logic strictly necessary for layout correctness, something like:
```
private void ensureNeighborEdgeConsistency() {
    for (LayoutNode n : nodes) {
        n.succs.removeIf(e -> !nodes.contains(e.to));
        n.preds.removeIf(e -> !nodes.contains(e.from));
    }
}
```
This does not entirely remove the overhead but at least mitigates it to an acceptable level.
While at it, could you remove the unused `sanityCheck...()` methods? Thanks!

src/utils/IdealGraphVisualizer/HierarchicalLayout/pom.xml
```diff
@@ -51,6 +51,11 @@
             <artifactId>Layout</artifactId>
```
chhagedorn, line 51: You should also update the copyright year of this file.

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
```diff
@@ -0,0 +1,1471 @@
...
+            e.relativeFrom = e.link.getFrom().getRelativePosition().x;
+            if (reversedLinks.contains(e.link)) {
+                e.relativeFrom = e.link.getTo().getRelativePosition().x;
+                reversedDown.add(e.relativeFrom);
+            }
+        }
+
+        // Whether the node has non-self reversed edges going downwards.
+        // If so, reversed edges going upwards are drawn to the left.
+        boolean hasReversedDown = reversedDown.size() > 0;
```
chhagedorn, line 283:
```suggestion
        boolean hasReversedDown = !reversedDown.isEmpty();
```

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
```diff
@@ -0,0 +1,1471 @@
...
+
+        SortedSet<Integer> reversedUp = null;
```
chhagedorn, line 285:
```suggestion
        SortedSet<Integer> reversedUp;
```

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
```diff
@@ -0,0 +1,1471 @@
...
+            for (LayoutEdge e : node.preds) {
+                e.relativeTo -= minX;
+            }
+
+            for (LayoutEdge e : node.succs) {
+                e.relativeFrom -= minX;
+            }
+
+            node.xOffset = -minX;
+            node.width += -minX;
```
chhagedorn, line 388: Somewhat confusing. Rather use:
```suggestion
            node.width -= minX;
```

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
```diff
@@ -0,0 +1,1471 @@
...
+            shouldRedrawLayout = false;
+        } else {
+            generateActions();
+
+            new BuildDatastructure().run();
+
+            findInitialReversedLinks();
+
+            // Only apply updates if there are any
+            if (linkActions.size() > 0 || vertexActions.size() > 0) {
```
chhagedorn, line 425:
```suggestion
            if (!linkActions.isEmpty() || !vertexActions.isEmpty()) {
```

src/utils/IdealGraphVisualizer/HierarchicalLayout/src/main/java/com/sun/hotspot/igv/hierarchicallayout/HierarchicalStableLayoutManager.java
```diff
@@ -0,0 +1,1471 @@
...
+                    e.to.y + e.to.yOffset + e.link.getTo().getRelativePosition().y);
+            points.add(p);
+            if (e.to.inOffsets.containsKey(e.relativeTo)) {
+                points.add(new Point(p.x,
+                        p.y + e.to.inOffsets.get(e.relativeTo) + e.link.getTo().getRelativePosition().y));
+            }
+
+            LayoutNode cur = e.from;
+            LayoutEdge curEdge = e;
+            while (cur.vertex == null && cur.preds.size() != 0) {
```
chhagedorn, line 1361:
```suggestion
            while (cur.vertex == null && !cur.preds.isEmpty()) {
```

chhagedorn (approved):
Thanks for the updates! There are only some easy clean-up suggestions left and a copyright year to be updated. Otherwise, it looks good to me. Thanks for your work!

As discussed, we can move everything else to follow-up tasks.

chhagedorn (approved):
Thanks for the updates, looks good!

tobiasholenstein (approved):

tobiasholenstein:
Thank you for your work @emmyyin ! 
Looks good now.

/sponsor

robcasloz:
The slowdown caused by `ensureNeighborEdgeConsistency()` is now documented in [JDK-8315316](https://bugs.openjdk.org/browse/JDK-8315316).

PR #14353, opened 2023-06-07
8309266: C2: assert(final_con == (jlong)final_int) failed: final value should be integer

quadhier:
This patch should fix [JDK-8309266](https://bugs.openjdk.org/browse/JDK-8309266). A `jtreg` test is also added. I'd appreciate any comments and reviews. Thanks in advance!

## Problem Analysis

For the following program,
```
public class Test {

    static boolean flag;

        for (int i = 0; i < 10000; i++) {
            flag = !flag;
            test();
        }
    }

    public static void test() {
        int limit = flag ? Integer.MAX_VALUE : 1000;

        int i = 0;
        while (i < limit) {
            i += 3;
            if (flag) {
                return;
            }
        }
    }
}
```
A `LoopLimitNode` will be generated and its `Limit` input is a `PhiNode`, as depicted in the following picture.

phi_as_limit

During `PhaseCCP`, the `LoopLimitNode::Value()` tries to calculate the constant final value:
https://github.com/openjdk/jdk/blob/16ebf47fe3b0fac7b67acfa589a26abf8843306b/src/hotspot/share/opto/loopnode.cpp#L2289-L2301

The problem is that the assertion in `line 2299` could fail during CCP though it must hold true at the end of CCP. Here is the reason: `PhaseCCP` initializes all nodes with the type `TOP` and iterates in an "arbitrary" order. The following order may happen:
```
28 IfTrue => 34 Region => 36 Phi => 195 LoopLimit => ... => 29 IfFalse
```
1. In `ProjNode::Value()` (`IfTrue` inherits it), the type of `IfTrue` is set to `Type::CONTROL`
https://github.com/openjdk/jdk/blob/fa791119f0b73cd1e110d6a62d3bed58fee5740a/src/hotspot/share/opto/multnode.cpp#L168-L171

2. In `PhiNode::Value()`, only `28 IfTrue`'s correspondence `33 ConI` gets merged (as `29 IfFalse` has not been dealt with yet), then it has a value of `int:max`.
https://github.com/openjdk/jdk/blob/fa791119f0b73cd1e110d6a62d3bed58fee5740a/src/hotspot/share/opto/cfgnode.cpp#L1269-L1277

3. In `LoopLimitNode::Value()`, it finds its `Limit` input `36 Phi` is constant, which triggers the assertion, and the assertion fails since the final value calculated from that constant limit (`int:max`) overflows.

## Solution
Move the overflow check to the end of CCP, where it must not fail.

---------
### Progress

### Issue
 * [JDK-8309266](https://bugs.openjdk.org/browse/JDK-8309266): C2: assert(final_con == (jlong)final_int) failed: final value should be integer (**Bug** - `"3"`)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14353/head:pull/14353` \
`$ git checkout pull/14353`

`$ git checkout pull/14353` \
`$ git pull https://git.openjdk.org/jdk.git pull/14353/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14353`

`$ git pr show -t 14353`

</details>

<a href="https://git.openjdk.org/jdk/pull/14353.diff">https://git.openjdk.org/jdk/pull/14353.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14353#issuecomment-1580735337)

### DIFF
--- a/test/hotspot/jtreg/compiler/loopopts/TestCountedLoopFinalValueMayOverflow.java
+++ b/test/hotspot/jtreg/compiler/loopopts/TestCountedLoopFinalValueMayOverflow.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/**
+ * @test
+ * @bug 8309266
+ * @summary Test that a counted loop generating an exact final value that may sometimes
+ *          overflow will not crash
+ *
+ * @run main/othervm -Xbatch -XX:CompileOnly=compiler.loopopts.TestCountedLoopFinalValueMayOverflow::test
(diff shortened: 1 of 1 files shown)


eme64:
@quadhier ok, good luck!

eme64:
@quadhier So you intend to keep working on this?

eme64:
@quadhier I hope I did not discourage you, my feedback yesterday was a bit scattered and maybe overwhealming, I'm sorry for that.
These things are not easy to get right. I was impressed how far you got!
Let me know if you want to take this back up, or want another task to work on - though a JBS account would help ;)

eme64:
@quadhier Ok, sounds good. Looking forward to your future PR's! :blush:

JesperIRL:
@quadhier Please let me know if there is anything you think is missing from the OpenJDK Developers' Guide, I'd be happy to work on any improvements.

Since you mention the guide, I assume that you have read https://openjdk.org/guide/index.html#contributing-to-an-openjdk-project and in particular https://openjdk.org/guide/index.html#socialize-your-change

PR #14355, opened 2023-06-07
8306841: Generational ZGC: NMT reports Java heap size larger than max heap size

stefank:
ZGC has separated the committing of physical memory from the mapping of the committed memory to virtual memory. It also has asynchronous, lazy unmapping of virtual memory from physical memory. This leads to a situation where multiple virtual memory areas can be mapped to the same physical memory. NMT has a strong assumption that there's a 1-to-1 correspondence between committed memory and its virtual memory areas. Because of this NMT and ZGC is not entirely compatible. ZGC has worked around this by adding NMT hooks where the virtual memory is mapped to the committed memory. This mostly works, but there are situations where we have multiple virtual memory areas mapped to the same physical memory, and that causes the NMT values to be inflated.

I propose that we move the NMT committed memory tracking from the mapping of virtual memory to the actual committing of physical memory.

FWIW, given that NMT and ZGC doesn't agree about how memory is committed, we have to fake the virtual memory addresses reported to NMT. This could probably be noticed if you look for the Java heap addresses in the NMT details output, but I don't see why anyone should be looking for those address for the Java heap in NMT. The interesting number is the amount of committed memory, not the exact addresses, IMHO. This isn't something that we change with this patch, but it can be worth understanding while looking at this Bug and the associated PR.

I've written a small sanity test for the NMT Java Heap values, however it's non-trivial to write a test that efficiently provokes this. I've verified this fix by manually running an over-provisioned SPECjbb2015 run, which results in a lot of splitting of ZGC heap regions, which in turn gives us multiple virtual memory area mapping for the same physical memory.

Side note: the lazy unmapping of virtual memory can cause other problems with too many virtual memory areas. The inflated NMT numbers have been a smoking gun showing us that issue. We are tracking that issue with [JDK-8308783](https://bugs.openjdk.org/browse/JDK-8308783).

---------
### Progress

### Issue
 * [JDK-8306841](https://bugs.openjdk.org/browse/JDK-8306841): Generational ZGC: NMT reports Java heap size larger than max heap size (**Bug** - `"4"`)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14355/head:pull/14355` \
`$ git checkout pull/14355`

`$ git checkout pull/14355` \
`$ git pull https://git.openjdk.org/jdk.git pull/14355/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14355`

`$ git pr show -t 14355`

</details>

<a href="https://git.openjdk.org/jdk/pull/14355.diff">https://git.openjdk.org/jdk/pull/14355.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14355#issuecomment-1580844579)

### DIFF
--- a/src/hotspot/share/gc/z/zPhysicalMemory.cpp
+++ b/src/hotspot/share/gc/z/zPhysicalMemory.cpp
@@ -276,6 +276,13 @@ void ZPhysicalMemoryManager::try_enable_uncommit(size_t min_capacity, size_t max
 }
 
 void ZPhysicalMemoryManager::nmt_commit(zoffset offset, size_t size) const {
+  // NMT expects a 1-to-1 mapping between virtual and physical memory.
+  // ZGC can temporarily have multiple virtual addresses pointing to
+  // the same physical memory.
+  //
+  // When this function is called we don't know where in the virtual memory
+  // this physical memory will be mapped. So we fake that the virtual memory
+  // address is the heap base + the given offset.
   const zaddress addr = ZOffset::address(offset);
   MemTracker::record_virtual_memory_commit((void*)untype(addr), size, CALLER_PC);
 }
@@ -320,6 +327,11 @@ bool ZPhysicalMemoryManager::commit(ZPhysicalMemory& pmem) {
 
     // Commit segment
     const size_t committed = _backing.commit(segment.start(), segment.size());
+
+    // Register with NMT
+    nmt_commit(segment.start(), committed);
+
+    // Register committed segment
     if (!pmem.commit_segment(i, committed)) {
       // Failed or partially failed
       return false;
@@ -341,6 +353,11 @@ bool ZPhysicalMemoryManager::uncommit(ZPhysicalMemory& pmem) {
 
     // Uncommit segment
     const size_t uncommitted = _backing.uncommit(segment.start(), segment.size());
+
+    // Unregister with NMT
+    nmt_uncommit(segment.start(), uncommitted);
+
+    // Deregister uncommitted segment
     if (!pmem.uncommit_segment(i, uncommitted)) {
       // Failed or partially failed
       return false;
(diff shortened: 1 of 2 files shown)


fisk (approved):
Looks good.

src/hotspot/share/gc/z/zPhysicalMemory.cpp
```diff
@@ -276,6 +276,13 @@ void ZPhysicalMemoryManager::try_enable_uncommit(size_t min_capacity, size_t max
 }
 
 void ZPhysicalMemoryManager::nmt_commit(zoffset offset, size_t size) const {
+  // NMT expects a 1-to-1 mapping between virtual and physical memory.
+  // ZGC can temporarily have multiple virtual addresses pointing to
+  // the same physical memory.
+  //
+  // When this function is called we don't know where in the virtual memory
+  // this physical memory will be mapped. So we fake that the virtual memory
+  // address is the heap base + the given offset.
```
tstuefe, line 285: Question of a casual ZGC source reader: when you talk about physical vs virtual here, you are not talking about the real physical vs virtual, right? You are talking about offsets into the ZGC backing file vs attach points of said offsets in the virtual address space?

tstuefe (approved):
Looks good. 

Question: why is this limited to generational ZGC? Just a decision not to fix old ZGC, or does it not happen with old ZGC?

> FWIW, given that NMT and ZGC doesn't agree about how memory is committed, we have to fake the virtual memory addresses reported to NMT. This could probably be noticed if you look for the Java heap addresses in the NMT details output, but I don't see why anyone should be looking for those address for the Java heap in NMT.

We do, but it is not such an important use case: in hs_err file "unknown pointer" printing, I use NMT to make sense of an otherwise unknown address.

stefank:
> Question: why is this limited to generational ZGC? Just a decision not to fix old ZGC, or does it not happen with old ZGC?

It's a little bit of both. We are mainly focusing on improving Generational ZGC, but I actually did create a Bug for the Singlegen ZGC (JDK-8309607) for this. The fix in this PR was actually an old patch for Generational ZGC that I had laying around that I had already tested. We can probably still fix this for Singlegen ZGC, but it needs a little bit more care since we have yet another layer with the multi-mapping.

stefank:
> We do, but it is not such an important use case: in hs_err file "unknown pointer" printing, I use NMT to make sense of an otherwise unknown address.

OK. I didn't think about that newish feature. However, if those pointers were actually into the Java Heap they wouldn't be "unknown pointers" but instead reported as uncolored ZGC pointers by the code in ZCollectedHeap::print_location (IIUC). Though I see your point, and maybe there's a way to rewrite NMT in the future.

src/hotspot/share/gc/z/zPhysicalMemory.cpp
stefank: Right, but the backing file is backed by physical memory and that's where the name comes from. The backing file is just the way for us to get hold of physical memory, which we can map into the virtual address space. I hope that makes sense.

tstuefe: It does, thank you.

PR #14357, opened 2023-06-07
8285368: Overhaul doc-comment inheritance

pavelrappo:
Please review this long-awaited change to documentation inheritance.

This change improves "methods comment algorithm" and introduces directed documentation inheritance. While "methods comment algorithm" -- automatic search for inheritable documentation -- has been improved, it still cannot read an author's mind so as to always find the documentation they intended. From now on, an author can state their intention, by providing an FQN of the superclass or superinterface from which to inherit documentation:
```
​{@inheritDoc S}
```
Which is exactly what I did to counterbalance some of the JDK API Documentation changes caused by the change to "methods comment algorithm".

---------
### Progress
- [x] Change requires CSR request [JDK-8287152](https://bugs.openjdk.org/browse/JDK-8287152) to be approved

### Issues
 * [JDK-8285368](https://bugs.openjdk.org/browse/JDK-8285368): Overhaul doc-comment inheritance (**Bug** - P3)
 * [JDK-6376959](https://bugs.openjdk.org/browse/JDK-6376959): Algorithm for Inheriting Method Comments seems to go not as documented (**Bug** - P3)
 * [JDK-6934301](https://bugs.openjdk.org/browse/JDK-6934301): Support directed inheriting of class comments with @inheritDoc (**Enhancement** - P3)
 * [JDK-8287152](https://bugs.openjdk.org/browse/JDK-8287152): Overhaul doc-comment inheritance (**CSR**)

### Reviewers
 * [Jonathan Gibbons](https://openjdk.org/census#jjg) (@jonathan-gibbons - **Reviewer**) ⚠️ Review applies to [011d7982](https://git.openjdk.org/jdk/pull/14357/files/011d7982546c4e268bb3b1fbc71bd14dd87fbec9)
 * [Roger Riggs](https://openjdk.org/census#rriggs) (@RogerRiggs - **Reviewer**) ⚠️ Review applies to [563a8761](https://git.openjdk.org/jdk/pull/14357/files/563a8761370afea9d9bec9b6e22842c726f8843f)
 * [Alexey Ivanov](https://openjdk.org/census#aivanov) (@aivanov-jdk - **Reviewer**) ⚠️ Review applies to [563a8761](https://git.openjdk.org/jdk/pull/14357/files/563a8761370afea9d9bec9b6e22842c726f8843f)
 * [Stuart Marks](https://openjdk.org/census#smarks) (@stuart-marks - **Reviewer**) ⚠️ Review applies to [563a8761](https://git.openjdk.org/jdk/pull/14357/files/563a8761370afea9d9bec9b6e22842c726f8843f)
 * [Martin Buchholz](https://openjdk.org/census#martin) (@Martin-Buchholz - **Reviewer**) ⚠️ Review applies to [563a8761](https://git.openjdk.org/jdk/pull/14357/files/563a8761370afea9d9bec9b6e22842c726f8843f)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14357/head:pull/14357` \
`$ git checkout pull/14357`

`$ git checkout pull/14357` \
`$ git pull https://git.openjdk.org/jdk.git pull/14357/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14357`

`$ git pr show -t 14357`

</details>

<a href="https://git.openjdk.org/jdk/pull/14357.diff">https://git.openjdk.org/jdk/pull/14357.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14357#issuecomment-1580940896)


src/jdk.compiler/share/classes/com/sun/tools/javac/parser/DocCommentParser.java
```diff
@@ -1301,13 +1301,33 @@ public DCTree parse(int pos) throws ParseException {
...
+                        //
+                        // Although it does not seem possible to plug in a custom
+                        // DocTreeFactory, for backward compatibility, we should
+                        // call the former method when there's no reference
+                        // rather the latter and pass null.
+                        //
+                        // Long story short: do NOT refactor the below if-else to
+                        //
+                        //     m.at(pos).newInheritDocTree(ref)
+                        //
```
jonathan-gibbons, line 1325: Although the scary warning is not wrong, I think it would have been enough to say something more concise, like "_use the original legacy method if no ref is given_"

pavelrappo:
For details on changes to "methods comment algorithm", see the associated CSR: https://bugs.openjdk.org/browse/JDK-8287152

pavelrappo:
Full disclosure. Here are all the files that are different before (1) and after (2) the change:

 - https://cr.openjdk.org/~prappo/8285368/1/api
 - https://cr.openjdk.org/~prappo/8285368/2/api

As you can see, the vast majority of differences are in "See Also" sections, which come from `@see` tags. Sadly, `@see` tags are inheritable "by omission". That said, `@see` is also underspecified, somewhat broken, and I believe is not as important as those parts of a doc comment that allow `{@inheritDoc}`.

I suggest that we integrate this PR and take care of `@see` in 22. But if the corrections are really required, I can try to introduce explicit `@see` tags, so there are no changes whatsoever.

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/taglets/InheritDocTaglet.java
```diff
@@ -70,25 +73,72 @@ public InheritDocTaglet() {
...
+            // (S is the same type as the type that contains {@inheritDoc S})
+            // due to irreflexivity of the "overrides" relationship.
+            //
+            // This way we do more work in erroneous case, but less in the typical
+            // case. We don't optimize for the former.
+            VisibleMemberTable visibleMemberTable = configuration.getVisibleMemberTable(supertype);
+            List<Element> methods = visibleMemberTable.getAllVisibleMembers(VisibleMemberTable.Kind.METHODS);
+            for (Element e : methods) {
+                ExecutableElement m = (ExecutableElement) e;
+                if (configuration.utils.elementUtils.overrides(method, m, (TypeElement) method.getEnclosingElement())) {
```
jonathan-gibbons, line 110: Suggestion for eventual future cleanup (not this PR): reduce need for chain fields like`config.utils.elementUtils`

jonathan-gibbons, line 110: Note that `configuration.utils` is available as `utils`

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/taglets/InheritDocTaglet.java
```diff
@@ -70,25 +73,72 @@ public InheritDocTaglet() {
...
         DocTree holderTag = path.getLeaf();
+
+        ExecutableElement src = null;
+        // 1. Does @inheritDoc specify a type?
+        if (inheritDoc.getSupertype() != null) {
+            // 2. Can we find that type?
+            var supertype = (TypeElement) ch.getReferencedElement(inheritDoc.getSupertype());
+            if (supertype == null) {
+                messages.error(inheritDocPath, "doclet.inheritDocBadSupertype");
+                return replacement;
```
jonathan-gibbons, line 94: `replacement` may be empty.
Consider using `HtmlDocletWriter.invalidTagOutput`.

(Later, we might want to provide a utility message to wrap `messages.error` and `invalidTagOutput`

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/taglets/InheritDocTaglet.java
```diff
@@ -70,25 +73,72 @@ public InheritDocTaglet() {
...
+                    assert !method.equals(m) : Utils.diagnosticDescriptionOf(method);
+                    src = m;
+                    break;
+                }
+            }
+            if (src == null) {
+                // "self-inheritors" and supertypes that do not contain a method
+                // that this method overrides
+                messages.error(inheritDocPath, "doclet.inheritDocBadSupertype");
+                return replacement;
```
jonathan-gibbons, line 120: Another case for `invalidTagOutput` ?

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/taglets/InheritableTaglet.java
```diff
@@ -52,7 +52,7 @@ public interface InheritableTaglet extends Taglet {
      * In the future, this could be reworked using some other mechanism,
      * such as throwing an exception.
```
jonathan-gibbons, line 53: Should `ThrowsTaglet.Failure` (eventually?) be declared here in `InheritableTaglet`

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/DocFinder.java
```diff
@@ -47,45 +44,34 @@ public class DocFinder {
...
-    DocFinder(Function<ExecutableElement, ExecutableElement> overriddenMethodLookup,
-              BiFunction<ExecutableElement, ExecutableElement, Iterable<ExecutableElement>> implementedMethodsLookup) {
+    DocFinder(Function<ExecutableElement, Iterable<? extends ExecutableElement>> overriddenMethodLookup) {
         this.overriddenMethodLookup = overriddenMethodLookup;
-        this.implementedMethodsLookup = implementedMethodsLookup;
     }
 
     @SuppressWarnings("serial")
-    public static final class NoOverriddenMethodsFound extends Exception {
+    public static final class NoOverriddenMethodFound extends Exception {
```
jonathan-gibbons, line 54: General comment, for future work:
Maybe we can combine/merge/unify this exception with `ThrowsTaglet.Failure`

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/Utils.java
```diff
@@ -655,6 +661,16 @@ public OverrideInfo overriddenMethod(ExecutableElement method) {
...
     }
 
+    public boolean isDirectSupertype(TypeElement overrider,
+                                     TypeElement overridden) {
+//        var t1 = overrider.asType();
+//        var t2 = overridden.asType();
+//        return typeUtils.directSupertypes(t1).stream()
+//                .anyMatch(t -> Objects.equals(typeUtils.asElement(t), typeUtils.asElement(t2)));
+
+        return true; /* disabled for causing issues in JDK API Documentation build */
```
jonathan-gibbons, line 671: Please describe the issues and/or provide a JBS issue, so that we can decide when to revert the code to the commented-out form.

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/Utils.java
```diff
@@ -2748,14 +2764,125 @@ public DocFinder docFinder() {
...
+
+                // consider only the declared methods for consistency with
+                // the existing facilities, such as Utils.overriddenMethod
+                // and VisibleMemberTable.getImplementedMethods
+                TypeElement peek = searchStack.peek();
+                if (peek == null) {
+                    next = null; // end-of-hierarchy
+                    break;
+                }
+                if (isPlainInterface(peek) && !isPublic(peek) && !isLinkable(peek)) {
```
jonathan-gibbons, line 2847: What is the significance of the `isPublic` check?
I'm developing a knee-jerk reaction to such checks, in the face of access-related command-line options.

test/langtools/jdk/javadoc/doclet/testDirectedInheritance/TestDirectedInheritance.java
```diff
@@ -0,0 +1,767 @@
...
+                public interface E2 extends I1 {
+                    /** @foo {@inheritDoc I1} */
+                    void m();
+                }
+                """);
+        // DocLint should neither prevent nor cause a crash. Explicitly check that
+        // there's no crash with DocLint on and off, but don't check that the exit
+        // code is OK, it likely isn't (after all, there's an unknown tag).
+        setAutomaticCheckNoStacktrace(true);
+        { // DocLint is on
```
jonathan-gibbons, line 75: "on" -> "explicit"

test/langtools/jdk/javadoc/doclet/testDirectedInheritance/TestDirectedInheritance.java
```diff
@@ -0,0 +1,767 @@
...
+            int i = 0;
+            for (var check : new String[]{":all", ":none", ""}) {
+                var outputDir = "out-DocLint-" + i++; // use separate output directories
+                javadoc("-Xdoclint" + check,
+                        "-d", base.resolve(outputDir).toString(),
+                        "--source-path", src.toString(),
+                        "x");
+            }
+        }
+        // DocLint is off
```
jonathan-gibbons, line 85: "off" -> "default"

test/langtools/jdk/javadoc/doclet/testDirectedInheritance/TestDirectedInheritance.java
```diff
@@ -0,0 +1,767 @@
...
+                &nbsp;<span class="element-name">m</span>()</div>
+                <div class="block">A.m()</div>""";
+        new OutputChecker("x/C1.html").check(m);
+        new OutputChecker("x/C2.html").check(m);
+        new OutputChecker(Output.OUT).setExpectFound(false)
+                .check("warning: not a direct supertype"); // no unexpected warnings
+    }
+
+    /*
+     * C1.m inherits documentation from B1.m explicitly and undirect.
```
jonathan-gibbons, line 289: possible typo: do you mean "indirect"

test/langtools/jdk/javadoc/doclet/testDirectedInheritance/TestDirectedInheritance.java
```diff
@@ -0,0 +1,767 @@
...
+     * allow a method doc comment to inherit from that method type's
+     * indirect (not immediate) supertypes. This is similar to how an
+     * implementation cannot write, say, super.super.m().
+     *
+     *
+     * Directed documentation inheritance was primarily designed to
+     * resolve possible ambiguity among immediate supertypes.
+     *
+     * For now a warning is issued if a doc comment inherits from
+     * an indirect supertype.
```
jonathan-gibbons, line 673: Not sure I agree with this. Can we discuss?

test/langtools/jdk/javadoc/doclet/testMethodCommentAlgorithm/TestMethodCommentsAlgorithm.java
```diff
@@ -0,0 +1,459 @@
...
+import com.sun.source.tree.VariableTree;
+import com.sun.source.util.DocTrees;
+import com.sun.source.util.JavacTask;
+import com.sun.source.util.TreePath;
+import com.sun.source.util.TreePathScanner;
+import javadoc.tester.JavadocTester;
+import jtreg.SkippedException;
+import toolbox.ToolBox;
+
+/*
```
jonathan-gibbons, line 61: Great description! :-)

test/langtools/jdk/javadoc/doclet/testMethodCommentAlgorithm/TestMethodCommentsAlgorithm.java
```diff
@@ -0,0 +1,459 @@
...
+    @Test
+    public void testInterfaceHierarchy(Path base) throws Exception {
+        Path p = Path.of(System.getProperty("test.src", ".")).toAbsolutePath();
+        while (!Files.exists(p.resolve("TEST.ROOT"))) {
+            p = p.getParent();
+            if (p == null) {
+                throw new SkippedException("can't find TEST.ROOT");
+            }
+        }
+        System.err.println("Test suite root: " + p);
```
jonathan-gibbons, line 273: You might want to use `JavadocTester.out` instead of `System.err`

test/langtools/jdk/javadoc/doclet/testMethodCommentAlgorithm/TestMethodCommentsAlgorithm.java
```diff
@@ -0,0 +1,459 @@
...
+ *
+ * 2. The documentation for java.lang.Object is slightly amended for
+ *    uniformity with test documentation and for additional test
+ *    coverage.
+ *
+ * 3. While documentation for java.lang.Object is currently inaccessible outside
+ *    of the JDK, these test mimic what happens when the JDK documentation is
+ *    built.
+ */
+public class TestMethodCommentsAlgorithm extends JavadocTester {
```
jonathan-gibbons, line 111: General comment for entire class:  great/clever test!

test/langtools/jdk/javadoc/doclet/testDirectedInheritance/TestDirectedInheritance.java
pavelrappo: Sure we can; this relates to an earlier comment of yours on Utils.isDirectSupertype here: https://github.com/openjdk/jdk/pull/14357#discussion_r1222053011

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/Utils.java
pavelrappo: I mentioned it in the test, here: TestDirectedInheritance.testNotDirectSupertype. I reckon referring to code from test is okay, but not the other way around.

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/Utils.java
pavelrappo: That directly mimics the current behaviour at addSuperInterfaces(Utils.java:802), which in turn is used at ImplementedMethods(VisibleMemberTable.java:1064) to collect overridden methods in superinterfaces. If you look around, it's quite an idiom and predates this PR.

test/langtools/jdk/javadoc/doclet/testDirectedInheritance/TestDirectedInheritance.java
jonathan-gibbons: The general criticism here is whether we should restrict in any way the set of super types from which one can inherit documentation, and/or what should the set of checks be?

For example, if a method in C inherits a method in B that has no comment but which inherits a method with a comment in A, then it seems unduly restrictive to stop the method in C explicitly referencing the method in A, as compared to only allowing a reference to B, which implicitly gets its comment from A.

That being said, there is merit in starting off with restrictions and loosening them in the face of experience, rather than trying to increase the stylistic restrictions later.

jonathan-gibbons: > Sure we can; this relates to an earlier comment of yours on Utils.isDirectSupertype here: [#14357 (comment)](https://github.com/openjdk/jdk/pull/14357#discussion_r1222053011)

Thanks for tying these together; I had not realized they were related.

jonathan-gibbons (approved):
Great work in a difficult area overflowing with Technical Debt.

I've made various (optional) suggestions, I'll approve for now, but will look again if you want to make ore changes.

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/taglets/InheritDocTaglet.java
pavelrappo: Done in d579a83cc4a.

test/langtools/jdk/javadoc/doclet/testDirectedInheritance/TestDirectedInheritance.java
pavelrappo: I'm neither a grammarian nor a native English speaker, but my hunch is that _indirect_ would be wrong here. But again, I'm open to corrections.

I believe "directed inhertiance" came from an analogy with the graph theory, where a graph can be directed and undirected. The former has pointy arrows, the latter does not.

By that analogy, documentation inheritance can be directed `{@inheritDoc <FQN>}` (I choose), or undirected `{@inheritDoc}` (JavaDoc chooses / I'm Feeling Lucky). In addition to that, documentation inheritance can be implicit (i.e. by omission), which works the same way as undirected but without extra keystrokes.

While those concepts are lacking better names, can I suggest we use the following almost-made-up adverbs: _directedly_ and _undirectedly_? As a side note, we badly need a JavaDoc vocabulary.

test/langtools/jdk/javadoc/doclet/testDirectedInheritance/TestDirectedInheritance.java
pavelrappo: Done in 93adde63ed0.

test/langtools/jdk/javadoc/doclet/testDirectedInheritance/TestDirectedInheritance.java
pavelrappo: Done in 93adde63ed0.

test/langtools/jdk/javadoc/doclet/testDirectedInheritance/TestDirectedInheritance.java
pavelrappo: Have a look at this commit: 3da4f473a7c

dfuch (commented):
I will only comment on the changes to the `java.base` classes.

I had a look at the changes to these and they looked right to me. Though I’m not an expert on collections it seemed that you picked the right ‘semantically closer parent’ to inherit from.
Probably @stuart-marks  could give a more authoritative opinion.

test/langtools/jdk/javadoc/doclet/testDirectedInheritance/TestDirectedInheritance.java
pavelrappo: As agreed out of band, I removed that check and the test for now in dea776e285b.

src/java.base/share/classes/java/util/TreeMap.java
```diff
@@ -1195,6 +1195,8 @@ public Collection<V> values() {
      * {@code Set.remove}, {@code removeAll}, {@code retainAll} and
      * {@code clear} operations.  It does not support the
      * {@code add} or {@code addAll} operations.
+     *
+     * @return {@inheritDoc SortedMap}
```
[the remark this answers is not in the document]
pavelrappo: If we were to (re-)structure doc comments in the way you propose, I'd suggest we do it in a separate, non-jdk.javadoc  PR. I was simply trying to keep JDK API documentation unchanged.

One thing with the `@return` documentation inherited from SortedMap is that compared to the first sentence of TreeMap, it provides one extra bit of information: a set view of the mappings contained in this map, **sorted in ascending key order**. In order to keep that bit, the suggested inline `{@return ...}` should include more than just the first sentence.

src/java.base/share/classes/java/util/concurrent/LinkedBlockingDeque.java
```diff
@@ -628,7 +628,9 @@ public boolean add(E e) {
     }
 
     /**
+     * {@inheritDoc BlockingDeque}
      * @throws NullPointerException if the specified element is null
+     * @return {@inheritDoc BlockingDeque}
      */
     public boolean offer(E e) {
```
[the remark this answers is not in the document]
pavelrappo: The directed `{@inheritDoc <FQN>}` works for the main description, `@throws`, `@param` and `@return` tags. If your question is about why that particular doc comment does not use this:

    @param e {@inheritDoc BlockingDeque}

then the answer is: it's not necessary for keeping the API documentation unchanged.

pavelrappo:
@Martin-Buchholz, could you please have a look at changes to collections and, in particular, those in java.util.concurrent.*?

pavelrappo:
Note that the change to "method comments algorithm" uncovers and fixes some bugs to Object- or Object-like- methods. 

Using the provided [before][] and [after][] outputs, compare documentation for:

 * toString() in java.time.chrono.{Hijrah,Japanese,Minguo,ThaiBuddhist}Date
 * hashCode() in Hashtable, ConcurrentHashMap.KeySetView, java.util.jar.Attributes, com.sun.net.httpserver.Headers, com.sun.management.GcInfo
 * equals() in ConcurrentHashMap.KeySetView, com.sun.net.httpserver.Headers, com.sun.management.GcInfo
 * clone() in javax.management.ImmutableDescriptor, javax.management.modelmbean.DescriptorSupport, javax.naming.directory.BasicAttribute, javax.naming.directory.BasicAttributes
 
Perhaps some of those could be further refined using directed documentation inheritance. For example, ConcurrentHashMap.KeySetView is first and foremost a set and only then a collection. That suggests that the documentation for Object-like methods could be inherited from (some) set rather than a "generic" collection.

Also, the javax.management.ImmutableDescriptor and javax.management.modelmbean.DescriptorSupport clone methods require some attention from area experts. It's likely that those methods' doc comments need to fully inherit documentation from Descriptor.clone(), not just its `@return` tag.

[before]: https://cr.openjdk.org/~prappo/8285368/1/api
[after]: https://cr.openjdk.org/~prappo/8285368/2/api

src/java.base/share/classes/java/util/TreeMap.java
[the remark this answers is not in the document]
AlanBateman: > If we were to (re-)structure doc comments in the way you propose, I'd suggest we do it in a separate, non-jdk.javadoc PR. I was simply trying to keep JDK API documentation unchanged.

Understood, it was just a passing comment that TreeMap::entrySet overrides the description so it's specifying what it returns and using inherited text for the @return description.

test/langtools/jdk/javadoc/doclet/testMethodCommentAlgorithm/TestMethodCommentsAlgorithm.java
pavelrappo: Fixed in:  cccffa660a8.

src/java.base/share/classes/java/util/concurrent/LinkedBlockingDeque.java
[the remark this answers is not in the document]
aivanov-jdk: I admit I can't parse this sentence:

> then the answer is that it's not necessary for keeping the API documentation unchanged.

Do you mean that you didn't add `@param e {@inheritDoc BlockingDeque}` here to keep the documentation unchanged?

You added `@inheritDoc` for `@param` tags in [FileCacheImageOutputStream.java](https://github.com/openjdk/jdk/pull/14357/files#diff-bc1e47b4c5c405826e7a062cabea04349ef72d5823b67e8616a680dcfcb5d4be) and others but not here.

So, do we need to add `@param` tags if the entire javadoc is inherited with `@inheritDoc`?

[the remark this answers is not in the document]
pavelrappo: Copying an individual doc element using `{@inheritDoc}` never adds meta information on the generated HTML page to indicate where that doc element was copied from. It so happens in this particular case that the doc element's contents are the same in both possible supertypes (BlockingDeque and Queue) and look like this:

    @param e the element to add

If the end result _on the HTML page_ will look identical no mater if I add `@param e {@inheritDoc BlockingDeque}` or not, I will not add it in _this_ PR. This PR's goal is not restructuring the doc comment; in fact, such restructuring would be an anti-goal. It might be a goal for subsequent, area-specific PRs.

Above I said that "`{@inheritDoc}` never adds meta information on the generated HTML page". You might have immediately thought about cases where you saw the "Description copied from interface/class" headings. However, those do not come from `{@inheritDoc}`; those come from empty comments (which inherit their content by omission, if you wish). So those headings DO indicate the source of the complete doc comment.

Now about FileCacheImageOutputStream.java. If I hand't added directed inheritance there, the updated algorithm would've picked a different comment, which wouldn't have been identical to what was picked before:

 - javax.imageio.stream.ImageInputStream#flushBefore
    ```
     * @param pos a {@code long} containing the length of the
     * stream prefix that may be flushed.
    ```

 - javax.imageio.stream.ImageOutputStream#flushBefore:
    ```
     * @param pos a {@code long} containing the length of the
     * stream prefix that may be flushed to the destination.
    ```
Hence, the correction. Does it clarify the matter?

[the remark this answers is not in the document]
pavelrappo: Also, for better or worse, a lone `{@inheritDoc}` never inherits "everything" from a supertype:
```
/** {@inheritDoc} */
```
It merely inherits the main description of a method. Other elements such as parameter, exception or return information are inherited individually.

pavelrappo:
core-libs, client: any reviewers?

RogerRiggs (approved):
Look ok for core-libs classes

src/java.base/share/classes/java/util/concurrent/LinkedBlockingDeque.java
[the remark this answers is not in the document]
aivanov-jdk: Thank you, it helps.

aivanov-jdk (approved):
Looks good to me for client changes.

stuart-marks (approved):
I reviewed the changes in java.base. They look fine. I support the notion of keeping these changes minimal, with the narrow goal of keeping the resulting javadoc output the same. There are many opportunities for improving the actual docs, but those should be handled separately.

Martin-Buchholz:
> I reviewed the changes in java.base. They look fine. I support the notion of keeping these changes minimal, with the narrow goal of keeping the resulting javadoc output the same. There are many opportunities for improving the actual docs, but those should be handled separately.

I agree with stuart-marks@

src/java.base/share/classes/java/util/concurrent/ConcurrentHashMap.java
```diff
@@ -1531,7 +1531,7 @@ private void readObject(java.io.ObjectInputStream s)
     // ConcurrentMap methods
 
     /**
-     * {@inheritDoc}
+     * {@inheritDoc ConcurrentMap}
```
Martin-Buchholz, line 1534: This clarifies the authors' intent, so is clear progress. 

Semantic difficulties remain:
- it would be nice to somehow declare that we never ever want to inherit doc from AbstractMap (or even at the module level, declare that AbstractFoo classes are specified only for subclass writers, not users).  Private inheritance?
- we never want to inherit AbstractFoo @implNotes
- ConcurrentMap does *not* have the same spec as ConcurrentHashMap, e.g. the latter does not permit null values.  Therefore one can argue that javadoc should *not* be inherited here.  Right now the main method spec from ConcurrentMap is perfectly suitable for ConcurrentHashMap, but one can imagine a future change to ConcurrentMap::putIfAbsent's spec that changes that, perhaps due to the null value handling difference.  We have a distasteful choice - brittle inheritance or copy-pasta.  In practice not so bad here, since these classes are maintained together.

Martin-Buchholz (approved):
We don't currently have a way to diff rendered javadoc, like my ancient BlenderRev hack?

pavelrappo:
> We don't currently have a way to diff rendered javadoc, like my ancient BlenderRev hack?

I've generated something you might find helpful: https://cr.openjdk.org/~prappo/8285368/specdiff/overview-summary.html

src/java.base/share/classes/java/util/concurrent/ConcurrentHashMap.java
pavelrappo: > * we never want to inherit AbstractFoo @implNotes

`@implNote`, `@apiNote` and `@implSpec` can only be inherited via explicit `{@inheritDoc}`:

    @implSpec {@inheritDoc}

> * ConcurrentMap does _not_ have the same spec as ConcurrentHashMap, e.g. the latter does not permit null values.  Therefore one can argue that javadoc should _not_ be inherited here.  Right now the main method spec from ConcurrentMap is perfectly suitable for ConcurrentHashMap, but one can imagine a future change to ConcurrentMap::putIfAbsent's spec that changes that, perhaps due to the null value handling difference.  We have a distasteful choice - brittle inheritance or copy-pasta.  In practice not so bad here, since these classes are maintained together.

That state of affairs predates this PR and is merely highlighted by it. Sadly, I'm not sure how JavaDoc can help here. Annotations, contracts, inspections, and doc tests come to mind; but none of these are supported by JavaDoc at the moment.

Martin-Buchholz: (There's nothing wrong with this PR)

Back in the 1980s there was a lot of optimism that OOP would be a big leap forward in software engineering.
But especially the "inheritance" part of OOP has been disappointing - unexpectedly brittle!
javadoc inheritance in collection classes was particularly interesting to me from this perspective.
Especially with the multiple inheritance that java was mostly reluctant to adopt.

PR #14360, opened 2023-06-07
8307526: [JFR] Better handling of tampered JFR repository

jaokim:
This change makes sure that "JFR Periodic Task" isn't interrupted by any errors. This can happen when an unfinished chunkfile is removed "in-flight", which would lead to the chunkfiles not being rotated properly, and the maxsize/maxage being ignored.

With this fix, when an unfinished chunkfile is detected, all of the chunkfiles in the recording are checked for existence (since one likely cause for this could be f.i. 'rm -r /tmp' being invoked, effectively deleting all chunkfiles). Upon seeing missing chunkfiles, an error is logged, and emitted to the recording to signal that some data might be missing.
A check to catch any missing chunkfiles is also added when dumping a recording -- this can happen if only finished chunkfiles have been removed, which is not detected in-flight.

Also, a check was added to the file purger to check whether the file had already been deleted to not fill the purge list with already removed files.

### Testing
 - [x] tier1-3
 - [x] jdk_jfr

---------
### Progress

### Issue
 * [JDK-8307526](https://bugs.openjdk.org/browse/JDK-8307526): [JFR] Better handling of tampered JFR repository (**Bug** - P2)

### Reviewers
 * [Erik Gahlin](https://openjdk.org/census#egahlin) (@egahlin - **Reviewer**) ⚠️ Review applies to [4f7427c7](https://git.openjdk.org/jdk/pull/14360/files/4f7427c7e391bc4f92afc929ab55547f7c6fd626)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14360/head:pull/14360` \
`$ git checkout pull/14360`

`$ git checkout pull/14360` \
`$ git pull https://git.openjdk.org/jdk.git pull/14360/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14360`

`$ git pr show -t 14360`

</details>

<a href="https://git.openjdk.org/jdk/pull/14360.diff">https://git.openjdk.org/jdk/pull/14360.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14360#issuecomment-1581014643)


src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java
```diff
@@ -448,11 +450,22 @@ private void startDiskMonitor() {
...
-            if (r != ignoreMe && r.getState() == RecordingState.RUNNING) {
-                r.appendChunk(chunk);
+        try {
+            chunk.finish(time);
+            for (PlatformRecording r : getRecordings()) {
+                if (r != ignoreMe && r.getState() == RecordingState.RUNNING) {
+                    r.appendChunk(chunk);
+                }
             }
+        } catch (FileNotFoundException e)  {
```
turbanoff, line 460:
```suggestion
        } catch (FileNotFoundException e) {
```

src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java
```diff
@@ -498,17 +511,24 @@ private void periodicTask() {
...
+                    if (jvm.shouldRotateDisk()) {
+                        rotateDisk();
+                    }
+                    if (isToDisk()) {
+                        EventLog.update();
+                    }
                 }
+                long minDelta = PeriodicEvents.doPeriodic();
+                wait = Math.min(minDelta, Options.getWaitInterval());
+            } catch(Throwable t) {
```
turbanoff, line 526:
```suggestion
            } catch (Throwable t) {
```

src/jdk.jfr/share/classes/jdk/jfr/internal/FilePurger.java
```diff
@@ -63,6 +63,13 @@ private static void removeOldest() {
     }
 
     private static boolean delete(SafePath p) {
+        try {
+            if (!SecuritySupport.exists(p)) {
+                return true;
+            }
+        } catch (IOException e) {
```
egahlin, line 70: Looks like there could be a race here. If the file is removed after the exists check, but before the file is deleted, the method may return false incorrectly.

jaokim: Yes, well spotted. However, does it matter? If this method return false, it means the file will still be in the list of paths to remove, and upon next purge, the SecuritySupport.exists(p) would see the file doesn't exist, and return true. 

I added it because SecuritySupport.delete() might throw an exception (via Files.delete()) if the file doesn't exists, causing non-existent files to be left in the purge list until removeOldest() removes them.

src/jdk.jfr/share/classes/jdk/jfr/internal/RepositoryChunk.java
```diff
@@ -174,4 +180,25 @@ public long getCurrentFileSize() {
...
+    }
+
+    void emitMissingChunkFileEvent() {
+        emitMissingChunkFileEvent(this.endTime);
+    }
+
+    void emitMissingChunkFileEvent(Instant endTime) {
+        String message = "Chunkfile \""+ chunkFile.toString() + "\" is missing.";
+        Logger.log(LogTag.JFR, LogLevel.ERROR, message + " Data loss might occur from " + startTime.toString() + (endTime != null ? " to " + endTime.toString() : ""));
+        ErrorThrownEvent.commit(startTime.getNano(), endTime != null ? Duration.between(startTime, endTime).toNanos() : 0L, message, MissingChunkFileError.class);
```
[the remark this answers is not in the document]
jaokim: I've skipped the MissingChunkFileError event when finding missing files during dumping (it would've only shown up in subsequent recordings anyways, so of little use). I instead throw the error when finding missing files in-flight; I think it's important to have something that hints to user that a recording might be missing data, and the error will show up in the recording that might be missing some data.

src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java
jaokim: Fixed this too. Thanks!

jaokim:
I've rewritten some parts of the solution.

- Removed the MissingChunkfileError to instead emit a jdk.DataLoss event for each missing chunkfile. The "amount" field contains the number of bytes missing, the "total" field of the event is set to 0, the event time is when the chunkfile was found missing.
- Moved the privileged IO into to transferChunksWithRetry method.

Please have a look again. Thanks!

src/hotspot/share/jfr/jni/jfrJniMethod.cpp
```diff
@@ -401,3 +401,7 @@ JVM_ENTRY_NO_ENV(jlong, jfr_host_total_memory(JNIEnv* env, jobject jvm))
   return os::physical_memory();
 #endif
 JVM_END
+
+JVM_ENTRY_NO_ENV(void, jfr_emit_data_loss(JNIEnv* env, jobject jvm, jlong bytes))
+  EventDataLoss::commit(bytes, 0L);
```
egahlin, line 406: If a value is missing, we use min_jlong to indicate not available or not applicable

src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java
```diff
@@ -498,17 +508,24 @@ private void periodicTask() {
...
+                // Catch everything and log, but don't allow it to end the periodic task
+                Logger.log(JFR_SYSTEM, ERROR, "Error in Periodic task: " + t.getClass().getName() + ", " + t.getMessage());
```
egahlin, line 525: Not sure if it matters, but you could get an OOM when string are concatenated and the thread will end.

egahlin (approved):

jaokim:
Thanks for reviewing, @turbanoff and @egahlin. Could I also request a sponsor for this? Thanks!

/integrate

PR #14361, opened 2023-06-07
8309545: Thread.interrupted from virtual thread needlessly resets interrupt status

AlanBateman:
Thread.interrupted is used to "get and clear" the current thread's interrupt status. When called from a virtual thread, the current implementation always clears the carrier's interrupt status. There is no need to do this when the interrupt status is not set, it can just read the interrupt status and return false in that case.

This was found by Sergey Kuksenko when comparing platform vs. virtual thread performance. Once CODETOOLS-7903476 is in a released version of JMH then we can start to accumulate benchmarks that execute in virtual threads.

Testing: tier1-5

---------
### Progress

### Issue
 * [JDK-8309545](https://bugs.openjdk.org/browse/JDK-8309545): Thread.interrupted from virtual thread needlessly resets interrupt status (**Bug** - P3)

### Reviewers
 * [Ron Pressler](https://openjdk.org/census#rpressler) (@pron - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14361/head:pull/14361` \
`$ git checkout pull/14361`

`$ git checkout pull/14361` \
`$ git pull https://git.openjdk.org/jdk.git pull/14361/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14361`

`$ git pr show -t 14361`

</details>

<a href="https://git.openjdk.org/jdk/pull/14361.diff">https://git.openjdk.org/jdk/pull/14361.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14361#issuecomment-1582214506)

### DIFF
--- a/src/java.base/share/classes/java/lang/VirtualThread.java
+++ b/src/java.base/share/classes/java/lang/VirtualThread.java
@@ -874,13 +874,14 @@ final class VirtualThread extends BaseVirtualThread {
     @Override
     boolean getAndClearInterrupt() {
         assert Thread.currentThread() == this;
-        synchronized (interruptLock) {
-            boolean oldValue = interrupted;
-            if (oldValue)
+        boolean oldValue = interrupted;
+        if (oldValue) {
+            synchronized (interruptLock) {
                 interrupted = false;
-            carrierThread.clearInterrupt();
-            return oldValue;
+                carrierThread.clearInterrupt();
+            }
         }
+        return oldValue;
     }
 
     @Override


src/java.base/share/classes/java/lang/VirtualThread.java
```diff
@@ -874,13 +874,14 @@ public boolean isInterrupted() {
     @Override
     boolean getAndClearInterrupt() {
         assert Thread.currentThread() == this;
-        synchronized (interruptLock) {
-            boolean oldValue = interrupted;
-            if (oldValue)
+        boolean oldValue = interrupted;
+        if (oldValue) {
+            synchronized (interruptLock) {
```
dholmes-ora, line 879: Don't you still need to read under the lock to synchronize with the interrupt method?

AlanBateman: The override of interrupt for virtual threads always sets the interrupt status as the first step. So if Thread.interrupted reads the interrupt status as true then it will synchronize. It's not too different to platform threads, I think you added a comment to Thread.interrupted about this when move the interrupted flag into Java.

pron (approved):

dholmes-ora (approved):
Changes look fine. Thanks.

jaikiran (approved):

PR #14362, opened 2023-06-07
8309532: java/lang/Class/getDeclaredField/FieldSetAccessibleTest should filter modules that depend on JVMCI

mlchung:
`FieldSetAccessibleTest` attempts to load all JDK classes and test `setAccessible` on their public fields to work properly.   It should filter the modules that directly and indirectly depend on `jdk.internal.vm.compiler` and  `jdk.internal.vm.compiler.management` as they are upgradeable and also provide APIs to add qualifed exports dynamically.

### DIFF
--- a/test/jdk/java/lang/Class/getDeclaredField/FieldSetAccessibleTest.java
+++ b/test/jdk/java/lang/Class/getDeclaredField/FieldSetAccessibleTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2014, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


test/jdk/java/lang/Class/getDeclaredField/FieldSetAccessibleTest.java
```diff
@@ -283,18 +288,64 @@ public Iterator<String> iterator() {
...
+            }
+
+            // reverse edges
+            moduleToDeps.keySet().forEach(u -> {
+                moduleToDeps.get(u)
+                            .forEach(v -> inverseDeps.get(v)
+                                                     .add(u));
+            });
+
+            Set<String> deployModules = Set.of("javafx.deploy", "jdk.deploy", "jdk.plugin", "jdk.javaws");
```
AlanBateman, line 314: Can deployModules be dropped?

mlchung: There are two other tests that should be updated.  I updated this one in this PR.  Will create another PR to drop those.

AlanBateman (approved):

test/jdk/java/lang/Class/getDeclaredField/FieldSetAccessibleTest.java
AlanBateman: Okay, that's fine.

PR #12993, opened 2023-03-12
JDK-8303950: [macos]Translucent Windows Flicker on Repaint

mickleness:
I'm confident about the new test case for this ticket, but the resolution is more invasive than I'd like. (It works, though.)

In short:
This introduces a new RenderingHint (in SunHints) to bypass the call in Window to `gg2d.fillRect(0, 0, getWidth(), getHeight());`

I left more detailed notes here about the proposed resolution:
https://github.com/openjdk/jdk/commit/1991fdac5dbf76ddaf73cc78a9f7c38370c9674c

I'm open to suggestions if anyone has a more elegant proposal to prevent the monitor from refreshing too soon?

---------
### Progress
- [ ] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8303950](https://bugs.openjdk.org/browse/JDK-8303950): [macos]Translucent Windows Flicker on Repaint

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/12993/head:pull/12993` \
`$ git checkout pull/12993`

Update a local copy of the PR: \
`$ git checkout pull/12993` \
`$ git pull https://git.openjdk.org/jdk.git pull/12993/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 12993`

View PR using the GUI difftool: \
`$ git pr show -t 12993`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jdk/pull/12993.diff">https://git.openjdk.org/jdk/pull/12993.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/12993#issuecomment-1465257059)

PR #14363, opened 2023-06-07
JDK-8303950: [macos]Translucent Windows Flicker on Repaint

mickleness:
# Problem Summary

For non-opaque windows, Window#paint calls `gg.fillRect(0, 0, getWidth(), getHeight())` before `super.paint(g)`.

This can cause flickering on Mac, and the flickering seems to have gotten much worse in recent JVMs. (See movie attachments to original ticket.)

# Discussion

This is my 2nd PR for this ticket. The original is [here](https://github.com/openjdk/jdk/pull/12993); that proposed change was IMO more invasive/scarier. It was auto-closed after 8 weeks of inactivity, and I'd rather offer this PR instead.

In that previous discussion Alan Snyder framed the core problem as a "lack of synchronization" (see [comment here](https://github.com/openjdk/jdk/pull/12993#issuecomment-1467528061)). If the monitor refreshes/flushes after we've called `fillRect` *but before we finish `super.paint`*: it makes sense that we'd see a flicker.

I agree with Alan, but I think the problem can also be framed as a mixing-Swing-with-AWT issue. (Which IMO will involve an easier fix.)

This PR is a low-risk change (relative to the previous PR) that intercepts calls to repaint a Window that is also RootPaneContainer. Now we'll redirect those calls to paint the JRootPane instead. This means we'll exclusively paint within Swing's/RepaintManager's double-buffered architecture, so we bypass the risky call to `fillRect` on the screen's Graphics2D. (And this change occurs within RepaintManager, so we're clearly already in Swing's architecture.)

So with this change: we paint everything to the double-buffer, and the *only time* we paint to the Window's Graphics2D is when have set up a AlphaComposite.Src and replace its contents with our buffer's contents.

# Tests

This PR includes a new test for 8303950 itself. This is pretty self-explanatory: we repaint a trivial animation for a few seconds and use the Robot to see if a pixel is the expected color.

This PR also includes a test called `bug8303950_legacyWindowPaintBehavior` that creates a grid of 4 windows with varying opacity/backgrounds:

image

I was surprised by how these windows rendered, but I don't think that's worth debating here. This test simply makes sure that we preserve this preexisting behavior. The broad "rules" appear to be:
1. If a JComponent identifies as opaque (see `JComponent.isOpaque`) then the JComponent's background is used. (In this case: that's the opaque green top two windows.) This is probably coming from `ComponentUI.update`
2. If a JRootPane has a translucent background, that color "wins". It doesn't composite on top of a window's translucent background: it simply *replaces* it. (See the blue-ish bottom-right window. Note it's blue -- not purple.) This is probably because the RepaintManager uses `AlphaComposite.Src` in a couple of places to replace its destination Graphics.
3. If a JRootPane has a null background color, then the Window's background color paints as expected. (See the red-ish bottom-left window.)

---------
### Progress

### Integration blocker
&nbsp;⚠️ Title mismatch between PR and JBS for issue [JDK-8303950](https://bugs.openjdk.org/browse/JDK-8303950)

### Issue
 * [JDK-8303950](https://bugs.openjdk.org/browse/JDK-8303950): [macos] Translucent Windows Flicker on Repaint (**Bug** - P3) ⚠️ Title mismatch between PR and JBS.

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14363/head:pull/14363` \
`$ git checkout pull/14363`

`$ git checkout pull/14363` \
`$ git pull https://git.openjdk.org/jdk.git pull/14363/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14363`

`$ git pr show -t 14363`

</details>

<a href="https://git.openjdk.org/jdk/pull/14363.diff">https://git.openjdk.org/jdk/pull/14363.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14363#issuecomment-1581265513)


test/jdk/javax/swing/RepaintManager/8303950/bug8303950_legacyWindowPaintBehavior.java
```diff
@@ -0,0 +1,183 @@
...
+
+                GraphicsConfiguration gc = GraphicsEnvironment.getLocalGraphicsEnvironment().getDefaultScreenDevice().getDefaultConfiguration();
+                Insets i = Toolkit.getDefaultToolkit().getScreenInsets(gc);
+                x += i.left;
+                y += i.top;
+
+                Window w1 = createWindow( WINDOW_BACKGROUND, null, x, y, 400, 400, false, "window 1");
+                Window w2 = createWindow( WINDOW_BACKGROUND, ROOTPANE_BACKGROUND, x + 400, y, 400, 400, false, "window 2");
+                Window w3 = createWindow( WINDOW_BACKGROUND, null, x, y + 400, 400, 400, true, "window 3");
+                Window w4 =  createWindow( WINDOW_BACKGROUND, ROOTPANE_BACKGROUND, x + 400, y + 400, 400, 400, true, "window 4");
```
turbanoff, line 68:
```suggestion
                Window w4 = createWindow( WINDOW_BACKGROUND, ROOTPANE_BACKGROUND, x + 400, y + 400, 400, 400, true, "window 4");
```

mickleness:
@aivanov-jdk IIRC this ticket (a P3) was initially assigned to you, and I asked to look at it instead. It looks like for now this PR is stalled, so please feel free to reassign the openJDK ticket (or make any other appropriate changes) if needed.

aivanov-jdk:
> @aivanov-jdk IIRC this ticket (a P3) was initially assigned to you, and I asked to look at it instead. It looks like for now this PR is stalled, so please feel free to reassign the openJDK ticket (or make any other appropriate changes) if needed.

@mickleness It's not assigned to me now. I can't even find that it was assigned to me.

mickleness:
OK, maybe I'm mistaken. Thanks for checking.

aivanov-jdk:
> I encountered an issue very similar to this problem, is there any chance to see progress on the fix ?

@bric3 This is not how it works… Yet feel free to propose a patch that addresses the problem. It could be based on this PR.

I looks @mickleness couldn't run the tests.

I haven't looked into the code changes or tests…

bric3:
Yes, I was more thinking about discussing the patch and PR than expecting the PR to be.erged right away, since the last comments don't really address the technical issue.

mickleness:
I'm happy to help with/discuss anything if anyone has any questions. This PR currently resolves the original issue I was seeing and it preserves the legacy painting behavior. Tested on Mac.

PR #14366, opened 2023-06-07
8309612: [REDO] JDK-8307153 JVMTI GetThreadState on carrier should return STATE_WAITING

sspitsyn:
This is REDO the fix of [JDK-8307153](https://bugs.openjdk.org/browse/JDK-8307153).
The last update of the fix in the review cycle was incorrect and incorrectly tested, so the issue has not been noticed. It is why the fix was backed out.
The issue is that the SUSPEND bit was missed in the JVMTI thread state of platform/carrier threads carrying virtual threads (see`JvmtiEnvBase::get_thread_state` function).

The first push/patch is the original fix of JDK-8307153.
The fix of the SUSPEND bit issue will be in the incremental update.
It is to simplify the review.

Testing:
 - TBD: mach5 tiers 1-5

---------
### Progress

### Issue
 * [JDK-8309612](https://bugs.openjdk.org/browse/JDK-8309612): [REDO] JDK-8307153 JVMTI GetThreadState on carrier should return STATE_WAITING (**Bug** - P3)

### Reviewers
 * [Chris Plummer](https://openjdk.org/census#cjplummer) (@plummercj - **Reviewer**) ⚠️ Review applies to [8f26e277](https://git.openjdk.org/jdk/pull/14366/files/8f26e277aa92640a581fb20bdc8acc8160c06482)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14366/head:pull/14366` \
`$ git checkout pull/14366`

`$ git checkout pull/14366` \
`$ git pull https://git.openjdk.org/jdk.git pull/14366/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14366`

`$ git pr show -t 14366`

</details>

<a href="https://git.openjdk.org/jdk/pull/14366.diff">https://git.openjdk.org/jdk/pull/14366.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14366#issuecomment-1581344269)

### DIFF
--- a/src/hotspot/share/prims/jvmtiEnvBase.cpp
+++ b/src/hotspot/share/prims/jvmtiEnvBase.cpp
@@ -730,7 +730,7 @@ JvmtiEnvBase::get_cthread_last_java_vframe(JavaThread* jt, RegisterMap* reg_map_
 }
 
 jint
-JvmtiEnvBase::get_thread_state(oop thread_oop, JavaThread* jt) {
+JvmtiEnvBase::get_thread_state_base(oop thread_oop, JavaThread* jt) {
   jint state = 0;
 
   if (thread_oop != nullptr) {
@@ -757,6 +757,19 @@ JvmtiEnvBase::get_thread_state(oop thread_oop, JavaThread* jt) {
 }
 
 jint
+JvmtiEnvBase::get_thread_state(oop thread_oop, JavaThread* jt) {
+  jint state;
+
+  if (is_thread_carrying_vthread(jt, thread_oop)) {
+    state = JVMTI_THREAD_STATE_ALIVE | JVMTI_THREAD_STATE_WAITING |
+            JVMTI_THREAD_STATE_WAITING_INDEFINITELY;
+  } else {
+    state = get_thread_state_base(thread_oop, jt);
+  }
+  return state;
+}
+
+jint
 JvmtiEnvBase::get_vthread_state(oop thread_oop, JavaThread* java_thread) {
   jint state = 0;
   bool ext_suspended = JvmtiVTSuspender::is_vthread_suspended(thread_oop);
@@ -770,7 +783,7 @@ JvmtiEnvBase::get_vthread_state(oop thread_oop, JavaThread* java_thread) {
     jint filtered_bits = JVMTI_THREAD_STATE_SUSPENDED | JVMTI_THREAD_STATE_INTERRUPTED;
 
     // This call can trigger a safepoint, so thread_oop must not be used after it.
-    state = get_thread_state(ct_oop, java_thread) & ~filtered_bits;
+    state = get_thread_state_base(ct_oop, java_thread) & ~filtered_bits;
   } else {
     jshort vt_state = java_lang_VirtualThread::state(thread_oop);
     state = (jint)java_lang_VirtualThread::map_state_to_thread_status(vt_state);
@@ -1708,13 +1721,13 @@ JvmtiEnvBase::suspend_thread(oop thread_oop, JavaThread* java_thread, bool singl
   if (java_thread->is_hidden_from_external_view()) {
     return JVMTI_ERROR_NONE;
   }
-  bool is_passive_cthread = is_passive_carrier_thread(java_thread, thread_h());
+  bool is_thread_carrying = is_thread_carrying_vthread(java_thread, thread_h());
 
   // A case of non-virtual thread.
   if (!is_virtual) {
     // Thread.suspend() is used in some tests. It sets jt->is_suspended() only.
     if (java_thread->is_carrier_thread_suspended() ||
-        (!is_passive_cthread && java_thread->is_suspended())) {
+        (!is_thread_carrying && java_thread->is_suspended())) {
       return JVMTI_ERROR_THREAD_SUSPENDED;
     }
     java_thread->set_carrier_thread_suspended();
@@ -1728,7 +1741,7 @@ JvmtiEnvBase::suspend_thread(oop thread_oop, JavaThread* java_thread, bool singl
   // An attempt to handshake-suspend a passive carrier thread will result in
   // suspension of mounted virtual thread. So, we just mark it as suspended
   // and it will be actually suspended at virtual thread unmount transition.
-  if (!is_passive_cthread) {
+  if (!is_thread_carrying) {
     assert(thread_h() != nullptr, "sanity check");
     assert(single_suspend || thread_h()->is_a(vmClasses::BaseVirtualThread_klass()),
            "SuspendAllVirtualThreads should never suspend non-virtual threads");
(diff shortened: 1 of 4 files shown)


src/hotspot/share/prims/jvmtiEnvBase.cpp
```diff
@@ -756,6 +756,17 @@ JvmtiEnvBase::get_thread_state(oop thread_oop, JavaThread* jt) {
   return state;
 }
 
+jint
+JvmtiEnvBase::get_thread_state(oop thread_oop, JavaThread* jt) {
+  jint state = get_thread_state_base(thread_oop, jt);
+
+  if (is_thread_carrying_vthread(jt, thread_oop)) {
+    state &= ~JVMTI_THREAD_STATE_RUNNABLE;
+    state |= JVMTI_THREAD_STATE_WAITING | JVMTI_THREAD_STATE_WAITING_INDEFINITELY;
```
plummercj, line 765: How about a comment here:

"Clear RUNNABLE state and add WAITING state because..."

src/hotspot/share/prims/jvmtiEnvBase.hpp
```diff
@@ -96,8 +96,8 @@ class JvmtiEnvBase : public CHeapObj<mtInternal> {
   static jvmtiError check_thread_list(jint count, const jthread* list);
   static bool is_in_thread_list(jint count, const jthread* list, oop jt_oop);
 
-  // check if thread_oop represents a passive carrier thread
-  static bool is_passive_carrier_thread(JavaThread* java_thread, oop thread_oop) {
+  // check if thread_oop represents a thread carrying virtual thread
```
plummercj, line 99:
```suggestion
  // check if thread_oop represents a thread carrying a virtual thread
```

plummercj (changes requested):

src/hotspot/share/prims/jvmtiEnvBase.cpp
alexmenkov, line 765: This does not look correct.
GetThreadState spec provides hierarchical set of questions to interpret thread state value.
JVMTI_THREAD_STATE_ALIVE | JVMTI_THREAD_STATE_WAITING | JVMTI_THREAD_STATE_WAITING_INDEFINITELY is only one branch and I'd expect all other bits are not set for this state.
Need to decide what do we want to report as carrier thread state for all possible values returned by get_thread_state_base().

sspitsyn: Thanks. Added comment.

src/hotspot/share/prims/jvmtiEnvBase.cpp
```diff
@@ -1725,10 +1736,10 @@ JvmtiEnvBase::suspend_thread(oop thread_oop, JavaThread* java_thread, bool singl
           (is_virtual && JvmtiVTSuspender::is_vthread_suspended(thread_h())),
          "sanity check");
 
-  // An attempt to handshake-suspend a passive carrier thread will result in
+  // An attempt to handshake-suspend a thread carrying virtual thread will result in
```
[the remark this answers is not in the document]
sspitsyn: Thanks. Updated now.

src/hotspot/share/prims/jvmtiEnvBase.hpp
sspitsyn: Thanks. Updated now.

src/hotspot/share/prims/jvmtiEnvBase.hpp
```diff
@@ -180,7 +180,7 @@ class JvmtiEnvBase : public CHeapObj<mtInternal> {
   static oop current_thread_obj_or_resolve_external_guard(jthread thread);
 
   // Return true if the thread identified with a pair <jt,thr_obj> is current.
-  // A passive carrier thread is not treated as current.
+  // A thread carrying virtual thread is not treated as current.
```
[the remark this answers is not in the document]
sspitsyn: Thanks. Updated now.

src/hotspot/share/prims/jvmtiEnvBase.cpp
sspitsyn: Good concern.
There are two bits (and the related RUNNABLE bit) that we care in this sub-tree of state bits: `SUSPENDED` and `INTERRUPTED`. This update clones these two bits. The RUNNABLE bit must be cleared.
A thread carrying a virtual thread can not be in native, blocked, parked, sleeping or waiting on some object.
The state returned by the `get_thread_state_base` is based on the call:
`    state = (jint)java_lang_Thread::get_thread_status(thread_oop);`
and addition of the derived from JavaThread bits: `SUSPENDED`, `INTERRUPTED` and `IN_NATIVE`.
The three bit derived from the JavaThread are not relevant.
This call has to be made directly:
`   state = (jint)java_lang_Thread::get_thread_status(thread_oop);`
The SUSPEND bit has to be based on the call:
`    java_thread->is_carrier_thread_suspended();`

plummercj (approved):

src/hotspot/share/prims/jvmtiEnvBase.cpp
alexmenkov: Do you need to check jt->is_interrupted(false) and set INTERRUPTED bit?
It looks like java_lang_Thread::get_thread_status(thread_oop) can only return RUNNABLE in the case and we clear it.

    if (is_thread_carrying_vthread(jt, thread_oop)) {
      jint state = JVMTI_THREAD_STATE_WAITING | JVMTI_THREAD_STATE_WAITING_INDEFINITELY;
      if (jt->is_carrier_thread_suspended()) {
        state |= JVMTI_THREAD_STATE_SUSPENDED;
      }
      if (jt->is_interrupted(false)) {
        state |= JVMTI_THREAD_STATE_INTERRUPTED;
      }
      return state;
    } else ...

alexmenkov: > A thread carrying a virtual thread can not be in native, blocked, parked, sleeping or waiting on some object.

Actually it can be in native.
And if I remember correctly synchronized block pins virtual thread, so inside synchronized we can get other states

sspitsyn: The INTERRUPTED bit we need has to be returned by the `java_lang_Thread::get_thread_status`.
Not completely sure but the bit jt->is_interrupted(false) can be set for the mounted virtual thread.
The JVMTI InterruptThread calls this function to set interrupt bit for non-virtual threads:
`  java_lang_Thread::set_interrupted(thread_obj, true);`

sspitsyn: Corrected the function `get_thread_state()` to make it more safe.
Only `ALIVE` and `INTERRUPTED` bits are taken from result of `java_lang_Thread::get_thread_status(thread_oop)`.

dholmes-ora: > A thread carrying a virtual thread can not be in native, blocked, parked, sleeping or waiting on some object.

A virtual thread can call native code, be blocked on an object monitor, or waiting on an object monitor. Only parking and sleeping are specialized for virtual threads in the list you gave.

sspitsyn: > > A thread carrying a virtual thread can not be in native, blocked, parked, sleeping or waiting on some object.

>A virtual thread can call native code, be blocked on an object monitor, or waiting on an object monitor. Only parking and sleeping are specialized for virtual threads in the list you gave.

This statement was about carrier thread when there is a virtual thread executed at the top.
We are getting state bits with the `java_lang_Thread::get_thread_status(thread_oop)` where the `thread_oop` belongs to the carrier thread. But you are talking about a virtual thread which, of course, can be in almost any state.

dholmes-ora: Thanks for clarifying - it gets very confusing as to which "thread" is being talked about. But if a virtual thread is mounted on this JavaThread then I thought the carrier thread's thread-oop is supposed to be in a blocked state?

sspitsyn: It was decided with Alan that it is okay to be waiting state.
Alan's comment in the original PR [https://github.com/openjdk/jdk/pull/14298](https://github.com/openjdk/jdk/pull/14298) was:
>  if the jt is carrying thread_oop and it's okay for the JVMTI state to reported as WAITING when waiting for something other than Object.wait.

AlanBateman: The mental model  is that the carrier is blocked so this is what an observer using the APIs should see. My recollection is that JVMTI_THREAD_STATE_WAITING was okay because there is a wriggle room in the JVM TI spec, it only uses Object.wait as an example. There may be a few rough edges to smooth down in this area. It's okay to take time with this PR and expand the tests to cover more cases and get more confident that there aren't more issues.

alexmenkov (approved):

src/hotspot/share/prims/jvmtiEnvBase.cpp
sspitsyn: We agreed with Alex to file a test RFE to improve test coverage in this area.

sspitsyn:
Chris and Alex, thank you for review!

PR #14368, opened 2023-06-07
JDK-8305645: System Tray icons get corrupted when Windows primary monitor changes

honkar-jdk:
There are two scenarios related to tray icon distortion.

1. Single Screen - when DPI / Scale is changed on the fly - this was resolved as part of [PR#8441](https://github.com/openjdk/jdk/pull/8441)

2. Multi Screen - when screens are set to different scales and the primary display is toggled. This is a variation of scenario one. 

Earlier Windows msg - [WM_DPICHANGED](https://learn.microsoft.com/en-us/windows/win32/hidpi/wm-dpichanged) was being used to update the tray icons. This message is sent when window DPI changes. WM_DPICHANGED msg is not received when taskbar switches to primary display under multiple screen scenario, hence the tray icon was still seen distorted in case 2.
 
As [WM_DISPLAYCHANGE](https://learn.microsoft.com/en-us/windows/win32/gdi/wm-displaychange) is received in both cases, it is used as the new message to trigger tray icon update. This message also works when the main display monitor gets disconnected accidently making the other display the primary display.

TrayIconScalingTest's  instructions have been updated to include both the cases described above.

### DIFF
--- a/src/java.desktop/windows/native/libawt/windows/awt_TrayIcon.cpp
+++ b/src/java.desktop/windows/native/libawt/windows/awt_TrayIcon.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2005, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2005, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
--- a/src/java.desktop/windows/native/libawt/windows/awt_TrayIcon.cpp
+++ b/src/java.desktop/windows/native/libawt/windows/awt_TrayIcon.cpp
@@ -262,7 +262,7 @@ LRESULT CALLBACK AwtTrayIcon::TrayWindowProc(HWND hwnd, UINT uMsg, WPARAM wParam
                 }
             }
             break;
-        case WM_DPICHANGED:
+        case WM_DISPLAYCHANGE:
             // Set the flag to update icon images, see WmTaskbarCreated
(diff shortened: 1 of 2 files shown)


mrserb:
The old fix did not add any scale checks since the updateNativeImage was called on dpi change only. If we will start to call updateNativeImage on each monitor change event then probably we should recreate images only if the scale was changed?

honkar-jdk:
@mrserb 

> The old fix did not add any scale checks since the updateNativeImage was called on dpi change only. If we will start to call updateNativeImage on each monitor change event then probably we should recreate images only if the scale was changed?

The update for any tray icons happens within `AwtTrayIcon::WmTaskbarCreated()` and for only the 'WM_DISPLAYCHANGE' events forwarded to `AwtTrayIcon::TrayWindowProc` . In case 2, we are ideally switching between 2 different scales because the 2 monitors are at different scales and the taskbar is added to the primary monitor which is at a different scale from the previous one.

I checked the case where both monitors are at same scale and the the primary display is toggled btw the two, WM_DISPLAYCHANGE is received but since WmTaskbarCreated() is not called in this scenario and as a result updateNativeImage() is not called as well.

mrserb:
Then it looks fine, thank you for clarification.

mrserb (approved):

aivanov-jdk:
> I checked the case where both monitors are at same scale and the the primary display is toggled btw the two, WM_DISPLAYCHANGE is received but WmTaskbarCreated() is not called in this scenario and as a result updateNativeImage() is not called as well.

So, we set the flag that the icon needs updating… but because the scale remains unchanged the Taskbar doesn't broadcast its `"TaskbarCreated"` message, therefore the icons aren't updated.

Sounds reasonable.

aivanov-jdk (approved):

PR #14371, opened 2023-06-07
8305104: Remove the old core reflection implementation

mlchung:
JEP 416 integrated in JDK 18 and since then, only a couple minor issues has been reported.  Those issues were related with exception being thrown with invalid arguments.   We propose to remove the old core reflection implementation in JDK 22.  The `-Djdk.reflect.useDirectMethodHandle=false` workaround to revert to the old implementation will stop to work.

---------
### Progress
- [x] Change requires CSR request [JDK-8309635](https://bugs.openjdk.org/browse/JDK-8309635) to be approved

### Issues
 * [JDK-8305104](https://bugs.openjdk.org/browse/JDK-8305104): Remove the old core reflection implementation (**Enhancement** - P4)
 * [JDK-8309635](https://bugs.openjdk.org/browse/JDK-8309635): Remove the old core reflection implementation (**CSR**)

### Reviewers
 * [David Holmes](https://openjdk.org/census#dholmes) (@dholmes-ora - **Reviewer**) ⚠️ Review applies to [d161a384](https://git.openjdk.org/jdk/pull/14371/files/d161a38443971f289d4c859e1f7b12f2ef4d0f99)
 * [Alan Bateman](https://openjdk.org/census#alanb) (@AlanBateman - **Reviewer**) ⚠️ Review applies to [fc164c71](https://git.openjdk.org/jdk/pull/14371/files/fc164c719d764008d690095f430703d29e4e0b5a)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14371/head:pull/14371` \
`$ git checkout pull/14371`

`$ git checkout pull/14371` \
`$ git pull https://git.openjdk.org/jdk.git pull/14371/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14371`

`$ git pr show -t 14371`

</details>

<a href="https://git.openjdk.org/jdk/pull/14371.diff">https://git.openjdk.org/jdk/pull/14371.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14371#issuecomment-1581544162)


test/jdk/java/lang/reflect/Field/NegativeTest.java
```diff
@@ -24,8 +24,7 @@
 /**
  * @test
  * @bug 8277451
- * @run testng/othervm -Djdk.reflect.useDirectMethodHandle=true NegativeTest
- * @run testng/othervm -Djdk.reflect.useDirectMethodHandle=false NegativeTest
+ * @run testng/othervm NegativeTest
```
liach, line 27: Does this still need othervm if it doesn't set system properties?

src/hotspot/share/classfile/verifier.cpp
```diff
@@ -290,13 +291,12 @@ bool Verifier::is_eligible_for_verification(InstanceKlass* klass, bool should_ve
     !(klass->is_shared() && klass->is_rewritten()) &&
 
     // As of the fix for 4486457 we disable verification for all of the
-    // dynamically-generated bytecodes associated with the 1.4
-    // reflection implementation, not just those associated with
+    // dynamically-generated bytecodes associated with
     // jdk/internal/reflect/SerializationConstructorAccessor.
     // NOTE: this is called too early in the bootstrapping process to be
     // guarded by Universe::is_gte_jdk14x_version().
     // Also for lambda generated code, gte jdk8
```
dholmes-ora, line 298: While you are here could you delete these version comments please - they are meaningless these days. Thanks.

dholmes-ora (approved):
Hotspot code and test changes look fine. Thanks.

src/java.base/share/classes/jdk/internal/reflect/ReflectionFactory.java
```diff
@@ -638,10 +575,7 @@ private static boolean disableSerialConstructorChecks() {
     // for the first few invocations of Methods and Constructors and
     // then switch to the bytecode-based implementations.
 
-    private static final Config DEFAULT_CONFIG = new Config(false, // noInflation
-                                                            15, // inflationThreshold
-                                                            ALL_MH_ACCESSORS, // useDirectMethodHandle
-                                                            false, // useNativeAccessorOnly
+    private static final Config DEFAULT_CONFIG = new Config(false, // useNativeAccessorOnly
```
AlanBateman, line 578: The block comment just before this will need updating.  There's another one in Config.config where it uses the default/native implementation in early startup.

AlanBateman (approved):
The  old + new implementations have been in four releases, which is good for shaking out any regressions/issues. Removing the old implementation in JDK 22 is good.

test/jdk/java/lang/reflect/Field/NegativeTest.java
mlchung: Updated.

src/hotspot/share/classfile/verifier.cpp
mlchung: Done.

AlanBateman (approved):

src/java.base/share/classes/jdk/internal/reflect/MethodHandleAccessorFactory.java
```diff
@@ -303,6 +303,17 @@ static void ensureClassInitialized(Class<?> defc) {
...
      * Returns true if NativeAccessor should be used.
+     *
+     * Native accessor, i.e. VM reflection implementation, is used if one of
+     * the following conditions is met:
+     * 1. during VM early startup and method handle support is fully initialized
+     * 2. a Java native method
+     * 3. -Djdk.reflect.useNativeAccessorOnly=true is set
+     * 4. the member takes a variable number of arguments and the last parameter
+     *    is an array (see details below)
+     * 5. the member's method type has an arity >= 255
```
ExE-Boss, line 314:
```suggestion
     * 1. during VM early startup before method handle support is fully initialized
     * 2. a Java native method
     * 3. -Djdk.reflect.useNativeAccessorOnly=true is set
     * 4. the member takes a variable number of arguments and the last parameter
     *    is not an array (see details below)
     * 5. the member's method type has an arity >= 255
```

mlchung: Fixed.  Thanks.

PR #14373, opened 2023-06-08
8309583: AArch64: Optimize firstTrue() when amount of elements < 8

changpeng1997:
This patch optimizes VectorMask.firstTrue() on Neon when there are 2 or 4 elements in vector registers.

VectorMask.firstTrue() should return VLEGNTH when vector mask is all false [1]. Current implementation uses rbit and then clz [2] to count leading zeros, then uses csel [3] (conditional select) to get the smaller value between VLENGTH and the number of unset lanes to ensure correctness.

This patch sets the 16th or 32nd bit as 1, when there are only 2 or 4 elements in boolean masks, before rbit and clz. With this trick, maximum value calculated in such case will be VLENGTH (2 or 4).

Test:
All vector and vectorapi test passed.

Performance:
The benchmark functions are in MaskQueryOperationsBenchmark.java [4]. This patch also modifies above benchmark to measure mask operations' performance more effectively.

Following data is collected on a 128-bit Neon machine.

Benchmark                                                                        (inputs)   Mode  Before    After    Units
MaskQueryOperationsBenchmark.testFirstTrueInt            1            thrpt  5952.670  7298.491 ops/ms
MaskQueryOperationsBenchmark.testFirstTrueInt            2            thrpt  5951.513  7297.620 ops/ms
MaskQueryOperationsBenchmark.testFirstTrueInt            3            thrpt  5953.048  7298.072 ops/ms
MaskQueryOperationsBenchmark.testFirstTrueLong        1            thrpt  3496.990  4003.188 ops/ms
MaskQueryOperationsBenchmark.testFirstTrueLong        2            thrpt  3497.755  4002.577 ops/ms
MaskQueryOperationsBenchmark.testFirstTrueLong        3            thrpt  3500.085  4002.471 ops/ms

[1]: https://docs.oracle.com/en/java/javase/20/docs/api/jdk.incubator.vector/jdk/incubator/vector/VectorMask.html#firstTrue()
[2]: https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/aarch64/aarch64_vector.ad#L5540
[3]: https://developer.arm.com/documentation/ddi0602/2021-12/Base-Instructions/CSEL--Conditional-Select-

---------
### Progress

### Issue
 * [JDK-8309583](https://bugs.openjdk.org/browse/JDK-8309583): AArch64: Optimize firstTrue() when amount of elements &lt; 8 (**Enhancement** - P4)

### Reviewers
 * [Andrew Haley](https://openjdk.org/census#aph) (@theRealAph - **Reviewer**) ⚠️ Review applies to [62a6522c](https://git.openjdk.org/jdk/pull/14373/files/62a6522cc61deafa6ea363e7523875e3aea7c796)
 * [Eric Liu](https://openjdk.org/census#eliu) (@e1iu - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14373/head:pull/14373` \
`$ git checkout pull/14373`

`$ git checkout pull/14373` \
`$ git pull https://git.openjdk.org/jdk.git pull/14373/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14373`

`$ git pr show -t 14373`

</details>

<a href="https://git.openjdk.org/jdk/pull/14373.diff">https://git.openjdk.org/jdk/pull/14373.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14373#issuecomment-1581809214)


theRealAph:
Where is the benchmark? You don't seem to have included it in this PR.

changpeng1997:
> Where is the benchmark? You don't seem to have included it in this PR.

@theRealAph

Sorry for the delay. Original performance was measured by a simple benchmark only measuring firstTrue()'s performance written by myself. When I wanted to add it to JDK I found an existing benchmark used to measure different mask operations' performance ([jdk/test/micro/org/openjdk/bench/jdk/incubator/vector/MaskQueryOperationsBenchmark.java at master · openjdk/jdk · GitHub](https://github.com/openjdk/jdk/blob/master/test/micro/org/openjdk/bench/jdk/incubator/vector/MaskQueryOperationsBenchmark.java)).  I tried to measure firstTrue()'s performance by this benchmark, but I found Blackhole‘s proportion of hottest region is too high, like following:

 

```

[Hottest Methods (after inlining)].............................................................. 57.15% c2, level 4 org.openjdk.jmh.infra.Blackhole::consumeFull, version 867 41.99% c2, level 4 org.openjdk.bench.jdk.incubator.vector.jmh_generated.MaskQueryOperationsBenchmark_testFirstTrueLong_jmhTest::testFirstTrueLong_thrpt_jmhStub, version 883.

```

 

So I spent some time on fixing this benchmark to measure mask operations' performance effectively. After this update, the proportion of blackhole is below 10% for each benchmark function. And I also updated the performance of firstTrue() measured by this benchmark when there are only 2 or 4 elements in boolean masks.

theRealAph:
Something is wrong with your setup.
You should be seeing this:
# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)

not this:
# Blackhole mode: full + dont-inline hint (default, use -Djmh.blackhole.autoDetect=true to auto-detect)

shipilev:
> Output before this patch: https://gist.github.com/changpeng1997/734aa176577bfff56f5a87db9c8db69a 
> Output after this patch: https://gist.github.com/changpeng1997/73098069b8f814310d6606dfd7dc56c5

Blackhole mode autodetection was added in JMH 1.33, and enabled in JMH 1.34. The logs above say they run with JMH 1.33. Current version is 1.36, you need to upgrade, @changpeng1997. 

Also, I notice that your before/after logs use different JVM modes, one uses `release`, and another uses `fastdebug`. These are not comparable.

theRealAph:
So I'm looking at the results of the patch and I see:

Before
```
Benchmark                                           (inputs)  Mode  Cnt   Score   Error  Units
MaskQueryOperationsBenchmark.testFirstTrueInt              1  avgt    3  69.547 ± 2.837  ns/op
MaskQueryOperationsBenchmark.testFirstTrueInt              2  avgt    3  69.549 ± 0.497  ns/op
MaskQueryOperationsBenchmark.testFirstTrueInt              3  avgt    3  69.506 ± 1.360  ns/op
```

After:
```
Benchmark                                           (inputs)  Mode  Cnt   Score   Error  Units
MaskQueryOperationsBenchmark.testFirstTrueInt              1  avgt    3  58.955 ± 0.838  ns/op
MaskQueryOperationsBenchmark.testFirstTrueInt              2  avgt    3  58.690 ± 2.940  ns/op
MaskQueryOperationsBenchmark.testFirstTrueInt              3  avgt    3  58.923 ± 1.088  ns/op

```

which corresponds with a change from

```
            0x00000001158ef748:   fmov	x11, d16
            0x00000001158ef74c:   rbit	x11, x11
            0x00000001158ef750:   clz	x11, x11
            0x00000001158ef754:   lsr	w11, w11, #3
           ;; 0x4
            0x00000001158ef758:   orr	w8, wzr, #0x4
            0x00000001158ef75c:   cmp	w11, w8
            0x00000001158ef760:   csel	w11, w8, w11, ge  // ge = tcont
``` 
to

```
            0x0000000115f3f8e8:   fmov	x14, d16
            0x0000000115f3f8ec:   orr	x14, x14, #0x100000000
            0x0000000115f3f8f0:   rbit	x14, x14
            0x0000000115f3f8f4:   clz	x14, x14
            0x0000000115f3f8f8:   lsr	w14, w14, #3
```

That's a pretty decent speedup when you consider that the benchmark is dominated by memory operations and vector->core register moves.

theRealAph:
If we care about memory ops, note that we can get a useful speedup with `match(Set dst (VectorMaskFirstTrue (LoadVector mem)))` but perhaps that's not worth doing.

```
Benchmark                                           (inputs)  Mode  Cnt   Score   Error  Units
MaskQueryOperationsBenchmark.testFirstTrueInt              1  avgt    3  49.591 ± 0.477  ns/op

```

I will say that in general if you have to work in the core integer processor on an in-memory vector , it might be worth loading straight into core registers rather than going via the SIMD regs. Maybe we should write a general-purpose function that bypasses the SIMD unit in all such cases.

theRealAph (approved):
OK.

changpeng1997:
> Something is wrong with your setup. You should be seeing this: `# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)`
> 
> not this: `# Blackhole mode: full + dont-inline hint (default, use -Djmh.blackhole.autoDetect=true to auto-detect)`

@theRealAph Sorry for the delay, I was on holiday last week.

I found that we can avoid the effects of blackhole by using ``` -Djmh.blackhole.autoDetect=true ```, so I have reset this benchmark.

Following is the performance of ``` testFirstTrueInt ``` and ``` testFirstTrueLong ``` before and after this patch:

```
Benchmark                                       bits    (inputs)  Mode  Before      After      Units
MaskQueryOperationsBenchmark.testFirstTrueInt   128           1  thrpt  520650.354  580091.081 ops/ms
MaskQueryOperationsBenchmark.testFirstTrueInt   128           2  thrpt  520677.937  580391.661 ops/ms
MaskQueryOperationsBenchmark.testFirstTrueInt   128           3  thrpt  519967.269  580705.642 ops/ms
MaskQueryOperationsBenchmark.testFirstTrueLong  128           1  thrpt  518563.126  575941.490 ops/ms
MaskQueryOperationsBenchmark.testFirstTrueLong  128           2  thrpt  517329.190  578848.383 ops/ms
MaskQueryOperationsBenchmark.testFirstTrueLong  128           3  thrpt  517987.339  577601.752 ops/ms
```

And following are the corresponding JMH output:
before my patch: https://gist.github.com/changpeng1997/3ebe4b7beea93716d9f29d4ef71641af
after my patch: https://gist.github.com/changpeng1997/d306957370eb0bdbb8e71b601440cdaa

We can see the C2 code of firstTrue().

changpeng1997:
> 

Thanks! Sorry for this mistake.

changpeng1997:
> > Output before this patch: https://gist.github.com/changpeng1997/734aa176577bfff56f5a87db9c8db69a
> > Output after this patch: https://gist.github.com/changpeng1997/73098069b8f814310d6606dfd7dc56c5
> 
> Blackhole mode autodetection was added in JMH 1.33, and enabled in JMH 1.34. The logs above say they run with JMH 1.33. Current version is 1.36, you need to upgrade, @changpeng1997.
> 
> Also, I notice that your before/after logs use different JVM modes, one uses `release`, and another uses `fastdebug`. These are not comparable.

@shipilev Thanks! Sorry for this mistake.

e1iu (approved):

PR #14375, opened 2023-06-08
8308869: C2: use profile data in subtype checks when profile has more than one class

rwestrel:
In this simple micro benchmark:

https://github.com/franz1981/java-puzzles/blob/main/src/main/java/red/hat/puzzles/polymorphism/RequireNonNullCheckcastScalability.java#L70

Performance drops sharply with polluted profile:

```
Benchmark                                         (typePollution)   Mode  Cnt     Score    Error   Units
RequireNonNullCheckcastScalability.isDuplicated1            false  thrpt   10  1453.372 ± 24.919  ops/us
```

to:

```
Benchmark                                         (typePollution)   Mode  Cnt   Score   Error   Units
RequireNonNullCheckcastScalability.isDuplicated1             true  thrpt   10  28.579 ± 2.280  ops/us
```

The test has 2 type checks to 2 different interfaces so caching with
`secondary_super_cache` doesn't help.

The micro-benchmark only uses 2 different concrete classes
(`DuplicatedContext` and `NonDuplicatedContext`) and they are recorded
in profile data at the type checks. But c2 only take advantage of
profile data at type checks if they report a single class.

What I propose is that the full blown type check expanded in
`Phase::gen_subtype_check()` takes advantage of profile data. So in
the case of the micro benchmark, before checking the
`secondary_super_cache`, generated code checks whether the object
being type checked is a `DuplicatedContext` or a
`NonDuplicatedContext`.

This works fairly well on this micro benchmark:

```
Benchmark                                         (typePollution)   Mode  Cnt    Score    Error   Units
RequireNonNullCheckcastScalability.isDuplicated1             true  thrpt   10  871.224 ± 20.750  ops/us
```

It also scales much better if there are multiple threads running the
same test (`secondary_super_cache` doesn't scale well: see
JDK-8180450).

Now if the micro-benchmark is changed according to the comment:

https://github.com/franz1981/java-puzzles/blob/d2d60af3d0dfe7a2567807395138edcb1d1c24f5/src/main/java/red/hat/puzzles/polymorphism/RequireNonNullCheckcastScalability.java#L62

so the type check hits in the `secondary_super_cache`, the current
code performs much better:

```
Benchmark                                         (typePollution)   Mode  Cnt    Score    Error   Units
RequireNonNullCheckcastScalability.isDuplicated1             true  thrpt   10  871.224 ± 20.750  ops/us
```

but leveraging profiling as explained above performs even better:

```
Benchmark                                         (typePollution)   Mode  Cnt     Score    Error   Units
RequireNonNullCheckcastScalability.isDuplicated1             true  thrpt   10  1165.474 ± 70.171  ops/us
```

I think it's actually likely that there's a performance advantage even
if profiling sees more than 2 types at a type check unless the profile
is heavily polluted. The problem is that, the way current profile data
is collected, we can't tell if the profile is heavily polluted
because, unlike profiling at virtual calls, there's no counter for non
recorded types. The `count` field is used to count failed type checks
instead. JVMCI added a `nonprofiled_count`. I thought about using that
one but it seems after looking at the way c2 uses the failed type
check count that it would be simpler to simply collect profile data at
type checks the way it's done at virtual calls. Indeed, C2 uses the
unique class reported by profile data only if there was no failed type
checks recorded in profile data but:

- at checkcasts, it also checks that it can prove the check would
  statically fold. That last check seems to be the one that matters.
  
- at instanceof, AFAICT, a profiled type that causes the instanceof to
  fail is as valuable as one that makes it succeed so it would be
  better to ignore failures reported by profiling.
  
I also discussed this briefly with Tom and he said graal doesn't need
the failed type check count.

So, in the patch I propose, I changed the way profile data is
collected so it works the same it does at virtual call. If this patch
is accepted, I'll need help with platforms other than x86 and aarch64.

I also modified the JVMCI code. BTW, I also wonder if
`VirtualCallData.getMethodProfile()` is not obsolete.

Finally, I changed `Phase::gen_subtype_check()` so it emits the extra
checks. That method is now called at macro expansion when profile data
is not longer available. So I attached profile data to the
`SubTypeCheck` node. For each profile data entry, 2 edges are added:
one for the klass, one for the profile frequency.

Because `SubTypeCheck` now has extra edges, it can happen that 2
`SubTypeCheck` nodes that perform the same subtype check don't common
during `IGVN` which can get in the way of some optimizations. I had to
make some adjustments to the logic of split if and code that looks for
dominating identical checks because of that.

/cc hotspot-compiler

---------
### Progress

### Issue
 * [JDK-8308869](https://bugs.openjdk.org/browse/JDK-8308869): C2: use profile data in subtype checks when profile has more than one class (**Enhancement** - P4)

### Reviewers
 * [Vladimir Ivanov](https://openjdk.org/census#vlivanov) (@iwanowww - **Reviewer**) ⚠️ Review applies to [4072e7ea](https://git.openjdk.org/jdk/pull/14375/files/4072e7ea544f00dd1b84dca02c8568a85ac9c345)
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - **Reviewer**) ⚠️ Review applies to [d6e9fa0f](https://git.openjdk.org/jdk/pull/14375/files/d6e9fa0f137e6007e78b0e7a3aef85052c9eaf38)

### Contributors
 * Fei Yang `<fyang@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14375/head:pull/14375` \
`$ git checkout pull/14375`

`$ git checkout pull/14375` \
`$ git pull https://git.openjdk.org/jdk.git pull/14375/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14375`

`$ git pr show -t 14375`

</details>

<a href="https://git.openjdk.org/jdk/pull/14375.diff">https://git.openjdk.org/jdk/pull/14375.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14375#issuecomment-1582419808)


src/jdk.internal.vm.ci/share/classes/jdk/vm/ci/hotspot/HotSpotMethodData.java
```diff
@@ -571,9 +566,8 @@ public int getExecutionCount(HotSpotMethodData data, int position) {
             return -1;
         }
 
-        @Override
         protected long getTypesNotRecordedExecutionCount(HotSpotMethodData data, int position) {
```
tkrodriguez, line 569: I think you can inline this method into the callers as it only existed to distinguish the count and nonprofiled_count

src/hotspot/share/oops/methodData.hpp
```diff
@@ -1088,21 +1088,8 @@ class ReceiverTypeData : public CounterData {
   friend class JVMCIVMStructs;
 protected:
   enum {
-#if INCLUDE_JVMCI
```
tkrodriguez, line 1091: I think you should add a comment to ReceiverTypeData explaining when count is incremented

tkrodriguez:
The updates look good.

TobiHartmann:
Impressive work, Roland!

I did not review the code in detail yet but here are some failures from preliminary testing:

`applications/ctw/modules/java_base_2.java` and some other CTW tests fails on AArch64:

```
#  Internal Error (/open/src/hotspot/cpu/aarch64/assembler_aarch64.hpp:548), pid=60565, tid=30979
#  assert(offset_ok_for_immed(offset(), size)) failed: must be, was: 32776, 3

Current CompileTask:
C1: 330461 92946    b  3       sun.security.tools.keytool.Resources::<clinit> (5678 bytes)

Stack: [0x000000016cc14000,0x000000016ce17000],  sp=0x000000016ce13ed0,  free space=2047k
V  [libjvm.dylib+0x1385054]  VMError::report_and_die(int, char const*, char const*, char*, Thread*, unsigned char*, void*, void*, char const*, int, unsigned long)+0x4bc  (assembler_aarch64.hpp:548)
V  [libjvm.dylib+0x13859f0]  VMError::report_and_die(Thread*, void*, char const*, int, char const*, char const*, char*)+0x40
V  [libjvm.dylib+0x6a9518]  report_vm_error(char const*, int, char const*, char const*, ...)+0x6c
V  [libjvm.dylib+0x112b08]  Address::encode(Instruction_aarch64*) const+0x230
V  [libjvm.dylib+0x1127d4]  Assembler::ld_st2(Register, Address const&, int, int, int)+0x258
V  [libjvm.dylib+0x42634c]  LIR_Assembler::emit_opTypeCheck(LIR_OpTypeCheck*)+0x6f0
V  [libjvm.dylib+0x412f44]  LIR_OpTypeCheck::emit_code(LIR_Assembler*)+0x20
V  [libjvm.dylib+0x41b7fc]  LIR_Assembler::emit_lir_list(LIR_List*)+0x1f4
V  [libjvm.dylib+0x41bae8]  LIR_Assembler::emit_block(BlockBegin*)+0x154
V  [libjvm.dylib+0x41b914]  LIR_Assembler::emit_code(BlockList*)+0xa4
V  [libjvm.dylib+0x3d0708]  Compilation::emit_code_body()+0x150
V  [libjvm.dylib+0x3d1234]  Compilation::compile_java_method()+0x39c
V  [libjvm.dylib+0x3d15a4]  Compilation::compile_method()+0x128
V  [libjvm.dylib+0x3d1b18]  Compilation::Compilation(AbstractCompiler*, ciEnv*, ciMethod*, int, BufferBlob*, bool, DirectiveSet*)+0x1e8
V  [libjvm.dylib+0x3d45d0]  Compiler::compile_method(ciEnv*, ciMethod*, int, bool, DirectiveSet*)+0x254
V  [libjvm.dylib+0x620750]  CompileBroker::invoke_compiler_on_method(CompileTask*)+0xa0c
V  [libjvm.dylib+0x61fa84]  CompileBroker::compiler_thread_loop()+0x3c8
V  [libjvm.dylib+0xa26908]  JavaThread::thread_main_inner()+0x334
V  [libjvm.dylib+0x12b0764]  Thread::call_run()+0x134
V  [libjvm.dylib+0x102317c]  thread_native_entry(Thread*)+0x160
C  [libsystem_pthread.dylib+0x706c]  _pthread_start+0x94

```

`compiler/jvmci/compilerToVM/ReprofileTest.java` fails with:

```
java.lang.AssertionError: 56 != 48
	at jdk.internal.vm.ci/jdk.vm.ci.hotspot.HotSpotMethodDataAccessor.getSize(HotSpotMethodDataAccessor.java:82)
	at jdk.internal.vm.ci/jdk.vm.ci.hotspot.HotSpotProfilingInfo.findBCI(HotSpotProfilingInfo.java:168)
	at jdk.internal.vm.ci/jdk.vm.ci.hotspot.HotSpotProfilingInfo.getExecutionCount(HotSpotProfilingInfo.java:138)
	at jdk.internal.vm.ci/jdk.vm.ci.meta.ProfilingInfo.toString(ProfilingInfo.java:149)
	at jdk.internal.vm.ci/jdk.vm.ci.hotspot.HotSpotProfilingInfo.toString(HotSpotProfilingInfo.java:202)
	at compiler.jvmci.compilerToVM.ReprofileTest.runSanityTest(ReprofileTest.java:101)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1593)
	at compiler.jvmci.compilerToVM.ReprofileTest.main(ReprofileTest.java:64)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at com.sun.javatest.regtest.agent.MainWrapper$MainTask.run(MainWrapper.java:138)
```

`serviceability/sa/ClhsdbCDSCore.java` fails with:

```
#  SIGSEGV (0xb) at pc=0x0000000103b40894, pid=6427, tid=9987
#
# JRE version: Java(TM) SE Runtime Environment (22.0) (fastdebug build 22-internal-2023-06-14-1110599.tobias.hartmann.jdk2)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (fastdebug 22-internal-2023-06-14-1110599.tobias.hartmann.jdk2, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, bsd-aarch64)
# Problematic frame:
# V  [libjvm.dylib+0x1308894]  Unsafe_PutInt(JNIEnv_*, _jobject*, _jobject*, long, int)+0x170

Stack: [0x000000016f7f8000,0x000000016f9fb000],  sp=0x000000016f9fa8c0,  free space=2058k
V  [libjvm.dylib+0x1308894]  Unsafe_PutInt(JNIEnv_*, _jobject*, _jobject*, long, int)+0x170
j  jdk.internal.misc.Unsafe.putInt(Ljava/lang/Object;JI)V+0 java.base@22-internal
j  jdk.internal.misc.Unsafe.putInt(JI)V+4 java.base@22-internal
j  CrashApp.main([Ljava/lang/String;)V+5
v  ~StubRoutines::call_stub 0x0000000113e5417c
V  [libjvm.dylib+0x9f275c]  JavaCalls::call_helper(JavaValue*, methodHandle const&, JavaCallArguments*, JavaThread*)+0x648
V  [libjvm.dylib+0xae32f8]  jni_invoke_static(JNIEnv_*, JavaValue*, _jobject*, JNICallType, _jmethodID*, JNI_ArgumentPusher*, JavaThread*)+0x25c
V  [libjvm.dylib+0xae9c7c]  jni_CallStaticVoidMethod+0x248
C  [libjli.dylib+0xb2bc]  JavaMain+0xd60
C  [libjli.dylib+0xd4c4]  ThreadJavaMain+0xc
C  [libsystem_pthread.dylib+0x726c]  _pthread_start+0x94
Java frames: (J=compiled Java code, j=interpreted, Vv=VM code)
j  jdk.internal.misc.Unsafe.putInt(Ljava/lang/Object;JI)V+0 java.base@22-internal
j  jdk.internal.misc.Unsafe.putInt(JI)V+4 java.base@22-internal
j  CrashApp.main([Ljava/lang/String;)V+5
v  ~StubRoutines::call_stub 0x0000000113e5417c
```

`serviceability/sa/TestPrintMdo.java` fails with:

```
 stderr: [Exception in thread "main" java.lang.InternalError: 72 144 0
	at jdk.hotspot.agent/sun.jvm.hotspot.oops.MethodData.dataAt(MethodData.java:282)
	at jdk.hotspot.agent/sun.jvm.hotspot.oops.MethodData.nextData(MethodData.java:318)
	at jdk.hotspot.agent/sun.jvm.hotspot.oops.MethodData.printDataOn(MethodData.java:359)
	at jdk.hotspot.agent/sun.jvm.hotspot.CommandProcessor$19$1.visit(CommandProcessor.java:938)
	at jdk.hotspot.agent/sun.jvm.hotspot.classfile.ClassLoaderData.classesDo(ClassLoaderData.java:107)
	at jdk.hotspot.agent/sun.jvm.hotspot.classfile.ClassLoaderDataGraph.classesDo(ClassLoaderDataGraph.java:84)
	at jdk.hotspot.agent/sun.jvm.hotspot.CommandProcessor$19.doit(CommandProcessor.java:926)
	at jdk.hotspot.agent/sun.jvm.hotspot.CommandProcessor.executeCommand(CommandProcessor.java:2212)
	at jdk.hotspot.agent/sun.jvm.hotspot.CommandProcessor.executeCommand(CommandProcessor.java:2182)
	at jdk.hotspot.agent/sun.jvm.hotspot.CommandProcessor.run(CommandProcessor.java:2053)
	at jdk.hotspot.agent/sun.jvm.hotspot.CLHSDB.run(CLHSDB.java:112)
	at jdk.hotspot.agent/sun.jvm.hotspot.CLHSDB.main(CLHSDB.java:44)
	at jdk.hotspot.agent/sun.jvm.hotspot.SALauncher.runCLHSDB(SALauncher.java:281)
	at jdk.hotspot.agent/sun.jvm.hotspot.SALauncher.main(SALauncher.java:500)
```

`testlibrary_tests/ir_framework/tests/TestIRMatching.java` fails with:

```
1) Method "public boolean ir_framework.tests.CheckCastArray.array()" - [Failed IR rules: 2]:
   * @IR rule 1: "@compiler.lib.ir_framework.IR(phase={DEFAULT}, applyIfCPUFeatureAnd={}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={}, applyIfAnd={}, failOn={"_#CHECKCAST_ARRAY#_"}, applyIfOr={}, applyIfNot={})"
     > Phase "PrintOptoAssembly":
       - failOn: Graph contains forbidden nodes:
         * Constraint 1: "(((?i:cmp|CLFI|CLR).*precise \\[.*:|.*(?i:mov|or).*precise \\[.*:.*\\R.*(cmp|CMP|CLR)))"
           - Matched forbidden node:
             * cmpl    R10, narrowklass: precise [java/lang/Object: 0x00007fcab601f3e8 * (java/lang/Cloneable,java/io/Serializable): :Constant:
   * @IR rule 3: "@compiler.lib.ir_framework.IR(phase={DEFAULT}, applyIfCPUFeatureAnd={}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={}, applyIfAnd={}, failOn={"_#C#CHECKCAST_ARRAY_OF#_", "MyClasss", "_#C#CHECKCAST_ARRAY_OF#_", "Object"}, applyIfOr={}, applyIfNot={})"
     > Phase "PrintOptoAssembly":
       - failOn: Graph contains forbidden nodes:
         * Constraint 2: "(((?i:cmp|CLFI|CLR).*precise \\[.*Object:|.*(?i:mov|or).*precise \\[.*Object:.*\\R.*(cmp|CMP|CLR)))"
           - Matched forbidden node:
             * cmpl    R10, narrowklass: precise [java/lang/Object:

2) Method "public java.lang.Object[] ir_framework.tests.CheckCastArray.arrayCopy(java.lang.Object[],java.lang.Class)" - [Failed IR rules: 1]:
   * @IR rule 1: "@compiler.lib.ir_framework.IR(phase={DEFAULT}, applyIfCPUFeatureAnd={}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={}, applyIfAnd={}, failOn={"_#CHECKCAST_ARRAYCOPY#_"}, applyIfOr={}, applyIfNot={})"
     > Phase "PrintOptoAssembly":
       - failOn: Graph contains forbidden nodes:
         * Constraint 1: "(.*((?i:call_leaf_nofp,runtime)|CALL,\\s?runtime leaf nofp|BCTRL.*.leaf call).*checkcast_arraycopy.*)"
           - Matched forbidden node:
             * 1a5     call_leaf_nofp,runtime  checkcast_arraycopy
```

rwestrel:
> I did not review the code in detail yet but here are some failures from preliminary testing:

Thanks for running tests. New commit should fix the issues you reported.

TobiHartmann:
`testlibrary_tests/ir_framework/tests/TestIRMatching.java` still fails: `Failed IR Rules (21) of Methods (8)`:

```
1) Method "public void ir_framework.tests.Traps.classCheck()" - [Failed IR rules: 3]:
   * @IR rule 1: "@compiler.lib.ir_framework.IR(phase={DEFAULT}, applyIfCPUFeatureAnd={}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={}, failOn={"_#TRAP#_"}, applyIfAnd={}, applyIfOr={}, applyIfNot={})"
     > Phase "PrintIdeal":
       - failOn: Graph contains forbidden nodes:
         * Constraint 1: "(\\d+(\\s){2}(CallStaticJava.*)+(\\s){2}===.*uncommon_trap.*reason)"
           - Matched forbidden nodes (2):
             * 39  CallStaticJava  === 33 6 7 8 9 (38 1 1 10 29 ) [[ 40 ]] # Static uncommon_trap(reason
             * 60  CallStaticJava  === 245 6 7 8 9 (59 1 1 10 24 ) [[ 61 ]] # Static uncommon_trap(reason
   * @IR rule 2: "@compiler.lib.ir_framework.IR(phase={DEFAULT}, applyIfCPUFeatureAnd={}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={}, failOn={"_#CLASS_CHECK_TRAP#_"}, applyIfAnd={}, applyIfOr={}, applyIfNot={})"
     > Phase "PrintIdeal":
       - failOn: Graph contains forbidden nodes:
         * Constraint 1: "(\\d+(\\s){2}(CallStaticJava.*)+(\\s){2}===.*uncommon_trap.*class_check)"
           - Matched forbidden node:
             * 60  CallStaticJava  === 245 6 7 8 9 (59 1 1 10 24 ) [[ 61 ]] # Static uncommon_trap(reason='class_check
   * @IR rule 3: "@compiler.lib.ir_framework.IR(phase={DEFAULT}, applyIfCPUFeatureAnd={}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={}, failOn={"_#NULL_CHECK_TRAP#_"}, applyIfAnd={}, applyIfOr={}, applyIfNot={})"
     > Phase "PrintIdeal":
       - failOn: Graph contains forbidden nodes:
         * Constraint 1: "(\\d+(\\s){2}(CallStaticJava.*)+(\\s){2}===.*uncommon_trap.*null_check)"
           - Matched forbidden node:
             * 39  CallStaticJava  === 33 6 7 8 9 (38 1 1 10 29 ) [[ 40 ]] # Static uncommon_trap(reason='null_check

[...]

```

rwestrel:
> `testlibrary_tests/ir_framework/tests/TestIRMatching.java` still fails: `Failed IR Rules (21) of Methods (8)`:

Thanks for running testing again. Hopefully the new commit fixes `TestIRMatching.java` for good.

iwanowww:
Nice enhancement, Roland!

(I believe the monomorphic case you refer to is covered by `GraphKit::maybe_cast_profiled_receiver()`.) 

I'm trying to understand what are the implications if you generate profile-based type guards early (during parsing). Any particular benefits from late expansion or downsides from early expansion? I'd expect that additional type info may be helpful (even though bimorphic/polymorphic cases are less useful than monomorphic one).

Handling it during parsing would relieve `SubTypeCheck` from caring about profile data and enable placing an uncommon trap on slow path for bimorphic case. (Doing that during macro expansion would require `SubTypeCheckNode` to keep JVM state.)

What are the implications if you make profiling changes separately? My understanding is you gain access to accurate probabilities and ability to distinguish between bi-/poly-/mega-morphic cases. Is it correct?

Speaking of alternative ways to pass profile info around, you could just embed `ciCallProfile` in `SubTypeCheck`. Any particular reasons not to do so?

rwestrel:
> I'm trying to understand what are the implications if you generate profile-based type guards early (during parsing). Any particular benefits from late expansion or downsides from early expansion? I'd expect that additional type info may be helpful (even though bimorphic/polymorphic cases are less useful than monomorphic one).

Late expansion allows some optimizations to trigger that wouldn't otherwise. Something like:

```
if (o.klass = profile1) {
  goto success; 
} else if (o.klass = profile2) {
  goto success;
} else if (o instanceof super) {
  goto success;
} else {
  goto failure;
}
```
is unlikely to optimize as well as:
```
if (o instanceof super) {
}
```
with split if, loop predication or finding a dominating if with an identical type check.

> Handling it during parsing would relieve `SubTypeCheck` from caring about profile data and enable placing an uncommon trap on slow path for bimorphic case. (Doing that during macro expansion would require `SubTypeCheckNode` to keep JVM state.)

Carrying the JVM state in the `SubTypeCheck` looks like too much extra complexity to me. With profiling of branches, we could get an uncommon trap in:

```
if (!(a instanceof super)) {
  // never taken
}
```

even if profile data reports more than a single receiver at the checkcast.

> What are the implications if you make profiling changes separately? My understanding is you gain access to accurate probabilities and ability to distinguish between bi-/poly-/mega-morphic cases. Is it correct?

Pushing this without the change to profile data collection? We have no way to tell if, when profile data reports 2 classes, they are common or not. So yes, that's correct.

> Speaking of alternative ways to pass profile info around, you could just embed `ciCallProfile` in `SubTypeCheck`. Any particular reasons not to do so?

It felt easier in terms of memory management. If we have some extra data embedded in the `SubTypeCheck` node, is it a pointer or the full data structure? If it is a pointer, do we clone the data on node clone? Try to reclaim memory on node destruction? Where should the data live so it's not destroyed while we need it but doesn't live longer that's required?

rwestrel:
> > Speaking of alternative ways to pass profile info around, you could just embed `ciCallProfile` in `SubTypeCheck`. Any particular reasons not to do so?
> 
> It felt easier in terms of memory management. If we have some extra data embedded in the `SubTypeCheck` node, is it a pointer or the full data structure? If it is a pointer, do we clone the data on node clone? Try to reclaim memory on node destruction? Where should the data live so it's not destroyed while we need it but doesn't live longer that's required?

Let me see if I can simplify that part.

rwestrel:
> Speaking of alternative ways to pass profile info around, you could just embed `ciCallProfile` in `SubTypeCheck`. Any particular reasons not to do so?

One thing to consider is that we don't necessarily want to common 2 `SubTypeCheck` nodes with the same object/super kass inputs. If we have something like:

```
if (...) {
  if (o instanceof C) {
  }
} else {
  if (o instanceof C) {
  }
}
```

There are 2 `SubTypeCheck` nodes with possibly different profile data. In order to common them, we would have to decide what profile data to keep (or drop profile data entirely) which could lead to less performant subtype checks. I think we want to decide on a case by case basis whether 2 `SubTypeCheck` nodes need to be commoned or not which is essentially what the logic I added does. Having extra edges makes that fairly straightforward.

iwanowww:
> Late expansion allows some optimizations to trigger that wouldn't otherwise. 

I agree that early expansion may hinder optimizations for polymorphic case.
But bimorphic case may enrich the receiver with more precise type information. 

> Carrying the JVM state in the SubTypeCheck looks like too much extra complexity to me.

Yes, I agree.

> It felt easier in terms of memory management. If we have some extra data embedded in the SubTypeCheck node, is it a pointer or the full data structure?

`ciCallProfile` has fixed size and is passed by value. Embedding the whole structure inside `SubTypeCheck` doesn't look problematic. It refers to CI entities which should be kept alive for the duration of the compilation.  

> One thing to consider is that we don't necessarily want to common 2 SubTypeCheck nodes with the same object/super kass inputs. 

I thought about commoning of SubTypeCheck nodes as well and there are other cases when it may be undesireable. For example, when 2 checks are performed in 2 subbranches with low frequencies, we don't want to place commoned check in a hot dominating block. 

I'd prefer to see `SubTypeCheck` to have control input which is explicitly relaxed to accommodate commoning.


Overall, I'm fine with late expansion of profile-guided type checks for now, but embedding profile data info SubTypeCheck should significantly simplify the patch without compromising the benefits. Also, enhancing profiling support separately may be a viable tradeoff as well. Inaccuracies in code shape classification don't look like a critical issue when the guards are introduced during macro expansion. 

We can explore other optimization opportunities later (e.g., profile for superclasses may help reflection case).

rwestrel:
> > It felt easier in terms of memory management. If we have some extra data embedded in the SubTypeCheck node, is it a pointer or the full data structure?
> 
> `ciCallProfile` has fixed size and is passed by value. Embedding the whole structure inside `SubTypeCheck` doesn't look problematic. It refers to CI entities which should be kept alive for the duration of the compilation.

Some `SubTypeCheck` nodes have no profile data associated with them. It doesn't seem right that all of them have to carry around `ciCallProfile`. That's maybe a small overhead at this point but we could profile more than 2 receivers in the future so that overhead could change.

> I'd prefer to see `SubTypeCheck` to have control input which is explicitly relaxed to accommodate commoning.

I don't how that would simplify things. `SubTypeCheck` is a `Cmp` node. As a consequence some optimizations that apply to `Cmp` nodes apply to it. `Cmp` nodes have no control. Setting the control to one of them will break things and require extra logic to accomodate the extra control (split if will likely break for instance). 

Beyond that, we'll need logic to find  `SubTypeCheck` nodes that be commoned which I expect would be in the same locations I already extra code. I don't understand what the control edge would make simpler.

> Overall, I'm fine with late expansion of profile-guided type checks for now, but embedding profile data info SubTypeCheck should significantly simplify the patch without compromising the benefits.

At this point, I fail to see how. I'm not saying the solution I propose is great but it seems to me the problem it solves need to be solved in any case.

> Also, enhancing profiling support separately may be a viable tradeoff as well. Inaccuracies in code shape classification don't look like a critical issue when the guards are introduced during macro expansion.

Is that out of concern that getting the code  done on all platforms will be too complicated?

iwanowww:
Overall, I'd prefer to leave commoning considerations for a separate enhancement. 

Embedding `ciCallProfile` looks to me much cleaner than exploding its content into node inputs.

Having profile info explicitly fed into `SubTypeCheck` as node inputs in practice defeats any possible sharing unless the nodes are constructed from the very same profile data. The types, their order, and frequencies have to perfectly match in order for commoning to happen. You already have  `IfNode::same_condition()` to alleviate some of the effects of broken sharing.

When you embed profiling info you are left with a choice how to common nodes (whether to take profiling info into account or not). But if you simply ignore it until macro expansion, the behavior will stay the same as it is now.

I prefer the patch to be focused on slow path case (reduce the frequency of secondary super cache checks & updates) and leave the rest for future considerations.  

As an example, it's still an open question for me should `IfNode::search_identical()` take profile info into account. Current patch ignores profile-related info (`IfNode::same_condition()` check), but maybe it is worth merging the profiles instead?  


> Some SubTypeCheck nodes have no profile data associated with them.

I don't consider footprint as an issue here. `SubTypeCheck`s are relatively rare and `ciCallProfile` size is quite small for any practical morphism limits. Additional profiling may introduce more about 1-2 additional slots (rather than 10s or 100s) and the main footprint hit will be on runtime side (in MDOs). 

> Is that out of concern that getting the code done on all platforms will be too complicated?

It does look like an excessive requirement, but I'm not too much concerned about it. If you think it's better to get the full support all at once, I'm perfectly fine with that. It just seems cleaner to refine profiling part separately. There are open questions which may be well out of scope for the proposed enhancement. 

For example, while `checkcast`/`aastore` behave very similarly to `invokevirtual`/`invokeinterface` (very low rate of failures), `instanceof` is different and can expose very high rates of failures (esp. in case of chained `instanceof` checks). Should we continue profiling for that? (Can C2 benefit from such info? I believe so: we could skip SSC check if failure rate is too high.) 

Also, I refrained from commenting on naming, but `ciCallProfile` does look confusing when it comes to `checkcast`,  `aastore`, and `instanceof` cases.

iwanowww (commented):
Thanks, Roland. IR shape looks much better now.

> I also still think that preventing commoning is important so some path doesn't end up with profile data from some other path. 

I took a closer look at the relevant code (in particular, I forgot that`PhaseMacroExpand::expand_subtypecheck_node` creates a dedicated copy for each user) and now agree with you that commoning between unrelated paths is undesirable. Moreover, I'm in favor of completely disabling sharing for `SubTypeCheck` node. Considering `IfNode::search_identical()` handles `SubTypeCheck` case now, I don't see much value in special handling for nodes without associated bytecode location info.

> I still think it's important to change profile collection to have data that is as accurate as possible. If the change needs to be split, I think profile collection changes should go in first.

What's the plan if we agree on adjusting profile collection? Should all the platforms be updated all at once? If not, how is it intended to work during transition period?

test/hotspot/jtreg/compiler/c2/irTests/ProfileAtTypeCheck.java
```diff
@@ -0,0 +1,412 @@
...
+ * @requires vm.compiler2.enabled
+ * @run driver compiler.c2.irTests.ProfileAtTypeCheck
+ */
+
+public class ProfileAtTypeCheck {
+    public static void main(String[] args) {
+        ArrayList<String> flags = new ArrayList<>();
+        flags.add("-XX:TypeProfileSubTypeCheckCommonThreshold=90");
+        if (!Platform.is32bit()) {
+            flags.add("-XX:-UseCompressedClassPointers");
```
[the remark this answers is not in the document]
rwestrel: It is to make it easier to match the IR. I could update the rules so it's not required.

rwestrel:
@iwanowww thanks for looking at this some more.

> I don't see much value in special handling for nodes without associated bytecode location info.

Right. I removed it.

> What's the plan if we agree on adjusting profile collection? Should all the platforms be updated all at once? If not, how is it intended to work during transition period?

I thought it would be better to update profile collection on all platforms as part of this PR initially so no platforms are left behind. But that's likely unrealistic. I pushed a new commit that I think would support both ways of doing profiling so there's no need to have all platforms in sync. What do you think?

src/hotspot/share/opto/c2_globals.hpp
```diff
@@ -771,6 +771,12 @@
                                                                             \
   product(bool, VerifyReceiverTypes, trueInDebug, DIAGNOSTIC,               \
           "Verify receiver types at runtime")                               \
+                                                                            \
+  product(intx, TypeProfileSubTypeCheckCommonThreshold, 50,                 \
```
iwanowww, line 775: Is it better to declare it diagnostic?

iwanowww (approved):
Looks good.

I assume SA code is not sensitive to receiver profiling details and the changes there are JVMCI-specific.

test/hotspot/jtreg/compiler/c2/irTests/ProfileAtTypeCheck.java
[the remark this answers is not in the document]
iwanowww: I'm fine with it either way.

test/hotspot/jtreg/compiler/c2/irTests/ProfileAtTypeCheck.java
```diff
@@ -0,0 +1,412 @@
...
+        }
+        flags.add("-XX:+IgnoreUnrecognizedVMOptions");
+        flags.add("-XX:+UseParallelGC");
```
iwanowww, line 47: The test fails when another GC is specified externally.

RealFYang:
Hello, we witnessed  the same problem on linux-riscv64 platform. So I prepared changes for this platform by referencing the aarch64 port.
[14375-riscv-v4.diff.txt](https://github.com/openjdk/jdk/files/12074890/14375-riscv-v4.diff.txt)
Tier1-3 tested and this also passed the newly added test by this PR. Could you please add this? Thanks.

src/hotspot/share/opto/c2_globals.hpp
rwestrel: I don't have a strong opinion but the similar `TypeProfileMajorReceiverPercent` is product.

test/hotspot/jtreg/compiler/c2/irTests/ProfileAtTypeCheck.java
rwestrel: Should be fixed in the new commit. I also improved the test a bit (added one test case with array store checks, duplicated the rules so there's no need to run with -XX:-UseCompressedClassPointers, run tests with only c1 profile collection and only interpreter profile collection to exercise new profile collection code paths independently).

rwestrel:
> Hello, we witnessed the same problem on linux-riscv64 platform. So I prepared changes for this platform by referencing the aarch64 port. [14375-riscv-v4.diff.txt](https://github.com/openjdk/jdk/files/12074890/14375-riscv-v4.diff.txt) Tier1-3 tested and this also passed the newly added test by this PR. Could you please add this? Thanks.

Added. Thanks for contributing riscv support.

rwestrel:
Anyone else for a review?

TobiHartmann:
I didn't get to review this yet but I plan to - probably only after a short vacation next week. I did run some performance and correctness testing though. Performance looks good (neutral). Correctness testing looks good too but found this single failure:

`ProfileAtTypeCheck` fails IR verification with `-ea -esa -XX:CompileThreshold=100 -XX:+UnlockExperimentalVMOptions -server -XX:-TieredCompilation`:

```
Failed IR Rules (11) of Methods (9)
-----------------------------------
1) Method "public static void compiler.c2.irTests.ProfileAtTypeCheck.test10(boolean)" - [Failed IR rules: 1]:
   * @IR rule 2: "@compiler.lib.ir_framework.IR(applyIfCPUFeatureAnd={}, phase={ITER_GVN1}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={"_#SUBTYPE_CHECK#_", "1"}, failOn={}, applyIfAnd={}, applyIfOr={}, applyIfNot={})"
     > Phase "Iter GVN 1":
       - counts: Graph contains wrong number of nodes:
         * Constraint 1: "(\\d+(\\s){2}(SubTypeCheck.*)+(\\s){2}===.*)"
           - Failed comparison: [found] 2 = 1 [given]
             - Matched nodes (2):
               * 45  SubTypeCheck  === _ 36 27  [[ 130 ]]  profiled at:  compiler.c2.irTests.ProfileAtTypeCheck::test10:3 !jvms: ProfileAtTypeCheck::test10 @ bci:3 (line 270)
               * 90  SubTypeCheck  === _ 82 27  [[ 126 ]]  profiled at:  compiler.c2.irTests.ProfileAtTypeCheck::test10:16 !jvms: ProfileAtTypeCheck::test10 @ bci:16 (line 272)

2) Method "public static void compiler.c2.irTests.ProfileAtTypeCheck.test12()" - [Failed IR rules: 2]:
   * @IR rule 1: "@compiler.lib.ir_framework.IR(applyIfCPUFeatureAnd={}, phase={AFTER_PARSING}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={"_#SUBTYPE_CHECK#_", "3"}, failOn={}, applyIfAnd={}, applyIfOr={}, applyIfNot={})"
     > Phase "After Parsing":
       - counts: Graph contains wrong number of nodes:
         * Constraint 1: "(\\d+(\\s){2}(SubTypeCheck.*)+(\\s){2}===.*)"
           - Failed comparison: [found] 0 = 3 [given]
           - No nodes matched!
   * @IR rule 2: "@compiler.lib.ir_framework.IR(applyIfCPUFeatureAnd={}, phase={PHASEIDEALLOOP_ITERATIONS}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={"_#SUBTYPE_CHECK#_", "1"}, failOn={}, applyIfAnd={}, applyIfOr={}, applyIfNot={})"
     > Phase "PhaseIdealLoop iterations":
       - NO compilation output found for this phase! Make sure this phase is emitted or remove it from the list of compile phases in the @IR rule to match on.

3) Method "public static void compiler.c2.irTests.ProfileAtTypeCheck.test15(java.lang.Object)" - [Failed IR rules: 1]:
   * @IR rule 3: "@compiler.lib.ir_framework.IR(applyIfCPUFeatureAnd={}, phase={MACRO_EXPANSION}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={"_#CMP_P#_", "5", "_#LOAD_KLASS#_", "1", "_#LOAD_NKLASS#_", "1", "_#PARTIAL_SUBTYPE_CHECK#_", "1"}, failOn={}, applyIfAnd={"UseCompressedClassPointers", "true", "UseParallelGC", "true"}, applyIfOr={}, applyIfNot={})"
     > Phase "Macro expand":
       - counts: Graph contains wrong number of nodes:
         * Constraint 1: "(\\d+(\\s){2}(CmpP.*)+(\\s){2}===.*)"
           - Failed comparison: [found] 3 = 5 [given]
             - Matched nodes (3):
               * 39  CmpP  === _ 10 38  [[ 40 ]]  !jvms: ProfileAtTypeCheck::test15 @ bci:5 (line 429)
               * 117  CmpP  === _ 116 38  [[ 118 ]] 
               * 123  CmpP  === _ 97 35  [[ 125 ]]  !orig=[100]

4) Method "public static void compiler.c2.irTests.ProfileAtTypeCheck.test2(java.lang.Object)" - [Failed IR rules: 2]:
   * @IR rule 1: "@compiler.lib.ir_framework.IR(applyIfCPUFeatureAnd={}, phase={AFTER_PARSING}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={}, failOn={"_#SUBTYPE_CHECK#_"}, applyIfAnd={}, applyIfOr={}, applyIfNot={})"
     > Phase "After Parsing":
       - failOn: Graph contains forbidden nodes:
         * Constraint 1: "(\\d+(\\s){2}(SubTypeCheck.*)+(\\s){2}===.*)"
           - Matched forbidden node:
             * 39  SubTypeCheck  === _ 30 21  [[ 53 ]]  profiled at:  compiler.c2.irTests.ProfileAtTypeCheck::test2:1 !jvms: ProfileAtTypeCheck::test2 @ bci:1 (line 102)
   * @IR rule 3: "@compiler.lib.ir_framework.IR(applyIfCPUFeatureAnd={}, phase={AFTER_PARSING}, applyIf={"UseCompressedClassPointers", "true"}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={"_#CMP_P#_", "2", "_#LOAD_NKLASS#_", "1"}, failOn={}, applyIfAnd={}, applyIfOr={}, applyIfNot={})"
     > Phase "After Parsing":
       - counts: Graph contains wrong number of nodes:
         * Constraint 1: "(\\d+(\\s){2}(CmpP.*)+(\\s){2}===.*)"
           - Failed comparison: [found] 1 = 2 [given]
             - Matched node:
               * 25  CmpP  === _ 10 24  [[ 26 ]]  !jvms: ProfileAtTypeCheck::test2 @ bci:1 (line 102)
         * Constraint 2: "(\\d+(\\s){2}(LoadNKlass.*)+(\\s){2}===.*)"
           - Failed comparison: [found] 0 = 1 [given]
           - No nodes matched!

5) Method "public static void compiler.c2.irTests.ProfileAtTypeCheck.test3(java.lang.Object)" - [Failed IR rules: 1]:
   * @IR rule 1: "@compiler.lib.ir_framework.IR(applyIfCPUFeatureAnd={}, phase={AFTER_PARSING}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={"_#SUBTYPE_CHECK#_", "1"}, failOn={}, applyIfAnd={}, applyIfOr={}, applyIfNot={})"
     > Phase "After Parsing":
       - counts: Graph contains wrong number of nodes:
         * Constraint 1: "(\\d+(\\s){2}(SubTypeCheck.*)+(\\s){2}===.*)"
           - Failed comparison: [found] 2 = 1 [given]
             - Matched nodes (2):
               * 40  SubTypeCheck  === _ 30 21  [[ 46 ]]  profiled at:  compiler.c2.irTests.ProfileAtTypeCheck::test3:1 !jvms: ProfileAtTypeCheck::test3 @ bci:1 (line 120)
               * 62  SubTypeCheck  === _ 30 21  [[ 67 ]]  profiled at:  compiler.c2.irTests.ProfileAtTypeCheck::test3:8 !jvms: ProfileAtTypeCheck::test3 @ bci:8 (line 121)

6) Method "public static void compiler.c2.irTests.ProfileAtTypeCheck.test5(java.lang.Object)" - [Failed IR rules: 1]:
   * @IR rule 3: "@compiler.lib.ir_framework.IR(applyIfCPUFeatureAnd={}, phase={MACRO_EXPANSION}, applyIf={"UseCompressedClassPointers", "true"}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={"_#CMP_P#_", "5", "_#LOAD_KLASS#_", "1", "_#LOAD_NKLASS#_", "1", "_#PARTIAL_SUBTYPE_CHECK#_", "1"}, failOn={}, applyIfAnd={}, applyIfOr={}, applyIfNot={})"
     > Phase "Macro expand":
       - counts: Graph contains wrong number of nodes:
         * Constraint 1: "(\\d+(\\s){2}(CmpP.*)+(\\s){2}===.*)"
           - Failed comparison: [found] 3 = 5 [given]
             - Matched nodes (3):
               * 25  CmpP  === _ 10 24  [[ 26 ]]  !jvms: ProfileAtTypeCheck::test5 @ bci:1 (line 161)
               * 108  CmpP  === _ 107 24  [[ 109 ]] 
               * 114  CmpP  === _ 88 21  [[ 116 ]]  !orig=[91]

7) Method "public static boolean compiler.c2.irTests.ProfileAtTypeCheck.test7(java.lang.Object)" - [Failed IR rules: 1]:
   * @IR rule 3: "@compiler.lib.ir_framework.IR(applyIfCPUFeatureAnd={}, phase={MACRO_EXPANSION}, applyIf={"UseCompressedClassPointers", "true"}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={"_#CMP_P#_", "5", "_#LOAD_KLASS#_", "1", "_#LOAD_NKLASS#_", "1", "_#PARTIAL_SUBTYPE_CHECK#_", "1"}, failOn={}, applyIfAnd={}, applyIfOr={}, applyIfNot={})"
     > Phase "Macro expand":
       - counts: Graph contains wrong number of nodes:
         * Constraint 1: "(\\d+(\\s){2}(CmpP.*)+(\\s){2}===.*)"
           - Failed comparison: [found] 3 = 5 [given]
             - Matched nodes (3):
               * 26  CmpP  === _ 10 25  [[ 27 ]]  !jvms: ProfileAtTypeCheck::test7 @ bci:1 (line 198)
               * 86  CmpP  === _ 85 25  [[ 87 ]] 
               * 92  CmpP  === _ 66 22  [[ 94 ]]  !orig=[69]

8) Method "public static void compiler.c2.irTests.ProfileAtTypeCheck.test8(java.lang.Object)" - [Failed IR rules: 1]:
   * @IR rule 3: "@compiler.lib.ir_framework.IR(applyIfCPUFeatureAnd={}, phase={MACRO_EXPANSION}, applyIf={"UseCompressedClassPointers", "true"}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={"_#CMP_P#_", "5", "_#LOAD_KLASS#_", "1", "_#LOAD_NKLASS#_", "1", "_#PARTIAL_SUBTYPE_CHECK#_", "1"}, failOn={}, applyIfAnd={}, applyIfOr={}, applyIfNot={})"
     > Phase "Macro expand":
       - counts: Graph contains wrong number of nodes:
         * Constraint 1: "(\\d+(\\s){2}(CmpP.*)+(\\s){2}===.*)"
           - Failed comparison: [found] 3 = 5 [given]
             - Matched nodes (3):
               * 25  CmpP  === _ 10 24  [[ 26 ]]  !jvms: ProfileAtTypeCheck::test8 @ bci:1 (line 216)
               * 108  CmpP  === _ 107 24  [[ 109 ]] 
               * 114  CmpP  === _ 88 21  [[ 116 ]]  !orig=[91]

9) Method "public static void compiler.c2.irTests.ProfileAtTypeCheck.test9(boolean,boolean,java.lang.Object,java.lang.Object)" - [Failed IR rules: 1]:
   * @IR rule 2: "@compiler.lib.ir_framework.IR(applyIfCPUFeatureAnd={}, phase={PHASEIDEALLOOP1}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={"_#SUBTYPE_CHECK#_", "2"}, failOn={}, applyIfAnd={}, applyIfOr={}, applyIfNot={})"
     > Phase "PhaseIdealLoop 1":
       - counts: Graph contains wrong number of nodes:
         * Constraint 1: "(\\d+(\\s){2}(SubTypeCheck.*)+(\\s){2}===.*)"
           - Failed comparison: [found] 1 = 2 [given]
             - Matched node:
               * 104  SubTypeCheck  === _ 166 87  [[ 140 ]]  profiled at:  compiler.c2.irTests.ProfileAtTypeCheck::test9:56 !jvms: ProfileAtTypeCheck::test9 @ bci:56 (line 253)
```

rwestrel:
> I didn't get to review this yet but I plan to - probably only after a short vacation next week. I did run some performance and correctness testing though. Performance looks good (neutral). Correctness testing looks good too but found this single failure:

Thank you! I pushed a fix to the test case that should take care of that one that one.

TobiHartmann (approved):
Great work, Roland. The changes look good to me. I just added some minor style suggestions of things I stumbled upon when reading the code.

What's the plan for PPC, S390, 32-bit ARM, Zero support?

rwestrel:
> Great work, Roland. The changes look good to me. I just added some minor style suggestions of things I stumbled upon when reading the code.

Thanks for the careful review. I pushed an update that takes care of your comments.

> What's the plan for PPC, S390, 32-bit ARM, Zero support?

Not much of a plan other than filing a bug and pinging those involved with the other architectures.

src/hotspot/share/opto/subtypenode.hpp
```diff
@@ -61,6 +61,7 @@ class SubTypeCheckNode : public CmpNode {
 #endif
 
 private:
+  // method/bci that for this subtype check so profile data can be retrieved after parsing is over
```
TobiHartmann, line 64:
```suggestion
  // method/bci for this subtype check so profile data can be retrieved after parsing is over
```

TobiHartmann (approved):
Thanks for making these changes. Looks good to me.

rwestrel:
@iwanowww @TobiHartmann thanks for the reviews. @tkrodriguez thanks for the help with the jvmci code.

rwestrel:
I filed https://bugs.openjdk.org/browse/JDK-8315750 as follow up bug for not yet supported platforms
FYI @TheRealMDoerr @offamitkumar @bulasevich ( you are listed as maintainer of those ports in https://wiki.openjdk.org/display/HotSpot/Ports )

TheRealMDoerr:
> I filed https://bugs.openjdk.org/browse/JDK-8315750 as follow up bug for not yet supported platforms
> FYI @TheRealMDoerr @offamitkumar @bulasevich ( you are listed as maintainer of those ports in https://wiki.openjdk.org/display/HotSpot/Ports )

Thanks! I've split the issue into 3 and assigned the PPC one to myself.

rwestrel:
FTR, I had to update the test case because it failed on 32 bits x86: the IR rules test whether compressed oops are on or off but that option is not supported on 32 bits.

PR #14378, opened 2023-06-08
8309305: sun/security/ssl/SSLSocketImpl/BlockedAsyncClose.java fails with jtreg test timeout

mpdonova:
This PR improves the reliability of the BlockedAsyncClose test by addressing an edge case/race condition between the two test threads. The purpose of the test is to verify that an SSLSocket can be closed if a thread is blocked in a write operation. 

The test starts a "write" thread that writes data to a socket until the output buffer was filled, causing the write operation to block. The main thread then calls `SSLSocket.close()`. The original code used `Thread.sleep(1000)` to wait for the write-thread to block. However, 1 second isn't always long enough and if the write-thread isn't blocked and the output buffer is full (or almost full), the `socket.close()` call may block when it tries to send the close_notify alert. This is the condition that caused this bug.

My change uses a Lock to determine if the write thread is blocked. In the write thread, the lock creates a critical section around the `write()` call. The main thread uses `tryLock()` with a timeout to determine that the write() call is taking too long and thus likely blocked. 

While there, I also updated the test to use the SSLContextTemplate class.

---------
### Progress

### Issue
 * [JDK-8309305](https://bugs.openjdk.org/browse/JDK-8309305): sun/security/ssl/SSLSocketImpl/BlockedAsyncClose.java fails with jtreg test timeout (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14378/head:pull/14378` \
`$ git checkout pull/14378`

`$ git checkout pull/14378` \
`$ git pull https://git.openjdk.org/jdk.git pull/14378/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14378`

`$ git pr show -t 14378`

</details>

<a href="https://git.openjdk.org/jdk/pull/14378.diff">https://git.openjdk.org/jdk/pull/14378.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14378#issuecomment-1582695843)

### DIFF
--- a/test/jdk/sun/security/ssl/SSLSocketImpl/BlockedAsyncClose.java
+++ b/test/jdk/sun/security/ssl/SSLSocketImpl/BlockedAsyncClose.java
@@ -87,7 +87,7 @@ public class BlockedAsyncClose extends SSLContextTemplate implements Runnable {
 
         // if the writeLock is not released by the other thread within 10
         // seconds it is probably blocked, and we can try to close the socket
-        while(writeLock.tryLock(10, TimeUnit.SECONDS)) {
+        while (writeLock.tryLock(10, TimeUnit.SECONDS)) {
             writeLock.unlock();
         }
 
--- a/test/jdk/sun/security/ssl/SSLSocketImpl/BlockedAsyncClose.java
+++ b/test/jdk/sun/security/ssl/SSLSocketImpl/BlockedAsyncClose.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


test/jdk/sun/security/ssl/SSLSocketImpl/BlockedAsyncClose.java
```diff
@@ -94,23 +84,22 @@ public BlockedAsyncClose() throws Exception {
...
 
         socket.setSoLinger(true, 10);
-        System.out.println("Calling Socket.close");
 
-        // Sleep for a while so that the write thread blocks by hitting the
-        // output stream buffer limit.
-        Thread.sleep(1000);
+        // if the writeLock is not released by the other thread within 10
+        // seconds it is probably blocked, and we can try to close the socket
+        while(writeLock.tryLock(10, TimeUnit.SECONDS)) {
```
turbanoff, line 90:
```suggestion
        while (writeLock.tryLock(10, TimeUnit.SECONDS)) {
```

mpdonova:
Can someone please review this? Thanks!

djelinski (approved):
LGTM.
Some OSes spontaneously increase the buffer size after a few seconds, unblocking blocked write operations. We also use 10 seconds limit in networking tests, seems to work pretty well. See https://github.com/openjdk/jdk/blob/9e4fc568a6f1a93c84a84d6cc5220c6eb4e546a5/test/jdk/java/net/httpclient/websocket/PendingOperations.java#L43-L47

PR #14379, opened 2023-06-08
8309660: C2: failed: XMM register should be 0-15 (UseKNLSetting and ConvF2HF)

eme64:
Context: `Float.floatToFloat16` -> `vcvtps2ph`.

**Problem**
```
vcvtps2ph
pre=Assembler::VEX_SIMD_66
opc=Assembler::VEX_OPCODE_0F_3A
VEX.128.66.0F3A
requires F16C
```
https://www.felixcloutier.com/x86/vcvtps2ph

So this is a non-AVX512 feature, and we should only use the registers `xmm0-15`.

There is also a AVX512 version, but it requires `AVX512VL and AVX512F`.

So on `x64`, we should only use registers `xmm0-15` if we do not have `AVX512VL`, and if we have it, then we can use `xmm0-31`.

**Suggested Solution**
As @sviswa7 suggested, we should just use the `vlRegF` instead of `regF`, see discussion in comments.

**Testing**
I simulated the patch on intel's `sde`. So now I'm confident that I don't generate code that uses `AVX512VL` registers (XMM16-31).

Running: tier1-6 + stress testing.

---------
### Progress

### Issue
 * [JDK-8309660](https://bugs.openjdk.org/browse/JDK-8309660): C2: failed: XMM register should be 0-15 (UseKNLSetting and ConvF2HF) (**Bug** - P4)

### Reviewers
 * [Sandhya Viswanathan](https://openjdk.org/census#sviswanathan) (@sviswa7 - **Reviewer**)
 * [Jatin Bhateja](https://openjdk.org/census#jbhateja) (@jatin-bhateja - **Reviewer**)

### Contributors
 * Sandhya Viswanathan `<sviswanathan@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14379/head:pull/14379` \
`$ git checkout pull/14379`

`$ git checkout pull/14379` \
`$ git pull https://git.openjdk.org/jdk.git pull/14379/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14379`

`$ git pr show -t 14379`

</details>

<a href="https://git.openjdk.org/jdk/pull/14379.diff">https://git.openjdk.org/jdk/pull/14379.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14379#issuecomment-1584313653)

### DIFF
--- a/src/hotspot/cpu/x86/x86.ad
+++ b/src/hotspot/cpu/x86/x86.ad
@@ -3638,7 +3638,7 @@ instruct sqrtD_reg(regD dst) %{
   ins_pipe(pipe_slow);
 %}
 
-instruct convF2HF_reg_reg(rRegI dst, regF src, regF tmp) %{
+instruct convF2HF_reg_reg(rRegI dst, vlRegF src, vlRegF tmp) %{
   effect(TEMP tmp);
   match(Set dst (ConvF2HF src));
   ins_cost(125);
@@ -3682,7 +3682,7 @@ instruct vconvF2HF_mem_reg(memory mem, vec src) %{
   ins_pipe( pipe_slow );
 %}
 
-instruct convHF2F_reg_reg(regF dst, rRegI src) %{
+instruct convHF2F_reg_reg(vlRegF dst, rRegI src) %{
   match(Set dst (ConvHF2F src));
   format %{ "vcvtph2ps $dst,$src" %}
   ins_encode %{
--- a/src/hotspot/cpu/x86/assembler_x86.cpp
+++ b/src/hotspot/cpu/x86/assembler_x86.cpp
@@ -1944,7 +1944,7 @@ void Assembler::vcvtdq2pd(XMMRegister dst, XMMRegister src, int vector_len) {
 
 void Assembler::vcvtps2ph(XMMRegister dst, XMMRegister src, int imm8, int vector_len) {
   assert(VM_Version::supports_evex() || VM_Version::supports_f16c(), "");
-  InstructionAttr attributes(vector_len, /* rex_w */ false, /* legacy_mode */ false, /* no_mask_reg */ true, /*uses_vl */ true);
+  InstructionAttr attributes(vector_len, /* rex_w */ false, /* legacy_mode */ false, /* no_mask_reg */ true, /*uses_vl */ false);
   int encode = vex_prefix_and_encode(src->encoding(), 0, dst->encoding(), VEX_SIMD_66, VEX_OPCODE_0F_3A, &attributes);
   emit_int24(0x1D, (0xC0 | encode), imm8);
 }
--- a/src/hotspot/cpu/x86/assembler_x86.cpp
+++ b/src/hotspot/cpu/x86/assembler_x86.cpp
@@ -1944,7 +1944,7 @@ void Assembler::vcvtdq2pd(XMMRegister dst, XMMRegister src, int vector_len) {
 
 void Assembler::vcvtps2ph(XMMRegister dst, XMMRegister src, int imm8, int vector_len) {
   assert(VM_Version::supports_evex() || VM_Version::supports_f16c(), "");
-  InstructionAttr attributes(vector_len, /* rex_w */ false, /* legacy_mode */ false, /* no_mask_reg */ true, /*uses_vl */ false);
+  InstructionAttr attributes(vector_len, /* rex_w */ false, /* legacy_mode */ false, /* no_mask_reg */ true, /*uses_vl */ true);
   int encode = vex_prefix_and_encode(src->encoding(), 0, dst->encoding(), VEX_SIMD_66, VEX_OPCODE_0F_3A, &attributes);
   emit_int24(0x1D, (0xC0 | encode), imm8);
 }
--- a/test/hotspot/jtreg/compiler/vectorization/TestFloatConversionsVector.java
+++ b/test/hotspot/jtreg/compiler/vectorization/TestFloatConversionsVector.java
@@ -58,7 +58,14 @@ public class TestFloatConversionsVector {
         }
     }
 
-    @Run(test = {"test_float_float16"}, mode = RunMode.STANDALONE)
+    @Test
+    public void test_float_float16_strided(short[] sout, float[] finp) {
+        for (int i = 0; i < finp.length/2; i++) {
+            sout[i*2] = Float.floatToFloat16(finp[i*2]);
+        }
+    }
+
+    @Run(test = {"test_float_float16", "test_float_float16_strided"}, mode = RunMode.STANDALONE)
     public void kernel_test_float_float16() {
         finp = new float[ARRLEN];
         sout = new short[ARRLEN];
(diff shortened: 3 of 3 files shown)


sviswa7:
vcvtps2ph is correct. The problem is that the register from XMM16-XMM31 is being passed for KNL. The higher bank register is  and AVX512 feature. The fix would be:
--- a/src/hotspot/cpu/x86/x86.ad
+++ b/src/hotspot/cpu/x86/x86.ad
@@ -3638,7 +3638,7 @@ instruct sqrtD_reg(regD dst) %{
   ins_pipe(pipe_slow);
 %}

-instruct convF2HF_reg_reg(rRegI dst, regF src, regF tmp) %{
+instruct convF2HF_reg_reg(rRegI dst, vlRegF src, vlRegF tmp) %{
   effect(TEMP tmp);
   match(Set dst (ConvF2HF src));
   ins_cost(125);
@@ -3682,7 +3682,7 @@ instruct vconvF2HF_mem_reg(memory mem, vec src) %{
   ins_pipe( pipe_slow );
 %}

-instruct convHF2F_reg_reg(regF dst, rRegI src) %{
+instruct convHF2F_reg_reg(vlRegF dst, rRegI src) %{
   match(Set dst (ConvHF2F src));
   format %{ "vcvtph2ps $dst,$src" %}
   ins_encode %{

eme64:
@sviswa7 Thanks for the quick response! Yes, I was finally able to test it with `sde`, and got an error:

```
$ /oracle-work/sde-external-9.21.1-2023-04-24-lin/sde -knl -- ./java -XX:UseAVX=3 -XX:+UnlockDiagnosticVMOptions -XX:+UseKNLSetting -XX:CompileCommand=compileonly,Test2::test -XX:CompileCommand=printcompilation,Test2::* -XX:-TieredCompilation -Xbatch Test2.java
CompileCommand: compileonly Test2.test bool compileonly = true
CompileCommand: PrintCompilation Test2.* bool PrintCompilation = true
  57913   82 %  b        Test2::test @ 2 (30 bytes)
  61114   83    b        Test2::test (30 bytes)
TID 1 SDE-ERROR: Executed instruction not valid for specified chip (KNL): 0x7fa0c11a5d75: vcvtps2ph xmm0, xmm16, 0x4
Instruction bytes are: 62 e3 7d 08 1d c0 04
```
Looks like I would have been indeed using `xmm16` which is not allowed on KNL.

Ok, I think I also understand why your fix works: `vlRegF` is defined with `constraint(ALLOC_IN_RC(float_reg_vl))`, and `float_reg_vl` is defined as
```
reg_class_dynamic float_reg_vl(float_reg_evex, float_reg_legacy, %{ VM_Version::supports_evex() && VM_Version::supports_avx512vl() %} );
```
`reg_class_dynamic` evaluates the condition (checks if we have `evex` and `avx512vl`), and if we have it it picks the larger `float_reg_evex` (XMM0-XMM31), else it picks `float_reg_legacy` (XMM0-XMM15).

@sviswa7 thanks for the patch!

merykitty:
You also need to do the same for `convF2HF_mem_reg`

eme64:
@merykitty I'm not sure if you deleted your comment, but you basically asked me to check if there are any other cases that need this fix.

I think this one is ok, because it already explicitly requires `avx512vl`.
```
instruct convF2HF_mem_reg(memory mem, regF src, kReg ktmp, rRegI rtmp) %{
  predicate((UseAVX > 2) && VM_Version::supports_avx512vl());
```

I'm not sure about all the vector cases, like this `instruct vconvHF2F(vec dst, vec src)`. Can someone tell me where to find the definition of register class `vec`, cannot seem to find it in `x86.ad`.

eme64:
@merykitty @sviswa7 @fg1417 Is there a way to stress-test the registers? It seems this bug only triggered because we had a moderately large unrolling factor, and then did not vectorize, leaving lots of instructions with probably a higher register pressure. Would be nice to have some sort of testing where we generate more (all?) of the possible register combinations. What do you think?

sviswa7:
> vconvHF2F

vconvHF2F and the reverse vconvF2HF are correct. The vec() operand class is defined in x86.ad. It is a generic vector, replaced by vecS, vecD, vecX, vecY, vecZ depending on the vector length. The operand class vecX e.g., allocates in register class vectorx_reg_vlbwdq. The vectorx_reg_vlbwdq is a dynamic register class which either gets the whole range from xmm0-xmm31 or the limited range xmm0-xmm15 based on whether AVX512vlbwdq feature availability.

merykitty:
@eme64 Yes that was my mistake, that node requires AVX512VL so `vlRegF` and `regF` are the same.

> Is there a way to stress-test the registers?

Can we randomise the allocated register during register allocation?

Thanks.

eme64:
@merykitty Yes, randomization would be great. I don't know much about the register allocator, so feel free to do something like that if you want and have time ;)

@sviswa7 Is there something you want me to change still?

sviswa7:
> @merykitty Yes, randomization would be great. I don't know much about the register allocator, so feel free to do something like that if you want and have time ;)
> 
> @sviswa7 Is there something you want me to change still?

No additional changes from my side.

sviswa7 (approved):

iwanowww:
> Is there a way to stress-test the registers?

As an idea for such a stress test mode, is it possible to make `regF`/`vlRegF`, `regD`/`vlRegD` (and `vec`/`legVec` family of register classes) disjoint sets (`xmm0-xmm15` and `xmm16-xmm31`)? It should be enough to trigger relevant asserts whenever an AD instruction is used.

eme64:
@iwanowww @merykitty @sviswa7 I think the best would really be randomization. But I have no expertise in the register allocation. Maybe we can just disable a random subset of registers? That would also stress spilling, which sometimes has lead to issues before.

Anyway. I think this fix is ok regardless. Can I get a second review, please? ;)

src/hotspot/cpu/x86/x86.ad
```diff
@@ -3682,7 +3682,7 @@ instruct vconvF2HF_mem_reg(memory mem, vec src) %{
   ins_pipe( pipe_slow );
 %}
 
-instruct convHF2F_reg_reg(regF dst, rRegI src) %{
+instruct convHF2F_reg_reg(vlRegF dst, rRegI src) %{
```
jatin-bhateja, line 3685: I think we can remove change from destination operand, it will automatically be assigned register from lower register bank for non-evex targets based on dynamic_register_classes. Problem is manly around incorrect register class for source operands since their definitions may not be constrained to allocate from lower register bank unless we use a legacy register class (vlRegF).

jatin-bhateja (approved):
Fix looks good to me.  dynamic register class "float_reg_vl" should be able to pick correct register allocation set for vex-encoded instruction on KNL.

eme64:
Thanks @sviswa7 for the help with the patch! Thanks @jatin-bhateja for the review!
/integrate

PR #14382, opened 2023-06-08
8309675: Generational ZGC: compiler/gcbarriers/UnsafeIntrinsicsTest.java fails in nmt_commit

stefank:
The failure happens because we have extra verification code that is incorrect to use when calling the NMT tracking code. 

The fix is to simply inline the address calculation code *without* the verification code.

This fixes a tier3 failure. I'm going to run this through more extensive testing but it would be good to get this reviewed.

---------
### Progress

### Issue
 * [JDK-8309675](https://bugs.openjdk.org/browse/JDK-8309675): Generational ZGC: compiler/gcbarriers/UnsafeIntrinsicsTest.java fails in nmt_commit (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14382/head:pull/14382` \
`$ git checkout pull/14382`

`$ git checkout pull/14382` \
`$ git pull https://git.openjdk.org/jdk.git pull/14382/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14382`

`$ git pr show -t 14382`

</details>

<a href="https://git.openjdk.org/jdk/pull/14382.diff">https://git.openjdk.org/jdk/pull/14382.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14382#issuecomment-1583121013)

### DIFF
--- a/src/hotspot/share/gc/z/zPhysicalMemory.cpp
+++ b/src/hotspot/share/gc/z/zPhysicalMemory.cpp
@@ -283,15 +283,15 @@ void ZPhysicalMemoryManager::nmt_commit(zoffset offset, size_t size) const {
   // When this function is called we don't know where in the virtual memory
   // this physical memory will be mapped. So we fake that the virtual memory
   // address is the heap base + the given offset.
-  const zaddress addr = ZOffset::address(offset);
-  MemTracker::record_virtual_memory_commit((void*)untype(addr), size, CALLER_PC);
+  const uintptr_t addr = ZAddressHeapBase + untype(offset);
+  MemTracker::record_virtual_memory_commit((void*)addr, size, CALLER_PC);
 }
 
 void ZPhysicalMemoryManager::nmt_uncommit(zoffset offset, size_t size) const {
   if (MemTracker::enabled()) {
-    const zaddress addr = ZOffset::address(offset);
+    const uintptr_t addr = ZAddressHeapBase + untype(offset);
     Tracker tracker(Tracker::uncommit);
-    tracker.record((address)untype(addr), size);
+    tracker.record((address)addr, size);
   }
 }
 


stefank:
The reason why this only happens on this test is that -XX:+ZVerifyOops is only turned on in this test.

stefank:
To understand the fix see the function I'm replacing:
```
inline zaddress ZOffset::address(zoffset offset) {
  return to_zaddress(untype(offset) | ZAddressHeapBase);
}
```

and the verification code in this function:
```
inline zaddress to_zaddress(uintptr_t value) {
  const zaddress addr = zaddress(value);
  assert_is_valid(addr);
  DEBUG_ONLY(dereferenceable_test(addr));
  return addr;
}
```

The crash happens because of the dereferenceable_test, which checks that the memory have been committed and is accessible. As described in nmt_commit, we fake the address we report to NMT, so we can't try to dereference the contents of that address.

stefank:
Yes, that's a good suggestion, though I think I'd prefer to stick with the current solution because then it is more clear that the code matches the comment in nmt_commit.

Thanks for reviewing!

dcubed-ojdk (approved):
Thumbs up. Yet another disconnect between NMT virtual memory
modeling versus Z virtual memory modeling?

dcubed-ojdk:
Did you run the revised test on all Tier3 platforms?

stefank:
> Thumbs up. Yet another disconnect between NMT virtual memory modeling versus Z virtual memory modeling?

Thanks for reviewing. Yes, in away. It's the workaround for this disconnect that triggered our verification code, which in turn found our workaround with the fake addresses. :)

stefank:
> Did you run the revised test on all Tier3 platforms?

I'm running tier1-7 with Generational ZGC. There's just a few more tests running, so I'm on the verge of integrating this.

PR #14383, opened 2023-06-08
8309303: jdk/internal/misc/VM/RuntimeArguments test ignores jdk/internal/vm/options

mlchung:
The `RuntimeArguments` test verifies the VM arguments returned by `jdk.internal.misc.VM::getRuntimeArguments` as expected.    This test fails when running with GraalVM as it was created with `jlink --add-options` and the application will always be launched with that additional set of VM options.  This PR updates the test to read if 'jdk/internal/vm/options` is present and includle those options in the expected list.

### DIFF
--- a/test/jdk/jdk/internal/misc/VM/RuntimeArguments.java
+++ b/test/jdk/jdk/internal/misc/VM/RuntimeArguments.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2023, Oracle and/or its affiliates. All rights reserved.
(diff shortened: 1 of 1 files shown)


dougxc:
I can confirm the change works on GraalVM but it would make sense to also test it directly in the JDK by creating a test JDK image with jlink that includes VM options.

mlchung:
> I can confirm the change works on GraalVM but it would make sense to also test it directly in the JDK by creating a test JDK image with jlink that includes VM options.

Already exists.  `test/jdk/tools/jlink/plugins/AddOptionsPluginTest.java`

dougxc (approved):

AlanBateman (approved):

PR #14386, opened 2023-06-08
8309688: Data race on java.io.ClassCache$CacheRef.strongReferent

caoman:
Hi all,

Could anyone review this small fix for a data race in java.io.ClassCache$CacheRef? This fix makes the code safer by making the code data-race-free.

---------
### Progress
- [x] Change must be properly reviewed (2 reviews required, with at least 2 [Reviewers](https://openjdk.org/bylaws#reviewer))

### Issue
 * [JDK-8309688](https://bugs.openjdk.org/browse/JDK-8309688): Data race on java.io.ClassCache$CacheRef.strongReferent (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14386/head:pull/14386` \
`$ git checkout pull/14386`

`$ git checkout pull/14386` \
`$ git pull https://git.openjdk.org/jdk.git pull/14386/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14386`

`$ git pr show -t 14386`

</details>

<a href="https://git.openjdk.org/jdk/pull/14386.diff">https://git.openjdk.org/jdk/pull/14386.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14386#issuecomment-1583458423)


caoman:
The pre-submit failures for windows are due to https://bugs.openjdk.org/browse/CODETOOLS-7903482.

caoman:
The presubmit failure on linux-x86 looks unrelated to this PR. It is https://bugs.openjdk.org/browse/JDK-8309592.

AlanBateman:
> Could anyone review this small fix for a data race in java.io.ClassCache$CacheRef? This fix makes the code safer by making the code data-race-free.

If I read the existing code correctly, ClassCache::get can't return null. If there are two threads racing then it's okay for both to read a non-null strong ref, one read non-null and the other null, or both read null (if someone else cleared it). It just means there may be fallback to the soft ref or if that gets cleared then the remove/re-compute the ClassValue. Maybe more eyes are needed but it could be that this TSAN is reporting a false positive here.

RogerRiggs:
I think the race is benign also but the change maintains the essential semantics.
The use of AtomicRef will have a very slight performance impact.

The semantic is that at-least-one caller gets the reference before it can become "only" soft-referenced and might be reclaimed.

The use of AtomicReference changes the semantic to be exactly-one caller gets the initial hard reference; other callers will be able to get the value from the soft-reference as long as the first caller holds the reference and perhaps beyond depending on memory pressure.

ClassCaches are used by serialization to provide ObjectStreamClass descriptors.
It is quite possible that multiple threads may be serializing the same classes.

caoman:
Agree with the analysis that this race looks benign, and the slight change in semantics is OK. It is still a true data race and TSAN is NOT reporting a false positive though.

In our experience, the best practice is to avoid all data races unless there's a strong reason that fixing the race has undesirable effect (e.g. noticeable performance penalty). A benign race could become problematic in the future with unsuspecting changes to this file, and has higher maintenance cost in general.

I hope we can actually fix this data race. If we decide not to, at least add a comment about this intentionally benign data race.

RogerRiggs (approved):
Looks fine.

src/java.base/share/classes/java/io/ClassCache.java
```diff
@@ -39,24 +40,20 @@
 
     private static class CacheRef<T> extends SoftReference<T> {
         private final Class<?> type;
-        private T strongReferent;
+        private final AtomicReference<T> strongReferent;
```
shipilev, line 43: This introduces a dereference. Maybe it should be `AtomicReferenceFieldUpdater` or a `VarHandle`, assuming there are no bootstrapping issues with this code.

src/java.base/share/classes/java/io/ClassCache.java
```diff
@@ -87,9 +84,8 @@ T get(Class<?> cl) {
             // We might still have strong referent, and can return it.
             // This guarantees progress for at least one thread on every CacheRef.
             // Clear the strong referent before returning to make the cache soft.
-            T strongVal = ref.getStrong();
+            T strongVal = ref.getAndClearStrong();
```
shipilev, line 87: This is a cache, so IIRC, the prevailing case is when CacheRef already exists in the map, and its `strong` is already `null`. So this change introduces an atomic op, which is mostly useless on this path. So we pay a latency price on most `get()`-s without much benefit.

shipilev (changes requested):
I think this data race is benign, because `ClassValue` gives us the required memory semantics. Introducing additional performance hogs seems weird in this case, see below.

Since TSAN complains about the unsynchronized conflicting accesses in `getStrong()` and `clearStrong()`, maybe the better solution would be marking `strong` as `volatile`? It would still be awkward, but we would "only" pay a price of volatile load on a hot path.

caoman:
Thank you @shipilev! Good point about the cost of extra AtomicReference instance, and slowing down the fast path with a CAS.
Agree that making `strongReferent` `volatile` is better and simpler.

For AtomicReferenceFieldUpdater or VarHandle, they are more complicated and I don't see much benefit over a simple `volatile` in this case.

shipilev:
Unfortunately, even the single `volatile` on that fast path seems to cost quite a bit :(

On M1:

```
# CONF=macosx-aarch64-server-release make images test TEST=micro:java.io.ObjectStreamClasses

Benchmark                       Mode  Cnt    Score   Error  Units

# Before
ObjectStreamClasses.testLookup  avgt    8  106,946 ± 0,691  ns/op

# After
ObjectStreamClasses.testLookup  avgt    8  119,984 ± 0,588  ns/op
```

So again, this looks like a benign data race, and we are penalizing the serialization a bit even with the lightweight fix. This is only to make TSAN happy, right? If so, I would rather keep the code as is.

There are plenty of benign data races in JDK, we should be able to somehow tell TSAN which are actually fine.

caoman:
Thanks for the benchmarking. I was unaware of the microbenchmark `java.io.ObjectStreamClasses`.
The performance penalty is likely related to aarch64 being slower on volatile load. I see no observable difference on x86-64:
```
Benchmark                       Mode  Cnt    Score    Error  Units
# Before
ObjectStreamClasses.testLookup  avgt    8  230.255 ± 10.540  ns/op

# After
ObjectStreamClasses.testLookup  avgt    8  230.291 ± 19.369  ns/op
```

Anyway, this race does look benign and I'm OK with just documenting this benign race, and suppress it in TSAN. Updated this PR to a comment-only change.

src/java.base/share/classes/java/io/ClassCache.java
```diff
@@ -87,6 +87,10 @@ T get(Class<?> cl) {
             // We might still have strong referent, and can return it.
             // This guarantees progress for at least one thread on every CacheRef.
             // Clear the strong referent before returning to make the cache soft.
+            // A benign data race exists on CacheRef.strongReferent field if
+            // multiple threads calls get() with the same Class parameter.
+            // Fixing this race could introduce noticeable performance penalty.
+            // See the review thread for JDK-8309688 for details.
```
shipilev, line 93: Feels like this comment belongs near `strongReferent` declaration. I suggest:

```
// This field is deliberately accessed without sychronization. ClassValue
// provides synchronization when CacheRef is published. However, when
// a thread reads this field, while another thread is clearing the field, it 
// would formally constitute a data race. But that data race is benign, and
// fixing it could introduce noticeable performance penalty, see JDK-8309688.
private T strongReferent;
```

shipilev (approved):
Right. I have suggestions about the comment.

src/java.base/share/classes/java/io/ClassCache.java
caoman: Thank you. Updated.

shipilev (approved):
I am okay with it. Since this is my comment, I'd like someone else to re-review :)

RogerRiggs (approved):
Looks fine

caoman:
Thank you for the reviews and suggestions!

PR #14389, opened 2023-06-09
JDK-8308047  java/util/concurrent/ScheduledThreadPoolExecutor/BasicCancelTest.java timed out and also had jcmd pipe errors

viktorklang-ora:
The test used a too-small heap for generic testing, and coupled with looking at `freeMemory()` it led to a less-than-ideal determinism of execution of the test.

### DIFF
--- a/test/jdk/ProblemList-generational-zgc.txt
+++ b/test/jdk/ProblemList-generational-zgc.txt
@@ -38,5 +38,4 @@ sun/tools/jhsdb/heapconfig/JMapHeapConfigTest.java 8307393   generic-all
 sun/tools/jhsdb/HeapDumpTestWithActiveProcess.java 8307393   generic-all
 
 com/sun/jdi/ThreadMemoryLeakTest.java          8307402 generic-all
-java/util/concurrent/ScheduledThreadPoolExecutor/BasicCancelTest.java 8308047 windows-x64
 java/util/concurrent/locks/Lock/OOMEInAQS.java 8309218 generic-all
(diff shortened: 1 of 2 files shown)


stefank (commented):
Did you test if this significantly changed the execution time of the test? I'm wondering if it would make sense to change this to something smaller, say `maxMemory / 32`?

viktorklang-ora:
@stefank Good point. The runtime didn't meaningfully change on my own machine, but I've now decresed the iterations proportionally to the memory, and made sure that it still errors out (locally) if it doesn't GC the tasks.

stefank (approved):
I'm happy with the changes, but make sure to get a Review from one of the maintainers of this code area. Thanks.

viktorklang-ora:
@stefank :+1:

@AlanBateman Would you want to review this one? 🤔

AlanBateman (approved):
This looks okay to me and I assume it tested on with all GCs/configurations. If it comes back then maybe an alternative way to test this would be a timed loop of submit+task. If the cancel policy isn't working then it would throw OOME very quickly with a 64Mb heap.

viktorklang-ora:
@AlanBateman Yes, i've run it extensively. I decided against running it in a timed loop because that would technically make it succeed if the application thread didn't get enough CPU-time. (And makes it difficult to know how it behaves on very slow machines)

viktorklang-ora:
@AlanBateman Let me know if you're able to sponsor this one :+1:

PR #14394, opened 2023-06-09
8281658: Add a security category to the java -XshowSettings option

coffeys:
New functionality in the -XshowSettings menu to display relevant information about JDK security configuration

---------
### Progress
- [x] Change requires CSR request [JDK-8308768](https://bugs.openjdk.org/browse/JDK-8308768) to be approved

### Issues
 * [JDK-8281658](https://bugs.openjdk.org/browse/JDK-8281658): Add a security category to the java -XshowSettings option (**Enhancement** - P3)
 * [JDK-8308768](https://bugs.openjdk.org/browse/JDK-8308768): Add a security category to the java -XshowSettings option (**CSR**)

### Reviewers
 * [Roger Riggs](https://openjdk.org/census#rriggs) (@RogerRiggs - **Reviewer**) ⚠️ Review applies to [7e6f5090](https://git.openjdk.org/jdk/pull/14394/files/7e6f5090044fbaa0e1fd498f74831b007b78c49a)
 * [Sean Mullan](https://openjdk.org/census#mullan) (@seanjmullan - **Reviewer**) ⚠️ Review applies to [559f698e](https://git.openjdk.org/jdk/pull/14394/files/559f698e68a2a6d20350b70a77ddfe54baaf9d00)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14394/head:pull/14394` \
`$ git checkout pull/14394`

`$ git checkout pull/14394` \
`$ git pull https://git.openjdk.org/jdk.git pull/14394/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14394`

`$ git pr show -t 14394`

</details>

<a href="https://git.openjdk.org/jdk/pull/14394.diff">https://git.openjdk.org/jdk/pull/14394.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14394#issuecomment-1584630658)


src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -57,26 +57,22 @@
...
+import java.security.Security;
 import java.text.MessageFormat;
 import java.text.Normalizer;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Locale;
+import java.util.*;
```
RogerRiggs, line 65: Wild card imports are discouraged. (Your IDE may have done this behind your back)

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -312,12 +318,104 @@ private static void printLocales() {
...
                 ostream.println();
-                ostream.print(INDENT + INDENT);
+                ostream.print(TWOINDENT);
+            }
+        }
+        ostream.print("\n");
+    }
+
+    private static void printSecuritySettings(String arg) {
+        if (arg.toLowerCase(Locale.ROOT).equals("properties")) {
```
RogerRiggs, line 328: How about a string switch with "all" and default: being the catchall. It would be easier to add others and avoid multiple calls to `toLowerCase`.
It might be a good idea for unrecognized settings to print a 1-liner reminder of the valid values, including `all`

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -312,12 +318,104 @@ private static void printLocales() {
...
+            String val = p.getProperty(key);
+            if (val.contains(",") && val.length() > 60) {
+                // split lines longer than 60 chars which have multiple values
+                ostream.println(TWOINDENT + key + "=");
+                List.of(val.split(",")).forEach(s -> ostream.println(THREEINDENT + s.trim()));
+            } else {
+                ostream.println(TWOINDENT + key + "=" + val);
+            }
+        }
+        ostream.print("\n");
```
RogerRiggs, line 354: or just `ostream.println();`

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -312,12 +318,104 @@ private static void printLocales() {
...
+    }
+
+    private static void printSecurityTLSConfig() {
+        SSLSocket ssls;
+        try {
+            ssls = (SSLSocket)
+                    SSLContext.getDefault().getSocketFactory().createSocket();
+        } catch (IOException | NoSuchAlgorithmException e) {
+            throw new RuntimeException(e);
```
RogerRiggs, line 363: That's pretty rude of the runtime; perhaps `throw InternalError("Unable to create secure socket")`.
It would give some clue that doesn't look like a coding bug.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -312,12 +318,104 @@ private static void printLocales() {
...
+        ostream.println(TWOINDENT + "Enabled Protocols:");
+        for (String s : ssls.getEnabledProtocols()) {
+            System.out.println(THREEINDENT + s);
+        }
+
+        System.out.println("\n" + TWOINDENT + "Enabled Cipher Suites:");
+        for (String s : ssls.getEnabledCipherSuites()) {
+            System.out.println(THREEINDENT + s);
+        }
+        ostream.print("\n");
```
RogerRiggs, line 376: `ostream.println();`

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -312,12 +318,104 @@ private static void printLocales() {
...
+            ostream.println(TWOINDENT + "Provider services: (type : algorithm)");
+            for (Provider.Service ps : p.getServices().stream()
+                    .sorted(Comparator.comparing(Provider.Service::getType)).toList()) {
+                ostream.println(THREEINDENT + ps.getType() + " : " + ps.getAlgorithm());
+            }
+        }
+        ostream.print("\n");
+    }
+
+    // return a string split across multiple lines which aims to limit max width
```
RogerRiggs, line 393: This utility method might be usable in more places if the indentation was an argument instead of hard coded.
For example, in `printSecurityProperties`.  It might need to be able to split in more places.

src/java.base/share/classes/sun/launcher/resources/launcher.properties
```diff
@@ -168,6 +168,14 @@ java.launcher.X.usage=\n\
 \                      show all property settings and continue\n\
 \    -XshowSettings:vm\n\
 \                      show all vm related settings and continue\n\
+\    -XshowSettings:security\n\
```
RogerRiggs, line 171: `all` should probably be documented.

AlanBateman:
There is a lot of output and does seem to overwhelm the java -XshowSettings output with detail that I wouldn't expect by default.  It's effectively printing conf/java.security and enumerating all the capabilities of the providers and cipher suites. So I think I agree with Sean's suggestion to require the :security sub-option for the detail. We could potentially do this for the available locales too. The -XshowSettings output could include a message at the end about using -XshowSettings:detailed or some such sub-option for more detailed output.

In passing, it looks like some of the output is going to stderr, some to stdout, so you might want to check that. Also StringTokenizer is legacy, I don't think we want that in the java launcher code, it looks like it just needs String.split.

coffeys:
Thanks for the reviews to date. All good comments, I'll implement.

Yes, the output of `-XShowSettings` is more verbose. Let me try and re-implement with a new user option. Perhaps, we could have some of the key security settings show up with the `-XShowSettings` option (static provider names (I should list them in preference order actually) and enabled TLS protocols ?) and then have some help output hinting about how to get verbose output ?

seanjmullan:
Another possibility is to put this under a new option, ex `-XshowSecuritySettings`.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -312,12 +318,104 @@ private static void printLocales() {
...
+                ostream.print(TWOINDENT);
+            }
+        }
+        ostream.print("\n");
+    }
+
+    private static void printSecuritySettings(String arg) {
+        if (arg.toLowerCase(Locale.ROOT).equals("properties")) {
+            printSecurityProperties();
+        } else if(arg.toLowerCase(Locale.ROOT).equals("providers")) {
```
turbanoff, line 330:
```suggestion
        } else if (arg.toLowerCase(Locale.ROOT).equals("providers")) {
```

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -315,12 +332,136 @@ private static void printLocales() {
...
                 ostream.println();
-                ostream.print(INDENT + INDENT);
+                ostream.print(TWOINDENT);
+            }
+        }
+        ostream.println();
+    }
+
+    private static void printSecuritySettings(String arg) {
+        switch (arg.toLowerCase(Locale.ROOT)) {
```
RogerRiggs, line 342: Drop the conversion to lower case: none of the other categories allow "sloppy-case" inputs.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -172,6 +181,13 @@ static void showSettings(boolean printToStderr, String optionFlag,
             case "locale":
                 printLocale();
                 break;
+            case "security":
+                if (opts.length > 2) {
+                    printSecuritySettings(opts[2].trim());
+                } else {
+                    printSecuritySettings("all");
+                }
```
RogerRiggs, line 189: Perhaps a slight preference for:
```
            var opt = opts.length > 2 ? opts[2].trim() : "all";
            printSecuritySettings(opt);
```

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -315,12 +332,136 @@ private static void printLocales() {
...
+    private static void printSecurityProviderConfig(boolean verbose) {
+        ostream.println(INDENT + "Security provider static configuration:");
+        for (Provider p : Security.getProviders()) {
+            if (verbose) {
+                // separate the views out
+                ostream.println(TWOINDENT + "-".repeat(40));
+            }
+            ostream.println(TWOINDENT + "Provider name: " + p.getName());
+            if (verbose) {
+                ostream.println(TWOINDENT + PROV_INFO_STRING + wrappedString(p.getInfo(), 80));
```
RogerRiggs, line 422: Wrapping the args at 80 might still produce line that is quite long.
Perhaps wrap the string after it was concatenated.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
coffeys: fair point. done

src/java.base/share/classes/sun/launcher/LauncherHelper.java
coffeys: logic updated. I think that's what you were hinting at Roger.

coffeys (commented):
Thanks for the comments Roger. Changes pushed. Some other minor issues corrected also while passing through code.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -315,12 +329,136 @@ private static void printLocales() {
...
+                ostream.println(TWOINDENT + "-".repeat(40));
+            }
+            ostream.println(TWOINDENT + "Provider name: " + p.getName());
+            if (verbose) {
+                ostream.println(wrappedString(PROV_INFO_STRING + p.getInfo(), 80));
+                ostream.println(TWOINDENT + "Provider services: (type : algorithm)");
+                Set<Provider.Service> services = p.getServices();
+               if (!services.isEmpty()) {
+                   services.stream()
+                           .sorted(Comparator.comparing(Provider.Service::getType))
```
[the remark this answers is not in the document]
seanjmullan: +1 to sorting by algorithm name within same type.

ascarpino:
I would recommend using the service syntax that is similar to Preferred Provider and is also similar to a [services control](https://mail.openjdk.org/pipermail/security-dev/2023-May/035819.html) on the security mailing list that is not yet approved.  `Signature.SHA1withRSA` instead of `Signature : SHA1withRSA`

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -172,6 +181,10 @@ static void showSettings(boolean printToStderr, String optionFlag,
             case "locale":
                 printLocale();
                 break;
```
AlanBateman, line 183: printLocale might need the same treatment to avoid printing the list of all locales by default.

AlanBateman: Have you tried putting the new code into say sun.launcher.SecuritySettings to avoid adding so much security conf code into LauncherHelper? Maybe eventually this will need to a more pluggable as -XshowSettings has grown a bit beyond what it was originally for.

AlanBateman:
The output from -XshowSettings vs. the large dump from -XshowSettings:security look good, I think the default vs. all split is quite good.  The format/style of the output is a bit different to the other sections in the -XshowSettings output. The other sections are more like name = value. What you have is okay, it just looks a bit different.  There are a few blank lines in the the "Security settings summary" output that you might want to cull. Also "Use -XshowSettings:security for verbose details" might be better as "... for security details". 

Also what would you think about the earlier suggesting to do the same for the locale output? The dump of all available locales seems excessive to print out by default.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -315,12 +329,136 @@ private static void printLocales() {
...
+    private static void printSecurityProperties() {
+        ostream.println(INDENT + "Security properties:");
+        Properties p = SharedSecrets.getJavaSecurityPropertiesAccess().getInitialProperties();
+        for (String key : p.stringPropertyNames().stream().sorted().toList()) {
+            String val = p.getProperty(key);
+            if (val.contains(",") && val.length() > 60) {
+                // split lines longer than 60 chars which have multiple values
+                ostream.println(TWOINDENT + key + "=");
+                List.of(val.split(",")).forEach(
+                        s -> ostream.println(THREEINDENT + s.trim() + ","));
```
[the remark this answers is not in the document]
coffeys: yes, I was hoping it wouldn't be an issue! modified code.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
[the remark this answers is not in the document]
coffeys: good point. done.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -315,12 +329,136 @@ private static void printLocales() {
...
+                printAllSecurityConfig();
+            }
+
+        }
+    }
+
+    // A non-verbose description of some core security configuration settings
+    private static void printSecuritySummarySettings() {
+        ostream.println("Security settings summary: " + "\n" +
+                INDENT + "Use \"-XshowSettings:security\" for verbose details\n");
```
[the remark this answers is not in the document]
coffeys: I think it might be better to delegate to `java -X` help output actually. tweaked code in this area. (I could see the subcommand options becoming inaccurate if we add more such subcommands in the future). -X help output should be our reference

src/java.base/share/classes/sun/launcher/LauncherHelper.java
coffeys: Logged https://bugs.openjdk.org/browse/JDK-8310201 to track the Locale changes Alan.

Good suggestion on factoring out the security code to a helper class. Done.

coffeys:
Incorporated all comments to date. Took Tony's point on board. The provider output format is now `type.service`

the refactoring of code into new security helper class might make it harder to determine recent changes. I committed the main changes before the refactor operation. Might make review easier :

https://github.com/openjdk/jdk/pull/14394/commits/1af66eb6362c15a92ce03b0ab8667d6d3a0cb673

Need to update details in CSR also.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -128,7 +129,7 @@ private static class ResourceBundleHolder {
         private static final ResourceBundle RB =
                 ResourceBundle.getBundle(defaultBundleName);
     }
-    private static PrintStream ostream;
+    static PrintStream ostream;
```
AlanBateman, line 132: I don't think we want SecuritySettings accessing LauncherHelper's field like this. I assume printSecuritySetting is just missing a PrintStream parameter so that it can be called to print the security settings to a given stream.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -267,7 +273,7 @@ private static void printPropertyValue(String key, String value) {
                 ostream.println(s);
                 first = false;
             } else { // following lines prefix with indents
-                ostream.println(INDENT + INDENT + s);
+                ostream.println(TWOINDENT + s);
```
AlanBateman, line 276: I assume these changes are per reverted now.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
AlanBateman: > Logged https://bugs.openjdk.org/browse/JDK-8310201 to track the Locale changes Alan.

Thanks, probably should treat this as a bug for JDK 21 so that a new addition doesn't change behavior between 21 and 22.

src/java.base/share/classes/sun/launcher/SecuritySettings.java
```diff
@@ -0,0 +1,207 @@
...
+        ostream = stream;
+        switch (arg) {
+            case "properties" -> printSecurityProperties();
+            case "providers"  -> printSecurityProviderConfig(true);
+            case "tls"        -> printSecurityTLSConfig(true);
+            case "all"        -> printAllSecurityConfig();
+            default           -> {
+                ostream.println("Unrecognized security subcommand. See \"java -X\" for help");
+                ostream.println("Printing all security settings");
+                printAllSecurityConfig();
```
RogerRiggs, line 66: The error message is going to get lost in the volume of settings.
Allowing bad command input reinforces learning the wrong suboption; though it may duplicate the help, I'd print the allowed options.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
coffeys: @AlanBateman  - I was planning to keep this trivial change. Does it not read better ?

src/java.base/share/classes/sun/launcher/SecuritySettings.java
coffeys: @RogerRiggs  - do you mean to print nothing in the "bad command input" scenario ? The current -XshowSettings launch behaviour prints all data if a bad value is passed to it. I was mimicking this for security subcommands. 

Are you suggesting to print the relevant part of the "java -X" help menu here or is restoring the message with value subcommands sufficient ?

src/java.base/share/classes/sun/launcher/SecuritySettings.java
```diff
@@ -0,0 +1,207 @@
...
+        printSecurityProviderConfig(true);
+        printSecurityTLSConfig(true);
+    }
+
+    private static void printSecurityProperties() {
+        ostream.println(INDENT + "Security properties:");
+        Properties p = SharedSecrets.getJavaSecurityPropertiesAccess().getInitialProperties();
+        for (String key : p.stringPropertyNames().stream().sorted().toList()) {
+            String val = p.getProperty(key);
+            if (val.contains(",") && val.length() > 60) {
```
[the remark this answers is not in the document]
coffeys: fair point. done.

src/java.base/share/classes/sun/launcher/SecuritySettings.java
RogerRiggs: Restoring the message with value subcommands is what I would prefer.
I do disagree with the original choice of printing everything if the command is wrong and would not propagate that behavior to the new command.  
Separately, I would consider changing the existing behavior but I suspect that would fail to achieve consensus based on compatibility concerns.

coffeys: Fair point. Modified the code so that we don't print any security settings if a bad security subcommand value is specified. Info message instead. I'll update the CSR on that point also.

Agreed on the compatibility concerns of changing the existing behaviour already present in this area.

AlanBateman:
I looked through the current version (7e6f5090) and you've got it to a good place, the launcher changes look fine.

The changes make me wonder if `-XshowSetting:aardvark` should be an error rather than default to print all settings. Something we should look at again.  Same thing for `-XshowSettings:system` on non-Linux, probably should have been a bit more discussion when that was added as its more about the container environment.

src/java.base/share/classes/sun/launcher/SecuritySettings.java
```diff
@@ -0,0 +1,213 @@
...
+    private static void printSecurityTLSConfig(boolean verbose) {
+        SSLSocket ssls;
+        try {
+            ssls = (SSLSocket)
+                    SSLContext.getDefault().getSocketFactory().createSocket();
+        } catch (IOException | NoSuchAlgorithmException e) {
+            throw new InternalError("Failed to create SSL socket");
+        }
+
+        ostream.println(INDENT + "Security TLS configuration:");
```
seanjmullan, line 123: What about also noting the name of the TLS/JSSE provider in this line, for example:

"Security TLS configuration (SunJSSE provider):"

This would be useful information if a customer is using a 3rd party JSSE provider, and it is selected before SunJSSE, as the defaults in that case may be different.

mlchung:
> The changes make me wonder if `-XshowSetting:aardvark` should be an error rather than default to print all settings. Something we should look at again. Same thing for `-XshowSettings:system` on non-Linux, probably should have been a bit more discussion when that was added as its more about the container environment.

I think throwing an error for an invalid suboption makes sense.   `-XshowSetting` without the suboption is default to print all settings, i.e. equivalent to `-XshowSetting:all`.   But the switch statment is missing the `all` case.   I think we should fix this and error for any invalid option.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -112,7 +112,7 @@ private LauncherHelper() {}
 
     private static StringBuilder outBuf = new StringBuilder();
 
-    private static final String INDENT = "    ";
+    private static final String INDENT = " ".repeat(4);
```
RogerRiggs, line 115: Revert this, its just making a simple constant into a runtime expression.

RogerRiggs (approved):

src/java.base/share/classes/sun/launcher/SecuritySettings.java
coffeys: good suggestion Sean. Patch updated.

coffeys:
> > The changes make me wonder if `-XshowSetting:aardvark` should be an error rather than default to print all settings. Something we should look at again. Same thing for `-XshowSettings:system` on non-Linux, probably should have been a bit more discussion when that was added as its more about the container environment.
> 
> I think throwing an error for an invalid suboption makes sense. `-XshowSetting` without the suboption is default to print all settings, i.e. equivalent to `-XshowSetting:all`. But the switch statment is missing the `all` case. I think we should fix this and error for any invalid option.

fair point Mandy - I hope to change such behavior in a new follow on patch that I'll deliver shortly: JDK-8311653

src/java.base/share/classes/sun/launcher/LauncherHelper.java
coffeys: Thanks Roger. Fair point. Done.

src/java.base/share/classes/sun/launcher/SecuritySettings.java
coffeys: I think we need to harden this area of code actually. Some JDK configurations may not have an SSL provider.

e.g. (removing JSSE from config)   

```
    See "java -X" for verbose security settings options
    Security provider static configuration: (in order of preference)
        Provider name: SUN
        Provider name: SunRsaSign
Exception in thread "main" java.lang.InternalError: Failed to create SSL socket
	at java.base/sun.launcher.SecuritySettings.printSecurityTLSConfig(SecuritySettings.java:121)
	at java.base/sun.launcher.SecuritySettings.printSecuritySummarySettings(SecuritySettings.java:75)
	at java.base/sun.launcher.LauncherHelper.showSettings(LauncherHelper.java:188)

```

seanjmullan (approved):

PR #14395, opened 2023-06-09
8308184: Launching java with large number of jars in classpath with java.protocol.handler.pkgs system property set can lead to StackOverflowError

jaikiran:
Can I please get a review of this change which proposes to fix the issue noted in https://bugs.openjdk.org/browse/JDK-8308184?

When an application is launched, the `app` classloader internally uses a `jdk.internal.loader.URLClassPath` to find and load the main class being launched. The `URLClassPath` uses a list of classpath entries to find resources. Depending on the classpath entry, the `URLClassPath` will use a relevant loader. A couple of such loaders are `URLClassPath$FileLoader` (for loading resources from directories) and `URLClassPath$JarLoader` (for loading resources from jar files).

`JarLoader` creates instances of `java.net.URL` to represent the jar file being loaded. `java.net.URL` uses protocol specific `java.net.URLStreamHandler` instance to handle connections to those URLs. When constructing an instance of `URL`, callers can pass a protocol handler. If it is not passed then the `URL` class looks for protocol handlers that might have been configured by the application. The `java.protocol.handler.pkgs` system property is the one which allows overriding the protocol handlers (even for the `jar` protocol). When this property is set, the `URL` class triggers lookup and classloading of the protocol handler classes.

The issue that is reported is triggered when the `java.protocol.handler.pkgs` system property is set and the classpath has too many jar files. `app` classloader triggers lookup of the main class and the `URLClassPath` picks up the first entry in the classpath and uses a `JarLoader` (in this example our classpath entries have a jar file at the beginning of the list). The `JarLoader` instantiates a `java.net.URL`, which notices that the `java.protocol.handler.pkgs` is set, so it now triggers lookup of a (different) class using the same classloader and thus the same `URLClassPath`. The `URLClassPath` picks the next classpath entry and then calls into the `URL` again through the `JarLoader`. This sequence ends up being re-entrant calls and given the large number of classpath entries, these re-entrant calls end up with a `StackOverflowError` as shown in the linked JBS issue.

The commit in this PR fixes this issue by using the system provided protocol handler implementation of the `jar` protocol in the `app` classloader. This results in the `URL` instances created through the `JarLoader` to use this specific handler instance. This allows the `app` classloader which is responsible for loading the application's main class to reliably use the system provided protocol handler.

Do note that this doesn't in any way impact the ability of applications to override the protocol handler for `jar` protocol. Applications can still continue to do it by setting the `java.protocol.handler.pkgs` system property to configure their application specific implementation for the `jar` protocol (or other protocols). `URL` instances created by the application code will continue to use their overridden `jar` protocol handler. The commit in this PR merely fixes the jar protocol handler that is used by the `app` classloader.

A new jtreg test has been added which reproduces this issue and verifies the fix. tier testing is in progress.

---------
### Progress

### Issue
 * [JDK-8308184](https://bugs.openjdk.org/browse/JDK-8308184): Launching java with large number of jars in classpath with java.protocol.handler.pkgs system property set can lead to StackOverflowError (**Bug** - P4)

### Reviewers
 * [Mandy Chung](https://openjdk.org/census#mchung) (@mlchung - **Reviewer**) ⚠️ Review applies to [b1260557](https://git.openjdk.org/jdk/pull/14395/files/b126055761b352b07794c9368aa958965ac077b4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14395/head:pull/14395` \
`$ git checkout pull/14395`

`$ git checkout pull/14395` \
`$ git pull https://git.openjdk.org/jdk.git pull/14395/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14395`

`$ git pr show -t 14395`

</details>

<a href="https://git.openjdk.org/jdk/pull/14395.diff">https://git.openjdk.org/jdk/pull/14395.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14395#issuecomment-1584736031)


src/java.base/share/classes/jdk/internal/loader/URLClassPath.java
```diff
@@ -177,12 +177,13 @@ public URLClassPath(URL[] urls, @SuppressWarnings("removal") AccessControlContex
...
      *
      * @param cp the class path string
+     * @param jarHandler the {@link URLStreamHandler} for {@code jar} protocol
      * @param skipEmptyElements indicates if empty elements are ignored or
      *        treated as the current working directory
      *
      * @apiNote Used to create the application class path.
      */
-    URLClassPath(String cp, boolean skipEmptyElements) {
+    URLClassPath(String cp, URLStreamHandler jarHandler, boolean skipEmptyElements) {
```
AlanBateman, line 186: This is a special constructor for the application class path so shouldn't have the jarHandle parameter. This should allow you to drop the changes to ClassLoaders.

src/java.base/share/classes/jdk/internal/loader/URLClassPath.java
```diff
@@ -209,7 +210,10 @@ public URLClassPath(URL[] urls, @SuppressWarnings("removal") AccessControlContex
 
         this.unopenedUrls = unopenedUrls;
         this.path = path;
-        this.jarHandler = null;
+        // this URLClassPath constructor is solely for the app classloader, for which we use
+        // the system provided jar protocol handler implementation for loading jars
+        // in the classpath
```
AlanBateman, line 215: The apiNote on the constructor makes it clear that the constructor is for the application class path.  Also just to say that the term used in the URL docs is "built-in protocol handler". Combined then it might be clearer if the comment says that the application class loader must use the built-in protocol handler.

src/java.base/share/classes/jdk/internal/loader/URLClassPath.java
```diff
@@ -209,7 +210,10 @@ public URLClassPath(URL[] urls, @SuppressWarnings("removal") AccessControlContex
...
+        this.jarHandler = new Handler();
```
AlanBateman, line 216: I wouldn't normally suggest dropping an import and using a qualified class name but "Handler" is a very generic class name, replacing it with "new sun.net.www.protocol.jar.Handler()" might be clearer. Another approach would be to move the registry/lookup of the built-in protocol handler from URL to a support class so it could be used in several places but that that might be more trouble that it is worth so what you have is okay.

test/jdk/java/net/URL/HandlersPkgPrefix/LargeClasspathWithPkgPrefix.java
```diff
@@ -0,0 +1,137 @@
...
+    private static void createJar(Path p) throws Exception {
+        JarBuilder jb = new JarBuilder(p.toString());
+        jb.addEntry("foobar.txt", "foobar".getBytes());
+        jb.build();
+        System.out.println("Created jar at " + p);
+    }
+
+    // javac -d <destDir> <javaFile>
+    private static void compile(Path javaFile, Path destDir) throws Exception {
+        String javacPath = JDKToolFinder.getJDKTool("javac");
```
mlchung, line 115: FYI.   `jdk.test.lib.compiler.CompilerUtils` can be used to compile classes in process.

mlchung (approved):
Using the built-in jar protocol handler looks right.   I like Alan's suggestion to use the fully qualified name in the source to make it clearer.

src/java.base/share/classes/jdk/internal/loader/URLClassPath.java
jaikiran: > I wouldn't normally suggest dropping an import and using a qualified class name but "Handler" is a very generic class name, replacing it with "new sun.net.www.protocol.jar.Handler()" might be clearer.

I agree. I have now updated the PR to use the fully qualified name at the call site and also adjusted the code comment as recommended.

test/jdk/java/net/URL/HandlersPkgPrefix/LargeClasspathWithPkgPrefix.java
jaikiran: Thank you Mandy, this is useful, I wasn't aware of this one. I have now updated the test to use this utility class.

test/jdk/java/net/URL/HandlersPkgPrefix/LargeClasspathWithPkgPrefix.java
```diff
@@ -0,0 +1,140 @@
...
+ * @test
+ * @bug 8308184
+ * @summary Verify that an application can be launched when the classpath contains large number of
+ *          jars and the java.protocol.handler.pkgs system property is set
+ * @library /test/lib/
+ * @build jdk.test.lib.util.JarBuilder jdk.test.lib.compiler.CompilerUtils
+ *        jdk.test.lib.process.ProcessTools
+ * @run driver LargeClasspathWithPkgPrefix
+ */
+public class LargeClasspathWithPkgPrefix {
```
AlanBateman, line 44: test/jdk/java/net/URL/HandlersPkgPrefix is an unusual location for the test, maybe it could move to test/jdk/sun/misc/URLClassPath so it's the same location as the other tests for URLClassPath. Separately (not suggesting it for this PR) is that test directory should probably be renamed to jdk/internal/loader as the URLClassPath moved in JDK 9.

src/java.base/share/classes/jdk/internal/loader/URLClassPath.java
```diff
@@ -209,7 +207,9 @@ public URLClassPath(URL[] urls, @SuppressWarnings("removal") AccessControlContex
 
         this.unopenedUrls = unopenedUrls;
         this.path = path;
-        this.jarHandler = null;
+        // the application class loader uses the built-in protocol handler to avoid protocol
+        // handler lookup when opening JAR files on the class path.
+        this.jarHandler = new sun.net.www.protocol.jar.Handler();
```
AlanBateman, line 212: Good, this looks much better.

AlanBateman (approved):
Thanks for taking this one. My guess is that this issue goes back 20+ years but not noticed because it's running with the system property to specify another location for protocol handlers is probably rare, and it seems a scanning of a large large class path to trigger it.

I assume you'll see the comment about moving the test to be with the other tests for URLClassPath.

jaikiran:
Hello Alan,

> Thanks for taking this one. My guess is that this issue goes back 20+ years but not noticed because it's running with the system property to specify another location for protocol handlers is probably rare, and it seems a scanning of a large large class path to trigger it.

When I started looking into this issue, I was curious why this wasn't reported in Java 8 since the reporter was merely switching Java version from 8 to a newer version. When I look into Java 8 code, I see that the `sun.misc.Launcher` in Java 8 specifically deals with this case by using the built-in protocol handler and uses the bootstrap classloader for loading that hander, while launching the application's main class:
https://github.com/openjdk/jdk8u-dev/blob/master/jdk/src/share/classes/sun/misc/Launcher.java#L53
https://github.com/openjdk/jdk8u-dev/blob/master/jdk/src/share/classes/sun/misc/Launcher.java#L512

I think that explains why it isn't seen in that version.

jaikiran:
> I assume you'll see the comment about moving the test to be with the other tests for URLClassPath.

I have updated the PR to change the location of this new test class. Additionally I've filed https://bugs.openjdk.org/browse/JDK-8309763 to move the `sun/misc/URLClassPath` tests to `jdk/internal/loader` directory.

AlanBateman:
> When I started looking into this issue, I was curious why this wasn't reported in Java 8 since the reporter was merely switching Java version from 8 to a newer version. When I look into Java 8 code, I see that the `sun.misc.Launcher` in Java 8 specifically deals with this case by using the built-in protocol handler

Thanks for the digging into that, I was curious why this hasn't come up before now. I suppose part of it is that it's rare to run with this system property to override or add URL protocol handlers. There are a number of JDK 1.0/1.1 era APIs that used this approach for pluggability. In time it may be possible to deprecate this mechanism, newer code use services and URLStreamHandlerProvider.

Anyway, it's good that you've tracked this down, thanks for spending time on it.

AlanBateman (approved):

jaikiran:
Thank you Alan and Mandy for the reviews. tier testing came back fine. I'll go ahead and merge this.

PR #14396, opened 2023-06-09
8309685: Fix -Wconversion warnings in assembler and register code

coleenp:
This change widens some int arguments, narrows some int returns and adds some casts and checked_casts where the compiler was doing the implicit conversion.  I maintained existing types - ie if the parameter was short, I used a cast to short.  Also int is used in places that might be better served as unsigned int but I didn't fix that because it would be too large and risky.  The registers encode an offset in an array, so it's safe to checked_cast<> to get their encoding. This fix is limited so that the types changed and casts added are intentional.
See CR for counts of -Wconversion warnings this resolves.
Tested with tier1-7, also tested with tier1 on all Oracle supported platforms.

---------
### Progress

### Issue
 * [JDK-8309685](https://bugs.openjdk.org/browse/JDK-8309685): Fix -Wconversion warnings in assembler and register code (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14396/head:pull/14396` \
`$ git checkout pull/14396`

`$ git checkout pull/14396` \
`$ git pull https://git.openjdk.org/jdk.git pull/14396/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14396`

`$ git pr show -t 14396`

</details>

<a href="https://git.openjdk.org/jdk/pull/14396.diff">https://git.openjdk.org/jdk/pull/14396.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14396#issuecomment-1587231846)


src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -2035,10 +2035,10 @@ void MacroAssembler::post_call_nop() {
...
   relocate(post_call_nop_Relocation::spec());
   InlineSkippedInstructionsCounter skipCounter(this);
-  emit_int8((int8_t)0x0f);
-  emit_int8((int8_t)0x1f);
-  emit_int8((int8_t)0x84);
-  emit_int8((int8_t)0x00);
+  emit_int8((uint8_t)0x0f);
+  emit_int8((uint8_t)0x1f);
+  emit_int8((uint8_t)0x84);
+  emit_int8((uint8_t)0x00);
```
jdksjolen, line 2041: Are these necessary? Signature is `void emit_int8(   uint8_t x1)`, so the compiler should figure out that they fit without the cast.

src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -2047,11 +2047,11 @@ void MacroAssembler::fat_nop() {
...
-    emit_int8((int8_t)0x26); // es:
-    emit_int8((int8_t)0x2e); // cs:
-    emit_int8((int8_t)0x64); // fs:
-    emit_int8((int8_t)0x65); // gs:
-    emit_int8((int8_t)0x90);
+    emit_int8((uint8_t)0x26); // es:
+    emit_int8((uint8_t)0x2e); // cs:
+    emit_int8((uint8_t)0x64); // fs:
+    emit_int8((uint8_t)0x65); // gs:
+    emit_int8((uint8_t)0x90);
```
jdksjolen, line 2054: Same here

src/hotspot/share/asm/assembler.hpp
```diff
@@ -298,15 +298,17 @@ class AbstractAssembler : public ResourceObj  {
   // ensure buf contains all code (call this before using/copying the code)
   void flush();
 
-  void emit_int8(   uint8_t x1)                                     { code_section()->emit_int8(x1); }
+  // These take full width arguments and cast down without checking. Many callers have sign extension converting up to
+  // full int width arguments, so this casts this away.
+  void emit_int8(   uint32_t x1)                                    { code_section()->emit_int8((uint8_t)x1); }
```
kimbarrett, line 303: It is unfortunate that emit_int8 has that name, rather than emit_uint8.
But that's probably water under the bridge; certainly so for this change.

It is also unfortunate that integral promotion and -Wconversion fight each
other pretty hard.  And implicit conversions make it easy for -Wconversion
failure cases to slip in while we don't have it enabled as part of normal
operations. Here's an alternative approach:

```
template<typename T, ENABLE_IF(std::is_same<T, uint8_t>::value)>
void emit_int8(T v) { ... }

template<typename T,
         ENABLE_IF(std::is_integral<T>::value),
         ENABLE_IF(!std::is_same<T, uint8_t>::value)>
void emit_int8(T v) { emit_int8(checked_cast<uint8_t>(v)); }
```

Maybe that's not so bad?

src/hotspot/share/code/vmreg.cpp
```diff
@@ -30,7 +30,7 @@
 // used by SA and jvmti, but it's a leaky abstraction: SA and jvmti
 // "know" that stack0 is an integer masquerading as a pointer. For the
 // sake of those clients, we preserve this interface.
-VMReg VMRegImpl::stack0 = (VMReg)VMRegImpl::stack_0()->value();
+VMReg VMRegImpl::stack0 = (VMReg)(intptr_t)VMRegImpl::stack_0()->value();
```
kimbarrett, line 33: So enlighten me - why the two-step cast?

kimbarrett (changes requested):
This is smaller and less painful than I expected.

src/hotspot/share/asm/assembler.hpp
coleenp: Thanks for the suggested change.  The sign extension for integral promotion was a problem.  The second emit_int8 will still assert for the callers (there are still some) that pass a type that gets sign extended to int to emit_in8.  I can see how many are left once the first function is added.  Maybe it's possible to fix them all.  We'll see.

src/hotspot/share/code/vmreg.cpp
coleenp: I changed value() to return an int rather than intptr_t, so the compiler objects:

src/hotspot/share/code/vmreg.cpp:33:27: error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]
   33 | VMReg VMRegImpl::stack0 = (VMReg)VMRegImpl::stack_0()->value();
      |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

coleenp (commented):
Thanks for reviewing this Kim.  I tried to limit the changes so that it's not just adding casts.  I'm glad this isn't as bad as you thought it would be.

src/hotspot/share/asm/assembler.hpp
coleenp: src/hotspot/cpu/x86/assembler_x86.cpp|4421 col 12| error: no matching function for call to 'Assembler::emit_int8(Assembler::ComparisonPredicate&)'                                                                                                                                          
||  4421 |   emit_int8(vcc);
||       |   ~~~~~~~~~^~~~~
|| In file included from /scratch/cphillim/hg/21more-conversion/src/hotspot/cpu/x86/assembler_x86.cpp:26:

But I made ComparisonPredicate enum a uint8_t.  Why didn't it match it?

  // Comparison predicates for integral types & FP types when using SSE
  enum ComparisonPredicate : uint8_t {

There are a bunch of these that don't like the templates.

coleenp: Taking out the ENABLE_IF<is_integral> at least allows it to compile but there are lots of crashes because of sign extension.

Maybe we can make checked_cast<> tolerant of that somehow?

```template <typename T2, typename T1>
constexpr T2 checked_cast(T1 thing) {
  T2 result = static_cast<T2>(thing);
  assert(static_cast<T1>(result) == thing, "must be");
  return result;
}```

Comment (coleenp) [reply src/hotspot/share/asm/assembler.hpp:303]: How about something like this?
```  template<typename T, ENABLE_IF(std::is_same<T, uint8_t>::value)>
  void emit_int8(T v) { code_section()->emit_int8(v); }

  template<typename T, ENABLE_IF(!std::is_same<T, uint8_t>::value)>
  void emit_int8(T v) {
    if (v < 0) {
      int8_t tmp = static_cast<int8_t>(v);
      assert(static_cast<T>(v) == tmp, "signed cast preserves value");
      emit_int8(static_cast<uint8_t>(v));
    } else {
      emit_int8(checked_cast<uint8_t>(v)); 
    }
  }
```

zifeihan:
Hi, may I ask if it's OK to add fix for RISC-V please? I checked and found that this port has similar issue.

zifeihan:
Hi, may I ask if it's OK to add fix for RISC-V please? I checked and found that this port has similar issue. 
```diff 
diff --git a/src/hotspot/cpu/riscv/register_riscv.hpp b/src/hotspot/cpu/riscv/register_riscv.hpp
index 7b547335cd0..7d25875fedd 100644
--- a/src/hotspot/cpu/riscv/register_riscv.hpp
+++ b/src/hotspot/cpu/riscv/register_riscv.hpp
@@ -70,7 +70,7 @@ class Register {
 
    public:
     // accessors
-    constexpr int raw_encoding() const { return this - first(); }
+    constexpr int raw_encoding() const { return checked_cast<int>(this - first()); }
     constexpr int     encoding() const { assert(is_valid(), "invalid register"); return raw_encoding(); }
     constexpr bool    is_valid() const { return 0 <= raw_encoding() && raw_encoding() < number_of_registers; }
 
@@ -187,7 +187,7 @@ class FloatRegister {
 
    public:
     // accessors
-    constexpr int raw_encoding() const { return this - first(); }
+    constexpr int raw_encoding() const { return checked_cast<int>(this - first()); }
     constexpr int     encoding() const { assert(is_valid(), "invalid register"); return raw_encoding(); }
     constexpr bool    is_valid() const { return 0 <= raw_encoding() && raw_encoding() < number_of_registers; }
 
@@ -297,7 +297,7 @@ class VectorRegister {
 
    public:
     // accessors
-    constexpr int raw_encoding() const { return this - first(); }
+    constexpr int raw_encoding() const { return checked_cast<int>(this - first()); }
     constexpr int     encoding() const { assert(is_valid(), "invalid register"); return raw_encoding(); }
     constexpr bool    is_valid() const { return 0 <= raw_encoding() && raw_encoding() < number_of_registers; }
```

coleenp:
Thanks @zifeihan I added riscv too.

src/hotspot/share/utilities/globalDefinitions.hpp
```diff
@@ -512,15 +513,20 @@ inline size_t pointer_delta(const MetaWord* left, const MetaWord* right) {
...
+constexpr T2 checked_cast(T1 thing) {
+#ifdef ASSERT
+  typedef typename std::make_signed<T1>::type S1;
+  S1 test = static_cast<S1>(thing);
+  S1 max = std::numeric_limits<T2>::max();
+  S1 min = (std::is_signed<T2>::value) ? std::numeric_limits<T2>::min() : -max;
+  assert(test <= max && test >= min, "out of range of destination type");
+#endif
+  return static_cast<T2>(thing);
 }
```
dean-long, line 530: I don't understand this change.  Isn't it confusing to treat unsigned types as signed here?  Unsigned types don't sign extend.  If emit_intXX can take signed values then shouldn't that be reflected in the types of the arguements? 
Also, -max should be -max - 1 for 2's complement, right?

coleenp: The conversion from an unsigned type to a smaller unsigned type should have just worked if you convert back, like the old checked_cast code did, but callers of emit_intxxx were sign extending so the conversion back didn't work.

Calls like emit_int8() are passed arguments like 0xf0, but there were multiple (~940) that are passed with something that did some bitwise calculation that was promoted up to int.

Maybe the emit_intxx functions should take signed int, and then I could just check range in the checked_cast() without converting the argument to signed.  That might work and still provide the range checking that we want.

Yes, it should be max-1.  Thank you for looking at this.

coleenp: Changing checked_cast<> to not convert to a signed type for T1 leads to very puzzling compiler errors.  For these emit_int8 etc functions, maybe a straight cast is the best thing, and leaving the definition checked_cast<> alone.  I'd be happy to do that as I had in my first patch.

dean-long: I don't know what checked_cast<> should do for negative values when going between signed --> unsigned and for values greater than signed max for unsigned --> signed.  It seems like both should fail because the sign changes.

coleenp:
I'm going to break this up more.

PR #14398, opened 2023-06-09
8307907: [ppc] Remove RTM locking implementation

TheRealMDoerr:
This PR removes RTM code for PPC64. The instructions are not supported by Power10 processors.
The situation is even worse in Power9 compatibility mode on Power10. The VM detects RTM to be available, but the processor is unable to complete any transaction successfully.
The feature had so many problems and we never supported it in production.


MBaesken (approved):
LGTM.
Maybe a release note would be nice because the change removed product flags.

TheRealMDoerr:
Thanks for the reviews! The flags were copied from x86, but they are officially only supported on x86: "Options related to RTM are available only for the Java HotSpot Server VM on x86 CPUs that support Transactional Synchronization Extensions (TSX)." [https://docs.oracle.com/en/java/javase/11/tools/java.html#GUID-3B1CE181-CD30-4178-9602-230B800D4FAE]

src/hotspot/cpu/ppc/macroAssembler_ppc.cpp
```diff
@@ -2644,14 +2196,6 @@ void MacroAssembler::compiler_fast_lock_object(ConditionRegister flag, Register
     bne(flag, failure);
   }
 
-#if INCLUDE_RTM_OPT
-  if (UseRTMForStackLocks && use_rtm) {
```
reinrich, line 2648: Can RTM related method parameters (e.g. `use_rtm`) be removed?

src/hotspot/cpu/ppc/macroAssembler_ppc.cpp
```diff
@@ -2769,19 +2300,6 @@ void MacroAssembler::compiler_fast_unlock_object(ConditionRegister flag, Registe
   assert(LockingMode != LM_LIGHTWEIGHT || flag == CCR0, "bad condition register");
   Label success, failure, object_has_monitor, notRecursive;
 
-#if INCLUDE_RTM_OPT
-  if (UseRTMForStackLocks && use_rtm) {
```
reinrich, line 2773: Can RTM related method parameters (e.g. `use_rtm`) be removed?

reinrich (commented):
Looks good (except 2 nits). I'm glad we get rid of this complex code. Thanks!

src/hotspot/cpu/ppc/macroAssembler_ppc.cpp
TheRealMDoerr: Good catch! There are even more parameters which are no longer needed. And I've found some more leftovers. I just removed all of them.

reinrich (approved):
Looks good!
Thanks, Richard.

src/hotspot/cpu/ppc/vm_version_ppc.cpp
tstuefe, line 1: I think you can remove PPC_FEATURE2_HTM_NOSC too, right?

TheRealMDoerr:
@tstuefe: Thanks for reviewing! I've removed all you requested (and even more). I decided to remove the instruction emitters, too. We don't need them any more.

tstuefe (approved):
This is a nice complexity reduction!

All good now.

TheRealMDoerr:
Thanks!
/integrate

PR #14399, opened 2023-06-09
8308609: java/lang/ScopedValue/StressStackOverflow.java fails with "-XX:-VMContinuations"

theRealAph:
Move `runWith()` from `VirtualThread` to `BaseVirtualThread`.

`BoundVirtualThread` does not use `runWith()` to run its task, so when a VM error occurs it can not recover scoped values.

Moving `runWith()` into the common subclass of both `VirtualThread` and `BoundVirtualThread` fixes the problem.

---------
### Progress

### Issue
 * [JDK-8308609](https://bugs.openjdk.org/browse/JDK-8308609): java/lang/ScopedValue/StressStackOverflow.java fails with "-XX:-VMContinuations" (**Bug** - P4)

### Reviewers
 * [Alan Bateman](https://openjdk.org/census#alanb) (@AlanBateman - **Reviewer**) ⚠️ Review applies to [b2663471](https://git.openjdk.org/jdk/pull/14399/files/b2663471519b031139fc9224d0fa1cc2955712e8)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14399/head:pull/14399` \
`$ git checkout pull/14399`

`$ git checkout pull/14399` \
`$ git pull https://git.openjdk.org/jdk.git pull/14399/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14399`

`$ git pr show -t 14399`

</details>

<a href="https://git.openjdk.org/jdk/pull/14399.diff">https://git.openjdk.org/jdk/pull/14399.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14399#issuecomment-1584890249)

### DIFF
--- a/src/java.base/share/classes/java/lang/BaseVirtualThread.java
+++ b/src/java.base/share/classes/java/lang/BaseVirtualThread.java
@@ -24,11 +24,6 @@
  */
 package java.lang;
 
-import jdk.internal.vm.annotation.ForceInline;
-import jdk.internal.vm.annotation.Hidden;
-
-import java.lang.ref.Reference;
-
 /**
  * Base class for virtual thread implementations.
  */
@@ -68,13 +63,5 @@ sealed abstract class BaseVirtualThread extends Thread
      * Makes available the parking permit to the given this virtual thread.
      */
     abstract void unpark();
-
-    @Hidden
-    @ForceInline
-    protected void runWith(Object bindings, Runnable op) {
-        ensureMaterializedForStackWalk(bindings);
-        op.run();
-        Reference.reachabilityFence(bindings);
-    }
 }
 
--- a/src/java.base/share/classes/java/lang/BaseVirtualThread.java
+++ b/src/java.base/share/classes/java/lang/BaseVirtualThread.java
@@ -24,6 +24,11 @@
  */
 package java.lang;
 
+import jdk.internal.vm.annotation.ForceInline;
+import jdk.internal.vm.annotation.Hidden;
+
+import java.lang.ref.Reference;
+
 /**
  * Base class for virtual thread implementations.
  */
@@ -63,5 +68,13 @@ sealed abstract class BaseVirtualThread extends Thread
      * Makes available the parking permit to the given this virtual thread.
      */
     abstract void unpark();
+
+    @Hidden
+    @ForceInline
+    protected void runWith(Object bindings, Runnable op) {
+        ensureMaterializedForStackWalk(bindings);
+        op.run();
+        Reference.reachabilityFence(bindings);
+    }
 }
 
--- a/test/jdk/ProblemList-Virtual.txt
+++ b/test/jdk/ProblemList-Virtual.txt
@@ -44,8 +44,6 @@ javax/management/remote/mandatory/loading/MissingClassTest.java 8145413 windows-
 
 javax/management/remote/mandatory/loading/RMIDownloadTest.java 8308366 windows-x64
 
-java/lang/ScopedValue/StressStackOverflow.java 8309646 generic-all
-
 java/lang/instrument/NativeMethodPrefixAgent.java 8307169 generic-all
 
 ##########
--- a/src/hotspot/share/prims/jvm.cpp
+++ b/src/hotspot/share/prims/jvm.cpp
@@ -1386,7 +1386,7 @@ JVM_ENTRY(jobject, JVM_FindScopedValueBindings(JNIEnv *env, jclass cls))
     InstanceKlass* holder = method->method_holder();
     if (name == vmSymbols::runWith_method_name()) {
       if ((holder == resolver.Carrier_klass
-           || holder == vmClasses::VirtualThread_klass()
+           || holder == vmClasses::BaseVirtualThread_klass()
            || holder == vmClasses::Thread_klass())) {
         loc = 1;
       }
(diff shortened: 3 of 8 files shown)


src/java.base/share/classes/java/lang/BaseVirtualThread.java
```diff
@@ -63,5 +68,13 @@ sealed abstract class BaseVirtualThread extends Thread
...
      */
     abstract void unpark();
+
+    @Hidden
+    @ForceInline
+    protected void runWith(Object bindings, Runnable op) {
+        ensureMaterializedForStackWalk(bindings);
+        op.run();
+        Reference.reachabilityFence(bindings);
+    }
```
AlanBateman, line 78: BaseVirtualThread <: Thread.  Would it be simpler, and avoid duplicate, to just change Thread.runWith to be a final method with default access. That would be accessibility to VirtualThread and BoundVirtualThread without duplication.

offamitkumar:
Hi Andrew, 

This test failure is no more reproducible on s390x now, would you consider undo-ing problemlisting done by [JDK-8303499](https://bugs.openjdk.org/browse/JDK-8303499), I'll close the parent task [JDK-8303498](https://bugs.openjdk.org/browse/JDK-8303498)

src/java.base/share/classes/java/lang/BaseVirtualThread.java
theRealAph: Sure, that makes sense.

AlanBateman:
One suggestion for the test is to have it re-run with -XX:-VMContinuations on platforms/ports that do do have VM continuations support. Some of the tests for virtual threads do the same thing to ensure this operating mode is tested. This should do it:

```
--- a/test/jdk/java/lang/ScopedValue/StressStackOverflow.java
+++ b/test/jdk/java/lang/ScopedValue/StressStackOverflow.java
@@ -21,8 +21,8 @@
  * questions.
  */
 
-/**
- * @test
+/*
+ * @test id=default
  * @summary StressStackOverflow the recovery path for ScopedValue
  * @enablePreview
  * @run main/othervm/timeout=300 -XX:-TieredCompilation StressStackOverflow
@@ -30,6 +30,13 @@
  * @run main/othervm/timeout=300 StressStackOverflow
  */
 
+/*
+ * @test id=no-vmcontinuations
+ * @requires vm.continuations
+ * @enablePreview
+ * @run main/othervm/timeout=300 -XX:+UnlockExperimentalVMOptions -XX:-VMContinuations StressStackOverflow
+ */
+
```

test/jdk/ProblemList-Virtual.txt
```diff
@@ -44,8 +44,6 @@ javax/management/remote/mandatory/loading/MissingClassTest.java 8145413 windows-
 
 javax/management/remote/mandatory/loading/RMIDownloadTest.java 8308366 windows-x64
 
-java/lang/ScopedValue/StressStackOverflow.java 8309646 generic-all
-
```
AlanBateman, line 48: That's the issue for the JTREG_TEST_THREAD_FACTORY=Virtual runs. I think we need to leave it excluded until that issue is diagnosed/fixed. 

However, in jdk/test/ProblemList.txt, it is excluded on linux-s390x as that platform doesn't seem to have, or use, VM continuations and so is using the fallback/alternative implementation. I think this entry can be removed now.

theRealAph: I think it's fixed now.

Two problems:
1. Test was not guaranteed to make forward progress. Fixed with a 3 minute timeout.
2. Test was very slow in some conditions. Likewise.

AlanBateman (approved):
The implementation changes look good.

The update to test looks okay too although I would like to see if we can do better in SVC and STS when there is a SO.  In passing, I think all usages of tlr should be replaced with ThreadLocalRandom.current().

theRealAph:
> The implementation changes look good.
> 
> The update to test looks okay too although I would like to see if we can do better in SVC and STS when there is a SO.

That would be nice, but it's something of a can of worms. 

> In passing, I think all usages of tlr should be replaced with ThreadLocalRandom.current().

Err, sure -- but why? I guess there may be something I don't understand here.

AlanBateman (approved):
I'd probably go further and drop the TLR parameter from the choose method too but what you have is fine, let's go with that.

AlanBateman:
I didn't spot this before it was integrated but test/hotspot/jtreg/serviceability/jvmti/thread/GetStackTrace/getstacktr03/libgetstacktr03.cpp is sensitive to the stack trace so its like of expected frames will be take account of runWith no longer be overridden in the sub-class.

dcubed-ojdk:
@theRealAph -
So the GHA results for this PR show that serviceability/jvmti/thread/GetStackTrace/getstacktr03/getstacktr03
failed. Why wasn't this addressed prior to integration? We saw 5 instances of this failure mode in
every Mach5 Tier1 and 20 instances in every Mach5 Tier3 after this PR was integrated. This was not
an intermittent failure.

I've applied the fix that @AlanBateman proposed in [JDK-8310211](https://bugs.openjdk.org/browse/JDK-8310211)
and that appears to have solved the issue for Mach5 Tier1. I'll have to wait for the
fix to reach Mach5 Tier3 before I can declare that one is also good.

PR #14404, opened 2023-06-09
8309622: Re-examine the cache mechanism in BaseLocale

naotoj:
This is stemming from the PR: https://github.com/openjdk/jdk/pull/14211 where aggressive GC can cause NPE in `BaseLocale$Key` class. I refactored the in-house cache with WeakHashMap, and removed the Key class as it is no longer needed (thus the original NPE will no longer be possible). Also with the new JMH test case, it gains some performance improvement:
```
(w/o fix)

Benchmark                       Mode  Cnt      Score     Error  Units
LocaleCache.testForLanguageTag  avgt   20   5781.275 ± 569.580  ns/op
LocaleCache.testLocaleOf        avgt   20  62564.079 ± 406.697  ns/op

(w/ fix)
Benchmark                       Mode  Cnt      Score     Error  Units
LocaleCache.testForLanguageTag  avgt   20   4801.175 ± 371.830  ns/op
LocaleCache.testLocaleOf        avgt   20  60394.652 ± 352.471  ns/op
```

---------
### Progress

### Issue
 * [JDK-8309622](https://bugs.openjdk.org/browse/JDK-8309622): Re-examine the cache mechanism in BaseLocale (**Bug** - P4)

### Reviewers
 * [Daniel Fuchs](https://openjdk.org/census#dfuchs) (@dfuch - **Reviewer**) ⚠️ Review applies to [fb1535ff](https://git.openjdk.org/jdk/pull/14404/files/fb1535ffa380f45e8d2d3c2810255fd15f4ec70c)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14404/head:pull/14404` \
`$ git checkout pull/14404`

`$ git checkout pull/14404` \
`$ git pull https://git.openjdk.org/jdk.git pull/14404/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14404`

`$ git pr show -t 14404`

</details>

<a href="https://git.openjdk.org/jdk/pull/14404.diff">https://git.openjdk.org/jdk/pull/14404.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14404#issuecomment-1585274122)


dfuch (approved):
The new code looks much easier to maintain than the old one. I couldn't figure out what the old code was actually trying to do, but reading the new code I now understand what it does. Caching, retrieval, hashCode and equals look good. From the code it's trivial to see that a new normalised BaseLocale will be created if none is found in the cache - so that will assuredly fix the NPE that was experienced previously when the cache was cleared. The main change is that base locales will now be cached using WeakReference rather than SoftReference, and thus probably collected more eagerly, but that actually looks like a positive change. Not a usual reviewer in this are so please get another reviewer before integrating (also I haven't looked at the test).

dfuch (commented):
One thing though - WeakHashMap is not thread safe - and I don't see any locks or synchronized to protect access to CACHE. Is that going to cause trouble if two threads compete to add a locale to the map? In other words - how confident are we that the race would be benign?

dfuch (changes requested):
I think this question should be answered before integrating.

dfuch:
> The main change is that base locales will now be cached using WeakReference rather than SoftReference, and thus probably collected more eagerly, but that actually looks like a positive change.

Hmm.... or not. Nothing references the key, so the binding would be immediately eligible for garbage collection.

RogerRiggs:
One other thing to consider...
If the BaseLocale entries are being GC'd sooner, then perhaps the value of the String.intern() is reduced.
Eliminating `.intern90` would make creating a normalized BaseLocale faster.

test/hotspot/jtreg/gc/TestAllocHumongousFragment.java
```diff
@@ -177,6 +177,20 @@
  *      TestAllocHumongousFragment
  */
 
+/*
```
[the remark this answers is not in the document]
naotoj: Moved the block next to `iu-aggressive`

naotoj:
> One thing though - WeakHashMap is not thread safe - and I don't see any locks or synchronized to protect access to CACHE. Is that going to cause trouble if two threads compete to add a locale to the map? In other words - how confident are we that the race would be benign?

Although I think the possible race is benign (it would simply overwrite the existing normalized one), I put it in a synchronization block as a safety measure, as the race would be quite rare for creating a BaseLocale instance. Tested with the microbench mark and I didn't see the number change.

naotoj:
> One other thing to consider... If the BaseLocale entries are being GC'd sooner, then perhaps the value of the String.intern() is reduced. Eliminating `.intern90` would make creating a normalized BaseLocale faster.

Some code inside the Locale still relies on `intern` and do `==` comparison. Yes, it would be faster but I think that would be out of the scope of this fix

src/java.base/share/classes/sun/util/locale/BaseLocale.java
```diff
@@ -165,8 +159,18 @@ public static BaseLocale getInstance(String language, String script,
             language = convertOldISOCodes(language);
         }
 
-        Key key = new Key(language, script, region, variant, false);
-        return Cache.CACHE.get(key);
+        // Obtain the "normalized" BaseLocale, using un-normalized
+        // BaseLocale as the key. The returned "normalized" instance
+        // can subsequently be used by the Locale instance which
+        // guarantees the locale components are properly cased/interned.
+        synchronized (BaseLocale.class) {
```
AlanBateman, line 166: The simplification is good but I wonder if this coarse locking is going to be a problem, do we need to use some concurrent to avoid contention here?

test/micro/org/openjdk/bench/java/util/LocaleCache.java
```diff
@@ -0,0 +1,73 @@
...
+public class LocaleCache {
+    @Benchmark
+    public void testForLanguageTag() {
+        Locale previous = null;
+        for (int count = 100; count > 0; count--) {
+            var l = Locale.forLanguageTag("foo");
+            if (previous != null && previous != l) {
+                throw new RuntimeException("Different Locale was created");
+            }
+            previous =  l;
```
turbanoff, line 58:
```suggestion
            previous = l;
```

test/micro/org/openjdk/bench/java/util/LocaleCache.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Benchmark
+    public void testLocaleOf() {
+        Locale previous = null;
+        for (int count = 100; count > 0; count--) {
+            var l = Locale.of("foo");
+            if (previous != null && previous != l) {
+                throw new RuntimeException("Different Locale was created");
+            }
+            previous =  l;
```
turbanoff, line 70:
```suggestion
            previous = l;
```

src/java.base/share/classes/sun/util/locale/BaseLocale.java
naotoj: Replaced it with a ReentrantLock

test/micro/org/openjdk/bench/java/util/LocaleCache.java
naotoj: Fixed

src/java.base/share/classes/sun/util/locale/BaseLocale.java
AlanBateman: > Replaced it with a ReentrantLock

The concern is that this is a system-wide lock and so creates the potential for contention when many threads are bashing on Locale.of and other methods. Moving to use the JDK's ReferenceKeyMap with a CHM, or roll something similar, would avoid that but I can't helping feeling that the lookup is already a bit complicated. Right now, obtaining a Locale will create a non-normalized BaseLocale to use as a key, use the key to get (or compute) the normalized BaseLocale, then the normalized BaseLocale as a key for the Locale cache. Since it's early in JDK 22 it makes me wonder if we should step back and re-think this.

On the WHM, then it is a bit surprising that a Locales don't keep the non-normalized BaseLocale reachable. I assume this means that entries will be expunged from the CACHE if there is a reference processing between lookups. Also using weak refs makes it harder to reason about when the other part of the caching (in LocaleObjectCache) is used soft refs. I think this is another reason to think about a larger overhaul.

dfuch: One possibility would be to have the whole cache be soft-referenced, instead of weak-referencing the key.
Something like Reference<Map<BaseLocale, BaseLocale>> where Reference is a SoftReference and Map is a ConcurrentHashMap. 
But then you'd have to deal with a non-final field for the reference to the cache, and creating a new map and a new cache when the reference to the cache gets cleared.

naotoj: Jim's `java.lang.runtime.ReferenceKeyMap` would be a possible solution for it, although it is currently package private.

src/java.base/share/classes/sun/util/locale/BaseLocale.java
```diff
@@ -165,8 +162,21 @@ public static BaseLocale getInstance(String language, String script,
...
+        // BaseLocale as the key. The returned "normalized" instance
+        // can subsequently be used by the Locale instance which
+        // guarantees the locale components are properly cased/interned.
+        LOCK.lock();
+        try {
+            return CACHE.computeIfAbsent(new BaseLocale(language, script, region, variant),
+                    (b) -> new BaseLocale(
+                            LocaleUtils.toLowerString(b.getLanguage()).intern(),
+                            LocaleUtils.toTitleString(b.getScript()).intern(),
+                            LocaleUtils.toUpperString(b.getRegion()).intern(),
```
bokken, line 175: don't lines 147 and 148 above already guarantee the language and region have correct case?

src/java.base/share/classes/sun/util/locale/BaseLocale.java
```diff
@@ -165,8 +162,21 @@ public static BaseLocale getInstance(String language, String script,
             language = convertOldISOCodes(language);
         }
 
-        Key key = new Key(language, script, region, variant, false);
-        return Cache.CACHE.get(key);
+        // Obtain the "normalized" BaseLocale, using un-normalized
+        // BaseLocale as the key. The returned "normalized" instance
+        // can subsequently be used by the Locale instance which
+        // guarantees the locale components are properly cased/interned.
```
bokken, line 168: Is it truly important that the String values are interned, or is the desire simply that all refer to the same "canonical" String instance?
my thought is that managing the canonical String instances could avoid WeakReferences and synchronized map access and just return a new BaseLocale each time.



```
    public static BaseLocale getInstance(String language, String script,
                                         String region, String variant) {
        if (script == null) {
            script = "";
        } else {
            script = getCanonicalString(script, LocaleUtils::toTitleString);
        }

        if (region == null) {
            region = "";
        } else {
            region = getCanonicalString(region, LocaleUtils::toUpperString);
        }

        if (language == null) {
            language = "";
        } else {
            language = getCanonicalString(language, LocaleUtils::toLowerString);
        }
        
        if (variant == null) {
            variant = "";
        } else {
            variant = getCanonicalString(variant, Function.identity());
        }
        ...
        return new BaseLocale(language, script, region, variant);
    }



    private static final ConcurrentMap<String, String> CANONICALIZED_STRINGS = new ConcurrentHashMap<>();

    static {
        // prime the old iso codes
        CANONICALIZED_STRINGS.put("he", "he");
        CANONICALIZED_STRINGS.put("iw", "iw");
        CANONICALIZED_STRINGS.put("id", "id");
        CANONICALIZED_STRINGS.put("in", "in");
        CANONICALIZED_STRINGS.put("ji", "ji");
        CANONICALIZED_STRINGS.put("yi", "yi");
    }
    
    private static String getCanonicalString(String value, Function<String, String> conversion) {
        String existing = CANONICALIZED_STRINGS.get(value);
        if (existing != null) {
            return existing;
        }

        String converted = conversion.apply(value);
        return CANONICALIZED_STRINGS.merge(converted, converted, (k,v) -> v);
    }
```

naotoj: Interning components in a Locale instance (through BaseLocale) is a long standing behavior. Changing it could break apps that might have relied on it (using `==` for comparison)

src/java.base/share/classes/sun/util/locale/BaseLocale.java
naotoj: 147,148 are in `getInstance()` and the BaseLocale key here is created with a constructor, which simply copies the passed arguments. So normalization is needed here

src/java.base/share/classes/sun/util/locale/BaseLocale.java
bokken: as in doing `"en" == locale.getLanguage()`?

In that case, the interned values could still be used as the canonical values, and the CHM can account both for case adjustments and interning (and likely faster than actually calling intern). I am curious how this does with your jmh test:

```

    private static final ConcurrentMap<String, String> CANONICALIZED_UPPER_STRINGS = new ConcurrentHashMap<>();
    private static final ConcurrentMap<String, String> CANONICALIZED_LOWER_STRINGS = new ConcurrentHashMap<>();
    private static final ConcurrentMap<String, String> CANONICALIZED_TITLE_STRINGS = new ConcurrentHashMap<>();
    private static final ConcurrentMap<String, String> CANONICALIZED_STRINGS = new ConcurrentHashMap<>();

    public static BaseLocale getInstance(String language, String script,
                                         String region, String variant) {
        if (script == null || script.isEmpty()) {
            script = "";
        } else {
            script = getCanonicalString(script, LocaleUtils::toTitleString, CANONICALIZED_TITLE_STRINGS);
        }

        if (region == null || region.isEmpty()) {
            region = "";
        } else {
            region = getCanonicalString(region, LocaleUtils::toUpperString, CANONICALIZED_UPPER_STRINGS);
        }

        if (language == null || language.isEmpty()) {
            language = "";
        } else {
            language = getCanonicalString(language, LocaleUtils::toLowerString, CANONICALIZED_LOWER_STRINGS);
        }

        if (variant == null || variant.isEmpty()) {
            variant = "";
        } else {
            variant = getCanonicalString(variant, Function.identity(), CANONICALIZED_STRINGS);
        }
        ...
        return new BaseLocale(language, script, region, variant);
    }

    private static String getCanonicalString(String value, Function<String, String> conversion, ConcurrentMap<String, String> map) {
        String existing = map.get(value);
        if (existing != null) {
            return existing;
        }

        String converted = conversion.apply(value).intern();
        // put the interned value in, keyed by the interned value
        map.putIfAbsent(converted, converted);

        // also put the interned value in for the given key, if not already present
        // this avoids future needs to apply a conversion to the value
        map.putIfAbsent(value, converted);
        return converted;
    }
```

src/java.base/share/classes/sun/util/locale/BaseLocale.java
bokken: perhaps I am missing something, but aren't `language` and `region` being reassigned to the case adjusted values? So the values used to create the "key" BaseLocale instance reflect that case adjustment? (language is again potentially adjusted at line 162 to account for old iso codes, but that never results in the wrong case.)

```
        language = LocaleUtils.toLowerString(language);
        region = LocaleUtils.toUpperString(region);
```

src/java.base/share/classes/sun/util/locale/BaseLocale.java
```diff
@@ -165,8 +160,16 @@ public static BaseLocale getInstance(String language, String script,
...
+        return CACHE.computeIfAbsent(new BaseLocale(language, script, region, variant),
```
liach, line 167: This `computeIfAbsent` will store the non-normalized base locale as a soft key in the cache, which is undesirable. I have commented on the base patch https://github.com/openjdk/jdk/pull/14684#discussion_r1266914091 to support distinct keys for querying and storing like those in the old `LocalObjectCache`, and we can then simply use a `ReferenceKeySet` for the cache.

naotoj: The reason I used non-normalized base locale as the key was that normalizing (interning) was slower than comparing the key. I would have to double check, and see if the distinct normalized key is faster.

src/java.base/share/classes/sun/util/locale/BaseLocale.java
```diff
@@ -90,6 +92,9 @@ public final class BaseLocale {
         }
     }
 
+    // Non-normalized to normalized BaseLocale cache for saving costly normalizations
+    private static final Map<BaseLocale, BaseLocale> CACHE = ReferencedKeyMap.create(true, ConcurrentHashMap::new);
```
liach, line 96: You can now simply use a `ReferencedKeySet<BaseLocale>` and call [`intern(T, UnaryOperator<T>)`](https://github.com/openjdk/jdk/blob/ce2a7ea40a22c652e5f8559c91d5eea197e2d708/src/java.base/share/classes/jdk/internal/util/ReferencedKeySet.java#L202) to cache a normalized variety computed from a non-normalized base.

test/jdk/jdk/internal/util/ReferencedKeyTest.java
liach, line 1: Is the test removal intentional?

src/java.base/share/classes/sun/util/locale/BaseLocale.java
naotoj: I need to double-check that it won't cause performance regression, will be visiting this later.

test/jdk/jdk/internal/util/ReferencedKeyTest.java
naotoj: Ah, thanks. My bad for unintentionally removing the test.

shipilev:
@naotoj, we are still seeing the CI failures due to this. Are you planning to finish this work? Thanks!

naotoj:
keep open

naotoj:
keep open

naotoj:
After a hiatus, I now got back to work on this. I think I reflected all the comments (and removed the performance bench because it no longer offers an improvement).

RogerRiggs (approved):
Much simpler and looks good.  Thanks

PR #10704, opened 2022-10-13
8240567: MethodTooLargeException thrown while creating a jlink image

koppor:
Fix for [JDK-8240567](https://bugs.openjdk.org/browse/JDK-8240567): "MethodTooLargeException thrown while creating a jlink image".

Java still has a 64kb limit: A method may not be longer than 64kb. The idea of the fix is to split up the generated methods in several smaller methods

---------
### Progress
- [ ] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Error
&nbsp;⚠️ Executable files are not allowed (file: src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java)

### Issue
 * [JDK-8240567](https://bugs.openjdk.org/browse/JDK-8240567): MethodTooLargeException thrown while creating a jlink image (**Bug** - P4)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/10704/head:pull/10704` \
`$ git checkout pull/10704`

Update a local copy of the PR: \
`$ git checkout pull/10704` \
`$ git pull https://git.openjdk.org/jdk.git pull/10704/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 10704`

View PR using the GUI difftool: \
`$ git pr show -t 10704`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jdk/pull/10704.diff">https://git.openjdk.org/jdk/pull/10704.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/10704#issuecomment-1278192697)

PR #14408, opened 2023-06-11
8240567: MethodTooLargeException thrown while creating a jlink image

koppor:
Fix for [JDK-8240567](https://bugs.openjdk.org/browse/JDK-8240567): "MethodTooLargeException thrown while creating a jlink image".

Java still has a 64kb limit: A method may not be longer than 64kb. The idea of the fix is to split up the generated methods in several smaller methods

This is a follow-up to https://github.com/openjdk/jdk/pull/10704. GitHub did not allow me to re-open the PR, because I did a force-push to have one commit.

---------
### Progress

### Issue
 * [JDK-8240567](https://bugs.openjdk.org/browse/JDK-8240567): MethodTooLargeException thrown while creating a jlink image (**Bug** - P4)

### Reviewers
 * [Mandy Chung](https://openjdk.org/census#mchung) (@mlchung - **Reviewer**) ⚠️ Review applies to [e112f03a](https://git.openjdk.org/jdk/pull/14408/files/e112f03aacc9194e49dbcf605dd538c3ca7ce162)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14408/head:pull/14408` \
`$ git checkout pull/14408`

`$ git checkout pull/14408` \
`$ git pull https://git.openjdk.org/jdk.git pull/14408/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14408`

`$ git pr show -t 14408`

</details>

<a href="https://git.openjdk.org/jdk/pull/14408.diff">https://git.openjdk.org/jdk/pull/14408.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14408#issuecomment-1586373375)


Siedlerchr:
We validated this by compiling JabRef against it and a) writing out the generated class file for the generated bytecode  and b)  verifying with -Xlog:init=debug -XX:+UnlockDiagnosticVMOptions -XX:+BytecodeVerificationLocal 

Is there a way to "parse" or call the verifyer after generation?

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
```diff
@@ -666,25 +667,140 @@ private void genIncubatorModules(ClassBuilder clb) {
...
+                clb.withMethodBody(
+                        helperMethodNamePrefix + index[0],
+                        MethodTypeDesc.of(CD_void, CD_MODULE_DESCRIPTOR.arrayType(), arrayListClassDesc),
+                        ACC_PUBLIC,
+                        cob -> {
+                            List<ModuleInfo> moduleInfosPackage = splitModuleInfos.get(index[0]);
+
+                            // Restore all (!) sets from parameter to local variables
+                            if (nextLocalVar > firstVariableForDedup) {
+                                // We need to go from the end to the beginning as we will probably overwrite position 2 (which holds the list at the beginning)
```
AlanBateman, line 754: Do you mind fixing all the comments as these 160+ lines make it impossible to look at the changes side-by-side again. You had fixed that in the original PR but it looks like they have come back with this PR.

AlanBateman (commented):
@koppor Is this ready for review? The other PR went through a dozens or so iterations before it was returned to draft. It seems like you were still battling with verifier errors. The comment on this PR says you it was created because a force-push so I can't tell if you the changes are ready or not.

koppor:
TL;DR: summary of the changes

- In the other PR we missed that `dedupSetBuilder` requires local variables in the method for caching
  - Code comments: `// Restore all (!) sets from parameter to local variables` and `// Store all new sets to List`
- In this PR, these variables are transferred from method to method by a `ArrayList`
- `0` is `this` when a method is called. `this` is needed to call the next helper method, thus we moved `BUILDER_VAR` to another place.
  - Code comments: `MAX_LOCAL_VARS + 1; // we need 0 for "this"` and on `MAX_LOCAL_VARS`: `// Constant chosen for this generator, can be higher in practice`

> @koppor Is this ready for review?

Yes, it is! We especially checked it with our (rather large) desktop application.

> The other PR went through a dozens or so iterations before it was returned to draft.

I remember 😅. Therefore, we first had this as draft, waited for the "on-site checks" and then opened it.

> It seems like you were still battling with verifier errors. 

We missed an important point and the other PR.

> The comment on this PR says you it was created because a force-push so I can't tell if you the changes are ready or not.

I am sorry for that. I weighted the requirement to have a single commit higher than reviewing the changes. The "excuse" for this decision is that nearly all code submitted back than has changed.

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
koppor: > Do you mind fixing all the comments as these 160+ lines

Done. I copied the (for me) important comments to the GitHub reply https://github.com/openjdk/jdk/pull/14408#issuecomment-1586615579.

> You had fixed that in the original PR but it looks like they have come back with this PR.

This is my job training 😇: Comment the core ideas (and do not verbatim repeat what is said by the code directly). - For me, the good thing is, the diff now has all the comments (https://github.com/openjdk/jdk/pull/14408/commits/23bbc0ce0c8fd8a4cd689c0260c5fbcb91b20046). OK, at least one has go to nirvana in all cases.

Siedlerchr: Yes, the verify errors are gone. Application (JabRef) starts sucessfully against the compilation. Is there a way to verify in a tesr the creation of the helper methos in the bytecode? e.g. something like getDeclaredMethods?

koppor:
@asontana At https://bugs.openjdk.org/browse/JDK-8240567?focusedCommentId=14588425&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14588425 you asked the following:
    
> Did you consider also constant pool 65k entries size limit? 
When module descriptors are split into multiple methods due to their code size, the system modules plugin may soon hit the class constant pool limit of 65k unique entries

I don't have a bugs.openjdk.org account; thus I reply here.

The constant pool is also an "interesting" limitation. We did not address it (yet) because the current patch is IMHO a major step forward: It unblocks feature development of our software: We can add new external libraries :tada:

Looking more close to that limit and the implications to the patch: I understand that this limit affects the number of methods in a class. If this is correct, let's assume that a project requires thousands of modules and would use half of the limit for the generation of the module descriptors. That would result in roughly 32,000 methods. The patch puts 50 module descriptors in one method. Assuming 32,000 methods leads to ~1.6 million modules for a Java program.

Currently, our project [JabRef](https://github.com/jabref/jabref/) has a total number of ~130 modules. We cannot move forward with our release as we cannot add new dependencies. We hope that other projects with many modules will benefit from this patch as well.

koppor:
@AlanBateman The diff to the "old" PR https://github.com/openjdk/jdk/pull/10704 is as follows. Maybe this helps at reviewing? IMHO this diff shows very good the new passing of the locals required by the deduplication functionality.

```diff
diff --git a/src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java b/src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
old mode 100755
new mode 100644
index 8e98a5b94e2..a87fa84ebe2
--- a/src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
+++ b/src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
@@ -118,7 +118,7 @@ public final class SystemModulesPlugin extends AbstractPlugin {
             ClassDesc.ofInternalName("jdk/internal/module/SystemModules");
     private static final ClassDesc CD_SYSTEM_MODULES_MAP =
             ClassDesc.ofInternalName(SYSTEM_MODULES_MAP_CLASSNAME);
-    private final boolean enabled;
+    private boolean enabled;
 
     public SystemModulesPlugin() {
         super("system-modules");
@@ -519,7 +519,7 @@ public final class SystemModulesPlugin extends AbstractPlugin {
 
         private static final int MAX_LOCAL_VARS = 256;
 
-        private final int BUILDER_VAR    = 0;
+        private final int BUILDER_VAR    = MAX_LOCAL_VARS + 1;
         private final int MD_VAR         = 1;  // variable for ModuleDescriptor
         private final int MT_VAR         = 1;  // variable for ModuleTarget
         private final int MH_VAR         = 1;  // variable for ModuleHashes
@@ -666,7 +666,7 @@ public final class SystemModulesPlugin extends AbstractPlugin {
          * Generate bytecode for moduleDescriptors method
          */
         private void genModuleDescriptorsMethod(ClassBuilder clb) {
-            if (moduleInfos.size() <= 71) {
+            if (moduleInfos.size() <= 75) {
                 clb.withMethodBody(
                         "moduleDescriptors",
                         MethodTypeDesc.of(CD_MODULE_DESCRIPTOR.arrayType()),
@@ -688,71 +688,98 @@ public final class SystemModulesPlugin extends AbstractPlugin {
                 return;
             }
 
-            // split up module infos in "consumable" packages
             List<List<ModuleInfo>> splitModuleInfos = new ArrayList<>();
             List<ModuleInfo> currentModuleInfos = null;
             for (int index = 0; index < moduleInfos.size(); index++) {
-                // The method is "manually split" based on the heuristics that 90 ModuleDescriptors are smaller than 64kb
-                // The number 10 is chosen "randomly" to be below the 64kb limit of a method
-                if (index % 10 == 0) {
-                    // Prepare new list
+                if (index % 50 == 0) {
                     currentModuleInfos = new ArrayList<>();
                     splitModuleInfos.add(currentModuleInfos);
                 }
                 currentModuleInfos.add(moduleInfos.get(index));
             }
-            // generate all helper methods
+
             final String helperMethodNamePrefix = "moduleDescriptorsSub";
+            final ClassDesc arrayListClassDesc = ClassDesc.ofInternalName("java/util/ArrayList");
+
+            final int firstVariableForDedup = nextLocalVar;
+
+            clb.withMethodBody(
+                    "moduleDescriptors",
+                    MethodTypeDesc.of(CD_MODULE_DESCRIPTOR.arrayType()),
+                    ACC_PUBLIC,
+                    cob -> {
+                        cob.constantInstruction(moduleInfos.size())
+                                .anewarray(CD_MODULE_DESCRIPTOR)
+                                .dup()
+                                .astore(MD_VAR);
+                        cob.new_(arrayListClassDesc)
+                           .dup()
+                           .invokespecial(arrayListClassDesc, "<init>", MethodTypeDesc.of(CD_void))
+                           .astore(nextLocalVar);
+                        cob.aload(0)
+                           .aload(MD_VAR)
+                           .aload(nextLocalVar)
+                           .invokevirtual(
+                                   this.classDesc,
+                                   helperMethodNamePrefix + "0",
+                                   MethodTypeDesc.of(CD_void, CD_MODULE_DESCRIPTOR.arrayType(), arrayListClassDesc)
+                           )
+                           .areturn();
+                    });
+
             final int[] globalCount = {0};
             for (final int[] index = {0}; index[0] < splitModuleInfos.size(); index[0]++) {
                 clb.withMethodBody(
                         helperMethodNamePrefix + index[0],
-                        MethodTypeDesc.of(CD_void, CD_MODULE_DESCRIPTOR.arrayType()),
+                        MethodTypeDesc.of(CD_void, CD_MODULE_DESCRIPTOR.arrayType(), arrayListClassDesc),
                         ACC_PUBLIC,
                         cob -> {
                             List<ModuleInfo> moduleInfosPackage = splitModuleInfos.get(index[0]);
-                            if (index[0] > 0) {
-                                // call last helper method
-                                cob.aload(0)
-                                   .aload(MD_VAR) // load first parameter, which is MD_VAR
-                                   .invokevirtual(
-                                           this.classDesc,
-                                           helperMethodNamePrefix + (index[0] - 1),
-                                           MethodTypeDesc.of(CD_void, CD_MODULE_DESCRIPTOR.arrayType())
-                                   );
+
+                            if (nextLocalVar > firstVariableForDedup) {
+                                for (int i = nextLocalVar-1; i >= firstVariableForDedup; i--) {
+                                    cob.aload(2)
+                                       .constantInstruction(i-firstVariableForDedup)
+                                       .invokevirtual(arrayListClassDesc, "get", MethodTypeDesc.of(CD_Object, CD_int))
+                                       .astore(i);
+                                }
                             }
+
                             for (int j = 0; j < moduleInfosPackage.size(); j++) {
                                 ModuleInfo minfo = moduleInfosPackage.get(j);
-                                // executed after the call, thus it is OK to overwrite index 0 (BUILDER_VAR)
                                 new ModuleDescriptorBuilder(cob,
                                         minfo.descriptor(),
                                         minfo.packages(),
                                         globalCount[0]).build();
                                 globalCount[0]++;
                             }
+
+                            if (index[0] + 1 < (splitModuleInfos.size())) {
+                                if (nextLocalVar > firstVariableForDedup) {
+                                    cob.new_(arrayListClassDesc)
+                                       .dup()
+                                       .invokespecial(arrayListClassDesc, "<init>", MethodTypeDesc.of(CD_void))
+                                       .astore(nextLocalVar);
+                                    for (int i = firstVariableForDedup; i < nextLocalVar; i++) {
+                                        cob.aload(nextLocalVar)
+                                           .aload(i)
+                                           .invokevirtual(arrayListClassDesc, "add", MethodTypeDesc.of(CD_boolean, CD_Object))
+                                           .pop();
+                                    }
+                                }
+                                cob.aload(0)
+                                   .aload(MD_VAR)
+                                   .aload(nextLocalVar)
+                                   .invokevirtual(
+                                           this.classDesc,
+                                           helperMethodNamePrefix + (index[0] + 1),
+                                           MethodTypeDesc.of(CD_void, CD_MODULE_DESCRIPTOR.arrayType(), arrayListClassDesc)
+                                   );
+                            }
+
                             cob.return_();
                         });
             }
-
-            // generate call to last helper method
-            clb.withMethodBody(
-                    "moduleDescriptors",
-                    MethodTypeDesc.of(CD_MODULE_DESCRIPTOR.arrayType()),
-                    ACC_PUBLIC,
-                    cob -> {
-                        cob.constantInstruction(moduleInfos.size())
-                           .anewarray(CD_MODULE_DESCRIPTOR)
-                           .astore(MD_VAR)
-                           .aload(MD_VAR) // storing for the return at the end
-                           .aload(0)
-                           .aload(MD_VAR)
-                           .invokevirtual(
-                                   this.classDesc,
-                                   helperMethodNamePrefix + (splitModuleInfos.size() - 1),
-                                   MethodTypeDesc.of(CD_void, CD_MODULE_DESCRIPTOR.arrayType())
-                           )
-                           .areturn();
-                    });
         }
 
         /**
diff --git a/test/jdk/tools/jlink/JLink100Modules.java b/test/jdk/tools/jlink/JLink100Modules.java
index 4a6111505d2..e71206c904f 100644
--- a/test/jdk/tools/jlink/JLink100Modules.java
+++ b/test/jdk/tools/jlink/JLink100Modules.java
@@ -27,8 +27,10 @@ import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.StringJoiner;
 import java.util.spi.ToolProvider;
+
 import tests.JImageGenerator;
 import tests.JImageGenerator.JLinkTask;
+
 /*
  * @test
  * @summary Make sure that 100 modules can be linked using jlink.
@@ -46,9 +48,9 @@ import tests.JImageGenerator.JLinkTask;
  */
 public class JLink100Modules {
     private static final ToolProvider JAVAC_TOOL = ToolProvider.findFirst("javac")
-                                                               .orElseThrow(() -> new RuntimeException("javac tool not found"));
+            .orElseThrow(() -> new RuntimeException("javac tool not found"));
     private static final ToolProvider JLINK_TOOL = ToolProvider.findFirst("jlink")
-                                                               .orElseThrow(() -> new RuntimeException("jlink tool not found"));
+            .orElseThrow(() -> new RuntimeException("jlink tool not found"));
 
     static void report(String command, String[] args) {
         System.out.println(command + " " + String.join(" ", Arrays.asList(args)));
@@ -77,9 +79,10 @@ public class JLink100Modules {
             StringBuilder builder = new StringBuilder("module ");
             builder.append(name).append(" {");
 
-            for (int j = 0; j < i; j++) {
-                builder.append("requires module" + j + "x;");
+            if (i != 0) {
+                builder.append("requires module0x;");
             }
+
             builder.append("}\n");
             Files.writeString(moduleDir.resolve("module-info.java"), builder.toString());
             mainModuleInfoContent.add(name);
@@ -106,7 +109,7 @@ public class JLink100Modules {
 
         String out = src.resolve("out").toString();
 
-        javac(new String[] {
+        javac(new String[]{
                 "-d", out,
                 "--module-source-path", src.toString(),
                 "--module", "bug8240567x"
@@ -116,6 +119,7 @@ public class JLink100Modules {
                 .modulePath(out)
                 .output(src.resolve("out-jlink"))
                 .addMods("bug8240567x")
-                .call().assertSuccess();
+                .call()
+                .assertSuccess();
     }
 }
```

AlanBateman:
> @AlanBateman The diff to the "old" PR #10704 is as follows. Maybe this helps at reviewing? IMHO this diff shows very good the new passing of the locals required by the deduplication functionality.

It's good that you've got to a patch that doesn't trip the verifier. It is one my list to look at the latest version.

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
```diff
@@ -666,25 +666,120 @@ private void genIncubatorModules(ClassBuilder clb) {
...
+                           .invokevirtual(
+                                   this.classDesc,
+                                   helperMethodNamePrefix + "0",
+                                   MethodTypeDesc.of(CD_void, CD_MODULE_DESCRIPTOR.arrayType(), arrayListClassDesc)
+                           )
                            .areturn();
                     });
+
+            final int[] globalCount = {0};
+            for (final int[] index = {0}; index[0] < splitModuleInfos.size(); index[0]++) {
```
mlchung, line 731:
```suggestion
            for (int n = 0, count = 0; n < splitModuleInfos.size(); count += splitModuleInfos.get(n).size(), n++) {
                int index = n;          // the index of which ModuleInfo being processed in the current batch
                int start = count;   // the start index to the return ModuleDescriptor array for the current batch
```

I prefer to avoid using the array to workaround the final variable used within lambda.   Free feel to rename these variables and comment to make it clear.

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
```diff
@@ -666,25 +666,120 @@ private void genIncubatorModules(ClassBuilder clb) {
...
+                clb.withMethodBody(
+                        helperMethodNamePrefix + index[0],
+                        MethodTypeDesc.of(CD_void, CD_MODULE_DESCRIPTOR.arrayType(), arrayListClassDesc),
+                        ACC_PUBLIC,
+                        cob -> {
+                            List<ModuleInfo> moduleInfosPackage = splitModuleInfos.get(index[0]);
```
mlchung, line 737: It's not a package.   Maybe `currentBatch`?

```suggestion
                            List<ModuleInfo> currentBatch = splitModuleInfos.get(index[0]);
```

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
```diff
@@ -666,25 +666,120 @@ private void genIncubatorModules(ClassBuilder clb) {
...
+            List<ModuleInfo> currentModuleInfos = null;
+            for (int index = 0; index < moduleInfos.size(); index++) {
+                if (index % 75 == 0) {
+                    currentModuleInfos = new ArrayList<>();
+                    splitModuleInfos.add(currentModuleInfos);
+                }
+                currentModuleInfos.add(moduleInfos.get(index));
+            }
+
+            final String helperMethodNamePrefix = "moduleDescriptorsSub";
```
mlchung, line 701: Perhaps a shorter name.   Also, prefer to drop the final modifier from these local variables (which adds noise).

```suggestion
            String helperMethodNamePrefix = "sub";
```

mlchung (commented):
Thanks for fixing this.   The approach looks okay.

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
```diff
@@ -666,25 +666,120 @@ private void genIncubatorModules(ClassBuilder clb) {
...
+                            cob.aload(MD_VAR)
+                               .areturn();
+                        });
+                return;
+            }
+
+            List<List<ModuleInfo>> splitModuleInfos = new ArrayList<>();
+            List<ModuleInfo> currentModuleInfos = null;
+            for (int index = 0; index < moduleInfos.size(); index++) {
+                if (index % 50 == 0) {
```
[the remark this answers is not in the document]
koppor: > The size of the generated moduleDescriptors is a function of the number of modules, the number of packages in the modules, number of exports, ... I think your approach is okay for now, meaning pick some threshold that you are confident will result in a method size < 64k.

I am very confident with 50 as 1) we tried out with a real life application using a high variation of dependencies (covering Apache libraries, Google libraries, and Oracle's JavaFX) and 2) it is half of the value of the number of modules where it broke at our software.

> It might be clearer to readers to use one threshold rather than two.

Above, the 75 was chosen, to avoid that anything in the JDK distribution is split (number was 73 to be precise, but I did want to have a "round" number). 

To be consistent, I adapted the lower split to 75. I tried it out with our app, and the app still works.

> The other limit that we will need to think about is the constant pool. That may be future work.

I am leaning towards "future work" here. Nevertheless, could you outline how that limit might be reached? It IMHO cannot be the numbmer of helper functions. The current "limit" leads to 32,000 (constant pool limit; leaving the other 32k to other things) x 75 (number of module descriptors covered) = 2.4 million module descriptors. In that case we might hit the call stack limit.

test/jdk/tools/jlink/JLink100Modules.java
```diff
@@ -0,0 +1,125 @@
...
+import tests.JImageGenerator;
+import tests.JImageGenerator.JLinkTask;
+
+/*
+ * @test
+ * @summary Make sure that 100 modules can be linked using jlink.
+ * @bug 8240567
+ * @library ../lib
+ * @modules java.base/jdk.internal.jimage
+ *          jdk.jdeps/com.sun.tools.classfile
```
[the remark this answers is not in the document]
koppor: If it is not present, I get following output (sorry for German - `./configure` doesn't allow `JAVA_TOOL_OPTIONS` to be set to `-Duser.language=en`

```
windir='C:\WINDOWS' \
    'c:\git-repositories\jdk\jdk\build\windows-x86_64-server-release\images\jdk\bin\javac' \
        -J-Xmx768m \
        -J-XX:MaxRAMPercentage=2.08333 \
        -J'-Dtest.boot.jdk=c:\progra~1\openjdk\jdk-20~1.1' \
        -J'-Djava.io.tmpdir=c:\git-repositories\jdk\jdk\build\windows-x86_64-server-release\test-support\jtreg_test_jdk_tools_jlink_JLink100Modules_java\tmp' \
        -J-ea \
        -J-esa \
        -J-Dtest.vm.opts='-Xmx768m -XX:MaxRAMPercentage=2.08333 -Dtest.boot.jdk=c:\progra~1\openjdk\jdk-20~1.1 -Djava.io.tmpdir=c:\git-repositories\jdk\jdk\build\windows-x86_64-server-release\test-support\jtreg_test_jdk_tools_jlink_JLink100Modules_java\tmp -ea -esa' \
        -J-Dtest.tool.vm.opts='-J-Xmx768m -J-XX:MaxRAMPercentage=2.08333 -J-Dtest.boot.jdk=c:\progra~1\openjdk\jdk-20~1.1 -J-Djava.io.tmpdir=c:\git-repositories\jdk\jdk\build\windows-x86_64-server-release\test-support\jtreg_test_jdk_tools_jlink_JLink100Modules_java\tmp -J-ea -J-esa' \
        -J-Dtest.compiler.opts= \
        -J-Dtest.java.opts= \
        -J-Dtest.jdk='c:\git-repositories\jdk\jdk\build\windows-x86_64-server-release\images\jdk' \
        -J-Dcompile.jdk='c:\git-repositories\jdk\jdk\build\windows-x86_64-server-release\images\jdk' \
        -J-Dtest.timeout.factor=4.0 \
        -J-Dtest.nativepath='c:\git-repositories\jdk\jdk\build\windows-x86_64-server-release\images\test\jdk\jtreg\native' \
        -J-Dtest.root='C:\git-repositories\jdk\jdk\test\jdk' \
        -J-Dtest.name=tools/jlink/JLink100Modules.java \
        -J-Dtest.file='C:\git-repositories\jdk\jdk\test\jdk\tools\jlink\JLink100Modules.java' \
        -J-Dtest.src='C:\git-repositories\jdk\jdk\test\jdk\tools\jlink' \
        -J-Dtest.src.path='C:\git-repositories\jdk\jdk\test\jdk\tools\jlink;C:\git-repositories\jdk\jdk\test\jdk\tools\lib' \
        -J-Dtest.classes='C:\git-repositories\jdk\jdk\build\windows-x86_64-server-release\test-support\jtreg_test_jdk_tools_jlink_JLink100Modules_java\classes\0\tools\jlink\JLink100Modules.d' \
        -J-Dtest.class.path='C:\git-repositories\jdk\jdk\build\windows-x86_64-server-release\test-support\jtreg_test_jdk_tools_jlink_JLink100Modules_java\classes\0\tools\jlink\JLink100Modules.d;C:\git-repositories\jdk\jdk\build\windows-x86_64-server-release\test-support\jtreg_test_jdk_tools_jlink_JLink100Modules_java\classes\0\tools\lib' \
        -J-Dtest.class.path.prefix='C:\git-repositories\jdk\jdk\test\jdk\tools\jlink;C:\git-repositories\jdk\jdk\build\windows-x86_64-server-release\test-support\jtreg_test_jdk_tools_jlink_JLink100Modules_java\classes\0\tools\lib' \
        -J-Dtest.modules='java.base/jdk.internal.jimage jdk.jlink/jdk.tools.jlink.internal jdk.jlink/jdk.tools.jlink.plugin jdk.jlink/jdk.tools.jmod jdk.jlink/jdk.tools.jimage jdk.compiler' \
        --add-modules java.base,jdk.jlink,jdk.compiler \
        --add-exports java.base/jdk.internal.jimage=ALL-UNNAMED \
        --add-exports jdk.jlink/jdk.tools.jlink.internal=ALL-UNNAMED \
        --add-exports jdk.jlink/jdk.tools.jlink.plugin=ALL-UNNAMED \
        --add-exports jdk.jlink/jdk.tools.jmod=ALL-UNNAMED \
        --add-exports jdk.jlink/jdk.tools.jimage=ALL-UNNAMED \
        -d 'C:\git-repositories\jdk\jdk\build\windows-x86_64-server-release\test-support\jtreg_test_jdk_tools_jlink_JLink100Modules_java\classes\0\tools\lib' \
        -sourcepath 'C:\git-repositories\jdk\jdk\test\jdk\tools\lib' \
        -classpath 'C:\git-repositories\jdk\jdk\build\windows-x86_64-server-release\test-support\jtreg_test_jdk_tools_jlink_JLink100Modules_java\classes\0\tools\lib' 'C:\git-repositories\jdk\jdk\test\jdk\tools\lib\tests\Helper.java' 'C:\git-repositories\jdk\jdk\test\jdk\tools\lib\tests\JImageGenerator.java' 'C:\git-repositories\jdk\jdk\test\jdk\tools\lib\tests\JImageValidator.java' 'C:\git-repositories\jdk\jdk\test\jdk\tools\lib\tests\Result.java'
direct:
C:\git-repositories\jdk\jdk\test\jdk\tools\lib\tests\JImageValidator.java:36: Fehler: Package com.sun.tools.classfile ist nicht sichtbar
import com.sun.tools.classfile.ClassFile;
                    ^
  (Package com.sun.tools.classfile wird in Modul jdk.jdeps deklariert, aber nicht in das unbenannte Modul exportiert)
C:\git-repositories\jdk\jdk\test\jdk\tools\lib\tests\JImageValidator.java:37: Fehler: Package com.sun.tools.classfile ist nicht sichtbar
import com.sun.tools.classfile.ConstantPoolException;
                    ^
  (Package com.sun.tools.classfile wird in Modul jdk.jdeps deklariert, aber nicht in das unbenannte Modul exportiert)
2 Fehler
```

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
```diff
@@ -666,25 +666,120 @@ private void genIncubatorModules(ClassBuilder clb) {
...
+                                        minfo.packages(),
+                                        index).build();
+                            }
+                            cob.aload(MD_VAR)
+                               .areturn();
+                        });
+                return;
+            }
+
+            List<List<ModuleInfo>> splitModuleInfos = new ArrayList<>();
```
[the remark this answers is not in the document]
koppor: I assume the "old" comments were too detailed. I removed them at [JabRef/jdk@`23bbc0c` (#2)](https://github.com/JabRef/jdk/pull/2/commits/23bbc0ce0c8fd8a4cd689c0260c5fbcb91b20046) to have the code reviewable. I can readd some of them if that helps.

I also added a compressed description of the idea at [`edd85c9` (#14408)](https://github.com/openjdk/jdk/pull/14408/commits/edd85c996125ce0a2950b0cfb95d1e387c35d5ff)

koppor:
I addressed all review comments. Tests are green. I also successfully tried with our desktop app (by backporting to JDK21).

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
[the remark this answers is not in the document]
mlchung: Have you considered passing the same ArrayList for saving and restoring local variables?  Currently each method creates one new array list and save the local variables from `firstVariableForDedup` to `nextLocalVar`, but the local variables from `firstVariableForDedup`  are already added to the array list created by the previous method.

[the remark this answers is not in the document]
mlchung: Ah, I see your commit.  I will check it out.

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
```diff
@@ -666,25 +666,120 @@ private void genIncubatorModules(ClassBuilder clb) {
          * Generate bytecode for moduleDescriptors method
          */
         private void genModuleDescriptorsMethod(ClassBuilder clb) {
+            if (moduleInfos.size() <= 75) {
```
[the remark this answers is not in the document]
mlchung: What I meant here is to change the `configure` method to allow the plugin to take an argument for example
```
    --system-modules batch-size=100
```

This argument is optional.   If not specified, the default value is 75.    This way you can tune the batch-size if we hit the method size limit for whatever reason.   And the test can also verify with different size if appropriate.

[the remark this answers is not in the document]
mlchung: `plugins.properties` needs to be updated to show the option added for this plugin.   

BTW, the current argument described in the usage is out-dated which needs update.
```
system-modules.argument=retainModuleTarget
```

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
```diff
@@ -533,7 +540,8 @@ static class SystemModulesClassGenerator {
 
         private static final int MAX_LOCAL_VARS = 256;
 
-        private final int BUILDER_VAR    = 0;
+        private final int BUILDER_VAR    = MAX_LOCAL_VARS + 1;
```
mlchung, line 543: why this one can't use the next available local variables?   i.e.  if no splitting, 1 is the next available slot (btw it looks like using 0 in the current implementation is a bug because this is an instance method).   If module descriptors are done in multiple helper methods, 3 is the next available slot.

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
```diff
@@ -533,7 +540,8 @@ static class SystemModulesClassGenerator {
...
+        private final int DEDUP_LIST_VAR = MAX_LOCAL_VARS + 2;
```
mlchung, line 544: Same comment as BUILDER_VAR.  But this dedup variable list is passed as a parameter to the helper method.  It's already on the local slot 2.

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
```diff
@@ -680,25 +692,138 @@ private void genIncubatorModules(ClassBuilder clb) {
...
+                                                            minfo.packages(),
+                                                            index).build();
+                            }
+                            cob.aload(MD_VAR)
+                               .areturn();
+                        });
+                return;
+            }
+
+            // Instead of adding the module descriptor calls in this method,
```
mlchung, line 718: What about this?

```
            // Split the module descriptors be created by multiple helper methods.
            // Each helper method "subi" creates the maximum N number of module descriptors
            //     mi, m{i+1} ...
            // to avoid exceeding the 64kb limit of method length.  Then it will call
            // "sub{i+1}" to creates the next batch of module descriptors m{i+n}, m{i+n+1}...
            // and so on.  During the construction of the module descriptors, the string sets and
            // modifier sets are deduplicated (see SystemModulesClassGenerator.DedupSetBuilder)
            // and cached in the locals. These locals are saved in an array list so
            // that the helper method can restore the local variables that may be
            // referenced by the bytecode generated for creating module descriptors.
            // Pseudo code looks like this:
            //
            // void subi(ModuleDescriptor[] mdescs, ArrayList<Object> localvars) {
            //      // assign localvars to local variables
            //      var l3 = localvars.get(0);
            //      var l4 = localvars.get(1);
            //        :
            //      // fill mdescs[i] to mdescs[i+n-1]
            //      mdescs[i] = ...
            //      mdescs[i+1] = ...
            //        :
            //      // save new local variables added
            //      localvars.add(lx)
            //      localvars.add(l{x+1})
            //        :
            //      sub{i+i}(mdescs, localvars);
            // }
```

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
```diff
@@ -680,25 +692,138 @@ private void genIncubatorModules(ClassBuilder clb) {
...
+                    splitModuleInfos.add(currentModuleInfos);
+                }
+                currentModuleInfos.add(moduleInfos.get(index));
+            }
+
+            String helperMethodNamePrefix = "sub";
+            ClassDesc arrayListClassDesc = ClassDesc.ofInternalName("java/util/ArrayList");
+
+            int firstVar = nextLocalVar;
+            var wrapper = new Object() {
```
mlchung, line 746: I think you can simply use a local variable at the beginning of each iteration.  See below.

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
```diff
@@ -680,25 +692,138 @@ private void genIncubatorModules(ClassBuilder clb) {
...
+                    currentModuleInfos = new ArrayList<>();
+                    splitModuleInfos.add(currentModuleInfos);
+                }
+                currentModuleInfos.add(moduleInfos.get(index));
+            }
+
+            String helperMethodNamePrefix = "sub";
+            ClassDesc arrayListClassDesc = ClassDesc.ofInternalName("java/util/ArrayList");
+
+            int firstVar = nextLocalVar;
```
mlchung, line 745:
```suggestion
            int dedupVarStart = nextLocalVar;
```

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
```diff
@@ -680,25 +692,138 @@ private void genIncubatorModules(ClassBuilder clb) {
...
+                                   this.classDesc,
+                                   helperMethodNamePrefix + "0",
+                                   MethodTypeDesc.of(CD_void, CD_MODULE_DESCRIPTOR.arrayType(), arrayListClassDesc)
+                           )
                            .areturn();
                     });
+
+            for (int n = 0, count = 0; n < splitModuleInfos.size(); count += splitModuleInfos.get(n).size(), n++) {
+                int index = n;       // the index of which ModuleInfo being processed in the current batch
+                int start = count;   // the start index to the return ModuleDescriptor array for the current batch
```
mlchung, line 777: add `int curDedupVar = nextLocalVar;`

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
```diff
@@ -680,25 +692,138 @@ private void genIncubatorModules(ClassBuilder clb) {
...
+                clb.withMethodBody(
+                        helperMethodNamePrefix + index,
+                        MethodTypeDesc.of(CD_void, CD_MODULE_DESCRIPTOR.arrayType(), arrayListClassDesc),
+                        ACC_PUBLIC,
+                        cob -> {
+                            cob.aload(2)
+                               .astore(DEDUP_LIST_VAR);
```
mlchung, line 784: I think you can keep slot 2 for the dedup var list by changing the start of `nextLocalVar` to 3?  Then you don't need to reassign it to another local.

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
```diff
@@ -680,25 +692,138 @@ private void genIncubatorModules(ClassBuilder clb) {
...
+
+                            if (nextLocalVar > firstVar) {
+                                for (int i = firstVar; i < nextLocalVar; i++) {
```
mlchung, line 787:
```suggestion
                                for (int i = firstLocalVar; i < curLocalVar; i++) {
```

mlchung, line 787:
```suggestion
                            if (curLocalVar > firstLocalVar) {
                                for (int i = firstLocalVar; i < curLocalVar; i++) {
```

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
```diff
@@ -680,25 +692,138 @@ private void genIncubatorModules(ClassBuilder clb) {
...
+                                ModuleInfo minfo = currentBatch.get(j);
+                                new ModuleDescriptorBuilder(cob,
+                                                            minfo.descriptor(),
+                                                            minfo.packages(),
+                                                            start + j).build();
+                            }
+
+                            if (index < splitModuleInfos.size() - 1) {
+                                if ((nextLocalVar > firstVar) && (nextLocalVar > wrapper.lastCopiedVar)) {
+                                    for (int i = wrapper.lastCopiedVar + 1; i < nextLocalVar; i++) {
```
mlchung, line 806: shouldn't it start with `wrapper.lastCopiedVar`? 

```suggestion
                            if (index < splitModuleInfos.size() - 1) {
                                if (nextLocalVar > firstLocalVar && nextLocalVar > curLocalVar) {
                                    for (int i = curLocalVar; i < nextLocalVar; i++) {
```

test/jdk/tools/jlink/JLink100Modules.java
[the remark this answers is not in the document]
mlchung: I see now.  It's used by JImageValidator and that can be replaced by the ClassFile API in the future.  Thanks.

mlchung (commented):
Thanks for the update.   Some comments below.   The test you add does not cause new locals be defined in the helper methods.   Do you think you can add such test case  i.e. new elements are added in the dedup var list by the helper functions?

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
koppor: I tried that multiple times and always failed.  With following hack, I "debugged" the value of nextLocalVar at the beginning of the split:

    Files.writeString(Path.of("C:\\temp\\test.log"), "nextlocavar" + nextLocalVar, StandardOpenOption.CREATE, StandardOpenOption.APPEND);

The values are always 4. I did not dare to overwrite these "magic" slots 1, 2, 3. Despite using IntelliJ's call hierarchy feature, I could not find out where in the code this happens or for what reasons.

I moved `BUILDER_VAR` and `DEDUP_LIST_VAR` two "away".

I think, using `0` was "intentional", because `this` was not needed in the method and thus, this slot could be reused.

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/SystemModulesPlugin.java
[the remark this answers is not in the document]
koppor: Added it at 9835a64018d. This is not for the main `./configure`. To which configure does that go?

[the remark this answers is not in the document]
mlchung: `SystemModulesPlugin::configure`

koppor:
> Thanks for the update. Some comments below. 

Thank you for the feedback! I didn't see the "proper" usage of effectively final variables.

> The test you add does not cause new locals be defined in the helper methods. Do you think you can add such test case i.e. new elements are added in the dedup var list by the helper functions?

I modified the "end-to-end" test in [`15d7448` (#14408)](https://github.com/openjdk/jdk/pull/14408/commits/15d7448b3132e6f0838ac6cb8eb2b939c6eb00f2).

Note that the numbers cannot be arbitrarily increased. Otherwise, one gets errors similar to following

    Fehler: java.lang.IllegalArgumentException: Code length 102397 is outside the allowed range in sub1(ModuleDescriptor[],ArrayList)void

test/jdk/tools/jlink/JLink100Modules.java
```diff
@@ -0,0 +1,125 @@
...
+                .modulePath(out)
+                .output(src.resolve("out-jlink"))
+                .addMods("bug8240567x")
+                .call()
+                .assertSuccess();
+
+        Path binDir = src.resolve("out-jlink").resolve("bin").toAbsolutePath();
+        Path bin = binDir.resolve("java");
+
+        ProcessBuilder processBuilder = new ProcessBuilder(bin.toString(), "-XX:+UnlockDiagnosticVMOptions", "-XX:+BytecodeVerificationLocal", "-m", "bug8240567x/testpackage.JLink100ModulesTest");
```
mlchung, line 118: please wrap this long line into multiple lines.

test/jdk/tools/jlink/JLink100Modules.java
```diff
@@ -0,0 +1,125 @@
...
+        processBuilder.inheritIO();
+        processBuilder.directory(binDir.toFile());
+        Process process = processBuilder.start();
+        int exitCode = process.waitFor();
+        if (exitCode != 0) throw new AssertionError("Exit code is not 0");
```
mlchung, line 123:
```suggestion
        if (exitCode != 0)
             throw new AssertionError("JLink100ModulesTest failed to launch");
```

mlchung (commented):
It's looking pretty good. 

About the test, I don't see `ArrayList::add` in the generated bytecode of `sub2-13`.   The dedup string set is used for the targets of qualified exports and opens and uses.   The modifiers set of `requires` is deduplicated but not the required module names.

I assume you want this be backported.  If so, we should give it some baked time after it's integrated to the main line.    I'm okay if you want to extend the test in a follow up.

test/jdk/tools/jlink/JLink100Modules.java
```diff
@@ -0,0 +1,129 @@
...
+        for (int i = 0; i < 1_000; i++) {
+            String name = "module" + i + "x";
+            Path moduleDir = Files.createDirectories(src.resolve(name));
+
+            StringBuilder moduleInfoContent = new StringBuilder("module ");
+            moduleInfoContent.append(name).append(" {\n");
+            // feed SystemModulesPlugin.SystemModulesClassGenerator.DedupSetBuilder
+            for (int j = 0; j < i % 20; j++) {
+                moduleInfoContent.append("  requires module" + j + "x;\n");
+            }
```
mlchung, line 75: Do you want to keep this change?   or at least take out the comment since this doesn't feed DedupSetBuilder.

Can you add a comment about "the numbers cannot be arbitrarily increased".

mlchung (approved):
I approve this patch.   I'll create a JBS issue to follow up the test update.

test/jdk/tools/jlink/JLink100Modules.java
koppor: I reverted at [`8494517` (#14408)](https://github.com/openjdk/jdk/pull/14408/commits/8494517103c7b303e28c8a8e15fab951b9ea28ea). Think, the comment is then not necessary?

mlchung:
I created https://bugs.openjdk.org/browse/JDK-8311591.   This issue should not be backport until JDK-8311591 is resolved.

mlchung:
@koppor do you have any update on a new test for [JDK-8311591](https://bugs.openjdk.org/browse/JDK-8311591)?

koppor:
> @koppor do you have any update on a new test for [JDK-8311591](https://bugs.openjdk.org/browse/JDK-8311591)?

Currently on a business trip. Will resume work on this next week the latest.

Siedlerchr:
@mlchung Pardon for the delay, we now created a PR with a test case https://github.com/openjdk/jdk/pull/15234

PR #14411, opened 2023-06-12
8309819: Clarify API note in Class::getName and MethodType::toMethodDescriptorString

liach:
The API specification for descriptorString not being a strict inverse of Class::forName and MethodType::fromDescriptorString are not entirely correct.

1. Class::descriptorString was never an inverse of Class::forName, which takes a binary name instead. The note about different class loaders is moved to getName, as ClassDesc requires an explicit lookup for resolution already.
2. MethodType::toMethodDescriptorString ends with a meaningless sentence: "fromMethodDescriptorString, because the latter requires a suitable class loader argument.", and the "Note:" section can be replaced with an `@apiNote`.
3. Both of these didn't mention hidden classes (or other non-nominally-describable classes) as a reason that prevents the inversion operation, in addition to distinct class loaders. Added valid method type descriptor/binary name as a prerequisite for the distinct class loader explanation.

A few user-defined anchor links are replaced with updated javadoc link tag format as well. The explicit html-style links in `@see` tags are unchanged in order to retain the non-code output.

The rendered specifications:
https://cr.openjdk.org/~liach/8309819/09/java.base/java/lang/Class.html
https://cr.openjdk.org/~liach/8309819/09/java.base/java/lang/invoke/MethodType.html

---------
### Progress
- [x] Change requires CSR request [JDK-8309823](https://bugs.openjdk.org/browse/JDK-8309823) to be approved

### Issues
 * [JDK-8309819](https://bugs.openjdk.org/browse/JDK-8309819): Clarify API note in Class::getName and MethodType::toMethodDescriptorString (**Bug** - P4)
 * [JDK-8309823](https://bugs.openjdk.org/browse/JDK-8309823): Clarify API note in Class::getName and MethodType::toMethodDescriptorString (**CSR**)

### Reviewers
 * [Mandy Chung](https://openjdk.org/census#mchung) (@mlchung - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14411/head:pull/14411` \
`$ git checkout pull/14411`

`$ git checkout pull/14411` \
`$ git pull https://git.openjdk.org/jdk.git pull/14411/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14411`

`$ git pr show -t 14411`

</details>

<a href="https://git.openjdk.org/jdk/pull/14411.diff">https://git.openjdk.org/jdk/pull/14411.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14411#issuecomment-1586802710)


src/java.base/share/classes/java/lang/invoke/MethodType.java
```diff
@@ -1223,15 +1224,16 @@ static MethodType fromDescriptor(String descriptor, ClassLoader loader)
...
      *
-     * <p>
-     * Note that this is not a strict inverse of {@link #fromMethodDescriptorString fromMethodDescriptorString}.
+     * @apiNote
+     * This is not a strict inverse of {@link #fromMethodDescriptorString fromMethodDescriptorString}
+     * because the latter requires a suitable class loader argument.
      * Two distinct classes which share a common name but have different class loaders
      * will appear identical when viewed within descriptor strings.
+     * In addition, the result string is not a bytecode descriptor if this
+     * method type cannot be {@linkplain ##descriptor described nominally}.
```
mlchung, line 1233: This method is equivalent to `MethodType::descriptorString` which specifies the result descriptor string if a method type descriptor not.  So I think this  sentence may not be needed.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -4542,8 +4548,7 @@ public Class<?>[] getNestMembers() {
      * ({@code "B", "C", "D", "F", "I", "J", "S", "Z", "V"}) (JVMS {@jvms 4.3.2}).
      *
      * @apiNote
-     * This is not a strict inverse of {@link #forName};
-     * distinct classes which share a common name but have different class loaders
+     * Distinct classes which share a common name but have different class loaders
```
mlchung, line 4551: I think this apiNote can be dropped.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -923,6 +923,12 @@ public boolean isSynthetic() {
      *     returns "[[[[[[[I"
      * </pre></blockquote>
      *
+     * @apiNote
+     * This is not a strict inverse of {@link #forName(String) forName};
+     * distinct classes may share a common name but have different class loaders.
+     * In addition, {@linkplain #isHidden() hidden} classes cannot be discovered
+     * by {@code forName}.
```
mlchung, line 930:
```suggestion
     * This is not a strict inverse of {@link #forName(String) forName}.
     * For normal classes, distinct classes which share a common name but
     * have different class loaders will have the identical name.
     * In addition, {@linkplain #isHidden() hidden} classes cannot be
     * discovered by {@code forName}.
```

mlchung (commented):
Looks good with minor suggestion.

src/java.base/share/classes/java/lang/invoke/MethodType.java
liach: Should we consolidate this distinct class loader note into `descriptorString` too?

mlchung: This method was added prior to `MethodTypeDesc`.    MethodTypeDesc::ofDescriptor` and `MethodTypeDesc::resolveConstantDesc` would be the alternative way to get back the method type.  I tend to think that `descriptorString` would not need that distinct class loader note (as it would use the Lookup class to resolve instead).

liach: Since the method type descriptor info is available in `descriptorString` but it's a reason `toMethodDescriptorString` is not a strict inverse of `fromMethodDescriptorString`, I propose to reword the notes sections to the following:
```
     * @apiNote
     * This is not a strict inverse of {@link #fromMethodDescriptorString
     * fromMethodDescriptorString} which requires a method type descriptor
     * (JVMS {@jvms 4.3.3}) and a suitable class loader argument. Two distinct
     * classes which share a common name but have different class loaders will
     * appear identical when viewed within descriptor strings.
     * <p>
     * This method is included for the benefit of applications that must
     * generate bytecodes that process method handles and {@code invokedynamic}.
```

src/java.base/share/classes/java/lang/invoke/MethodType.java
```diff
@@ -1168,7 +1168,8 @@ Invokers invokers() {
      * <p>
      * Note that it is possible to encounter method types which cannot be
```
mlchung, line 1169: BTW this can be converted to `@apiNote` as well.

src/java.base/share/classes/java/lang/invoke/MethodType.java
mlchung: That's good.

mlchung (approved):

src/java.base/share/classes/java/lang/invoke/MethodType.java
liach: Done.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -923,6 +923,13 @@ public boolean isSynthetic() {
      *     returns "[[[[[[[I"
      * </pre></blockquote>
      *
+     * @apiNote
+     * This is not a strict inverse of {@link #forName(String) forName}.
+     * For normal classes, distinct classes which share a common name but
```
AlanBateman, line 928: Readers may wonder what "normal classes" means here, it might be simpler to just drop introducing that term and just use the structure from the note removed from descriptorString.

liach:
In addition, `MethodType` has a section about "nominal descriptor" while that for `Class` resides in `descriptorString`. Should I add additional links there for occurrences of "described nominally" to `Class::descriptorString`, such as to `Class::describeConstable`, like the links in `MethodType`?

src/java.base/share/classes/java/lang/Class.java
liach: This might wish to refer to "classes that can be described nominally", but same might be said for hidden classes as well. So your suggestion is better.

mlchung:
`Class::descriptorString` specifies clearly whether the result string is a field descriptor and whether it can be used to produce `ClassDesc` for all cases.   I feel that "described nominally" is quite clear what it means.

src/java.base/share/classes/java/lang/Class.java
liach: On a second thought, these are all `getName` results that cannot be parsed by `forName`:
 - Primitive type names
 - Hidden class names
 - Arrays whose component types are hidden classes

I suggest:
```
     * This is not a strict inverse of {@link #forName(String) forName(String)}.
     * Distinct classes which share a common name but have different class
     * loaders will have the identical name. {@linkplain #isPrimitive()
     * Primitive types}, {@linkplain #isHidden() hidden} classes, and arrays
     * with hidden class as element types cannot be discovered by {@code forName}.
```

mlchung: That reads fine with one minor tweak:

... and array classes whose element type is a hidden class or interface cannot be discovered by {@code forName}.

liach:
If this version looks good, I will re-finalize the CSR.

src/java.base/share/classes/java/lang/invoke/MethodType.java
```diff
@@ -1165,10 +1165,12 @@ Invokers invokers() {
...
      * Any class or interface name embedded in the descriptor string will be
      * resolved by the given loader (or if it is null, on the system class loader).
-     * <p>
-     * Note that it is possible to encounter method types which cannot be
+     *
+     * @apiNote
+     * It is possible to encounter method types which cannot be
      * constructed by this method, because their component types are
-     * not all reachable from a common class loader.
+     * not all reachable from a common class loader, or some component types
```
AlanBateman, line 1172: "reachable from a common class loader" pre-dates your change but I think it should say visible to the given class loader.

src/java.base/share/classes/java/lang/invoke/MethodType.java
```diff
@@ -1223,15 +1225,15 @@ static MethodType fromDescriptor(String descriptor, ClassLoader loader)
...
-     * <p>
-     * Note that this is not a strict inverse of {@link #fromMethodDescriptorString fromMethodDescriptorString}.
-     * Two distinct classes which share a common name but have different class loaders
-     * will appear identical when viewed within descriptor strings.
+     * @apiNote
+     * This is not a strict inverse of {@link #fromMethodDescriptorString
+     * fromMethodDescriptorString} which requires a method type descriptor
+     * (JVMS {@jvms 4.3.3}) and a suitable class loader argument. Two distinct
+     * classes which share a common name but have different class loaders will
+     * appear identical when viewed within descriptor strings.
```
AlanBateman, line 1233: "will appear identical when viewed within descriptor strings".  Would it be clearer to say that have equal descriptor strings?

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -923,6 +923,14 @@ public boolean isSynthetic() {
...
      * </pre></blockquote>
      *
+     * @apiNote
+     * This is not a strict inverse of {@link #forName(String) forName(String)}.
+     * Distinct classes which share a common name but have different class
+     * loaders will have the identical name. {@linkplain #isPrimitive()
+     * Primitive types}, {@linkplain #isHidden() hidden} classes, and array
+     * classes whose element type is a hidden class or interface cannot be
+     * discovered by {@code forName}.
+     *
```
AlanBateman, line 933: This is an API note in Class::getName. It seems excessive to be saying that primitive types, hidden classes, ... can't be discovered with Class.forName. It might be better to just drop the second half of this note.

mlchung: I think so too.  The spec describes the result of a hidden class is of the form: `{@code N + '/' + <suffix>}`.  If really needed, we can add in that bullet saying "The result is not a binary name."

liach: The 3-arg forName specifies it uses the 'getName' format. It's good to update that to refer to binary names, but that API can find array classes, and array classes aren't quite binary names.

mlchung:
The api note is primarily regarding about 1-N Class instances from a class name because of different class loader.   For cases that are not a loadable class name or not a type descriptor, the spec of `Class::forName` and `MethodType::fromMethodDescriptorString` should specify clearly.   If not, we should clarify those specs rather than explaining the negative cases in the api note.   

I think  these API specs can be improved especially `forName` doesn't talk about the name for array type which is a fully-qualified name.  I'll create a PR for that and see if that helps the clarification.

mlchung:
For `Class::getName`, I think #14528 should clarify it.

For `MethodType`, I suggest the following:
```
--- a/src/java.base/share/classes/java/lang/invoke/MethodType.java
+++ b/src/java.base/share/classes/java/lang/invoke/MethodType.java
@@ -1161,28 +1161,29 @@ class MethodType
     }
 
     /**
-     * Finds or creates an instance of a method type, given the spelling of its bytecode descriptor.
-     * Convenience method for {@link #methodType(java.lang.Class, java.lang.Class[]) methodType}.
+     * Finds or creates an instance of a method type of the given method descriptor
+     * (JVMS {@jvms 4.3.3}). This method is a convenience method for
+     * {@link #methodType(java.lang.Class, java.lang.Class[]) methodType}.
      * Any class or interface name embedded in the descriptor string will be
-     * resolved by the given loader (or if it is null, on the system class loader).
+     * resolved by the given loader (or if it is {@code null}, on the system class loader).
      *
      * @apiNote
      * It is possible to encounter method types which cannot be
      * constructed by this method, because their component types are
-     * not all reachable from a common class loader, or some component types
-     * cannot be {@linkplain ##descriptor described nominally}.
+     * not visible from a common class loader.
      * <p>
      * This method is included for the benefit of applications that must
      * generate bytecodes that process method handles and {@code invokedynamic}.
      * @param descriptor a bytecode-level type descriptor string "(T...)T"
      * @param loader the class loader in which to look up the types
      * @return a method type matching the bytecode-level type descriptor
-     * @throws NullPointerException if the string is null
-     * @throws IllegalArgumentException if the string is not well-formed
+     * @throws NullPointerException if the string is {@code null}
+     * @throws IllegalArgumentException if the string is not a method descriptor
      * @throws TypeNotPresentException if a named type cannot be found
      * @throws SecurityException if the security manager is present and
      *         {@code loader} is {@code null} and the caller does not have the
      *         {@link RuntimePermission}{@code ("getClassLoader")}
+     * @jvms 4.3.3 Method Descriptors
      */
     public static MethodType fromMethodDescriptorString(String descriptor, ClassLoader loader)
         throws IllegalArgumentException, TypeNotPresentException
@@ -1230,7 +1231,7 @@ class MethodType
      * fromMethodDescriptorString} which requires a method type descriptor
      * (JVMS {@jvms 4.3.3}) and a suitable class loader argument. Two distinct
      * classes which share a common name but have different class loaders will
-     * appear identical when viewed within descriptor strings.
+     * have the identical descriptor strings.
      * <p>
      * This method is included for the benefit of applications that must
      * generate bytecodes that process method handles and {@code invokedynamic}.
```

With these clarification, would you think it's clearer?  CNFE or IAE is thrown for the negative cases and no need to mention in API note.

src/java.base/share/classes/java/lang/invoke/MethodType.java
liach: How about this:
"A method type produced by changing a component type to a distinct class with the same name but a different class loader will have the same descriptor string"

liach:
This patch, rendered API specification, and the CSR have all been updated, taking notes from recommendations in #14528 ([JDK-8310242](https://bugs.openjdk.org/browse/JDK-8310242)). Please review again, thanks!

src/java.base/share/classes/java/lang/invoke/MethodType.java
```diff
@@ -1220,19 +1222,19 @@ static MethodType fromDescriptor(String descriptor, ClassLoader loader)
...
 
     /**
-     * Returns a descriptor string for the method type.  This method
+     * {@return the descriptor string for this method type} This method
      * is equivalent to calling {@link #descriptorString() MethodType::descriptorString}.
      *
-     * <p>
-     * Note that this is not a strict inverse of {@link #fromMethodDescriptorString fromMethodDescriptorString}.
-     * Two distinct classes which share a common name but have different class loaders
-     * will appear identical when viewed within descriptor strings.
```
mlchung, line 1229: Re-reading this, I think this api note reads fine.   OTOH I find a little confusing how you rephrased it.  "A method type produced by changing a component class object to...".   I understand you tried to make it crystal clear two method types may have the identical descriptor strings even if its parameter type or return type are distinct Class objects.   

I think the original version is clear enough and less lengthy.

liach:
```
     * Two distinct classes which share a common name but have different class loaders
     * produce identical component descriptors in descriptor strings.
```
Does this sound good? This also addresses alan's wording concerns.

mlchung: Why can't just say "identical descriptor strings"?   It's unclear what "component descriptor" is.

liach: "component descriptor" tried to refer to both parameter types and return type descriptors. How about "produce identical class descriptors in method descriptors?"

mlchung: `MethodType::descriptorString` refers to method descriptor and `Class::descriptorString` refers to class descriptor.   I don't think it helps to say "produce identical class descriptors in method descriptors".

liach: I think we can put it simple:
"Two distinct classes which share a common name but have different class loaders cannot be distinguished in descriptor strings"
which stands for both class and method descriptors. How does this sound?

mlchung: What about "Two distinct {@code MethodType} objects can have an identical descriptor string as distinct classes can have the same name but different class loaders."?

This api note basically says that from a given descriptor string, `fromMethodDescriptorString` may produce a different `MethodType` with a different class loader.

mlchung (approved):
This version looks good to me.

liach:
I have published the updated javadoc with Class.forName spec update integrated: https://cr.openjdk.org/~liach/8309819/09/java.base/java/lang/Class.html

The CSR has been re-finalized as well.

PR #13082, opened 2023-03-17
8304425: ClassHierarchyResolver from Reflection

liach:
Add API to explore Class Hierarchy with a `ClassLoader` or a `Lookup` with proper privileges, with tests.

This addition is useful in case classes at runtime are not loaded from the system class loader, such as Proxy. This is also useful to APIs that generate bytecode with a `Lookup` object, such as a custom single-abstract-method class implementations from a method handle.

See https://mail.openjdk.org/pipermail/classfile-api-dev/2023-March/000249.html as well.

Current questions, which I wish to discuss with @asotona:
1. Should the resolver fail fast on `IllegalAccessException` from the lookup? This usually indicates the hierarchy resolver is set up improperly, and proceeding may simply yield verification errors in class loading that are hard to track. For bytecode-generating APIs, throwing access errors for the Lookup eagerly is also more preferable than later silent generation failure.
2. Whether the default resolver should be reading from jrt alone, reflection alone, or jrt then reflection. I personally believe reflection alone is more reliable, for classes may redefined with instrumentation or jfr, which may not be reflected in the system resources.
3. In addition, I don't think chaining system class loader reflection after system resource retrieval is really meaningful: is there any case where reflection always works while the system resource retrieval always fails?

PR #14412, opened 2023-06-12
8309838: Classfile API Util.toBinaryName and other cleanup

liach:
Thanks to @exe-boss for the reviews in #13082. This cleanup patch fixed the typos you pointed out and removed other redundancies in classfile Util as well.

### DIFF
--- a/src/java.base/share/classes/jdk/internal/classfile/impl/Util.java
+++ b/src/java.base/share/classes/jdk/internal/classfile/impl/Util.java
@@ -54,10 +54,6 @@ public class Util {
     private Util() {
     }
 
-    public static String arrayOf(CharSequence s) {
-        return "[" + s;
-    }
-
     public static int parameterSlots(MethodTypeDesc mDesc) {
         int count = 0;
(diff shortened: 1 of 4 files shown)


src/java.base/share/classes/jdk/internal/classfile/impl/Util.java
```diff
@@ -85,70 +81,19 @@ public static int maxLocals(int flags, MethodTypeDesc mDesc) {
...
-                }
-            }
-        };
-    }
-
-    public static String returnDescriptor(String s) {
-        return s.substring(s.indexOf(')') + 1);
+    public static String toBinaryName(ClassDesc cd) {
+        return toInternalName(cd).replace('/', '.');
     }
```
ExE-Boss, line 90: Note that this method is called with array class descriptors as well, so the correct implementation would be:
```suggestion
    public static String toBinaryName(ClassDesc cd) {
        return (cd.isArray() ? cd.descriptorString() : toInternalName(cd))
            .replace('/', '.');
    }
```

liach:
@asotona Can you please review this when you are free?

PR #14423, opened 2023-06-12
JDK-8309870: Using -proc:full should be considered requesting explicit annotation processing

jddarcy:
Address a small omission to the recent addition of "javac -proc:full", regard "-proc:full" as explicitly requesting annotation processing. Also add "-A" to the list of options indicating explicit annotation processing was requested.

The intention is to backport this bug fix to JDK 21 too.

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/main/Option.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/main/Option.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2006, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2006, 2023, Oracle and/or its affiliates. All rights reserved.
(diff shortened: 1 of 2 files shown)


src/jdk.compiler/share/classes/com/sun/tools/javac/main/Option.java
```diff
@@ -89,6 +89,11 @@
  * or deletion without notice.</b></p>
  */
 public enum Option {
+    /*
+     * Maintenance note: when adding new annotation processing related
+     * options, the list of options regarded as requesting explicit
+     * annotation processing in JavaCompiler should be updated.
+     */
```
jonathan-gibbons, line 96: (minor) consider moving this option to before the declaration of `Option` so that it seems less like it applies to the first member.  It could even be an `@implNote` on the enum.

jonathan-gibbons (approved):
Approved, with suggestion to relocate comment.

PR #14432, opened 2023-06-12
JDK-8306819: Consider disabling the compiler's default active annotation processing

jddarcy:
Change annotation processing to be opt-in rather than opt-out.

---------
### Progress
- [x] Change requires CSR request [JDK-8317544](https://bugs.openjdk.org/browse/JDK-8317544) to be approved

### Issues
 * [JDK-8306819](https://bugs.openjdk.org/browse/JDK-8306819): Consider disabling the compiler's default active annotation processing (**Enhancement** - P4)
 * [JDK-8317544](https://bugs.openjdk.org/browse/JDK-8317544): Consider disabling the compiler's default active annotation processing (**CSR**)

### Reviewers
 * [Vicente Romero](https://openjdk.org/census#vromero) (@vicente-romero-oracle - **Reviewer**) ⚠️ Review applies to [0ef998fd](https://git.openjdk.org/jdk/pull/14432/files/0ef998fd243c3f8df72ffede17cb9f1c0e439145)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14432/head:pull/14432` \
`$ git checkout pull/14432`

`$ git checkout pull/14432` \
`$ git pull https://git.openjdk.org/jdk.git pull/14432/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14432`

`$ git pr show -t 14432`

</details>

<a href="https://git.openjdk.org/jdk/pull/14432.diff">https://git.openjdk.org/jdk/pull/14432.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14432#issuecomment-1712654334)


jddarcy:
Keep alive.

jddarcy:
With the preparatory work of

8310061: Note if implicit annotation processing is being used

in JDK 21, please review this first cut at removing implicit annotation processing by default.

In this first iteration, I left the javac note in the properties file. In a subsequent iteration, we may decided to keep the old policy or earlier source levels, etc. Depending on the final disposition of that point, the property key should be fully expunged.

Once it is clearer what the eventual policy will be, I'll file the corresponding CSR.

rbygrave:
> Based on my observations, nowadays annotation processing is hardly used anymore and if it's used it mostly occurs after the actual compilation process in a build tool

Annotation processing use is fundamental to newer frameworks like Quarkus, Micronaut and Spring framework is also adopting more use of annotation processing. This is because it enables frameworks and libraries to move some work from runtime to build time for reduced memory consumption and faster startup (replacing reflection, dynamic proxies etc).

For myself I have annotation processors that:
- Perform Dependency Injection 
- Generate JSON adapters (think Jackson without any runtime reflection, all the fastest JSON libs use annotation processing)
- Generate Rest Http Clients (think JAX-RS Jersey client implemented as source code generation)
- Generate Http Server controllers (think JAX-RS Jersey server implementation as source code generation)
- Generate Validation adapters (think Jakarta Validation implemented as source code generation)
- Generate "Builders" for record types

Its also common to use annotation processing for mapping/adapting between types (e.g. MapStruct)

I'd suggest the use of annotation processing is actually taking off in the last say 5 years as we especially look to deploy applications into cloud hosted environments where we pay money for the amount of CPU and Memory used.  Annotation processing means that a **_LOT_** of libraries can move some or even most of their functionality to build time reducing the CPU required to initialise the library and the total memory consumption.


> Based on my observations, nowadays annotation processing is hardly used anymore

To me I see the exact opposite happening and it's driven by cloud adoption and the desire to move work from runtime to build time.

altrisi:
This doesn't remove annotation processors, just disables them running by default with no input from the developer (I assume if they were simply in the compiler classpath, which could happen by accident). As mentioned, it can still be enabled by its flag, and it's likely build systems will set that flag for you if you have declared you use annotation processors for your build.

I also disagree in the hardly used anymore part though, but I have no data.

liach:
@rbygrave This patch isn't aimed to remove AP. This aims to avoid unintentional AP pollutions in classpath, like if one AP is present in a dependency and thus your classpath but you don't want it to run. Such classpath separation have already been promoted by build tools like Gradle.

After all, your build tool should be setting the `-proc` flag for you already, just like setting the classpath flags. You can first upgrade to Java 21, which has patch https://bugs.openjdk.org/browse/JDK-8310061 that warns you about implicit APs, before upgrading to 22 that changes the default behavior.

SentryMan:
> @rbygrave This patch isn't aimed to remove AP. 

Pretty sure he was only speaking against the statement that AP is an unused feature. 

> This aims to avoid unintentional AP pollutions in classpath, like if one AP is present in a dependency and thus your classpath but you don't want it to run. 

In practice is this a thing that happens? Most processors I've seen (like Lombok or record builder for example) are typically added as provided or optional scope dependencies. (and so cannot be transitively pulled)

rob-bygrave:
> This patch isn't aimed to remove AP. This aims to avoid unintentional AP pollutions in classpath

Yes, I understand that.

> Such classpath separation have already been promoted by build tools like Gradle.

**_BUT_** it has **_NOT_** been promoted by **_Maven_**. In the bug report and in this PR I do not see any explicit recognition of how this change will impact the community that is using Maven and I am suggesting that this is a significant proportion of the community.

For example, all the libraries that use annotation processing that document maven use could be impacted along the lines that their documentation will now be at least misleading (and sometimes incorrect). 

This change might indeed be a good long term change but it seems to be being made without any recognition or understanding of how impactful it will be.

SentryMan:
> BUT it has NOT been promoted by Maven. In the bug report and in this PR I do not see any explicit recognition of how this change will impact the community using Maven. I am suggesting that this is a significant proportion of the community.

If this is the new reality I can deal with it, but it would indeed be nice to know that the full gravity of this breaking change was acknowledged and taken into account when making this decision. (The initial proposal seems to make the assertion 
 that annotation processing is a niche tool that isn't used much)

jddarcy:
> For whoever wondering about the impact of this breaking change, in JDK 21, the previous release and the previous LTS release, there's a new warning when the default active AP is used: #14499 https://bugs.openjdk.org/browse/JDK-8310061

To expand on and reiterate points already made by @liach , the goal here is only to change whether or not annotation processing is run implicitly in a case like

`javac -cp $CLASSPATH file`

 Accomplishing this is being broken up over (at least) two releases. In JDK 21, there were two related changes:
 
 [JDK-8308245](https://bugs.openjdk.org/browse/JDK-8308245): Add -proc:full to describe current default annotation processing policy
  [JDK-8310061] (https://bugs.openjdk.org/browse/JDK-8310061): Note if implicit annotation processing is being used

The `-proc:full` option corresponds to the long-standing old default policy on annotation processing. Quoting the note added in the second bug,

    Annotation processing is enabled because one or more processors were found
    on the class path. A future release of javac may disable annotation processing
    unless at least one processor is specified by name (-processor), or a search
    path is specified (--processor-path, --processor-module-path), or annotation
    processing is enabled explicitly (-proc:only, -proc:full).
    Use -Xlint:-options to suppress this message.
    Use -proc:none to disable annotation processing.

The message is a note rather than a waning to avoid tripping up people who run their builds with `-Werror`. Besides a release note in JDK 21, we did extra outreach on the quality discuss alias to publicize this change (https://mail.openjdk.org/pipermail/quality-discuss/2023-July/001122.html).

So, starting in JDK 21, if you want to get the old look-on-the-classpath behavior, it is sufficient to use `-proc:full`. This also will work in JDK 22 after the changes in this PR go back. Breaking the change up over multiple release helps avoid a more disruptive flag day since `-proc:full` can be used before and after to request the desired behavior. HTH

dmlloyd:
> > This aims to avoid unintentional AP pollutions in classpath, like if one AP is present in a dependency and thus your classpath but you don't want it to run.
> 
> In practice is this a thing that happens? Most processors I've seen (like Lombok or record builder for example) are typically added as provided or optional scope dependencies. (and so cannot be transitively pulled)

In fact this is a thing that can cause problems. At present, annotation processor authors have to be cautious about including dependencies because commonly used libraries (such as Guava or Eclipse Collections) can conflict with other processors and/or the user's project with surprising (and sometimes subtle) results.

I'm not sure about other build tools, but Maven does supply a means to include an annotation processor on a (relatively) isolated class path, i.e. separately from the application's class path. Though this is somewhat more cumbersome, it is certainly safer.

SentryMan:
> I'm not sure about other build tools, but Maven does supply a means to include an annotation processor on a (relatively) isolated class path, i.e. separately from the application's class path. Though this is somewhat more cumbersome, it is certainly safer.

Maven's `annotationProcessorPaths` has its own CP issues when compared to using implicit processing. Even now I have code in our processor to check if the processor is [currently running in `annotationProcessorPaths`](https://github.com/avaje/avaje-inject/blob/13ce75156de310d078e1520231fe91cc7d5874b9/inject-generator/src/main/java/io/avaje/inject/generator/ExternalProvider.java#L17-L28) since the cp can get wonky when using it. (especially when dealing with JPMS)

jddarcy:
> > > This aims to avoid unintentional AP pollutions in classpath, like if one AP is present in a dependency and thus your classpath but you don't want it to run.
> > 
> > 
> > In practice is this a thing that happens? Most processors I've seen (like Lombok or record builder for example) are typically added as provided or optional scope dependencies. (and so cannot be transitively pulled)
> 
> In fact this is a thing that can cause problems. At present, annotation processor authors have to be cautious about including dependencies because commonly used libraries (such as Guava or Eclipse Collections) can conflict with other processors and/or the user's project with surprising (and sometimes subtle) results.
> 
> I'm not sure about other build tools, but Maven does supply a means to include an annotation processor on a (relatively) isolated class path, i.e. separately from the application's class path. Though this is somewhat more cumbersome, it is certainly safer.

Yes, in terms of good build hygiene, I recommend separately specifying the path for annotation processors and the path for general compiler dependencies, even if they happen to be the same path.

Likewise,, specifying --release $N is a good practice too.

test/langtools/tools/javac/6341866/T6341866.java
```diff
@@ -67,7 +67,7 @@ enum ImplicitType {
 
     enum AnnoType {
         NONE,           // no annotation processing
-        SERVICE,        // implicit annotation processing, via ServiceLoader
+     // SERVICE,        // implicit annotation processing, via ServiceLoader
```
vicente-romero-oracle, line 70: why keeping this if it won't be used?

jddarcy: > why keeping this if it won't be used?

Sure; I can delete that before the changeset goes back.

AB-xdev:
Oh there is now a PR for my (bug) report. That's nice to see (although I somehow didn't get the notification for it lol).

Anyway here some additional information from my perspective:
<details><summary>Backstory of this issue</summary>

So a while ago we updated a project to Java 17 and we somehow got this error when using ``records``:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.10.1:compile (default-compile) on project maven-module-xxx: Compilation failure
[ERROR] D:\REDACTED\src\main\java\xxx\Job.java: Internal compiler error: java.lang.Exception: javax.lang.model.element.UnknownElementException: Unknown element: "de.xxx.Parameters" at org.eclipse.jdt.internal.compiler.apt.dispatch.RoundDispatcher.handleProcessor(RoundDispatcher.java:172)
[ERROR] -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.10.1:compile (default-compile) on project maven-module-xxx: Compilation failure
D:\REDACTED\src\main\java\xxx\Job.java: Internal compiler error: java.lang.Exception: javax.lang.model.element.UnknownElementException: Unknown element: "de.xxx.Parameters" at org.eclipse.jdt.internal.compiler.apt.dispatch.RoundDispatcher.handleProcessor(RoundDispatcher.java:172)

    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:215)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:148)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
    at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:56)
    at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:128)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:305)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:192)
    at org.apache.maven.DefaultMaven.execute (DefaultMaven.java:105)
    at org.apache.maven.cli.MavenCli.execute (MavenCli.java:957)
    at org.apache.maven.cli.MavenCli.doMain (MavenCli.java:289)
    at org.apache.maven.cli.MavenCli.main (MavenCli.java:193)
    at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)
    at jdk.internal.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:77)
    at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke (DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke (Method.java:568)
    at org.codehaus.plexus.classworlds.launcher.Launcher.launchEnhanced (Launcher.java:282)
    at org.codehaus.plexus.classworlds.launcher.Launcher.launch (Launcher.java:225)
    at org.codehaus.plexus.classworlds.launcher.Launcher.mainWithExitCode (Launcher.java:406)
    at org.codehaus.plexus.classworlds.launcher.Launcher.main (Launcher.java:347)
    at org.codehaus.classworlds.Launcher.main (Launcher.java:47)
Caused by: org.apache.maven.plugin.compiler.CompilationFailureException: Compilation failure
D:\REDACTED\src\main\java\xxx\Job.java: Internal compiler error: java.lang.Exception: javax.lang.model.element.UnknownElementException: Unknown element: "de.xxx.Parameters" at org.eclipse.jdt.internal.compiler.apt.dispatch.RoundDispatcher.handleProcessor(RoundDispatcher.java:172)

    at org.apache.maven.plugin.compiler.AbstractCompilerMojo.execute (AbstractCompilerMojo.java:1310)
    at org.apache.maven.plugin.compiler.CompilerMojo.execute (CompilerMojo.java:198)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:137)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:210)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:148)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
    at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:56)
    at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:128)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:305)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:192)
    at org.apache.maven.DefaultMaven.execute (DefaultMaven.java:105)
    at org.apache.maven.cli.MavenCli.execute (MavenCli.java:957)
    at org.apache.maven.cli.MavenCli.doMain (MavenCli.java:289)
    at org.apache.maven.cli.MavenCli.main (MavenCli.java:193)
    at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)
    at jdk.internal.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:77)
    at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke (DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke (Method.java:568)
    at org.codehaus.plexus.classworlds.launcher.Launcher.launchEnhanced (Launcher.java:282)
    at org.codehaus.plexus.classworlds.launcher.Launcher.launch (Launcher.java:225)
    at org.codehaus.plexus.classworlds.launcher.Launcher.mainWithExitCode (Launcher.java:406)
    at org.codehaus.plexus.classworlds.launcher.Launcher.main (Launcher.java:347)
    at org.codehaus.classworlds.Launcher.main (Launcher.java:47)
```

The error was kind of confusing and the only thing hinting to a annotation processor was the ``handleProcessor`` in the stacktrace.

As it later turned out this was caused by [``org.hibernate.validator:hibernate-validator-annotation-processor``](https://mvnrepository.com/artifact/org.hibernate.validator/hibernate-validator-annotation-processor) (→ https://hibernate.atlassian.net/browse/HV-1863) which was somehow on the classpath.

So there was an annotation processor that was kind of magically executed with no log output so there was 0 chance of finding out that it was even executed, so that's one of the reasons why I created this issue.

</details>

<details><summary>About your discussion of AP pollutions in classpath</summary>

> > This aims to avoid unintentional AP pollutions in classpath, like if one AP is present in a dependency and thus your classpath but you don't want it to run.
> In practice is this a thing that happens? Most processors I've seen (like Lombok or record builder for example) are typically added as provided or optional scope dependencies. (and so cannot be transitively pulled)

> Is it really such a common thing that people accidentally include unrelated annotation processor dependencies transitively via maven?

I can show a very simple example:
[Log4J2s PluginProcessor](https://github.com/apache/logging-log4j2/blob/c7f0f15dd6d37fc8f23689b975163ec2df80a9ec/log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/processor/PluginProcessor.java) (registered via Java Service Loading [here](https://github.com/apache/logging-log4j2/blob/c7f0f15dd6d37fc8f23689b975163ec2df80a9ec/log4j-core/src/main/resources/META-INF/services/javax.annotation.processing.Processor)) is likely included and executed in a lot of projects that don't require and know about it.

> Yes, in terms of good build hygiene, I recommend separately specifying the path for annotation processors and the path for general compiler dependencies, even if they happen to be the same path.

There is 2 problems with this statement:
1. The average developer that compiles with Maven oder Gradle doesn't know about this "good practice" and there is nothing hinting to it. Update: This is (partially) addressed with [JDK-8310061](https://bugs.openjdk.org/browse/JDK-8310061)
2. Even if this is the case and you split it into to separate "paths" there is no way in preventing e.g. Log4J2s PluginProcessor getting into the "general compiler dependencies" because it's inside log4j-core.jar (without this jar logging no longer works)

</details>

I again want to say that my intention was not to remove Annotation processing but simply considering disabling it by default due to the security concerns mentioned in the issue.

Also thank you for your efforts and the progress already done with [JDK-8310061](https://bugs.openjdk.org/browse/JDK-8310061) I think the new message is already a good step into the right direction :)

vicente-romero-oracle (approved):
lgtm

DanielThomas:
After much head scratching attempting to run the benchmarks for 8309130, I realised this change broke microbenchmarks:
```
$ make test TEST="micro:java.lang.ArraysSort"
Building target 'test' in configuration 'linux-x86_64-server-release'
...
Test selection 'micro:java.lang.ArraysSort', will run:
* micro:java.lang.ArraysSort

Running test 'micro:java.lang.ArraysSort'
Exception in thread "main" java.lang.RuntimeException: ERROR: Unable to find the resource: /META-INF/BenchmarkList
	at org.openjdk.jmh.runner.AbstractResourceReader.getReaders(AbstractResourceReader.java:98)
	at org.openjdk.jmh.runner.BenchmarkList.find(BenchmarkList.java:124)
	at org.openjdk.jmh.runner.Runner.internalRun(Runner.java:252)
	at org.openjdk.jmh.runner.Runner.run(Runner.java:208)
	at org.openjdk.jmh.Main.main(Main.java:71)
Finished running test 'micro:java.lang.ArraysSort'
Test report is stored in build/linux-x86_64-server-release/test-results/micro_java_lang_ArraysSort
```

forax:
Yes, JMH is using an annotation processor which now does not run anymore

jaikiran:
Hello @DanielThomas, thank you reporting and narrowing down the microbenchmark failures. This is now being addressed in https://bugs.openjdk.org/browse/JDK-8317802.

jddarcy:
> Hello @DanielThomas, thank you reporting and narrowing down the microbenchmark failures. This is now being addressed in https://bugs.openjdk.org/browse/JDK-8317802.

Thanks for taking care of that issue @jaikiran .

jddarcy:
> > This patch isn't aimed to remove AP. This aims to avoid unintentional AP pollutions in classpath
> 
> Yes, I understand that.
> 
> > Such classpath separation have already been promoted by build tools like Gradle.
> 
> **_BUT_** it has **_NOT_** been promoted by **_Maven_**. In the bug report and in this PR I do not see any explicit recognition of how this change will impact the community that is using Maven and I am suggesting that this is a significant proportion of the community.
> 
> For example, all the libraries that use annotation processing that document maven use could be impacted along the lines that their documentation will now be at least misleading (and sometimes incorrect).
> 
> This change might indeed be a good long term change but it seems to be being made without any recognition or understanding of how impactful it will be.

Maven is listed as a participant in the OpenJDK Quality Outreach effort (https://wiki.openjdk.org/display/quality/Quality+Outreach).

The prior change of the javac emitting the note was explicitly raised on the quality discuss alias in July 2023 (https://mail.openjdk.org/pipermail/quality-discuss/2023-July/thread.html), ahead of the GA of JDK 21 in September.

The change in this PR is likewise publicized this the October 2023 quality discuss message (https://mail.openjdk.org/pipermail/quality-discuss/2023-October/001129.html), months ahead of the expected GA of JDK 23 in March 2024.

SentryMan:
Just out of curiosity though, why didn't this (pretty sizable breaking change) get a JEP? The dynamic agents being disabled by default got one.

jddarcy:
> Just out of curiosity though, why didn't this (pretty sizable breaking change) get a JEP? The dynamic agents being disabled by default got one.

The short answer is despite the Sturm und Drang of any discussions Reddit, I don't agree with the assessment that this PR constitutes a "pretty sizable breaking change."

To summarize: the capabilities of javac with respect to annotation processing are not changing in any way. The only aspect that is changing is what is the default policy if no explicit annotation processing configuration options are given.

If a build environment is already using any of the javac options `-processor`, `--processor-path`, `--processor-module-path`, or `-proc:only` nothing changes.

If a build environment does not have annotation processors on the class path, nothing changes.

In JDK 21, javac prints a note -- not a warning for a reason I'll explain below -- to identify the situations where implicit annotation processing is being relied on. Now it is true that users might not see this note, but communication about the JDK 21 change was publicized on the quality-discuss list to help make sure interested parties were informed of the change. The reason the javac message is a note and not a warning is so that build environments running under `-Werror` will not (spuriously) fail solely due to communicating the possibility of a future policy change.

Starting in JDK 21, a build environment or other javac invocation can be explicitly configured to request the old default policy by using `-proc:full`. All the other existing annotation processing configuration options (`-processor`, `-processor-path`, ...) are still supported and mean exactly the same thing as before.

Only in JDK 22 will the default policy actually be change to _not_ look on the class path for annotation processors (unless `-proc:full` is also given). However, there need not be any flag day for changing to JDK 22 specifically since the option to request the old policy is accepted in JDK 21.

rbygrave:
> there need not be any flag day for changing to JDK 22 specifically since the option to request the old policy is accepted in JDK 21

Just to say, most of our libs build against **11, 17, 21 and EA/22**. So for this reason our projects we can't trivially make use of `-proc:full` introduced in 21 as that fails the builds using 11 and 17 and instead we needed to introduce build profiles for this.

jddarcy:
> > there need not be any flag day for changing to JDK 22 specifically since the option to request the old policy is accepted in JDK 21
> 
> Just to say, most of our libs build against **11, 17, 21 and EA/22**. So for this reason our projects we can't trivially make use of `-proc:full` introduced in 21 as that fails the builds using 11 and 17 and instead we needed to introduce build profiles for this.

Noted; thanks.

jddarcy:
> > I have learned very late (about a month ago, and through JMH bug reports!) about this change.
> > Looking at incoming bug reports, I can see this is looking to break significant amount of day-to-day things that rely on annotation processing to work by default. This includes existing JMH benchmarks, existing JCStress tests, and that is only OpenJDK tools so far! The proposed fix, "enable the annotation processing in build" seems to underestimate how many already existing build configs _are there_, and how widespread annotation processing is. So this is a harbinger of a widespread breakage when updating to JDK 22 and beyond.
> > Yet, I see only tiny discussion about this in either here or in the CSR. It mostly just states that disabling annotation processors is the right thing to do to solve the externally reported problematic scenario. Normally, given the impact for changing a policy like this, I would have expected to see a JEP-sized discussion that weighs pros and cons for mitigation strategies, polls what problems real projects would run into, discusses to what extent we want to deal with currently supported JDK releases, etc.
> > After more digging, the only (?) discussions I could find is [Reddit post](https://www.reddit.com/r/java/comments/17f7dha/consider_disabling_the_compilers_default_active/) and this PR comments, which contain some valid questions, concerns, and scenarios the discussion for a change like this should consider.
> > Yes, there was a warning message. This only highlights that hardly anyone reads the warning messages, especially buried somewhere in CI logs. This also does not capture the unfortunate reality that JDK 21 only starts to see major use, so most users did not even had the opportunity to see the warning message! In other words, warning messages are inefficient tools to bring quick attention to the issue that proposes to change the JDK behavior in a considerable way.
> > IMO, JEPs are done for this reason, even for less impactful things. So I suggest we revert [JDK-8306819](https://bugs.openjdk.org/browse/JDK-8306819) from JDK 22, and start over with the JEP targeting JDK 23 or even 24. I believe this would be a right thing to do at this point.
> 
> Thanks for the input @shipilev ; I'm looking into options to amend this change later in JDK 22.

Upon further reflection, we'll be deferring the policy change to disable implicit annotation processing from JDK 22 to JDK 23. The changeset to re-enable the old policy is out for review as https://github.com/openjdk/jdk/pull/16988.

In addition, we plan to backport support for the `-proc:full` option to our update releases: https://bugs.openjdk.org/browse/JDK-8308245

jddarcy:
> > > there need not be any flag day for changing to JDK 22 specifically since the option to request the old policy is accepted in JDK 21
> > 
> > 
> > Just to say, most of our libs build against **11, 17, 21 and EA/22**. So for this reason our projects we can't trivially make use of `-proc:full` introduced in 21 as that fails the builds using 11 and 17 and instead we needed to introduce build profiles for this.
> 
> Noted; thanks.

PS Support for -proc:full has been backported to Oracle's 17u, 11u, and 8u release trains (https://bugs.openjdk.org/browse/JDK-8321416, https://bugs.openjdk.org/browse/JDK-8321418, https://bugs.openjdk.org/browse/JDK-8321419) as well as the OpenJDK 17u and 11u release trains (https://bugs.openjdk.org/browse/JDK-8324670, https://bugs.openjdk.org/browse/JDK-8324804). The feature is on track to appear in the April 2024 set of releases.

jddarcy:
FYI, PR to implement the change in annotation processing default policy in JDK 23 out for review:

https://github.com/openjdk/jdk/pull/19331

see also discussion thread

https://mail.openjdk.org/pipermail/jdk-dev/2024-May/009028.html

PR #14440, opened 2023-06-13
8310275: Bug in assignment operator of ReservedMemoryRegion

schmelter-sap:
---------
### Progress

### Issue
 * [JDK-8310275](https://bugs.openjdk.org/browse/JDK-8310275): Bug in assignment operator of ReservedMemoryRegion (**Bug** - P4)

### Reviewers
 * [Johan Sjölen](https://openjdk.org/census#jsjolen) (@jdksjolen - **Reviewer**)
 * [Thomas Stuefe](https://openjdk.org/census#stuefe) (@tstuefe - **Reviewer**) ⚠️ Review applies to [4d4138d0](https://git.openjdk.org/jdk/pull/14440/files/4d4138d0916177dcd58e1345bb9df24ea1aea5a7)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14440/head:pull/14440` \
`$ git checkout pull/14440`

`$ git checkout pull/14440` \
`$ git pull https://git.openjdk.org/jdk.git pull/14440/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14440`

`$ git pr show -t 14440`

</details>

<a href="https://git.openjdk.org/jdk/pull/14440.diff">https://git.openjdk.org/jdk/pull/14440.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14440#issuecomment-1619728959)


jdksjolen (commented):
Hi,

Instead of having this be `TEST_VM`, why not simply have it be a regular `TEST`? Then you avoid interference from the VM. You'd only have to ensure that `VirtualMemoryTracker::initailize` is called first.

schmelter-sap:
Calling just VirtualMemoryTracker::initialize() is not enough. E.g. the vm page size is not initialized and this leads to errors.

dholmes-ora (commented):
Functional fix looks good - how would this error manifest in practice? Could any existing test be affected?

Thanks.

schmelter-sap:
> ... how would this error manifest in practice? Could any existing test be affected?

We found it during gtests on a JDK 17 alpine build, since in this version the JavaTestThread class overwrote the post:_run() method without deregistering the stack at the VirtualMemoryManager. JDK-8215948 fixed this, so I would not expect any tests to be affected.

Outside of test the code is only executed if a thread calls AttachCurrentThread() and exits without calling DetachCurrentThread.().

dholmes-ora:
> > ... how would this error manifest in practice? Could any existing test be affected?
> 
> Outside of test the code is only executed if a thread calls AttachCurrentThread() and exits without calling DetachCurrentThread.().

Sorry but how can a thread not detaching before exiting cause our buggy RMR assignment operator to be called??

schmelter-sap:
> Sorry but how can a thread not detaching before exiting cause our buggy RMR assignment operator to be called??

In the VirtualMemoryTracker::add_reserved_region() method we detect that a newly registered reserved region if overlapping with an already existing reserved region. If that already existing region has type mtThreadStack we assume that the thread ended without notifying the virtual memory tracker and fix this by assigning the new reserved region to the old one. And that is where the assignment operator is called (see https://github.com/openjdk/jdk/blob/e9daf4a0185b90762d2bdd38d86fe93b4822ea08/src/hotspot/share/services/virtualMemoryTracker.cpp#L373).

tstuefe:
> > Sorry but how can a thread not detaching before exiting cause our buggy RMR assignment operator to be called??
> 
> In the VirtualMemoryTracker::add_reserved_region() method we detect that a newly registered reserved region if overlapping with an already existing reserved region. If that already existing region has type mtThreadStack we assume that the thread ended without notifying the virtual memory tracker and fix this by assigning the new reserved region to the old one. And that is where the assignment operator is called (see
> 
> https://github.com/openjdk/jdk/blob/e9daf4a0185b90762d2bdd38d86fe93b4822ea08/src/hotspot/share/services/virtualMemoryTracker.cpp#L373
> ).

And libc's typically cache and reuse thread stacks, so this error is not unlikely.

tstuefe (commented):
Fix itself is fine, obviously. This is an old bug and we should backport it.

I propose a simpler fix that tests the fixed assignment directly:

```
TEST_VM(NMT, ReservedRegionCopy) {
  address dummy1 = (address)0x10000000;
  ReservedMemoryRegion region1(dummy1, os::vm_page_size(), CALLER_PC);
  region1.add_committed_region(dummy1, os::vm_page_size(), CALLER_PC);

  address dummy2 = (address)0x20000000;
  ReservedMemoryRegion region2(dummy2, os::vm_page_size(), CALLER_PC);
  region2.add_committed_region(dummy2, os::vm_page_size(), CALLER_PC);

  region2 = region1;

  CommittedRegionIterator itr = region2.iterate_committed_regions();
  const CommittedMemoryRegion* rgn = itr.next();
  ASSERT_EQ(rgn->base(), dummy1); // Now we should see dummy1
  rgn = itr.next();
  ASSERT_EQ(rgn, (const CommittedMemoryRegion*)nullptr); // and nothing else
}
```

Tested, fails with stock and succeeds with your patch.

Could be probably dumbed down even further if we extended the RMR interface to expose committed regions.

Could also be extended to check that the other properties are copied correctly.

tstuefe (approved):
Great, thanks for the fix!

Cheers Thomas

jdksjolen (approved):
Happy to see a much improved test, thanks for the hard work.

LGTM.

dholmes-ora (approved):

tstuefe:
Sure, go ahead.

PR #14444, opened 2023-06-13
6664309: Docking point of a floating toolbar changes after closing

prsadhuk:
Issue is if we create a floating JToolBar at the EAST side and drag it and then if we close the floating toolbar, then it gets docket at NORTH side instead of the EAST side where it was docked before being floated.
This is because of some contentious code in BasicToolBarUI where we change the constaint to NORTH if it is EAST/WEST.
I could not find any JBS which did that code change.

Albeit, removing that piece of code does not cause regression in "clientlibs" jtreg/jck tests in addition to solve this anomaly, so thought, removing the code seems prudent..

---------
### Progress

### Issue
 * [JDK-6664309](https://bugs.openjdk.org/browse/JDK-6664309): Docking point of a floating toolbar changes after closing (**Bug** - P4)

### Reviewers
 * [Tejesh R](https://openjdk.org/census#tr) (@TejeshR13 - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14444/head:pull/14444` \
`$ git checkout pull/14444`

`$ git checkout pull/14444` \
`$ git pull https://git.openjdk.org/jdk.git pull/14444/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14444`

`$ git pr show -t 14444`

</details>

<a href="https://git.openjdk.org/jdk/pull/14444.diff">https://git.openjdk.org/jdk/pull/14444.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14444#issuecomment-1588982750)


test/jdk/javax/swing/JToolBar/TestToolBarConstraint.java
```diff
@@ -0,0 +1,101 @@
...
+
+import java.awt.BorderLayout;
+import javax.swing.JButton;
+import javax.swing.JFrame;
+import javax.swing.JPanel;
+import javax.swing.JToolBar;
+import javax.swing.GroupLayout;
+import javax.swing.SwingUtilities;
+import javax.swing.SwingConstants;
+import javax.swing.WindowConstants;
```
TejeshR13, line 41: Importing `WindowConstants` can be removed, since its not used.

test/jdk/javax/swing/JToolBar/TestToolBarConstraint.java
```diff
@@ -0,0 +1,101 @@
...
+                PassFailJFrame.addTestWindow(frame);
+                PassFailJFrame.positionTestWindow(frame, PassFailJFrame.Position.HORIZONTAL);
+            } catch (Exception e) {
+                e.printStackTrace();
+            }
+        });
+        passFailJFrame.awaitAndCheck();
+    }
+
+    private static void creatUI() {
```
TejeshR13, line 69: Typo for method name(`createUI`).

test/jdk/javax/swing/JToolBar/TestToolBarConstraint.java
prsadhuk: ok

TejeshR13 (approved):

prrace:
So there's two things at play here
(1) the orientation property set on the toolbar, which governs how the components on the toolbar are laid out.
(2) the special interpretation by JToolBar of being laid out on one side of a BorderLayout.
I don't know that I really understand why
>For drag-out to work correctly, it is recommended that you add JToolBar instances to one of the four "sides"
> of a container whose layout manager is a BorderLayout, and do not add children to any of the other four "sides".

my guess is that it just has the right characteristics - the CENTER component would be free to use for the rest
of the window, and empty sides are not an issue, and you can easily make it undock from any edge, and redock
on any edge without affecting layout of the rest of the window.

I've played around with it and it seems to me that the intent of the code you are removing is that
a HORIZONTAL layout implies it was docked on a side that would be HORIZONTAL - but in that
case why NORTH and not SOUTH ?

And I've verified that its perfectly possible, if a bit weird, to have HORIZONTAL toolbar on the EAST
i.e toolbar buttons laid out left to right.

So conflating orientation of the toolbar with the edge on which it is docked is probably wrong.

BTW you should add to the bug evaluation that this code was introduced by the fix for

https://bugs.openjdk.org/browse/JDK-4700351

The reasoning there seems to be a bit back and forth and confusing as to what views won out,
but I do not agree with all the comment.
For example I don't agree that "Also, when placed in a BorderLayout, the orientation should change
to the appropriate value, even if the orientation has been set explicitly"

But PROBABLY most people putting a toolbar on the EAST will make it vertical so I expect few apps
to see an unwelcome change in behaviour here .. but there's always a risk of that.

prrace (approved):

prsadhuk:
> But PROBABLY most people putting a toolbar on the EAST will make it vertical so I expect few apps
> to see an unwelcome change in behaviour here .. but there's always a risk of that.

I am not sure on this point being "unwelcome change"....I think even if there is a change in behaviour, it is as per expectation as 
If a user docks a "horizontal" toolbar in EAST and then it is dragged and floated and then closed, I think it should be expected to dock the toolbar as it is ie HORIZONTAL in EAST side, that is what it was before docking. If orientation is changed to VERTICAL while in floating state and then closed, then the toolbar should be docked at EAST in VERTICAL state

As you correctly pointed out it is not correct that `"the orientation should change to the appropriate value, even if the orientation has been set explicitly"` so orientation and docking point should not be changed by the implementation.
Also, it was mentioned in that bug that `"The documentation should be changed to clarify this. "` which was not done even though the implementation was done so it seems prudent to rectify the implementation.

PR #19, opened 2020-09-05
8252657: JVMTI agent is not unloaded when Agent_OnAttach is failed

YaSuenag:
If `Agent_OnAttach()` in JVMTI agent which is attempted to load via JVMTI.agent_load dcmd is failed, it would not be unloaded.  
We've [discussed it on serviceability-dev](https://mail.openjdk.java.net/pipermail/serviceability-dev/2020-September/032839.html). This PR is a continuation of that.

This PR also includes to call `Agent_OnUnload()` when `Agent_OnAttach()` failed.

How to reproduce:

1. Build JVMTI agent for test
```
$ git clone https://github.com/YaSuenag/jvmti-examples.git
$ cd jvmti-examples/helloworld/out/build
$ cmake ../..
```

2. Run JShell

3. Load JVMTI agent via `jcmd <PID> JVMTI.agent_load` with "error" ("error" means `Agent_OnAttach()` returns JNI_ERR)
```
$ jcmd
89456 jdk.jshell.execution.RemoteExecutionControl 45651
89547 sun.tools.jcmd.JCmd
89436 jdk.jshell/jdk.internal.jshell.tool.JShellToolProvider
$ jcmd 89436 JVMTI.agent_load `pwd`/libhelloworld.so error
89436:
return code: -1
```

4. Check loaded libraries via `jcmd <PID> VM.dynlibs`
```
$ jcmd 89436 VM.dynlibs | grep libhelloworld
7f2f8b06b000-7f2f8b06c000 r--p 00000000 fd:00 11818202 /home/ysuenaga/github/jvmti-examples/helloworld/out/build/libhelloworld.so
7f2f8b06c000-7f2f8b06d000 r-xp 00001000 fd:00 11818202 /home/ysuenaga/github/jvmti-examples/helloworld/out/build/libhelloworld.so
7f2f8b06d000-7f2f8b06e000 r--p 00002000 fd:00 11818202 /home/ysuenaga/github/jvmti-examples/helloworld/out/build/libhelloworld.so
7f2f8b06e000-7f2f8b06f000 r--p 00002000 fd:00 11818202 /home/ysuenaga/github/jvmti-examples/helloworld/out/build/libhelloworld.so
7f2f8b06f000-7f2f8b070000 rw-p 00003000 fd:00 11818202 /home/ysuenaga/github/jvmti-examples/helloworld/out/build/libhelloworld.so
```

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
[...]

PR #14445, opened 2023-06-13
8309258: RISC-V: Add riscv_hwprobe syscall

robehn:
Hi, please consider.

Linux kernel 6.4 will come with the new syscall https://docs.kernel.org/riscv/hwprobe.html to determine CPU and features.
RV cpus features/capabilities can vastly differ and it is not feasible for users to manually enable the correct feature set.
Today the VM uses the ELF aux vector and cpuinfo to gather some information about CPU capabilities.

Currently features are track with a bit field, this is insufficient.
There are many capabilities and these can have values attached to them.
CPU features should also be possible to turn if we can determine vendor (hwprobe).

This patchs adds the syscall, uses the syscall in combination with the aux and cpuinfo to enable features by default.
If there is a vendor specific path it calls that in addition.
Then we build the feature string(and bit field) and update flags accordingly.

Tested t1 and hwprobe with:
https://lore.kernel.org/qemu-devel/7f8d733df6e9b6151e9efb843d55441348805e70.camel@rivosinc.com/

---------
### Progress

### Issue
 * [JDK-8309258](https://bugs.openjdk.org/browse/JDK-8309258): RISC-V: Add riscv_hwprobe syscall (**Enhancement** - P4)

### Reviewers
 * [Feilong Jiang](https://openjdk.org/census#fjiang) (@feilongjiang - Author) ⚠️ Review applies to [55f2462c](https://git.openjdk.org/jdk/pull/14445/files/55f2462c35e6a8d2cf3d3672b60947d29febe557)
 * [Thomas Stuefe](https://openjdk.org/census#stuefe) (@tstuefe - **Reviewer**) ⚠️ Review applies to [eb8fe28c](https://git.openjdk.org/jdk/pull/14445/files/eb8fe28ce91b8822a1f634688c8f1c52912b25c1)
 * [Fei Yang](https://openjdk.org/census#fyang) (@RealFYang - **Reviewer**) ⚠️ Review applies to [55f2462c](https://git.openjdk.org/jdk/pull/14445/files/55f2462c35e6a8d2cf3d3672b60947d29febe557)
 * [Ludovic Henry](https://openjdk.org/census#luhenry) (@luhenry - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14445/head:pull/14445` \
`$ git checkout pull/14445`

`$ git checkout pull/14445` \
`$ git pull https://git.openjdk.org/jdk.git pull/14445/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14445`

`$ git pr show -t 14445`

</details>

<a href="https://git.openjdk.org/jdk/pull/14445.diff">https://git.openjdk.org/jdk/pull/14445.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14445#issuecomment-1588990658)

### DIFF
--- a/src/hotspot/cpu/riscv/vm_version_riscv.hpp
+++ b/src/hotspot/cpu/riscv/vm_version_riscv.hpp
@@ -111,35 +111,35 @@ class VM_Version : public Abstract_VM_Version {
   // unaligned_access Unaligned memory accesses (unknown, unspported, emulated, slow, firmware, fast)
   // satp mode SATP bits (number of virtual addr bits) mbare, sv39, sv48, sv57, sv64
 
-  #define NO_BIT (BitsPerWord+1) // nth_bit will return 0 on values larger than BitsPerWord
+  #define RV_NO_FLAG_BIT (BitsPerWord+1) // nth_bit will return 0 on values larger than BitsPerWord
 
   // declaration name  , extension name,    bit pos ,in str, mapped flag)
-  #define RV_FEATURE_FLAGS(decl)                                                             \
-  decl(ext_I           , "i"           , ('I' - 'A'), true , NO_UPDATE_DEFAULT)              \
-  decl(ext_M           , "m"           , ('M' - 'A'), true , NO_UPDATE_DEFAULT)              \
-  decl(ext_A           , "a"           , ('A' - 'A'), true , NO_UPDATE_DEFAULT)              \
-  decl(ext_F           , "f"           , ('F' - 'A'), true , NO_UPDATE_DEFAULT)              \
-  decl(ext_D           , "d"           , ('D' - 'A'), true , NO_UPDATE_DEFAULT)              \
-  decl(ext_C           , "c"           , ('C' - 'A'), true , UPDATE_DEFAULT(UseRVC))         \
-  decl(ext_Q           , "q"           , ('Q' - 'A'), true , NO_UPDATE_DEFAULT)              \
-  decl(ext_H           , "h"           , ('H' - 'A'), true , NO_UPDATE_DEFAULT)              \
-  decl(ext_V           , "v"           , ('V' - 'A'), true , UPDATE_DEFAULT(UseRVV))         \
-  decl(ext_Zicbom      , "Zicbom"      ,      NO_BIT, true , UPDATE_DEFAULT(UseZicbom))      \
-  decl(ext_Zicboz      , "Zicboz"      ,      NO_BIT, true , UPDATE_DEFAULT(UseZicboz))      \
-  decl(ext_Zicbop      , "Zicbop"      ,      NO_BIT, true , UPDATE_DEFAULT(UseZicbop))      \
-  decl(ext_Zba         , "Zba"         ,      NO_BIT, true , UPDATE_DEFAULT(UseZba))         \
-  decl(ext_Zbb         , "Zbb"         ,      NO_BIT, true , UPDATE_DEFAULT(UseZbb))         \
-  decl(ext_Zbc         , "Zbc"         ,      NO_BIT, true , NO_UPDATE_DEFAULT)              \
-  decl(ext_Zbs         , "Zbs"         ,      NO_BIT, true , UPDATE_DEFAULT(UseZbs))         \
-  decl(ext_Zicsr       , "Zicsr"       ,      NO_BIT, true , NO_UPDATE_DEFAULT)              \
-  decl(ext_Zifencei    , "Zifencei"    ,      NO_BIT, true , NO_UPDATE_DEFAULT)              \
-  decl(ext_Zic64b      , "Zic64b"      ,      NO_BIT, true , UPDATE_DEFAULT(UseZic64b))      \
-  decl(ext_Zihintpause , "Zihintpause" ,      NO_BIT, true , UPDATE_DEFAULT(UseZihintpause)) \
-  decl(mvendorid       , "VendorId"    ,      NO_BIT, false, NO_UPDATE_DEFAULT)              \
-  decl(marchid         , "ArchId"      ,      NO_BIT, false, NO_UPDATE_DEFAULT)              \
-  decl(mimpid          , "ImpId"       ,      NO_BIT, false, NO_UPDATE_DEFAULT)              \
-  decl(unaligned_access, "Unaligned"   ,      NO_BIT, false, NO_UPDATE_DEFAULT)              \
-  decl(satp_mode       , "SATP"        ,      NO_BIT, false, NO_UPDATE_DEFAULT)              \
+  #define RV_FEATURE_FLAGS(decl)                                                                \
+  decl(ext_I           , "i"           ,    ('I' - 'A'), true , NO_UPDATE_DEFAULT)              \
+  decl(ext_M           , "m"           ,    ('M' - 'A'), true , NO_UPDATE_DEFAULT)              \
+  decl(ext_A           , "a"           ,    ('A' - 'A'), true , NO_UPDATE_DEFAULT)              \
+  decl(ext_F           , "f"           ,    ('F' - 'A'), true , NO_UPDATE_DEFAULT)              \
+  decl(ext_D           , "d"           ,    ('D' - 'A'), true , NO_UPDATE_DEFAULT)              \
+  decl(ext_C           , "c"           ,    ('C' - 'A'), true , UPDATE_DEFAULT(UseRVC))         \
+  decl(ext_Q           , "q"           ,    ('Q' - 'A'), true , NO_UPDATE_DEFAULT)              \
+  decl(ext_H           , "h"           ,    ('H' - 'A'), true , NO_UPDATE_DEFAULT)              \
+  decl(ext_V           , "v"           ,    ('V' - 'A'), true , UPDATE_DEFAULT(UseRVV))         \
+  decl(ext_Zicbom      , "Zicbom"      , RV_NO_FLAG_BIT, true , UPDATE_DEFAULT(UseZicbom))      \
+  decl(ext_Zicboz      , "Zicboz"      , RV_NO_FLAG_BIT, true , UPDATE_DEFAULT(UseZicboz))      \
+  decl(ext_Zicbop      , "Zicbop"      , RV_NO_FLAG_BIT, true , UPDATE_DEFAULT(UseZicbop))      \
+  decl(ext_Zba         , "Zba"         , RV_NO_FLAG_BIT, true , UPDATE_DEFAULT(UseZba))         \
+  decl(ext_Zbb         , "Zbb"         , RV_NO_FLAG_BIT, true , UPDATE_DEFAULT(UseZbb))         \
+  decl(ext_Zbc         , "Zbc"         , RV_NO_FLAG_BIT, true , NO_UPDATE_DEFAULT)              \
+  decl(ext_Zbs         , "Zbs"         , RV_NO_FLAG_BIT, true , UPDATE_DEFAULT(UseZbs))         \
+  decl(ext_Zicsr       , "Zicsr"       , RV_NO_FLAG_BIT, true , NO_UPDATE_DEFAULT)              \
+  decl(ext_Zifencei    , "Zifencei"    , RV_NO_FLAG_BIT, true , NO_UPDATE_DEFAULT)              \
+  decl(ext_Zic64b      , "Zic64b"      , RV_NO_FLAG_BIT, true , UPDATE_DEFAULT(UseZic64b))      \
+  decl(ext_Zihintpause , "Zihintpause" , RV_NO_FLAG_BIT, true , UPDATE_DEFAULT(UseZihintpause)) \
+  decl(mvendorid       , "VendorId"    , RV_NO_FLAG_BIT, false, NO_UPDATE_DEFAULT)              \
+  decl(marchid         , "ArchId"      , RV_NO_FLAG_BIT, false, NO_UPDATE_DEFAULT)              \
+  decl(mimpid          , "ImpId"       , RV_NO_FLAG_BIT, false, NO_UPDATE_DEFAULT)              \
+  decl(unaligned_access, "Unaligned"   , RV_NO_FLAG_BIT, false, NO_UPDATE_DEFAULT)              \
+  decl(satp_mode       , "SATP"        , RV_NO_FLAG_BIT, false, NO_UPDATE_DEFAULT)              \
 
   #define DECLARE_RV_FEATURE(NAME, PRETTY, BIT, FSTRING, FLAGF)        \
   struct NAME##RVFeatureValue : public RVFeatureValue {                \
--- a/src/hotspot/cpu/riscv/vm_version_riscv.hpp
+++ b/src/hotspot/cpu/riscv/vm_version_riscv.hpp
@@ -2,6 +2,7 @@
  * Copyright (c) 1997, 2020, Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2014, 2020, Red Hat Inc. All rights reserved.
  * Copyright (c) 2020, 2023, Huawei Technologies Co., Ltd. All rights reserved.
+ * Copyright (c) 2023, Rivos Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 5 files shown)


robehn:
Thanks @markdryan for pointing out that we should free the uarch string.

feilongjiang (commented):
Thanks for the great work, it looks much better now.

src/hotspot/cpu/riscv/vm_version_riscv.hpp
```diff
@@ -30,55 +31,174 @@
...
+  // Enables features in _feature_list
+  static void setup_cpu_available_features();
+  // Helper for specific queries
+  static void os_aux_features();
+  static char* os_uarch_additional_features();
+  static void vendor_features();
+  // Vendors specific features
+  static void rivos_features();
+
+  // Determine vector length iff ext_C/UseRVC
```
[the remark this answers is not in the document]
robehn: Yes, thanks! I'll fix it!

robehn:
> Thanks for the great work, it looks much better now.

Thank you!

src/hotspot/os_cpu/linux_riscv/vm_version_linux_riscv.cpp
```diff
@@ -25,39 +26,42 @@
...
 #include "runtime/os.inline.hpp"
 #include "runtime/vm_version.hpp"
 
 #include <asm/hwcap.h>
+#include <ctype.h>
 #include <sys/auxv.h>
 
 #ifndef HWCAP_ISA_I
-#define HWCAP_ISA_I  (1 << ('I' - 'A'))
+#define HWCAP_ISA_I  (1ULL << ('I' - 'A'))
```
tstuefe, line 40: nit: use the nth_bit() macro?

src/hotspot/os_cpu/linux_riscv/vm_version_linux_riscv.cpp
```diff
@@ -70,68 +74,166 @@
...
-  if (!strcmp(_vm_mode, "sv39")) {
+void VM_Version::setup_cpu_available_features() {
+
+  assert(ext_I.feature_bit() == HWCAP_ISA_I, "Bit for I must follow Linux HWCAP");
+  assert(ext_M.feature_bit() == HWCAP_ISA_M, "Bit for M must follow Linux HWCAP");
+  assert(ext_A.feature_bit() == HWCAP_ISA_A, "Bit for A must follow Linux HWCAP");
+  assert(ext_F.feature_bit() == HWCAP_ISA_F, "Bit for F must follow Linux HWCAP");
+  assert(ext_D.feature_bit() == HWCAP_ISA_D, "Bit for D must follow Linux HWCAP");
+  assert(ext_C.feature_bit() == HWCAP_ISA_C, "Bit for C must follow Linux HWCAP");
+  assert(ext_V.feature_bit() == HWCAP_ISA_V, "Bit for V must follow Linux HWCAP");
```
tstuefe, line 90: I'm somewhat confused about the hw_cap feature bits. Kernel asm/hwcap.h seems only to define caps up to C for riscv. This code mentions (in VMVersion) Q, H, V. 

Are these future flags?

Here we don't check for Q, should we check that too?

tstuefe (approved):
I just got curious. Small nits and questions.

It seems odd to add individual vendor capabilities, but since its democratic and every vendor can add theirs, I think this is fine.

src/hotspot/os_cpu/linux_riscv/vm_version_linux_riscv.cpp
```diff
@@ -70,68 +74,166 @@
...
+
+  RiscvHwprobe::probe_features();
+  os_aux_features();
```
feilongjiang, line 93: Looks like `probe_features` and `os_aux_features` are doing the same thing (enabling supported extensions). Should we call them both? Maybe it's better to call `os_aux_features` when the `probe_features` syscall is not supported? e.g.:

```cpp
if (!RiscvHwprobe::probe_features()) {
  os_aux_features();
}
```

src/hotspot/os_cpu/linux_riscv/vm_version_linux_riscv.cpp
```diff
@@ -70,68 +74,166 @@
...
+  switch (mvendorid.value()) {
+    case RIVOS_MVENDORID:
+    rivos_features();
+    break;
+    default:
+    break;
+  }
+}
+
+void VM_Version::rivos_features() {
```
RealFYang, line 212: It looks to me not safe to only check vendor ID when enabling those features. I thinks it's possible for a vendor to have different CPU models equipped different RISC-V extensions. So it's more reasonable to check `marchid` & `mimpid` here which helps distinguish Architecture ID & Implementaion ID of the CPU.

src/hotspot/cpu/riscv/vm_version_riscv.hpp
```diff
@@ -30,55 +31,174 @@
...
+  //
+  // Zicsr Control and Status Register (CSR) Instructions
+  // Zifencei Instruction-Fetch Fence
+  // Zic64b Cache blocks must be 64 bytes in size, naturally aligned in the address space.
+  // Zihintpause Pause instruction HINT
+  //
+  // Other features and settings
+  // mvendorid Manufactory JEDEC
+  // marchid   Manufactory prop.
+  // mimpid    Manufactory prop.
```
RealFYang, line 110: I think the code comments for marchid and mimpid could be improved.

RealFYang (commented):
Thanks for the work. Two comments after a cursory look.

src/hotspot/os_cpu/linux_riscv/vm_version_linux_riscv.cpp
robehn: This is a moving target more is getting added.
The basic standard ones follow this patterns.
E.g. https://lore.kernel.org/lkml/20230428190609.3239486-2-evan@rivosinc.com/

Yes Q should be added and the other ones. I'll just add Q in this change-set,

src/hotspot/os_cpu/linux_riscv/vm_version_linux_riscv.cpp
robehn: Yes, ok, sure!

src/hotspot/os_cpu/linux_riscv/vm_version_linux_riscv.cpp
robehn: Yes, correct.
The idea was that the vendor can add these checks in it's specific path.
For us enabling these basic features are fine, since they are common to march/mimpid, hence we do not yet need to check these.

Hope that explains it.

src/hotspot/cpu/riscv/vm_version_riscv.hpp
robehn: Ok, I'll fix something!

src/hotspot/os_cpu/linux_riscv/vm_version_linux_riscv.cpp
robehn: > Thanks for the work. Two comments after a cursory look.

Thanks!

robehn:
> Thanks for the work. Two comments after a cursory look.

Thanks!

src/hotspot/os_cpu/linux_riscv/vm_version_linux_riscv.cpp
RealFYang: I see. Maybe leave a code comment for your case?

src/hotspot/os_cpu/linux_riscv/vm_version_linux_riscv.cpp
```diff
@@ -70,68 +74,166 @@
...
+  assert(ext_V.feature_bit() == HWCAP_ISA_V, "Bit for V must follow Linux HWCAP");
+
+  RiscvHwprobe::probe_features();
+  os_aux_features();
+  char* uarch = os_uarch_additional_features();
+  vendor_features();
+
+  char buf[1024] = {};
+  if (uarch != nullptr && strcmp(uarch, "") != 0) {
+    snprintf(buf, sizeof(buf), "%s,", uarch);
```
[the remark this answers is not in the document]
robehn: I did an alternative.

src/hotspot/cpu/riscv/vm_version_riscv.hpp
```diff
@@ -30,55 +31,174 @@
...
-  static void c2_initialize();
-#endif // COMPILER2
+  friend RiscvHwprobe;
+ private:
+  class RVFeatureValue {
+    const char* _pretty;
+    bool        _feature_string;
+    uint64_t    _feature_bit;
+    bool        _enabled;
+    int64_t     _value;
```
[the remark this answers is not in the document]
robehn: Fixed

src/hotspot/os_cpu/linux_riscv/vm_version_linux_riscv.cpp
robehn: Sure fixed

src/hotspot/os_cpu/linux_riscv/vm_version_linux_riscv.cpp
robehn: Added H and Q.

robehn:
Pushed an update after review comments, hope I got all!

tstuefe:
> Pushed an update after review comments, hope I got all!

Still good. Thanks for taking my suggestions.

RealFYang:
I tried to do a native release build on HiFive Unmatched board with the latest version and unfortunately I got SIGILL problem:
```
#
#
#  SIGILL (0x4) at pc=0x0000003f9053eb34, pid=64268, tid=64273
#
# JRE version:  (22.0) (build )
# Java VM: OpenJDK 64-Bit Server VM (22-internal-adhoc.fyang.jdk, mixed mode, tiered, compressed oops, compressed class ptrs, g1 gc, linux-riscv64)
# Problematic frame:
#
[error occurred during error reporting (printing problematic frame), id 0xb, SIGSEGV (0xb) at pc=0x0000003f9df03b0e]
# Core dump will be written. Default location: Core dumps may be processed with "/usr/share/apport/apport -p%p -s%s -c%c -d%d -P%P -u%u -g%g -- %E" (or dumping to /home/fyang/jdk/make/core.64268)
#
#

---------------  S U M M A R Y ------------

Command Line: -Xms64M -Xmx1600M --add-exports=java.base/jdk.internal.module=ALL-UNNAMED build.tools.jigsaw.AddPackagesAttribute /home/fyang/jdk/build/linux-riscv64-server-release/jdk

Host: RISCV64, 4 cores, 15G, Ubuntu 22.04.2 LTS
Time: Thu Jun 15 13:11:35 2023 CST elapsed time: 0.209789 seconds (0d 0h 0m 0s)
```

robehn:
> I tried to do a native release build on HiFive Unmatched board with the latest version and unfortunately I got SIGILL problem:
> 
> ```
> #
> # A fatal error has been detected by the Java Runtime Environment:
> #
> #  SIGILL (0x4) at pc=0x0000003f9053eb34, pid=64268, tid=64273
> #
> # JRE version:  (22.0) (build )
> # Java VM: OpenJDK 64-Bit Server VM (22-internal-adhoc.fyang.jdk, mixed mode, tiered, compressed oops, compressed class ptrs, g1 gc, linux-riscv64)
> # Problematic frame:
> #
> [error occurred during error reporting (printing problematic frame), id 0xb, SIGSEGV (0xb) at pc=0x0000003f9df03b0e]
> # Core dump will be written. Default location: Core dumps may be processed with "/usr/share/apport/apport -p%p -s%s -c%c -d%d -P%P -u%u -g%g -- %E" (or dumping to /home/fyang/jdk/make/core.64268)
> #
> #
> 
> ---------------  S U M M A R Y ------------
> 
> Command Line: -Xms64M -Xmx1600M --add-exports=java.base/jdk.internal.module=ALL-UNNAMED build.tools.jigsaw.AddPackagesAttribute /home/fyang/jdk/build/linux-riscv64-server-release/jdk
> 
> Host: RISCV64, 4 cores, 15G, Ubuntu 22.04.2 LTS
> Time: Thu Jun 15 13:11:35 2023 CST elapsed time: 0.209789 seconds (0d 0h 0m 0s)
> ```

Thanks, it seem most likely I have a bug and enabled an extension not supported.
I guess you are not running a kernel with hw_probe.
If possible can you run with -XX:+PrintFlagsFinal ?

RealFYang:
PS:
Checking the flags with -XX:+PrintFlagsFinal, looks like this will enable some features which are not available on this board.
```
fyang@ubuntu:~/jdk$ ./build/linux-riscv64-server-release/jdk/bin/java -XX:+PrintFlagsFinal -XX:+UnlockExperimentalVMOptions -version &> ~/flags.txt
Aborted (core dumped)

fyang@ubuntu:~/jdk$ grep "ARCH experimental" ~/flags.txt
     bool UseRVA22U64                              = false                           {ARCH experimental} {default}
     bool UseRVV                                   = false                           {ARCH experimental} {default}
     bool UseZba                                   = true                            {ARCH experimental} {default}
     bool UseZbb                                   = true                            {ARCH experimental} {default}
     bool UseZbs                                   = true                            {ARCH experimental} {default}
     bool UseZic64b                                = true                            {ARCH experimental} {default}
     bool UseZicbom                                = true                            {ARCH experimental} {default}
     bool UseZicbop                                = true                            {ARCH experimental} {default}
     bool UseZicboz                                = true                            {ARCH experimental} {default}
     bool UseZihintpause                           = true                            {ARCH experimental} {default}
```

RealFYang:
> 
> Thanks, it seem most likely I have a bug and enabled an extension not supported. I guess you are not running a kernel with hw_probe. If possible can you run with -XX:+PrintFlagsFinal ?

Yes, you are right. I am using Ubuntu 22.04.2 LTS which is powered by Linux 5.19.0-1019-generic.

robehn:
Thanks again.

I have jumped back and forth with bit number vs bit.
This should fix it, and make it more clear.

I would very much appreciate if you could retest when you got time. (only have qemu a.t.m.)

src/hotspot/os_cpu/linux_riscv/vm_version_linux_riscv.cpp
```diff
@@ -70,68 +82,175 @@
...
-          uarch[strcspn(uarch, "\n")] = '\0';
-          _uarch = uarch;
-          break;
+          mode = VM_Version::parse_satp_mode(p);
+        }
+      }
+      if (ret == nullptr) {
+        if (strncmp(buf, "uarch", sizeof "uarch" - 1) == 0) {
+          ret = os::strdup(p + 2);
+          ret[strcspn(p, "\n")] = '\0';
```
RealFYang, line 196: This line was triggering some NMT corruption errors when running tier1 test. I think this should be: `ret[strcspn(ret, "\n")] = '\0';`.

RealFYang (changes requested):

src/hotspot/os_cpu/linux_riscv/vm_version_linux_riscv.cpp
robehn: Since ret is identical to 'p' (but larger), the number of chars is identical.
I don't have a system which return uarch in cpuinfo, and I have not seen one, so I can't test this path.
What are you testing on?

`rehn@rv:~/source/gdb/binutils-gdb$ uname -a ; grep -i "uarch\|mmu" /proc/cpuinfo  | uniq 
Linux rv 6.2.0-19-generic #19.1-Ubuntu SMP Fri Mar 31 12:41:53 UTC 2023 riscv64 riscv64 riscv64 GNU/Linux
mmu		: sv48
`

I don't mind pushing your suggested change, but I fail to understand how that can fix something :)

RealFYang: Here is dump of file /proc/cpuinfo on hifive unmatched board:
```
$ cat /proc/cpuinfo
processor       : 0
hart            : 2
isa             : rv64imafdc
mmu             : sv39
uarch           : sifive,u74-mc

processor       : 1
hart            : 1
isa             : rv64imafdc
mmu             : sv39
uarch           : sifive,u74-mc

processor       : 2
hart            : 3
isa             : rv64imafdc
mmu             : sv39
uarch           : sifive,u74-mc

processor       : 3
hart            : 4
isa             : rv64imafdc
mmu             : sv39
uarch           : sifive,u74-mc
```

Since `ret = os::strdup(p + 2);`, I don't think `ret` and 'p' are identical. And here is what the old code does [1].

[1] https://github.com/openjdk/jdk/blob/master/src/hotspot/os_cpu/linux_riscv/vm_version_linux_riscv.cpp#LL128-L129

robehn:
> Updated change LGTM. Thanks.

Thank you!

robehn:
I did a last minute change to our common features.
I'll integrate tomorrow, thanks all for reviewing!

I'll ask for approval to backport to 21.

DingliZhang:
Hi, I noticed fastdebug building crashed when running `java -version` on SG2042 (64 core c910):

```
$ ./java -version
#
#
#  SIGILL (0x4) at pc=0x0000003f73a70892, pid=347134, tid=347135
#
# JRE version:  (22.0) (fastdebug build )
# Java VM: OpenJDK 64-Bit Server VM (fastdebug 22-internal-adhoc.zhangdingli.jdk, mixed mode, tiered, compressed oops, compressed class ptrs, g1 gc, linux-riscv64)
# Problematic frame:
# v  ~StubRoutines::jint_disjoint_arraycopy 0x0000003f73a70892
#
# Core dump will be written. Default location: Core dumps may be processed with "/usr/share/apport/apport -p%p -s%s -c%c -d%d -P%P -u%u -g%g -- %E" (or dumping to /home/perfxlab01/jdkbuild/compare/jdk_after/bin/core.347134)
#
# An error report file with more information is saved as:
# /home/perfxlab01/jdkbuild/compare/jdk_after/bin/hs_err_pid347134.log
[0.727s][warning][os] Loading hsdis library failed
#
#
Aborted (core dumped)
```
The instruct at the pc of crash is `020f6007`, which is `vle32.v v0,(t5)`.
Complete log: https://cr.openjdk.org/~dzhang/hs_err_pid347134.log

System Information:
```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.10
Release:	22.10
Codename:	kinetic

$ uname -a
Linux riscv-02 6.1.31 #6 SMP Tue Jul 11 18:48:27 CST 2023 riscv64 riscv64 riscv64 GNU/Linux

$ cat /proc/cpuinfo 
processor	: 0
hart		: 4
isa		: rv64imafdcv
mmu		: sv39
mvendorid	: 0x5b7
marchid		: 0x0
mimpid		: 0x0
...
```

The release building `java -version` doesn't cause a crash, but when set it as JAVA_HOME and then executes `jtreg -version`, it will crash for similar reasons. We can see the flag under release building:
```
$ ./java -XX:+PrintFlagsFinal -XX:+UnlockExperimentalVMOptions -version | grep UseRVV
     bool UseRVV                                   = true                            {ARCH experimental} {default}
     bool UseRVVForBigIntegerShiftIntrinsics       = true                                 {ARCH product} {default}
```
The problem looks like c910 only supports rvv0.7.1, while openjdk only supports rvv1.0. `UseRVV` is automatically enabled on SG2042, even though it's a experimental VMOption. Maybe we should set `UPDATE_DEFAULT(UseRVV)` to `NO_UPDATE_DEFAULT` for now or do something depending on the version of rvv?

I would appreciate it if some advice is given, thank you!

luhenry:
@DingliZhang I expect that it gets enabled because `hwcap` sets the `V` bit to 1, which sets `ext_V.enabled = true` which, when calling `update_flags` sets `UseRVV` to true, correct?

`riscv_hwprobe` does support versioning (checking the exact details). So if anything, we should indeed not set `UseRVV` when probing via hwcap, but we _should_ set `UseRVV` when probing via hwprobe (given we have RVV 1.0 of course).

PR #14450, opened 2023-06-13
8309939: HttpClient should not use Instant.now() as Instant source for deadlines

dfuch:
The HttpClient uses `Instant.now()` to create deadlines for timeouts. This could have undesirable effects since `Instant.now()` is linked to the wall clock, which is not monotonic. This fix changes the HttpClient to use a monotonic instant source based on `System.nanoTime()` for the purpose of setting and comparing deadlines.

---------
### Progress

### Issue
 * [JDK-8309939](https://bugs.openjdk.org/browse/JDK-8309939): HttpClient should not use Instant.now() as Instant source for deadlines (**Bug** - P4)

### Reviewers
 * [Daniel Jeliński](https://openjdk.org/census#djelinski) (@djelinski - **Reviewer**) ⚠️ Review applies to [e5f6cd60](https://git.openjdk.org/jdk/pull/14450/files/e5f6cd60667fe2da36bfe2380f7ba020259a4fa0)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14450/head:pull/14450` \
`$ git checkout pull/14450`

`$ git checkout pull/14450` \
`$ git pull https://git.openjdk.org/jdk.git pull/14450/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14450`

`$ git pr show -t 14450`

</details>

<a href="https://git.openjdk.org/jdk/pull/14450.diff">https://git.openjdk.org/jdk/pull/14450.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14450#issuecomment-1589582710)


src/java.net.http/share/classes/jdk/internal/net/http/ConnectionPool.java
```diff
@@ -124,17 +126,22 @@ public int hashCode() {
...
     ConnectionPool(long clientId) {
-        this("ConnectionPool("+clientId+")");
+        this(clientId, TimeSource.source());
     }
 
-    /**
-     * There should be one of these per HttpClient.
-     */
-    private ConnectionPool(String tag) {
+    ConnectionPool(long clientId, TimeSource timeSource) {
```
jaikiran, line 132: Hello Daniel, as far as I can see, no one calls this constructor outside of this very class and this seems to be always passed a constant `TimeSource.source()`. So maybe we don't have to have a `timeSource` field in this class and instead just use `TimeSource.now()` at call sites within this class (and `ExpiryList`) whenever we want `now()`?

src/java.net.http/share/classes/jdk/internal/net/http/common/TimeSource.java
```diff
@@ -0,0 +1,129 @@
...
+                // around.
+                // The use of Integer.MAX_VALUE is arbitrary.
+                // Any value not too close to Long.MAX_VALUE
+                // would do.
+                nanoSource = new NanoSource(now, nanos);
+            }
+            return now;
+        }
+
+        // @ForceInline
```
jaikiran, line 82: Should these couple of commented out `@ForceInline` be deleted?

src/java.net.http/share/classes/jdk/internal/net/http/common/TimeSource.java
```diff
@@ -0,0 +1,129 @@
...
+        Instant instant(long nanos, long delay) {
+            Instant now = first.plusNanos(delay);
+            if (!isInWindow(delay)) {
+                // Shifts the time reference (firstNanos) to
+                // prevent issues that may be caused by
+                // System.nanoTime() - firstNanos wrapping
+                // around.
+                // The use of Integer.MAX_VALUE is arbitrary.
+                // Any value not too close to Long.MAX_VALUE
+                // would do.
```
jaikiran, line 76: The `Integer.MAX_VALUE` gets used in `isInWindow()` method implemented, through the use of `TIME_WINDOW` member. Should this comment about `Integer.MAX_VALUE` instead be moved as a field comment for `TIME_WINDOW`? That way it's closer to where the `Integer.MAX_VALUE` is actually used.

src/java.net.http/share/classes/jdk/internal/net/http/common/TimeSource.java
```diff
@@ -0,0 +1,129 @@
...
+ * The granularity of instants returned is identical to the granularity
+ * of {@link System#nanoTime()}. This time source has the same property
+ * of monotonicity than the {@link System#nanoTime()} it is based on.
+ */
+public final class TimeSource implements InstantSource {
+    private static volatile NanoSource nanoSource = new NanoSource();
+    private NanoSource localSource = nanoSource;
+    private static final TimeSource SOURCE = new TimeSource();
+
+    private static final class NanoSource implements InstantSource {
```
djelinski, line 46: This class doesn't need to implement InstantSource

src/java.net.http/share/classes/jdk/internal/net/http/common/TimeSource.java
```diff
@@ -0,0 +1,129 @@
...
+        NanoSource() {
+            this(Instant.now(), System.nanoTime());
+        }
+        NanoSource(Instant first, long firstNanos) {
+            this.first = first;
+            this.firstNanos = firstNanos;
+        }
+
+        @Override
+        public Instant instant() {
```
djelinski, line 59: This method is not used anywhere

src/java.net.http/share/classes/jdk/internal/net/http/common/TimeSource.java
```diff
@@ -0,0 +1,129 @@
...
+ * purpose of handling timeouts. This time source is intentionally not
+ * based on the system clock, in order to not be sensitive to clock skews
+ * caused by changes to the wall clock. Consequently, callers should use
+ * instants returned by this time source solely for the purpose of
+ * comparing them with other instants returned by this same time source.
+ * The granularity of instants returned is identical to the granularity
+ * of {@link System#nanoTime()}. This time source has the same property
+ * of monotonicity than the {@link System#nanoTime()} it is based on.
+ */
+public final class TimeSource implements InstantSource {
```
djelinski, line 41: Given that all instances of TimeSource are essentially the same (through the use of shared `nanoSource`), can we add a private constructor and make this class a singleton?

src/java.net.http/share/classes/jdk/internal/net/http/ConnectionPool.java
dfuch: This constructor should in fact take an InstantSource, not a TimeSource. It was added in case we wanted to upgrade the whitebox ConnectionPoolTest to be able to create an instance of ConnectionPool with a custom InstantSource.

src/java.net.http/share/classes/jdk/internal/net/http/common/TimeSource.java
dfuch: Done

src/java.net.http/share/classes/jdk/internal/net/http/common/TimeSource.java
dfuch: Done

src/java.net.http/share/classes/jdk/internal/net/http/common/TimeSource.java
dfuch: It doesn't but it is an InstantSource - so it feels right that it implements InstanceSource.

dfuch: Good point

src/java.net.http/share/classes/jdk/internal/net/http/common/TimeSource.java
dfuch: Good point. That's an oversight.

src/java.net.http/share/classes/jdk/internal/net/http/common/TimeSource.java
```diff
@@ -0,0 +1,128 @@
...
+import java.time.Instant;
+import java.time.InstantSource;
+
+/**
+ * An {@link InstantSource} based on {@link System#nanoTime()} for the
+ * purpose of handling timeouts. This time source is intentionally not
+ * based on the system clock, in order to not be sensitive to clock skews
+ * caused by changes to the wall clock. Consequently, callers should use
+ * instants returned by this time source solely for the purpose of
+ * comparing them with other instants returned by this same time source.
```
RogerRiggs, line 36: Do not use `Instant` as the value from this time source. It is not comparable with the real Instants and would be misleading lead to bugs.

RogerRiggs (changes requested):

src/java.net.http/share/classes/jdk/internal/net/http/common/TimeSource.java
RogerRiggs: Wrapping the Instant is safe/fine.
However, it could quite a bit simpler (it seems) to just use the raw `long` nanoTime values.
Though milliseconds or seconds would be sufficient for the timeouts.
The complexity of the volatile read and needing to reset the nanoSource could be avoided.

dfuch: Working with Instant (now Deadline) and  Duration is much safer and easier than working with longs representing nano time, especially when you have a sorted list of deadlines.

src/java.net.http/share/classes/jdk/internal/net/http/common/Deadline.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    @Override
+    public int compareTo(Deadline o) {
+        return deadline.compareTo(o.deadline);
+    }
+
+    @Override
+    public String toString() {
+        return deadline.toString();
```
djelinski, line 203: I'm not sure about using `Instant.toString` here; if we ever log a `Deadline` instance along with the current time and the clock skews, we will need to read the code along with the logs to figure out which date represents a deadline and which one represents an instant.
I guess the problem would also go away if we initialized the first NanoSource with something other than Instant.now().

dfuch: I can modify the toString to be 
```
  return "Deadline(" + deadline.toString() + ")";
```
if you think it's better.

But I do believe that using a fake instant for the origin of the NanoSource would be way more confusing.

djelinski: We can also avoid using that method. I guess we can always deal with it later if it proves to be a problem.

djelinski (approved):
Using nanoTime as @RogerRiggs suggested is worth exploring, but IMO the current version is good enough for internal use.

PR #14457, opened 2023-06-13
 JDK-8308398 Move SunEC crypto provider into java.base 

ascarpino:
Hi,

I need a code review for moving the contents of the jdk.crypto.ec module into java.base.  This moves the SunEC JCE Provider (Elliptic Curve) into java.base.  EC has always been separate from the base module/pkg because of its dependence on a native library.  That library was removed in JDK 16.  An empty jdk.crypto.ec module will remain for compatibility, but marked as deprecated with the intent to be removed in a future release.

There should be no compatibility risk for application using EC through JCE. There are no public API changes to EC, XEC, and EdDSA classes .  Applications that unwisely accessing internal EC classes will need to use the java.base module.

Thanks

Tony

---------
### Progress
- [x] Change requires CSR request [JDK-8308602](https://bugs.openjdk.org/browse/JDK-8308602) to be approved

### Issues
 * [JDK-8308398](https://bugs.openjdk.org/browse/JDK-8308398): Move SunEC crypto provider into java.base (**Enhancement** - P3)
 * [JDK-8308602](https://bugs.openjdk.org/browse/JDK-8308602): Move SunEC crypto provider into java.base (**CSR**)

### Reviewers
 * [Valerie Peng](https://openjdk.org/census#valeriep) (@valeriepeng - **Reviewer**) ⚠️ Review applies to [cf8e954d](https://git.openjdk.org/jdk/pull/14457/files/cf8e954d2b6c4a2ff2a86731c681ed5eb7a4b6ee)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14457/head:pull/14457` \
`$ git checkout pull/14457`

`$ git checkout pull/14457` \
`$ git pull https://git.openjdk.org/jdk.git pull/14457/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14457`

`$ git pr show -t 14457`

</details>

<a href="https://git.openjdk.org/jdk/pull/14457.diff">https://git.openjdk.org/jdk/pull/14457.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14457#issuecomment-1595268521)


src/jdk.crypto.ec/share/classes/module-info.java
```diff
@@ -31,6 +31,7 @@
  * @moduleGraph
  * @since 9
  */
+@Deprecated(since="22", forRemoval = true)
 module jdk.crypto.ec {
-    provides java.security.Provider with sun.security.ec.SunEC;
+    //provides java.security.Provider with sun.security.ec.SunEC;
```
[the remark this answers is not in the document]
AlanBateman: Yes, it should be removed.

AlanBateman:
> This moves the SunEC JCE Provider (Elliptic Curve) into java.base. EC has always been separate from the base module/pkg because of its dependence on a native library. That library was removed in JDK 16. 

The proposed changes look okay, meaning it should be okay to have the SunEC provider in java.base. However, the motivation isn't clear as there isn't an issue with JCE providers in java.base using native code. I know there were non-technical issues with libsunec in the past but that would haven't prevent the SunEC code form being compiled into java.base.

I assume the main implications of the change is that 3rd party JCE providers signed with an EC certificate can now be deployed on the module path.  Another way to solve that issue is that delay verification of signed JARs until the boot layer is created - if we did that, would you still want to move the SunEC provider into java.base?

src/jdk.crypto.ec/share/classes/module-info.java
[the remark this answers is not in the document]
ascarpino: Yep

ascarpino:
> > This moves the SunEC JCE Provider (Elliptic Curve) into java.base. EC has always been separate from the base module/pkg because of its dependence on a native library. That library was removed in JDK 16.
> 
> The proposed changes look okay, meaning it should be okay to have the SunEC provider in java.base. However, the motivation isn't clear as there isn't an issue with JCE providers in java.base using native code. I know there were non-technical issues with libsunec in the past but that would haven't prevent the SunEC code form being compiled into java.base.
> 

From what I was told, the native library was one of the reasons it was not in the base pkg before modularization and just remained so afterwards.

> I assume the main implications of the change is that 3rd party JCE providers signed with an EC certificate can now be deployed on the module path. Another way to solve that issue is that delay verification of signed JARs until the boot layer is created - if we did that, would you still want to move the SunEC provider into java.base? Maybe you want it in java.base so there is an alternative to RSA in all run-time images?

Just before this review went out I tried JDK-8215932.  At this point, I'm unable to reproduce the original problem with EC JCE signed jars.  I talked to the submitter and at this point I'm not sure if JDK-8215932 was incorrect or it was fixed elsewhere. That doesn't reduce the motivation to remove the module.

AlanBateman:
> From what I was told, the native library was one of the reasons it was not in the base pkg before modularization and just remained so afterwards.

Maybe you are thinking about the size of libsunec or non-technical issues that meant it wasn't included by some distributions? There weren't an issue with deciding which providers to include to java.base. I think the motivation for having the SunEC provider in java.base now is probably TLS so there are more secure cipher suites available for those that create a small run-image with jlink and don't include all security providers.

ascarpino:
> 

TLS is mentioned in the CSR as a reason for the merger.  With EC not being in java.base and an alternative for RSA, it could lead to slimmed down, custom images not having the best crypto algorithms available moving in the future.

test/jdk/sun/security/ec/ecModuleCheck.java
```diff
@@ -0,0 +1,45 @@
...
+/* This test verifies that with jdk.crypto.ec loaded that the EC modules is
+ * available. The KeyPairGenerator is just to verify SunEC is working.  Other
+ * tests access internal sun.security.ec APIs from java.base (see TestEC.java)
+ */
+public class ecModuleCheck {
+    public static void main(String[] args) throws Exception {
+        if (!ModuleLayer.boot().findModule("jdk.crypto.ec").isPresent()) {
+            throw new AssertionError("jdk.crypto.ec module does not exist");
+        }
+        System.out.println("jdk.crypto.ec module exists");
```
AlanBateman, line 43: `@modules jdk.crypto.ec` in the test description means this test will not be selected/executed if the jdk.crypto.ec is not the run-time image. It also means the test with run with `--add-modules jdk.crypto.ec` which is not what you want. I think this is closer to what you want in this test:

```
    assertTrue(ModuleFinder.ofSystem().find("jdk.crypto.ec").isPresent());
    assertFalse(ModuleLayer.boot().findModule("jdk.crypto.ec").isPresent());    
```

This tests that the jdk.crypto.ec is observable but is not resolved.

AlanBateman:
> > Maybe you are thinking about the size of libsunec or non-technical issues that meant it wasn't included by some distributions? There weren't an issue with deciding which providers to include to java.base. I think the motivation for having the SunEC provider in java.base now is probably TLS so there are more secure cipher suites available for those that create a small run-image with jlink and don't include all security providers.
> 
> Yes, I think the motivation is more that Elliptic Curve Cryptography is a widely used form of crypto and should be in java.base. I haven't tried this, but I think TLS 1.3 would simply not work if you just had java.base in your runtime.

I think we've converged on the right motivation. If would be good to check if there are TLS tests that could run with --limit-modules java.base, that would give confidence that the API/implementation will work when the run-time image only contains java.base.

ascarpino:
> I think we've converged on the right motivation. If would be good to check if there are TLS tests that could run with --limit-modules java.base, that would give confidence that the API/implementation will work when the run-time image only contains java.base.

I can add a case for one of the TLS tests I changed in this PR

src/java.base/share/classes/sun/security/jca/ProviderConfig.java
```diff
@@ -176,60 +176,63 @@ Provider getProvider() {
...
-                    p = doLoadProvider();
-                } finally {
-                    isLoading = false;
-                }
-            }
+            };
             provider = p;
         }
-        return p;
+        return provider;
```
valeriepeng, line 235: Why changing it to return 'provider', is this change intentional? IIRC, the general pattern is to return the local reference instead of a volatile variable.

ascarpino: Ok

valeriepeng (approved):

AlanBateman (approved):

PR #14460, opened 2023-06-14
8309663: test fails "assert(check_alignment(result)) failed: address not aligned: 0x00000008baadbabe"

alexmenkov:
If virtual thread has frames in stackChunks, need to apply load barriers before processing nmethods (javaVFrame::locals() and javaVFrame::expressions() do it internally)

Testing: tier1-tier5;
400 runs of VThreadStackRefTest.java test on linux-x64 and linux-x64-debug with "-XX:+UseZGC -Xcomp -XX:-TieredCompilation"

---------
### Progress

### Issue
 * [JDK-8309663](https://bugs.openjdk.org/browse/JDK-8309663): test fails "assert(check_alignment(result)) failed: address not aligned: 0x00000008baadbabe" (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14460/head:pull/14460` \
`$ git checkout pull/14460`

`$ git checkout pull/14460` \
`$ git pull https://git.openjdk.org/jdk.git pull/14460/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14460`

`$ git pr show -t 14460`

</details>

<a href="https://git.openjdk.org/jdk/pull/14460.diff">https://git.openjdk.org/jdk/pull/14460.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14460#issuecomment-1590277238)

### DIFF
--- a/src/hotspot/share/prims/jvmtiTagMap.cpp
+++ b/src/hotspot/share/prims/jvmtiTagMap.cpp
@@ -2320,7 +2320,10 @@ bool StackRefCollector::do_frame(vframe* vf) {
       // Follow oops from compiled nmethod.
       if (jvf->cb() != nullptr && jvf->cb()->is_nmethod()) {
         _blk->set_context(_thread_tag, _tid, _depth, method);
-        jvf->cb()->as_nmethod()->oops_do(_blk);
+        // Need to apply load barriers for unmounted vthreads.
+        nmethod* nm = jvf->cb()->as_nmethod();
+        nm->run_nmethod_entry_barrier();
+        nm->oops_do(_blk);
         if (_blk->stopped()) {
           return false;
         }
--- a/test/hotspot/jtreg/ProblemList-zgc.txt
+++ b/test/hotspot/jtreg/ProblemList-zgc.txt
@@ -45,4 +45,3 @@ vmTestbase/gc/gctests/MemoryEaterMT/MemoryEaterMT.java        8289582   windows-
 
 vmTestbase/nsk/monitoring/MemoryPoolMBean/isCollectionUsageThresholdExceeded/isexceeded002/TestDescription.java 8298302 generic-all
 vmTestbase/nsk/sysdict/vm/stress/chain/chain007/chain007.java 8298991 linux-x64
-serviceability/jvmti/vthread/FollowReferences/VThreadStackRefTest.java#default 8309663 linux-x64


alexmenkov:
added serviceability as it was not added automatically

src/hotspot/share/prims/jvmtiTagMap.cpp
```diff
@@ -2320,7 +2320,10 @@ bool StackRefCollector::do_frame(vframe* vf) {
       // Follow oops from compiled nmethod.
       if (jvf->cb() != nullptr && jvf->cb()->is_nmethod()) {
         _blk->set_context(_thread_tag, _tid, _depth, method);
-        jvf->cb()->as_nmethod()->oops_do(_blk);
+        // Need to apply load barriers for unmounted vthreads.
+        nmethod* nm = jvf->cb()->as_nmethod();
+        nm->run_nmethod_entry_barrier();
```
sspitsyn, line 2325: This looks good in general.
However, I'm looking at the `stackChunkOopDesc::do_barriers0()` implementation and wonder if a similar trick is needed for interpreter frames (to support `Class Redefinition`):
```
template <stackChunkOopDesc::BarrierType barrier, ChunkFrames frame_kind, typename RegisterMapT>
void stackChunkOopDesc::do_barriers0(const StackChunkFrameStream<frame_kind>& f, const RegisterMapT* map) {
  // We need to invoke the write barriers so as not to miss oops in old chunks that haven't yet been concurrently scanned
  assert (!f.is_done(), "");

  if (f.is_interpreted()) {
    Method* m = f.to_frame().interpreter_frame_method();
    // Class redefinition support
    m->record_gc_epoch();
  } else if (f.is_compiled()) {
    nmethod* nm = f.cb()->as_nmethod();
    // The entry barrier takes care of having the right synchronization
    // when keeping the nmethod alive during concurrent execution.
    nm->run_nmethod_entry_barrier();
    // There is no need to mark the Method, as class redefinition will walk the
    // CodeCache, noting their Methods
  }
  . . .
```

alexmenkov: > This looks good in general. However, I'm looking at the `stackChunkOopDesc::do_barriers0()` implementation and wonder if a similar trick is needed for interpreter frames (to support `Class Redefinition`):
> 

This code collects references from thread stack, I don't see how class redefinition can affect them

sspitsyn: Thanks. I agree with it.
The stackChunkOopDesc implementation needs it as it does some stack chunk updates.

sspitsyn (approved):
Looks good.
Thanks,
Serguei

alexmenkov:
Ping. Need one more reviewer.

fisk (approved):
Looks good. It isn't obvious that we need to follow nmethod oops that are not reachable through the current JVM state, but if we do, that's a good fix.

PR #9327, opened 2022-06-29
8288882: JFileChooser - empty (0 bytes) file is displayed as 1 KB

kumarabhi006:
JFileChooser - empty file size issue fixed. 
For empty file, now the size 0 bytes.
Manual Test Case "ZeroFileSizeCheck.java" created.

PR #14464, opened 2023-06-14
8301606: JFileChooser file chooser details view "size" label cut off in Metal Look&Feel

TejeshR13:
"size" label which is _RIGHT_ aligned is cut off on header cell. The issue is not only w.r.t to `JFileChooser` rather it is part of `JTable`. The root caused is found to be that in metal L&F the border insets is set to `(2,2,2,0)` meaning the right most inset value is 0. Hence when UIScaling increases the issue will be visible clearly. The fix addresses the issue by setting the right `inset` to 2 similar to other `inset` values. (Though the reason for setting it to 0 is unclear since it was initial load). 
CI testing shows green.
After the fix at 225% scaling:
image

---------
### Progress

### Issue
 * [JDK-8301606](https://bugs.openjdk.org/browse/JDK-8301606): JFileChooser file chooser details view "size" label cut off in Metal Look&amp;Feel (**Bug** - P3)

### Reviewers
 * [Abhishek Kumar](https://openjdk.org/census#abhiscxk) (@kumarabhi006 - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14464/head:pull/14464` \
`$ git checkout pull/14464`

`$ git checkout pull/14464` \
`$ git pull https://git.openjdk.org/jdk.git pull/14464/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14464`

`$ git pr show -t 14464`

</details>

<a href="https://git.openjdk.org/jdk/pull/14464.diff">https://git.openjdk.org/jdk/pull/14464.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14464#issuecomment-1590830667)


prsadhuk:
Although this might be a bug in insets but for this issue, I guess in native filechooser, "Size" is left aligned which is what we follow..
I think in Java Windows L&F too, it is right aligned which I think we should rectify otherwise it can come as another bug that we are not following native look..

test/jdk/javax/swing/JFileChooser/FileChooserSizeLabelCutOffTest.java
```diff
@@ -0,0 +1,139 @@
...
+        PassFailJFrame.positionTestWindow(frame, PassFailJFrame.Position.TOP_LEFT_CORNER);
+
+        frame.setDefaultCloseOperation(WindowConstants.EXIT_ON_CLOSE);
+        jfc.setControlButtonsAreShown(false);
+        jfc.setDialogType(JFileChooser.CUSTOM_DIALOG);
+
+        frame.add(jfc, BorderLayout.CENTER);
+        frame.pack();
+        frame.setVisible(true);
+    }
```
kumarabhi006, line 102: Please add a blank line between method implementation.

test/jdk/javax/swing/JFileChooser/FileChooserSizeLabelCutOffTest.java
```diff
@@ -0,0 +1,139 @@
...
+    private static void switchToDetailsView() {
+
+        AbstractButton detailsBtn = findDetailsButton(jfc);
+        if (detailsBtn == null) {
+            throw new Error("'Details' button not found in JFileChooser");
+        }
+        detailsBtn.doClick();
+    }
+    private static AbstractButton findDetailsButton(final Container container) {
```
kumarabhi006, line 111: Please add a blank line here as well.

test/jdk/javax/swing/JFileChooser/FileChooserSizeLabelCutOffTest.java
```diff
@@ -0,0 +1,139 @@
...
+        Component result = findComponent(container,
+                c -> c instanceof JToggleButton button
+                        && "Details".equals(button.getToolTipText()));
+        return (AbstractButton) result;
+    }
+
+    private static JTable findTable(final Container container) {
```
kumarabhi006, line 118: This method is not called at any place, can it be removed?

TejeshR13:
> Although this might be a bug in insets but for this issue, I guess in native filechooser, "Size" is left aligned which is what we follow.. I think in Java Windows L&F too, it is right aligned which I think we should rectify otherwise it can come as another bug that we are not following native look..

Yeah, can do that, will update that also in this PR.

aivanov-jdk:
> Although this might be a bug in insets but for this issue, I guess in native filechooser, "Size" is left aligned which is what we follow.. I think in Java Windows L&F too, it is right aligned which I think we should rectify otherwise it can come as another bug that we are not following native look..

Java L&F may not follow what native File Explorers / Navigators do — after all, it's a Java Look-and-Feel.

On Windows, Explorer displays the Size header left-aligned as all other headers, yet the values in the Size column are right-aligned. If Swing doesn't allow different alignment for the header cell and data cells in a column, then we should preserve right-alignment.

test/jdk/javax/swing/JFileChooser/FileChooserSizeLabelCutOffTest.java
TejeshR13: Yeah, was actually trying to do that but couldn't get exact coordinates of header cell. Later on forgot to remove the unused methods.

TejeshR13:
> > Although this might be a bug in insets but for this issue, I guess in native filechooser, "Size" is left aligned which is what we follow.. I think in Java Windows L&F too, it is right aligned which I think we should rectify otherwise it can come as another bug that we are not following native look..
> 
> Java L&F may not follow what native File Explorers / Navigators do — after all, it's a Java Look-and-Feel.
> 
> On Windows, Explorer displays the Size header left-aligned as all other headers, yet the values in the Size column are right-aligned. If Swing doesn't allow different alignment for the header cell and data cells in a column, then we should preserve right-alignment.

Hmm, I do have a fix if we are planning to make it similar to native (Window, Linux and MacOS). And FYI, windows native API  gives the alignment for columns irrespective of header/data, which is RIGHT alignment for `Size` column. If it is still debatable, then will it be better to open a new bug on Size alignment issue and handle it there?

test/jdk/javax/swing/JFileChooser/FileChooserSizeLabelCutOffTest.java
TejeshR13: And also it's not possible to verify with background color because anti-aliasing, I guess.

aivanov-jdk:
> > > Although this might be a bug in insets but for this issue, I guess in native filechooser, "Size" is left aligned which is what we follow.. I think in Java Windows L&F too, it is right aligned which I think we should rectify otherwise it can come as another bug that we are not following native look..
> > 
> > 
> > Java L&F may not follow what native File Explorers / Navigators do — after all, it's a Java Look-and-Feel.
> > On Windows, Explorer displays the Size header left-aligned as all other headers, yet the values in the Size column are right-aligned. If Swing doesn't allow different alignment for the header cell and data cells in a column, then we should preserve right-alignment.
> 
> Hmm, I do have a fix if we are planning to make it similar to native (Window, Linux and MacOS). And FYI, windows native API [getdetailsof](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ishellfolder2-getdetailsof) gives the alignment for columns irrespective of header/data, which is RIGHT alignment for `Size` column. If it is still debatable, then will it be better to open a new bug on Size alignment issue and handle it there?

Here's a screenshot of Windows File Explorer:

Windows File Explorer: Size column is right-aligned, its header is left-aligned

As you can see, the Size column values are right-aligned, its header remains left-aligned as other headers.

In [this comment](https://github.com/openjdk/jdk/pull/9327#issuecomment-1204836731), @kumarabhi006 attached a screenshot of a File Navigator in Oracle Linux: it shows left-aligned for the Size column.

If it's decided to change the alignment of the Size column in Metal L&F, it must be done separately in its own bug. But I don't think we should change it.

aivanov-jdk:
> > > Although this might be a bug in insets but for this issue, I guess in native filechooser, "Size" is left aligned which is what we follow.. I think in Java Windows L&F too, it is right aligned which I think we should rectify otherwise it can come as another bug that we are not following native look..
> > 
> > 
> > Java L&F may not follow what native File Explorers / Navigators do — after all, it's a Java Look-and-Feel.
> > On Windows, Explorer displays the Size header left-aligned as all other headers, yet the values in the Size column are right-aligned. If Swing doesn't allow different alignment for the header cell and data cells in a column, then we should preserve right-alignment.
> 
> Hmm, I do have a fix if we are planning to make it similar to native (Window, Linux and MacOS). And FYI, windows native API [getdetailsof](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ishellfolder2-getdetailsof) gives the alignment for columns irrespective of header/data, which is RIGHT alignment for `Size` column. If it is still debatable, then will it be better to open a new bug on Size alignment issue and handle it there?

Here's a screenshot of Windows File Explorer:

Windows File Explorer: Size column is right-aligned, its header is left-aligned

As you can see, the Size column values are right-aligned, its header remains left-aligned as other headers.

In [this comment](https://github.com/openjdk/jdk/pull/9327#issuecomment-1204836731), @kumarabhi006 attached a screenshot of a File Navigator in Oracle Linux: it shows left-aligned for the Size column.

If it's decided to change the alignment of the Size column in Metal L&F, it must be done separately in its own bug. But I don't think we should change it.

test/jdk/javax/swing/JFileChooser/FileChooserSizeLabelCutOffTest.java
aivanov-jdk: > And also it's not possible to verify with background color because anti-aliasing, I guess.

Why is it? There's no background anti-aliasing, the background is filled with solid color.

You need to use [Robot.createMultiResolutionScreenCapture](https://docs.oracle.com/en/java/javase/17/docs/api/java.desktop/java/awt/Robot.html#createMultiResolutionScreenCapture(java.awt.Rectangle)) to get the real screen pixels rather than scaled down image as if uiScale=1.0.

You may need to tweak the position of the capture and the area where you sample the pixels.

TejeshR13: Yeah, verifying with JTable will be better idea.

prrace (changes requested):
This change will have broad impact everywhere the Border is used.
And it seems to me to be a workaround for the problem that Metal isn't
correctly calculating the length of the text at the scaled resolution.

SwingUtilities2.stringWidth()  ought to be using the correct FontMetrics here for the scale.
Or perhaps its not being used at all.
So go find the code that does the measurement and then positions the text and understand why it is wrong.
Perhaps its rounding the x location incorrectly for the right aligned text ?

aivanov-jdk:
> This change will have broad impact everywhere the Border is used. And it seems to me to be a workaround for the problem that Metal isn't correctly calculating the length of the text at the scaled resolution.

It is possible that the width of the sting isn't calculated correctly, which could explain “cut off” label. At the same time, the text flushed to the very edge doesn't look either. When there's a margin, text looks better and is easier to read. Perhaps, we should address both issues.

prrace:
> > This change will have broad impact everywhere the Border is used. And it seems to me to be a workaround for the problem that Metal isn't correctly calculating the length of the text at the scaled resolution.
> 
> It is possible that the width of the sting isn't calculated correctly, which could explain “cut off” label. At the same time, the text flushed to the very edge doesn't look either. When there's a margin, text looks better and is easier to read. Perhaps, we should address both issues.

So long as such a margin doesn't just become another workaround fix, so

Step 1 - measure it correctly
Step 2 - add a pixel of clear margin space for added aesthetics.

TejeshR13:
> > > This change will have broad impact everywhere the Border is used. And it seems to me to be a workaround for the problem that Metal isn't correctly calculating the length of the text at the scaled resolution.
> > 
> > 
> > It is possible that the width of the sting isn't calculated correctly, which could explain “cut off” label. At the same time, the text flushed to the very edge doesn't look either. When there's a margin, text looks better and is easier to read. Perhaps, we should address both issues.
> 
> So long as such a margin doesn't just become another workaround fix, so
> 
> Step 1 - measure it correctly Step 2 - add a pixel of clear margin space for added aesthetics.

I have been spending some time in analyzing the String width and position, they seems to be fine. And what I have observed is the border of JTableHeader seems to be having an offset (from 100% to 225%), i.e., it is varying w.r.t scaling, maybe a rounding factor too. Since the border is moved to left slightly (maybe 1pixel), so we are seeing that the string is "cut off". I guess if we work on correcting the JTableMetal Border along with adding a margin space (Right), the issue will be fixed? (Screen shot added - Maybe have to zoom-in to see the difference in border w.r.t data cell border)

JTableSizeHeaderScalingComparision

aivanov-jdk:
> Since the border is moved to left slightly (maybe 1pixel), so we are seeing that the string is "cut off". I guess if we work on correcting the JTableMetal Border along with adding a margin space (Right), the issue will be fixed? (Screen shot added - Maybe have to zoom-in to see the difference in border w.r.t data cell border)

It's very likely the reason why. We saw that borders may render inconsistently: 1-, 2- or 3-pixel lines (for 225%).

I zoomed in your screenshot. The 100%-scale case has 1-pixel shadow and 1-pixel highlight. It seems the text is cut off by 1 pixel, the black pixels of ‘e’ are at the very edge of the shadow pixels. I'd expect to see blue-ish pixels of text antialiasing to the right of the black text pixels.

In 225%-scale case, the first separator looks better. It consists of 2-pixel shadow, 1-pixel background, 1-pixel highlight for the next header. On the background-colored pixel, I see pixels of text, blue-ish pixels of antialiasing. So the position of the text is still a bit off. It's no wonder if you take into account that it's 1-pixel off even for 100% case.

The second separator has more issues: 2-pixel shadow, 2-pixel background, 1-pixel highlight. There should be no background pixels between the shadow and highlight but there are… It's an inherent problem of scaled graphics. That background pixels also have parts of the text, 2 pixels.

In short, it seems there's still a text positioning issue… as if it doesn't account for antialiasing.

TejeshR13:
> > Since the border is moved to left slightly (maybe 1pixel), so we are seeing that the string is "cut off". I guess if we work on correcting the JTableMetal Border along with adding a margin space (Right), the issue will be fixed? (Screen shot added - Maybe have to zoom-in to see the difference in border w.r.t data cell border)
> 
> It's very likely the reason why. We saw that borders may render inconsistently: 1-, 2- or 3-pixel lines (for 225%).
> 
> I zoomed in your screenshot. The 100%-scale case has 1-pixel shadow and 1-pixel highlight. It seems the text is cut off by 1 pixel, the black pixels of ‘e’ are at the very edge of the shadow pixels. I'd expect to see blue-ish pixels of text antialiasing to the right of the black text pixels.
> 
> In 225%-scale case, the first separator looks better. It consists of 2-pixel shadow, 1-pixel background, 1-pixel highlight for the next header. On the background-colored pixel, I see pixels of text, blue-ish pixels of antialiasing. So the position of the text is still a bit off. It's no wonder if you take into account that it's 1-pixel off even for 100% case.
> 
> The second separator has more issues: 2-pixel shadow, 2-pixel background, 1-pixel highlight. There should be no background pixels between the shadow and highlight but there are… It's an inherent problem of scaled graphics. That background pixels also have parts of the text, 2 pixels.
> 
> In short, it seems there's still a text positioning issue… as if it doesn't account for antialiasing.

Ok, in summary there exists 2 major issues:
1. The text should positioned till the blue-ish pixels is just behind the header border for each header cell.
2. The header border to be aligned with data grid lines.
Right?

aivanov-jdk:
> Ok, in summary there exists 2 major issues:

Yes.

> 1. The text should positioned till the blue-ish pixels is just behind the header border for each header cell.

Let's focus on this one here. Position the text so that it touches the shadow line of the header, the add the 2-pixel margin on the right just like it's done for other sides.

> 2. The header border to be aligned with data grid lines.

There have been a few instances where border rendering is fixed by removing the scale transform. The same can be done here. If we go for it, it's better to do it separately.

TejeshR13:
@prrace @aivanov-jdk I don't see any difference w.r.t to Label position/width for different scaling factors. 
The width is computed in [stringWidth(String)](https://github.com/openjdk/jdk/blob/48e61c1df53c11ed49603abd70a3dd62a25f7be5/src/java.desktop/share/classes/sun/font/FontDesignMetrics.java#L457) which is same for all scaling factors (though their are slight differences between scaling factor, its corrected due to 0.5 addition at the end). The position is computed in [layoutCompoundLabelImpl](https://github.com/openjdk/jdk/blame/526dba1a2942e444bf11d03d8eaf014b5ef20ccf/src/java.desktop/share/classes/javax/swing/SwingUtilities.java#L1073) which seems to be correct, since the calculation for XPos which is `dx = (viewR.x + viewR.width) - (labelR_x + labelR_width);` seems straight forward. 
Is there anything else which I should check or can we move ahead with current fix i.e., setting margin to 2pixels for RIGHT Aligned header text for Metal L&F?
Parallelly I will raise a bug and PR for header border alignment.

aivanov-jdk:
@TejeshR13 I agree it looks correct, I can't spot any differences. Does `JLabel` always position the text correctly? If you make `JLabel` right-aligned, do you see any pixels which aren't painted as expected?

TejeshR13:
> @TejeshR13 I agree it looks correct, I can't spot any differences. Does `JLabel` always position the text correctly? If you make `JLabel` right-aligned, do you see any pixels which aren't painted as expected?

I didn't any differences in pixels, nor any computational issues specific to right-alignment.

TejeshR13:
@prrace @aivanov-jdk @kumarabhi006 Any comments on this PR?

test/jdk/javax/swing/JTableHeader/JTableHeaderLabelRightAlignTest.java
```diff
@@ -0,0 +1,117 @@
...
+import java.awt.image.BufferedImage;
+
+import java.io.File;
+import javax.imageio.ImageIO;
+
+/*
+ * @test
+ * @bug 8301606
+ * @library /java/awt/regtesthelpers
+ * @build PassFailJFrame
```
aivanov-jdk, line 42:
```suggestion
```
The test is no longer manual, it uses no libraries.

test/jdk/javax/swing/JTableHeader/JTableHeaderLabelRightAlignTest.java
```diff
@@ -0,0 +1,117 @@
...
+        int expectedRGB;
+        BufferedImage imgHeader;
+        double w;
+        double h;
+        String[][] data = {
+                { "1", "1", "Green"},
+                { "2", "2", "Blue"}
+        };
+
+        String[] columnNames = { "Size", "Size", "Size"};
```
aivanov-jdk, line 76: Do you really need three columns and two rows?

aivanov-jdk (changes requested):

test/jdk/javax/swing/JTableHeader/JTableHeaderLabelRightAlignTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+        table = new JTable(data, columnNames);
+        table.setSize(WIDTH,HEIGHT);
+        ((JLabel)table.getTableHeader().getDefaultRenderer())
+                .setHorizontalAlignment( JLabel.RIGHT );
+
+        final JTableHeader header = table.getTableHeader();
+        TableCellRenderer renderer = header.getDefaultRenderer();
+        header.setDefaultRenderer(renderer);
```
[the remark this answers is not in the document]
TejeshR13: Yeah, not required.

test/jdk/javax/swing/JTableHeader/JTableHeaderLabelRightAlignTest.java
```diff
@@ -0,0 +1,117 @@
...
+            g2d.dispose();
+        }
+
+        verticalLineCol = (int)(table.getTableHeader().
+                getColumnModel().getColumn(0).getWidth() * SCALE);
+        expectedRGB = imgHeader.getRGB(verticalLineCol, 1);
+
+        for (int i = 1; i < imgHeader.getHeight()-3; i++) {
+            for (int j = verticalLineCol; j < verticalLineCol + 1; j++) {
+                if (expectedRGB != imgHeader.getRGB(j, i)) {
```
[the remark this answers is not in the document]
TejeshR13: For height, I am using BufferedImage height, not TableHeader height. So I guess it is fine.

test/jdk/javax/swing/JTableHeader/JTableHeaderLabelRightAlignTest.java
TejeshR13: Not required. Reduced to two columns and one rows.

test/jdk/javax/swing/JTableHeader/JTableHeaderLabelRightAlignTest.java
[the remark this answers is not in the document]
aivanov-jdk: Blimey! Sorry for my confusion, you're right.

test/jdk/javax/swing/JTableHeader/JTableHeaderLabelRightAlignTest.java
```diff
@@ -94,10 +90,10 @@ private static void test() {
                     try {
                         ImageIO.write(imgHeader, "png",
                                 new File("FailureImage.png"));
-                    } catch (IOException ioException){
+                    } catch (IOException ioException) {
                         ioException.printStackTrace();
                     }
-                    throw new RuntimeException("Test Failed");
+                    throw new RuntimeException("Test Failed at Pos_x : " + x + ", Pos_y : " + y );
```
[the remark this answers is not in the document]
TejeshR13: I don't think we should verify the colors to the left of shadow border, for this particular test verifying right column of header border would do the job. VerticalLine is at the same location, can remove the for-loop.

TejeshR13:
@aivanov-jdk @kumarabhi006 Please let me know if there are any more review suggestions.

test/jdk/javax/swing/JTableHeader/JTableHeaderLabelRightAlignTest.java
```diff
@@ -0,0 +1,101 @@
...
+        for (int y = 1; y < (imgHeader.getHeight() - 3); y++) {
+            if (expectedRGB != imgHeader.getRGB(x, y)) {
+                try {
+                    ImageIO.write(imgHeader, "png",
+                            new File("FailureImage.png"));
+                } catch (IOException ioException) {
+                    ioException.printStackTrace();
+                }
+                throw new RuntimeException("Test Failed at <" +
+                        x + ", " + y + ">");
```
aivanov-jdk, line 96:
```suggestion
                throw new RuntimeException("Test Failed at <" + x + ", " + y + ">");
```
It doesn't fit 80-column limit, but I think it is okay to keep it on one line without wrapping. It doesn't affect the readability of the code, and it's only 5 chars above the limit.

test/jdk/javax/swing/JTableHeader/JTableHeaderLabelRightAlignTest.java
```diff
@@ -0,0 +1,101 @@
...
+        Graphics2D g2d = imgHeader.createGraphics();
+        g2d.scale(SCALE, SCALE);
+        try {
+            header.paint(g2d);
+        } finally {
+            g2d.dispose();
+        }
+
+        int x = (int)(table.getTableHeader().
+                getColumnModel().getColumn(0).getWidth() * SCALE);
```
aivanov-jdk, line 84:
```suggestion
        int x = (int)(table.getTableHeader()
                .getColumnModel().getColumn(0).getWidth() * SCALE);
```
It's a matter of (my) preference, and it's stated in Java Coding Style that wrapping the operator to the next line makes it clearer it's a continuation line.

I'd wrap on each method call:
```java
        int x = (int)(table.getTableHeader()
                           .getColumnModel()
                           .getColumn(0)
                           .getWidth() * SCALE);
```

aivanov-jdk (approved):
Looks good to me.

aivanov-jdk (approved):

test/jdk/javax/swing/JTableHeader/JTableHeaderLabelRightAlignTest.java
```diff
@@ -88,15 +88,18 @@ private static void test() {
...
-                    ioException.printStackTrace();
-                }
+                saveBufferedImage(imgHeader, "failureImage.png");
                 throw new RuntimeException("Test Failed at <" + x + ", " + y + ">");
             }
         }
         System.out.println("Test Passed");
     }
+
+    private static void saveBufferedImage(BufferedImage image, String fileName) {
```
aivanov-jdk, line 98: `saveImage` should be enough as in #14766?

TejeshR13: I have used `saveBufferedImage` in both the test/PR, I think it should be fine.

aivanov-jdk: My point is that `saveImage` clearly conveys the meaning, you pass a `BufferedImage`, which implies you save a buffered image. Whatever, but shorter name is better in my opinion, and it's used in many other tests.

test/jdk/javax/swing/JTableHeader/JTableHeaderLabelRightAlignTest.java
```diff
@@ -46,10 +46,16 @@ public class JTableHeaderLabelRightAlignTest {
...
     private static final double SCALE = 2.25;
+    private static boolean testStatus = true;
+    private static String failureMsg;
 
     public static void main(String[] args) throws Exception {
         UIManager.setLookAndFeel("javax.swing.plaf.metal.MetalLookAndFeel");
         SwingUtilities.invokeAndWait(JTableHeaderLabelRightAlignTest::test);
+        if (!testStatus) {
+            throw new RuntimeException(failureMsg);
+        }
```
aivanov-jdk, line 57:
```suggestion
        if (failureMsg != null) {
            throw new RuntimeException(failureMsg);
        }
```
`testStatus` is redudant.

aivanov-jdk (changes requested):
I still prefer the previous version where the exception was thrown directly.

In fact, you don't create real UI, the test doesn't depend on processing events — the entire test can be run on main thread *safely*. Get rid of `invokeAndWait` altogether?

test/jdk/javax/swing/JTableHeader/JTableHeaderLabelRightAlignTest.java
```diff
@@ -0,0 +1,98 @@
...
+    public static void main(String[] args) throws Exception {
+        UIManager.setLookAndFeel("javax.swing.plaf.metal.MetalLookAndFeel");
+        String[][] data = {
+                {"1", "1"}
+        };
+
+        String[] columnNames = {"Size", "Size"};
+
+        JTable table = new JTable(data, columnNames);
+        table.setSize(WIDTH, HEIGHT);
```
aivanov-jdk, line 59: Setting size of the table seems unnecessary, the test works correctly without it. If so, both `WIDTH` and `HEIGHT` could be removed.

test/jdk/javax/swing/JTableHeader/JTableHeaderLabelRightAlignTest.java
```diff
@@ -0,0 +1,98 @@
...
+        Graphics2D g2d = imgHeader.createGraphics();
+        g2d.scale(SCALE, SCALE);
+        try {
+            header.paint(g2d);
+        } finally {
+            g2d.dispose();
+        }
+
+        int x = (int) (table.getTableHeader().getColumnModel().getColumn(0)
+                            .getWidth() * SCALE);
```
aivanov-jdk, line 79:
```suggestion
        int x = (int)(table.getTableHeader()
                           .getColumnModel()
                           .getColumn(0)
                           .getWidth() * SCALE);
```

To make it look similar to `TableHeaderBorderPositionTest.java`:

https://github.com/openjdk/jdk/blob/be04454fce1e298655154603700580d11bad2c5f/test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java#L82-L85

After all, these two tests are very similar.

test/jdk/javax/swing/JTableHeader/JTableHeaderLabelRightAlignTest.java
```diff
@@ -0,0 +1,98 @@
...
+        ((JLabel)table.getTableHeader().getDefaultRenderer())
+                .setHorizontalAlignment(JLabel.RIGHT);
+
+        final JTableHeader header = table.getTableHeader();
+        Dimension size = header.getPreferredSize();
+        header.setSize(size);
+
+        int w = (int)Math.ceil(SCALE * size.width);
+        int h = (int)Math.ceil(SCALE * size.height);
+        BufferedImage imgHeader = new BufferedImage(w, h, BufferedImage.TYPE_INT_RGB);
```
aivanov-jdk, line 69:
```suggestion
        int w = (int)Math.ceil(SCALE * size.width);
        int h = (int)Math.ceil(SCALE * size.height);
        BufferedImage imgHeader =
                new BufferedImage((int)Math.ceil(SCALE * size.width),
                        (int)Math.ceil(SCALE * size.height),
                        BufferedImage.TYPE_INT_RGB);
```
Inline `w` and `h` as in `TableHeaderBorderPositionTest.java`?

https://github.com/openjdk/jdk/blob/be04454fce1e298655154603700580d11bad2c5f/test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java#L66-L69

aivanov-jdk (changes requested):

aivanov-jdk (approved):

aivanov-jdk (approved):

PR #14465, opened 2023-06-14
8303916: ThreadLists.java inconsistent results

kevinjwalls:
This test fails when additional threads are created, as it compares thread lists from different sources and sees a mismatch.

It needs to tolerate some threads with certain names appearing, e.g. JVMCI and ForkJoin threads.  If there are extra threads in the thread ID list from the ThreadMBean, remove them from the thread count before comparing.

When it fails, additionally print the ids and names of threads that it got from the MBean.

I found the test quite confusing, the naming of thread counts, arrays of threads, and arrays of thread stacktraces from different sources, so have renamed some things.

---------
### Progress

### Issue
 * [JDK-8303916](https://bugs.openjdk.org/browse/JDK-8303916): ThreadLists.java inconsistent results (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14465/head:pull/14465` \
`$ git checkout pull/14465`

`$ git checkout pull/14465` \
`$ git pull https://git.openjdk.org/jdk.git pull/14465/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14465`

`$ git pr show -t 14465`

</details>

<a href="https://git.openjdk.org/jdk/pull/14465.diff">https://git.openjdk.org/jdk/pull/14465.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14465#issuecomment-1591037835)


plummercj (commented):
Overall looks good, but I'm surprised you only need to filter "ForkJoinPool" and "JVMCI". What about threads like "Common-Cleaner"? 

https://github.com/openjdk/jdk/blob/57b8251241e2044d5039ce162bf4637a9b2e5466/test/hotspot/jtreg/vmTestbase/nsk/share/jdi/EventFilters.java#L333

kevinjwalls:
> 

Thanks - that's a useful reference list.  For this test, it's never managed to not have Common-Cleaner in the initial set of threads it observes, I think...  But it's very easy to see a couple of ForkJoinPool-x-worker-x threads, and see one more of them by the time it gets threads from the MXBean.  That's the way I see it failing for us now.  If we do see others appear, I hope having added the extra printing it will be easy to add other names, if they affect this test...

dholmes-ora (approved):
The test assumes that it can find a set of steady-state threads running that does not change, but that is not the case. I don't find this test particularly useful in what it attempts to do. It would make more sense to have a set of test threads that it looks for via each method, and which it can control termination of and thus check that they disappear. Adding an allow-list of dynamically appearing threads seems somewhat fragile, but I guess we've really only seen two problematic scenarios:
- Running with virtual threads
- Running with JVMCI
so the fix "works".

plummercj (approved):

lmesnik (approved):

AlanBateman:
I agree with David and would be concerned this test will be whack-a-mole to deal with other cases. It would be a lot simpler to have a test that creates a number of threads, then uses Thread.getAllThreadStacks + ThreadMXBean.getAllThreadIds to that the expected list of threads are included. If the current thread is not a virtual thread then it can check that it is included too. If you really want to check that both methods return the same list of threads then I think it will need to loop + backoff when the threads aren't the same.

kevinjwalls:
Thanks for the comments and reviews.

I did look at the history and this test exists because of previous confusion about thread groups, where the top thread group and the MXBean thread lists could be different.  So if this test creates new threads and checks they are in both lists, it loses the original purpose as I read it.

I'm not saying it's a great test.  But it successfully monitored that problem since 2004, and lasted surprisingly well even before being moved to an "othervm" !
 
If there's any mole-whacking, I think the new logging should make it easy to see the name of the mole, and if it's a thread that's expected to appear possibly between the observations, then it can be added to the list.  It is just those two new things, JVMCI and ForkJoinPool, which are known to fool it so far.  We should make this change and watch for moles appearing, or indeed disappearing.

dholmes-ora:
> I did look at the history and this test exists because of previous confusion about thread groups, where the top thread group and the MXBean thread lists could be different. So if this test creates new threads and checks they are in both lists, it loses the original purpose as I read it.

Can you elaborate what the issue was please. I would have to suspect the wrong "top" ThreadGroup was being used (main vs. system).

kevinjwalls:
> 

Sure, this was:

5047639: threadGroup.enumerate() ignores the Signal Dispatcher thread
https://bugs.openjdk.org/browse/JDK-5047639
(2004)

top ThreadGroup enumerate returns one less thread than ThreadMBean's getThreadCount.

Notes: SignalDispatcher started by the VM, question on whether ThreadMBean should include system threads.
Threads only added to a group if started with Thread.start()



The test wants to ensure that thread group and mbean are consistent, it doesn't particularly look for any thread belonging to a particular group, but would be happy as long as all threads are either in both samples or neither.

PR #14466, opened 2023-06-14
8049347: HTMLDocument throws NPE for Block Tag

prsadhuk:
Trying to iterate for HEAD HTML tag or for any other "Block" Tag results in NPE due to some spurious unhandled code which just return NULL.
Removing the code solves this problem and also does not cause any regression in clientlibs test.

---------
### Progress

### Integration blocker
&nbsp;⚠️ Title mismatch between PR and JBS for issue [JDK-8049347](https://bugs.openjdk.org/browse/JDK-8049347)

### Issue
 * [JDK-8049347](https://bugs.openjdk.org/browse/JDK-8049347): HTMLDocument.getIterator throws NPE for Block Tag (**Enhancement** - P4) ⚠️ Title mismatch between PR and JBS.

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14466/head:pull/14466` \
`$ git checkout pull/14466`

`$ git checkout pull/14466` \
`$ git pull https://git.openjdk.org/jdk.git pull/14466/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14466`

`$ git pr show -t 14466`

</details>

<a href="https://git.openjdk.org/jdk/pull/14466.diff">https://git.openjdk.org/jdk/pull/14466.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14466#issuecomment-1591137454)


src/java.desktop/share/classes/javax/swing/text/html/HTMLDocument.java
```diff
@@ -545,10 +545,6 @@ public StyleSheet getStyleSheet() {
      * @see javax.swing.text.html.HTML.Tag
      */
     public Iterator getIterator(HTML.Tag t) {
-        if (t.isBlock()) {
-            // TBD
-            return null;
```
aivanov-jdk, line 550: This doesn't look right to me. The comment suggests the implementation of something like `BlockIterator` is missing.

I would expect that iterator for `HTML.Tag.HEAD` returns all elements inside `<head>` element. Yet the comment for the `next` method

https://github.com/openjdk/jdk/blob/931625a9304ec2761ca9035d69fd33f6beadb124/src/java.desktop/share/classes/javax/swing/text/html/HTMLDocument.java#L1993-L1997

suggests, the iterator goes over all instances of the specified tag in the document.

What does the iterator return if you pass it `HTML.Tag.HEAD` or `HTML.Tag.P`?

prsadhuk: It returns a LeafIterator object with HEAD or P tag respectively

aivanov-jdk: Obviously. :)

Does it do what the `getIterator` method promise to do? It doesn't because it iterates over `LeafElement`-s only:

https://github.com/openjdk/jdk/blob/931625a9304ec2761ca9035d69fd33f6beadb124/src/java.desktop/share/classes/javax/swing/text/html/HTMLDocument.java#L1993-L1998

So, the `next` method moves the iterator to the next leaf element:

https://github.com/openjdk/jdk/blob/931625a9304ec2761ca9035d69fd33f6beadb124/src/java.desktop/share/classes/javax/swing/text/html/HTMLDocument.java#L2037-L2044

Note that it uses `isLeaf` to stop the generic `ElementIterator`.

---

It looks as if `HTMLDocument.getIterator` is public as an implementation detail: it's used to iterate over `HTML.Tag.A` to scroll the text component to the anchor and to determine whether the document is a frame set or not.

Removing the `if` condition removes the NPE but the method doesn't support Block Elements.

Should we update the javadoc to state the method shouldn't be used by apps? And to throw `UnsupportedOperationException` if a block tag is passed instead of returning null?

aivanov-jdk: > > Does it do what the `getIterator` method promise to do? It doesn't because it iterates over `LeafElement`-s only:
> 
> As per the spec wordings, _The [getIterator(HTML.Tag t)](https://docs.oracle.com/en/java/javase/20/docs/api/java.desktop/javax/swing/text/html/HTMLDocument.html#getIterator(javax.swing.text.html.HTML.Tag)) method can also be used for finding all occurrences of the specified HTML tag in the document._
> 
> which I think is what we get with the fix an Iterator object with next block tag

Really? All the evidence I provided above shows the current implementation *skips* block tags because it takes into account only elements for which [`isLeaf`](https://github.com/openjdk/jdk/blob/83d92672d4c2637fc37ddd873533c85a9b083904/src/java.desktop/share/classes/javax/swing/text/AbstractDocument.java#L2636) method returns `true`, that is `LeafElement`s. Block tags are represented with `BranchElement`, its [`isLeaf`](https://github.com/openjdk/jdk/blob/83d92672d4c2637fc37ddd873533c85a9b083904/src/java.desktop/share/classes/javax/swing/text/AbstractDocument.java#L2481) method returns `false`.

prrace:
I'm inclined to agree this looks like unfinished work.
The original author seems to have thought block tags needed to be handled differently.
I don't see any evidence this was never intended to be used by apps any more than much of the rest of the API.

prsadhuk:
> I'm inclined to agree this looks like unfinished work. The original author seems to have thought block tags needed to be handled differently. I don't see any evidence this was never intended to be used by apps any more than much of the rest of the API.

Yes, I also think it so now that it probably will need a "BlockIterator" implementation as mentioned and calling `UnsupportedOperationException` may not be appropriate and it also does not seem right to update spec to mention the same which will cast in stone and might cause issue later on if we need to support it in future..

PR #14467, opened 2023-06-14
8304835 jdk/jfr/event/oldobject/TestArrayInformation.java fails with "Could not find event with class ... as (leak) object"

egahlin:
Could I have a review of a test fix?

The test fails intermittently because it's hard to guarantee that an object is being sampled. Fix is to repeat until one is found.

Testing: jdk/jdk/event/oldObject/TestArrayInformation.java

Thanks
Erik

---------
### Progress

### Issue
 * [JDK-8304835](https://bugs.openjdk.org/browse/JDK-8304835): jdk/jfr/event/oldobject/TestArrayInformation.java fails with "Could not find event with class ... as (leak) object" (**Bug** - P4)

### Reviewers
 * [Aleksey Shipilev](https://openjdk.org/census#shade) (@shipilev - **Reviewer**) ⚠️ Review applies to [f7cfa4f7](https://git.openjdk.org/jdk/pull/14467/files/f7cfa4f7a78e832dfe6a8e5b4665c2c9859dd5e7)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14467/head:pull/14467` \
`$ git checkout pull/14467`

`$ git checkout pull/14467` \
`$ git pull https://git.openjdk.org/jdk.git pull/14467/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14467`

`$ git pr show -t 14467`

</details>

<a href="https://git.openjdk.org/jdk/pull/14467.diff">https://git.openjdk.org/jdk/pull/14467.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14467#issuecomment-1591171487)

### DIFF
--- a/test/jdk/jdk/jfr/event/oldobject/TestArrayInformation.java
+++ b/test/jdk/jdk/jfr/event/oldobject/TestArrayInformation.java
@@ -56,20 +56,26 @@ public class TestArrayInformation {
     public static void main(String[] args) throws Exception {
         WhiteBox.setWriteAllObjectSamples(true);
 
-        try (Recording recording = new Recording()) {
-            recording.enable(EventNames.OldObjectSample).withoutStackTrace().with("cutoff", "infinity");
-            recording.start();
-            for(int i = 0; i < 25; i++) {
-              leak.add( buildNestedArray(CHAIN_DEPTH));
+        while (true) {
+            try (Recording recording = new Recording()) {
+                recording.enable(EventNames.OldObjectSample).withoutStackTrace().with("cutoff", "infinity");
+                recording.start();
+                for(int i = 0; i < 25; i++) {
+                  leak.add( buildNestedArray(CHAIN_DEPTH));
+                }
+                recording.stop();
+                List<RecordedEvent> events = Events.fromRecording(recording);
+                Events.hasEvents(events);
+                if (verifyObjectArray(events)) {
+                    return;
+                }
             }
-            recording.stop();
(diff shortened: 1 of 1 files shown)


shipilev (commented):
So now the "failing" test would timeout? This does not look right.

How many tries it usually takes to get a lucky sample? Take that estimate and do about 10x more iterations and fail, if still not found?

shipilev:
> > So now the "failing" test would timeout? This does not look right.
> > How many tries it usually takes to get a lucky sample? Take that estimate and do about 10x more iterations and fail, if still not found?
> 
> The test will timeout, not sure that is a problem. Looking at the log, it's obvious why. I don't know how many times are needed, so why not run as many as possible? This is the pattern we have used for other jfr tests.

Well, my concern is that timeouts are mostly treated as the infrastructural problems. I don't think anyone expects the test failure to manifest as timeout: the expectation is that tests complete in reasonable time, whether they pass or fail. Running until timeout is reached also makes the test dependent on TIMEOUT_FACTOR and other infrastructural tunings.

But you're right, JFR tests already do this, so it makes little sense in trying to address it in this PR. It would be nice to see if JFR tests can avoid timeouting on failure, some time later, though.

shipilev:
Merge from master to get Windows GHA builds, and thus Windows GHA tests fixed.

shipilev (approved):
Apart from a stylistic nit below, looks fine.

mgronlun (approved):

PR #14472, opened 2023-06-14
8310037: Parallel/Serial: Space Counters classes leak helper instances

zhengyu123:
Please review this small fix for memory leak.

---------
### Progress

### Issue
 * [JDK-8310037](https://bugs.openjdk.org/browse/JDK-8310037): Parallel/Serial: Space Counters classes leak helper instances (**Bug** - P4)

### Reviewers
 * [Yumin Qi](https://openjdk.org/census#minqi) (@yminqi - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14472/head:pull/14472` \
`$ git checkout pull/14472`

`$ git checkout pull/14472` \
`$ git pull https://git.openjdk.org/jdk.git pull/14472/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14472`

`$ git pr show -t 14472`

</details>

<a href="https://git.openjdk.org/jdk/pull/14472.diff">https://git.openjdk.org/jdk/pull/14472.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14472#issuecomment-1591656147)

### DIFF
--- a/src/hotspot/share/gc/parallel/spaceCounters.cpp
+++ b/src/hotspot/share/gc/parallel/spaceCounters.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2002, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2002, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
--- a/src/hotspot/share/gc/parallel/spaceCounters.cpp
+++ b/src/hotspot/share/gc/parallel/spaceCounters.cpp
@@ -34,7 +34,8 @@
 
 SpaceCounters::SpaceCounters(const char* name, int ordinal, size_t max_size,
                              MutableSpace* m, GenerationCounters* gc) :
-   _object_space(m) {
+   _object_space(m),
+   _perf_used_helper(nullptr) {
 
   if (UsePerfData) {
     EXCEPTION_MARK;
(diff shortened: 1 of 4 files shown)


src/hotspot/share/gc/parallel/spaceCounters.cpp
```diff
@@ -59,8 +60,9 @@ SpaceCounters::SpaceCounters(const char* name, int ordinal, size_t max_size,
                                    _object_space->capacity_in_bytes(), CHECK);
 
     cname = PerfDataManager::counter_name(_name_space, "used");
+    _perf_used_helper = new MutableSpaceUsedHelper(_object_space);
     _used = PerfDataManager::create_variable(SUN_GC, cname, PerfData::U_Bytes,
-                                    new MutableSpaceUsedHelper(_object_space),
+                                    _perf_used_helper,
```
tschatzl, line 65: The same issue affects serial gc in the `CSpaceCounters` file; since these are very minor changes, could you do this in the same change (renaming the CR too)?
Otherwise please file a bug for serial gc too.

tschatzl, line 65: Pre-existing: the formatting of `PerfDataManager::create*` calls is rather unusual. I filed [JDK-8310116](https://bugs.openjdk.org/browse/JDK-8310116) for that.

tschatzl (approved):
The change is good, and I'm approving it. The same issue exists for serial gc (in `CSpaceCounters`), so maybe this could be dealt with here too, but feel free to defer to a separate CR.

Fwiw, this seems to be one of those memory leaks of (single) instances that would be cleaned out automatically at VM exit anyway, am I correct?

tschatzl (approved):
Still good. Thanks.

stefank:
> Fwiw, this seems to be one of those memory leaks of (single) instances that would be cleaned out automatically at VM exit anyway, am I correct?

And IIUC this code will never be called. How did you test the added code?

yminqi (approved):
LGTM

zhengyu123:
> > Fwiw, this seems to be one of those memory leaks of (single) instances that would be cleaned out automatically at VM exit anyway, am I correct?
> 
> And IIUC this code will never be called. How did you test the added code?

It does seem that the destructors are never called. I have no idea why it shows up in memory leak profiling, while `_space_name` does not.

zhengyu123:
Since we never deallocate counters, there is no point to fix this.

PR #14473, opened 2023-06-14
8167252: Some of Charset.availableCharsets() does not contain itself

naotoj:
Adding themselves into their `contains()` method will fix it.

---------
### Progress

### Issue
 * [JDK-8167252](https://bugs.openjdk.org/browse/JDK-8167252): Some of Charset.availableCharsets() does not contain itself (**Bug** - P4)

### Reviewers
 * [Iris Clark](https://openjdk.org/census#iris) (@irisclark - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14473/head:pull/14473` \
`$ git checkout pull/14473`

`$ git checkout pull/14473` \
`$ git pull https://git.openjdk.org/jdk.git pull/14473/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14473`

`$ git pr show -t 14473`

</details>

<a href="https://git.openjdk.org/jdk/pull/14473.diff">https://git.openjdk.org/jdk/pull/14473.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14473#issuecomment-1591651303)

### DIFF
--- a/test/jdk/java/nio/charset/Charset/Contains.java
+++ b/test/jdk/java/nio/charset/Charset/Contains.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -23,7 +23,7 @@
 
 /* @test
  * @summary Unit test for charset containment
- * @bug 6798572
+ * @bug 6798572 8167252
  * @modules jdk.charsets
  */
 
@@ -93,6 +93,8 @@ public class Contains {
         ck(cp1252, cp1252, true);
 
         checkUTF();
+
+        selfContainmentTest();
     }
 
     static void checkUTF() throws Exception {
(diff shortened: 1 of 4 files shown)


bplb (approved):
Looks all right.

AlanBateman (approved):

irisclark (approved):

test/jdk/java/nio/charset/Charset/Contains.java
```diff
@@ -103,6 +105,24 @@ static void checkUTF() throws Exception {
                    true);
     }
 
+    static void selfContainmentTest() {
```
LanceAndersen, line 108: Be nice to add a comment regarding the purpose of the test

LanceAndersen (approved):
Changes look fine over all.  

It might be worth considering updating this to use junit at some point as a separate clean up activity as it could further streamline the test

test/jdk/java/nio/charset/Charset/Contains.java
AlanBateman: The spec is that Charset::contains return true for self so maybe it could be renamed to containsSelfTest, might help with the understanding the purpose (along with a comment)

naotoj: Good point. Will address these in the next iteration

AlanBateman (approved):

LanceAndersen (approved):

bplb (approved):
Re-approved.

irisclark (approved):

jaikiran:
Hello Naoto, should `sun.util.PropertyResourceBundleCharset` be fixed too? Its `contains()` currently just returns `false`. Then there is `sun.font.X11Johab` in `java.desktop` module whose `contains()` appears to have a typo and is checking against an unrelated `X11GB18030_1` instance.

AlanBateman:
> Hello Naoto, should `sun.util.PropertyResourceBundleCharset` be fixed too? 

This is JDK internal, it shouldn't leak out via the APIs. If there is a way for it to leak out then it would require a compliant contains method but I suspect it's internal only.

jaikiran:
Hello Alan,

> > Hello Naoto, should `sun.util.PropertyResourceBundleCharset` be fixed too?
> 
> This is JDK internal, it shouldn't leak out via the APIs. If there is a way for it to leak out then it would require a compliant contains method but I suspect it's internal only.

You are right this is an internal class (I just found these by checking which classes extend `Charset`). I now paid a closer attention to the title of this issue and this is more about the Charset(s) that are returned through the public API, so I think it's OK to leave these internal classes as-is at least in this PR.

naotoj:
Thanks for the review, Jai.
As to `PropertyResourceBundleCharset`, yes it is an internal implementation as Alan mentioned, to auto detect UTF-8/ISO-8859-1, so even though `contains()` does not include itself, it is fine.

PR #14475, opened 2023-06-14
8310053: VarHandle and slice handle derived from layout are lacking alignment check

JornVernee:
Add missing alignment checks for the alignment constraint of the root layout of a VarHandle created through `MemoryLayout::varHandle` and a MethodHandle `MemoryLayout::sliceHandle`.

Testing: `jdk_foreign` test suite

---------
### Progress
- [x] Change requires CSR request [JDK-8310068](https://bugs.openjdk.org/browse/JDK-8310068) to be approved

### Issues
 * [JDK-8310053](https://bugs.openjdk.org/browse/JDK-8310053): VarHandle and slice handle derived from layout are lacking alignment check (**Bug** - P3)
 * [JDK-8310068](https://bugs.openjdk.org/browse/JDK-8310068): VarHandle and slice handle derived from layout are lacking alignment check (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14475/head:pull/14475` \
`$ git checkout pull/14475`

`$ git checkout pull/14475` \
`$ git pull https://git.openjdk.org/jdk.git pull/14475/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14475`

`$ git pr show -t 14475`

</details>

<a href="https://git.openjdk.org/jdk/pull/14475.diff">https://git.openjdk.org/jdk/pull/14475.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14475#issuecomment-1592049952)

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
@@ -440,6 +440,10 @@ public sealed interface MemoryLayout permits SequenceLayout, GroupLayout, Paddin
      * long size = select(elements).byteSize();
      * MemorySegment slice = segment.asSlice(offset, size);
      * }
+     * <p>
+     * The segment to be must be <a href="MemorySegment.html#segment-alignment">aligned</a> according to the {@linkplain
+     * #byteAlignment() alignment constraint} of the root layout (this layout). Note that this can be more strict
+     * (but not less) than the alignment constraint of the selected value layout.
      *
      * @apiNote The returned method handle can be used to obtain a memory segment slice, similarly to {@link MemorySegment#asSlice(long, long)},
      * but more flexibly, as some indices can be specified when invoking the method handle.
--- a/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
@@ -380,7 +380,7 @@ public sealed interface MemoryLayout permits SequenceLayout, GroupLayout, Paddin
      * Additionally, the provided dynamic values must conform to bounds which are derived from the layout path, that is,
      * {@code 0 <= x_i < b_i}, where {@code 1 <= i <= n}, or {@link IndexOutOfBoundsException} is thrown.
      * <p>
-     * The accessed address must be <a href="MemorySegment.html#segment-alignment">aligned</a> according to the {@linkplain
+     * The base address must be <a href="MemorySegment.html#segment-alignment">aligned</a> according to the {@linkplain
      * #byteAlignment() alignment constraint} of the root layout (this layout). Note that this can be more strict
      * (but not less) than the alignment constraint of the selected value layout.
      * <p>
--- a/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
@@ -441,9 +441,9 @@ public sealed interface MemoryLayout permits SequenceLayout, GroupLayout, Paddin
      * MemorySegment slice = segment.asSlice(offset, size);
      * }
      * <p>
-     * The segment to be must be <a href="MemorySegment.html#segment-alignment">aligned</a> according to the {@linkplain
-     * #byteAlignment() alignment constraint} of the root layout (this layout). Note that this can be more strict
-     * (but not less) than the alignment constraint of the selected value layout.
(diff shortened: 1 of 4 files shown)


src/java.base/share/classes/jdk/internal/foreign/LayoutPath.java
```diff
@@ -194,16 +197,11 @@ public VarHandle dereferenceHandle(boolean adapt) {
...
 
         VarHandle handle = Utils.makeSegmentViewVarHandle(valueLayout);
-        for (int i = strides.length - 1; i >= 0; i--) {
-            MethodHandle collector = MethodHandles.insertArguments(MH_ADD_SCALED_OFFSET, 2,
-                    strides[i],
-                    bounds[i]);
-            // (J, ...) -> J to (J, J, ...) -> J
-            // i.e. new coord is prefixed. Last coord will correspond to innermost layout
-            handle = MethodHandles.collectCoordinates(handle, 1, collector);
+        handle = MethodHandles.collectCoordinates(handle, 1, offsetHandle());
```
mcimadamore, line 200: nice simplification!

test/jdk/java/foreign/TestLayoutPaths.java
```diff
@@ -134,6 +135,38 @@ public void testByteOffsetHandleBadRange() {
...
+        assertEquals(struct.byteAlignment(), 4);
+
+        try (Arena arena = Arena.ofConfined()) {
+            MemorySegment seg = arena.allocate(struct.byteSize() + 2, struct.byteAlignment()).asSlice(2);
+            assertEquals(seg.address() % JAVA_SHORT.byteAlignment(), 0); // should be aligned
+            assertNotEquals(seg.address() % struct.byteAlignment(), 0); // should not be aligned
+
+            String expectedMessage = "Target offset incompatible with alignment constraints: " + struct.byteAlignment();
+
+            try {
```
mcimadamore, line 152: Can assertThrows be used to simplify this pattern?

mcimadamore (approved):

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -380,6 +380,10 @@ default MethodHandle byteOffsetHandle(PathElement... elements) {
      * Additionally, the provided dynamic values must conform to bounds which are derived from the layout path, that is,
      * {@code 0 <= x_i < b_i}, where {@code 1 <= i <= n}, or {@link IndexOutOfBoundsException} is thrown.
      * <p>
+     * The accessed address must be <a href="MemorySegment.html#segment-alignment">aligned</a> according to the {@linkplain
```
mcimadamore, line 383: Very nice and clear!

test/jdk/java/foreign/TestLayoutPaths.java
JornVernee: Yeah, somewhat. I thought `assertThrows` was only a junit thing, but it looks like TestNG also has an `expectThrows` which does the same and returns the thrown exception so that we can check the message.

JornVernee:
I've switched to `expectThrows`. As discussed offline, I've also dropped the alignment check from the inner accessed element, as this is redundant due to layouts enforcing alignment of nested elements already. So, if the root layout is aligned, so is the element. (this turned out to be simper than I initially thought)

mcimadamore (approved):
Bonus points!

mcimadamore:
I'm wondering... for dereference path elements, we already perform a root layout check as part of the memory segment expansion - e.g. we use the target layout to resize the segment (and perform an alignment check). So in that case there would still be duplicate alignment checks.

In such cases, we need to make sure that either (a) only the first var handle in the dereference chain gets the proper alignment adaptation on the input segment or that (b) they all do, but then we omit the alignment checks that are generated as part of calling VH::get on the address layout (but that might be harder to do, as that logic is shared).

JornVernee:
> only the first var handle in the dereference chain gets the proper alignment adaptation on the input segment

That would be easy enough to do, we just check for `derefAdapters.length == 0`. I'll add this too.

JornVernee:
I've gone ahead and implemented this. I also added a test to make sure that alignment checks still happen when using deref path elements.

I also fixed a minor javadoc issue, in the doc for varHandle. It said that we checked the alignment of the accessed address, but we check the alignment of the base address (a term which is defined earlier).

PR #14476, opened 2023-06-14
8309974: some JVMCI tests fail when VM options include -XX:+EnableJVMCI

dougxc:
This PR fixes a few JVMCI tests to run the `ClassFileInstaller` step with `@run main/othervm` instead of `@run driver` so that it inherits extra VM options passed to jtreg. This is required for these tests as they use JVMCI and module exports are needed when running `ClassFileInstaller` on the test class itself.

Thanks to @jonathan-gibbons for providing this solution.

---------
### Progress

### Issue
 * [JDK-8309974](https://bugs.openjdk.org/browse/JDK-8309974): some JVMCI tests fail when VM options include -XX:+EnableJVMCI (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14476/head:pull/14476` \
`$ git checkout pull/14476`

`$ git checkout pull/14476` \
`$ git pull https://git.openjdk.org/jdk.git pull/14476/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14476`

`$ git pr show -t 14476`

</details>

<a href="https://git.openjdk.org/jdk/pull/14476.diff">https://git.openjdk.org/jdk/pull/14476.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14476#issuecomment-1591908879)

### DIFF
--- a/test/hotspot/jtreg/compiler/jvmci/events/JvmciShutdownEventTest.java
+++ b/test/hotspot/jtreg/compiler/jvmci/events/JvmciShutdownEventTest.java
@@ -35,9 +35,9 @@
  *
  * @build compiler.jvmci.common.JVMCIHelpers
  *        compiler.jvmci.events.JvmciShutdownEventListener
- * @run driver jdk.test.lib.FileInstaller ./JvmciShutdownEventTest.config
+ * @run main/othervm jdk.test.lib.FileInstaller ./JvmciShutdownEventTest.config
  *     ./META-INF/services/jdk.vm.ci.services.JVMCIServiceLocator
- * @run driver jdk.test.lib.helpers.ClassFileInstaller
+ * @run main/othervm jdk.test.lib.helpers.ClassFileInstaller
  *      compiler.jvmci.common.JVMCIHelpers$EmptyHotspotCompiler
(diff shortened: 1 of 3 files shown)


dholmes-ora (approved):
I must admit I am surprised by this situation. I never imagined that `@module` directives would also be applied to the action to run the `ClassfileInstaller` as it should not need anything of that kind - it is just a simple file copying utility. But I can see how jtreg has no way to know that. Makes me wonder though whether `@modules` should only ever be applied to `main/othervm`? We not only recommend using driver mode for `ClassfileInstaller` we actively went through and changed a number of tests to do that!

So the fix seems fine in so far as the situation we have today, but I'm not convinced there isn't a jtreg issue here that needs further examination.

Thanks.

dcubed-ojdk:
If a bug has been filed for this JTREG issue that @jonathan-gibbons is investigating,
then please make sure that the bug for this PR is linked to it.

dougxc:
Thanks for the review David.

/integrate

PR #14478, opened 2023-06-14
8310054: ScrollPane insets are incorrect

aivanov-jdk:
After the size of `ScrollPane` child component changes, it recalculates the size of the scroll bars and hides or shows them as necessary. This situation is handled in `WScrollPanePeer.childResized`:

https://github.com/openjdk/jdk/blob/ee4ab6709ebaf8a1b1e9f297a7c53205987f3eba/src/java.desktop/windows/classes/sun/awt/windows/WScrollPanePeer.java#L104-L109

After [JDK-8297923](https://bugs.openjdk.org/browse/JDK-8297923), `setSpans` is run asynchronously on the toolkit thread. Thus, `setInsets` calculates the incorrect values because `setSpans` hasn't updated the sizes yet.

This is a regression from [JDK-8297923](https://bugs.openjdk.org/browse/JDK-8297923).

**Fix**

The insets are updated in the native implementation of `setSpans` directly.

The native implementation of `setInsets` method is run on the toolkit thread as a synchronous call because the insets are used in [`WScrollPanePeer.initialize`](https://github.com/openjdk/jdk/blob/ee4ab6709ebaf8a1b1e9f297a7c53205987f3eba/src/java.desktop/windows/classes/sun/awt/windows/WScrollPanePeer.java#L62-L67).

---------
### Progress

### Issue
 * [JDK-8310054](https://bugs.openjdk.org/browse/JDK-8310054): ScrollPane insets are incorrect (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14478/head:pull/14478` \
`$ git checkout pull/14478`

`$ git checkout pull/14478` \
`$ git pull https://git.openjdk.org/jdk.git pull/14478/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14478`

`$ git pr show -t 14478`

</details>

<a href="https://git.openjdk.org/jdk/pull/14478.diff">https://git.openjdk.org/jdk/pull/14478.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14478#issuecomment-1591928661)

### DIFF
--- a/src/java.desktop/windows/native/libawt/windows/awt_ScrollPane.cpp
+++ b/src/java.desktop/windows/native/libawt/windows/awt_ScrollPane.cpp
@@ -525,6 +525,7 @@ void AwtScrollPane::_SetSpans(void *param)
             parentWidth, parentHeight, childWidth, childHeight);
         s->RecalcSizes(parentWidth, parentHeight, childWidth, childHeight);
         s->VerifyState();
+        s->SetInsets(env);
     }
 ret:
    env->DeleteGlobalRef(self);
@@ -718,7 +719,7 @@ Java_sun_awt_windows_WScrollPanePeer_setInsets(JNIEnv *env, jobject self)
 {
     TRY
 
-    AwtToolkit::GetInstance().SyncCall(AwtScrollPane::_SetInsets,
+    AwtToolkit::GetInstance().InvokeFunction(AwtScrollPane::_SetInsets,
         env->NewGlobalRef(self));
     // global ref is deleted in _SetInsets()
 
--- a/src/java.desktop/windows/classes/sun/awt/windows/WScrollPanePeer.java
+++ b/src/java.desktop/windows/classes/sun/awt/windows/WScrollPanePeer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1996, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1996, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -24,7 +24,13 @@
  */
 package sun.awt.windows;
 
-import java.awt.*;
+import java.awt.Adjustable;
+import java.awt.Component;
+import java.awt.Dimension;
+import java.awt.Insets;
+import java.awt.Point;
+import java.awt.ScrollPane;
+import java.awt.ScrollPaneAdjustable;
 import java.awt.event.AdjustmentEvent;
 import java.awt.peer.ScrollPanePeer;
 
@@ -105,7 +111,6 @@ final class WScrollPanePeer extends WPanelPeer implements ScrollPanePeer {
         ScrollPane sp = (ScrollPane)target;
         Dimension vs = sp.getSize();
         setSpans(vs.width, vs.height, width, height);
-        setInsets();
     }
 
     synchronized native void setSpans(int viewWidth, int viewHeight,
--- a/test/jdk/java/awt/ScrollPane/ScrollPaneExtraScrollBar.java
+++ b/test/jdk/java/awt/ScrollPane/ScrollPaneExtraScrollBar.java
@@ -23,7 +23,7 @@
 
 /*
   @test
-  @bug 4152524
+  @bug 4152524 8310054
   @requires os.family=="windows"
   @summary Test that scroll pane doesn't have scroll bars visible when it is
   shown for the first time with SCROLLBARS_AS_NEEDED style
--- a/test/jdk/java/awt/ScrollPane/ScrollPaneExtraScrollBar.java
+++ b/test/jdk/java/awt/ScrollPane/ScrollPaneExtraScrollBar.java
@@ -57,7 +57,7 @@ public class ScrollPaneExtraScrollBar {
             sp = new ScrollPane(ScrollPane.SCROLLBARS_AS_NEEDED);
             sp.add(new Button("TEST"));
             f.add("Center", sp);
-            f.pack();
+            // Frame must not be packed, otherwise the bug isn't reproduced
             f.setLocationRelativeTo(null);
             f.setVisible(true);
         });


prrace:
Hmm, I did ask during the previous review if there was any reason to block with
_SetScrollPos or "the case below" - ie  _SetSpans... but perhaps it is more complex than that.

I need some clarification
Before any changes, using SyncCall , _SetSpans was called synchronously.
After the first change to use InvokeFunctionLater,  _SetSpans was called asynchronously .
Now with InvokeFunction,  _SetSpans is again called synchronously

So why do you also need the change to set it in native ?

The difference I can see is that although it is synchronous it now requires jumping on to the Toolkit thread
and so other pending processing on the Toolkit thread will run first.

But I'm not sure that's relevant here.

So can you expand on why you also had to move the call to setInsets() to native ?

honkar-jdk (commented):
@aivanov-jdk If just the `_SetSpans()` on native is changed to run as  `InvokeFunction` instead of `InvokeFunctionLater` both the tests - ScrollPaneLeakTest.java & ScrollPaneExtraScrollBar.java pass without additional code changes to WScrollPanePeer.java.childResized and _setInsets.

aivanov-jdk:
> @aivanov-jdk If just the `_SetSpans()` on native is changed to run as `InvokeFunction` instead of `InvokeFunctionLater` both the tests - ScrollPaneLeakTest.java & ScrollPaneExtraScrollBar.java pass without additional code changes to WScrollPanePeer.java.childResized and _setInsets.

I know it. But I think it's bad. If both `_SetSpans` and `_SetInsets` are synchronous by using `InvokeFunction`, you block EDT to run some code on the toolkit thread twice:

1. EDT calls `_SetSpans`, it gets blocked until `SetSpans` completes on the toolkit thread;
2. EDT calls `_SetInsets`, it gets blocked until `SetInsets` completes on the toolkit thread.

I think blocking is bad. I coalesced these two calls into one asynchronous call. There's no EDT blocking with the same result.

aivanov-jdk:
> Hmm, I did ask during the previous review if there was any reason to block with _SetScrollPos or "the case below" - ie _SetSpans... but perhaps it is more complex than that.

I didn't find any reason to block either.

I considered `_SetInsets` for using `InvokeFunction` or `InvokeFunctionLater`. I deliberately decided not to.

But I eventually missed the case where `SetInsets` *depends* on the results of `SetSpans` when they're called together in `WScrollPanePeer.childResized`.

(In my initial approach, I used `InvokeFunction` because I overlooked `InvokeFunctionLater` at first; they're not close in the header file and in the implementation files. Yet I was bothered by blocking EDT where it wasn't needed. I found `InvokeFunctionLater` and used it — the result was the same.)

> I need some clarification Before any changes, using SyncCall , _SetSpans was called synchronously. After the first change to use InvokeFunctionLater, _SetSpans was called asynchronously .

Yes, when using `SyncCall` both `_SetSpans` and `_SetInsets` were called synchronously, so `_SetInsets` could use the updated spans.

> Now with InvokeFunction, _SetSpans is again called synchronously

No, it's still called *asynchronously*. 

> So why do you also need the change to set it in native ?

It is the reason why I moved `setInsets` from Java into native implementation.

Without moving SetInsets into the native, both `_SetSpans` and `_SetInsets` need to blocking calls that follow each other. In my opinion, it's not a good approach. I elaborated on it in [my reply to Harshitha's comment](https://github.com/openjdk/jdk/pull/14478#issuecomment-1593817579).

 
> The difference I can see is that although it is synchronous it now requires jumping on to the Toolkit thread and so other pending processing on the Toolkit thread will run first.
> 
> But I'm not sure that's relevant here.
> 
> So can you expand on why you also had to move the call to setInsets() to native ?

To avoid making both `_SetSpans` and `_SetInsets` synchronous calls;
To avoid another JNI call when both actions could be easily coalesced.

`SetInsets` must always be called after `SetSpans`. Why do we need to call both from Java side, where native implementation for `SetSpans` can as easily call it directly?

honkar-jdk:
> > @aivanov-jdk If just the `_SetSpans()` on native is changed to run as `InvokeFunction` instead of `InvokeFunctionLater` both the tests - ScrollPaneLeakTest.java & ScrollPaneExtraScrollBar.java pass without additional code changes to WScrollPanePeer.java.childResized and _setInsets.
> 
> I know it. But I think it's bad. If both `_SetSpans` and `_SetInsets` are synchronous by using `InvokeFunction`, you block EDT to run some code on the toolkit thread twice:
> 
> 1. EDT calls `_SetSpans`, it gets blocked until `SetSpans` completes on the toolkit thread;
> 2. EDT calls `_SetInsets`, it gets blocked until `SetInsets` completes on the toolkit thread.
> 
> I think blocking is bad. I coalesced these two calls into one asynchronous call. There's no EDT blocking with the same result.

Thank you for the details. I'll play around with the 2 tests, trace out the calls and check if we are hitting any other corner cases.

aivanov-jdk:
Following the same way of thought, I'd like to coalesce `setInsets`, `getInsets` followed by `setScrollPosition` in `WScrollPanePeer.initialize` into one native call as well. I believe this would be more efficient. I've already submitted [JDK-8310056](https://bugs.openjdk.org/browse/JDK-8310056) for it.

That change is related to this but I'd rather keep them separate: it's safer to make small changes. Moreover they're independent.

prrace:
> Now with InvokeFunction, _SetSpans is again called synchronously
No, it's still called *asynchronously*. 


=== 

  INLINE void * InvokeFunction(void*(*ftn)(void)) {
        return (void *)SendMessage(WM_AWT_INVOKE_VOID_METHOD, (WPARAM)ftn, 0);

https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-sendmessage

Sends the specified message to a window or windows.
The SendMessage function calls the window procedure for the specified window and does not return until the window procedure has processed the message.

===

So how is that not synchronous ?

honkar-jdk:
> Oh wait, that is _SetInsets you updated here. I thought I had these calls straight but I need to look again.

@prrace That is correct. It is_SetInsets which has been changed to be called as `InvokeFunction` and _SetSpans is still being called as `InvokeFunctionLater` as per @aivanov-jdk previous fix.

prrace (approved):
I think I got it now.

aivanov-jdk:
> So I think I see with the benefit of your recent comments.
> It seemed that reverting _SetSpans to synchronous would fix this and somehow I read that was what you were doing.
> But you didn't want to do that and you were forced into making _SetInsets synchronous because of the initalization usage
> and hence moving the update of insets needed by setSpans avoided that.

Yes, that's right.

honkar-jdk (approved):
PR changes look good. All ScrollPane related tests work as expected with the changes.

src/java.desktop/windows/native/libawt/windows/awt_ScrollPane.cpp
```diff
@@ -525,6 +525,7 @@ void AwtScrollPane::_SetSpans(void *param)
             parentWidth, parentHeight, childWidth, childHeight);
         s->RecalcSizes(parentWidth, parentHeight, childWidth, childHeight);
         s->VerifyState();
+        s->SetInsets(env);
```
mrserb, line 528: I guess we should call "s->VerifyState();" after the "s->SetInsets(env);" similar to what we do in the "AwtScrollPane::_SetInsets"

aivanov-jdk: > I guess we should call "s->VerifyState();" after the "s->SetInsets(env);" similar to what we do in the "AwtScrollPane::_SetInsets"

@mrserb An interesting question… You're right, I've introduced an inconsistency.

Yet I don't think it matters because `SetInsets` doesn't change any members of `AwtScrollPane` object which are asserted in `VerifyState`. In addition to it, `VerifyState` is a no-op in release builds.

I'll take this into account in [JDK-8310056](https://bugs.openjdk.org/browse/JDK-8310056).

PR #14479, opened 2023-06-14
8310033: Clarify return value of Java Time compareTo methods

RogerRiggs:
In java.time packages, clarify timeline order javadoc to mention "before" and "after" in the value of the `compareTo` method return values. 
Add javadoc @see tags to isBefore and isAfter methods

Replace use of "negative" and positive with "less than zero" and "greater than zero" in javadoc \@return
The term "positive" is ambiguous, zero is considered positive and indicates equality.

---------
### Progress
- [x] Change requires CSR request [JDK-8310064](https://bugs.openjdk.org/browse/JDK-8310064) to be approved

### Issues
 * [JDK-8310033](https://bugs.openjdk.org/browse/JDK-8310033): Clarify return value of Java Time compareTo methods (**Bug** - P4)
 * [JDK-8310064](https://bugs.openjdk.org/browse/JDK-8310064): Clarify Java Time documentation of the return value of compareTo methods (**CSR**)

### Reviewers
 * [Brian Burkhalter](https://openjdk.org/census#bpb) (@bplb - **Reviewer**) ⚠️ Review applies to [cd997ede](https://git.openjdk.org/jdk/pull/14479/files/cd997edefc69de9e039f84b324dd5ccbdb3265a2)
 * [Stephen Colebourne](https://openjdk.org/census#scolebourne) (@jodastephen - Author)
 * [Pavel Rappo](https://openjdk.org/census#prappo) (@pavelrappo - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14479/head:pull/14479` \
`$ git checkout pull/14479`

`$ git checkout pull/14479` \
`$ git pull https://git.openjdk.org/jdk.git pull/14479/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14479`

`$ git pr show -t 14479`

</details>

<a href="https://git.openjdk.org/jdk/pull/14479.diff">https://git.openjdk.org/jdk/pull/14479.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14479#issuecomment-1595190736)


bplb (approved):
Looks good.

src/java.base/share/classes/java/time/Duration.java
```diff
@@ -1419,7 +1419,8 @@ public Duration truncatedTo(TemporalUnit unit) {
      * It is "consistent with equals", as defined by {@link Comparable}.
      *
      * @param otherDuration the other duration to compare to, not null
-     * @return the comparator value, negative if less, positive if greater
+     * @return the comparator value is less than zero if the other duration is before,
```
LanceAndersen, line 1422: General comment.  I am not sure that "other duration" is quite clear.  It seems that javadoc for many `compareTo` methods have not standardized on wording for describing the argument being checked.   Some include "argument" when the documentation.    Perhaps change "other duration" -> "duration argument"?

I think LocalDate::compareTo and its use of "other" is perhaps a better  example of why we might want to beef the wording up a bit

RogerRiggs: The argument name is `otherDuration` would putting the argument name in {@code} font put sufficient focus on the argument?

LanceAndersen: I think that would be fine as well as then it is clear we are referring to the argument.

bplb (approved):
Using `{@code}` looks good.

pavelrappo:
While the new wording is clearly an improvement, this reads weird:

> @return the comparator value is less...

For readability, I'd use _which_ or _that_, depending on your preferred flavo ... flavour of English.
```
@return the comparator value that is less...
```

jensli:
>  The term "positive" is ambiguous, zero is considered positive and indicates equality.

Where did you get this idea? A "positive" means a number that is _greater_ than zero.

See both these dictionaries:

https://dictionary.cambridge.org/dictionary/english/positive
https://www.merriam-webster.com/dictionary/positive

jensli:
> While the new wording is clearly an improvement, this reads weird:
> 
> > @return the comparator value is less...
> 

I suggest you use the safe form as the old text, which used a comma:

> ```
> @return the comparator value, that is negative...
> ```

src/java.base/share/classes/java/time/Duration.java
```diff
@@ -1419,7 +1419,8 @@ public Duration truncatedTo(TemporalUnit unit) {
      * It is "consistent with equals", as defined by {@link Comparable}.
      *
      * @param otherDuration the other duration to compare to, not null
-     * @return the comparator value, negative if less, positive if greater
+     * @return the comparator value is less than zero if the {@code otherDuration} is before,
```
jodastephen, line 1422: There is no concept of before and after when talking about a `Duration`

src/java.base/share/classes/java/time/Instant.java
```diff
@@ -1272,8 +1272,11 @@ public long toEpochMilli() {
      * It is "consistent with equals", as defined by {@link Comparable}.
      *
      * @param otherInstant  the other instant to compare to, not null
-     * @return the comparator value, negative if less, positive if greater
+     * @return the comparator value is less than zero if the {@code otherInstant} is before,
+     *          zero if they are equal, greater than zero if the {@code otherInstant} is after
```
jodastephen, line 1276:
```suggestion
     * @return the comparator value - less than zero if {@code otherInstant} is before this instant,
     *          zero if they are equal, greater than zero if {@code otherInstant} is after this instant
```

Saying `the {@code otherInstant}` doesn't read correctly in text.

`the comparator value` needs punctuation after it in order to make sense

src/java.base/share/classes/java/time/chrono/AbstractChronology.java
```diff
@@ -656,7 +656,8 @@ void addFieldValue(Map<TemporalField, Long> fieldValues, ChronoField field, long
      * Subclasses must compare any additional state that they store.
      *
      * @param other  the other chronology to compare to, not null
-     * @return the comparator value, negative if less, positive if greater
+     * @return the comparator value is less than zero if the {@code other} ID string is before,
```
jodastephen, line 659: A `Chronology` is compared by ID, not before/after

src/java.base/share/classes/java/time/ZoneOffset.java
```diff
@@ -714,7 +714,9 @@ public Temporal adjustInto(Temporal temporal) {
      * The comparison is "consistent with equals", as defined by {@link Comparable}.
      *
      * @param other  the other date to compare to, not null
-     * @return the comparator value, negative if less, positive if greater
+     * @return the comparator value is less than zero if the {@code other} is before,
```
jodastephen, line 717: It is tricky to describe an offset as before or after. If you are going to change this you will need a much better description

src/java.base/share/classes/java/time/OffsetTime.java
```diff
@@ -1281,7 +1281,11 @@ public long toEpochSecond(LocalDate date) {
      * use {@link ChronoField#NANO_OF_DAY} as a comparator.
      *
      * @param other  the other time to compare to, not null
-     * @return the comparator value, negative if less, positive if greater
+     * @return the comparator value is less than zero if the {@code other} is before,
```
jodastephen, line 1284: Using before/after here is very wrong. As described in the text above, two instances can be at the same point in time, yet still have a defined sort order.

src/java.base/share/classes/java/time/OffsetDateTime.java
```diff
@@ -1801,7 +1802,11 @@ public long toEpochSecond() {
      * consistent with {@code equals()}.
      *
      * @param other  the other date-time to compare to, not null
-     * @return the comparator value, negative if less, positive if greater
+     * @return the comparator value is less than zero if the {@code other} is before,
```
jodastephen, line 1805: As per `OffsetTime` this definition is wrong. The comparison order is not before/after.

src/java.base/share/classes/java/time/MonthDay.java
```diff
@@ -675,7 +675,11 @@ public LocalDate atYear(int year) {
      * It is "consistent with equals", as defined by {@link Comparable}.
      *
      * @param other  the other month-day to compare to, not null
-     * @return the comparator value, negative if less, positive if greater
+     * @return the comparator value is less than zero if the {@code other} is before,
```
jodastephen, line 678: Using before/after here could be confusing, as January could be considered to be before or after July (since the year is not defined).

src/java.base/share/classes/java/time/chrono/ChronoZonedDateTime.java
```diff
@@ -569,7 +569,11 @@ default long toEpochSecond() {
      * This default implementation performs the comparison defined above.
      *
      * @param other  the other date-time to compare to, not null
-     * @return the comparator value, negative if less, positive if greater
+     * @return the comparator value is less than zero if the {@code other} is before,
```
jodastephen, line 572: The comparison is not before/after, as the ID is taken into account

src/java.base/share/classes/java/time/chrono/Chronology.java
```diff
@@ -807,7 +807,8 @@ default boolean isIsoBased() {
      * It is "consistent with equals", as defined by {@link Comparable}.
      *
      * @param other  the other chronology to compare to, not null
-     * @return the comparator value, negative if less, positive if greater
+     * @return the comparator value is less than zero if the {@code other} ID string is before,
```
jodastephen, line 810: Not before/after

src/java.base/share/classes/java/time/zone/ZoneOffsetTransition.java
```diff
@@ -400,7 +400,8 @@ List<ZoneOffset> getValidOffsets() {
      * The offsets are ignored, making this order inconsistent with equals.
      *
      * @param transition  the transition to compare to, not null
-     * @return the comparator value, negative if less, positive if greater
+     * @return the comparator value is less than zero if the {@code transition} is before,
```
jodastephen, line 403: Not before/after

jodastephen (changes requested):
As things stand, this PR makes things worse not better I'm afraid.

RogerRiggs:
> Where did you get this idea? A "positive" means a number that is _greater_ than zero.

One of my colleagues with a strong math background has corrected many of my API javadoc comments seeking to avoid any ambiguity.  See reference below about positive integers.

[home](https://www.math.net/) / [primary math](https://www.math.net/primary-math) / [number](https://www.math.net/number) / positive numbers
Positive numbers

A positive number is any number that is greater than 0. **Unlike positive integers**, which include 0 and the [natural numbers](https://www.math.net/natural-numbers), positive numbers include [fractions](https://www.math.net/fraction), [decimals](https://www.math.net/decimal), and other types of [numerals](https://www.math.net/numerals).

jensli:
> A positive number is any number that is greater than 0. **Unlike positive integers**, which include 0

math.net seems pretty alone in this. I find the notion bizarre. There seems to be an overwhelming majority of sources that consider 0 to be neither negative nor positive.

src/java.base/share/classes/java/time/MonthDay.java
RogerRiggs: The `isBefore` and `isAfter` methods use that terminology and Month is an enum with defined order January thru December and use the `compareTo` method to compute before/after.

jddarcy:
> > A positive number is any number that is greater than 0. **Unlike positive integers**, which include 0
> 
> math.net seems pretty alone in this. I find the notion bizarre. There seems to be an overwhelming majority of sources that consider 0 to be neither negative nor positive.

Yes, all the math conventions I'm aware of is that integer zero is neither positive nor negative. For example, the signum method returns 0 for 0, -1 for negative values, and 1 for positive values.

rgiulietti:
Most modern authors in math topics use positive to mean > 0, negative to mean < 0, and 0 is neither.

In some contexts, like in the IEEE 754 standard for floating-point numbers, there are both distinguishable positive and a negative binary zeros. There's no "unsigned" or "agnostic" zero there.

The natural numbers include 0 in foundational topics like set theory (where 0 is _defined_ as the empty set), and exclude it in others.

BTW, [math.net](https://www.math.net/positive-numbers) contradicts itself later on the same page. In the "Is 0 positive or negative?" section it states "Positive numbers are defined as any number that is greater than zero".

jddarcy:
> 

Adding a bit more context, floating-point arithmetic is an approximation to real arithmetic. Real arithmetic has a single zero value. The signed zeros of IEEE 754 floating-point are a computational artifact.

More specifically, IEEE 754 floating-point is an approximation to the real numbers extended with infinite values. There are two ways to extend the reals with infinity: affine (signed infinities) and projective (one unsigned infinity). Early drafts of the IEEE 754 standard supported both affine and project infinity, as did the original 8087 floating-point which predated IEEE 754. The final IEEE 754 only supports signed infinities and because there are two infinities, there are two distinguished (but numerically equal) zero values.

The floating-point signum methods in the Java math library return a same-signed zero for a zero input; +1.0 is *not* returned for +0.0 and -1.0 is *not* returned for -0.0. While a zero value has a sign bit in its representation, neither zero value is "positive" or "negative" in the mathematical sense of the term.

RogerRiggs:
@jodastephen Please review updated descriptions.  Thanks

pavelrappo:
> Adding a bit more context, floating-point arithmetic is an approximation to real arithmetic. Real arithmetic has a single zero value. The signed zeros of IEEE 754 floating-point are a computational artifact.
> 
> More specifically, IEEE 754 floating-point is an approximation to the real numbers extended with infinite values. There are two ways to extend the reals with infinity: affine (signed infinities) and projective (one unsigned infinity). Early drafts of the IEEE 754 standard supported both affine and project infinity, as did the original 8087 floating-point which predated IEEE 754. The final IEEE 754 only supports signed infinities and because there are two infinities, there are two distinguished (but numerically equal) zero values.
> 
> The floating-point signum methods in the Java math library return a same-signed zero for a zero input; +1.0 is _not_ returned for +0.0 and -1.0 is _not_ returned for -0.0. While a zero value has a sign bit in its representation, neither zero value is "positive" or "negative" in the mathematical sense of the term.



> > Where did you get this idea? A "positive" means a number that is _greater_ than zero.
> 
> One of my colleagues with a strong math background has corrected many of my API javadoc comments seeking to avoid any ambiguity. See reference below about positive integers.
> 
> [home](https://www.math.net/) / [primary math](https://www.math.net/primary-math) / [number](https://www.math.net/number) / positive numbers Positive numbers
> 
> A positive number is any number that is greater than 0. **Unlike positive integers**, which include 0 and the [natural numbers](https://www.math.net/natural-numbers), positive numbers include [fractions](https://www.math.net/fraction), [decimals](https://www.math.net/decimal), and other types of [numerals](https://www.math.net/numerals).

Late to the party. FWIW, the only way I can think of 0 as of a positive integer is when it is represented using fixed-width two's complement (e.g. Java's byte, short, int, and long). In that case, the sign bit for 0 is cleared, just like that of a positive integer and unlike that of a negative integer, assuming the same representation.

For example, the following expression will tell you if an `int i` is non-negative, i.e. zero or positive:

    i >>> 31 == 0

pavelrappo:
> math ba



> > Where did you get this idea? A "positive" means a number that is _greater_ than zero.
> 
> One of my colleagues with a strong math background has corrected many of my API javadoc comments seeking to avoid any ambiguity. See reference below about positive integers.
> 
> [home](https://www.math.net/) / [primary math](https://www.math.net/primary-math) / [number](https://www.math.net/number) / positive numbers Positive numbers
> 
> A positive number is any number that is greater than 0. **Unlike positive integers**, which include 0 and the [natural numbers](https://www.math.net/natural-numbers), positive numbers include [fractions](https://www.math.net/fraction), [decimals](https://www.math.net/decimal), and other types of [numerals](https://www.math.net/numerals).

Late to the party. FWIW, the only way I can think of 0 as of a positive integer is when it is represented using fixed-width two's complement (e.g. Java's byte, short, int, and long). In that case, the sign bit for 0 is cleared, just like that of a positive integer and unlike that of a negative integer, assuming the same representation.

For example, the following expression will tell you if an `int i` is non-negative, i.e. zero or positive:

    i >>> 31 == 0

pavelrappo:
> @jodastephen Please review updated descriptions. Thanks

I'm not @jodastephen, obviously, but we both suggested doing something with punctuation/grammar as this reads funny/peculiar:

    @return the comparator value is less than zero

Maybe our suggestions got lost in this huge discussion, or you've decided to address them later.

RogerRiggs:
Please re-review; its doc-only so can be still included in JDK 21.

src/java.base/share/classes/java/time/LocalDateTime.java
```diff
@@ -1807,7 +1807,10 @@ public ZonedDateTime atZone(ZoneId zone) {
      * chronology is also considered, see {@link ChronoLocalDateTime#compareTo}.
      *
      * @param other  the other date-time to compare to, not null
-     * @return the comparator value, negative if less, positive if greater
+     * @return the comparator value, that is less than zero if this is before {@code other},
```
jodastephen, line 1810: This text is incorrect. It does not match the description above re chronologies.

src/java.base/share/classes/java/time/LocalDate.java
```diff
@@ -1988,7 +1988,10 @@ public long toEpochSecond(LocalTime time, ZoneOffset offset) {
      * chronology is also considered, see {@link java.time.chrono.ChronoLocalDate#compareTo}.
      *
      * @param other  the other date to compare to, not null
-     * @return the comparator value, negative if less, positive if greater
+     * @return the comparator value, that is less than zero if this is before {@code other},
```
jodastephen, line 1991: This text is incorrect. It does not match the description above re chronologies.

src/java.base/share/classes/java/time/OffsetDateTime.java
```diff
@@ -1801,7 +1802,10 @@ public long toEpochSecond() {
      * consistent with {@code equals()}.
      *
      * @param other  the other date-time to compare to, not null
-     * @return the comparator value, negative if less, positive if greater
+     * @return the comparator value, that is the comparison with the {@code other}'s instant, if they are not equal;
```
jodastephen, line 1805: This text is incorrect. It does not match the description above re offsets. (This would be a suitable description for `compareInstant`)

src/java.base/share/classes/java/time/OffsetTime.java
```diff
@@ -1281,7 +1281,11 @@ public long toEpochSecond(LocalDate date) {
      * use {@link ChronoField#NANO_OF_DAY} as a comparator.
      *
      * @param other  the other time to compare to, not null
-     * @return the comparator value, negative if less, positive if greater
+     * @return the comparator value, that is the comparison of the UTC equivalent {@code other} instant,
+     *          if they are not equal, and if the UTC equivalent {@code other} instant is equal,
+     *          the comparison of this local date-time with {@code other} local date-time
```
jodastephen, line 1286: This text is incorrect. It compares the local time, not the local date-time, when  the instants are equal

src/java.base/share/classes/java/time/chrono/ChronoLocalDate.java
```diff
@@ -700,7 +700,10 @@ default long toEpochDay() {
      * This default implementation performs the comparison defined above.
      *
      * @param other  the other date to compare to, not null
-     * @return the comparator value, negative if less, positive if greater
+     * @return the comparator value, that is less than zero if this is before {@code other},
```
jodastephen, line 703: This text is incorrect. It does not match the description above re chronologies.

jodastephen (changes requested):

src/java.base/share/classes/java/time/OffsetDateTime.java
RogerRiggs: Line 1806 describes the comparison of the local time if the instants are equal.
```and if equal to the {@code other}'s instant, the comparison of the {@code other}'s local date-time```

jodastephen (approved):
LGTM

RogerRiggs:
Thank you for your careful review.

pavelrappo (approved):
Looks good to me.

naotoj (approved):
+1

PR #14481, opened 2023-06-15
8310076: Reduce inclusion of bytecodeStream.hpp

iklam:
The popular header file generateOopMap.hpp includes bytecodeStream.hpp, but it doesn't actually needs the full declaration of the BytecodeStream class.

This inclusion can be replaced with a simple forward declaration of the BytecodeStream class.

This reduces the number of HotSpot .o files that include bytecodeStream.hpp from 571 to 21.

### DIFF
--- a/src/hotspot/share/interpreter/oopMapCache.cpp
+++ b/src/hotspot/share/interpreter/oopMapCache.cpp
@@ -23,6 +23,7 @@
  */
 
 #include "precompiled.hpp"
+#include "interpreter/bytecodeStream.hpp"
 #include "interpreter/oopMapCache.hpp"
 #include "logging/log.hpp"
 #include "logging/logStream.hpp"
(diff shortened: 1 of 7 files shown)


dholmes-ora (approved):
Seems fine. Thanks.

coleenp (approved):
Looks good but why would generateOopMap.hpp be a popular header file?

iklam:
> Looks good but why would generateOopMap.hpp be a popular header file?

It's included via frame_x86.inline.hpp -> oopMapCache.hpp -> generateOopMap.hpp.

None of these includes seem necessary. I'll try to fix that in a separate issue. See [JDK-8310225](https://bugs.openjdk.org/browse/JDK-8310225)

iklam:
Thanks @coleenp and @dholmes-ora for the review.

Passed tier 1, tier 2, and builds-tier5

/integrate

PR #14484, opened 2023-06-15
JDK-8310107: os::trace_page_sizes_for_requested_size should name alignment as requested page size

tstuefe:
Somewhat trivial renaming patch.

`os::trace_page_sizes_for_requested_size()` uses "alignment" to name what is actually the requested page size (which may differ from the actual page size of a region, which is traced independently). I believe the naming stems from ReservedSpace and os::reserve_...() misusing "alignment" for "page size", but that would be a bigger RFE to clean up. One step at the time.

As a first step, I propose to make naming for os::trace_page_sizes_for_requested_size() clearer. 

Patch renames "alignment" to "requested pagesize" in both printout and code.

It also reshuffles the parameters somewhat to have requested properties at the beginning, followed by the actual properties of the region.

Printout before, for a failed 2M paged allocation (notice the difference between requested and actual page size, which is here named "alignment"):

```
OpenJDK 64-Bit Server VM warning: Failed to reserve and commit memory using large pages. req_addr: 0x0000000000000000 bytes: 2097152
[0.012s][info][pagesize] Block Offset Table: req_size=2M base=0x00007f3518200000 page_size=4K alignment=2M size=2M
```

Printout now:

```
OpenJDK 64-Bit Server VM warning: Failed to reserve and commit memory using large pages. req_addr: 0x0000000000000000 bytes: 2097152
[0.011s][info][pagesize] Block Offset Table: req_size=2M req_page_size=2M base=0x00007fda0e400000 size=2M page_size=4K
```

### DIFF
--- a/src/hotspot/share/runtime/os.hpp
+++ b/src/hotspot/share/runtime/os.hpp
@@ -414,10 +414,10 @@ class os: AllStatic {
                                const size_t size);
   static void trace_page_sizes_for_requested_size(const char* str,
                                                   const size_t requested_size,
-                                                  const size_t page_size,
-                                                  const size_t alignment,
+                                                  const size_t requested_page_size,
                                                   const char* base,
-                                                  const size_t size);
+                                                  const size_t size,
+                                                  const size_t page_size);
 
   static size_t vm_allocation_granularity() { return OSInfo::vm_allocation_granularity(); }
 
--- a/src/hotspot/share/gc/g1/g1CollectedHeap.cpp
+++ b/src/hotspot/share/gc/g1/g1CollectedHeap.cpp
@@ -1307,10 +1307,10 @@ G1RegionToSpaceMapper* G1CollectedHeap::create_aux_memory_mapper(const char* des
 
   os::trace_page_sizes_for_requested_size(description,
                                           size,
-                                          page_size,
                                           preferred_page_size,
                                           rs.base(),
-                                          rs.size());
+                                          rs.size(),
+                                          page_size);
 
   return result;
 }
--- a/src/hotspot/share/runtime/os.cpp
+++ b/src/hotspot/share/runtime/os.cpp
@@ -1675,23 +1675,23 @@ void os::trace_page_sizes(const char* str,
 
 void os::trace_page_sizes_for_requested_size(const char* str,
                                              const size_t requested_size,
-                                             const size_t page_size,
-                                             const size_t alignment,
+                                             const size_t requested_page_size,
                                              const char* base,
-                                             const size_t size) {
+                                             const size_t size,
+                                             const size_t page_size) {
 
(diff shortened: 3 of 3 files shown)


tstuefe:
x86 error unrelated.

stefank (commented):
I think this looks reasonable and we should also change `os::trace_page_sizes` to have the same order when printing `base`, `size`, and `page_size`.

tstuefe:
> I think this looks reasonable and we should also change `os::trace_page_sizes` to have the same order when printing `base`, `size`, and `page_size`.

Thank you, @stefank . New version changes printing- and argument order of  `os::trace_page_sizes` to equal `os::trace_page_sizes_for_requested_size()`.

While checking the affected tests, I noticed that gc/g1/TestLargePageUseForAuxMemory.java is broken when running on machines with 1G pages configured. Opened a separate issue https://bugs.openjdk.org/browse/JDK-8310216 to track that.

src/hotspot/share/runtime/os.hpp
```diff
@@ -405,19 +405,19 @@ class os: AllStatic {
   // passed to page_size_for_region() and page_size should be the result of that
   // call.  The (optional) base and size parameters should come from the
   // ReservedSpace base() and size() methods.
-  static void trace_page_sizes(const char* str, const size_t* page_sizes, int count);
+//  static void trace_page_sizes(const char* str, const size_t* page_sizes, int count);
```
stefank, line 408: Should this be removed?

stefank (approved):
Looks good. One nit:

tstuefe:
> Looks good. One nit:

Thank you! 

I removed a leftover prototype that had no caller nor implementation. Also a last test fix.

stefank (approved):

test/hotspot/jtreg/gc/g1/TestLargePageUseForAuxMemory.java
```diff
@@ -126,6 +126,8 @@ static void testVM(String what, long heapsize, boolean cardsShouldUseLargePages,
         pb = ProcessTools.createJavaProcessBuilder(getOpts(heapsize, true));
 
         OutputAnalyzer output = new OutputAnalyzer(pb.start());
+        output.reportDiagnosticSummary();
```
dholmes-ora, line 129: This should come last as if any `output.shouldxxx()` fails then the summary will be printed anyway, so we don't need to see it twice.

dholmes-ora (approved):
Functional changes seem fine. I don't know the tests so can't comment on those other than the seemingly unrelated change to print the summary.

Thanks.

tstuefe:
> Functional changes seem fine. I don't know the tests so can't comment on those other than the seemingly unrelated change to print the summary.
> 
> Thanks.

Thank you David. I removed the debug output.

PR #14486, opened 2023-06-15
8310110: Shenandoah: Trace page sizes

tstuefe:
`-Xlog:pagesize` now produced this printout

```
[0.006s][info][pagesize] Heap: req_size=5G base=0x00000006c0000000 page_size=1G alignment=1G size=5G
[0.006s][info][pagesize] Mark Bitmap: req_size=160M base=0x00007fc4c0000000 page_size=1G alignment=1G size=1G
[0.011s][info][pagesize] Aux Bitmap: req_size=160M base=0x00007fc4c0000000 page_size=1G alignment=1G size=1G
[0.011s][info][pagesize] Region Storage: req_size=320K base=0x00007fc3c0000000 page_size=1G alignment=1G size=1G
[0.011s][info][pagesize] Collection Set: req_size=16K base=0x0000000000010000 page_size=4K alignment=4K size=16K
```

Note that this printout is from a machine with 1G pages and -XX:+UseLargePages; it nicely demonstrates how much memory we waste in that case on support structures; https://bugs.openjdk.org/browse/JDK-8310111 tracks that.

Also note that I reuse `os::trace_page_sizes_for_requested_size` here, which I think can be improved and made clearer (alignment is actually requested pagesize). This is tracked separately in https://github.com/openjdk/jdk/pull/14484 .

---------
### Progress

### Issue
 * [JDK-8310110](https://bugs.openjdk.org/browse/JDK-8310110): Shenandoah: Trace page sizes (**Enhancement** - P4)

### Reviewers
 * [Roman Kennke](https://openjdk.org/census#rkennke) (@rkennke - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14486/head:pull/14486` \
`$ git checkout pull/14486`

`$ git checkout pull/14486` \
`$ git pull https://git.openjdk.org/jdk.git pull/14486/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14486`

`$ git pr show -t 14486`

</details>

<a href="https://git.openjdk.org/jdk/pull/14486.diff">https://git.openjdk.org/jdk/pull/14486.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14486#issuecomment-1594212310)


shipilev (commented):
The idea looks good. 

I would prefer not to have a one-liner helper method, and instead inline the uses like this:

```
   if (ShenandoahVerify) {
     ReservedSpace verify_bitmap(_bitmap_size, bitmap_page_size);
+    os::trace_page_sizes_for_requested_size("Verify Bitmap", bitmap_size_orig,
+                                            bitmap.page_size(), bitmap_page_size,
+                                            bitmap.base(), bitmap.size());
     if (!verify_bitmap.special()) {
```

Also, let's wait for #14484 to land first?

shipilev:
Oh, also, we try to keep the synopsis format like this: "8310110: Shenandoah: Trace page sizes"

tstuefe:
> The idea looks good.
> 
> I would prefer not to have a one-liner helper method, and instead inline the uses like this:
> 
> ```
>    if (ShenandoahVerify) {
>      ReservedSpace verify_bitmap(_bitmap_size, bitmap_page_size);
> +    os::trace_page_sizes_for_requested_size("Verify Bitmap", bitmap_size_orig,
> +                                            bitmap.page_size(), bitmap_page_size,
> +                                            bitmap.base(), bitmap.size());
>      if (!verify_bitmap.special()) {
> ```

Okay.

> 
> Also, let's wait for #14484 to land first?

Certainly. Mind reviewing #14484 ? It's very simple.

tstuefe:
@shipilev Thanks for reviewing. Here you go.

shipilev (approved):
Looks okay to me!

I would prefer the style to be:

```
  os::trace_page_sizes_for_requested_size("Mark Bitmap",
                                          bitmap_size_orig, bitmap_page_size,
                                          bitmap.base(),
                                          bitmap.size(), bitmap.page_size());
```

...so that we have:

```
line 1: label
line 2: original size, page size
line 3: base addr
line 4: actual size, page size
```

Would be easier to check lines 2 and 4 match?

shipilev (approved):
Looks good, thanks!

Please make sure Windows builds/tests pass -- I think GHA in current master should be good. I see you merged recently, so we might just wait for GHA to complete.

shipilev:
> The newly added Shenandoah test uncovered an old bug where the Aux bitmap would not be madvised correctly for use with THP. Tracked by https://bugs.openjdk.org/browse/JDK-8310388, and I disable the test for now.

OK, assuming we get the consensus around JDK-8310388 soon, we can wait for that to land before this test? :)

shipilev:
> > > The newly added Shenandoah test uncovered an old bug where the Aux bitmap would not be madvised correctly for use with THP. Tracked by https://bugs.openjdk.org/browse/JDK-8310388, and I disable the test for now.
> > 
> > OK, assuming we get the consensus around JDK-8310388 soon, we can wait for that to land before this test? :)
> 
> The test still makes sense for explicit huge pages, or?

Yes, it does. I just want to minimize churn and simplify backports :)

tstuefe:
> > The test still makes sense for explicit huge pages, or?
> 
> Yes, it does. I just want to minimize churn and simplify backports :) So fewer issues to backport with fewer dependencies would be better.

I'll remove the test from this PR.

shipilev (approved):
All right, this works too.

tstuefe:
Ping... may I have a second review? This is simple stuff.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -293,27 +314,35 @@ jint ShenandoahHeap::initialize() {
...
 
     for (uintptr_t addr = min; addr <= max; addr <<= 1u) {
       char* req_addr = (char*)addr;
       assert(is_aligned(req_addr, cset_align), "Should be aligned");
-      ReservedSpace cset_rs(cset_size, cset_align, os::vm_page_size(), req_addr);
+      ReservedSpace cset_rs(cset_size, cset_align, cset_page_size, req_addr);
       if (cset_rs.is_reserved()) {
         assert(cset_rs.base() == req_addr, "Allocated where requested: " PTR_FORMAT ", " PTR_FORMAT, p2i(cset_rs.base()), addr);
         _collection_set = new ShenandoahCollectionSet(this, cset_rs, sh_rs.base());
+        rs = &cset_rs;
```
rkennke, line 332: I'm probably missing something here, but doesn't cset_rs go out of scope, and rs would point to undefined stuff?

tstuefe: :-( you are of course right. This only worked out of accident since ReservedSpace has no dtor.

rkennke: > :-( you are of course right. This only worked out of accident since ReservedSpace has no dtor.

Could you copy-assign the object, instead?

shipilev: Or just null-check the resulting `_collection_set` and print the message then. Don't need to mess with locals then?

tstuefe: Okay, fixed. Copying ReservedSpace is fine, we do it in other places too. This thing is used like a value object in most places.

shipilev (approved):

rkennke (approved):
Looks good now, thanks!

tstuefe:
I verified manually that macos m1 builds.

Thanks @shipilev and @rkennke!

/integrate

PR #14490, opened 2023-06-15
C2: assert(final_con == (jlong)final_int) failed: final value should be integer

enothum:
**Acknowledgments**: Thanks to @quadhier for the preliminary work on this issue: https://github.com/openjdk/jdk/pull/14353

**JDK-8309266**: TestLoopLimitOverflowDuringCCP.java causes an assertion error (overflow check) in LoopLimitNode::Value. To fix the issue I added a check in LoopLimitNode::Value that verifies that the input nodes are ConI type nodes. 

Previously, TestLoopLimitOverflowDuringCCP would cause the assertion error in LoopLimitNode::Value, during PhaseCCP::analyze. The problem originated from PhaseCCP initializing all types to TOP, resulting in the Phi node from `int limit = flag ? 1000 : Integer.MAX_VALUE` being temporarily considered as Integer.MAX_VALUE. This happens as the Node for the Integer.MAX_VALUE case was already analyzed by CCP while the Node for the 1000 case was still initialized to TOP. When resolving the value of the Phi node, Integer.MAX_VALUE and TOP get merged to Integer.MAX_VALUE, which is then processed by LoopLimitNode::Value as a constant, resulting in the integer overflow.

By checking that the input nodes are ConI nodes in LoopLimitNode::Value, we avoid Phi nodes being misinterpreted during PhaseCCP. If the Phi nodes turn out to be constant they should rather be first transformed to ConI nodes.

### DIFF
--- a/src/hotspot/share/opto/loopnode.cpp
+++ b/src/hotspot/share/opto/loopnode.cpp
@@ -2282,11 +2282,13 @@ const Type* LoopLimitNode::Value(PhaseGVN* phase) const {
   if (limit_t  == Type::TOP) return Type::TOP;
   if (stride_t == Type::TOP) return Type::TOP;
 
-  int stride_con = stride_t->is_int()->get_con();
-  if (stride_con == 1)
-    return bottom_type();  // Identity
+  if(in(Init)->is_ConI() && in(Limit)->is_ConI() && in(Stride)->is_ConI()){
+    int stride_con = stride_t->is_int()->get_con();
+    if (stride_con == 1) {
+      return bottom_type();  // Identity
+    }
 
-  if (init_t->is_int()->is_con() && limit_t->is_int()->is_con()) {
+    assert(init_t->is_int()->is_con() && limit_t->is_int()->is_con(), "init_t and limit_t should have constant values");
     // Use jlongs to avoid integer overflow.
     jlong init_con   =  init_t->is_int()->get_con();
     jlong limit_con  = limit_t->is_int()->get_con();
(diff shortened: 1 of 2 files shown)


rwestrel:
Let's say CCP is running. Init and limit are not `ConINode`s at this point but once CCP is over, it will have discovered that they are actually constants so will make them `ConINode`s. The change you're making will have prevented CCP from propagating the constants through `LoopLimitNode` making CCP more pessimistic that it needs to be.

src/hotspot/share/opto/loopnode.cpp
```diff
@@ -2311,8 +2311,15 @@ const Type* LoopLimitNode::Value(PhaseGVN* phase) const {
     int final_int = (int)final_con;
     // The final value should be in integer range since the loop
     // is counted and the limit was checked for overflow.
-    assert(final_con == (jlong)final_int, "final value should be integer");
-    return TypeInt::make(final_int);
+    // Assert checks for overflow only if all input nodes are ConINodes, as during CCP
+    // there might be a temporary overflow from PhiNodes see JDK-8309266
+    assert(in(Init)->is_ConI() && in(Limit)->is_ConI() && in(Stride)->is_ConI() \
```
rwestrel, line 2316: Do we really need a backslash here?

rwestrel (approved):
Looks reasonable to me.

src/hotspot/share/opto/loopnode.cpp
chhagedorn: You can wrap lines in asserts without backslashes.

enothum: Ahh yes. Not sure why, I probably just used the backslash after having read it in some other parts of the code. Sorry for the confusion from my side

enothum:
Thanks @chhagedorn and @rwestrel for the feedback and reviews!

PR #14491, opened 2023-06-15
8241800: Disable IPV6_MULTICAST_ALL to prevent interference from all multicast groups

Michael-Mc-Mahon:
Hi,

Can I get the following change reviewed please? It disables the IPV6_MULTICAST_ALL socket option on UDP sockets on Linux systems which support the option (kernel version 4.20+). This has the effect of disabling the surprising default behavior of multicast sockets on Linux where sockets can receive packets sent to groups the receiver has not joined, if some other socket has joined that group on the same system.

The equivalent behavior has already been implemented for IPv4 sockets.

Thanks,
Michael.

---------
### Progress

### Issue
 * [JDK-8241800](https://bugs.openjdk.org/browse/JDK-8241800): Disable IPV6_MULTICAST_ALL to prevent interference from all multicast groups (**Enhancement** - P3)

### Reviewers
 * [Alan Bateman](https://openjdk.org/census#alanb) (@AlanBateman - **Reviewer**) ⚠️ Review applies to [3e900f1e](https://git.openjdk.org/jdk/pull/14491/files/3e900f1e8f727ab590fd94b7ef1cc5a5f359cc9d)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14491/head:pull/14491` \
`$ git checkout pull/14491`

`$ git checkout pull/14491` \
`$ git pull https://git.openjdk.org/jdk.git pull/14491/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14491`

`$ git pr show -t 14491`

</details>

<a href="https://git.openjdk.org/jdk/pull/14491.diff">https://git.openjdk.org/jdk/pull/14491.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14491#issuecomment-1592817207)


src/java.base/unix/native/libnio/ch/Net.c
```diff
@@ -295,6 +299,16 @@ Java_sun_nio_ch_Net_socket0(JNIEnv *env, jclass cl, jboolean preferIPv6,
...
         }
+        if (level == IPPROTO_IPV6) {
+            if ((setsockopt(fd, level, IPV6_MULTICAST_ALL, (char*)&arg, sizeof(arg)) < 0) &&
+                (errno != ENOPROTOOPT)) {
+                JNU_ThrowByNameWithLastError(env,
+                                         JNU_JAVANETPKG "SocketException",
+                                         "Unable to set IPV6_MULTICAST_ALL");
+                close(fd);
+                return -1;
+            }
```
AlanBateman, line 310: It might be better to move this to just below the attempt to set IPV6_MULTICAST_HOPS as that code only runs for IPv6 datagrams.

That said, I'm a bit concerned that this will fail on older kernels. Are you sure the error is ENOPROTOOPT?

Michael-Mc-Mahon: Yes, the error is the same as for the IPv4 option. If you like I can change test to call the option and check an exception is not thrown in those cases.

Michael-Mc-Mahon:
> 

It's certainly the case that the IPv4 option can be called on AF_INET6 sockets but only affects the IPv4 stack. Which suggests the IPv6 option only affects the IPv6 stack. I'll see if I can confirm that is true.

Michael-Mc-Mahon:
> > 
> 
> It's certainly the case that the IPv4 option can be called on AF_INET6 sockets but only affects the IPv4 stack. Which suggests the IPv6 option only affects the IPv6 stack. I'll see if I can confirm that is true.

I just checked and the two options are independent of each other, applying only to the IPv4 stack and IPv6 stack respectively.

src/java.base/unix/native/libnio/ch/Net.c
```diff
@@ -308,6 +312,17 @@ Java_sun_nio_ch_Net_socket0(JNIEnv *env, jclass cl, jboolean preferIPv6,
...
+        /* Disable IPV6_MULTICAST_ALL if option supported */
+        arg = 0;
+        if ((setsockopt(fd, IPPROTO_IPV6, IPV6_MULTICAST_ALL, (char*)&arg, sizeof(arg)) < 0) &&
+            (errno != ENOPROTOOPT)) {
+            JNU_ThrowByNameWithLastError(env,
+                                     JNU_JAVANETPKG "SocketException",
+                                     "Unable to set IPV6_MULTICAST_ALL");
+            close(fd);
+            return -1;
+        }
```
AlanBateman, line 325: This implementation change looks okay, assuming ENOPROTOOPT on older kernels. I haven't had time to study the test yet, tricky scenario to test reliably.

AlanBateman:
Can you look at test/jdk/java/nio/channels/DatagramChannel/Promiscuous.java? I wonder if this should be updated to exercise the test with two IPv6 multicast groups? In passing, I assume the check for the OS version isn't needed in this test anymore.

As regards the new test then I think the naming should be a bit more consistent with the existing tests if possible. The existing test for IP_MULTICAST_ALL is Promiscuous so maybe the new test is PromiscuousIPv6 or something with "IPv6" in the name?

The test launches "uname -r", maybe it can use the Platform test infra class instead.

The class initializer probes for interfaces that support IPv6 multicasting. Can this be moved to the main method so that it's easier to see in main why the test is skipping?

If < 4.20 then I think the test should just skip on Linux. No need to create a socket + close it. If we have an issue here then all DatagramChannel tests would fail.

The select(2000) might not be enough on slow machines. So if we are keeping this test then I think it will need a few rounds of cleanup to make it more robust.

src/java.base/unix/native/libnio/ch/Net.c
Michael-Mc-Mahon: I am running some repeat 50 test runs of the datagram tests on the newer Linux systems.

Michael-Mc-Mahon:
> Can you look at test/jdk/java/nio/channels/DatagramChannel/Promiscuous.java? I wonder if this should be updated to exercise the test with two IPv6 multicast groups? In passing, I assume the check for the OS version isn't needed in this test anymore.
> 
Okay
> As regards the new test then I think the naming should be a bit more consistent with the existing tests if possible. The existing test for IP_MULTICAST_ALL is Promiscuous so maybe the new test is PromiscuousIPv6 or something with "IPv6" in the name?
> 
Right
> The test launches "uname -r", maybe it can use the Platform test infra class instead.
> 
Ah, I missed that Platform can return the os version numbers. Will change that.
> The class initializer probes for interfaces that support IPv6 multicasting. Can this be moved to the main method so that it's easier to see in main why the test is skipping?
> 
ok
> If < 4.20 then I think the test should just skip on Linux. No need to create a socket + close it. If we have an issue here then all DatagramChannel tests would fail.
> 
ok, I just added that after the previous round. I will take it out.
> The select(2000) might not be enough on slow machines. So if we are keeping this test then I think it will need a few rounds of cleanup to make it more robust.

I'll increase that. I suspect the test doesn't need othervm mode for that matter. Will change that too.

test/jdk/java/nio/channels/DatagramChannel/Promiscuous.java
```diff
@@ -195,21 +196,6 @@ static void test(ProtocolFamily family,
...
-        if (os.equals("Linux")) {
-            String osversion = System.getProperty("os.version");
-            String[] vers = osversion.split("\\.", 0);
-            int major = Integer.parseInt(vers[0]);
-            if (major < 3) {
-                System.out.format("Kernel version is %s, test skipped%n", osversion);
-                return;
-            }
-        }
-
```
AlanBateman, line 212: Version check removed, good!

test/jdk/java/nio/channels/DatagramChannel/Promiscuous.java
```diff
@@ -222,8 +208,8 @@ public static void main(String[] args) throws IOException {
             InetAddress source = config.ip4Addresses(nif).iterator().next();
             test(INET, nif, ip4Group1, ip4Group2);
 
-            // Solaris and Linux allow IPv6 sockets join IPv4 multicast groups
-            if (os.equals("Linux"))
+            // Linux allows IPv6 sockets join IPv4 multicast groups
+            if (Platform.isLinux())
```
AlanBateman, line 212: If you can, it would be useful to double check it is Linux specific, I would expect it should pass on Windows and macOS too.

test/jdk/java/nio/channels/DatagramChannel/PromiscuousIPv6.java
```diff
@@ -145,6 +147,7 @@ static void receiveDatagram(DatagramChannel dc,
 
     static void test(ProtocolFamily family,
                      NetworkInterface nif,
+                     boolean bindToWildCard,
```
AlanBateman, line 150: Wildcard rather than WildCard :-)

AlanBateman:
> So, it turns out there is already an IPv6 version of the Promiscuous test. It only required a small change to this test to verify the behavior with the wildcard address. I updated it, and removed the new test

Good!

Michael-Mc-Mahon:
So, the test Promiscuous.java (the IPv4 version) can run on all platforms now. The IPv6 version (PromiscuousIPv6.java) still does not work on Windows.

test/jdk/java/nio/channels/DatagramChannel/Promiscuous.java
```diff
@@ -43,6 +43,7 @@
 import java.util.stream.Collectors;
 
 import jdk.test.lib.NetworkConfiguration;
+import jdk.test.lib.Platform;
```
AlanBateman, line 46: I assume this isn't needed for the current patch.

test/jdk/java/nio/channels/DatagramChannel/Promiscuous.java
```diff
@@ -222,9 +208,8 @@ public static void main(String[] args) throws IOException {
             InetAddress source = config.ip4Addresses(nif).iterator().next();
             test(INET, nif, ip4Group1, ip4Group2);
 
-            // Solaris and Linux allow IPv6 sockets join IPv4 multicast groups
-            if (os.equals("Linux"))
-                test(UNSPEC, nif, ip4Group1, ip4Group2);
+            // test IPv6 sockets joining IPv4 multicast groups
+            test(UNSPEC, nif, ip4Group1, ip4Group2);
```
AlanBateman, line 212: The SAP and/or IBM folks that maintain the AIX port will probably have to adjust this to skip on that platform.

test/jdk/java/nio/channels/DatagramChannel/PromiscuousIPv6.java
```diff
@@ -196,17 +203,14 @@ static void test(ProtocolFamily family,
...
         } else {
-            String osVersion = System.getProperty("os.version");
-            String prefix = "3.10.0";
-            if (osVersion.startsWith(prefix)) {
-                throw new SkippedException(
-                        String.format("The behavior under test is known NOT to work on '%s' kernels", prefix));
-            }
+            int major = Platform.getOsVersionMajor();
+            int minor = Platform.getOsVersionMinor();
+            hasIPV6MulticastAll = Platform.isOSX() || (major > 4) || ((major == 4 && minor >= 20));
```
AlanBateman, line 213: The version check should be Linux only.

Michael-Mc-Mahon: It runs the additional tests if it is MacOSX or if Linux and kernel > 4.20.

AlanBateman: It looks like it's good the version check on AIX too. I think you'll need to include a check for isLinux to avoid a follow-up bug from the maintainers of that port.

Michael-Mc-Mahon: Oh, I see there is a missing pair of braces. I will fix that

Michael-Mc-Mahon: Ok then. I'll add explicit tests for Linux and Windows, and the AIX folks can add that if/when they verify it works.

test/jdk/java/nio/channels/DatagramChannel/Promiscuous.java
```diff
@@ -192,24 +193,13 @@ static void test(ProtocolFamily family,
         }
     }
 
+    private static boolean supportedByPlatform() {
```
AlanBateman, line 196: Do you mind adding a /**/ comment to this method to say that it returns true if platform allows an IPv6 socket join an IPv4 multicast group?

test/jdk/java/nio/channels/DatagramChannel/PromiscuousIPv6.java
```diff
@@ -194,19 +201,22 @@ static void test(ProtocolFamily family,
         }
     }
 
+    private static boolean supportedByPlatform() {
```
AlanBateman, line 204: This method too, I think it needs a comment to this method to say that it returns true if platform allows an IPv6 socket join an IPv4 multicast group without interference.

PR #14494, opened 2023-06-15
8310130: C2: assert(false) failed: scalar_input is neither phi nor a matchin reduction

eme64:
Removed a spurious assert before optimization bailout.

I assumed that the "scalar input" of a Reduction node must always be either a Phi or another Reduction node. But that is incorrect, partial vectorization can lead to a reduction node chain where we have vector reductions and scalar reductions at the same time. In those cases, we cannot move the UnorderedReductions out of the loop, so a optimization bailout is appropriate.

I assessed the other asserts in `PhaseIdealLoop::move_unordered_reduction_out_of_loop`, and I think they are all justified. However, one assert would have lead to a `continue` in production, which would not break out of the nested loop correctly. I changed it to a `return`, so that would be a bailout from the optimization. This assert should not be triggered because in `SuperWord::mark_reductions` we forbid that a reduction node has any uses inside the loop except for the successor node in the reduction chain.

I have one regression test delivered by the fuzzer, and one that I constructed myself after understanding the issue.
Testing up to tier6 and stress testing.

---------
### Progress

### Issue
 * [JDK-8310130](https://bugs.openjdk.org/browse/JDK-8310130): C2: assert(false) failed: scalar_input is neither phi nor a matchin reduction (**Bug** - P3)

### Reviewers
 * [Christian Hagedorn](https://openjdk.org/census#chagedorn) (@chhagedorn - **Reviewer**) ⚠️ Review applies to [53e8913e](https://git.openjdk.org/jdk/pull/14494/files/53e8913ec04a7a466ef8e0fc68c8d78bdecf80f9)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14494/head:pull/14494` \
`$ git checkout pull/14494`

`$ git checkout pull/14494` \
`$ git pull https://git.openjdk.org/jdk.git pull/14494/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14494`

`$ git pr show -t 14494`

</details>

<a href="https://git.openjdk.org/jdk/pull/14494.diff">https://git.openjdk.org/jdk/pull/14494.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14494#issuecomment-1596778249)

### DIFF
--- a/test/hotspot/jtreg/compiler/loopopts/superword/TestUnorderedReductionPartialVectorization.java
+++ b/test/hotspot/jtreg/compiler/loopopts/superword/TestUnorderedReductionPartialVectorization.java
@@ -61,7 +61,7 @@ public class TestUnorderedReductionPartialVectorization {
     @Test
     @IR(counts = {IRNode.LOAD_VECTOR, "> 0",
                   IRNode.OR_REDUCTION_V, "> 0",},
-        applyIfCPUFeatureOr = {"sse4.1", "true", "asimd", "true"})
+        applyIfCPUFeatureOr = {"sse4.1", "true"})
     static long test1(int[] data, long sum) {
         for (int i = 0; i < data.length; i++) {
             // Mixing int and long ops means we only end up allowing half of the int
--- a/test/hotspot/jtreg/compiler/loopopts/superword/TestUnorderedReductionPartialVectorization.java
+++ b/test/hotspot/jtreg/compiler/loopopts/superword/TestUnorderedReductionPartialVectorization.java
@@ -71,7 +71,7 @@ public class TestUnorderedReductionPartialVectorization {
             // with vectorization. That means we have a mixed scalar/vector reduction
             // chain. This way it is possible that a vector-reduction has a scalar
             // reduction as input, which is neigher a phi nor a vector reduction.
-            // In such a case, we must bail out of the optimization in 
+            // In such a case, we must bail out of the optimization in
             // PhaseIdealLoop::move_unordered_reduction_out_of_loop
             int v = data[i]; // int read
             data[0] = 0;     // ruin the first pack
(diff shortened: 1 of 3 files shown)


vnkozlov (approved):
Looks fine.

test/hotspot/jtreg/compiler/loopopts/superword/TestUnorderedReductionPartialVectorization.java
```diff
@@ -0,0 +1,97 @@
...
+
+import compiler.lib.ir_framework.*;
+
+public class TestUnorderedReductionPartialVectorization {
+    static final int RANGE = 1024;
+    static final int ITER  = 10;
+
+    public static void main(String[] args) {
+        TestFramework.runWithFlags("-Xbatch", "-XX:-TieredCompilation",
+                                   "-XX:CompileCommand=compileonly,compiler.loopopts.superword.TestUnorderedReductionPartialVectorization::test*");
```
chhagedorn, line 43: It should probably also trigger without the flags by just specifying `TestFramework.run()` as `test1()` is not using other methods and the IR framework will implicitly use `-Xbatch` and wait for the compilation of `test1()` to be finished.

chhagedorn (approved):
Looks good!

chhagedorn (approved):

test/hotspot/jtreg/compiler/loopopts/superword/TestUnorderedReductionPartialVectorization.java
chhagedorn: Nice!

eme64:
Just saw a x86 (32bit) IR rule failure for `compiler.loopopts.superword.TestUnorderedReductionPartialVectorization.test1`. Will investigate.

vnkozlov (approved):
Update is good

eme64:
Thanks @vnkozlov @chhagedorn for the reviews!
/integrate

PR #14496, opened 2023-06-15
8310157: Allow void-returning filters for MethodHandles::collectCoordinates

JornVernee:
DRAFT

### DIFF
--- a/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
+++ b/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
@@ -8188,8 +8188,8 @@ assertEquals("boojum", (String) catTrace.invokeExact("boo", "jum"));
      * filter function and the target var handle is then called on the modified (usually shortened)
      * coordinate list.
      * <p>
-     * If {@code R} is the return type of the filter (which cannot be void), the target var handle must accept a value of
-     * type {@code R} as its coordinate in position {@code pos}, preceded and/or followed by
+     * If {@code R} is the return type of the filter, then, unless {@code R} is void, the target var handle must accept
+     * a value of type {@code R} as its coordinate in position {@code pos}, preceded and/or followed by
      * any coordinate not passed to the filter.
      * No coordinates are reordered, and the result returned from the filter
      * replaces (in order) the whole subsequence of coordinates originally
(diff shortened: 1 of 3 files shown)


src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -8188,8 +8188,8 @@ public static VarHandle permuteCoordinates(VarHandle target, List<Class<?>> newC
      * filter function and the target var handle is then called on the modified (usually shortened)
      * coordinate list.
      * <p>
-     * If {@code R} is the return type of the filter (which cannot be void), the target var handle must accept a value of
-     * type {@code R} as its coordinate in position {@code pos}, preceded and/or followed by
+     * If {@code R} is the return type of the filter, then, unless {@code R} is void, the target var handle must accept
```
mcimadamore, line 8191: Not introduced by you - but this para seems to be repeated twice-ish. I think we should try to consolidate the two, if we can.

src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -8198,9 +8198,10 @@ public static VarHandle permuteCoordinates(VarHandle target, List<Class<?>> newC
      * The argument types (if any) of the filter
      * replace zero or one coordinate types of the target var handle, at position {@code pos},
      * in the resulting adapted var handle.
-     * The return type of the filter must be identical to the
+     * If the return type of the filter is <em>not</em> void, it must be identical to the
      * coordinate type of the target var handle at position {@code pos}, and that target var handle
-     * coordinate is supplied by the return value of the filter.
+     * coordinate is supplied by the return value of the filter. If the return type of the filter <em>is</em> void,
```
mcimadamore, line 8203: This is the other para I was talking about in the other comment.

mcimadamore (approved):
Looks good (thanks for taking care of this). The javadoc can probably be consolidated a bit further.

JornVernee:
Thanks for the suggestion. I've rewritten these two paragraphs. I now list the behavior of two cases non-void/void separately.

JornVernee:
Updated the javadoc for the void case after some offline discussion on the wording. It should be much simpler now.

mcimadamore (approved):

JornVernee:
Still under review.

PR #14497, opened 2023-06-15
8283214: [macos] Screen magnifier does not show the magnified text for JcomboBox

kumarabhi006:
The issue exist only for non-editable combobox and the root cause is accessible object is not created due to incorrect index returned from component class which results in no a11y API invoked.

Proposed solution is to return the correct accessible child from getAccessibleChild method which is AquaComboBoxButton (arrowButton) instance and that results in invoking the a11y APIs to return the current selected item in combobox. 

Further when the application comes up first time the accessible name is not set for current displayed item in JCombobox that is handled in AquaComboBoxButton which will take care for the current selected item as well as if user modifies the selection by drop-down list.

CI link is posted in JBS.

---------
### Progress

### Issue
 * [JDK-8283214](https://bugs.openjdk.org/browse/JDK-8283214): [macos] Screen magnifier does not show the magnified text for JComboBox (**Bug** - P3)

### Reviewers
 * [Artem Semenov](https://openjdk.org/census#asemenov) (@savoptik - Committer) ⚠️ Review applies to [db0a35b7](https://git.openjdk.org/jdk/pull/14497/files/db0a35b72dede1d90acfb5487f08efeb09cb6ef2)
 * [Alexander Zuev](https://openjdk.org/census#kizune) (@azuev-java - **Reviewer**)
 * [Alexey Ivanov](https://openjdk.org/census#aivanov) (@aivanov-jdk - **Reviewer**) ⚠️ Review applies to [6b41f668](https://git.openjdk.org/jdk/pull/14497/files/6b41f668d02fa5913b65a7b9b60cf8eaa44de55b)

### Contributors

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14497/head:pull/14497` \
`$ git checkout pull/14497`

`$ git checkout pull/14497` \
`$ git pull https://git.openjdk.org/jdk.git pull/14497/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14497`

`$ git pr show -t 14497`

</details>

<a href="https://git.openjdk.org/jdk/pull/14497.diff">https://git.openjdk.org/jdk/pull/14497.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14497#issuecomment-1593358602)


src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxButton.java
```diff
@@ -154,6 +156,11 @@ public void paintComponent(final Graphics g) {
             } else {
                 painter.state.set(isSquare ? Widget.BUTTON_POP_UP_SQUARE : Widget.BUTTON_POP_UP);
             }
+
+            AccessibleContext ac = this.getAccessibleContext();
+            if (ac != null && (comboBox.getSelectedItem() != null)) {
+                ac.setAccessibleName(comboBox.getSelectedItem().toString());
```
savoptik, line 162: Why are you using the currently selected element for the button name and not a specific name? Whether it will sound right when using a screen reader. Wouldn't this be misleading to a blind user? Especially since it will only change when this method is called.

savoptik, line 162: Are you sure that the toString() method always provides a valid string, and not a descriptive JSON or some other character set that is useless for a blind user?

savoptik (commented):
@azuev-java @NikitkoCent please review.

src/java.desktop/share/classes/javax/swing/plaf/basic/BasicComboBoxUI.java
```diff
@@ -1132,21 +1127,24 @@ public Accessible getAccessibleChild(JComponent c, int i) {
         // 0 = the popup
         // 1 = the editor
```
NikitkoCent, line 1128: The comment here needs to be updated

src/java.desktop/share/classes/javax/swing/plaf/basic/BasicComboBoxUI.java
```diff
@@ -1118,12 +1118,7 @@ else if (comboBox.getModel().getSize() > 0) {
     // This is currently hacky...
     @Override
     public int getAccessibleChildrenCount(JComponent c) {
-        if ( comboBox.isEditable() ) {
-            return 2;
-        }
-        else {
-            return 1;
-        }
+        return 2;
```
NikitkoCent, line 1121: I think the logic here has to be equal to the logic in `getAccessibleChild `, otherwise `getAccessibleChildrenCount` returns 2 even when `getAccessibleChild(..., 1)` returns null.

src/java.desktop/share/classes/javax/swing/plaf/basic/BasicComboBoxUI.java
```diff
@@ -1132,21 +1127,24 @@ public Accessible getAccessibleChild(JComponent c, int i) {
...
+            case 1:
+                if (comboBox.isEditable()
+                        && (editor instanceof Accessible)) {
+                    AccessibleContext ac = ((Accessible) editor).getAccessibleContext();
+                    ac.setAccessibleParent(comboBox);
+                    return (Accessible) editor;
+                } else if (!comboBox.isEditable()
+                        && (arrowButton instanceof Accessible)) {
+                    return (Accessible) arrowButton;
+                }
```
NikitkoCent, line 1146: How about to idiomatize the code a bit by using Java 14's enhanced `instanceof`?

NikitkoCent (changes requested):

src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxButton.java
kumarabhi006: > Why are you using the currently selected element for the button name and not a specific name?

As it is required to show the magnified text for the selected item in JComboBox. Setting it to some specific name always results into that name irrespective of selected item in JCombobox.

> Whether it will sound right when using a screen reader. Wouldn't this be misleading to a blind user?

I checked with VO announcement with or without this chunk of code, it is same in both case. 

> Especially since it will only change when this method is called.

This method is called whenever the current selection is changed so accessible name is set to correct item.

kumarabhi006: I tested several times, didn't get such kind of result and as per java docs toString method returns `a string representation of the object.` So, I think it will be a valid string.

src/java.desktop/share/classes/javax/swing/plaf/basic/BasicComboBoxUI.java
kumarabhi006: Sure, I will update.

src/java.desktop/share/classes/javax/swing/plaf/basic/BasicComboBoxUI.java
kumarabhi006: Can you please elaborate more or what logic can be added here?

src/java.desktop/share/classes/javax/swing/plaf/basic/BasicComboBoxUI.java
kumarabhi006: Yeah it can be done to use enhanced `instanceOf`.

src/java.desktop/share/classes/javax/swing/plaf/basic/BasicComboBoxUI.java
kumarabhi006: Updated.

src/java.desktop/share/classes/javax/swing/plaf/basic/BasicComboBoxUI.java
NikitkoCent: I mean `getAccessibleChild(..., 1)` can still return null (when both `if`s of `case 1` resolve to false). Let's consider the following code:
```java
public void doSmthWith(JComponent component, BasicComboBoxUI ui) {
    if (ui.getAccessibleChildrenCount(component) > 1) {
        // well, we have at least 2 components, so we can expect to get a valid object below
        final var ac = ui.getAccessibleChild(component, 1)
                                  .getAccessibleContext(); // <= possible NPE
        // ...
    }
}
```

I think we have to return 1 from `getAccessibleChildrenCount(...)` for the cases when `getAccessibleChild(..., 1)` returns `null`

kumarabhi006: I think it is not needed to return 1 from `getAccessibleChildrenCount(...)`.  `getAccessibleChildrenCount(...)` is used to get the exact number of accessible child for a component and afterwards `getAccessibleChild(..., index)` method can be invoked to get the child.

It should be ok to return null if the switch case condition fails or index doesn't match with case statement. Although chances of returning null is very unlikely for index 0 and 1.

src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxButton.java
savoptik: JSON with enumeration of fields is also a text description of the object.
But if you checked that there is no garbage, then good.

savoptik (approved):

kumarabhi006:
@azuev-java  Please review this PR.

src/java.desktop/share/classes/javax/swing/plaf/basic/BasicComboBoxUI.java
```diff
@@ -1118,35 +1118,33 @@ else if (comboBox.getModel().getSize() > 0) {
...
-        // 1 = the editor
+        // 1 = the editor for editable combobox and the arrow button for non-editable combobox
         switch ( i ) {
-        case 0:
-            if ( popup instanceof Accessible ) {
-                AccessibleContext ac = ((Accessible) popup).getAccessibleContext();
-                ac.setAccessibleParent(comboBox);
-                return(Accessible) popup;
-            }
-            break;
```
aivanov-jdk, line 1141: Why do you change indentation? To minimise the number of changed lines, please keep the existing indentation.

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
```diff
@@ -0,0 +1,93 @@
...
+        try {
+            SwingUtilities.invokeAndWait(
+                    TestJComboBoxScreenMagnifier::createAndShowUI);
+            passFailJFrame.awaitAndCheck();
+        } finally {
+            SwingUtilities.invokeAndWait(() -> {
+                if (frame != null) {
+                    frame.dispose();
+                }
+            });
```
aivanov-jdk, line 68: This is redundant, `PassFailJFrame` automatically handles disposing of windows which are added using the `addTestWindow` method.

victordyakov:
JCK open bug: https://bugs.openjdk.org/browse/JCK-7319452

src/java.desktop/share/classes/javax/swing/plaf/basic/BasicComboBoxUI.java
aivanov-jdk: It's a matter of style.

The [old conventions](https://www.oracle.com/docs/tech/java/codeconventions.pdf)¹ suggest that `case` statements shouldn't be indented. The code followed this style.

[A latest draft](https://cr.openjdk.org/~alundblad/styleguide/index-v6.html) which is often referred to explicitly states that `case` statements should be indented with four spaces and the statements inside should be be indented with additional four spaces.

The updated code follows the new style. Since you're updating the code in the `switch` statement, it's fine to change the indentation… provided we agree to *follow the new style*.

¹ The [HTML version for `switch` statement](https://www.oracle.com/java/technologies/javase/codeconventions-statements.html#468) is completely *unreadable now*.

victordyakov:
JCK challenge public bug https://bugs.openjdk.org/browse/JCK-7319452

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
```diff
@@ -0,0 +1,93 @@
...
+ * questions.
+ */
+
+/*
+ * @test
+ * @bug 8283214
+ * @library /java/awt/regtesthelpers
+ * @build PassFailJFrame
+ * @requires (os.family == "mac")
+ * @summary Verifies if JComboBox selected itemmagnifies using
```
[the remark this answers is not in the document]
kumarabhi006: Updated.

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
kumarabhi006: Updated.

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
```diff
@@ -0,0 +1,93 @@
...
+            });
+        }
+    }
+    private static void createAndShowUI() {
+        frame = new JFrame("JComboBox A11Y Screen Magnifier Test");
+
+        String[] fruits = new String[] {"Apple", "Orange",
+                "Mango", "Pine Apple", "Banana"};
+        JComboBox<String> comboBox = new JComboBox<String>(fruits);
+        JPanel fruitPanel = new JPanel(new GridLayout(1,2));
```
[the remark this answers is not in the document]
kumarabhi006: Updated.

src/java.desktop/share/classes/javax/swing/plaf/basic/BasicComboBoxUI.java
kumarabhi006: > It's a matter of style.
> 
> The [old conventions](https://www.oracle.com/docs/tech/java/codeconventions.pdf)¹ suggest that `case` statements shouldn't be indented. The code followed this style.
> 
> [A latest draft](https://cr.openjdk.org/~alundblad/styleguide/index-v6.html) which is often referred to explicitly states that `case` statements should be indented with four spaces and the statements inside should be be indented with additional four spaces.
> 
> The updated code follows the new style. Since you're updating the code in the `switch` statement, it's fine to change the indentation… provided we agree to _follow the new style_.
> 
> ¹ The [HTML version for `switch` statement](https://www.oracle.com/java/technologies/javase/codeconventions-statements.html#468) is completely _unreadable now_.

The indented code is more readable and as you pointed it is inline with new style too. So, I am ok with the current indentation if we are on same page.

src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxButton.java
kumarabhi006: I tried with the `Integer` object and it works fine. Please suggest if anything else I can test.

aivanov-jdk: > I tried with the `Integer` object and it works fine. Please suggest if anything else I can test.

Obviously, it does.

What about a custom object which doesn't override `toString`? Although this shouldn't be a problem either.

What if `JComboBox` uses a non-standard renderer which takes a property from an object? Let's assume, `JComboBox` is populated with `Rectangle` objects, a custom renderer is set which displays the value of the `bottom` field in the combo box. Your code, if I understand it correctly, will default to _full rectangle_ rather than what's really displayed.

How does `JTable` handle this kind of situation? Or `JList`? Swing relies on renderers which makes it capable of displaying an arbitrary object (it uses `toString` by default, if I'm not mistaken), a renderer could provide a custom view of an object in a cell, in a list.

kumarabhi006: I need to check with custom renderer. 

If I understand correctly, with current fix the `getSelectedItem` method will return the `rectangle` object and then calling `toString` method will return rectangle's dimension instead of the displayed text in JComboBox and that may result incorrectly. Right?

aivanov-jdk: > If I understand correctly, with current fix the `getSelectedItem` method will return the `rectangle` object and then calling `toString` method will return rectangle's dimension instead of the displayed text in JComboBox and that may result incorrectly. Right?

Yes.

I wonder if other components support this scenario.

kumarabhi006: > What if `JComboBox` uses a non-standard renderer which takes a property from an object? Let's assume, `JComboBox` is populated with `Rectangle` objects, a custom renderer is set which displays the value of the `bottom` field in the combo box. Your code, if I understand it correctly, will default to _full rectangle_ rather than what's really displayed.
> 
You were right. At least in this case the current fix returns _full_rectangle_ dimension in a string form in stead of the text displayed in JComboBox. 
I think it is required to look for alternate fix, probably to get the displayed text from JComboBox and set it as an accessibleName.

 >How does JTable handle this kind of situation? Or JList?

I case of JTable and JList, the accessible context is of type JLabel which returns the accessibleName in string format.

aivanov-jdk: > > What if `JComboBox` uses a non-standard renderer which takes a property from an object? Let's assume, `JComboBox` is populated with `Rectangle` objects, a custom renderer is set which displays the value of the `bottom` field in the combo box. Your code, if I understand it correctly, will default to _full rectangle_ rather than what's really displayed.
> 
> You were right. At least in this case the current fix returns _full_rectangle_ dimension in a string form instead of the text displayed in JComboBox.

Thank you for confirming it.

> I had a doubt here if a custom renderer is used to populate the JComboBox, isn't it required to `override` the `toString` method to return useful information from user end?

No, it's not required to override the `toString` method if the programmer doesn't use it.

> If not then I think it is required to look for alternate fix, probably to get the displayed text from JComboBox and set it as an accessibleName.

Could be…

How would you get the text? It should be provided by a renderer…

> > How does JTable handle this kind of situation? Or JList?
> 
> I case of JTable and JList, the accessible context is of type JLabel which returns the accessibleName in string format.

This doesn't really answer the question. Do `JTable` and `JList` correctly display the magnified text when a custom renderer is used? In the case with rectangle and a custom renderer? Do they still display text from `toString`?

If neither `JTable` nor `JList` handle this situation, it's okay to go with `toString` as you implemented.

aivanov-jdk: How does `JTable` handle the situation where an image is displayed? You can use SwingSet2 for testing: it displays colored text and images.

kumarabhi006: > This doesn't really answer the question. Do JTable and JList correctly display the magnified text when a custom renderer is used? In the case with rectangle and a custom renderer? Do they still display text from toString?

Checked with JList and it correctly displayed the magnified text for a rectangle and custom renderer.

aivanov-jdk: > > This doesn't really answer the question. Do JTable and JList correctly display the magnified text when a custom renderer is used? In the case with rectangle and a custom renderer? Do they still display text from toString?
> 
> Checked with JList and it correctly displayed the magnified text for a rectangle and custom renderer. The `toString` method of `Rectangle` class is not invoked.

It means you have to look into how it's done in JList so that JCheckBox works in a similar way.

kumarabhi006: > It means you have to look into how it's done in JList so that JCheckBox works in a similar way.

For a JList, whenever `setText` is invoked from renderer's `getListCellRendererComponent` method, JLabel's text is set to the value displayed in JList. So, a11y API invokes the `getAccessibleName` method for `AccessibleJLabel` and that returns the correct string.

But in case of JComboBox, a11y API invokes the `getAccessibleName` method for `AccessibleJButton` and that returns an empty string.

To fix this we need to get the current displayed text is JComboBox as the other way around is not correct for custom renderer.

aivanov-jdk: > But in case of JComboBox, a11y API invokes the `getAccessibleName` method for `AccessibleJButton` and that returns an empty string.

And this seems to be the problem. The `getAccessibleName` method should be on JComboBox, or its editor, which should return the result of `getAccessibleName` of the renderer component.

Does this work with editable combo box?

kumarabhi006: > And this seems to be the problem. The getAccessibleName method should be on JComboBox, or its editor, which should return the result of getAccessibleName of the renderer component.

Since the a11y APIs are called upon the `AquaComboBoxButton` (subclass of JButton) instance, need to set the text of the button to current displayed text of JComboBox. JComboBox contains `getAccessibleName` method but it is not invoked.

> Does this work with editable combo box?

Yeah, it worked for editable JComboBox. For editable JComboBox, the string is returned by NavigableTextAccessibility and it returns the current text displayed in editor.

kumarabhi006: Accessible name is set by getting the current text displayed in JComboBox (using `renderer getText()` method) in stead of setting it by JComboBox  `getSelectedItem().toString() `method. Tested with custom renderer and it shows the magnified text as displayed in JComboBox.

aivanov-jdk: > > And this seems to be the problem. The getAccessibleName method should be on JComboBox, or its editor, which should return the result of getAccessibleName of the renderer component.
> 
> Since the a11y APIs are called upon the `AquaComboBoxButton` (subclass of JButton) instance, need to set the text of the button to current displayed text of JComboBox. JComboBox contains `getAccessibleName` method but it is not invoked.

This is weird. I expect the accessibility API is called on the `JComboBox` in both cases where it's editable or not. Could this be a bug? Should we fix it in a different way?

> > Does this work with editable combo box?
> 
> Yeah, it worked for editable JComboBox. For editable JComboBox, the string is returned by NavigableTextAccessibility and it returns the current text displayed in editor.

That's good.

And it makes me wonder why non-editable combo box behaves so differently.

aivanov-jdk: > > This is weird. I expect the accessibility API is called on the JComboBox in both cases where it's editable or not. Could this be a bug? Should we fix it in a different way?
> 
> AFAIK, I don't think that it is a bug because the arrow button in Aqua LAF is an instance of AquaComboBoxButton. So, native side invoked a11y API based on component. Similarly, For non-editable JComboBox, editor is an instance of JTextField and a11y API invoked on NavigableTextAccessibility.

That's the problem: *why* is it called on *the combo button* instead of the combo box itself? It doesn't look right.

In case of editable `JComboBox`, it consists of two components: an editor and the button. I assume the implementation of accessible interfaces gets called on the first child of the combo box. It's about right for the editable component: the editor is the main component of the combo box.

But it doesn't look right to me that the drop-down button is the main component of the combo box if it's non-editable. The combo box should handle the accessibility request itself in this case. It still uses the same renderer which can't be put into editable state. The renderer is never added into containment hierarchy.

Or there could be a *dummy* editor which handles the accessibility.

kumarabhi006: > But it doesn't look right to me that the drop-down button is the main component of the combo box if it's non-editable. The combo box should handle the accessibility request itself in this case. It still uses the same renderer which can't be put into editable state. The renderer is never added into containment hierarchy.

In case of non-editable combobox there is no editor as such and arrowButton is the main component. When native side request a11y information, accessible interface for combo button is returned (as only two accessible component are present. One is arrow button and other is popup). Subsequent accessible APIs are invoked for combo button to get accessible name.

aivanov-jdk: > > But it doesn't look right to me that the drop-down button is the main component of the combo box if it's non-editable. The combo box should handle the accessibility request itself in this case. It still uses the same renderer which can't be put into editable state. The renderer is never added into containment hierarchy.
> 
> In case of non-editable combobox there is no editor as such and arrowButton is the main component. When native side request a11y information, accessible interface for combo button is returned (as only two accessible component are present. One is arrow button and other is popup). Subsequent accessible APIs are invoked for combo button to get accessible name.

Do you think it is correct?

I admit refactoring / rewriting accessibility implementation for `JComboBox` is out of scope in this bug. Yet the current way doesn't feel right.

kumarabhi006: > Do you think it is correct?
>
> I admit refactoring / rewriting accessibility implementation for JComboBox is out of scope in this bug. Yet the current way doesn't feel right.

I think so. 
If I am not wrong the non-editable JComboBox doesn't have any editor, so the arrowButton width and height is equal to the combobox width and height. This results in arrowbutton rendering for entire display area of combobox. Due to this we are getting the accessible interface invoked for combobox button.

aivanov-jdk: Yes, non-editable combo box doesn't have the editor. Yet the arrowButton is not what displays the text. It is for this reason that I think it is wrong that the button becomes the main component which handles accessibility for combo box if it's not editable. It is not the button that displays the text.

Setting accessible name in `doRendererPaint` doesn't feel right either.

kumarabhi006: As per our discussion, I didn't find other way to set the accessible name.

src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxButton.java
```diff
@@ -228,6 +230,17 @@ protected void doRendererPaint(final Graphics g, final ButtonModel buttonModel,
 
         rendererPane.paintComponent(g, c, this, left, top, cWidth, height, shouldValidate); // h - (insets.top + insets.bottom) );
 
+        // fix for 8283214
```
mrserb, line 233: we usually do not mention the bugid in the comments, it can be found in the git history, just describe why it is needed.

src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxButton.java
```diff
@@ -228,6 +230,17 @@ protected void doRendererPaint(final Graphics g, final ButtonModel buttonModel,
...
+        // set the accessible name to the displayed text in JComboBox.
+        // screen magnifier queries to get the accessible name to display magnified text.
+
+        AccessibleContext ac = this.getAccessibleContext();
+        if (ac != null && !editable && c instanceof JLabel label) {
+            if (label.getText() != null) {
+                ac.setAccessibleName(label.getText());
+            }
+        }
```
mrserb, line 242: How it will work in other L&Fs?

src/java.desktop/share/classes/javax/swing/plaf/basic/BasicComboBoxUI.java
mrserb: please double-check how it will work in Windows, it might be possible that jaws expects 1.

src/java.desktop/share/classes/javax/swing/plaf/basic/BasicComboBoxUI.java
```diff
@@ -1118,35 +1118,33 @@ else if (comboBox.getModel().getSize() > 0) {
...
-                 && (editor instanceof Accessible) ) {
-                AccessibleContext ac = ((Accessible) editor).getAccessibleContext();
-                ac.setAccessibleParent(comboBox);
-                return(Accessible) editor;
-            }
-            break;
+            case 0:
+                if ( popup instanceof Accessible accessiblePopup) {
+                    AccessibleContext ac = accessiblePopup.getAccessibleContext();
+                    ac.setAccessibleParent(comboBox);
```
mrserb, line 1133: do we need to reset it back at some point?

src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxButton.java
kumarabhi006: In case of other LAFs except **Aqua** in MacOS, the accessible component invoked for screen magnifier is an instance of `AccessibleJComboBox` and subsequently the a11y API called on `AccessibleJComboBox` to return the accessible name.

But for aqua, the accessible component invoked for screen magnifier is an instance for `AquaComboBoxButton`, so it is required to return the accessible button instance for a11y APIs to get called for retrieving accessible name.

src/java.desktop/share/classes/javax/swing/plaf/basic/BasicComboBoxUI.java
kumarabhi006: I don't think so, as comboBox is the accessible parent for popup. Moreover, this piece of code is unchanged for the current fix.

src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxButton.java
kumarabhi006: Updated.

kumarabhi006:
> please double-check how it will work in Windows, it might be possible that jaws expects 1.

I will check this and update if there is any change in announcement with JAWS with this fix.

src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxButton.java
```diff
@@ -33,6 +33,8 @@
 import apple.laf.JRSUIState;
 import apple.laf.JRSUIConstants.*;
```
aivanov-jdk, line 34: I think we should expand all wild card imports.

src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxButton.java
aivanov-jdk: That's exactly my concern: in other L&Fs `AccessibleJComboBox` handles the scenario yet in Aqua L&F it is handled by a different component which is not a problem okay per se. What is a problem is that retrieving the accessible name depends on paint:
* if `AquaComboBoxButton.paint` hasn't been called for whatever reason, there's no accessible name;
* if the value in the combo box is changed, the accessible name isn't updated until paint occurs.

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
```diff
@@ -0,0 +1,85 @@
...
+                "System Preference -> Accessibility -> Zoom -> " +
+                "Select ( Enable Hover Text) \n\n" +
+            "2) Move the mouse over the combobox selected item by pressing  " +
+                "\"cmd\" button.\n\n" +
+            "3) If magnified label is visible, Press Pass else Fail.";
+
+    public static void main(String[] args) throws InterruptedException,
+             InvocationTargetException {
+        PassFailJFrame passFailJFrame = new PassFailJFrame(
+                "JMenu Screen Magnifier Test Instructions", INSTRUCTIONS, 5, 12, 40);
```
aivanov-jdk, line 58:
```suggestion
                "JComboBox Screen Magnifier Test Instructions", INSTRUCTIONS, 5, 12, 40);
```

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
```diff
@@ -0,0 +1,85 @@
...
+
+        SwingUtilities.invokeAndWait(TestJComboBoxScreenMagnifier::createAndShowUI);
+        passFailJFrame.awaitAndCheck();
+    }
+    private static void createAndShowUI() {
```
aivanov-jdk, line 63:
```suggestion
    }

    private static void createAndShowUI() {
```

Please add a blank line between methods.

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
```diff
@@ -0,0 +1,85 @@
...
+import javax.swing.JLabel;
+import javax.swing.JPanel;
+import javax.swing.SwingUtilities;
+
+public class TestJComboBoxScreenMagnifier {
+    private static JFrame frame;
+    private static final String INSTRUCTIONS =
+            "1) Enable Screen magnifier on the Mac \n\n" +
+                "System Preference -> Accessibility -> Zoom -> " +
+                "Select ( Enable Hover Text) \n\n" +
```
aivanov-jdk, line 50:
```suggestion
                "Select \"Enable Hover Text\"\n\n" +
```

src/java.desktop/share/classes/javax/swing/plaf/basic/BasicComboBoxUI.java
```diff
@@ -1118,35 +1118,33 @@ else if (comboBox.getModel().getSize() > 0) {
...
-        case 1:
-            if ( comboBox.isEditable()
-                 && (editor instanceof Accessible) ) {
-                AccessibleContext ac = ((Accessible) editor).getAccessibleContext();
-                ac.setAccessibleParent(comboBox);
-                return(Accessible) editor;
-            }
-            break;
+            case 0:
+                if ( popup instanceof Accessible accessiblePopup) {
```
aivanov-jdk, line 1131:
```suggestion
                if (popup instanceof Accessible accessiblePopup) {
```

src/java.desktop/share/classes/javax/swing/plaf/basic/BasicComboBoxUI.java
kumarabhi006: > please double-check how it will work in Windows, it might be possible that jaws expects 1.

@mrserb I observed the difference in announcement with JAWS with the fix. 
Since the magnifier is used only for macos, proposed fix is now handled in AquaComboBoxUI instead of BasicComboBoxUI. Changes in BasicComboBoxUI is reverted back.

src/java.desktop/share/classes/javax/swing/plaf/basic/BasicComboBoxUI.java
kumarabhi006: This code segment is reverted back and handled in `AquaComboBoxUI` now.

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
kumarabhi006: Updated.

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
kumarabhi006: Updated.

src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxButton.java
kumarabhi006: Updated.

src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxButton.java
kumarabhi006: > That's exactly my concern: in other L&Fs `AccessibleJComboBox` handles the scenario yet in Aqua L&F it is handled by a different component which is not a problem okay per se. What is a problem is that retrieving the accessible name depends on paint:
> 
>     * if `AquaComboBoxButton.paint` hasn't been called for whatever reason, there's no accessible name;
> 
>     * if the value in the combo box is changed, the accessible name isn't updated until paint occurs.

I rechecked again but couldn't find any other way to either set accessible name or get the current selected item other than `AquaComboBoxButton.paint`.

kumarabhi006:
@aivanov-jdk Updated the suggested changes. Do you have any other suggestions?

kumarabhi006:
@mrserb Checked with JAWS on windows and updated accordingly. 
Please have a look.

src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxButton.java
mrserb: >But for aqua, the accessible component invoked for screen magnifier is an instance for AquaComboBoxButton, so >it is required to return the accessible button instance for a11y APIs to get called for retrieving accessible name.

Can we inject some code into the AquaComboBoxButton so it will know what name should be used if that button is used "inside" combobox?

kumarabhi006: > Can we inject some code into the AquaComboBoxButton so it will know what name should be used if that button is used "inside" combobox?

I didn't get you.

kumarabhi006: > Can we inject some code into the AquaComboBoxButton so it will know what name should be used if that button is used "inside" combobox?
If I understood this correctly, we are setting the accessible name for the button and that can be retrieved by `getAccessibleName `API for the button if it is required inside combobox .

src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxButton.java
```diff
@@ -228,6 +249,16 @@ protected void doRendererPaint(final Graphics g, final ButtonModel buttonModel,
 
         rendererPane.paintComponent(g, c, this, left, top, cWidth, height, shouldValidate); // h - (insets.top + insets.bottom) );
 
+        // set the accessible name to the displayed text in JComboBox.
+        // screen magnifier queries to get the accessible name to display magnified text.
```
azuev-java, line 253: I still do not think that setting accessible name in paint is a good idea. Setting it during the initialization and adding a property change listener to the accessible context to track changes seems like a better way to do it.

kumarabhi006: Ok, I will check this.

aivanov-jdk: > I still do not think that setting accessible name in paint is a good idea.

It's been my biggest concern. But I didn't even think about it this way.

The problem is that the displayed value is known only when the component is painted. Other components seem to handle it… somehow.

kumarabhi006: > Setting it during the initialization and adding a property change listener to the accessible context to track changes seems like a better way to do it.

@azuev-java We have a ItemListener for JComboBox defined in AquaComboBoxUI here to handle any changes in combobox selected item.

https://github.com/openjdk/jdk/blob/9e6cb620486ac7b0adaefeb2000babf3ea31207f/src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxUI.java#L135

Tried setting the `accessibleName` for arrowButton in `itemStateChanged` method. So whenever combobox value  changes using list selection, **accessibleName is set to the selected Item**.

For initial accessibleName, it can be set to the first value of data model inside `AquaComboBoxButton constructor` code.

`However, this doesn't work with custom renederer.`

kumarabhi006: @azuev-java Please review.

src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxButton.java
```diff
@@ -65,6 +86,12 @@ public void setArmed(final boolean armed) {
         });
 
         setEnabled(comboBox.isEnabled());
+
+        AccessibleContext ac = this.getAccessibleContext();
+        if (ac != null && !comboBox.isEditable() && list.getModel().getSize() > 0) {
+            ac.setAccessibleName(String.valueOf(list.getModel()
+                    .getElementAt(0)));
+        }
```
azuev-java, line 94: So i see it updates accessible name whenever selection is changed. What about the initial value? Will the combobox initial value be displayed if it was never changed?

kumarabhi006: > What about the initial value? Will the combobox initial value be displayed if it was never changed?

The initial value is set in `AquaComboBoxButton` constructor. Yes, initial value is displayed with screen magnifier even if it has never changed.

azuev-java (approved):
Looks good now.

aivanov-jdk:
This is definitely an improvement compared to the complete lack of magnified text on hover, but it still doesn't handle custom renderers.

I've been digging into how this works in Metal but I couldn't figure it out completely. Eventually, it gets to `JList` of the popup, the list returns the renderer component as its child at the specified position — this is how text from the renderer gets used as the accessible name.

This should work for Aqua L&F too, all UI classes are derived from `BasicUI`, but for whatever reason it doesn't, and I didn't figure it out unfortunately.

However, I came up with an alternative solution which handles custom renderers nicely. You can find my version at [commit eee17ca](https://github.com/openjdk/jdk/commit/eee17ca7fc07fb87997040589bde83f5fad9fdea), it shows the changes to your latest version in the PR. You can also view [the diff to jdk/master](https://github.com/openjdk/jdk/compare/master...aivanov-jdk:jdk:eee17ca7fc07fb87997040589bde83f5fad9fdea).

Ideally, I'd like to make Aqua L&F classes work like Metal or Nimbus do. I spent quite a lot of time, and it requires even more time to understand how it gets to `JList.AccessibleJList` and `AccessibleJListChild`.

kumarabhi006:
> This is definitely an improvement compared to the complete lack of magnified text on hover, but it still doesn't handle custom renderers.
> 
> I've been digging into how this works in Metal but I couldn't figure it out completely. Eventually, it gets to `JList` of the popup, the list returns the renderer component as its child at the specified position — this is how text from the renderer gets used as the accessible name.
> 
> This should work for Aqua L&F too, all UI classes are derived from `BasicUI`, but for whatever reason it doesn't, and I didn't figure it out unfortunately.
> 
> However, I came up with an alternative solution which handles custom renderers nicely. You can find my version at [commit eee17ca](https://github.com/openjdk/jdk/commit/eee17ca7fc07fb87997040589bde83f5fad9fdea), it shows the changes to your latest version in the PR. You can also view [the diff to jdk/master](https://github.com/openjdk/jdk/compare/master...aivanov-jdk:jdk:eee17ca7fc07fb87997040589bde83f5fad9fdea).
> 
> Ideally, I'd like to make Aqua L&F classes work like Metal or Nimbus do. I spent quite a lot of time, and it requires even more time to understand how it gets to `JList.AccessibleJList` and `AccessibleJListChild`.

Thanks @aivanov-jdk for your detailed analysis. I will go through the alternate fix you suggested and check the behavior of VoiceOver and will run the CI jobs to ensure there is no regression as such.

aivanov-jdk:
@kumarabhi006 I attached the app that I used for testing to [JDK-8283214](https://bugs.openjdk.org/browse/JDK-8283214): `MagnifierTest.java`

kumarabhi006:
@aivanov-jdk I tested with your suggested solution and it looks good to me. Observed a test failure (regression) while running the CI jobs with the fix. Modified the test to fix the issue.
Now, CI jobs are ok. Please review.

src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxButton.java
```diff
@@ -169,12 +198,15 @@ public void paintComponent(final Graphics g) {
         }
     }
 
-    protected void doRendererPaint(final Graphics g, final ButtonModel buttonModel, final boolean editable, final Insets insets, int left, int top, int width, int height) {
+    final Component getRendererComponent() {
```
aivanov-jdk, line 201:
```suggestion
    private Component getRendererComponent() {
```

I wanted to use this method in `AquaComboBoxUI` but eventually I added `updateAccessibleName` instead, thus this method can be made `private`.

src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxUI.java
```diff
@@ -133,7 +139,13 @@ protected ItemListener createItemListener() {
         return new ItemListener() {
             long lastBlink = 0L;
             public void itemStateChanged(final ItemEvent e) {
+
                 if (e.getStateChange() != ItemEvent.SELECTED) return;
```
aivanov-jdk, line 143:
```suggestion
            public void itemStateChanged(final ItemEvent e) {
                if (e.getStateChange() != ItemEvent.SELECTED) return;
```
The blank line at the start of the method isn't necessary, is it?

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+
+/*
+ * @test
+ * @bug 8283214
+ * @library /java/awt/regtesthelpers
+ * @build PassFailJFrame
+ * @requires (os.family == "mac")
+ * @summary Verifies if JComboBox selected item magnifies using
+ * screen magnifier a11y tool.
```
aivanov-jdk, line 31:
```suggestion
 * @summary Verifies if item selected in JComboBox magnifies using
 *          screen magnifier a11y tool
```

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
```diff
@@ -0,0 +1,86 @@
...
+
+        String[] fruits = new String[] {"Apple", "Orange",
+                "Mango", "Pine Apple", "Banana"};
+        JComboBox<String> comboBox = new JComboBox<String>(fruits);
+        JPanel fruitPanel = new JPanel(new GridLayout(1, 2));
+        JLabel fruitLabel = new JLabel("Fruits : ", JLabel.CENTER);
+        fruitLabel.getAccessibleContext().setAccessibleName("Fruits Label");
+        fruitPanel.add(fruitLabel);
+        fruitPanel.add(comboBox);
+        comboBox.getAccessibleContext().setAccessibleName("Fruit jCombobox");
```
aivanov-jdk, line 75:
```suggestion
        comboBox.getAccessibleContext().setAccessibleName("Fruit Combo box");
```

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
```diff
@@ -0,0 +1,86 @@
...
+
+        SwingUtilities.invokeAndWait(TestJComboBoxScreenMagnifier::createAndShowUI);
+        passFailJFrame.awaitAndCheck();
+    }
+
+    private static void createAndShowUI() {
+        frame = new JFrame("JComboBox A11Y Screen Magnifier Test");
+
+        String[] fruits = new String[] {"Apple", "Orange",
+                "Mango", "Pine Apple", "Banana"};
```
aivanov-jdk, line 68:
```suggestion
                "Mango", "Pineapple", "Banana"};
```

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
```diff
@@ -0,0 +1,86 @@
...
+import javax.swing.JComboBox;
+import javax.swing.JFrame;
+import javax.swing.JLabel;
+import javax.swing.JPanel;
+import javax.swing.SwingUtilities;
+
+public class TestJComboBoxScreenMagnifier {
+    private static JFrame frame;
+    private static final String INSTRUCTIONS =
+            "1) Enable Screen magnifier on the Mac \n\n" +
```
aivanov-jdk, line 48:
```suggestion
            "1) Enable Screen magnifier on the Mac\n\n" +
```

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
```diff
@@ -0,0 +1,86 @@
...
+                "System Preference -> Accessibility -> Zoom -> " +
+                "Select \"Enable Hover Text\"\n\n" +
+            "2) Move the mouse over the combobox selected item by pressing  " +
+                "\"command\" button.\n\n" +
```
aivanov-jdk, line 52:
```suggestion
            "2) Move the mouse over the combo box and press  " +
                "\"Command\" button.\n\n" +
```

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
```diff
@@ -0,0 +1,86 @@
...
+            "3) If magnified label is visible, Press Pass else Fail.";
```
aivanov-jdk, line 53:
```suggestion
            "3) If magnified label is visible, press Pass else Fail.";
```

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    private static void createAndShowUI() {
+        frame = new JFrame("JComboBox A11Y Screen Magnifier Test");
+
+        String[] fruits = new String[] {"Apple", "Orange",
+                "Mango", "Pine Apple", "Banana"};
+        JComboBox<String> comboBox = new JComboBox<String>(fruits);
+        JPanel fruitPanel = new JPanel(new GridLayout(1, 2));
+        JLabel fruitLabel = new JLabel("Fruits : ", JLabel.CENTER);
```
aivanov-jdk, line 71:
```suggestion
        JLabel fruitLabel = new JLabel("Fruits:", JLabel.CENTER);
```

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
```diff
@@ -0,0 +1,86 @@
...
+        fruitLabel.getAccessibleContext().setAccessibleName("Fruits Label");
+        fruitPanel.add(fruitLabel);
+        fruitPanel.add(comboBox);
+        comboBox.getAccessibleContext().setAccessibleName("Fruit jCombobox");
+        frame.getContentPane().add(fruitPanel, BorderLayout.CENTER);
+
+        PassFailJFrame.addTestWindow(frame);
+        PassFailJFrame.positionTestWindow(frame,
+                PassFailJFrame.Position.HORIZONTAL);
+        frame.setLocationRelativeTo(null);
```
aivanov-jdk, line 81:
```suggestion
```
It's not needed, the location of the frame is set explicitly by the above call to `PassFailJFrame.positionTestWindow`.

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
aivanov-jdk, line 1: You can use HTML for instructions and builder pattern for configuring `PassFailJFrame` if you like. I'm just advertising new features.

test/jdk/javax/swing/JComboBox/6567433/UpdateUIRecursionTest.java
aivanov-jdk, line 1: You should update the copyright year.

I wonder if it makes sense to resolve warnings about raw classes.

aivanov-jdk (changes requested):

aivanov-jdk:
> Observed a test failure (regression) while running the CI jobs with the fix.

The failed test is `javax/swing/JComboBox/6567433/UpdateUIRecursionTest.java`. It fails with `NullPointerException` because `PropertyChangeListener` in `AquaComboBoxUI` updates the accessible name of the arrow button. It does os before the accessible name is actually needed.

We should consider getting the accessible name lazily.

> Modified the test to fix the issue. Now, CI jobs are ok. Please review.

It looks fine to me.

But this change breaks backward compatibility: the failed test demonstrates it. If another application does something similar, it will throw NPE now.

You and I proved that adding `null`-checks in the listener or `updateAccessibleName` doesn't help. The only way to avoid it is to catch NPE.

This brings us to:

> Ideally, I'd like to make Aqua L&F classes work like Metal or Nimbus do. …

This way gets the accessible name *lazily*: it's requested only *when needed*. Therefore you should submit another bug to address the limitation of the current approach.

aivanov-jdk:
> > Ideally, I'd like to make Aqua L&F classes work like Metal or Nimbus do. …
> 
> This way gets the accessible name _lazily_: it's requested only _when needed_. Therefore you should submit another bug to address the limitation of the current approach.

Eventually, I've found the perfect solution: [commit 516af4c](https://github.com/openjdk/jdk/commit/516af4c62f59984f2d3c44bfb8ecae6fc2e88810) and [diff to jdk/master](https://github.com/openjdk/jdk/compare/master...aivanov-jdk:jdk:516af4c62f59984f2d3c44bfb8ecae6fc2e88810)

The `javax/swing/JComboBox/6567433/UpdateUIRecursionTest.java` test passes in its unmodified version, which means the new version does not break the backward compatibility. Due to this, 8283214 should be added to the list of bugids in `UpdateUIRecursionTest`.

kumarabhi006:
> > > Ideally, I'd like to make Aqua L&F classes work like Metal or Nimbus do. …
> > 
> > 
> > This way gets the accessible name _lazily_: it's requested only _when needed_. Therefore you should submit another bug to address the limitation of the current approach.
> 
> Eventually, I've found the perfect solution: [commit 516af4c](https://github.com/openjdk/jdk/commit/516af4c62f59984f2d3c44bfb8ecae6fc2e88810) and [diff to jdk/master](https://github.com/openjdk/jdk/compare/master...aivanov-jdk:jdk:516af4c62f59984f2d3c44bfb8ecae6fc2e88810)
> 
> The `javax/swing/JComboBox/6567433/UpdateUIRecursionTest.java` test passes in its unmodified version, which means the new version does not break the backward compatibility. Due to this, 8283214 should be added to the list of bugids in `UpdateUIRecursionTest`.

So, I need to revert the changes that is done in the current commit and just add the bug id 8283214... right?

aivanov-jdk:
> So, I need to revert the changes that is done in the current commit and just add the bug id 8283214... right?

Yes, that's right.

test/jdk/javax/swing/JComboBox/6567433/UpdateUIRecursionTest.java
aivanov-jdk: > > I wonder if it makes sense to resolve warnings about raw classes.
> 
> I didn't get this. Which warnings need to be resolved?

```text
test/jdk/javax/swing/JComboBox/6567433/UpdateUIRecursionTest.java:61: warning: [unchecked] unchecked call to JComboBox(E[]) as a member of the raw type JComboBox
        combo = new JComboBox(listData);
                ^
  where E is a type-variable:
    E extends Object declared in class JComboBox
test/jdk/javax/swing/JComboBox/6567433/UpdateUIRecursionTest.java:62: warning: [unchecked] unchecked call to setRenderer(ListCellRenderer<? super E>) as a member of the raw type JComboBox
        combo.setRenderer(this);
                         ^
  where E is a type-variable:
    E extends Object declared in class JComboBox
2 warnings
```

However, now the changes will be limited to adding a new bugid. As such, let's leave the file intact.

src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxButton.java
kumarabhi006: Updated.

src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxUI.java
kumarabhi006: Removed.

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
kumarabhi006: Updated.

test/jdk/javax/accessibility/JComboBox/TestJComboBoxScreenMagnifier.java
```diff
@@ -0,0 +1,91 @@
...
+/*
+ * @test
+ * @bug 8283214
+ * @library /java/awt/regtesthelpers
+ * @build PassFailJFrame
+ * @requires (os.family == "mac")
+ * @summary Verifies if item selected in JComboBox magnifies using
+ *          screen magnifier a11y tool
+ * @run main/manual TestJComboBoxScreenMagnifier
+ */
```
aivanov-jdk, line 33: Could you please move the jtreg tags below the imports, above the class declaration? This way, they're easier to see in an IDE. Not required though, some people still prefer the tags above the imports.

kumarabhi006: Updated.

aivanov-jdk (approved):
Looks good to me.

Someone else should also approve it.

src/java.desktop/macosx/classes/com/apple/laf/AquaComboBoxButton.java
```diff
@@ -233,4 +255,26 @@ protected void doRendererPaint(final Graphics g, final ButtonModel buttonModel,
...
+    private final class AccessibleAquaComboBoxButton extends AccessibleJButton {
+        @Override
+        public String getAccessibleName() {
+            String name = super.getAccessibleName();
+            if (name == null || name.isEmpty()) {
+                if (!comboBox.isEditable() && comboBox.getSelectedItem() != null) {
+                    Component c = getRendererComponent();
+                    name = c.getAccessibleContext().getAccessibleName();
+                }
+            }
```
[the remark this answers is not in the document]
kumarabhi006: Conditions combined.

> Since the arrow button is an internal component, we may drop calling super. However, I'm unsure about it.

I think we can keep this as it is.

kumarabhi006:
@azuev-java As suggested by @aivanov-jdk  the current fix ensures that screen magnifier works well with the default renderer as well as custom renderer. Ci jobs are green with this fix.

Can you please re-review this PR?

azuev-java (approved):

PR #14498, opened 2023-06-15
8310133: Effectivelly final condition not enforced in guards for binding variables from the same case

lahodaj:
Consider code containing snippet like:
```
...
case String s when s.isEmpty() -> {
s = null;
}
...
````

The `s` binding variable is obviously not effectively final, but the specification currently requires variables used inside the guard to be effectively final or final.

The cause here is in a position check - for error reporting we need to sent a shared variable `currentTree` to the case, but the effectively final check will only check variables declared before the start of `currentTree` (to avoid reporting issues for variables inside a lambda, for example). The proposed fix is to keep the case inside `currentTree`, but ensure the guard's starting position is used, to allow a proper effectively final check.

---------
### Progress

### Issue
 * [JDK-8310133](https://bugs.openjdk.org/browse/JDK-8310133): Effectivelly final condition not enforced in guards for binding variables from the same case (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14498/head:pull/14498` \
`$ git checkout pull/14498`

`$ git checkout pull/14498` \
`$ git pull https://git.openjdk.org/jdk.git pull/14498/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14498`

`$ git pr show -t 14498`

</details>

<a href="https://git.openjdk.org/jdk/pull/14498.diff">https://git.openjdk.org/jdk/pull/14498.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14498#issuecomment-1593385391)

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Flow.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Flow.java
@@ -3233,7 +3233,7 @@ public class Flow {
         void checkEffectivelyFinal(DiagnosticPosition pos, VarSymbol sym) {
             if (currentTree != null &&
                     sym.owner.kind == MTH &&
-                    sym.pos < currentTree.getStartPosition()) {
+                    sym.pos < getCurrentTreeStartPosition()) {
                 switch (currentTree.getTag()) {
                     case CLASSDEF:
                     case CASE:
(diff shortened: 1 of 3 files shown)


vicente-romero-oracle (approved):
looks sensible

PR #14499, opened 2023-06-15
JDK-8310061: Note if implicit annotation processing is being used

jddarcy:
Emit a note warning of the possibility of changing the implicit running of annotation processors in a future release.

---------
### Progress

### Issue
 * [JDK-8310061](https://bugs.openjdk.org/browse/JDK-8310061): Note if implicit annotation processing is being used (**Bug** - P3)

### Reviewers
 * [Brian Goetz](https://openjdk.org/census#briangoetz) (@briangoetz - **Reviewer**) ⚠️ Review applies to [6cbeec6c](https://git.openjdk.org/jdk/pull/14499/files/6cbeec6c0a567616c87930bde594fc6b02514291)
 * [Vicente Romero](https://openjdk.org/census#vromero) (@vicente-romero-oracle - **Reviewer**) ⚠️ Review applies to [6cbeec6c](https://git.openjdk.org/jdk/pull/14499/files/6cbeec6c0a567616c87930bde594fc6b02514291)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14499/head:pull/14499` \
`$ git checkout pull/14499`

`$ git checkout pull/14499` \
`$ git pull https://git.openjdk.org/jdk.git pull/14499/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14499`

`$ git pr show -t 14499`

</details>

<a href="https://git.openjdk.org/jdk/pull/14499.diff">https://git.openjdk.org/jdk/pull/14499.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14499#issuecomment-1595131036)


briangoetz (commented):
New diagnostic messages look reasonable

jddarcy:
The goal of the change in this PR is to emit a message to users who may be unknowingly relying on default implicit annotation processing that the long-standing default may be changing in a future release. The message is a Note rather than a Warning to avoid running afoul of build setups using configurations like "-Xlint:all -Werror".

Semantically, the note about implicit processors is treated as on by default. The note is *not* emitted if there are explicit configuration options related to annotation processing, including the new-in-JDK-21 "-proc:full". and not emitted if the "options" lint check is explicitly disabled, either through "-Xlint:-options" or "-Xlint:none".

A targeted regression test for the change to follow.

jddarcy:
PS To be clear, the goal is to get this change into JDK 22 and then backport it into JDK 21.

briangoetz (approved):

src/jdk.compiler/share/classes/com/sun/tools/javac/main/JavaCompiler.java
```diff
@@ -232,6 +236,10 @@ else if (option.equals("class"))
      */
     public Log log;
 
+    /** Whether or not the options lint category was initially disabled
```
jonathan-gibbons, line 239: You've copied my pattern of misuse.  "Whether or not" is not recommended use, whether or not you like it.
The consensus is to use just `whether`;  `whether or not` means `regardless`

FWIW, I still like the `whether or not` form!

src/jdk.compiler/share/classes/com/sun/tools/javac/main/JavaCompiler.java
```diff
@@ -424,6 +432,9 @@ public JavaCompiler(Context context) {
         moduleFinder.moduleNameFromSourceReader = this::readModuleName;
 
         options = Options.instance(context);
+        optionsCheckingInitiallyDisabled =
+            options.isSet(Option.XLINT_CUSTOM, "-options") ||
+            options.isSet(Option.XLINT_CUSTOM, "none");
```
jonathan-gibbons, line 437: You have a `context`, so you can use get a `Lint` object with `Lint.instance(context)` and then use `isEnabled(LintCategory)`

src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties
```diff
@@ -1769,6 +1769,14 @@ compiler.note.proc.messager=\
 compiler.note.multiple.elements=\
     Multiple elements named ''{1}'' in modules ''{2}'' were found by javax.lang.model.util.Elements.{0}.
 
+compiler.note.implicit.annotation.processing=\
+    Annotation processing is enabled because one or more processors were found\n\
+    on the class path. A future release of javac may disable annotation processing\n\
+    unless at least one processor is specified by name (-processor), or a search\n\
+    path is specified (--processor-path, --processor-module-path), or annotation\n\
+    processing is enabled explicitly (-proc:only, -proc:full).\n\
+    Use -Xlint:-options to suppress this message.
```
jonathan-gibbons, line 1778: You ought to be able to suppress the message by addressing the underlying case, not just by disabling the check.
For example, I would expect `-proc:none` to disable the message as well.

test/langtools/tools/javac/diags/examples.not-yet.txt
```diff
@@ -214,4 +214,4 @@ compiler.err.preview.without.source.or.release
 compiler.misc.illegal.signature                               # the compiler can now detect more non-denotable types before class writing
 
 # this one needs a forged class file to be reproduced
-compiler.err.annotation.unrecognized.attribute.name
\ No newline at end of file
+compiler.err.annotation.unrecognized.attribute.name
```
jonathan-gibbons: Is it just me? I cannot see any difference in these lines...  Is there a difference in trailing whitespace?

src/jdk.compiler/share/classes/com/sun/tools/javac/main/JavaCompiler.java
jddarcy: I'm slightly abusing the existing lint options category here. I want to have the effect that a new options-related link check is on by default. There are lint categories that are enabled by default, but options isn't one of them.

I'm implementing that policy by treating the implicit annotation processing check as enabled unless options checking has been explicitly disabled, which is what the code above does and is different than the standard is-options-lint-category enabled check.

test/langtools/tools/javac/processing/options/TestNoteOnImplicitProcessing.java
```diff
@@ -0,0 +1,274 @@
...
+        Path servicesFile =
+            apDir
+            .resolve("META-INF")
+            .resolve("services")
+            .resolve(Processor.class.getCanonicalName());
+        tb.writeFile(servicesFile,
+                     processorName);
+
+        // write out processor source file
+        tb.writeFile(processorName + ".java",
```
jonathan-gibbons, line 106: If you use `writeJavaFiles`, it will infer the filename from the source code.

test/langtools/tools/javac/processing/options/TestNoteOnImplicitProcessing.java
```diff
@@ -0,0 +1,274 @@
...
+                         @Override
+                         public SourceVersion getSupportedSourceVersion() {
+                             return SourceVersion.latest();
+                         }
+                     }
+                     """);
+
+        // Compile the processor
+        new JavacTask(tb)
+            .files(processorName + ".java")
```
jonathan-gibbons, line 135: there's a utility method `ToolBox.findJavaFiles` which is generally useful in this sort of context

jonathan-gibbons (approved):
Approved, with some optional suggestions for stylistic improvement

test/langtools/tools/javac/diags/examples.not-yet.txt
jddarcy: IIRC, it was an end-of-file difference, but it isn't being reflected in the webrev either.

src/jdk.compiler/share/classes/com/sun/tools/javac/main/JavaCompiler.java
jddarcy: PS Will add a clarifying comment that the idiom in question is intentional.

src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties
jddarcy: That observation is true and I'll add -proc:none to the test case.

I'll add a sentence that -proc:none will disable processing.

jonathan-gibbons (approved):

PR #14501, opened 2023-06-15
8306446: java/lang/management/ThreadMXBean/Locks.java transient failures

kevinjwalls:
This test iterates an array of ThreadInfos in a few places (e.g. in the method doCheck()), and needs to tolerate and ignore nulls, in case a thread finishes and the test hits an NPE.

There are other calls like "TM.getThreadInfo(tid).getLockName()" which might often be risky, but if the threads are blocked as they are here, they can't be terminating, so this usage is safe.

The test has additional problems when started in a virtual thread.  ThreadMXBean.getThreadInfo() methods only return a ThreadInfo for platform threads.  The test needs to avoid some checks if mainThread is virtual.

In assertNoLock, it needs to not object to a thread holding a lock on a VirtualThread object is not relevant.
Also the loop in doChecks which follows a chain of locks... This needs to recognise that ForkJoinPool thead is not worth pursuing.  It's not one of the very narrow set of threads this test cares about.

Despite these exclusions, the test does some reasonable verification work when MainThread is virtual.  This test historically cam in with a general "JVM monitoring and management API" change, it is not testing a particular fix.

There's a failure condition in doCheck() which will not make the test fail: if it logs "TEST FAILED" in its final for loop, there is no failure.  Make the loop count the failures, and throw if there are any.

Also, while looking into this...  The variable names in some methods are confusing. In checkBlockedObject(), let's use "threadName" rather than "result" if we are finding a thread name, and let's not reuse the same result variable for a lockName later in the method.

The logs from this test are hard to read and verify, I find it better if the lock objects OBJB and OBJC are of classes other than Object, so you get to read, e.g.:
LockAThread blocked on Locks$ObjectB@4691fdfd
(ObjectB, not just Object).

---------
### Progress

### Issue
 * [JDK-8306446](https://bugs.openjdk.org/browse/JDK-8306446): java/lang/management/ThreadMXBean/Locks.java transient failures (**Bug** - P4)

### Reviewers
 * [Chris Plummer](https://openjdk.org/census#cjplummer) (@plummercj - **Reviewer**) ⚠️ Review applies to [02ba2a31](https://git.openjdk.org/jdk/pull/14501/files/02ba2a31e2bde12560e27768847d381a442eada3)
 * [Serguei Spitsyn](https://openjdk.org/census#sspitsyn) (@sspitsyn - **Reviewer**) ⚠️ Review applies to [02ba2a31](https://git.openjdk.org/jdk/pull/14501/files/02ba2a31e2bde12560e27768847d381a442eada3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14501/head:pull/14501` \
`$ git checkout pull/14501`

`$ git checkout pull/14501` \
`$ git pull https://git.openjdk.org/jdk.git pull/14501/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14501`

`$ git pr show -t 14501`

</details>

<a href="https://git.openjdk.org/jdk/pull/14501.diff">https://git.openjdk.org/jdk/pull/14501.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14501#issuecomment-1597634277)


test/jdk/java/lang/management/ThreadMXBean/Locks.java
```diff
@@ -61,10 +64,12 @@ private static void assertNoLock(Thread t) {
             return;
         }
         String name = t.getName();
+        // If t is Virtual, it will hold a lock on a VirtualThread, which we ignore.
         Optional<ThreadInfo> result = Arrays.stream(
                 TM.getThreadInfo(TM.getAllThreadIds(), true, true))
                                             .filter(Objects::nonNull)
                                             .filter(i -> name.equals(i.getLockOwnerName()))
+                                            .filter(i  -> !i.getLockName().contains("java.lang.VirtualThread"))
```
turbanoff, line 72:
```suggestion
                                            .filter(i -> !i.getLockName().contains("java.lang.VirtualThread"))
```

AlanBateman: ThreadMXBean::getAllThreadsIds returns an array of the thread IDs of platform threads, it doesn't include virtual threads, so I don't know why this filter is added.

kevinjwalls: Whitespace correction, thanks.

kevinjwalls: The filter was because of a failure where it's called on a real thread that happens to own a VirtualThread lock, like:

"ForkJoinPool-1-worker-1" daemon prio=5 Id=32 WAITING on java.lang.VirtualThread@289cb0b4 owned by "MainThread" Id=31

...actually my comment at line 67 isn't exactly correct, will update it.

dholmes-ora:
> In assertNoLock, it needs to not object to a thread holding a lock on a VirtualThread object is not relevant.

Can you elaborate on this please. I'm unclear how the test knows whether a thread may be holding any locks or not, but why is locking a VirtualThread instance a special case??

AlanBateman:
> Can you elaborate on this please. I'm unclear how the test knows whether a thread may be holding any locks or not, but why is locking a VirtualThread instance a special case??

Carriers are reported as waiting, the string representation of the object that they are waiting on is the identity string of the virtual thread that is mounted.

kevinjwalls:
Thanks Alan, yes-
All the TEST really knows is the couple of threads it creates, and the lock objects it creates.  When it says "assert no lock held", it means to assert that none of the locks that it created and cares about are held.

When MainThread is Virtual, as started by jtreg, the test asserts that no locks it cares about are held by MainThread.  When there is an additional locked object (a VirtualThread), that surprises the test. Hence the filter on getLockName() in assertNoLock.

test/jdk/java/lang/management/ThreadMXBean/Locks.java
```diff
@@ -61,10 +64,12 @@ private static void assertNoLock(Thread t) {
             return;
         }
         String name = t.getName();
+        // Carrier Thread can hold a lock on a VirtualThread, which we ignore.
```
plummercj, line 67: This comment's placement can make the user think the whole point of the code that follows is to ignore virtual threads here. Maybe it would be better if placed right before the added filter() call.

test/jdk/java/lang/management/ThreadMXBean/Locks.java
```diff
@@ -362,15 +366,20 @@ public static void main(String args[]) throws Exception {
                 t2.start();
 
                 p.arriveAndAwaitAdvance(); // Phase 1 (blocking)
+
                 assertThreadState(t2, Thread.State.BLOCKED);
-                checkBlockedObject(t2, OBJC, mainThread);
+                if (!mainThread.isVirtual()) {
```
plummercj, line 371: Although the you covered the reason for this exclusion in the PR description, I think a comment is warranted here, and also for line 380 below.

test/jdk/java/lang/management/ThreadMXBean/Locks.java
```diff
@@ -421,19 +433,22 @@ private static void findThreadsBlockedOn(Object o, long[] expectedThreads)
...
+        doCheck(allThreadInfos, lock, expectedThreads);
     }
 
     private static void doCheck(ThreadInfo[] infos, String lock, long[] expectedThreads)
             throws Exception {
         ThreadInfo ownerInfo = null;
         // Find the thread who is blocking on lock
         for (ThreadInfo info : infos) {
+            if (info == null) {
+                continue; // Missing thread, ignore
```
plummercj, line 450: Maybe comment on why it might be missing (thread completed already).

test/jdk/java/lang/management/ThreadMXBean/Locks.java
```diff
@@ -445,11 +460,18 @@ private static void doCheck(ThreadInfo[] infos, String lock, long[] expectedThre
...
         int count = 0;
         threads[count++] = ownerInfo.getThreadId();
         while (ownerInfo.getThreadState() == Thread.State.BLOCKED) {
             ownerInfo = findOwnerInfo(infos, lock);
+            log("ownerInfo = %s", ownerInfo);
+            if (ownerInfo.getThreadName().contains("ForkJoinPool")) {
+                log ("skipping %s", ownerInfo);
+                lock = ownerInfo.getLockName();
+                continue;
+            }
```
plummercj, line 474: What happens if you don't exclude these ForkJoinPool threads? What about the myriad of other threads that the JVM always starts up. Should they not also be skipped if you are going to skip ForkJoinPool threads?

plummercj (changes requested):

test/jdk/java/lang/management/ThreadMXBean/Locks.java
kevinjwalls: Thanks for the feedback, updated.
On this point, the method is only invoked on a very limited set of locks and threads.  It is called to examine LockAThread and LockBThread, and the test main thread (if not virtual).  So it will not (should not!) see a lock owner which is some other thread, it's not a useful generic utility method.

sspitsyn (commented):
This looks pretty good to me.
Is the test not failing anymore with your update?

kevinjwalls:
> This looks pretty good to me. Is the test not failing anymore with your update?

Thanks Serguei, yes it's running reliably, including with virtual threads, with this change (I just ran it, 50 times per platform, to re-check).

test/jdk/java/lang/management/ThreadMXBean/Locks.java
sspitsyn: This is not that easy to make this conclusion.
I'm thinking if placing a comment would help.

sspitsyn (approved):

test/jdk/java/lang/management/ThreadMXBean/Locks.java
kevinjwalls: Sure.  I checked again and without this "skip ForkJoinPool" logic,  JTREG_TEST_THREAD_FACTORY=Virtual causes failures.

This method checks in all the threadInfos, that the specified lock is blocking the expectedThreads (our test threads t1, t2 and mainThread if it's not Virtual)

If it sees a FJP thread that is blocked on a java.lang.VirtualThread, and that's not one of the expected threads, so the counting is wrong and we fail.

I cannot say I like the loop. 8-)

plummercj (approved):

kevinjwalls:
Thanks for the reviews!  I will add one small comment and integrate.

PR #14502, opened 2023-06-15
8310170: Use sp's argument to improve performance of outputStream::indent and remove SP_USE_TABS

jdksjolen:
Hi,

Please consider this small enhancement. `sp` takes a count argument, this was never used by `indent`, let's just use it.

---------
### Progress

### Issue
 * [JDK-8310170](https://bugs.openjdk.org/browse/JDK-8310170): Use sp's argument to improve performance of outputStream::indent and remove SP_USE_TABS (**Enhancement** - P4)

### Reviewers
 * [Aleksey Shipilev](https://openjdk.org/census#shade) (@shipilev - **Reviewer**) ⚠️ Review applies to [d9ee4c2f](https://git.openjdk.org/jdk/pull/14502/files/d9ee4c2f8093848b1b62b4890fa90fdd2530d4c3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14502/head:pull/14502` \
`$ git checkout pull/14502`

`$ git checkout pull/14502` \
`$ git pull https://git.openjdk.org/jdk.git pull/14502/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14502`

`$ git pr show -t 14502`

</details>

<a href="https://git.openjdk.org/jdk/pull/14502.diff">https://git.openjdk.org/jdk/pull/14502.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14502#issuecomment-1594333513)

### DIFF
--- a/src/hotspot/share/utilities/ostream.cpp
+++ b/src/hotspot/share/utilities/ostream.cpp
@@ -186,18 +186,9 @@ void outputStream::put(char ch) {
   write(buf, 1);
 }
 
-#define SP_USE_TABS false
-
 void outputStream::sp(int count) {
   if (count < 0)  return;
-  if (SP_USE_TABS && count >= 8) {
-    int target = position() + count;
-    while (count >= 8) {
-      this->write("\t", 1);
-      count -= 8;
-    }
-    count = target - position();
-  }
+
   while (count > 0) {
     int nw = (count > 8) ? 8 : count;
     this->write("        ", nw);
--- a/src/hotspot/share/utilities/ostream.cpp
+++ b/src/hotspot/share/utilities/ostream.cpp
@@ -257,7 +257,7 @@ void outputStream::date_stamp(bool guard,
 }
 
 outputStream& outputStream::indent() {
-  while (_position < _indentation) sp();
+  sp(_indentation - _position);
   return *this;
 }
 


shipilev (approved):
Looks okay. I thought what would happen if `_indentation < _position`, but `sp()` seems to handle `count < 0` fine.

dholmes-ora (approved):
There's potentially a subtle difference in behaviour here if SP_USE_TABS were ever true - but as it is defined to be false there is no difference in practice. Maybe we should just cull the code related to SP_USE_TABS and not pretend we could do something different here?

Thanks.

jdksjolen:
Hi,

Sorry for the slow response from me on this.

@dholmes-ora, let's remove SP_USE_TABS in this PR. I'll change the ticket description.

src/hotspot/share/utilities/ostream.cpp
```diff
@@ -186,18 +186,9 @@ void outputStream::put(char ch) {
   write(buf, 1);
 }
 
-#define SP_USE_TABS false
-
 void outputStream::sp(int count) {
   if (count < 0)  return;
```
tstuefe, line 190: Is this  needed?

tstuefe (approved):
Good.

src/hotspot/share/utilities/ostream.cpp
jdksjolen: The `count < 0` check? Yes, as Shipilev mentioned:

>Looks okay. I thought what would happen if _indentation < _position, but sp() seems to handle count < 0 fine.

So it's nice that we don't have to care at the call site.

tstuefe: Ah, okay. Missed that.

dholmes-ora (approved):
Looks good.

As an aside I was disappointed to discover that the indentation facility is far less useful than it might be. I had hoped you could set the indentation and then have all following log output indented, until you reset it. But indentation only happens when you explicitly call `indent()` so you can't call e.g. `val->print_on(stream)` and have the caller control indentation (without `print_on` directly supporting it).

jdksjolen:
One structured concurrency test fails on GHA on Linux-x86. That can't be thanks to this patch. Integrating.

/integrate

PR #14504, opened 2023-06-15
8308694: Clarify reversed() default methods' implementation requirements

stuart-marks:
Can I get a preliminary review of the wording for Deque.reversed()? If the text is good, I'll make corresponding changes to the implSpecs of the other reversed() default methods, namely those in List, SortedMap, and SortedSet and then file the CSR.

### DIFF
--- a/src/java.base/share/classes/java/util/Deque.java
+++ b/src/java.base/share/classes/java/util/Deque.java
@@ -617,8 +617,13 @@ public interface Deque<E> extends Queue<E>, SequencedCollection<E> {
      * {@inheritDoc}
      *
      * @implSpec
-     * The implementation in this interface returns an instance of a reverse-ordered
-     * Deque that delegates its operations to this Deque.
+     * The implementation in this interface returns a reverse-ordered Deque
+     * view that delegates its operations to this Deque. Methods that are
+     * order-insensitive delegate to the corresponding method of this Deque.
+     * Methods that are order-sensitive delegate to the corresponding but
+     * inverse-ordered method of this Deque. For example, the {@code getFirst}
+     * method of the returned view delegates to the {@code getLast} method of
+     * this Deque.
      *
      * @return a reverse-ordered view of this collection, as a {@code Deque}
      * @since 21


bplb (commented):
Reads well.

stuart-marks:
@bplb OK thanks. But now that I've posted this, I'm having second thoughts. :-)

The problem is that this spec is trying to be more specific...and the more specific something is, the greater the possibility is of inaccuracy. What I had written is _mostly_ true, but it isn't true for _all_ operations. Some operations in particular such as `addAll` actually have to do rather more complex things than delegating operations to the backing Deque. See the implementation for other examples:

https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/ReverseOrderDequeView.java

I don't want to specify the exact behavior of every method. Issues of a similar magnitude occur with the other reversed views. For example, getting the iterator of a reversed SortedSet involves some heavy lifting:

https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/ReverseOrderSortedSetView.java#L206

and I certainly don't want to specify all of that.

So, while in fact the general sense is true that the operations of the reversed view are all implemented via calls to the backing collection, I need to be deliberately vague in certain respects.

src/java.base/share/classes/java/util/Deque.java
```diff
@@ -617,8 +617,13 @@
...
      * @implSpec
-     * The implementation in this interface returns an instance of a reverse-ordered
-     * Deque that delegates its operations to this Deque.
+     * The implementation in this interface returns a reverse-ordered Deque
+     * view that delegates its operations to this Deque. Methods that are
+     * order-insensitive delegate to the corresponding method of this Deque.
+     * Methods that are order-sensitive delegate to the corresponding but
+     * inverse-ordered method of this Deque. For example, the {@code getFirst}
+     * method of the returned view delegates to the {@code getLast} method of
+     * this Deque.
```
RogerRiggs, line 626: The description is general and using 'For example" avoids having to be specific about every operation.
The term "inverse-ordered" is new to this context and perhaps should be "reverse-ordered".
"inverse" does appear in a few other class descriptions for SequencedXXX but its relation to "reverse" may not be clear.

bchristi-git: Rather than describing how the methods delegate, I think we mostly want to say that 
order-sensitive operations are performed in reverse order.

I agree about, "For example", and I would keep the `getFirst` -> `getLast` part - that's a good, straighforward example.

stuart-marks:
OK please take another look.

RogerRiggs (approved):
Looks good.

stuart-marks:
CSR [JDK-8311845](https://bugs.openjdk.org/browse/JDK-8311845) ready for review.

PR #14506, opened 2023-06-16
8309667: TLS handshake fails because of ConcurrentModificationException in PKCS12KeyStore.engineGetEntry

wangweij:
The `attributes` field inside the `PKCS12KeyStore.Entry` class might be modified and retrieved at the same time. Make it concurrent.

The test uses some reflection to get this field and try updating it in multiple threads.

---------
### Progress

### Issue
 * [JDK-8309667](https://bugs.openjdk.org/browse/JDK-8309667): TLS handshake fails because of ConcurrentModificationException in PKCS12KeyStore.engineGetEntry (**Bug** - P4)

### Reviewers
 * [Xue-Lei Andrew Fan](https://openjdk.org/census#xuelei) (@XueleiFan - **Reviewer**) ⚠️ Review applies to [1d817b6f](https://git.openjdk.org/jdk/pull/14506/files/1d817b6f23c7b5e0b7b0e81b1d966981f319e38c)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14506/head:pull/14506` \
`$ git checkout pull/14506`

`$ git checkout pull/14506` \
`$ git pull https://git.openjdk.org/jdk.git pull/14506/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14506`

`$ git pr show -t 14506`

</details>

<a href="https://git.openjdk.org/jdk/pull/14506.diff">https://git.openjdk.org/jdk/pull/14506.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14506#issuecomment-1593976176)

### DIFF
--- a/src/java.base/share/classes/sun/security/pkcs12/PKCS12KeyStore.java
+++ b/src/java.base/share/classes/sun/security/pkcs12/PKCS12KeyStore.java
@@ -237,7 +237,7 @@ public final class PKCS12KeyStore extends KeyStoreSpi {
             this.keyId = keyId;
             this.alias = alias;
             this.trustedKeyUsage = trustedKeyUsage;
-            this.attributes = new HashSet<>();
+            this.attributes = ConcurrentHashMap.newKeySet();
             if (attributes != null) {
                 this.attributes.addAll(attributes);
             }
@@ -697,7 +697,7 @@ public final class PKCS12KeyStore extends KeyStoreSpi {
                 throw new KeyStoreException("Unsupported Key type");
             }
 
-            entry.attributes = new HashSet<>();
+            entry.attributes = ConcurrentHashMap.newKeySet();
             if (attributes != null) {
                 entry.attributes.addAll(attributes);
             }
@@ -2432,7 +2432,7 @@ public final class PKCS12KeyStore extends KeyStoreSpi {
             String alias = null;
             byte[] keyId = null;
             ObjectIdentifier[] trustedKeyUsage = null;
-            Set<PKCS12Attribute> attributes = ConcurrentHashMap.newKeySet();
+            Set<PKCS12Attribute> attributes = new HashSet<>();
 
             if (attrSet != null) {
                 for (int j = 0; j < attrSet.length; j++) {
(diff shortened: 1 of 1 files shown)


XueleiFan (approved):
Nice catch!

src/java.base/share/classes/sun/security/pkcs12/PKCS12KeyStore.java
```diff
@@ -1434,7 +1435,7 @@ public synchronized void engineSetEntry(String alias, KeyStore.Entry entry,
     private Set<KeyStore.Entry.Attribute> getAttributes(Entry entry) {
 
         if (entry.attributes == null) {
-            entry.attributes = new HashSet<>();
+            entry.attributes = ConcurrentHashMap.newKeySet();
```
dfuch, line 1438: It seems that there is still a risk that two threads may compete here, then you may end up with one thread working with a different copy of the attributes, no longer tied to the entry. E.g:

Thread#1 sees attributes is null, Thread#2 sees attributes is null, both set attributes and only one of them win. The caller in the second thread (that lost) is given the "dangling" map that has not been set, and if it modifies it then modifications will be lost.

Also I see in total four places that do:
```
[entry|this].attributes = new HashSet<>();
```

wouldn't you need to modify all four places in the same manner?

djelinski (changes requested):
Would it be possible to only catch checked exceptions in `X509KeyManagerImpl.getEntry`? Currently the exception is discarded without a trace, which made debugging this a bit of a challenge.

src/java.base/share/classes/sun/security/pkcs12/PKCS12KeyStore.java
dfuch: I wonder if this would work, if you don't want to use final or volatile + CAS:

```
 class Entry { 
     ...
     private Set<KeyStore.Entry.Attribute> createAttributesMap() {
          synchronized (this) {
              var attributes = this.attributes;
              if (attributes == null) {
                  return attributes = this.attributes = ConcurrentHashMap.newKeySet();
              } else return attributes;
           }
       }
       
 }
 
 ...
 
      if (entry.attributes == null) {
          entry.attributes = entry.createAttributesMap();
      }
      
```

PR #14448, opened 2023-06-13
JDK-8309934: Update GitHub Actions to use JDK 17 for building jtreg

sormuras:
Please review this change to use the pre-installed JDK 17 for building jtreg when running on GitHub Actions.

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8309934](https://bugs.openjdk.org/browse/JDK-8309934): Update GitHub Actions to use JDK 17 for building jtreg (**Bug** - P4)

### Reviewers
 * [Erik Joelsson](https://openjdk.org/census#erikj) (@erikj79 - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14448/head:pull/14448` \
`$ git checkout pull/14448`

Update a local copy of the PR: \
`$ git checkout pull/14448` \
`$ git pull https://git.openjdk.org/jdk.git pull/14448/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 14448`

View PR using the GUI difftool: \
`$ git pr show -t 14448`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jdk/pull/14448.diff">https://git.openjdk.org/jdk/pull/14448.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14448#issuecomment-1589352777)

PR #14507, opened 2023-06-16
8310183: Update GitHub Actions to use boot JDK for building jtreg

sormuras:
Please review this change to use the boot JDK for building jtreg when running on GitHub Actions.

This is a best-effort follow-up change to
- #14448
which didn't have the desired results - the `Bad address` error does still appear with using the pre-installed JDKs 11 and 17.

Tests using the boot JDK for building jtreg seem to be successful and more stable.

### DIFF
--- a/.github/actions/get-jtreg/action.yml
+++ b/.github/actions/get-jtreg/action.yml
@@ -57,7 +57,7 @@ runs:
     - name: 'Build JTReg'
       run: |
         # Build JTReg and move files to the proper locations
-        bash make/build.sh --jdk bootjdk/jdk
+        bash make/build.sh
         mkdir ../installed
         mv build/images/jtreg/* ../installed
       working-directory: jtreg/src
--- a/.github/actions/get-jtreg/action.yml
+++ b/.github/actions/get-jtreg/action.yml
@@ -57,7 +57,7 @@ runs:
     - name: 'Build JTReg'
       run: |
         # Build JTReg and move files to the proper locations
-        bash make/build.sh --jdk "$(realpath bootjdk/jdk)
+        bash make/build.sh --jdk "$(realpath bootjdk/jdk)"
         mkdir ../installed
         mv build/images/jtreg/* ../installed
       working-directory: jtreg/src
--- a/.github/actions/get-jtreg/action.yml
+++ b/.github/actions/get-jtreg/action.yml
@@ -54,17 +54,10 @@ runs:
         path: jtreg/src
       if: steps.get-cached-jtreg.outputs.cache-hit != 'true'
 
-    - name: 'Install JDK'
-      uses: actions/setup-java@v3
-      with:
-        distribution: 'oracle'
-        java-version: '17'
-      if: steps.get-cached-jtreg.outputs.cache-hit != 'true'
-
     - name: 'Build JTReg'
       run: |
         # Build JTReg and move files to the proper locations
-        bash make/build.sh --jdk "$JAVA_HOME"
+        bash make/build.sh --jdk "$(realpath bootjdk/jdk)
         mkdir ../installed
         mv build/images/jtreg/* ../installed
       working-directory: jtreg/src
--- a/.github/actions/get-jtreg/action.yml
+++ b/.github/actions/get-jtreg/action.yml
@@ -54,10 +54,17 @@ runs:
         path: jtreg/src
       if: steps.get-cached-jtreg.outputs.cache-hit != 'true'
 
+    - name: 'Install JDK'
+      uses: actions/setup-java@v3
+      with:
+        distribution: 'oracle'
+        java-version: '17'
+      if: steps.get-cached-jtreg.outputs.cache-hit != 'true'
+
     - name: 'Build JTReg'
       run: |
         # Build JTReg and move files to the proper locations
-        bash make/build.sh
+        bash make/build.sh --jdk "$JAVA_HOME"
         mkdir ../installed
         mv build/images/jtreg/* ../installed
       working-directory: jtreg/src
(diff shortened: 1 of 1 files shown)


shipilev (commented):
Wait, so why does it fix the bug? Is it a MSYS path conversion bug? If so, could we "just" `cp` everything from `JAVA_HOME_17_X64` to local path and try that?

My concern with hooking jtreg to boot JDK would require jtreg to be buildable with STSes, like JDK 22 for eventual JDK 23 builds. Is there a plan/commitment for jtreg to do it?

sormuras:
> Wait, so why does it fix the bug? Is it a MSYS path conversion bug?

It does not fix the bug, it works around it. Something (in MSYS) failes to work (calling `javac`) that used to work without problem until some week ago (new version of MSYS?), and looking into the workflow step log, it worked (calling same `javac`) just a second before the failing call. I tried to reproduce it locally and also on a Codespaces instance, but never get the `Bad address` error to appear.

Do you have a better idea how to trace to the underlying issue on GHA?

>  If so, could we "just" `cp` everything from `JAVA_HOME_17_X64` to local path and try that?

I didn't try that out. Might work, too. Yet, there is an installation available at a local path: `bootjdk/jdk`.

> My concern with hooking jtreg to boot JDK would require jtreg to be buildable with STSes, like JDK 22 for eventual JDK 23 builds. Is there a plan/commitment for jtreg to do it?

Copied from https://github.com/openjdk/jtreg/blob/master/doc/building.md

> The script is intended to be run in a Unix-like shell, such as bash on Linux or Mac OS X, or with Cygwin, MSYS2, or WSL on Windows. At a minimum, you must either set the JAVA_HOME environment variable or specify the location of the JDK to be used to build jtreg with the --jdk command-line option. **It must be a recent build of JDK 11 or later.**

So yes, "JDK 11 or later" includes 22 and following releases.

shipilev (approved):
All right then. Having a functioning GHA is better that being perfect in selecting the jtreg build JDK.

jaikiran:
I had a bit more look at this today. This most certainly is some issue in the `msys2` package that gets installed on these VMs. And this appears to be some recent regression. I went back and looked at some of my personal jobs which were passing without issues and I can see that those were on `7efe20baefed56359985e327d329042cde2434ff` commit of `msys2/setup-msys2@v2` github action. It appears that there has been a recent release of this action around 3 weeks back and the GitHub actions jobs are now bringing in `cf96e00c0aab3788743aaf63b64146f0d383cee9` https://github.com/msys2/setup-msys2/commit/cf96e00c0aab3788743aaf63b64146f0d383cee9

The place where this fails when building jtreg, is when launching `javac` with a `CLASSPATH` environment variable set. There's no obvious issues in that variable value - a few commands before, similar values for that variable have worked fine with `javac`. This will need a bit more investigation to understand why `make` fails to launch `javac` with this version of `msys2`. 

For now though, I pinned the `msys2/setup-msys2` action in our JDK's github action's workflow file to the commit that was working and that got my builds past this issue. Without that change I can consistently reproduce the issue every time.

I have created a branch which that change and reverted the commit in this PR to see if the jobs run fine https://github.com/openjdk/jdk/pull/14516.  It's currently in progress https://github.com/jaikiran/jdk/actions/runs/5288890247. If that works, then perhaps we should just use that pinned version of `msys2/setup-msys2`, until we report and have this issue fixed in that package?

jerboaa:
> If that works, then perhaps we should just use that pinned version of `msys2/setup-msys2`, until we report and have this issue fixed in that package?

+1

sormuras:
Thanks for the 🕵️ work, @jaikiran 

I would suggest to keep building jtreg with boot JDK and also to pin the version of `msys2/setup-msys2`. This way, we keep a bit more control over parts being in use by the sanity check workflow.

jonathan-gibbons (commented):
nice

jonathan-gibbons:
To those investigating the msys issue, do we know why and where `CLASSPATH` is being used?   In general, that's a deprecated style of use.

tstuefe:
Are the failing GHAs, eg in this PR: https://github.com/iklam/jdk/actions/runs/5292467822/jobs/9579580675 , related to this patch?

sormuras:
Yes, they seem related.

The error says in occasion: `realpath: bootjdk/jdk: No such file or directory`
Although one step before is has: `mv "$jdk_root"/* bootjdk/jdk/`

https://github.com/iklam/jdk/actions/runs/5292467822/jobs/9579580675#step:3:163

And yes, all worked on this run https://github.com/sormuras/jdk/actions/runs/5280848764 that was checked before merging this change. 🤔

sormuras:
In the light of that, coming back to @jaikiran's suggestion from above might makes more sense:

> I have created a branch which that change and reverted the commit in this PR to see if the jobs run fine https://github.com/openjdk/jdk/pull/14516. It's currently in progress https://github.com/jaikiran/jdk/actions/runs/5288890247. If that works, then perhaps we should just use that pinned version of msys2/setup-msys2, until we report and have this issue fixed in that package?

jonathan-gibbons:
> In the light of that, coming back to @jaikiran's suggestion from above might make more sense:
> 
> > I have created a branch which that change and reverted the commit in this PR to see if the jobs run fine #14516. It's currently in progress https://github.com/jaikiran/jdk/actions/runs/5288890247. If that works, then perhaps we should just use that pinned version of msys2/setup-msys2, until we report and have this issue fixed in that package?

Agreed that pinning msys2 seems the way to go at this time.

jaikiran:
> To those investigating the msys issue, do we know why and where `CLASSPATH` is being used? In general, that's a deprecated style of use.

The make files in jtreg build have a bunch of `CLASSPATH` usages while launching `javac`. For example here:

https://github.com/openjdk/jtreg/blob/master/make/jtreg.gmk#L46
https://github.com/openjdk/jtreg/blob/master/make/jtdiff.gmk#L36

I'll experiment today to move away from `CLASSPATH` environment variable and use javac's classpath option instead.

PR #14508, opened 2023-06-16
8308855: ARM32: TestBooleanVector crashes after 8300257

rwestrel:
In JDK-8300257, I removed by mistake some logic that's required to
compute the loop alignment. This change puts it back.


vnkozlov (approved):
Good.

TobiHartmann (approved):
Looks good.

rwestrel:
@TobiHartmann @vnkozlov thanks for the reviews.

PR #14509, opened 2023-06-16
8310187: Improve Generational ZGC jtreg testing

xmas92:
The current implementation for testing generational ZGC with jtreg is implemented with a filter on the mode flag `ZGenerational`. Because of this only environments which set this flag explicitly will run most of the tests. 
So they get missed in Github Actions and for developers running jtreg locally without supplying the `ZGenerational` flag.

The proposed change here is to introduce two new jtreg requirement properties, `vm.gc.ZGenerational` and `vm.gc.ZSingelgen`. These flags will effectively behave the same as the existing `vm.gc.<GC>` flags but also take the specific ZGC mode in account.

If no gc flags are supplied to jtreg and the `vm.gc.Z` is true (the build includes ZGC) both `vm.gc.ZGenerational` and `vm.gc.ZSinglegen` will be true.

If `-XX:+UseZGC` is supplied then both `vm.gc.ZGenerational` and `vm.gc.ZSinglegen` will also be true.

If `-XX:{+,-}ZGenerational` is supplied then either  `vm.gc.ZGenerational` or `vm.gc.ZSinglegen` be true depending on the flags value.

And if `vm.gc.Z` is false both `vm.gc.ZGenerational` and `vm.gc.ZSinglegen` will be false.

This change also splits the relevant tests into two distinct runs for the two modes. And the respective test ids are set to `ZGenerational` or `ZSinglegen` to make it easier to distinguish the runs.

This also solves the issue that some compiler tests will never run with generational ZGC unless the `TEST_VM_FLAGLESS` is set. This is because the current filter `vm.opt.final.ZGenerational` requires the flag to be explicit, but these compiler tests uses `vm.flagless`. 

The introduction of  `vm.gc.ZGenerational` and `vm.gc.ZSinglegen` harmonizes the way you specify generational / single gen ZGC test with the way it is done for other gcs with `vm.gc.<GC>`

To support this feature the Whitebox API is extended with `isDefaultVMFlag` to enable checking if `ZGenerational` is default or not.

`vm.opt.final.ZGenerational` is still kept because we still have some reasons to filter based on the supplied flags. 
- `test/hotspot/jtreg/gc/cslocker/TestCSLocker.java` is disabled when running with ZGenerational
- `test/jdk/java/lang/ProcessBuilder/CloseRace.java` is ran with a different max heap size for ZGenerational, but it is not the intent to dispatch the test to both G1 and generational ZGC if Generational ZGC is not specified.  

`test/jdk/java/lang/management/MemoryMXBean/MemoryTest.java` was also changed to not filter but instead dispatch. However unsure if this change should be included. The change is isolated in 5c0c0bdab9012a5f35dd4ab3a2847374578c8b9f . It seems more appropriate to actually test the MXBeans by default with no explicit gc, but I am open reverting this part.

The plan is to also backport this to 21 to improve the testing until release. The changes here should fall under `P1–P5 doc/test changes`

Running tests in Oracle CI. (tier1-tier5)
Verified that both `ZGenerational` and `ZSinglegen` tests get run in GHA.
Running tests in builds with `--disable-jvm-feature-zgc` to verify no regression.

---------
### Progress

### Issue
 * [JDK-8310187](https://bugs.openjdk.org/browse/JDK-8310187): Improve Generational ZGC jtreg testing (**Bug** - P4)

### Reviewers
 * [Stefan Karlsson](https://openjdk.org/census#stefank) (@stefank - **Reviewer**) ⚠️ Review applies to [e571b87f](https://git.openjdk.org/jdk/pull/14509/files/e571b87f44c496e55d998d708940d4b4c1d9e63b)
 * [Erik Österlund](https://openjdk.org/census#eosterlund) (@fisk - **Reviewer**) ⚠️ Review applies to [e571b87f](https://git.openjdk.org/jdk/pull/14509/files/e571b87f44c496e55d998d708940d4b4c1d9e63b)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14509/head:pull/14509` \
`$ git checkout pull/14509`

`$ git checkout pull/14509` \
`$ git pull https://git.openjdk.org/jdk.git pull/14509/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14509`

`$ git pr show -t 14509`

</details>

<a href="https://git.openjdk.org/jdk/pull/14509.diff">https://git.openjdk.org/jdk/pull/14509.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14509#issuecomment-1594386417)

### DIFF
--- a/test/jdk/java/lang/Thread/virtual/stress/Skynet.java
+++ b/test/jdk/java/lang/Thread/virtual/stress/Skynet.java
@@ -22,7 +22,7 @@
  */
 
 /*
- * @test
+ * @test id=default
  * @summary Stress test virtual threads with a variation of the Skynet 1M benchmark
  * @requires vm.continuations
  * @requires !vm.debug | vm.gc != "Z"
--- a/test/jdk/java/lang/Thread/virtual/stress/Skynet.java
+++ b/test/jdk/java/lang/Thread/virtual/stress/Skynet.java
@@ -25,14 +25,25 @@
  * @test
  * @summary Stress test virtual threads with a variation of the Skynet 1M benchmark
  * @requires vm.continuations
+ * @requires !vm.debug | vm.gc != "Z"
  * @run main/othervm/timeout=300 -Xmx1g Skynet
  */
 
 /*
- * @test
+ * @test id=ZSinglegen
  * @requires vm.debug == true & vm.continuations
- * @requires vm.gc.Z
+ * @requires vm.gc.ZSinglegen
  * @run main/othervm/timeout=300 -XX:+UnlockDiagnosticVMOptions
+ *     -XX:+UseZGC -XX:-ZGenerational
+ *     -XX:+ZVerifyOops -XX:ZCollectionInterval=0.01 -Xmx1g Skynet
+ */
+
+/*
+ * @test id=ZGenerational
+ * @requires vm.debug == true & vm.continuations
+ * @requires vm.gc.ZGenerational
+ * @run main/othervm/timeout=300 -XX:+UnlockDiagnosticVMOptions
+ *     -XX:+UseZGC -XX:+ZGenerational
  *     -XX:+ZVerifyOops -XX:ZCollectionInterval=0.01 -Xmx1g Skynet
  */
 
--- a/src/hotspot/share/prims/whitebox.cpp
+++ b/src/hotspot/share/prims/whitebox.cpp
@@ -1296,6 +1296,11 @@ WB_ENTRY(jboolean, WB_IsConstantVMFlag(JNIEnv* env, jobject o, jstring name))
   return (flag != nullptr) && flag->is_constant_in_binary();
 WB_END
 
+WB_ENTRY(jboolean, WB_IsDefaultVMFlag(JNIEnv* env, jobject o, jstring name))
+  const JVMFlag* flag = getVMFlag(thread, env, name);
+  return (flag != nullptr) && flag->is_default();
+WB_END
+
 WB_ENTRY(jboolean, WB_IsLockedVMFlag(JNIEnv* env, jobject o, jstring name))
   const JVMFlag* flag = getVMFlag(thread, env, name);
   return (flag != nullptr) && !(flag->is_unlocked() || flag->is_unlocker());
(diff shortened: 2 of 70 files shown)


stefank (approved):
Looks good.

fisk (approved):
Looks good.

dcubed-ojdk:
Hmmm... hopefully `vm.gc.ZSingelgen` is really `vm.gc.ZSinglegen`. If so you might
want to fix the PR description.

test/jdk/java/lang/Thread/virtual/stress/Skynet.java
```diff
@@ -25,14 +25,25 @@
  * @test
```
AlanBateman, line 25: Would you mind adding "id=default" to this test? If a test has several `@test` tags, some with an identifier, and some without, then it can be tricky to understand test selection and exclusion.

xmas92:
> Hmmm... hopefully `vm.gc.ZSingelgen` is really `vm.gc.ZSinglegen`. If so you might want to fix the PR description.

What part are you worried about? When would `vm.gc.ZSingelgen` be invalid? 

Also what part of the PR requires fixing?

AlanBateman:
> > Hmmm... hopefully `vm.gc.ZSingelgen` is really `vm.gc.ZSinglegen`. If so you might want to fix the PR description.
> 
> What part are you worried about? When would `vm.gc.ZSingelgen` be invalid?
> 
> Also what part of the PR requires fixing?

I assume Dan is pointing out the typo in the PR description ("e" and "l" have been transposed).

tstuefe (approved):
I'm not the most qualified for ZGC, but I looked this over with a mind on the planned 21 backport. I wanted to check if any of the existing legacy ZGC tests had been changed, either accidentally or deliberately. This seems not be the case.

How much time do we spend now in GHAs on the additional Zgenerational tests? In any case, this makes sense.

Cheers, Thomas

xmas92:
> How much time do we spend now in GHAs on the additional Zgenerational tests? In any case, this makes sense.

This adds around 40 more tests per platform. The runtime of GHA seems to have a lot of run to run variance. Most new tests are in the `hs/tier1 gc` which has a lower runtime than a few other jobs. So until the heavies jobs gets split into multiple jobs,   this should have a negligible affect the overall GHA runtime

xmas92:
Thanks for all the reviews.
/integrate

prrace:
I have no issues with the enhanced Java 2D tests to cover both generational and original ZGC.
But I'll leave it to hotspot reviews to formally approve the fix since those tests are such a tiny part of this.

PR #14510, opened 2023-06-16
8310192: RISC-V: Merge vector min & max instructs with similar match rules

zifeihan:
Hi, We merged vector min and max instructions with similar matching rules in this PR, and modified some comments of the copy_memory function in stubGenerator_riscv.cpp.

Please take a look and have some reviews. Thanks a lot.

## Testing:
- [x] Tier1 tests (release)
- [x] Tier2 tests (release)
- [x] Tier3 tests (release)
- [x] test/jdk/jdk/incubator/vector (fastdebug)

### DIFF
--- a/src/hotspot/cpu/riscv/stubGenerator_riscv.cpp
+++ b/src/hotspot/cpu/riscv/stubGenerator_riscv.cpp
@@ -851,47 +851,6 @@ class StubGenerator: public StubCodeGenerator {
 
   Label copy_f, copy_b;
 
-  // All-singing all-dancing memory copy.
-  //
-  // Copy count units of memory from s to d.  The size of a unit is
-  // step, which can be positive or negative depending on the direction
-  // of copy.  If is_aligned is false, we align the source address.
-  //
-  /*
-   * if (is_aligned) {
-   *   if (count >= 32)
-   *     goto copy32_loop;
-   *   if (count >= 8)
-   *     goto copy8_loop;
-   *   goto copy_small;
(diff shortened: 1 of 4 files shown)


src/hotspot/cpu/riscv/stubGenerator_riscv.cpp
```diff
@@ -851,47 +851,6 @@ class StubGenerator: public StubCodeGenerator {
 
   Label copy_f, copy_b;
 
-  // All-singing all-dancing memory copy.
-  //
-  // Copy count units of memory from s to d.  The size of a unit is
-  // step, which can be positive or negative depending on the direction
-  // of copy.  If is_aligned is false, we align the source address.
-  //
-  /*
```
luhenry, line 860: Why the removal of that comment?

luhenry (approved):

src/hotspot/cpu/riscv/stubGenerator_riscv.cpp
zifeihan: > Why the removal of that comment?

Thanks for the review, this comment does not match the actual assembly logic, where the align logic of source and dst address  is judged first, not only align the source address. 
https://github.com/openjdk/jdk/blob/b412fc79c3c2548df10918090beedaf6b2d08d96/src/hotspot/cpu/riscv/stubGenerator_riscv.cpp#L987-L993

RealFYang (approved):

PR #14517, opened 2023-06-16
8302865: Illegal bytecode for break from if with instanceof pattern matching condition

lahodaj:
Please see the CSR for the exact wording of the specification change:
https://bugs.openjdk.org/browse/JDK-8310016

There are two changes to flow scoping under this PR:

a) for code like:

```
X: if (!(o instanceof String s)) {
    break X;
}
System.err.println(s);
```

`s` won't be in the scope for `System.err.println` anymore (as that leads to an invalid bytecode anyway). I was experimenting with several possibilities to fix this, and the final proposal is (when we are about to add some bindings to the scope), look for an immediately enclosing labelled statement(s), and there's some, look for a break to this statement(s).

b) for code like:

```
X: {
         while (!(o instanceof String s)) {
             break X;
         }
         System.err.println(s);
}
```

javac implements a strict interpretation of the specification, which does not include `s` in the scope for the `System.err.println` (due to the `break` out of the `while`). The specification is clarified, so that only breaks to the loop are considered (together with breaks to the enclosing labelled statements), so flow scoping handling for loops is adjusted to only consider breaks to the loop.

---------
### Progress
- [x] Change requires CSR request [JDK-8310016](https://bugs.openjdk.org/browse/JDK-8310016) to be approved

### Issues
 * [JDK-8302865](https://bugs.openjdk.org/browse/JDK-8302865): Illegal bytecode for break from if with instanceof pattern matching condition (**Bug** - P3)
 * [JDK-8310016](https://bugs.openjdk.org/browse/JDK-8310016): Illegal bytecode for break from if with instanceof pattern matching condition (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14517/head:pull/14517` \
`$ git checkout pull/14517`

`$ git checkout pull/14517` \
`$ git pull https://git.openjdk.org/jdk.git pull/14517/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14517`

`$ git pr show -t 14517`

</details>

<a href="https://git.openjdk.org/jdk/pull/14517.diff">https://git.openjdk.org/jdk/pull/14517.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14517#issuecomment-1594657649)

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
@@ -1447,12 +1447,7 @@ public class Attr extends JCTree.Visitor {
     public void visitDoLoop(JCDoWhileLoop tree) {
         attribStat(tree.body, env.dup(tree));
         attribExpr(tree.cond, env, syms.booleanType);
-        if (!breaksOutOf(tree, tree.body)) {
-            //include condition's body when false after the while, if cannot get out of the loop
-            MatchBindings condBindings = matchBindings;
-            condBindings.bindingsWhenFalse.forEach(env.info.scope::enter);
-            condBindings.bindingsWhenFalse.forEach(BindingSymbol::preserveBinding);
-        }
+        handleLoopConditionBindings(matchBindings, tree, tree.body);
         result = null;
     }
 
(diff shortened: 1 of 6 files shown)


vicente-romero-oracle (approved):
looks sensible

PR #14520, opened 2023-06-16
8310160: Make GC APIs for handling archive heap objects agnostic of GC policy

ashu-mehra:
This PR adds GC APIs to be implemented by all collectors for proper handling of archive heap space. Currently only G1 is updated to use these APIs which just involves renaming the existing G1 APIs.
In addition to that filemap.cpp is updated to replace calls to `G1CollectedHeap::heap()` with `Universe::heap()` to avoid G1 specific code as much as possible.

At many places in filemap.cpp heap range is requested from GC. All collectors except ZGC have contiguous heap and set `CollectedHeap::_reserved` to the heap range, so it can be easily exposed to the CDS code. This is done in this patch through `CollectedHeap::reserved` API. But for ZGC the heap can be discontiguous which makes it tricky to expose the heap range.
Another point to note is that most of the usage for heap range is for logging purpose, but there is one place where it is used for setting the `mapping_offset` in `FileMapInfo::write_region()` based on the heap start. So purely based on the functional requirement, we only need the heap start address, not the range.

To keep things simple and considering ZGC does not currently support archive heap, i refrained from tackling the issue of discontiguous heap range in this PR.

---------
### Progress

### Issue
 * [JDK-8310160](https://bugs.openjdk.org/browse/JDK-8310160): Make GC APIs for handling archive heap objects agnostic of GC policy (**Sub-task** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14520/head:pull/14520` \
`$ git checkout pull/14520`

`$ git checkout pull/14520` \
`$ git pull https://git.openjdk.org/jdk.git pull/14520/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14520`

`$ git pr show -t 14520`

</details>

<a href="https://git.openjdk.org/jdk/pull/14520.diff">https://git.openjdk.org/jdk/pull/14520.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14520#issuecomment-1594794285)

### DIFF
--- a/src/hotspot/share/cds/filemap.cpp
+++ b/src/hotspot/share/cds/filemap.cpp
@@ -69,7 +69,6 @@
 #include "utilities/defaultStream.hpp"
 #include "utilities/ostream.hpp"
 #if INCLUDE_G1GC
-#include "gc/g1/g1CollectedHeap.hpp"
 #include "gc/g1/heapRegion.hpp"
 #endif
 
@@ -211,12 +210,11 @@ void FileMapHeader::populate(FileMapInfo *info, size_t core_region_alignment,
       _heap_begin = CompressedOops::begin();
       _heap_end = CompressedOops::end();
     } else {
-#if INCLUDE_G1GC
-      address start = (address)G1CollectedHeap::heap()->reserved().start();
-      address end = (address)G1CollectedHeap::heap()->reserved().end();
+      assert(UseG1GC, "used by G1 only");
+      address start = (address)Universe::heap()->reserved().start();
+      address end = (address)Universe::heap()->reserved().end();
       _heap_begin = HeapShared::to_requested_address(start);
       _heap_end = HeapShared::to_requested_address(end);
-#endif
     }
   }
   _compressed_oops = UseCompressedOops;
@@ -1569,9 +1567,8 @@ void FileMapInfo::write_region(int region, char* base, size_t size,
       mapping_offset = (size_t)((address)requested_base - CompressedOops::base());
       assert((mapping_offset >> CompressedOops::shift()) << CompressedOops::shift() == mapping_offset, "must be");
     } else {
-#if INCLUDE_G1GC
-      mapping_offset = requested_base - (char*)G1CollectedHeap::heap()->reserved().start();
-#endif
+      assert(UseG1GC, "used by G1 only");
+      mapping_offset = requested_base - (char*)Universe::heap()->reserved().start();
     }
 #endif // INCLUDE_CDS_JAVA_HEAP
   } else {
@@ -2025,17 +2022,17 @@ bool FileMapInfo::can_use_heap_region() {
   log_info(cds)("    heap range = [" PTR_FORMAT " - "  PTR_FORMAT "]",
                 p2i(header()->heap_begin()), p2i(header()->heap_end()));
 
-  log_info(cds)("The current max heap size = " SIZE_FORMAT "M, HeapRegion::GrainBytes = " SIZE_FORMAT,
-                MaxHeapSize/M, HeapRegion::GrainBytes);
+  log_info(cds)("The current max heap size = " SIZE_FORMAT "M", MaxHeapSize/M);
+  if (UseG1GC) {
+    log_info(cds)("HeapRegion::GrainBytes = " SIZE_FORMAT, HeapRegion::GrainBytes);
+  }
   log_info(cds)("    narrow_klass_base = " PTR_FORMAT ", narrow_klass_shift = %d",
                 p2i(CompressedKlassPointers::base()), CompressedKlassPointers::shift());
   log_info(cds)("    narrow_oop_mode = %d, narrow_oop_base = " PTR_FORMAT ", narrow_oop_shift = %d",
                 CompressedOops::mode(), p2i(CompressedOops::base()), CompressedOops::shift());
   log_info(cds)("    heap range = [" PTR_FORMAT " - "  PTR_FORMAT "]",
-                UseCompressedOops ? p2i(CompressedOops::begin()) :
-                                    UseG1GC ? p2i((address)G1CollectedHeap::heap()->reserved().start()) : 0L,
-                UseCompressedOops ? p2i(CompressedOops::end()) :
-                                    UseG1GC ? p2i((address)G1CollectedHeap::heap()->reserved().end()) : 0L);
+                UseCompressedOops ? p2i(CompressedOops::begin()) : p2i((address)Universe::heap()->reserved().start()),
+                UseCompressedOops ? p2i(CompressedOops::end()) : p2i((address)Universe::heap()->reserved().end()));
 
   if (narrow_klass_base() != CompressedKlassPointers::base() ||
       narrow_klass_shift() != CompressedKlassPointers::shift()) {
@@ -2094,14 +2091,15 @@ bool FileMapInfo::map_heap_region() {
 
     // Make sure we map at the very top of the heap - see comments in
     // init_heap_region_relocation().
-    MemRegion heap_range = G1CollectedHeap::heap()->reserved();
+    MemRegion heap_range = Universe::heap()->reserved();
     assert(heap_range.contains(_mapped_heap_memregion), "must be");
 
-    address heap_end = (address)heap_range.end();
-    address mapped_heap_region_end = (address)_mapped_heap_memregion.end();
-    assert(heap_end >= mapped_heap_region_end, "must be");
-    assert(heap_end - mapped_heap_region_end < (intx)(HeapRegion::GrainBytes),
-           "must be at the top of the heap to avoid fragmentation");
+    if (UseG1GC) {
+      address heap_end = (address)heap_range.end();
+      address mapped_heap_region_end = (address)_mapped_heap_memregion.end();
+      assert(heap_end - mapped_heap_region_end < (intx)(HeapRegion::GrainBytes),
+             "must be at the top of the heap to avoid fragmentation");
+    }
 #endif
 
     ArchiveHeapLoader::set_mapped();
@@ -2112,7 +2110,7 @@ bool FileMapInfo::map_heap_region() {
 }
 
 bool FileMapInfo::map_heap_region_impl() {
-  assert(UseG1GC, "the following code assumes G1");
+  assert(UseG1GC, "used by G1 only");
 
   FileMapRegion* r = region_at(MetaspaceShared::hp);
   size_t size = r->used();
@@ -2126,7 +2124,7 @@ bool FileMapInfo::map_heap_region_impl() {
   log_info(cds)("Preferred address to map heap data (to avoid relocation) is " INTPTR_FORMAT, p2i(requested_start));
 
   // allocate from java heap
-  HeapWord* start = G1CollectedHeap::heap()->alloc_archive_region(word_size, (HeapWord*)requested_start);
+  HeapWord* start = Universe::heap()->alloc_archive_space(word_size, (HeapWord*)requested_start);
   if (start == nullptr) {
     log_info(cds)("UseSharedSpaces: Unable to allocate java heap region for archive heap.");
     return false;
@@ -2141,7 +2139,7 @@ bool FileMapInfo::map_heap_region_impl() {
                               addr, _mapped_heap_memregion.byte_size(), r->read_only(),
                               r->allow_exec());
   if (base == nullptr || base != addr) {
-    dealloc_heap_region();
+    handle_failed_mapping();
     log_info(cds)("UseSharedSpaces: Unable to map at required address in java heap. "
                   INTPTR_FORMAT ", size = " SIZE_FORMAT " bytes",
                   p2i(addr), _mapped_heap_memregion.byte_size());
@@ -2150,7 +2148,7 @@ bool FileMapInfo::map_heap_region_impl() {
 
   r->set_mapped_base(base);
   if (VerifySharedSpaces && !r->check_region_crc()) {
-    dealloc_heap_region();
+    handle_failed_mapping();
     log_info(cds)("UseSharedSpaces: mapped heap region is corrupt");
     return false;
   }
@@ -2173,7 +2171,7 @@ bool FileMapInfo::map_heap_region_impl() {
     char* bitmap_base = map_bitmap_region();
     if (bitmap_base == NULL) {
       log_info(cds)("CDS heap cannot be used because bitmap region cannot be mapped");
-      dealloc_heap_region();
+      handle_failed_mapping();
       unmap_region(MetaspaceShared::hp);
       _heap_pointers_need_patching = false;
       return false;
@@ -2210,18 +2208,13 @@ void FileMapInfo::patch_heap_embedded_pointers() {
 void FileMapInfo::fixup_mapped_heap_region() {
   if (ArchiveHeapLoader::is_mapped()) {
     assert(!_mapped_heap_memregion.is_empty(), "sanity");
-
-    // Populate the archive regions' G1BlockOffsetTableParts. That ensures
-    // fast G1BlockOffsetTablePart::block_start operations for any given address
-    // within the archive regions when trying to find start of an object
-    // (e.g. during card table scanning).
-    G1CollectedHeap::heap()->populate_archive_regions_bot_part(_mapped_heap_memregion);
+    Universe::heap()->fixup_archive_space(_mapped_heap_memregion);
   }
 }
 
 // dealloc the archive regions from java heap
-void FileMapInfo::dealloc_heap_region() {
-  G1CollectedHeap::heap()->dealloc_archive_regions(_mapped_heap_memregion);
+void FileMapInfo::handle_failed_mapping() {
+  Universe::heap()->handle_archive_space_failure(_mapped_heap_memregion);
 }
 #endif // INCLUDE_CDS_JAVA_HEAP
 
--- a/src/hotspot/share/cds/filemap.cpp
+++ b/src/hotspot/share/cds/filemap.cpp
@@ -211,8 +211,8 @@ void FileMapHeader::populate(FileMapInfo *info, size_t core_region_alignment,
       _heap_end = CompressedOops::end();
     } else {
       assert(UseG1GC, "used by G1 only");
-      address start = (address)Universe::heap()->reserved().start();
-      address end = (address)Universe::heap()->reserved().end();
+      address start = (address)Universe::heap()->reserved_region().start();
+      address end = (address)Universe::heap()->reserved_region().end();
       _heap_begin = HeapShared::to_requested_address(start);
       _heap_end = HeapShared::to_requested_address(end);
     }
@@ -1568,7 +1568,7 @@ void FileMapInfo::write_region(int region, char* base, size_t size,
       assert((mapping_offset >> CompressedOops::shift()) << CompressedOops::shift() == mapping_offset, "must be");
     } else {
       assert(UseG1GC, "used by G1 only");
-      mapping_offset = requested_base - (char*)Universe::heap()->reserved().start();
+      mapping_offset = requested_base - (char*)Universe::heap()->reserved_region().start();
     }
 #endif // INCLUDE_CDS_JAVA_HEAP
   } else {
@@ -2031,8 +2031,8 @@ bool FileMapInfo::can_use_heap_region() {
   log_info(cds)("    narrow_oop_mode = %d, narrow_oop_base = " PTR_FORMAT ", narrow_oop_shift = %d",
                 CompressedOops::mode(), p2i(CompressedOops::base()), CompressedOops::shift());
   log_info(cds)("    heap range = [" PTR_FORMAT " - "  PTR_FORMAT "]",
-                UseCompressedOops ? p2i(CompressedOops::begin()) : p2i((address)Universe::heap()->reserved().start()),
-                UseCompressedOops ? p2i(CompressedOops::end()) : p2i((address)Universe::heap()->reserved().end()));
+                UseCompressedOops ? p2i(CompressedOops::begin()) : p2i((address)Universe::heap()->reserved_region().start()),
+                UseCompressedOops ? p2i(CompressedOops::end()) : p2i((address)Universe::heap()->reserved_region().end()));
 
   if (narrow_klass_base() != CompressedKlassPointers::base() ||
       narrow_klass_shift() != CompressedKlassPointers::shift()) {
@@ -2091,7 +2091,7 @@ bool FileMapInfo::map_heap_region() {
 
     // Make sure we map at the very top of the heap - see comments in
     // init_heap_region_relocation().
-    MemRegion heap_range = Universe::heap()->reserved();
+    MemRegion heap_range = Universe::heap()->reserved_region();
     assert(heap_range.contains(_mapped_heap_memregion), "must be");
 
     if (UseG1GC) {
--- a/src/hotspot/share/cds/filemap.cpp
+++ b/src/hotspot/share/cds/filemap.cpp
@@ -2124,7 +2124,7 @@ bool FileMapInfo::map_heap_region_impl() {
   log_info(cds)("Preferred address to map heap data (to avoid relocation) is " INTPTR_FORMAT, p2i(requested_start));
 
   // allocate from java heap
-  HeapWord* start = Universe::heap()->alloc_archive_space(word_size, (HeapWord*)requested_start);
+  HeapWord* start = Universe::heap()->allocate_archive_space(word_size, (HeapWord*)requested_start);
   if (start == nullptr) {
     log_info(cds)("UseSharedSpaces: Unable to allocate java heap region for archive heap.");
     return false;
--- a/src/hotspot/share/cds/filemap.cpp
+++ b/src/hotspot/share/cds/filemap.cpp
@@ -1567,7 +1567,6 @@ void FileMapInfo::write_region(int region, char* base, size_t size,
       mapping_offset = (size_t)((address)requested_base - CompressedOops::base());
       assert((mapping_offset >> CompressedOops::shift()) << CompressedOops::shift() == mapping_offset, "must be");
     } else {
-      assert(UseG1GC, "used by G1 only");
       mapping_offset = requested_base - (char*)Universe::heap()->reserved_region().start();
     }
 #endif // INCLUDE_CDS_JAVA_HEAP
@@ -2094,12 +2093,10 @@ bool FileMapInfo::map_heap_region() {
     MemRegion heap_range = Universe::heap()->reserved_region();
     assert(heap_range.contains(_mapped_heap_memregion), "must be");
 
-    if (UseG1GC) {
-      address heap_end = (address)heap_range.end();
-      address mapped_heap_region_end = (address)_mapped_heap_memregion.end();
-      assert(heap_end - mapped_heap_region_end < (intx)(HeapRegion::GrainBytes),
-             "must be at the top of the heap to avoid fragmentation");
-    }
+    address heap_end = (address)heap_range.end();
+    address mapped_heap_region_end = (address)_mapped_heap_memregion.end();
+    assert(heap_end - mapped_heap_region_end < (intx)(HeapRegion::GrainBytes),
+           "must be at the top of the heap to avoid fragmentation");
 #endif
 
     ArchiveHeapLoader::set_mapped();
--- a/src/hotspot/share/gc/shared/collectedHeap.hpp
+++ b/src/hotspot/share/gc/shared/collectedHeap.hpp
@@ -207,7 +207,7 @@ class CollectedHeap : public CHeapObj<mtGC> {
   }
 
  public:
-  MemRegion reserved() const {
+  MemRegion reserved_region() const {
     return _reserved;
   }
 
--- a/src/hotspot/share/gc/shared/collectedHeap.hpp
+++ b/src/hotspot/share/gc/shared/collectedHeap.hpp
@@ -207,6 +207,9 @@ class CollectedHeap : public CHeapObj<mtGC> {
   }
 
  public:
+  MemRegion reserved() const {
+    return _reserved;
+  }
 
   static inline size_t filler_array_max_size() {
     return _filler_array_max_size;
@@ -517,6 +520,20 @@ class CollectedHeap : public CHeapObj<mtGC> {
   virtual bool can_load_archived_objects() const { return false; }
   virtual HeapWord* allocate_loaded_archive_space(size_t size) { return nullptr; }
   virtual void complete_loaded_archive_space(MemRegion archive_space) { }
+
+  // Commit the heap memory for CDS heap archive area according to the requested size.
+  // Preferred address is treated as a hint for the location of the archive area in the heap.
+  // The returned memory range may or may start at the preferred address.
+  // Return MemRegion corresponding to the commited heap memory.
+  virtual HeapWord* alloc_archive_space(size_t word_size, HeapWord* preferred_addr) { return nullptr; }
+
+  // This must be called after alloc_archive_space, and after class loading has occurred.
+  // GC policy can take necessary actions to make the archive space usable,
+  // such as inserting filler objects to make it parseable, or updating block offset table.
+  virtual void fixup_archive_space(MemRegion range) { return; }
+
+  // Handle any failure in loading the CDS archive area.
+  virtual void handle_archive_space_failure(MemRegion range) { return; }
 
   virtual bool is_oop(oop object) const;
   // Non product verification and debugging.
--- a/src/hotspot/share/gc/shared/collectedHeap.hpp
+++ b/src/hotspot/share/gc/shared/collectedHeap.hpp
@@ -525,9 +525,9 @@ class CollectedHeap : public CHeapObj<mtGC> {
   // Preferred address is treated as a hint for the location of the archive area in the heap.
   // The returned memory range may or may start at the preferred address.
   // Return MemRegion corresponding to the commited heap memory.
-  virtual HeapWord* alloc_archive_space(size_t word_size, HeapWord* preferred_addr) { return nullptr; }
+  virtual HeapWord* allocate_archive_space(size_t word_size, HeapWord* preferred_addr) { return nullptr; }
 
-  // This must be called after alloc_archive_space, and after class loading has occurred.
+  // This must be called after allocate_archive_space, and after class loading has occurred.
   // GC policy can take necessary actions to make the archive space usable,
   // such as inserting filler objects to make it parseable, or updating block offset table.
   virtual void fixup_archive_space(MemRegion range) { return; }
--- a/src/hotspot/share/cds/archiveHeapWriter.cpp
+++ b/src/hotspot/share/cds/archiveHeapWriter.cpp
@@ -333,7 +333,7 @@ size_t ArchiveHeapWriter::copy_one_source_obj_to_buffer(oop src_obj) {
 void ArchiveHeapWriter::set_requested_address(ArchiveHeapInfo* info) {
   assert(!info->is_used(), "only set once");
   assert(UseG1GC, "must be");
-  address heap_end = (address)G1CollectedHeap::heap()->reserved().end();
+  address heap_end = (address)G1CollectedHeap::heap()->reserved_region().end();
   log_info(cds, heap)("Heap end = %p", heap_end);
 
   size_t heap_region_byte_size = _buffer_used;
--- a/src/hotspot/share/cds/filemap.hpp
+++ b/src/hotspot/share/cds/filemap.hpp
@@ -548,7 +548,7 @@ public:
   bool  validate_boot_class_paths() NOT_CDS_RETURN_(false);
   bool  validate_app_class_paths(int shared_app_paths_len) NOT_CDS_RETURN_(false);
   bool  map_heap_region_impl() NOT_CDS_JAVA_HEAP_RETURN_(false);
-  void  dealloc_heap_region() NOT_CDS_JAVA_HEAP_RETURN;
+  void  handle_failed_mapping() NOT_CDS_JAVA_HEAP_RETURN;
   bool  can_use_heap_region();
   bool  load_heap_region() NOT_CDS_JAVA_HEAP_RETURN_(false);
   bool  map_heap_region() NOT_CDS_JAVA_HEAP_RETURN_(false);
--- a/src/hotspot/share/cds/heapShared.cpp
+++ b/src/hotspot/share/cds/heapShared.cpp
@@ -532,9 +532,9 @@ void HeapShared::archive_objects(ArchiveHeapInfo *heap_info) {
 
     log_info(cds)("Heap range = [" PTR_FORMAT " - "  PTR_FORMAT "]",
                    UseCompressedOops ? p2i(CompressedOops::begin()) :
-                                       p2i((address)G1CollectedHeap::heap()->reserved().start()),
+                                       p2i((address)G1CollectedHeap::heap()->reserved_region().start()),
                    UseCompressedOops ? p2i(CompressedOops::end()) :
-                                       p2i((address)G1CollectedHeap::heap()->reserved().end()));
+                                       p2i((address)G1CollectedHeap::heap()->reserved_region().end()));
     copy_objects();
 
     CDSHeapVerifier::verify();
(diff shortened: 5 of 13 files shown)


ashu-mehra:
fyi - @iklam @tschatzl

src/hotspot/share/cds/filemap.cpp
```diff
@@ -2141,7 +2139,7 @@ bool FileMapInfo::map_heap_region_impl() {
                               addr, _mapped_heap_memregion.byte_size(), r->read_only(),
                               r->allow_exec());
   if (base == nullptr || base != addr) {
-    dealloc_heap_region();
+    handle_failed_mapping();
```
iklam, line 2142: `handle_failed_mapping()` is called for several reasons where we are not able to use the archive heap:
- mmap failed for the `hp` or `bm` regions
- CRC check failure

So I think the wording of "mapping" is too specific and doesn't cover all cases (and also precludes the possibility of other failures in the future.

Also "handle" could imply a possibility of recovering from the failure, which isn't the case.

I think a more neutral name would be better. Maybe `release_archive_space()` in contrast of `alloc_archive_space` (or even `allocate_archive_space`, as we prefer proper words than short hands).

src/hotspot/share/cds/filemap.cpp
```diff
@@ -1569,9 +1567,8 @@ void FileMapInfo::write_region(int region, char* base, size_t size,
       mapping_offset = (size_t)((address)requested_base - CompressedOops::base());
       assert((mapping_offset >> CompressedOops::shift()) << CompressedOops::shift() == mapping_offset, "must be");
     } else {
-#if INCLUDE_G1GC
-      mapping_offset = requested_base - (char*)G1CollectedHeap::heap()->reserved().start();
-#endif
+      assert(UseG1GC, "used by G1 only");
```
tstuefe, line 1570: Is this assert needed? If yes, it should be guarding the whole HeapShared::is_heap_region() branch, not just this else branch. OTOH we already assert "HeapShared::can_write()", which is a stronger assert.

src/hotspot/share/gc/shared/collectedHeap.hpp
```diff
@@ -518,6 +521,20 @@ class CollectedHeap : public CHeapObj<mtGC> {
...
+  // Return MemRegion corresponding to the commited heap memory.
+  virtual HeapWord* alloc_archive_space(size_t word_size, HeapWord* preferred_addr) { return nullptr; }
+
+  // This must be called after alloc_archive_space, and after class loading has occurred.
+  // GC policy can take necessary actions to make the archive space usable,
+  // such as inserting filler objects to make it parseable, or updating block offset table.
+  virtual void fixup_archive_space(MemRegion range) { return; }
+
+  // Handle any failure in loading the CDS archive area.
+  virtual void handle_archive_space_failure(MemRegion range) { return; }
```
tstuefe, line 536: For both `fixup_archive_space` and `handle_archive_space_failure`: Why do we need to pass in the MemRegion? The only valid argument in both cases is the MemRegion that had been previously established by a call to `alloc_archive_space`. 

Why not let CollectedHeap remember the MemRegion established by `alloc_archive_space`, thereby also guaranteeing that `alloc_archive_space` can only be called once, and refer to that in the followup calls?

src/hotspot/share/gc/shared/collectedHeap.hpp
```diff
@@ -518,6 +521,20 @@ class CollectedHeap : public CHeapObj<mtGC> {
...
+  // Commit the heap memory for CDS heap archive area according to the requested size.
+  // Preferred address is treated as a hint for the location of the archive area in the heap.
+  // The returned memory range may or may start at the preferred address.
+  // Return MemRegion corresponding to the commited heap memory.
+  virtual HeapWord* alloc_archive_space(size_t word_size, HeapWord* preferred_addr) { return nullptr; }
+
+  // This must be called after alloc_archive_space, and after class loading has occurred.
+  // GC policy can take necessary actions to make the archive space usable,
+  // such as inserting filler objects to make it parseable, or updating block offset table.
+  virtual void fixup_archive_space(MemRegion range) { return; }
```
tstuefe, line 533: Small nit about the naming of these callbacks. "fixup_archive_space" implies knowledge about what the collector does, as does "handle_archive_space_failure". But you want to abstract away collector work from CDS. @iklam already remarked on "handle..". 

I see these methods both more as callbacks, and as such they could be named after the point at which they are called, e.g. "on_cds_post_load" and "on_cds_load_error". Or similar. Named after when they are called, but assuming no knowledge of what the collector does at that point, if anything at all.

Up to you. This is very bikesheddy, I'm also fine with the names as they are.

src/hotspot/share/cds/filemap.cpp
tstuefe: > I think a more neutral name would be better. Maybe `release_archive_space()` in contrast of `alloc_archive_space` (or even `allocate_archive_space`, as we prefer proper words than short hands).

"on_load_error"?

src/hotspot/share/cds/filemap.cpp
```diff
@@ -2094,14 +2091,15 @@ bool FileMapInfo::map_heap_region() {
...
-    MemRegion heap_range = G1CollectedHeap::heap()->reserved();
+    MemRegion heap_range = Universe::heap()->reserved();
     assert(heap_range.contains(_mapped_heap_memregion), "must be");
 
-    address heap_end = (address)heap_range.end();
-    address mapped_heap_region_end = (address)_mapped_heap_memregion.end();
-    assert(heap_end >= mapped_heap_region_end, "must be");
-    assert(heap_end - mapped_heap_region_end < (intx)(HeapRegion::GrainBytes),
-           "must be at the top of the heap to avoid fragmentation");
+    if (UseG1GC) {
```
tstuefe, line 2097: This condition is odd. We only call FileMapInfo::map_heap_region() for G1, right? Maybe just assert at the start of the function for `ArchiveHeapLoader::can_map()` instead?

src/hotspot/share/cds/filemap.cpp
ashu-mehra: Looking at it again, you are right. I don't think we need this assert. I will remove it.

src/hotspot/share/gc/shared/collectedHeap.hpp
```diff
@@ -207,6 +207,9 @@ class CollectedHeap : public CHeapObj<mtGC> {
   }
 
  public:
+  MemRegion reserved() const {
+    return _reserved;
+  }
```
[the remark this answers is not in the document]
ashu-mehra: ok, I will do that.

src/hotspot/share/cds/filemap.cpp
ashu-mehra: To be honest I don't like these names either and am open for suggestions. I liked @tstuefe's suggestion for `on_cds_post_load` and `on_cds_load_error` but probably name them `on_archive_heap_post_load` and `on_archive_heap_load_error`.
Also in CDS code "load" is currently associated with the mechanism used by non-G1 collecters to read the archive space into the heap memory and "map" is associated with the G1 collector's mechanism to mmap the archive space. For instance FileMapInfo::map_or_load_heap_region.
Going forward I think we would need to get rid of this distinction, at least in public APIs and just use the term "load". Internally the "loading" may be achieved through IO read or mmap-ing the archive space.
@iklam what do you think?

ashu-mehra: I will also update `alloc_archive_space` to `allocate_archive_space`. Thanks @iklam for pointing that out.

src/hotspot/share/gc/shared/collectedHeap.hpp
ashu-mehra: I don't see any reason why we can't do that. Only reason could be that we need the MemRegion in FileMapInfo as well, so we end up duplicating the MemRegion in CollectedHeap and FileMapInfo. 
But looking at the uses of `_mapped_heap_memregion` in FileMapInfo.cpp I think cds only needs the start address of the archive space in the heap. Size of the region must be same as the used bytes of `MetaspaceShared::hp` region. So we can ideally replace `_mapped_heap_memregion` with something like`_archive_space_load_address`.
@iklam @tschatzl what is your opinion?

src/hotspot/share/cds/filemap.cpp
ashu-mehra: On second thought I think there is no need for `ArchiveHeapLoader::can_map()` either. `map_heap_region()` gets called from `FileMapInfo::map_or_load_heap_region` only if `ArchiveHeapLoader::can_map()` is true.

src/hotspot/share/gc/shared/collectedHeap.hpp
```diff
@@ -207,6 +207,9 @@ class CollectedHeap : public CHeapObj<mtGC> {
   }
 
  public:
+  MemRegion reserved_region() const {
+    return _reserved;
+  }
```
albertnetymk, line 212: Could you extract out this (a new CollectedHeap API) to its own ticket/PR? (It seems important enough to not be hidden inside a CDS PR.)

ashu-mehra: Sure. I will open an issue and PR for this.

fisk (changes requested):
It's a bit disappointing for a PR aiming to make heap archiving GC agnostic, to make assumptions about GC internal memory layout, that doesn't apply to all collectors.
We have discussed previously with @iklam an approach where materializing archived objects uses the normal object allocation APIs. That would for real make the heap archiving mechanism GC agnostic. I would rather see that being prototyped, than a not GC agnostic approach that we might throw away right after it gets integrated, in favour of the more GC agnostic approach.

iklam:
> It's a bit disappointing for a PR aiming to make heap archiving GC agnostic, to make assumptions about GC internal memory layout, that doesn't apply to all collectors. We have discussed previously with @iklam an approach where materializing archived objects uses the normal object allocation APIs. That would for real make the heap archiving mechanism GC agnostic. I would rather see that being prototyped, than a not GC agnostic approach that we might throw away right after it gets integrated, in favour of the more GC agnostic approach.

There are a few problems with using the regular heap allocation API to reserve space for the archived objects:

- On G1, if the required (contiguous) space is larger than a single G1 region, it will end up inside a huge region
- If we instead attempt to allocate several blocks, each the size of a regular G1 region, there's no guarantee that these blocks would be contiguous. Fixing the pointers across discontiguous blocks would be messy
- We could allocate one block for each individual archived object. This will be most portable across GCs, but the pointer relocation will be slow. (Also, we have to handle the possibility of collecting in the middle of allocating these blocks)
- If we allocate one or multiple larger blocks, and fill them with the archived objects, we still need a per-GC specific way to fix-up these blocks at the very end. This is what we are doing today with Serial and Parallel:
    - https://github.com/openjdk/jdk/commit/6a573b888d4d3322b9165562f85e1b7b781a5ff1#diff-d53bf148d2758636cb8c6a54595515610a4c52d953e23e7bfdeef1106c14f626
    - https://github.com/openjdk/jdk/commit/234c17e8ff7b9b9713e9fded693381f985d34d2a#diff-b8bae64d029923161d88b761e6b775454924218e5f27f30130610da56d0bbefd

iklam:
> > It's a bit disappointing for a PR aiming to make heap archiving GC agnostic, to make assumptions about GC internal memory layout, that doesn't apply to all collectors. We have discussed previously with @iklam an approach where materializing archived objects uses the normal object allocation APIs. That would for real make the heap archiving mechanism GC agnostic. I would rather see that being prototyped, than a not GC agnostic approach that we might throw away right after it gets integrated, in favour of the more GC agnostic approach.
> 
> There are a few problems with using the regular heap allocation API to reserve space for the archived objects:
> 
> * On G1, if the required (contiguous) space is larger than a single G1 region, it will end up inside a huge region
> * If we instead attempt to allocate several blocks, each the size of a regular G1 region, there's no guarantee that these blocks would be contiguous. Fixing the pointers across discontiguous blocks would be messy
> * We could allocate one block for each individual archived object. This will be most portable across GCs, but the pointer relocation will be slow. (Also, we have to handle the possibility of collecting in the middle of allocating these blocks)
> * If we allocate one or multiple larger blocks, and fill them with the archived objects, we still need a per-GC specific way to fix-up these blocks at the very end. This is what we are doing today with Serial and Parallel:
>   
>   * [6a573b8#diff-d53bf148d2758636cb8c6a54595515610a4c52d953e23e7bfdeef1106c14f626](https://github.com/openjdk/jdk/commit/6a573b888d4d3322b9165562f85e1b7b781a5ff1#diff-d53bf148d2758636cb8c6a54595515610a4c52d953e23e7bfdeef1106c14f626)
>   * [234c17e#diff-b8bae64d029923161d88b761e6b775454924218e5f27f30130610da56d0bbefd](https://github.com/openjdk/jdk/commit/234c17e8ff7b9b9713e9fded693381f985d34d2a#diff-b8bae64d029923161d88b761e6b775454924218e5f27f30130610da56d0bbefd)

In the future, the Leyden project may prefer to have a "permanent" section of the archived objects that are never collected/moved. That would require more collector-specific code.

ashu-mehra:
My whole idea behind this is to have common GC APIs that the CDS can use to manage archive heap. As long as collector internals are properly abstracted out and hidden behind the APIs, CDS wouldn't need to worry about what the individual collectors do internally.
I agree the ideal scenario would be to use object allocation APIs but considering the issues @iklam pointed out, it doesn't look like a viable approach. @fisk do you agree with this assessment? Do you have any other alternative approach in mind?

iklam:
Based on discussion with @fisk, I created an RFE for investigating his idea. Please see https://bugs.openjdk.org/browse/JDK-8310823

It seems quite promising to me, and will greatly reduce (or eliminate) the interface between CDS and the collectors.

I would like to keep the current performance as possible, so I am leaning towards having an API for CDS to tell the collector about preferred location of the archived objects, to allow mmap'ing and reduce/avoid relocation. But such an hinting API will be much smaller than the ones proposed by this PR.

(We'd also need an reverse API for the collector to tell CDS what the preferred address would be, during CDS dump time).

iklam:
> Based on discussion with @fisk, I created an RFE for investigating his idea. Please see https://bugs.openjdk.org/browse/JDK-8310823
> 
> It seems quite promising to me, and will greatly reduce (or eliminate) the interface between CDS and the collectors.
> 
> I would like to keep the current performance as possible, so I am leaning towards having an API for CDS to tell the collector about preferred location of the archived objects, to allow mmap'ing and reduce/avoid relocation. But such an hinting API will be much smaller than the ones proposed by this PR.
> 
> (We'd also need an reverse API for the collector to tell CDS what the preferred address would be, during CDS dump time).

I've done a very simple (and rough) implementation of @fisk 's idea, without any optimizations. Every relocation is done via a hashtable lookup. For the default CDS archive, this makes about 48000 relocations on start-up.

https://github.com/openjdk/jdk/compare/master...iklam:jdk:8310823-materialize-cds-heap-with-regular-alloc?expand=1

```
$ perf stat -r 40 java --version > /dev/null
          0.015065 +- 0.000228 seconds time elapsed  ( +-  1.51% )

$ perf stat -r 40 java -XX:+NewArchiveHeapLoading --version > /dev/null
          0.020598 +- 0.000215 seconds time elapsed  ( +-  1.04% )

$ perf stat -r 40 java -XX:+UseSerialGC --version > /dev/null
          0.013929 +- 0.000229 seconds time elapsed  ( +-  1.64% )

$ perf stat -r 40 java -XX:+UseSerialGC -XX:+NewArchiveHeapLoading --version > /dev/null
          0.019107 +- 0.000222 seconds time elapsed  ( +-  1.16% )
```

The cost of the individual object allocation and relocation is about 5ms.

So far the slow down doesn't seem too outrages.

My next step is to optimize the relocation code to see how much faster it can get.

I noticed that the objects in `ArchiveHeapLoader::newcode_runtime_allocate_objects()` are allocated in at least two contiguous blocks, due to TLAB overflow.

ashu-mehra:
@iklam in your performance tests for "java -version" what is the "heap data relocation delta"? Is it non-zero? If so, can you also add the numbers for the runs with -Xmx128m which would correspond to the best case where no relocation is done just to add another data point.

iklam:
> @iklam in your performance tests for "java -version" what is the "heap data relocation delta"? Is it non-zero? If so, can you also add the numbers for the runs with -Xmx128m which would correspond to the best case where no relocation is done just to add another data point.

I first ran `java -Xshare:dump` so all the subsequent `java --version` runs use the same heap size as dump time. As a result, my "before" runs had a heap relocation delta of zero, which should correspond to the best start-up time.

Anyway, I just wanted to get a rough baseline. I will do more detailed benchmarking after implementing the optimized relocation for the `-XX:+NewArchiveHeapLoading` case.

ashu-mehra:
> I first ran java -Xshare:dump so all the subsequent java --version runs use the same heap size as dump time. As a result, my "before" runs had a heap relocation delta of zero, which should correspond to the best start-up time.

Okay, thanks for clarifying. I thought `java --version` runs were using the default archive.

iklam:
> > I first ran java -Xshare:dump so all the subsequent java --version runs use the same heap size as dump time. As a result, my "before" runs had a heap relocation delta of zero, which should correspond to the best start-up time.
> 
> Okay, thanks for clarifying. I thought `java --version` runs were using the default archive.

I haven't done any optimizations yet, but I fixed a few problems in the slow-path code. 

https://github.com/openjdk/jdk/compare/master...iklam:jdk:8310823-materialize-cds-heap-with-regular-alloc?expand=1

```
# Before: no relocation
$ perf stat -r 40 java --version > /dev/null
          0.015872 +- 0.000238 seconds time elapsed  ( +-  1.50% )

# Before: force relocation (quick)
$ perf stat -r 40 java -Xmx4g --version > /dev/null
          0.016691 +- 0.000385 seconds time elapsed  ( +-  2.31% )

# Before: force relocation ("quick relocation not possible")
$ perf stat -r 40 java -Xmx2g --version > /dev/null
          0.017385 +- 0.000230 seconds time elapsed  ( +-  1.32% )

# After
$ perf stat -r 40 java -XX:+NewArchiveHeapLoading --version > /dev/null
          0.018780 +- 0.000225 seconds time elapsed  ( +-  1.20% )
```

So the slow path is just about 3ms slower than the fastest "before" case.

Looking at the detailed timing break down (`os::thread_cpu_time()` = ns):

```
$ java -XX:+NewArchiveHeapLoading -Xlog:cds+gc --version
[0.006s][info][cds,gc] Num objs                    :                24184
[0.006s][info][cds,gc] Num bytes                   :              1074640
[0.006s][info][cds,gc] Per obj bytes               :                   44
[0.006s][info][cds,gc] Num references (incl nulls) :                87109
[0.006s][info][cds,gc] Num references relocated    :                43225
[0.006s][info][cds,gc] Allocation Time             :              1605084 <<<< A
[0.006s][info][cds,gc] Relocation Time             :              1246894
[0.006s][info][cds,gc] Table(s) dispose Time       :                 1306

$ java -XX:+NewArchiveHeapLoading -XX:NewArchiveHeapNumAllocs=2 -Xlog:cds+gc --version
[0.006s][info][cds,gc] Allocation Time             :              2203781 <<<< B

$ java -XX:+NewArchiveHeapLoading -XX:NewArchiveHeapNumAllocs=-1 -Xlog:cds+gc --version
[0.003s][info][cds,gc] Allocation Time             :               282125 <<<< C

$ java -XX:+NewArchiveHeapLoading -XX:NewArchiveHeapNumAllocs=0 -Xlog:cds+gc --version
[0.004s][info][cds,gc] Allocation Time             :               854249 <<<< D
```

The cost of allocating all the object is about 0.6 ms (B - A). Determining the size of the objects cost about 0.3ms (C). The cost of storing the relocation into the hastables is about 0.6 ms (D - C).

I hope to implement a fast path for relocation that avoids using the hash tables at all. If we can get the total alloc + reloc time to be about 1.5ms, then it would be just as fast as before when relocation is enabled.

tstuefe:
> > @iklam in your performance tests for "java -version" what is the "heap data relocation delta"? Is it non-zero? If so, can you also add the numbers for the runs with -Xmx128m which would correspond to the best case where no relocation is done just to add another data point.
> 
> I first ran `java -Xshare:dump` so all the subsequent `java --version` runs use the same heap size as dump time. As a result, my "before" runs had a heap relocation delta of zero, which should correspond to the best start-up time.
> 

I must be missing something basic, but I don't understand Ashu's question nor the answer.

You now allocate space for each object individually, then memcopy it to the allotted space, right? How could there *not* be relocation involved, for either absolute or narrow oop? The chance that the oop is allocated at exactly the same address or at the same offset to heap start as in the dump JVM  must be vanishingly low, right?

iklam:
> > I first ran `java -Xshare:dump` so all the subsequent `java --version` runs use the same heap size as dump time. As a result, my "before" runs had a heap relocation delta of zero, which should correspond to the best start-up time.
> 
> I must be missing something basic, but I don't understand Ashu's question nor the answer.
> 
> You now allocate space for each object individually, then memcopy it to the allotted space, right? How could there _not_ be relocation involved, for either absolute or narrow oop? The chance that the object is allocated at exactly the same address or at the same offset to heap start as in the dump JVM must be vanishingly low, right?

In my command lines,  `java --version` runs the existing code (mmap the archive heap). The new code (individual oop allocation) is used only with the `-XX:+NewArchiveHeapLoading` flag.

iklam:
> I hope to implement a fast path for relocation that avoids using the hash tables at all. If we can get the total alloc + reloc time to be about 1.5ms, then it would be just as fast as before when relocation is enabled.

I've implemented a fast relocation lookup. It currently uses a table of the same size as the archived heap objects, but I can reduce that to 1/2 the size.

See https://github.com/openjdk/jdk/compare/master...iklam:jdk:8310823-materialize-cds-heap-with-regular-alloc?expand=1

This is implemented by about 330 lines of code in archiveHeapLoader.cpp. The code is templatized to try out different approaches (like `-XX:+NahlRawAlloc` and `-XX:+NahlUseAccessAPI`), so it can be further simplified.

There's only one thing that's not yet implemented -- the equivalent of `ArchiveHeapLoader::patch_native_pointers()`. I'll do that next.

```
$ java  -XX:+NewArchiveHeapLoading -Xmx128m -Xlog:cds+gc --version
[0.004s][info][cds,gc] Delayed allocation records alloced: 640
[0.004s][info][cds,gc] Load Time: 1388458
```

The whole allocation + reloc takes about 1.4ms. It's about 1.25ms slower in the worst case (when the "old" code doesn't have to relocate -- see the `(**)` in the table below). It's 0.8ms slower when the "old" code has to relocate.

```
All times are in ms, for "java --version"

====================================
Dump: java -Xshare:dump -Xmx128m

G1         old        new       diff
 128m   14.476     15.754     +1.277 (**)
8192m   15.359     16.085     +0.726


Serial     old        new
 128m   13.442     14.241     +0.798
8192m   13.740     14.532     +0.791

====================================
Dump: java -Xshare:dump -Xmx8192m

G1         old        new       diff
 128m   14.975     15.787     +0.812
2048m   16.239     17.035     +0.796
8192m   14.821     16.042     +1.221 (**)


Serial     old        new
 128m   13.444     14.167     +0.723
8192m   13.717     14.502     +0.785
```

While the code is slower than before, it's a lot simpler. It works on all collectors. I tested on ZGC, but I think Shenandoah should work as well.

The cost is about 1.3 ms per MB of archived heap objects. This may be acceptable as it's a small fraction of JVM bootstrap. We have about 1MB of archived objects now,  and we don't expect this size to drastically increase in the near future.

The extra memory cost is:

- a temporary in-memory copy of the archived heap objects
- a temporary table of 1/2 the size of the archived heap objects

The former can be reduced by reading the archived objects in a stream. The latter can be reduced by a more elaborate relocation algorithm that assumes that most of the allocated objects are in a contiguous block. Such changes may cause further slow down.

ashu-mehra:
@iklam can you please elaborate a bit on relocation optimizations being done by the patch. Without any background on the idea, it is difficult to infer it from the code.

iklam:
> @iklam can you please elaborate a bit on relocation optimizations being done by the patch. Without any background on the idea, it is difficult to infer it from the code.

The algorithm tries to materialize all objects and relocate their oop fields in a single pass. (Each object has a "stream address" (its location in the input stream) and a "materialized address" (its location in the runtime heap))

- Materialize one object from the input stream
- Enter the materialized address of this object in the `reloc_table`. Since the input stream is contiguous, we can index `reloc_table` by computing the offset of the `stream` address of this object to the bottom of the input stream.
- For each non-null oop pointer in the materialized object:
    -  If the pointee's stream address is lower than that of the current object, update the pointer with the pointee's materialized address, which is already in `reloc_table`
    - Otherwise, enter the location of this pointer into `reloc_table`, as a linked-list of the `Dst` type, at the `reloc_table` of the pointee. When the pointee is materialized, it walks its own `Dst` list, and relocate all pointers to itself.

My branch contains a separate patch for [JDK-8251330](https://bugs.openjdk.org/browse/JDK-8251330) -- the input stream is ordered such that:
- the first 50% of the input stream contains no pointers, so relocation can be skipped altogether
- in the remaining input stream, about 90% of the 43225 pointers are pointing below the current object, so they can be relocated quickly. Less than 640 `Dst` are needed for the "delayed relocation".

fisk:
Excellent work, @iklam! This work looks very promising to me. Seems like the price we have to pay for the generality is very small indeed, and well worth it. Thank you for doing this.

ashu-mehra:
@iklam I agree this is a much better approach and makes the whole process truly collector agnostic. Great work, specially the optimization to re-order the objects.

Given that this has minimal impact on performance, are we good to go ahead with this approach now?

One issue I noticed while doing some testing with Shenandoah collector is probably worth pointing out here:
When using `-XX:+NahlRawAlloc` with very small heap size like -Xmx4m or -Xmx8m the java process freezes. . This happens because the allocations for archive objects causes pacing to kick in and the main thread waits on `ShenandoahPacer::_wait_monitor` [0] to be notified by ShenandoahPeriodicPacerNotify [1]. But the WatcherThread which is responsible for executing the `ShenandoahPeriodicPacerNotify` task does run the periodic tasks until VM init is done [2][3]. So the main thread is stuck now.

I guess if we do the wait in `ShenandoahPacer::pace_for_alloc` only after VM init is completed, it would resolve this issue.

I haven't noticed this with `-XX:-NahlRawAlloc`, not sure why that should make any difference.

Here are the stack traces:

main thread:
```
#5  0x00007f5a1fafbafc in PlatformMonitor::wait (this=this@entry=0x7f5a180f6c78, millis=<optimized out>, millis@entry=10) at src/hotspot/os/posix/mutex_posix.hpp:124
#6  0x00007f5a1faa3f9c in Monitor::wait (this=0x7f5a180f6c70, timeout=timeout@entry=10) at src/hotspot/share/runtime/mutex.cpp:254
#7  0x00007f5a1fc2d3bd in ShenandoahPacer::wait (time_ms=10, this=0x7f5a180f6a20) at src/hotspot/share/gc/shenandoah/shenandoahPacer.cpp:286
#8  ShenandoahPacer::pace_for_alloc (this=0x7f5a180f6a20, words=<optimized out>) at src/hotspot/share/gc/shenandoah/shenandoahPacer.cpp:263
#9  0x00007f5a1fbfc7e1 in ShenandoahHeap::allocate_memory (this=0x7f5a180ca590, req=...) at src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp:855
#10 0x00007f5a1fbfcb5c in ShenandoahHeap::mem_allocate (this=<optimized out>, size=<optimized out>, gc_overhead_limit_was_exceeded=<optimized out>) at src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp:931
#11 0x00007f5a1f2402c2 in NewQuickLoader::mem_allocate_raw (size=6) at src/hotspot/share/cds/archiveHeapLoader.cpp:493
#12 NewQuickLoaderImpl<true, true>::allocate (this=<optimized out>, __the_thread__=<optimized out>, size=<synthetic pointer>: 6, stream=0x7f5a1d228850) at src/hotspot/share/cds/archiveHeapLoader.cpp:712
#13 NewQuickLoaderImpl<true, true>::load_archive_heap_inner<false, false, 3> (__the_thread__=0x7f5a180b8810, stream_top=0x7f5a1d28e168, stream=0x7f5a1d228850, stream_bottom=0x7f5a1d204000, this=0x7f5a1edfe9c0)
    at src/hotspot/share/cds/archiveHeapLoader.cpp:634
#14 NewQuickLoaderImpl<true, true>::load_archive_heap (this=this@entry=0x7f5a1edfe9c0, __the_thread__=__the_thread__@entry=0x7f5a180b8810) at src/hotspot/share/cds/archiveHeapLoader.cpp:603
#15 0x00007f5a1f22da2a in ArchiveHeapLoader::new_fixup_region (__the_thread__=__the_thread__@entry=0x7f5a180b8810) at src/hotspot/share/cds/archiveHeapLoader.cpp:806
#16 0x00007f5a1f22f00b in ArchiveHeapLoader::fixup_region () at src/hotspot/share/cds/archiveHeapLoader.cpp:90
#17 0x00007f5a1fde5bd5 in vmClasses::resolve_all (__the_thread__=__the_thread__@entry=0x7f5a180b8810) at src/hotspot/share/classfile/vmClasses.cpp:145
#18 0x00007f5a1fd186a9 in SystemDictionary::initialize (__the_thread__=__the_thread__@entry=0x7f5a180b8810) at src/hotspot/share/classfile/systemDictionary.cpp:1616
#19 0x00007f5a1fd8b960 in Universe::genesis (__the_thread__=__the_thread__@entry=0x7f5a180b8810) at src/hotspot/share/memory/universe.cpp:356
#20 0x00007f5a1fd8d093 in universe2_init () at src/hotspot/share/memory/universe.cpp:977
#21 0x00007f5a1f6e10f9 in init_globals2 () at src/hotspot/share/runtime/init.cpp:150
#22 0x00007f5a1fd6c729 in Threads::create_vm (args=<optimized out>, canTryAgain=canTryAgain@entry=0x7f5a1edfedaf) at src/hotspot/share/runtime/threads.cpp:568
#23 0x00007f5a1f7af69e in JNI_CreateJavaVM_inner (args=<optimized out>, penv=0x7f5a1edfee58, vm=0x7f5a1edfee50) at src/hotspot/share/prims/jni.cpp:3577
#24 JNI_CreateJavaVM (vm=0x7f5a1edfee50, penv=0x7f5a1edfee58, args=<optimized out>) at src/hotspot/share/prims/jni.cpp:3668
#25 0x00007f5a206d81ef in InitializeJVM (ifn=<synthetic pointer>, penv=0x7f5a1edfee58, pvm=0x7f5a1edfee50) at src/java.base/share/native/libjli/java.c:1506
#26 JavaMain (_args=<optimized out>) at src/java.base/share/native/libjli/java.c:415
#27 0x00007f5a206dbe29 in ThreadJavaMain (args=<optimized out>) at src/java.base/unix/native/libjli/java_md.c:650
#28 0x00007f5a20582907 in start_thread (arg=<optimized out>) at pthread_create.c:444
#29 0x00007f5a20608870 in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
```

Watcher thread:
```
#4  ___pthread_cond_timedwait64 (cond=0x7f5a202f0e30 <mutex_init()::PeriodicTask_lock_storage+48>, mutex=<optimized out>, abstime=0x7f5a1d126d20) at pthread_cond_wait.c:643
#5  0x00007f5a1fafbafc in PlatformMonitor::wait (this=this@entry=0x7f5a202f0e08 <mutex_init()::PeriodicTask_lock_storage+8>, millis=<optimized out>, millis@entry=100) at src/hotspot/os/posix/mutex_posix.hpp:124
#6  0x00007f5a1faa3f19 in Monitor::wait_without_safepoint_check (this=this@entry=0x7f5a202f0e00 <mutex_init()::PeriodicTask_lock_storage>, timeout=timeout@entry=100) at src/hotspot/share/runtime/mutex.cpp:226
#7  0x00007f5a1fabcded in MonitorLocker::wait (this=<synthetic pointer>, timeout=100) at src/hotspot/share/runtime/mutexLocker.hpp:254
#8  WatcherThread::sleep (this=this@entry=0x7f5a1810a290) at src/hotspot/share/runtime/nonJavaThread.cpp:189
#9  0x00007f5a1fabce61 in WatcherThread::run (this=0x7f5a1810a290) at src/hotspot/share/runtime/nonJavaThread.cpp:249
#10 0x00007f5a1fd5f49f in Thread::call_run (this=0x7f5a1810a290) at src/hotspot/share/runtime/thread.cpp:217
#11 0x00007f5a1faf0a9a in thread_native_entry (thread=0x7f5a1810a290) at src/hotspot/os/linux/os_linux.cpp:779
#12 0x00007f5a20582907 in start_thread (arg=<optimized out>) at pthread_create.c:444
#13 0x00007f5a20608870 in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
```

[0] https://github.com/openjdk/jdk/blob/2d05d3545c8fe4d9e5ad3cee673fc938f84d1901/src/hotspot/share/gc/shenandoah/shenandoahPacer.cpp#L263
[1] https://github.com/openjdk/jdk/blob/2d05d3545c8fe4d9e5ad3cee673fc938f84d1901/src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp#L75-L78
[2] https://github.com/openjdk/jdk/blob/2d05d3545c8fe4d9e5ad3cee673fc938f84d1901/src/hotspot/share/runtime/threads.cpp#L804-L808
[3] https://github.com/openjdk/jdk/blob/2d05d3545c8fe4d9e5ad3cee673fc938f84d1901/src/hotspot/share/runtime/nonJavaThread.cpp#L288

iklam:
Hi @ashu-mehra thanks for testing the patch. I think we all agree that the minor performance impact is acceptable because the code is simpler and more portable. I'll try to clean up my patch and start a PR.

BTW, I have implemented a simpler relocation algorithm with similar performance. It uses less memory and hopefully will be easier to understand as well. The algorithm is described in comments inside archiveHeapLoader.cpp

https://github.com/openjdk/jdk/compare/master...iklam:jdk:8310823-materialize-cds-heap-with-regular-alloc.alternative-relocation?expand=1

As a prerequisite, I'll start a PR for [JDK-8251330: Reorder CDS archived heap to speed up relocation](https://bugs.openjdk.org/browse/JDK-8251330)

Regarding raw allocation, it doesn't seem to be too much faster, so maybe we should disable it, at least for the initial integration.

```
$ (for i in {1..6}; do perf stat -r 100 java -XX:+NewArchiveHeapLoading -XX:-NahlRawAlloc --version > /dev/null; done) 2>&1 | grep elapsed
         0.0162332 +- 0.0000914 seconds time elapsed  ( +-  0.56% )
         0.0161316 +- 0.0000228 seconds time elapsed  ( +-  0.14% )
         0.0161171 +- 0.0000250 seconds time elapsed  ( +-  0.16% )
         0.0161311 +- 0.0000231 seconds time elapsed  ( +-  0.14% )
         0.0161433 +- 0.0000244 seconds time elapsed  ( +-  0.15% )
         0.0161121 +- 0.0000271 seconds time elapsed  ( +-  0.17% )
$ (for i in {1..6}; do perf stat -r 100 java -XX:+NewArchiveHeapLoading -XX:+NahlRawAlloc --version > /dev/null; done) 2>&1 | grep elapsed
         0.0160640 +- 0.0000973 seconds time elapsed  ( +-  0.61% )
         0.0159320 +- 0.0000221 seconds time elapsed  ( +-  0.14% )
         0.0159910 +- 0.0000272 seconds time elapsed  ( +-  0.17% )
         0.0159406 +- 0.0000230 seconds time elapsed  ( +-  0.14% )
         0.0159930 +- 0.0000252 seconds time elapsed  ( +-  0.16% )
         0.0159670 +- 0.0000296 seconds time elapsed  ( +-  0.19% )
$ (for i in {1..6}; do perf stat -r 100 java -XX:-NewArchiveHeapLoading -XX:+NahlRawAlloc --version > /dev/null; done) 2>&1 | grep elapsed
         0.0149069 +- 0.0000932 seconds time elapsed  ( +-  0.63% )
         0.0148363 +- 0.0000259 seconds time elapsed  ( +-  0.17% )
         0.0148077 +- 0.0000218 seconds time elapsed  ( +-  0.15% )
         0.0148377 +- 0.0000212 seconds time elapsed  ( +-  0.14% )
         0.0148411 +- 0.0000245 seconds time elapsed  ( +-  0.17% )
         0.0148504 +- 0.0000258 seconds time elapsed  ( +-  0.17% )
```

DanHeidinga:
> The cost is about 1.3 ms per MB of archived heap objects. This may be acceptable as it's a small fraction of JVM bootstrap. We have about 1MB of archived objects now, and we don't expect this size to drastically increase in the near future.

Looking ahead to Project Leyden, I wouldn't be surprised if the current 1MB of archived heap became much larger.  Demo apps on Graal Native Image are often ~4MB of image heap, and while it's not an apples-to-apples comparison, it suggests that somewhere between 5MB & 10MB isn't unreasonable for Leyden.

Using 10MB as a baseline for easy math,  1.3ms/MB * 10MB = 13 ms for the new code?  And (1.3ms-0.8ms) = 0.5ms/MB * 10MB = 5ms for the old code?  Assuming I've interpreted the numbers correctly and importably that they scale linearly, it seems worth preserving the mmap approach for collectors that can support it.

Does that seem reasonable?  And justify preserving the mmap approach?

fisk:
> > The cost is about 1.3 ms per MB of archived heap objects. This may be acceptable as it's a small fraction of JVM bootstrap. We have about 1MB of archived objects now, and we don't expect this size to drastically increase in the near future.
> 
> 
> 
> Looking ahead to Project Leyden, I wouldn't be surprised if the current 1MB of archived heap became much larger.  Demo apps on Graal Native Image are often ~4MB of image heap, and while it's not an apples-to-apples comparison, it suggests that somewhere between 5MB & 10MB isn't unreasonable for Leyden.
> 
> 
> 
> Using 10MB as a baseline for easy math,  1.3ms/MB * 10MB = 13 ms for the new code?  And (1.3ms-0.8ms) = 0.5ms/MB * 10MB = 5ms for the old code?  Assuming I've interpreted the numbers correctly and importably that they scale linearly, it seems worth preserving the mmap approach for collectors that can support it.
> 
> 
> 
> Does that seem reasonable?  And justify preserving the mmap approach?

I think Leyden has bigger fish to fry. If we start seeing this being costly, then there are still more cards we can play to optimize the GC agnostic approach, such as e.g. exploiting the mostly contiguous nature of our allocations. But I'd like to wait and see if it's worth it or not, compred to potentially bigger fish we can spend time on instead. I'm happy with current numbers, and see a path forward should we need them to improve going forward.

iklam:
> > @iklam I agree this is a much better approach and makes the whole process truly collector agnostic. Great work, specially the optimization to re-order the objects.
> > Given that this has minimal impact on performance, are we good to go ahead with this approach now?
> > One issue I noticed while doing some testing with Shenandoah collector is probably worth pointing out here: When using `-XX:+NahlRawAlloc` with very small heap size like -Xmx4m or -Xmx8m the java process freezes. . This happens because the allocations for archive objects causes pacing to kick in and the main thread waits on `ShenandoahPacer::_wait_monitor` [0] to be notified by ShenandoahPeriodicPacerNotify [1]. But the WatcherThread which is responsible for executing the `ShenandoahPeriodicPacerNotify` task does run the periodic tasks until VM init is done [2][3]. So the main thread is stuck now.
> > I guess if we do the wait in `ShenandoahPacer::pace_for_alloc` only after VM init is completed, it would resolve this issue.
> 
> Yes, Shenandoah pacing should not run before VM init is complete. This seems like a separate issue.
> 
> Great work in this PR, though!

I have done more testing with my prototype for JDK-8310823 (Materialize CDS archived objects with regular GC allocation) [1]  [2] and found a few issues:

- The minimum heap requirement has increased significantly. We allocate many objects during VM start-up. This may fill up the young generation for some collectors, causing the VM to exit as a GC is not yet possible at this stage. Since the generation sizing policy is different in each GC (and not really controllable via `-XX:NewSize`), the failure mode with small heap size becomes unpredictable. I think this a functional regression, which is more serious that the performance regression in start-up time.
- Although it's possible to enable archive heap objects for ZGC with my JDK-8310823 patch , there's only very marginal improvement in start-up time (probably because ZGC is doing many other tasks concurrently during start-up)

As we expect more heap objects to be archived in Project Leyden, we need a solution that scales well without blowing up the minimum heap size. For example, with the mainline, an 18MB archived heap can be mapped with -Xmx20M, but with my patch, I needed to use -Xmx40M.

To reduce the minimum heap requirement, @fisk has developed a prototype that materializes the archived objects incrementally. However, it's not clear whether future Leyden developments would support such incremental loading.

As we don't see any immediate benefits for JDK-8310823, I would suggest putting that on hold for now, until we get better understanding of the requirements from Leyden.

------
[1] https://github.com/openjdk/jdk/pull/15730
[2] https://bugs.openjdk.org/browse/JDK-8310823

PR #14521, opened 2023-06-16
8309976: Add microbenchmark for stressing code cache

ericcaspole:
Most benchmarks have a relatively small code footprint compared to enterprise applications. While trying to model an application with a very large code footprint, we developed this JMH with its own classloader generating the desired number of classes from the string literal in the file, using the existing InMemoryJavaCompiler. Then these classes are are instantiated to the desired count, and methods are called in those objects, which can fill up the code cache, possibly causing code cache sweeping or compiler shut-off. 
This allows to create a simulation of a large application with arbitrary java heap and code cache footprint, and take advantage of the benefits of JMH at the same time.
The defaults are set very low by default and the intent is that they would be customized for any given study.

---------
### Progress

### Issue
 * [JDK-8309976](https://bugs.openjdk.org/browse/JDK-8309976): Add microbenchmark for stressing code cache (**Enhancement** - P4)

### Reviewers
 * [Claes Redestad](https://openjdk.org/census#redestad) (@cl4es - **Reviewer**) ⚠️ Review applies to [6922b9dd](https://git.openjdk.org/jdk/pull/14521/files/6922b9dddec3ce10213c4ec1a5cf5aacacd628d7)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14521/head:pull/14521` \
`$ git checkout pull/14521`

`$ git checkout pull/14521` \
`$ git pull https://git.openjdk.org/jdk.git pull/14521/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14521`

`$ git pr show -t 14521`

</details>

<a href="https://git.openjdk.org/jdk/pull/14521.diff">https://git.openjdk.org/jdk/pull/14521.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14521#issuecomment-1594838388)


cl4es (approved):
Consider changing the bug summary to "Add microbenchmark for stressing code cache".

test/micro/org/openjdk/bench/vm/compiler/CodeCacheStress.java
```diff
@@ -0,0 +1,447 @@
...
+                + "    int padCP = 0;"
+                + "    int padCQ = 0;"
+                + "    int padCR = 0;"
+                + "    int padCS = 0;"
+                + "    int padCT = 0;"
+                + " "
+                + "    int instD = 0;"
+                + " "
+                + " "
+                + "   public Integer get( Map m, String k, Integer depth) { "
```
shipilev, line 188:
```suggestion
                + "   public Integer get(Map m, String k, Integer depth) { "
```

...and later...

test/micro/org/openjdk/bench/vm/compiler/CodeCacheStress.java
```diff
@@ -0,0 +1,447 @@
...
+        for (int index = 0; index < compiledClasses.length; index++) {
+            try {
+                Class c = chooseClass();
+                Object r = chooseInstance(c);
+                Method m = chooseMethod(c);
+                assert m != null;
+                Map map = chooseMap();
+                Integer result = callTheMethod(m, r, k, map);
+                assert result != null && result >= v;
+                sum += result;
```
shipilev, line 438: This performs boxed addition.

shipilev (changes requested):
I think the benchmark code needs massaging for style and other issues.
See e.g. the cursory review:

ericcaspole:
@shipilev - ping, Aleksey is this version OK with you? 
Thanks, 
Eric

test/micro/org/openjdk/bench/vm/compiler/CodeCacheStress.java
```diff
@@ -0,0 +1,445 @@
...
+            } else {
+                return super.findClass(name);
+            }
+        }
+    }
+
+    CodeCacheStress.BenchLoader loader1 = new CodeCacheStress.BenchLoader();
+
+    final String k = "key";
+    final Integer v = 1000;
```
shipilev, line 321: What are these fields? Do they need to be here? Do they need to be `static final` and be on top?

test/micro/org/openjdk/bench/vm/compiler/CodeCacheStress.java
```diff
@@ -0,0 +1,445 @@
...
+
+    final String methodNames[] = {
+            "get"
+    };
```
shipilev, line 325: Does it have to be an array? There is only a single element, should it be a single `static final` field?

test/micro/org/openjdk/bench/vm/compiler/CodeCacheStress.java
```diff
@@ -0,0 +1,445 @@
...
+                + "       if (depth > 0) {"
+                + "         instA += ((depth % 2) + staticA);"
+                + "         return (Integer) m.get(k) + get2(m, k, --depth);"
+                + "       } else {"
+                + "         setA(depth);"
+                + "         return (Integer) m.get(k)+ 10;"
+                + "       }"
+                + "   }"
+                + " "
+                + "   public Integer get2( Map m, String k, Integer depth) { "
```
shipilev, line 196: Here and later `get2( ` has redundant spaces after the parenthesis.

test/micro/org/openjdk/bench/vm/compiler/CodeCacheStress.java
```diff
@@ -0,0 +1,445 @@
...
+
+    // How many instances of each generated class to create and call in the measurement phase
+    @Param({"100"})
+    public int instanceCount;
+
+    byte[][] compiledClasses;
+    Class[] loadedClasses;
+    String[] classNames;
+
+    int index = 0;
```
shipilev, line 82: Why this `index` is here? It seems not be used except that within the `setupClasses`.

shipilev (commented):
Some more things that are visible from here are below. These are not _all_ the problematic things that are there, please take a closer look at the file from the style perspective.

test/micro/org/openjdk/bench/vm/compiler/CodeCacheStress.java
ericcaspole: They are used both for filling the maps passed as arguments and the key is passed later as a parameter when calling these generated methods. They could be static, yes.

test/micro/org/openjdk/bench/vm/compiler/CodeCacheStress.java
ericcaspole: In earlier versions it called more than 1 method in the generated classes, and I would prefer to keep that capability. It could be static, yes.

test/micro/org/openjdk/bench/vm/compiler/CodeCacheStress.java
ericcaspole: Thanks, fixed in the next rev.

test/micro/org/openjdk/bench/vm/compiler/CodeCacheStress.java
ericcaspole: Using index in this way to coordinate creating classes named like "B"+index is a convention that we have used in earlier tests such as hotspot/jtreg/serviceability/jvmti/RedefineClasses/TestMultipleClasses.java

shipilev (approved):

PR #14522, opened 2023-06-16
8307927: C2: "malformed control flow" with irreducible loop

rwestrel:
The test contains a loop nest with 2 loops. The outer loop is an
irreducible loop. The safepoint for that loop is also in the inner
loop. Because `IdealLoopTree::check_safepts()` ignores irreducible
loops, that safepoint is not marked as being required and is
eliminated from the inner loop. The inner loop is then optimized out
and the outer loop becomes an infinite loop with no safepoint (a
single node loop). That, in turn, causes the loop to be eliminated
because it has not use and the assert fires.

The fix I propose is to make `IdealLoopTree::check_safepts()` work
with irreducible loops. I think
`IdealLoopTree::allpaths_check_safepts()` can be used for that. When
working on this I wondered if that method could be called with a loop
whose head has more than 3 inputs. I couldn't write a test case with
an irreducible loop whose head had more than 3 inputs but I added an
assert in the method and ran some testing. That assert fired so I also
propose to tweak the method so it's robust in that case.

---------
### Progress

### Issue
 * [JDK-8307927](https://bugs.openjdk.org/browse/JDK-8307927): C2: "malformed control flow" with irreducible loop (**Bug** - P4)

### Reviewers
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - **Reviewer**) ⚠️ Review applies to [3bb72120](https://git.openjdk.org/jdk/pull/14522/files/3bb721201f9eb453e92aa58ff0ed814c25b1297b)
 * [Emanuel Peter](https://openjdk.org/census#epeter) (@eme64 - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14522/head:pull/14522` \
`$ git checkout pull/14522`

`$ git checkout pull/14522` \
`$ git pull https://git.openjdk.org/jdk.git pull/14522/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14522`

`$ git pr show -t 14522`

</details>

<a href="https://git.openjdk.org/jdk/pull/14522.diff">https://git.openjdk.org/jdk/pull/14522.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14522#issuecomment-1594932385)


TobiHartmann (approved):
Looks reasonable to me. All tests passed.

@eme64 Please have a look as well.

src/hotspot/share/opto/loopnode.cpp
```diff
@@ -3514,9 +3514,9 @@ bool IdealLoopTree::beautify_loops( PhaseIdealLoop *phase ) {
 // encountered.  Helper for check_safepts.
 void IdealLoopTree::allpaths_check_safepts(VectorSet &visited, Node_List &stack) {
```
eme64, line 3515: @rwestrel you should update the description here. Suggestion:
```
Allpaths backwards scan. Starting at the head, traversing all backedges, and the body. Terminating each path at first safepoint encountered.  Helper for check_safepts.
```

eme64: Also the line above is not accurate enough anymore:
`_required_safept->push(n);  // save the one closest to the tail`

For one: could there not be multiple such SafePoints? If so: what does it mean to take "the closest"? And we may have multiple backedges / tails, now that we allow irreducible loops.

eme64: Other than that the fix looks reasonable to me, thanks for the fix @rwestrel !

rwestrel:
@eme64 thanks for reviewing this (and making suggestions). Does the updated change look ok?

eme64 (approved):
Thanks for the changes @rwestrel , looks good now!

rwestrel:
@TobiHartmann @eme64 thanks for the reviews.

PR #10395, opened 2022-09-22
JDK-8200337: Generalize see and link tags for user-defined anchors

hns:
Please review a a new feature to allow `@link`, `@linkplain` and `@see` tags to link to arbitrary URI fragments in the generated documentation (including in auxiliary `doc-files` documentation).

The changes in module `jdk.compiler` are mostly cleanup changes retained from earlier versions of the patch. The current proposed version uses a very simple change in `ReferenceParser` to avoid parsing the member name section of the reference when a non-member fragment is encountered.

The implementation introduces a new form of reference with a double hash mark (`##`) separator. This is a change from the previous implementation which also auto-recognized URI fragments and documentation paths by looking for `-` characters which are not allowed in member names. This feature was removed upon further consideration because it makes the feature more complex and less recognizable.
 
Links to auxiliary documentation files follow the same rules. They are recognized by looking for `/` characters in the fragment name. This means that ordinary `id` attribute values must not contain `/`, while auxiliary file paths must contain a `/` character. Both restrictions should be easy to sustain.

One thing that is difficult for this feature is to provide a good link label if no label is supplied in the tag. In contrast to program element names a fragment name does usually not make a good human readable name. The solution is to use the fragment name as default label text. I expect that the feature will usually be used with a user provided label.

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change requires a CSR request to be approved

### Issues
 * [JDK-8200337](https://bugs.openjdk.org/browse/JDK-8200337): Generalize see and link tags for user-defined anchors
 * [JDK-8294195](https://bugs.openjdk.org/browse/JDK-8294195): Generalize see and link tags for user-defined anchors (**CSR**)

### Reviewers
 * [Jonathan Gibbons](https://openjdk.org/census#jjg) (@jonathan-gibbons - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk pull/10395/head:pull/10395` \
`$ git checkout pull/10395`

Update a local copy of the PR: \
`$ git checkout pull/10395` \
`$ git pull https://git.openjdk.org/jdk pull/10395/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 10395`

[...]

PR #14528, opened 2023-06-16
8310242: Clarify the name parameter to Class::forName

mlchung:
This PR clarifies the spec of the 3-arg Class::forName  regarding the format of the name for an array type which is of the form: one or more of "[" + binary name of the element type + ";'.

---------
### Progress
- [x] Change requires CSR request [JDK-8310244](https://bugs.openjdk.org/browse/JDK-8310244) to be approved

### Issues
 * [JDK-8310242](https://bugs.openjdk.org/browse/JDK-8310242): Clarify the name parameter to Class::forName (**Bug** - P4)
 * [JDK-8310244](https://bugs.openjdk.org/browse/JDK-8310244): Clarify the name parameter to Class::forName (**CSR**)

### Reviewers
 * [Roger Riggs](https://openjdk.org/census#rriggs) (@RogerRiggs - **Reviewer**) ⚠️ Review applies to [004f35d5](https://git.openjdk.org/jdk/pull/14528/files/004f35d591e1e61ce725a0dea2e2775d1e204234)
 * [Chen Liang](https://openjdk.org/census#liach) (@liach - Author)
 * [Alan Bateman](https://openjdk.org/census#alanb) (@AlanBateman - **Reviewer**) ⚠️ Review applies to [852bd774](https://git.openjdk.org/jdk/pull/14528/files/852bd7740fb7a83e3b15b128a6d319efe4aae17b)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14528/head:pull/14528` \
`$ git checkout pull/14528`

`$ git checkout pull/14528` \
`$ git pull https://git.openjdk.org/jdk.git pull/14528/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14528`

`$ git pr show -t 14528`

</details>

<a href="https://git.openjdk.org/jdk/pull/14528.diff">https://git.openjdk.org/jdk/pull/14528.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14528#issuecomment-1595241362)


src/java.base/share/classes/java/lang/Class.java
```diff
@@ -438,8 +438,10 @@ private static Class<?> forName(String className, Class<?> caller)
      * obtain any of the {@code Class} objects representing primitive
      * types or void.
      *
-     * <p> If {@code name} denotes an array class, the component type of
-     * the array class is loaded but not initialized.
+     * <p> If {@code name} denotes an array class, i.e. the name consists of
+     * one or more {@code '['} followed by the binary name of the element type
+     * and {@code ';'}, the element type of the array class is loaded but
+     * not initialized.
```
RogerRiggs, line 444: Do arrays of primitive types need special treatment?  For example, "[I"

mlchung: Thanks for catching it.  Updated.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -438,10 +438,15 @@ private static Class<?> forName(String className, Class<?> caller)
...
-     * and {@code ';'}, the element type of the array class is loaded but
-     * not initialized.
+     * <p> If {@code name} denotes an array class whose element type
+     * is a class or interface, i.e. the name consists of one or more {@code '['}
+     * followed by the binary name of the element type and {@code ';'}
+     * the element type of the array class is loaded but not initialized.
+     *
+     * <p> To find an array class of a primitive type, the name string
+     * consists of one or more {@code '['} followed by the field type descriptor
+     * of the primitive type (JVMS {@jvms 4.3.2 }) e.g. {@code "[I"} to
```
RogerRiggs, line 448: Can we avoid mixing "i.e." and "e.g"?
There are more "For example" occurrences than either "i.e." or "e.g." 
Your call.

mlchung: ok.  is the revised version better?

RogerRiggs (approved):
Looks good, thanks

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -438,8 +438,15 @@ private static Class<?> forName(String className, Class<?> caller)
      * obtain any of the {@code Class} objects representing primitive
      * types or void.
      *
-     * <p> If {@code name} denotes an array class, the component type of
-     * the array class is loaded but not initialized.
+     * <p> If {@code name} denotes an array class whose element type
+     * is a class or interface, the name consists of one or more {@code '['}
+     * followed by the binary name of the element type and {@code ';'},
```
liach, line 443: There's an L between [ and component binary name, like [Ljava.lang.Object;

liach (changes requested):

src/java.base/share/classes/java/lang/Class.java
mlchung: Updated.

AlanBateman (approved):
This changes look okay but they do make me wonder if we should add a note to the 2-arg Class.forName to make it clear that it does not allow class names for array types, the class name has to be a binary name.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -392,9 +392,10 @@ static String typeVarBounds(TypeVariable<?> typeVar) {
      * caller frame on the stack (e.g. when called directly from a JNI
      * attached thread), the system class loader is used.
      *
-     * @param      className   the fully qualified name of the desired class.
-     * @return     the {@code Class} object for the class with the
-     *             specified name.
+     * @param     className the <a href="#binary-name">binary name</a> of the class
```
jensli, line 395: The new text is probably more correct, but it is less easy to understand. I think "fully qualified name" of a class is a well-known term, but "binary name" is rather unknown.

I suggest changing the text to mention both terms. For example like this:

```suggestion
     * @param     className  the <a href="#binary-name">binary name</a>) of the class, for example its fully qualified name
```

jensli, line 395: Where do these links lead? I don't find any section with this ID in the `Class` Javadoc. Should they lead to `ClassLoader`? That Javadoc has such a section.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -423,11 +424,10 @@ private static Class<?> forName(String className, Class<?> caller)
     /**
      * Returns the {@code Class} object associated with the class or
      * interface with the given string name, using the given class loader.
-     * Given the fully qualified name for a class or interface (in the same
-     * format returned by {@code getName}) this method attempts to
-     * locate and load the class or interface.  The specified class
-     * loader is used to load the class or interface.  If the parameter
-     * {@code loader} is null, the class is loaded through the bootstrap
+     * Given the {@linkplain ##binary-name binary name} for a class or interface,
```
jensli, line 427: Is double hash a mistake?

ExE-Boss: That’s a new **Javadoc** feature available since **JDK 20**: [JDK‑8200337] ([GH‑10395])

[JDK‑8200337]: https://bugs.openjdk.org/browse/JDK-8200337 "[JDK‑8200337] Generalize see and link tags for user‑defined anchors"
[GH‑10395]: https://github.com/openjdk/jdk/pull/10395

src/java.base/share/classes/java/lang/Class.java
liach: Binary name is defined in JLS. Fully qualified name is ambiguous as it may refer to the internal name or a descriptor string.

dholmes-ora: This seems a specification change as the binary-name and FQN are not always the same. In particular for a member class `M` in class `C` the binary name is `C$M`, whereas the FQN is `C.M`.

mlchung:
I added an API note in the 2-arg forName that it does not support loading of array types.  Also fixed the links.   A new test is added to show the valid and invalid names.

src/java.base/share/classes/java/lang/Class.java
AlanBateman: > Binary name is a long-standing behavior. It's a spec bug.

Yes, I view this PR as just specifying long standing behavior. The name used for array class may be surprising but I don't think it can be changed.

mlchung:
I improved the spec of `forName` to make it clear of the name for nested types and array types.   It now refers to the table in `getName` about the name of an array class.   

@dholmes-ora does that clarify and address the concerns you had?

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -423,45 +424,52 @@ private static Class<?> forName(String className, Class<?> caller)
...
+     * an attempt will be made to locate a user-defined class in the unnamed package
+     * whose name is {@code I} instead.
      *
-     * <p> If {@code name} denotes an array class, the component type of
-     * the array class is loaded but not initialized.
-     *
-     * <p> For example, in an instance method the expression:
+     * <p> To obtain {@code Class} object associated with an array class,
+     * the name consists of one or more {@code '['} representing the depth
+     * of the array class, followed by the element type as encoded in
```
dholmes-ora, line 444: Suggestion
```
+     * <p> To obtain the {@code Class} object associated with an array class,
+     * the name consists of one or more {@code '['} representing the depth
+     * of the array nesting, followed by the element type as encoded in
```

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -423,45 +424,52 @@ private static Class<?> forName(String className, Class<?> caller)
...
+     * {@linkplain ##nameFormat the table} specified in {@code Class.getName()}.
      *
+     * <p> Examples:
      * {@snippet lang="java" :
-     * Class.forName("Foo")
+     * Class<?> threadClass = Class.forName("java.lang.Thread", false, currentLoader);
+     * Class<?> stringArrayClass = Class.forName("[Ljava.lang.String;", false, currentLoader);
+     * Class<?> intArrayClass = Class.forName("[[[I", false, currentLoader);
```
dholmes-ora, line 451: The variable name doesn't match the actual class loaded. It might be clearer to add a trailing comment `// Class of int[][][]` ?

test/jdk/java/lang/Class/forName/ForNameNames.java
```diff
@@ -0,0 +1,84 @@
...
+                Arguments.of("java.lang.String", String.class),
+                Arguments.of("[Ljava.lang.String;", String[].class),
+                Arguments.of("ForNameNames$Inner", Inner.class),
+                Arguments.of("[[I", int[][].class)
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("testCases")
+    void testForName(String cn, Class<?> expected) throws ClassNotFoundException {
```
dholmes-ora, line 52: Should this also test that `c.getName().equals(cn)`?

dholmes-ora (changes requested):
A few small suggestions, otherwise looks good.

Thanks.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -423,45 +424,52 @@ private static Class<?> forName(String className, Class<?> caller)
     /**
      * Returns the {@code Class} object associated with the class or
      * interface with the given string name, using the given class loader.
```
liach, line 426: Should we update the summary to `... associated with the class or interface or array with the given string name...`

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -423,45 +424,52 @@ private static Class<?> forName(String className, Class<?> caller)
...
      *
-     * <p> If {@code name} denotes a primitive type or void, an attempt
-     * will be made to locate a user-defined class in the unnamed package whose
-     * name is {@code name}. Therefore, this method cannot be used to
-     * obtain any of the {@code Class} objects representing primitive
-     * types or void.
+     * <p> This method cannot be used to obtain any of the {@code Class} objects
+     * representing primitive types or void, hidden classes or interfaces,
+     * or array classes whose element type is a hidden class or interface.
+     * If {@code name} denotes a primitive type or void, for example {@code I},
```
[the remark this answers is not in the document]
mlchung: One main point for this example to show is that it will find a class in unnamed package named `I`.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -423,45 +424,52 @@ private static Class<?> forName(String className, Class<?> caller)
...
      * }
      *
-     * is equivalent to:
+     * <p> A call to {@code getName()} on the {@code Class} object returned
+     * from {@code forName(}<i>N</i>{@code )} returns <i>N</i>.
      *
-     * {@snippet lang="java" :
-     * Class.forName("Foo", true, this.getClass().getClassLoader())
-     * }
+     * <p> A call to {@code forName("[L}<i>N</i>{@code ;")} causes the element type
```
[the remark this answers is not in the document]
mlchung: I considered that.   The spec and the example already describe multi-dimensional array.  I think one can infer from this single-dimensional array example.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -441,14 +441,14 @@ private static Class<?> forName(String className, Class<?> caller)
      *
      * <p> To obtain {@code Class} object associated with an array class,
```
dholmes-ora, line 442: To obtain _the_ Class object ...

test/jdk/java/lang/Class/forName/ForNameNames.java
```diff
@@ -0,0 +1,92 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+import static org.junit.jupiter.api.Assertions.*;
+
+public class ForNameNames {
+    static class Inner {}
+    static Stream<Arguments> testCases() {
+        return Stream.of(
+                Arguments.of("java.lang.String", String.class),
+                Arguments.of("[Ljava.lang.String;", String[].class),
+                Arguments.of("ForNameNames$Inner", Inner.class),
```
AlanBateman, line 45: For completes, you could add "[LForNameNames$Inner;" to this list of test cases.

AlanBateman (approved):

src/java.base/share/classes/java/lang/Class.java
mlchung: I want to keep it as is.  The first paragraph talks about a class or interface and the other paragraphs talk about array classes, primitive types and hidden classes.

mlchung:
I attached the specdiff to the [CSR](https://bugs.openjdk.org/browse/JDK-8310244) - please review.

liach (approved):

dholmes-ora (approved):

PR #14530, opened 2023-06-17
8310264: In PhaseChaitin::Split defs and phis are leaked

jdksjolen:
Hi,

`defs` and `phis` are leaked as they are resource allocated but not protected by a `ResourceMark`. The intention might have been for these to also live in the `split_arena`.. This change is the most conservative one, however, and does fix the memory leak.

Please consider, thanks.

Johan

---------
### Progress

### Issue
 * [JDK-8310264](https://bugs.openjdk.org/browse/JDK-8310264): In PhaseChaitin::Split defs and phis are leaked (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14530/head:pull/14530` \
`$ git checkout pull/14530`

`$ git checkout pull/14530` \
`$ git pull https://git.openjdk.org/jdk.git pull/14530/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14530`

`$ git pr show -t 14530`

</details>

<a href="https://git.openjdk.org/jdk/pull/14530.diff">https://git.openjdk.org/jdk/pull/14530.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14530#issuecomment-1595797285)

### DIFF
--- a/src/hotspot/share/opto/reg_split.cpp
+++ b/src/hotspot/share/opto/reg_split.cpp
@@ -571,8 +571,8 @@ uint PhaseChaitin::Split(uint maxlrg, ResourceArea* split_arena) {
   }
 
   // Keep track of DEFS & Phis for later passes
-  Node_List defs{split_arena, 8};
-  Node_List phis{split_arena, 16};
+  Node_List defs(split_arena, 8);
+  Node_List phis(split_arena, 16);
 
   //----------PASS 1----------
   //----------Propagation & Node Insertion Code----------
--- a/src/hotspot/share/opto/reg_split.cpp
+++ b/src/hotspot/share/opto/reg_split.cpp
@@ -571,8 +571,8 @@ uint PhaseChaitin::Split(uint maxlrg, ResourceArea* split_arena) {
   }
 
   // Keep track of DEFS & Phis for later passes
-  Node_List defs{split_arena, 128}; // Typically less than 128
-  Node_List phis{split_arena, 256}; // Typically less than 256
+  Node_List defs{split_arena, 8};
+  Node_List phis{split_arena, 16};
 
   //----------PASS 1----------
   //----------Propagation & Node Insertion Code----------
--- a/src/hotspot/share/opto/reg_split.cpp
+++ b/src/hotspot/share/opto/reg_split.cpp
@@ -499,7 +499,6 @@ uint PhaseChaitin::Split(uint maxlrg, ResourceArea* split_arena) {
 
   // Free thread local resources used by this method on exit.
   ResourceMark rm(split_arena);
-  ResourceMark rm2;
 
   uint                 bidx, pidx, slidx, insidx, inpidx, twoidx;
   uint                 non_phi = 1, spill_cnt = 0;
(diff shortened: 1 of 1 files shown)


TobiHartmann (approved):
Looks good to me.

chhagedorn (approved):
Looks good! I'm wondering, why we don't stack allocate both `Node_List` instead of `new Node_List()`. But regardless of that, we should indeed add a `ResourceMark`.

shipilev:
Looking at  https://github.com/openjdk/jdk/commit/f0d08c04f1312045c5e6f77935b04fe90967a186 -- it would seem that `def` and `phi` just missed the allocation in `split_arena`? I think it would a cleaner/safer to change `def`/`phi` to be allocated in `split_arena`, rather than doing the blank `ResourceMark` here. Are we sure nothing RA-allocated in used after we leave `Split`?

src/hotspot/share/opto/reg_split.cpp
```diff
@@ -569,8 +565,14 @@ uint PhaseChaitin::Split(uint maxlrg, ResourceArea* split_arena) {
...
-  for( slidx = 0; slidx < spill_cnt; slidx++ )
-    UP_entry[slidx] = new VectorSet(split_arena);
+  // Each containing at most spill_cnt * _cfg.number_of_blocks() entries.
+  for (slidx = 0; slidx < spill_cnt; slidx++) {
+    UP_entry[slidx] = new(split_arena) VectorSet(split_arena);
+  }
+
+  // Keep track of DEFS & Phis for later passes
+  Node_List defs{split_arena, 8};
+  Node_List phis{split_arena, 16};
```
TobiHartmann, line 575: Why do you use aggregate initialization instead of constructor invocation here?

TobiHartmann (approved):
Good catch! Looks good to me otherwise.

src/hotspot/share/opto/reg_split.cpp
jdksjolen: This will call the constructor, and has since C++11 I believe.  However, I'm clearly being inconsistent here and with the VectorSet change above. There's no reason that I picked brace initializer other than it being the 'modern' way: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-list

I'll revert this change and use regular parens to be more stylistically consistent.

TobiHartmann: Thanks for the background, I wasn't aware of that. I don't have a strong opinion but consistency in the same area would be nice.

vnkozlov: Please, change to normal `()`. Using '{}' is very confusing for not modern C++ experts and affects maintainability of this code.

jdksjolen:
@chhagedorn , @TobiHartmann , would you mind re-approving this PR? Thank you.

vnkozlov (approved):
Good.

jdksjolen:
Thank you.

/integrate

PR #14533, opened 2023-06-19
8309894: compiler/vectorapi/VectorLogicalOpIdentityTest.java fails on SVE system with UseSVE=0

XiaohongGong:
This test fails with several IR check failures when run on ARM SVE systems with vm option `-XX:UseSVE=0`. Here is one of the failed IR rule:

```
@IR(counts = {IRNode.AND_V, "1"}, applyIfCPUFeatureOr = {"sve", "true", "avx512", "true"})
public static void testAndMaskSameValue1()
```
The specified IR in the test depends on the platform's predicate feature. Hence the IR check can be applied only on ARM SVE or X86 AVX512 systems. But with `-XX:UseSVE=0` on ARM SVE machines, JVM will disable SVE feature for compiler. But the CPU feature is not changed. To guarantee the IR rule is run with SVE as expected, it has to add another condition like `applyIf = {"UseSVE", ">0"}`. Consider `UseSVE` is an ARM specific option, it can be used only on ARM CPUs. So the right IR rules should be:

```
@IR(counts = {IRNode.AND_V, "1"}, applyIfCPUFeature = {"sve", "true"}, applyIf = {"UseSVE", "> 0"})
@IR(counts = {IRNode.AND_V, "1"}, applyIfCPUFeature = {"avx512", "true"})
public static void testAndMaskSameValue1()

```
This patch also changes the check order of conditions for a IR rule. It's better to check `applyIfCPUFeature` before `applyIf`, in case the vm option is invalid on running hardware, which makes test throw exception and abort.

Verified on X86 systems with `UseAVX=1/2/3` by removing the test from ProblemList.txt, and SVE systems with `UseSVE=0/1`.

---------
### Progress

### Issue
 * [JDK-8309894](https://bugs.openjdk.org/browse/JDK-8309894): compiler/vectorapi/VectorLogicalOpIdentityTest.java fails on SVE system with UseSVE=0 (**Bug** - P4)

### Reviewers
 * [Emanuel Peter](https://openjdk.org/census#epeter) (@eme64 - **Reviewer**) ⚠️ Review applies to [de33dd21](https://git.openjdk.org/jdk/pull/14533/files/de33dd21af98ff7db4271d3f0c402a851ac8e373)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14533/head:pull/14533` \
`$ git checkout pull/14533`

`$ git checkout pull/14533` \
`$ git pull https://git.openjdk.org/jdk.git pull/14533/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14533`

`$ git pr show -t 14533`

</details>

<a href="https://git.openjdk.org/jdk/pull/14533.diff">https://git.openjdk.org/jdk/pull/14533.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14533#issuecomment-1596372007)


eme64:
@XiaohongGong Thanks for looking into this. But it seems to me this is not the same approach as we are taking with x86 SSE and AVX, where the `UseAVX` and `UseSSE` flags affect both the VM features and also the `applyIfCPUFeature` from the IR framework. We check in all sorts of places for `sve` feature, so why do you now only change it for this particular test?

XiaohongGong:
> @XiaohongGong Thanks for looking into this. But it seems to me this is not the same approach as we are taking with x86 SSE and AVX, where the `UseAVX` and `UseSSE` flags affect both the VM features and also the `applyIfCPUFeature` from the IR framework. We check in all sorts of places for `sve` feature, so why do you now only change it for this particular test?
> 
> https://github.com/openjdk/jdk/blob/8d899925dc281c5dabbef14d85a6df807f8d300e/src/hotspot/cpu/x86/vm_version_x86.cpp#L954-L955
> 
> Can you do a similar thing in `src/hotspot/cpu/aarch64/vm_version_aarch64.cpp` ?
> 
> It would be nice not to have to check for the flag and the features in every test, but just for the features. And the features should depend on what is present on the hardware, minus the restrictions by the flags.

Thanks for looking at this PR @eme64 ! Yes, that's the main difference between aarch64 and x86 platforms.  It actually makes things simpler that changing the CPU features based on the vm option.  But per my understanding, CPU features are the hardware's feature which is the objective fact, while the `UseSVE` are the JVM's option that people can set different values. And they cannot be mixed. Besides, x86 just mask off the CPU features for JVM instead of really changing the hardware's features. I'm not sure, but I'm afraid doing such changes like x86 may have some risks in current aarch64's backend. 

> We check in all sorts of places for `sve` feature, so why do you now only change it for this particular test?

For each SVE test, we have tried to add flag `UseSVE=1` in the test's `main` function to make sure this option is not changed by others, and current test is run with the  expected sve feature. For example:
```
        TestFramework.runWithFlags("--add-modules=jdk.incubator.vector",
                                   "-XX:UseSVE=1");
    }
```
For this test, we cannot add such an option in the test file, since it is also used to test other platforms like x86.

eme64:
@XiaohongGong I see, you are worried that it would take a lot of work in the `aarch64` code? So in the backend you are using the `UseSVE` flag instead of feature support?

Are you going to add the `UseSVE` flag to all these cases too?

```
emanuel@emanuel-oracle:/oracle-work/jdk-fork8/open$ grep sve test/hotspot/jtreg/compiler/ -r
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(failOn = IRNode.AND_VB, applyIfCPUFeatureAnd = {"asimd", "true", "sve", "false"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(failOn = IRNode.AND_VS, applyIfCPUFeatureAnd = {"asimd", "true", "sve", "false"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(counts = {IRNode.AND_VI, "1"}, applyIfCPUFeatureOr = {"sve", "true", "avx512", "true"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(counts = {IRNode.AND_VL, "1"}, applyIfCPUFeatureOr = {"sve", "true", "avx512", "true"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(counts = {IRNode.AND_VI, "1"}, applyIfCPUFeatureOr = {"sve", "true", "avx512", "true"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(failOn = IRNode.OR_VB, applyIfCPUFeatureAnd = {"asimd", "true", "sve", "false"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(failOn = IRNode.OR_VB, applyIfCPUFeatureAnd = {"asimd", "true", "sve", "false"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(counts = {IRNode.OR_VI, "1"}, applyIfCPUFeatureOr = {"sve", "true", "avx512", "true"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(counts = {IRNode.OR_VL, "1"}, applyIfCPUFeatureOr = {"sve", "true", "avx512", "true"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(counts = {IRNode.OR_VI, "1"}, applyIfCPUFeatureOr = {"sve", "true", "avx512", "true"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(failOn = IRNode.XOR_VS, applyIfCPUFeatureAnd = {"asimd", "true", "sve", "false"})
test/hotspot/jtreg/compiler/vectorization/runner/LoopArrayIndexComputeTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/LoopArrayIndexComputeTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/LoopArrayIndexComputeTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/LoopArrayIndexComputeTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayTypeConvertTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx512dq", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayTypeConvertTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayTypeConvertTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayTypeConvertTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayTypeConvertTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/BasicLongOpTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx512dq", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/BasicLongOpTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "sse4.1", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/BasicLongOpTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "sse4.1", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/BasicLongOpTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "sse4.1", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayIndexFillTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayIndexFillTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayIndexFillTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayIndexFillTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayIndexFillTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayIndexFillTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayIndexFillTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/LoopReductionOpTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/LoopReductionOpTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
```

I just really don't like the duplication of IR rules, and the checking of both the flag and the cpu feature.

Another solution: we filter out the `sve` feature at the IR framework, if we have `UseSVE=0`?

eme64:
> Besides, x86 just mask off the CPU features for JVM instead of really changing the hardware's features.

That is the question: what is the ground truth. On x86, the idea is to keep the flags and the cpu features in sync. So if the hardware does not support a flag value to be larger (eg trying to set `UseAVX=3` if only avx2 is available, we force it down to `UseAVX=2`). And if a flag restricts cpu features, we mask them off.

Of course, the hardware may support more features than the VM assumes. But still, that allows us to do some "cross cpu" testing - we can simulate avx1 features on a avx2 machine for example.

eme64:
@fg1417 @jatin-bhateja What do you think about the consistency of arm / intel hardware flags for `SVE` and `AVX`?

XiaohongGong:
> @XiaohongGong I see, you are worried that it would take a lot of work in the aarch64 code? So in the backend you are using the UseSVE flag instead of feature support?

Yes, that may need more overhead on AArch64 code. I think the changing should not only limit to `UseSVE` flag and the `sve` cpu feature, but also to all other flags. We have to keep the design consistent. 

And yes, for the current backend, we're almost using `UseSVE` flag instead of the cpu feature check.

> test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(failOn = IRNode.AND_VB, applyIfCPUFeatureAnd = {"asimd", "true", "sve", "false"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(failOn = IRNode.AND_VS, applyIfCPUFeatureAnd = {"asimd", "true", "sve", "false"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(counts = {IRNode.AND_VI, "1"}, applyIfCPUFeatureOr = {"sve", "true", "avx512", "true"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(counts = {IRNode.AND_VL, "1"}, applyIfCPUFeatureOr = {"sve", "true", "avx512", "true"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(counts = {IRNode.AND_VI, "1"}, applyIfCPUFeatureOr = {"sve", "true", "avx512", "true"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(failOn = IRNode.OR_VB, applyIfCPUFeatureAnd = {"asimd", "true", "sve", "false"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(failOn = IRNode.OR_VB, applyIfCPUFeatureAnd = {"asimd", "true", "sve", "false"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(counts = {IRNode.OR_VI, "1"}, applyIfCPUFeatureOr = {"sve", "true", "avx512", "true"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(counts = {IRNode.OR_VL, "1"}, applyIfCPUFeatureOr = {"sve", "true", "avx512", "true"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(counts = {IRNode.OR_VI, "1"}, applyIfCPUFeatureOr = {"sve", "true", "avx512", "true"})
test/hotspot/jtreg/compiler/vectorapi/VectorLogicalOpIdentityTest.java:    @IR(failOn = IRNode.XOR_VS, applyIfCPUFeatureAnd = {"asimd", "true", "sve", "false"})
test/hotspot/jtreg/compiler/vectorization/runner/LoopArrayIndexComputeTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/LoopArrayIndexComputeTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/LoopArrayIndexComputeTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/LoopArrayIndexComputeTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayTypeConvertTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx512dq", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayTypeConvertTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayTypeConvertTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayTypeConvertTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayTypeConvertTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/BasicLongOpTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx512dq", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/BasicLongOpTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "sse4.1", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/BasicLongOpTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "sse4.1", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/BasicLongOpTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "sse4.1", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayIndexFillTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayIndexFillTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayIndexFillTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayIndexFillTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayIndexFillTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayIndexFillTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/ArrayIndexFillTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/LoopReductionOpTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"},
test/hotspot/jtreg/compiler/vectorization/runner/LoopReductionOpTest.java:    @IR(applyIfCPUFeatureOr = {"sve", "true", "avx2", "true"}

I'm afraid we have to modify all these tests, and what I know is my colleague @pfustc will take charge of the vectorization tests  once this PR is merged. Besides, we have to only care about the rules that `sve` is `true` like `"@IR(applyIfCPUFeatureOr = {"sve", "true", ...}`. If the `sve` feature is `false`, `UseSVE` is always `0` in the VM, which is synced.

pfustc:
> I discussed it with @chhagedorn and @TobiHartmann . They also think it is better to have flags and cpu features in sync. Because the flags really should restrict what the VM uses everywhere. We want to make sure the restrictions apply in the VM, if we check for features or flags.

Hi @eme64,

Having flags and cpu features in sync sounds a good idea. However, of all platforms supported by HotSpot, only x86 does in this way. Even on x86, only AVX & SSE related features have such sync at present. There is no sync for other feature strings and flags, like `UseSHA`, `UseAES` and etc. That's the reason we think current approach of x86 is more like a workaround for IR tests only. If you believe keeping this kind of sync is a better approach, I'd suggest doing this for all CPU features on all platforms in another RFE.

theRealAph:
> Thanks for looking at this PR @eme64 ! Yes, that's the main difference between aarch64 and x86 platforms. It actually makes things simpler that changing the CPU features based on the vm option. But per my understanding, CPU features are the hardware's feature which is the objective fact, while the `UseSVE` are the JVM's option that people can set different values. And they cannot be mixed. Besides, x86 just mask off the CPU features for JVM instead of really changing the hardware's features. I'm not sure, but I'm afraid doing such changes like x86 may have some risks in current aarch64's backend.

It might, but it sounds like it's the right thing to do as soon as possible;e after the JDK 21 fork.

theRealAph:
> > I discussed it with @chhagedorn and @TobiHartmann . They also think it is better to have flags and cpu features in sync. Because the flags really should restrict what the VM uses everywhere. We want to make sure the restrictions apply in the VM, if we check for features or flags.

I agree.

> Having flags and cpu features in sync sounds a good idea. However, of all platforms supported by HotSpot, only x86 does in this way. Even on x86, only AVX & SSE related features have such sync at present. There is no sync for other feature strings and flags, like `UseSHA`, `UseAES` and etc. That's the reason we think current approach of x86 is more like a workaround for IR tests only. If you believe keeping this kind of sync is a better approach, I'd suggest doing this for all CPU features on all platforms in another RFE.

That's an interesting suggestion, but having to keep all of the back ends in lock-step is an intolerable constraint.

XiaohongGong:
> > Thanks for looking at this PR @eme64 ! Yes, that's the main difference between aarch64 and x86 platforms. It actually makes things simpler that changing the CPU features based on the vm option. But per my understanding, CPU features are the hardware's feature which is the objective fact, while the `UseSVE` are the JVM's option that people can set different values. And they cannot be mixed. Besides, x86 just mask off the CPU features for JVM instead of really changing the hardware's features. I'm not sure, but I'm afraid doing such changes like x86 may have some risks in current aarch64's backend.
> 
> It might, but it sounds like it's the right thing to do as soon as possible after the JDK 21 fork.

Thanks for looking at this issue @theRealAph. OK, I will take an investigation for this, and try to make a change.

XiaohongGong:
Hi there, I'v filed the vm option and cpu feature sync issue here for AArch64: https://bugs.openjdk.org/browse/JDK-8311130, and will address the comment with it. Thanks again for the advice!

Hi @eme64 , besides the sync issue, does the change to IR framework make sense to you? Currently, if we use an architecture specific vm options with `applyIf` for an IR check, and run the test on another different architecture, the whole test will fail by throwing exceptions, even if we add the `applyIfCPUFeature` to do the cpu check. The changes in the IR framework can fix this issue. 

If that part seems fine to you, maybe we can let this PR in first? Since the test failure will noise our internal ci testing. WDYT? Thanks!

eme64:
@XiaohongGong I totally agree with the changes to the IR framework (having `applyIfCPUFeature` before `applyIf`).

Otherwise, using both `UseSVE=0` and `sve, false` is a temporary fix that should be reverted after [JDK-8311130](https://bugs.openjdk.org/browse/JDK-8311130). I'm accepting it as a temporary fix only. Who will do the real fix?

I was a bit afraid not keeping the CPU feature and the VM flag in sync could also lead to issues in the backend of aarch64. But it does indeed seem that we only use `UseSVE`,  and never `VM_Version::supports_sve()`. Still, someone might use them synonymous in the future and expect that they are in sync.

Actually, since there are only so few uses of `VM_Version::supports_sve()`, is the risk not very low to just mask off the feature now directly with this fix? That fix does not look so complicated as I feared. What do you think?

Anyway, I just launched testing for commit 1: tier1-6 plus stress testing. Will report back on Monday probably.

XiaohongGong:
> @XiaohongGong I totally agree with the changes to the IR framework (having `applyIfCPUFeature` before `applyIf`).

Thanks a lot!

> Otherwise, using both `UseSVE=0` and `sve, false` is a temporary fix that should be reverted after [JDK-8311130](https://bugs.openjdk.org/browse/JDK-8311130). I'm accepting it as a temporary fix only. Who will do the real fix?

We (Arm) will do the real fix. `UseSVE=0` is needed when `sve, true`, which only affects this test now. And yes, I can revert these IR checks once the real fix is in. 

> I was a bit afraid not keeping the CPU feature and the VM flag in sync could also lead to issues in the backend of aarch64. But it does indeed seem that we only use `UseSVE`, and never `VM_Version::supports_sve()`. Still, someone might use them synonymous in the future and expect that they are in sync.

Agree, although we only use `UseSVE` in backend now.

> Actually, since there are only so few uses of `VM_Version::supports_sve()`, is the risk not very low to just mask off the feature now directly with this fix? That fix does not look so complicated as I feared. What do you think?

I prefer fixing that in a separate patch. One reason is syncing the vm options and cpu features is a refactory to AArch64 backend for me. It has other relative cpu features specific to different SVE systems besides `sve`. For example, the `svebitperm` which exists after sve2.  We have to take a consideration for them as well. Besides, although the changes is not so big, we have to do more testing to make sure no regressions are involved. 

And besides the `UseSVE`, do you think it's necessary to sync other options as well?

> Anyway, I just launched testing for commit 1: tier1-6 plus stress testing. Will report back on Monday probably.

Thanks for doing this!

eme64:
@XiaohongGong Testing for commit 1 looks good.

I don't have the expertise on the other CPU features. But from a distance, I'd say yes. The idea with all the CPU features is that we can simulate less powerful machines with more powerful machines. But I would first fix the SVE features, and handle the others separately. Maybe for those a separate discussion needs to happen first.

eme64 (approved):
Accepted as a temporary fix that has to be reverted with [JDK-8311130](https://bugs.openjdk.org/browse/JDK-8311130).

XiaohongGong:
> Accepted as a temporary fix that has to be reverted with [JDK-8311130](https://bugs.openjdk.org/browse/JDK-83T11130).

Thanks for the review and testing! I will revert the IR test part with JDK-8311130.

chhagedorn (approved):
> Accepted as a temporary fix that has to be reverted with [JDK-8311130](https://bugs.openjdk.org/browse/JDK-8311130).

I agree with this temporary fix and reverting it again when syncing the flags with the CPU features in JDK-8311130.

XiaohongGong:
Thanks for the review @chhagedorn ! I'v addressed the comments in latest commit.

chhagedorn (approved):
Thanks for the update, looks good!

PR #14536, opened 2023-06-19
8310299: C2: 8275201 broke constant folding of array store check in some cases

rwestrel:
Before 8275201, loading the element klass of an array returned:

```
TypeKlassPtr::make(tkls->ptr(), elem, 0/*offset*/);
```

that is exact if the array type is exact. I changed it to:

```
tkls->is_aryklassptr()->elem();
```

When the array type is exact (newly allocated array for instance) but
the element class has subclasses, this doesn't return an exact class
(so the logic is different from the one that was there before). That
affects array store checks that no longer constant fold.


TobiHartmann (approved):
Looks good to me.

sviswa7:
@rwestrel Thanks a lot for this fix. It fully recovers the performance drop that we have been observing since JDK 19.

PR #14539, opened 2023-06-19
8310308: IR Framework: check for type and size of vector nodes

eme64:
For some changes to `SuperWord`, and maybe auto-vectorization in general, I want to strengthen the IR Framework.

**Motivation**
I want to not just find the relevant IR nodes, but also assert that they have the maximal length that they could have on the respective platform (given the CPU features and `MaxVectorSize`). Without this verification it is possible that a future change leads to a regression where we still vectorize but at shorter vector widths as before - leading to performance loss.

**How to use it**

All `IRNode`s in `test/hotspot/jtreg/compiler/lib/ir_framework/IRNode.java` that are created with `vectorNode` are now all matched with their `type` and `size`. The regex might now look something like this:

`"(\d+(\s){2}(VectorCastF2X.*)+(\s){2}===.*vector[A-Za-z]\[8\]:\{int\})"`
which would match with IR nodes dumped like that:
`1150  VectorCastF2X  === _ 1151  [[ 1146 ]]  #vectory[8]:{int} ...`

The goal was to keep it simple and straight forward. In most cases, you can just use the nodes as before, and implicitly we now check for maximal size automatically. However, in some cases we want to ensure there is no or only a limited number of nodes (`failOn` or comparison `<` or `<=` or `=0`) - in those cases we usually want to make sure there is not any node of any size, so we match with any size by default. The size can also explicitly be constrained using `IRNode.VECTOR_SIZE`.

Some examples:
1. `@IR(counts = {IRNode.LOAD_VECTOR_I,  " >0 "})` -> search for a `LoadVector` node with `type` `int`, and maximal `size` possible on the machine (limited by CPU features and `MaxVectorSize`). This is the most common use case.
2. `@IR(failOn = { IRNode.LOAD_VECTOR_L, IRNode.STORE_VECTOR })` -> fail if there is a `LoadVector` with type `long`, of `any` size.
3. `@IR(counts = { IRNode.XOR_VI, IRNode.VECTOR_SIZE_4, " > 0 "})` -> find at least one `XorV` node with type `int` and exactly `4` elements. Useful for VectorAPI when the vector species is fixed.
4. `@IR(counts = { IRNode.LOAD_VECTOR_D, IRNode.VECTOR_SIZE + "min(4, max_double)", " >0 " })` -> search for a `LoadVector` node with `type` `double`, and `size` exactly equals to `min(4, max_double)` (so 4 elements, or if the hardware allows fewer `doubles`, then that number).
5. `@IR(counts = { IRNode.ABS_VF, IRNode.VECTOR_SIZE + "min(LoopMaxUnroll, max_float)", ">= 1" })` -> find at least one `AbsV` nodes with type `float`, and the `size` exactly equals to the smaller of `LoopMaxUnroll` or the maximal size allowed for `floats` (useful for tests where the `LoopMaxUnroll` is artificially lowered, which sometimes prevents the maximal filling of vectors).
6. `@IR(counts = {IRNode.VECTOR_CAST_I2F, IRNode.VECTOR_SIZE + "min(max_int, max_float)", ">0"})` -> find at least one `VectorCastI2X` node that casts to type `float`, and where the size is exactly equals to the smaller maximal size for `ints` and `floats`. This is helpful when there are multiple types in the loop, and the number of elements is limited by the sizes of multiple types.

I had to change lots of occurrences, hence you can find many more examples in the tests.

**Details**

Vector nodes that should be tested for `type` and `size` now are to be created with `VECTOR_PREFIX` and `vectorNode`, see `IRNode.java`.

When specifying such a `vectorNode` in an IR rule, one first uses the `irNodePlaceholder` (eg `Load_VECTOR_I`), and following it one can optionally add a `IRNode.VECTOR_SIZE` specifier,  which is then parsed by `parseVectorNodeSize`. This allows either naming a concrete size (eg `IRNode.VECTOR_SIZE_8`), a tag (`IRNode.VECTOR_SIZE + "<tag>"`) where the the tag can be one of the tags listed in `parseVectorNodeSizeTag`, or a `min(...)` clause which computes the minimum value of a comma separated list of tags. As a last resort one can match for any size (`IRNode.VECTOR_SIZE_ANY`).

The maximal vector size for any type is computed in `getMaxElementsForType`, under consideration of the CPU features and the `MaxVectorSize`.

**Changes to tests**

Unfortunately, I had to change a lot of IR rules, though not substantially. Most changes are because we usually had nodes like `MAX_V` or `LOAD_VECTOR` which matched for any type, and I had to create one node per type now (eg `MAX_VF, MAX_VD`, or `LOAD_VECTOR_I, LOAD_VECTOR_L, LOAD_VECTOR_F, ...`). While this was a lot of work, it is still good to know that we are generating the nodes with the correct types.

In the VectorAPI tests there were many which required concrete sizes due to the concrete size of the vector species. This is nice to test, since it guarantees that the vector species indeed generate the expected vector sizes.

A few tests required more attention, where I had to use patterns like `IRNode.VECTOR_SIZE + "min(...)"`. These are especially interesting, as they test cases like mixed types (eg casting between types).

A few tests had loop iteration counts that were too small (maybe 512), such that the loops were not sufficiently unrolled to reach the maximal vector width. This happened especially with byte cases, which require an unrolling factor of 64 to fill 512bit registers. I increased the loop iteration counts, such that we can also properly test the largest vector widths. This improves our test coverage.

**Future Work**

There are a few nodes that I did not yet handle with `vectorNode` (eg `VECTOR_REINTERPRET`, `OR_V_MASK`, `MACRO_LOGIC_V`, `LOAD_VECTOR_GATHER(_MASKED)`). Some of these only have very few tests and are all from the Vector API which was not my priority here. They can easily be converted should the need arise in the future.

While looking at lots of IR tests I also came up with these RFE's:
[JDK-8310891](https://bugs.openjdk.org/browse/JDK-8310891) C2 SuperWord tests: move platform requirements to IR rules
[JDK-8310523](https://bugs.openjdk.org/browse/JDK-8310523) Add IR tests for nodes that have too few IR tests yet
[JDK-8310533](https://bugs.openjdk.org/browse/JDK-8310533) [IR Framework] Add possibility to automatically verify that a test method always returns the same result

**Testing**

tier1-tier6 and stress-testing.

---------
### Progress

### Issue
 * [JDK-8310308](https://bugs.openjdk.org/browse/JDK-8310308): IR Framework: check for type and size of vector nodes (**Enhancement** - P4)

### Reviewers
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - **Reviewer**) ⚠️ Review applies to [ca0b576f](https://git.openjdk.org/jdk/pull/14539/files/ca0b576f9c4cd1c4cbe16f0bb40071cad35e45a0)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14539/head:pull/14539` \
`$ git checkout pull/14539`

`$ git checkout pull/14539` \
`$ git pull https://git.openjdk.org/jdk.git pull/14539/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14539`

`$ git pr show -t 14539`

</details>

<a href="https://git.openjdk.org/jdk/pull/14539.diff">https://git.openjdk.org/jdk/pull/14539.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14539#issuecomment-1609352391)


chhagedorn (commented):
This is a great enhancement! Thanks for working on that.

I have left some comments in the IR framework code (will also have a look at the test updates later) but here are some more general comments:

- We should provide some better description when misusing the new feature. Example:
```
@IR(counts = {IRNode.LOAD_VL, IRNode.VECTOR_SIZE + "min(4)", ">0"})
@IR(counts = {IRNode.LOAD_VL, IRNode.VECTOR_SIZE + "min()", ">0"})
```
Output:
```
 - Provided invalid value "_@min(4)" after comparator "=", node IRNode.LOAD_VL, in count string "_@min(4)" for IR rule 2 at private static long compiler.loopopts.superword.TestGeneralizedReductions.testReductionOnPartiallyUnrolledLoopWithSwappedInputs(long[]).
 - Provided invalid value "_@min()" after comparator "=", node IRNode.LOAD_VL, in count string "_@min()" for IR rule 3 at private static long compiler.loopopts.superword.TestGeneralizedReductions.testReductionOnPartiallyUnrolledLoopWithSwappedInputs(long[]).
```
It's not evidently clear what's wrong here or how a user can correct this. You might want to play around with other wrong usages of the new features and check if the format violation is precise enough. You could also add these wrong usages to `TestBadFormat.java`.
- We should have a (sanity) test that explicitely uses `IRNode.VECTOR_SIZE_ANY` and `IRNode.VECTOR_SIZE_MAX`.
- We should also make sure to have some sanity tests for all the different variations that are now possible with the new features (if not already covered by your updated tests).

test/hotspot/jtreg/compiler/c2/TestMinMaxSubword.java
```diff
@@ -61,7 +61,7 @@ public class TestMinMaxSubword {
     // as Java APIs for Math.min/max do not support integer subword types and superword
     // should not generate vectorized Min/Max nodes for them.
     @Test
-    @IR(failOn = {IRNode.MIN_V})
+    @IR(failOn = {IRNode.MIN_VI, IRNode.MIN_VF, IRNode.MIN_VD})
```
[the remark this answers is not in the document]
eme64: I'm not sure it is worth it now. There are very few use cases, this is one of them.

test/hotspot/jtreg/compiler/lib/ir_framework/driver/irmatching/irrule/IRRule.java
```diff
@@ -44,10 +45,10 @@ public class IRRule implements Matchable {
     private final IR irAnno;
     private final MatchableMatcher matcher;
 
-    public IRRule(int ruleId, IR irAnno, Compilation compilation) {
+    public IRRule(int ruleId, IR irAnno, Compilation compilation, VMInfo vmInfo) {
         this.ruleId = ruleId;
         this.irAnno = irAnno;
-        this.matcher = new MatchableMatcher(new CompilePhaseIRRuleBuilder(irAnno, compilation).build());
+        this.matcher = new MatchableMatcher(new CompilePhaseIRRuleBuilder(irAnno, compilation).build(vmInfo));
```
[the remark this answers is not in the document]
eme64: For consistency with what exactly? How would I pass it on/down from the `CompilePhaseIRRuleBuilder` constructor?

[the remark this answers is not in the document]
chhagedorn: In the IR framework, I'm usually passing the required information to the constructor of the builder class and then just call `build()`. You could also do the same here since you otherwise need to pass the `vmInfo` around inside the `CompilePhaseIRRuleBuilder` class to the different private methods.

eme64:
Thanks @chhagedorn , I addressed some of your feedback already. More to come tomorrow.

test/hotspot/jtreg/compiler/lib/ir_framework/IRNode.java
```diff
@@ -178,22 +246,22 @@ public class IRNode {
 
     public static final String ADD_REDUCTION_VD = PREFIX + "ADD_REDUCTION_VD" + POSTFIX;
     static {
-        superWordNodes(ADD_REDUCTION_VD, "AddReductionVD");
+        beforeMatchingNameRegex(ADD_REDUCTION_VD, "AddReductionVD");
```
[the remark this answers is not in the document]
eme64: Actually, the Vector API can also generate these nodes, and then they would already exist at parsing. I wanted to make these nodes vectorNodes as well, but they do not have a vector output, only a vector input. Hence I cannot match their size.
We currently do not have any reduction IR tests for the Vector API. I tracked it in [JDK-8310523](https://bugs.openjdk.org/browse/JDK-8310523).

Actually, the reduction nodes are the only `superWordNodes`. I think we can eventually remove `superWordNodes` completely.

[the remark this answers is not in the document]
eme64: Still, I reverted them to `superWordNodes` for now.

test/hotspot/jtreg/compiler/lib/ir_framework/IRNode.java
```diff
@@ -1747,6 +2218,151 @@ public static boolean isCompositeIRNode(String irCompositeNodeString) {
...
+        }
+        // Try to parse any tags, convert to comma separated list of ints
+        sizeString = parseVectorNodeSizeTag(sizeString, typeString, vmInfo);
+        // Parse comma separated list of numbers
+        String[] sizes = sizeString.split(",");
+        String regex = "";
+        for (int i = 0; i < sizes.length; i++) {
+            int s = 0;
+            try {
+                s = Integer.parseInt(sizes[i]);
```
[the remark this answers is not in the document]
eme64: I'm reporting an error on negative numbers now. But I'm not sure I want to limit it to power of 2 or a size limit. Especially when uses flags/tags this could not always hold.

eme64:
@chhagedorn I went through all your review suggestions. The changes are pushed. Rerunning testing...

chhagedorn (commented):
Thanks for doing the updates! I have some more comments. I've also went through the test changes which look good.

test/hotspot/jtreg/compiler/c2/irTests/TestVectorConditionalMove.java
```diff
@@ -203,7 +203,10 @@ private double cmoveDGTforD(double a, double b, double c, double d) {
 
     // Compare 2 values, and pick one of them
     @Test
-    @IR(counts = {IRNode.LOAD_VECTOR, ">0", IRNode.VECTOR_MASK_CMP, ">0", IRNode.VECTOR_BLEND, ">0", IRNode.STORE_VECTOR, ">0"},
+    @IR(counts = {IRNode.LOAD_VECTOR_F, ">0",
+                  IRNode.VECTOR_MASK_CMP_F, ">0",
+                  IRNode.VECTOR_BLEND_F, ">0",
+                  IRNode.STORE_VECTOR, ">0"},
```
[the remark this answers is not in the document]
chhagedorn: Right :-) I've filed [JDK-8311238](https://bugs.openjdk.org/browse/JDK-8311238) to keep track of that.

chhagedorn (approved):
Looks good, thanks for doing all the updates!

test/hotspot/jtreg/compiler/lib/ir_framework/driver/irmatching/irrule/constraint/raw/RawCountsConstraint.java
```diff
@@ -52,9 +56,42 @@ public CompilePhase defaultCompilePhase() {
...
+                                         comparison.getGivenValue() + "\" is useless, please only use positive numbers.");
+                return true; // max
+            }
+            case ">=" -> {
+                TestFormat.checkNoReport(comparison.getGivenValue() > 0, "Node count comparison \">=" +
+                                         comparison.getGivenValue() + "\" is useless, please only use strictly positive numbers with greater-equal.");
+                return true; // max
+            }
+            case "!=" -> throw new TestFormatException("Not-equal comparator not supported for node count: \"" +
+                                                       comparison.getComparator() + "\". Please rewrite the rule.");
```
TobiHartmann, line 86: No need to call `comparison.getComparator()` again here.

test/hotspot/jtreg/compiler/lib/ir_framework/driver/irmatching/irrule/constraint/raw/RawCountsConstraint.java
```diff
@@ -52,9 +56,42 @@ public CompilePhase defaultCompilePhase() {
         return rawIRNode.defaultCompilePhase();
     }
 
+    private boolean expectMaxSizeForVectorNode() {
+        switch (comparison.getComparator()) {
+            case "<" -> {
+                TestFormat.checkNoReport(comparison.getGivenValue() > 1, "Node count comparison \"<" +
```
TobiHartmann, line 62: What if `comparison.getGivenValue()` is negative?

test/hotspot/jtreg/compiler/lib/ir_framework/driver/irmatching/parser/VMInfo.java
```diff
@@ -0,0 +1,71 @@
...
+    public String getStringValue(String key) {
+        TestFramework.check(isKey(key), "VMInfo does not contain \"" + key + "\"");
+        return keyValueMap.get(key);
+    }
+
+    public long getLongValue(String key) {
+        try {
+            return Long.parseLong(getStringValue(key));
+        } catch (NumberFormatException e) {
+            throw new TestFrameworkException("VMInfo value for \"" + key + "\" is not long, got \"" + getStringValue(key) + "\"");
```
TobiHartmann, line 59:
```suggestion
            throw new TestFrameworkException("VMInfo value for \"" + key + "\" is not a long, got \"" + getStringValue(key) + "\"");
```

TobiHartmann (approved):
Nice enhancement! I skimmed through the changes and it looks good to me.

test/hotspot/jtreg/compiler/lib/ir_framework/driver/irmatching/irrule/constraint/raw/RawCountsConstraint.java
eme64: A rule like that would always fail.

eme64: Will add more checks to give more precise messages.

eme64:
Thanks @chhagedorn for all the help with getting this to designed, debugged and reviewed!
@TobiHartmann for the reviews and suggestions!
I'm now running some last pre-integration tests and will integrate ASAP to avoid more merge conflicts.

eme64:
Ah, it turns out on our testing infrastructure it looks all green, but on GitHub actions I have 36 failing tests. And I think they are almost all related. Will need to investigate.

eme64:
The issue is that CascadeLake has `MaxVectorSize=64` , but internally only uses 32 byte. Except if you explicitly set `-XX:MaxVectorSize=64` (not it is no longer "default_cascade_lake"), then it actually uses 64 byte. This makes it impossible to just read `MaxVectorSize`. If the value is 64, the VM may now use 32 byte (default) or 64 byte (if explicitly set).

test/hotspot/jtreg/compiler/lib/ir_framework/driver/irmatching/irrule/checkattribute/parsing/RawIRNode.java
```diff
@@ -51,14 +55,85 @@ public CompilePhase defaultCompilePhase() {
...
+            if (!vmInfo.canTrustVectorSize()) {
+                // Parse it already just to get errors before we overwrite size
+                IRNode.parseVectorNodeSize(size, type, vmInfo);
+            }
+        }
+
+        // Set default values in some cases:
+        if (!userPostfix.isValid() || !vmInfo.canTrustVectorSize()) {
+            switch (bound) {
+                case Comparison.Bound.LOWER -> {
```
chhagedorn, line 92:
```suggestion
                case LOWER -> {
```

test/hotspot/jtreg/compiler/lib/ir_framework/driver/irmatching/irrule/checkattribute/parsing/RawIRNode.java
```diff
@@ -51,14 +55,85 @@ public CompilePhase defaultCompilePhase() {
...
+                        // No size specified, so assume maximal size
+                        size = IRNode.VECTOR_SIZE_TAG_MAX;
+                    } else {
+                        System.out.println("WARNING: you are on a system with \"canTrustVectorSize == false\" (default Cascade Lake).");
+                        System.out.println("         The lower bound rule for \"" + node + "\" is now performed with");
+                        System.out.println("         \"IRNode.VECTOR_SIZE_TAG_ANY\" instead of \"IRNode.VECTOR_SIZE_TAG_MAX\".");
+                        size = IRNode.VECTOR_SIZE_TAG_ANY;
+                    }
+                }
+                case Comparison.Bound.UPPER -> {
```
chhagedorn, line 105:
```suggestion
                case UPPER -> {
```

test/hotspot/jtreg/compiler/lib/ir_framework/driver/irmatching/irrule/checkattribute/parsing/RawIRNode.java
```diff
@@ -51,14 +55,85 @@ public CompilePhase defaultCompilePhase() {
...
+                    }
+                }
+                case Comparison.Bound.EQUAL -> {
+                    if (vmInfo.canTrustVectorSize()) {
+                        // No size specified, so assume maximal size
+                        size = IRNode.VECTOR_SIZE_TAG_MAX;
+                    } else {
+                        // Equal comparison to a strictly positive number would lead us to an impossible
+                        // situation: we might have to know the exact vector size or else we count too many
+                        // or too few cases. Because of this we forbid this case in general.
```
chhagedorn, line 127: The comment is outdated and could, for example, be updated to:
```suggestion
                        // or too few cases. We therefore skip such a constraint and treat it as success.
```

test/hotspot/jtreg/compiler/lib/ir_framework/driver/irmatching/irrule/constraint/SuccessConstraintCheck.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.List;
+
+/**
+ * This class provides a check that always succeeds.
+ *
+ * @see Constraint
+ */
+class SuccessConstraintCheck implements ConstraintCheck {
+    public SuccessConstraintCheck() {}
```
chhagedorn, line 38: This default constructor can be removed since it's implicitly added for us.

```suggestion
```

test/hotspot/jtreg/compiler/lib/ir_framework/driver/irmatching/parser/VMInfo.java
```diff
@@ -0,0 +1,119 @@
...
+        System.err.println("MaxVectorSize: " + getLongValue("MaxVectorSize"));
+        System.err.println("MaxVectorSizeIsDefault: " + getLongValue("MaxVectorSizeIsDefault"));
+        System.err.println("LoopMaxUnroll: " + getLongValue("LoopMaxUnroll"));
+        System.err.println("UseAVX: " + getLongValue("UseAVX"));
+        System.err.println("UseAVXIsDefault: " + getLongValue("UseAVXIsDefault"));
+        if (isDefaultCascadeLake()) {
+            System.err.println(" -> You are on default Cascade Lake");
+            System.err.println("   -> SuperWord expected to run with 32 byte, not 64 byte, VectorAPI expected to use 64 byte");
+            System.err.println("   -> \"canTrustVectorSize == false\", some vector node IR rules are made weaker.");
+        }
```
chhagedorn, line 67: Is this a leftover from debugging? If you want to print this information for debugging purposes, I suggest to move this code to `VMInfoParser` and additionally guard it with `VERBOSE || PRINT_IR_ENCODING`. The name `PRINT_IR_ENCODING` is not completely correct here but we might want to clean this up separately at some other point in time.

You can keep the verification of calling the `get*()` methods here, though.

test/hotspot/jtreg/compiler/lib/ir_framework/driver/irmatching/parser/VMInfo.java
```diff
@@ -0,0 +1,119 @@
...
+        return isCascadeLake() &&
+               getLongValue("MaxVectorSizeIsDefault") == 1 &&
+               getLongValue("UseAVXIsDefault") == 1 &&
+               getLongValue("UseAVX") > 2;
+    }
+
+    /**
+     * Some platforms do not behave as expected, and one cannot trust that the vectors
+     * make use of the full MaxVectorSize. For Cascade Lake we by default only use
+     * 32 bytes for SuperWord even though MaxVectorSize is 64. But the VectorAPI still
```
chhagedorn, line 108: Suggestion: For Cascade Lake, we only use 32 bytes for SuperWord by default even though MaxVectorSize is 64.

chhagedorn (approved):
I have only some minor comments left, otherwise, the update looks good to me!

test/hotspot/jtreg/compiler/lib/ir_framework/driver/irmatching/parser/VMInfo.java
eme64: I can just remove it. It is not necessary any more I think.

TobiHartmann (approved):

chhagedorn (approved):
Thanks for the updates, looks good!

eme64:
Thanks @TobiHartmann @chhagedorn for all the discussions, help and reviews!
/integrate

PR #14540, opened 2023-06-19
8310311: Serial: move Generation::contribute_scratch to DefNewGeneration

albertnetymk:
Simple cleanup to move a method from base-class to  sub-class.

Test: hotspot_gc

---------
### Progress

### Issue
 * [JDK-8310311](https://bugs.openjdk.org/browse/JDK-8310311): Serial: move Generation::contribute_scratch to DefNewGeneration (**Enhancement** - P4)

### Reviewers
 * [Kim Barrett](https://openjdk.org/census#kbarrett) (@kimbarrett - **Reviewer**) ⚠️ Review applies to [04d3a6b0](https://git.openjdk.org/jdk/pull/14540/files/04d3a6b0eaffcef2f12eba9710393679e40cdbc9)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14540/head:pull/14540` \
`$ git checkout pull/14540`

`$ git checkout pull/14540` \
`$ git pull https://git.openjdk.org/jdk.git pull/14540/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14540`

`$ git pr show -t 14540`

</details>

<a href="https://git.openjdk.org/jdk/pull/14540.diff">https://git.openjdk.org/jdk/pull/14540.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14540#issuecomment-1597081677)

### DIFF
--- a/src/hotspot/share/gc/serial/defNewGeneration.cpp
+++ b/src/hotspot/share/gc/serial/defNewGeneration.cpp
@@ -981,27 +981,18 @@ bool DefNewGeneration::no_allocs_since_save_marks() {
   return to()->saved_mark_at_top();
 }
 
-void DefNewGeneration::contribute_scratch(ScratchBlock*& list, Generation* requestor,
-                                         size_t max_alloc_words) {
-  if (requestor == this || _promotion_failed) {
+void DefNewGeneration::contribute_scratch(void*& scratch, size_t& num_words) {
+  if (_promotion_failed) {
     return;
   }
-  assert(GenCollectedHeap::heap()->is_old_gen(requestor), "We should not call our own generation");
 
-  /* $$$ Assert this?  "trace" is a "MarkSweep" function so that's not appropriate.
-  if (to_space->top() > to_space->bottom()) {
-    trace("to_space not empty when contribute_scratch called");
-  }
-  */
+  constexpr static size_t MinFreeScratchWords = 100;
 
   ContiguousSpace* to_space = to();
-  assert(to_space->end() >= to_space->top(), "pointers out of order");
-  size_t free_words = pointer_delta(to_space->end(), to_space->top());
+  const size_t free_words = pointer_delta(to_space->end(), to_space->top());
   if (free_words >= MinFreeScratchWords) {
-    ScratchBlock* sb = (ScratchBlock*)to_space->top();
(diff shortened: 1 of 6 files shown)


albertnetymk:
> if GenCollectedHeap::young_gen() returned a DefNewGeneration right away;

I tried that but it affects too much other code, so I went with the current approach.

> This also removes the need for some helpers for the scratch space, like sorting the list (which is always very degenerate anyway).

I was planning to do that in a followup PR. The current PR is mostly a structural-change affecting both young & old-gen, while further cleanup is specific to young-gen, so I decided to not merge them into one PR.

src/hotspot/share/gc/serial/defNewGeneration.cpp
```diff
@@ -981,27 +981,18 @@ bool DefNewGeneration::no_allocs_since_save_marks() {
...
     return;
   }
-  assert(GenCollectedHeap::heap()->is_old_gen(requestor), "We should not call our own generation");
 
-  /* $$$ Assert this?  "trace" is a "MarkSweep" function so that's not appropriate.
-  if (to_space->top() > to_space->bottom()) {
-    trace("to_space not empty when contribute_scratch called");
-  }
-  */
+  constexpr static size_t MinFreeScratchWords = 100;
```
tschatzl, line 989: Is there a reason this needs to be a `constexpr static` when a simple `const` seems to be sufficient?

albertnetymk: Dropped `static`. `constexpr` makes it more a num literal. (The generated release asm is identical anyway.)

kimbarrett: I think constexpr doesn't do anything here beyond what const would do. I'm not sure about this, but
I think I'd prefer to reserve the use of constexpr only where it (might) matter.

albertnetymk: Changed to const, but I still believe constexpr should be used for all compile-time known values.

tschatzl (approved):

albertnetymk:
Thanks for the review.

/integrate

PR #14543, opened 2023-06-19
8310239: Add missing cross modifying fence in nmethod entry barriers

fisk:
In fact, there is a current race in the nmethod entry barriers, where what we are doing violates the AMD APM (cf. APM volume 2 section 7.6.1 https://www.amd.com/system/files/TechDocs/24593.pdf). 
In particular, if the compare instruction of the nmethod entry barrier is not yet patched and we call a slow path on thread 1, then before taking the nmethod entry lock, another thread 2 could fix and disarm the nmethod. Then thread 1 will observe *data* suggesting the nmethod has been patched, but never re-executes the patched compare (which might indeed still be stale), hence not qualifying for asynchronous cross modifying code, and neither do we run a serializing cpuid instruction, qualifying for synchronous cross modifying code. In this scenario, we can indeed start executing the nmethod instructions, while observing inconsistent concurrent patching effects, where some instructions will be updated and some not.

The following patch ensures that x86 nmethod entry barriers execute cross modifying fence after calling into the VM, where another thread could have disarmed the nmethod. I also ensured the other platforms perform their fencing after the VM call, instead of before - including a cross_modify_fence in the shared code for OSR nmethod entries. While fencing before will flush out the instruction pipeline, and it shouldn't be populated with problematic instructions until after we start executing the nmethod again, it feels unnecessary to fence on the wrong side of the modifications it wishes to guard, and hence not strictly following the synchronous cross modifying fence recipe.

I'm currently running tier1-5 and running performance testing in aurora. In the interest of time, I'm opening this PR before getting the final result, and will report the results when they come in.

---------
### Progress

### Issue
 * [JDK-8310239](https://bugs.openjdk.org/browse/JDK-8310239): Add missing cross modifying fence in nmethod entry barriers (**Bug** - P2)

### Reviewers
 * [Axel Boldt-Christmas](https://openjdk.org/census#aboldtch) (@xmas92 - **Reviewer**) ⚠️ Review applies to [f3c15b91](https://git.openjdk.org/jdk/pull/14543/files/f3c15b9121ddcc186e2b6d81e210de3e720d074f)
 * [Dean Long](https://openjdk.org/census#dlong) (@dean-long - **Reviewer**) ⚠️ Review applies to [f3c15b91](https://git.openjdk.org/jdk/pull/14543/files/f3c15b9121ddcc186e2b6d81e210de3e720d074f)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14543/head:pull/14543` \
`$ git checkout pull/14543`

`$ git checkout pull/14543` \
`$ git pull https://git.openjdk.org/jdk.git pull/14543/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14543`

`$ git pr show -t 14543`

</details>

<a href="https://git.openjdk.org/jdk/pull/14543.diff">https://git.openjdk.org/jdk/pull/14543.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14543#issuecomment-1597393757)


src/hotspot/cpu/x86/stubGenerator_x86_32.cpp
```diff
@@ -3818,6 +3818,11 @@ class StubGenerator: public StubCodeGenerator {
     __ cmpl(rax, 1); // 1 means deoptimize
     __ jcc(Assembler::equal, deoptimize_label);
 
+    // In case a concurrent thread disarmed the nmethod, we need to ensure the new instructinos
```
dholmes-ora, line 3821: Typo: instructinos

src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
```diff
@@ -3163,6 +3163,11 @@ address StubGenerator::generate_method_entry_barrier() {
   __ cmpl(rax, 1); // 1 means deoptimize
   __ jcc(Assembler::equal, deoptimize_label);
 
+  // In case a concurrent thread disarmed the nmethod, we need to ensure the new instructinos
```
dholmes-ora, line 3166: Typo: instructinos

fisk: Thanks for spotting that! Fixed.

fisk:
Performance testing shows no significant difference, and tier1-5 looks good.

src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp
```diff
@@ -5488,6 +5474,23 @@ class StubGenerator: public StubCodeGenerator {
 
     __ reset_last_Java_frame(true);
 
+    BarrierSetAssembler* bs_asm = BarrierSet::barrier_set()->barrier_set_assembler();
+
+    // In case a concurrent thread disarmed the nmethod, we need to ensure the new instructions
```
theRealAph, line 5479:
```suggestion
    // In case a concurrent thread disarmed this nmethod, we need to ensure the new instructions
```
I think you mean this nmethod?

xmas92 (approved):
This fix looks sound with respect to the AMD specification. 

Given the nature of the bug and how hard it would be to reproduce, I can only hope this is the cause. 
However having this fixed makes it easier to reason about what can happen.

fisk:
Thanks for the review @xmas92!

merykitty:
x86 also has `serialize` which is a bare minimal serializing instruction, would that make any difference here? Thanks a lot.

fisk:
> x86 also has `serialize` which is a bare minimal serializing instruction, would that make any difference here? Thanks a lot.

According to internal perf testing, this slow path call is simply too cold to show up in profiles really.

fisk:
I'm about to change my focus to guarding my garden against blood thirsty deer. So I'll delegate integration to somebody else. /integrate delegate

fisk:
So I survived the deer. Any takers on this one?

dean-long:
At first glance, this seems like an optimization to me, and not a correctness issue.  At worst, we continue to execute the stale compare and go through the slow path unnecessarily.  I don't think we are violating AMD APM, at least for the compare for the entry barrier.  However, there may be a problem if what that barrier is protecting, the nmethod code, was patched in a way that is not concurrent-execution-safe.  If that is the case, or just to be safe, then executing a serializing instruction seems like a good idea.  But we don't do that for all threads.  For any thread that sees the patched instruction, we will just take the fast path and not executed the serializing instruction.  Does visibility of the patched entry point instruction as code imply all patching to the nmethod is also visible, as both data and instructions?  I'm not convinced it does.

theRealAph:
Do we need an ISB on AArch64-specifc code? There, the guard value is data, not an immediate field.

dean-long:
I was thinking the issue Erik was worried about is patching of nmethod code as data at safepoints.  I seem to recall there are certain kinds of stubs that need to patch more than one word, so it is done at a safepoint.  I see code in SafepointMechanism::update_poll_values() that calls OrderAccess::cross_modify_fence() in certain cases, so maybe that's  enough?

theRealAph:
> I was thinking the issue Erik was worried about is patching of nmethod code as data at safepoints. I seem to recall there are certain kinds of stubs that need to patch more than one word, so it is done at a safepoint. I see code in SafepointMechanism::update_poll_values() that calls OrderAccess::cross_modify_fence() in certain cases, so maybe that's enough?

It certainly should be.

fisk:
> At first glance, this seems like an optimization to me, and not a correctness issue. At worst, we continue to execute the stale compare and go through the slow path unnecessarily. I don't think we are violating AMD APM, at least for the compare for the entry barrier. However, there may be a problem if what that barrier is protecting, the nmethod code, was patched in a way that is not concurrent-execution-safe. If that is the case, or just to be safe, then executing a serializing instruction seems like a good idea. But we don't do that for all threads. For any thread that sees the patched instruction, we will just take the fast path and not executed the serializing instruction. Does visibility of the patched entry point instruction as code imply all patching to the nmethod is also visible, as both data and instructions? I'm not convinced it does.

Unfortunately, it is not just an optimization. For example, ZGC has always patched code oop relocations in the nmethod entry barriers on x86_64, which is guarded by the nmethod entry barrier. The assumption is that if the nmethod immediate oops are patched first, and the guard value (immediate of the cmp instruction) is patched after, then if a thread sees the new cmp instruction, it will also see the new oop immediates. And that is indeed what the "asynchronous" cross modifying code description ensures will work in the AMD APM. So that all checks out.
What doesn't check out though unfortunately, is when the opposite happens and we see the old compare value, and we take a slow path. After the slow path has executed, we don't re-execute the guarding cmp instruction. Instead we enter the nmethod, assuming that surely that should be safe after taking the slow path. But as the cmp isn't re-executed, this slow path code only uses *data* to communicate that the guard value has been updated. That is precisely what is described as "synchronous" cross modifying code in the AMD APM. Another thread could have *just* flipped that data to look fine concurrently. We load some the data that signals that the code has been updated, and then need to either fall back and re-execute the cmp to guard the code with the asynchronous behaviour, or issue a cross-modifying code fence, to ensure that the patched oop immediates become visible.
In summary, the fast-path is asynchronous cross-modifying code, and it's fine without the fence. But the slow path uses synchronous cross-modifying code, but without correctly using cross modification fence, which is required for correctness.
Now, this problem becomes bigger with generational ZGC, because we patch more things than just oop immediates, but the principle and problem is precisely the same.

fisk:
> Do we need an ISB on AArch64-specifc code? There, the guard value is data, not an immediate field.
> In other words, what instruction has just been patched that we need to make visible?

On AArch64 we only use synchronous cross-modifying code, we just hide the expensive in slow paths using a epoch trick that proves that most executions don't need a fence. So that should all be fine. Sometimes I wonder if we should use that trick on x86_64 as well.

fisk:
Any takers?

theRealAph:
> Any takers?

Sure, when you've answered my question... :-)

theRealAph:
> > Do we need an ISB on AArch64-specifc code? There, the guard value is data, not an immediate field.
> > In other words, what instruction has just been patched that we need to make visible?
> 
> On AArch64 we only use synchronous cross-modifying code, we just hide the expensive in slow paths using a epoch trick that proves that most executions don't need a fence. So that should all be fine. Sometimes I wonder if we should use that trick on x86_64 as well.

I don't understand this reply. On AArch64 we don't patch code, we patch data. So why do we need to add a missing ISB to AArch64?

fisk:
> > > Do we need an ISB on AArch64-specifc code? There, the guard value is data, not an immediate field.
> 
> > > In other words, what instruction has just been patched that we need to make visible?
> 
> > 
> 
> > On AArch64 we only use synchronous cross-modifying code, we just hide the expensive in slow paths using a epoch trick that proves that most executions don't need a fence. So that should all be fine. Sometimes I wonder if we should use that trick on x86_64 as well.
> 
> 
> 
> I don't understand this reply. On AArch64 we don't patch code, we patch data. So why do we need to add a missing ISB to AArch64?

Generational ZGC patches code too.

dean-long:
Wouldn't it be better to put a cross_modify_fence() at the end of BarrierSetNMethod::nmethod_entry_barrier()?  I don't see any code patching after that.  Then we don't need it in the platform-specific generate_method_entry_barrier().  And the cross_modify_fence() could be condition depending on if code was actually patched, which it sounds like doesn't happen on aarch64 unless Generational ZGC is used?

dean-long:
> The assumption is that if the nmethod immediate oops are patched first, and the guard value (immediate of the cmp instruction) is patched after, then if a thread sees the new cmp instruction, it will also see the new oop immediates.   And that is indeed what the "asynchronous" cross modifying code description ensures will work in the AMD APM. So that all checks out.

I guess this is a separate issue from this patch, but where does the AMD APM guarantee that?

dean-long:
In particular, I'm wondering about branch prediction.  The "beyond reach of the thread's current control flow" to me only rules out pre-fetching code that is truly unreachable (ignoring unconditional branches).  What about this scenario:
1) Thread 1 reaches 1st instruction of nmethod, and predicts that the entry barrier slow path branch will not be taken, so it loads the some number of instructions past the branch into the pipeline, including instructions with oop immediates.
2) Before Thread 1 reaches the entry barrier compare, another thread calls the same nmethod, takes the slow path, patches oops in instructions, and disarms the entry barrier
3) Thread 1 sees the disarmed conditional branch and continues to execute the previously fetched pipeline which contains stale oops.

fisk:
> Wouldn't it be better to put a cross_modify_fence() at the end of BarrierSetNMethod::nmethod_entry_barrier()? I don't see any code patching after that. Then we don't need it in the platform-specific generate_method_entry_barrier(). And the cross_modify_fence() could be condition depending on if code was actually patched, which it sounds like doesn't happen on aarch64 unless Generational ZGC is used?

That's a good point. Perhaps BarrierSetNMethod::nmethod_stub_entry_barrier is the better place though, as we sometimes call the nmethod entry barrier from the runtime.

fisk:
> > The assumption is that if the nmethod immediate oops are patched first, and the guard value (immediate of the cmp instruction) is patched after, then if a thread sees the new cmp instruction, it will also see the new oop immediates.   And that is indeed what the "asynchronous" cross modifying code description ensures will work in the AMD APM. So that all checks out.
> 
> I guess this is a separate issue from this patch, but where does the AMD APM guarantee that?

In the APM, volume 2 (cf. https://www.amd.com/content/dam/amd/en/documents/processor-tech-docs/programmer-references/24593.pdf), section 7.6.1 under "Asynchronous modification", it says ""

> > The assumption is that if the nmethod immediate oops are patched first, and the guard value (immediate of the cmp instruction) is patched after, then if a thread sees the new cmp instruction, it will also see the new oop immediates.   And that is indeed what the "asynchronous" cross modifying code description ensures will work in the AMD APM. So that all checks out.
> 
> I guess this is a separate issue from this patch, but where does the AMD APM guarantee that?

Hmm, it used to be in Volume 2, section 7.6.1. But in the latest revision, 3.41 from this summer, I can't find it any more. Strange.

theRealAph:
> > 
> > I guess this is a separate issue from this patch, but where does the AMD APM guarantee that?
> 
> Hmm, it used to be in Volume 2, section 7.6.1. But in the latest revision, 3.41 from this summer, I can't find it any more. Strange.

I wonder if they may be making it up as they go along.

dean-long:
This looks much better.  Is the cross modify fence always needed, or can it be conditional based on nmethod_patching_type()?

fisk:
> This looks much better.  Is the cross modify fence always needed, or can it be conditional based on nmethod_patching_type()?

It could be conditional, but the patching type abstraction is currently platform specific, and here we are in shared code. I'm not sure if elevating that to the shared level is worth it, given that I found it not to impact performance. What do you think?

dean-long:
If there is no impact on performance, especially on aarch64, then I'm OK with it as is, or maybe with a comment in the code saying it could be made conditional if needed.

fisk:
> If there is no impact on performance, especially on aarch64, then I'm OK with it as is, or maybe with a comment in the code saying it could be made conditional if needed.

Thank you. I added a comment describing that it can be made conditional if needed in the future.

dean-long (approved):

xmas92 (approved):

src/hotspot/share/gc/shared/barrierSetNMethod.cpp
```diff
@@ -186,6 +186,14 @@ int BarrierSetNMethod::nmethod_stub_entry_barrier(address* return_address_ptr) {
   // Called upon first entry after being armed
   bool may_enter = bs_nm->nmethod_entry_barrier(nm);
 
+  // In case a concurrent thread disarmed the nmethod, we need to ensure the new instructions
+  // are made visible using a cross modify fence. Note that this is synchronous cross modifying
```
theRealAph, line 190:
```suggestion
  // are made visible, by using a cross modify fence. Note that this is synchronous cross modifying
```

theRealAph (approved):

fisk:
Thanks for the reviews, @theRealAph, @dean-long and @xmas92!

PR #14545, opened 2023-06-19
8308984: Relativize last_sp (and top_frame_sp) in interpreter frames

fbredber:
Implementation of relativized last_sp (top_frame_sp on PowerPC) in interpreter frames for x64, aarch64, ppc64le and riscv.
Not relativized last_sp on arm, zero and s390 but done some changes to cope with the changed generic code.

By changing the "last_sp" member in interpreter frames from being an absolute address into an offset that is relative to the frame pointer, we don't need to change the value as we freeze and thaw frames of virtual threads. This is since we might freeze and thaw from and to different worker threads, so the absolute address to locals might change, but the offset from the frame pointer will be constant.

This subtask only handles "last_sp" (and its close equivalent "top_frame_sp" on PowerPC). The relativization of other interpreter frame members are handled in other subtasks to JDK-8289296.

Tested tier1-tier7 on supported platforms. The rest was sanity tested using Qemu.

---------
### Progress

### Issue
 * [JDK-8308984](https://bugs.openjdk.org/browse/JDK-8308984): Relativize last_sp (and top_frame_sp) in interpreter frames (**Sub-task** - P4)

### Reviewers
 * [Patricio Chilano Mateo](https://openjdk.org/census#pchilanomate) (@pchilano - **Reviewer**)
 * [Hao Sun](https://openjdk.org/census#haosun) (@shqking - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14545/head:pull/14545` \
`$ git checkout pull/14545`

`$ git checkout pull/14545` \
`$ git pull https://git.openjdk.org/jdk.git pull/14545/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14545`

`$ git pr show -t 14545`

</details>

<a href="https://git.openjdk.org/jdk/pull/14545.diff">https://git.openjdk.org/jdk/pull/14545.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14545#issuecomment-1655610258)

### DIFF
--- a/src/hotspot/cpu/aarch64/templateInterpreterGenerator_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/templateInterpreterGenerator_aarch64.cpp
@@ -466,7 +466,7 @@ address TemplateInterpreterGenerator::generate_return_entry_for(TosState state, 
 
   // Restore stack bottom in case i2c adjusted stack
   __ ldr(rscratch1, Address(rfp, frame::interpreter_frame_last_sp_offset * wordSize));
-  __ lea(esp, Address(rfp, rscratch1, Address::lsl(3)));
+  __ lea(esp, Address(rfp, rscratch1, Address::lsl(Interpreter::logStackElementSize)));
   // and null it as marker that esp is now tos until next java call
   __ str(zr, Address(rfp, frame::interpreter_frame_last_sp_offset * wordSize));
   __ restore_bcp();
@@ -523,7 +523,7 @@ address TemplateInterpreterGenerator::generate_deopt_entry_for(TosState state,
 
   // Restore expression stack pointer
   __ ldr(rscratch1, Address(rfp, frame::interpreter_frame_last_sp_offset * wordSize));
-  __ lea(esp, Address(rfp, rscratch1, Address::lsl(3)));
+  __ lea(esp, Address(rfp, rscratch1, Address::lsl(Interpreter::logStackElementSize)));
   // null last_sp until next java call
   __ str(zr, Address(rfp, frame::interpreter_frame_last_sp_offset * wordSize));
 
@@ -1870,7 +1870,7 @@ void TemplateInterpreterGenerator::generate_throw_exception() {
 
   // Restore the last_sp and null it out
   __ ldr(rscratch1, Address(rfp, frame::interpreter_frame_last_sp_offset * wordSize));
-  __ lea(esp, Address(rfp, rscratch1, Address::lsl(3)));
+  __ lea(esp, Address(rfp, rscratch1, Address::lsl(Interpreter::logStackElementSize)));
   __ str(zr, Address(rfp, frame::interpreter_frame_last_sp_offset * wordSize));
 
   __ restore_bcp();
--- a/src/hotspot/cpu/aarch64/templateInterpreterGenerator_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/templateInterpreterGenerator_aarch64.cpp
@@ -465,7 +465,8 @@ address TemplateInterpreterGenerator::generate_return_entry_for(TosState state, 
   address entry = __ pc();
 
   // Restore stack bottom in case i2c adjusted stack
-  __ ldr(esp, Address(rfp, frame::interpreter_frame_last_sp_offset * wordSize));
+  __ ldr(rscratch1, Address(rfp, frame::interpreter_frame_last_sp_offset * wordSize));
+  __ lea(esp, Address(rfp, rscratch1, Address::lsl(3)));
   // and null it as marker that esp is now tos until next java call
   __ str(zr, Address(rfp, frame::interpreter_frame_last_sp_offset * wordSize));
   __ restore_bcp();
@@ -521,7 +522,8 @@ address TemplateInterpreterGenerator::generate_deopt_entry_for(TosState state,
   __ restore_sp_after_call();  // Restore SP to extended SP
 
   // Restore expression stack pointer
-  __ ldr(esp, Address(rfp, frame::interpreter_frame_last_sp_offset * wordSize));
+  __ ldr(rscratch1, Address(rfp, frame::interpreter_frame_last_sp_offset * wordSize));
+  __ lea(esp, Address(rfp, rscratch1, Address::lsl(3)));
   // null last_sp until next java call
   __ str(zr, Address(rfp, frame::interpreter_frame_last_sp_offset * wordSize));
 
@@ -1867,7 +1869,8 @@ void TemplateInterpreterGenerator::generate_throw_exception() {
                        /* notify_jvmdi */ false);
 
   // Restore the last_sp and null it out
-  __ ldr(esp, Address(rfp, frame::interpreter_frame_last_sp_offset * wordSize));
+  __ ldr(rscratch1, Address(rfp, frame::interpreter_frame_last_sp_offset * wordSize));
+  __ lea(esp, Address(rfp, rscratch1, Address::lsl(3)));
   __ str(zr, Address(rfp, frame::interpreter_frame_last_sp_offset * wordSize));
 
   __ restore_bcp();
--- a/src/hotspot/share/runtime/continuationFreezeThaw.cpp
+++ b/src/hotspot/share/runtime/continuationFreezeThaw.cpp
@@ -432,7 +432,6 @@ private:
   inline void set_top_frame_metadata_pd(const frame& hf);
   inline void patch_pd(frame& callee, const frame& caller);
   void adjust_interpreted_frame_unextended_sp(frame& f);
-  static inline void prepare_freeze_interpreted_top_frame(const frame& f);
   static inline void relativize_interpreted_frame_metadata(const frame& f, const frame& hf);
 
 protected:
@@ -809,13 +808,11 @@ NOINLINE freeze_result FreezeBase::recurse_freeze(frame& f, frame& caller, int c
     return recurse_freeze_compiled_frame(f, caller, callee_argsize, callee_interpreted);
   } else if (f.is_interpreted_frame()) {
     assert((_preempt && top) || !f.interpreter_frame_method()->is_native(), "");
-    if (top) {
-      if (_preempt && f.interpreter_frame_method()->is_native()) {
-        // int native entry
-        return freeze_pinned_native;
-      }
-      prepare_freeze_interpreted_top_frame(f);
+    if (_preempt && top && f.interpreter_frame_method()->is_native()) {
+      // int native entry
+      return freeze_pinned_native;
     }
+
     return recurse_freeze_interpreted_frame(f, caller, callee_argsize, callee_interpreted);
   } else if (_preempt && top && ContinuationHelper::Frame::is_stub(f.cb())) {
     return recurse_freeze_stub_frame(f, caller);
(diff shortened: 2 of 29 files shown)


fbredber:
I've done basic loom testing on ppc64le and riscv64 using Qemu, but would appreciate if @reinrich and @RealFYang could take it for a real test drive.

reinrich:
> I've done basic loom testing on ppc64le and riscv64 using Qemu, but would appreciate if @reinrich and @RealFYang could take it for a real test drive.

PPC code looks good. I've tested `jtreg:test/jdk/jdk/internal/vm/Continuation` successfully. Will do more testing until monday.

RealFYang (changes requested):
Hi, This has passed hotspot_loom, jdk_loom and tier1-3 tests on linux-riscv64 platform. Minor suggestions for riscv code. Thanks.

reinrich:
> > I've done basic loom testing on ppc64le and riscv64 using Qemu, but would appreciate if @reinrich and @RealFYang could take it for a real test drive.
> 
> PPC code looks good. I've tested `jtreg:test/jdk/jdk/internal/vm/Continuation` successfully. Will do more testing until monday.

So there are crashes in our tests with this pr applied.
E.g. `make test TEST=runtime/CommandLine/OptionsValidation/TestOptionsWithRangesDynamic.java` crashes always. To me it seems that the issue is unrelated to virtual threads. It has to be caused by relativization of `top_frame_sp`
Unfortunately I cannot look into it since I'm actually on vacation. @TheRealMDoerr might want to look into it. He's coming back from vacation in a couple of days.

src/hotspot/cpu/aarch64/templateInterpreterGenerator_aarch64.cpp
```diff
@@ -465,7 +465,8 @@ address TemplateInterpreterGenerator::generate_return_entry_for(TosState state,
   address entry = __ pc();
 
   // Restore stack bottom in case i2c adjusted stack
-  __ ldr(esp, Address(rfp, frame::interpreter_frame_last_sp_offset * wordSize));
+  __ ldr(rscratch1, Address(rfp, frame::interpreter_frame_last_sp_offset * wordSize));
+  __ lea(esp, Address(rfp, rscratch1, Address::lsl(3)));
```
theRealAph, line 469:
```suggestion
  __ lea(esp, Address(rfp, rscratch1, Address::lsl(Interpreter::logStackElementSize)));
```
I'm not sure that `logStackElementSize` buys us anything, but we might as well be consistent in this patch.

theRealAph:
I think you need to correct `frame::describe` to make it print the relativized SP in an easy-to-understand form.

src/hotspot/share/runtime/continuationFreezeThaw.cpp
```diff
@@ -808,11 +809,13 @@ NOINLINE freeze_result FreezeBase::recurse_freeze(frame& f, frame& caller, int c
...
     assert((_preempt && top) || !f.interpreter_frame_method()->is_native(), "");
-    if (_preempt && top && f.interpreter_frame_method()->is_native()) {
-      // int native entry
-      return freeze_pinned_native;
+    if (top) {
+      if (_preempt && f.interpreter_frame_method()->is_native()) {
+        // int native entry
+        return freeze_pinned_native;
+      }
+      prepare_freeze_interpreted_top_frame(f);
```
pchilano, line 817: Probably better to handle setting last_sp for this special case in freeze_start_frame_safepoint_stub(). But in any case the preemption code is not there yet so I would just remove these changes and prepare_freeze_interpreted_top_frame().

fbredber:
Thank you @reinrich for pointing out the failing `runtime/CommandLine/OptionsValidation/TestOptionsWithRangesDynamic.java` test. I used `R11_scratch1` inside `InterpreterMacroAssembler::add_monitor_to_stack()` which apparently is not so good because it's the same register as is used for the local `esp` register. Since `R12_scratch2` is also used I looked around the code and figured I should be able to  use `R23_tmp3` as a scratch register. It seems to work. Now the test passes.

Do you or @TheRealMDoerr have any objections to using `R23_tmp3` as a scratch register in `InterpreterMacroAssembler::add_monitor_to_stack()` ?

TheRealMDoerr:
I'd use `Rtemp2`. Note that one of the computations can be simplified:

```diff
diff --git a/src/hotspot/cpu/ppc/interp_masm_ppc_64.cpp b/src/hotspot/cpu/ppc/interp_masm_ppc_64.cpp
index 00722171058..721a1467ed8 100644
--- a/src/hotspot/cpu/ppc/interp_masm_ppc_64.cpp
+++ b/src/hotspot/cpu/ppc/interp_masm_ppc_64.cpp
@@ -1994,10 +1994,10 @@ void InterpreterMacroAssembler::add_monitor_to_stack(bool stack_is_empty, Regist
          "size of a monitor must respect alignment of SP");
 
   resize_frame(-monitor_size, /*temp*/esp); // Allocate space for new monitor
-  sub(R11_scratch1, R1_SP, esp); // esp contains fp
-  sradi(R11_scratch1, R11_scratch1, Interpreter::logStackElementSize);
+  subf(Rtemp2, esp, R1_SP); // esp contains fp
+  sradi(Rtemp2, Rtemp2, Interpreter::logStackElementSize);
   // Store relativized top_frame_sp
-  std(R11_scratch1, _ijava_state_neg(top_frame_sp), esp); // esp contains fp
+  std(Rtemp2, _ijava_state_neg(top_frame_sp), esp); // esp contains fp
 
   // Shuffle expression stack down. Recall that stack_base points
   // just above the new expression stack bottom. Old_tos and new_tos
diff --git a/src/hotspot/cpu/ppc/templateInterpreterGenerator_ppc.cpp b/src/hotspot/cpu/ppc/templateInterpreterGenerator_ppc.cpp
index 7df95d5672e..e3a8b3fbd94 100644
--- a/src/hotspot/cpu/ppc/templateInterpreterGenerator_ppc.cpp
+++ b/src/hotspot/cpu/ppc/templateInterpreterGenerator_ppc.cpp
@@ -1061,9 +1061,8 @@ void TemplateInterpreterGenerator::generate_fixed_frame(bool native_call, Regist
   __ std(R0, _ijava_state_neg(oop_tmp), R1_SP); // only used for native_call
 
   // Store sender's SP and this frame's top SP.
-  __ subf(R12_scratch2, Rtop_frame_size, R1_SP);
   __ std(R21_sender_SP, _ijava_state_neg(sender_sp), R1_SP);
-  __ sub(R12_scratch2, R12_scratch2, R1_SP);
+  __ neg(R12_scratch2, Rtop_frame_size);
   __ sradi(R12_scratch2, R12_scratch2, Interpreter::logStackElementSize);
   // Store relativized top_frame_sp
   __ std(R12_scratch2, _ijava_state_neg(top_frame_sp), R1_SP);
```

fbredber:
Yea Rtemp2 seems to work fine @TheRealMDoerr. Since Rtemp1 is used as an in-parameter (esp) and therefore not usable for me, I thought that Rtemp2 was also an in-parameter and not usable for me (the declaration of slot is very similar to esp). But after your comment I realized that Rtemp2 is not an in-parameter, so it's usable for me as a temp register. Also thankyou for drawing my attention to the fact that one of the computations could be simplified.

TheRealMDoerr:
Thanks for taking care of PPC64! Test results look good so far.

src/hotspot/share/runtime/frame.cpp
```diff
@@ -1621,8 +1621,7 @@ void FrameValues::print_on(outputStream* st, int min_index, int max_index, intpt
       const char* spacer = "          " LP64_ONLY("        ");
       st->print_cr(" %s  %s %s", spacer, spacer, fv.description);
     } else {
-      if (on_heap
```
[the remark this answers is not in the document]
fbredber: Thank you for pointing that out @shqking. Yes I think we should remove the parameter. I'll fix it.

theRealAph:
> I think you need to correct `frame::describe` to make it print the relativized SP in an easy-to-understand form.

Ping.

fbredber:
> > I think you need to correct `frame::describe` to make it print the relativized SP in an easy-to-understand form.
> 
> Ping.

I looked into frame::describe and I realized that it already contained code to deal with relativized values (by @AlanBateman and adopted to PowerPC by @reinrich).  It works like this: If a member of the fixed frame is on the heap and has a value between -99 to 99 it's value will be printed as a signed integer instead of a hexadecimal value. 

I think the basic thought is something like this: Members of the fixed frame are pointers and therefore don't hold values [-99..99], but if they do, they are probably relativized values so they should be printed as integers. And they should be interpreted as pointing to a stack index relative to the frame pointer. 

Up until I started to change fixed frame pointers so that they are permanently relativized (JDK-8289296), the relativized pointers only occurred on the heap (after a freeze). Henze the on_heap condition.
```
      if (on_heap
          && *fv.location != 0 && *fv.location > -100 && *fv.location < 100

```
I decided to go with the prior art of Alan and Richard, but since the relativized values can now appear both on the stack and on the heap I removed the on_heap condition. I've tested the output on several platforms, and the output seems ok to me.

I'm not sure how to interpret your "Ping." comment. 
I know the code change in frame::describe is very small, but it extends the way relativized pointers (on the heap) was shown before so that relativized pointers on the stack are shown in the same way. 

Does this answer your "Ping" @theRealAph?

reinrich:
Wouldn't it be good to print derelativized pointers for frames on stack? Maybe print both?

theRealAph:
> I looked into frame::describe and I realized that it already contained code to deal with relativized values (by @AlanBateman and adopted to PowerPC by @reinrich). It works like this: If a member of the fixed frame is on the heap and has a value between -99 to 99 it's value will be printed as a signed integer instead of a hexadecimal value.

Sure, but we need to see what they point to. In other words, every relativized point should be _decoded_ into its pointer form. Otherwise relativization, a good thing, becomes obfuscation, a bad thing.

fbredber:
> Sure, but we need to see what they point to. In other words, every relativized point should be _decoded_ into its pointer form. Otherwise relativization, a good thing, becomes obfuscation, a bad thing.

I see what you mean, but It can also be very confusing if you use a debugger to hex dump the memory were the fixed frame is located, and you see something else than is shown by frame::describe.

My opinion is that it's important to show the relativzed pointers the same way, regardless if they are on the stack or on the heap. And also somehow show that they are relativized.

Also since this PR is really just about relativizing last_sp, I think it would be nice if it could be integrated with the small change I did to frame::describe. After that I can create a new issue to handle the broader picture of how to show relativized pointers in frame::describe.

theRealAph:
> Also since this PR is really just about relativizing last_sp, I think it would be nice if it could be integrated with the small change I did to frame::describe. After that I can create a new issue to handle the broader picture of how to show relativized pointers in frame::describe.

That sounds good to me. Thanks.

fbredber:
Created [Improve how relativized pointers are printed by frame::describe](https://bugs.openjdk.org/browse/JDK-8314508) to handle the broader picture of how to show relativized pointers in frame::describe.

pchilano (approved):
Thanks for the fixes @fbredber, looks good to me.

Patricio

theRealAph (approved):

shqking (approved):
Thanks for your update on "on_heap" parameter.
AArch64 part looks good to me.

PR #14547, opened 2023-06-19
8310335 JFR: Modernize collections and switch statements

egahlin:
Could I have a review of an update to the jfr modules to use more recent APIs and language constructs.

Testing: test/jdk/jdk/jfr

Thanks
Erik

---------
### Progress

### Issue
 * [JDK-8310335](https://bugs.openjdk.org/browse/JDK-8310335): JFR: Modernize collections and switch statements (**Enhancement** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14547/head:pull/14547` \
`$ git checkout pull/14547`

`$ git checkout pull/14547` \
`$ git pull https://git.openjdk.org/jdk.git pull/14547/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14547`

`$ git pr show -t 14547`

</details>

<a href="https://git.openjdk.org/jdk/pull/14547.diff">https://git.openjdk.org/jdk/pull/14547.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14547#issuecomment-1597756256)

### DIFF
--- a/src/jdk.jfr/share/classes/jdk/jfr/consumer/RecordedObject.java
+++ b/src/jdk.jfr/share/classes/jdk/jfr/consumer/RecordedObject.java
@@ -464,7 +464,6 @@ public sealed class RecordedObject
         if (o instanceof Character c) {
             return c;
         }
-
         throw newIllegalArgumentException(name, "char");
     }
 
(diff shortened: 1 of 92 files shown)


src/jdk.jfr/share/classes/jdk/jfr/consumer/RecordedObject.java
```diff
@@ -502,12 +502,11 @@ public final short getShort(String name) {
         if (o instanceof Byte b) {
             return b;
         }
-        if (o instanceof UnsignedValue unsigned) {
```
SirYwell, line 505: I think this can be further simplified, e.g.,
```java
    return switch(o) {
        case Short s -> s;
        case Byte b -> b;
        case UnsignedValue(Short s) -> s;
        case UnsignedValue(Byte b) -> (short) Byte.toUnsignedInt(b);
        case null, default -> throw newIllegalArgumentException(name, "short");
    };
```
What do you think?

egahlin: I thought about having everything in a switch statements, not just for this method, but I wasn't sure about performance. 99% of the calls will probably be trapped in the first instanceof check. Users are likely to use the method matching the datatype. A fast path might pay off.

I did some measurements and I couldn't spot a difference, so I will change to one  switch statement. Thanks.

mgronlun (approved):

PR #14552, opened 2023-06-20
Make MapN ValueBased

minborg:
This is a PR with a sketch of enabling `@ValueBased` `Map` implementations. It also solves some race conditions that existed in previous implementations.

---------
### Progress
- [ ] Commit message must refer to an issue

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14552/head:pull/14552` \
`$ git checkout pull/14552`

`$ git checkout pull/14552` \
`$ git pull https://git.openjdk.org/jdk.git pull/14552/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14552`

`$ git pr show -t 14552`

</details>

<a href="https://git.openjdk.org/jdk/pull/14552.diff">https://git.openjdk.org/jdk/pull/14552.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/util/AbstractMap.java
+++ b/src/java.base/share/classes/java/util/AbstractMap.java
@@ -25,6 +25,10 @@
 
 package java.util;
 
+import jdk.internal.misc.Unsafe;
+import jdk.internal.vm.annotation.Stable;
+
+import java.util.concurrent.locks.StampedLock;
 import java.util.stream.Stream;
 import java.util.function.Consumer;
 import java.util.function.IntFunction;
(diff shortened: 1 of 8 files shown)


src/java.base/share/classes/java/util/AbstractMap.java
```diff
@@ -304,32 +311,24 @@ public void clear() {
...
-     *}</pre>
-     */
-    transient Set<K>        keySet;
-    transient Collection<V> values;
+    private static final long VIEWS
+            = UNSAFE.objectFieldOffset(AbstractMap.class, "views");
+
+    // Todo: Investigate if we can make this field final
+    // Set via Unsafe
+    private transient Views<K , V> views;
```
ExE-Boss, line 319: Note that this field still disqualifies `ImmutableCollections.Map*` from being `@ValueBased`.

liach:
Not too sure about how Valhalla will roll out, but for the 2 views, I think they should be made value-based as well; do the views hold any field other than the enclosing map instance? If they are single-object wrappers, I think we might as well just drop the caching, which just prevents object creation (which will be pointless once valhalla arrives)

minborg:
I am closing this draft PR as it would be better to remove the fields all together.

PR #14555, opened 2023-06-20
8140527: JInternalFrame has incorrect title button width

prsadhuk:
Title buttons under Widows Classic L&F got their sizes from the XP desktop theme in which button width can be bigger than height. It is construed as XP bug where sizes aren't updated properly so it uses height units for width for XP and later windows. The proposed fix uses the [same technique](https://github.com/openjdk/jdk/blob/a0595761ef35c4eec8cb84326a869b9473cd5bba/src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/WindowsInternalFrameTitlePane.java#L78-L82) for Classic and forces title buttons to be square and to fit the frame title in height.

Before fix SwingSet2 demo (Windows Classic InternalFrame)
image

After fix
image

---------
### Progress

### Issues
 * [JDK-8140527](https://bugs.openjdk.org/browse/JDK-8140527): JInternalFrame has incorrect title button width (**Bug** - P3)
 * [JDK-8139392](https://bugs.openjdk.org/browse/JDK-8139392): JInternalFrame has incorrect padding (**Bug** - P4)

### Reviewers
 * [Tejesh R](https://openjdk.org/census#tr) (@TejeshR13 - Committer)

### Contributors

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14555/head:pull/14555` \
`$ git checkout pull/14555`

`$ git checkout pull/14555` \
`$ git pull https://git.openjdk.org/jdk.git pull/14555/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14555`

`$ git pr show -t 14555`

</details>

<a href="https://git.openjdk.org/jdk/pull/14555.diff">https://git.openjdk.org/jdk/pull/14555.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14555#issuecomment-1598515219)


test/jdk/javax/swing/JInternalFrame/InternalFrameTitleButtonTest.java
```diff
@@ -0,0 +1,112 @@
...
+                    ((BasicInternalFrameTitlePane)((BasicInternalFrameUI)
+                            iframe.getUI()).getNorthPane());
+            int height = title.getHeight();
+            Insets insets = title.getInsets();
+            height = height - insets.top - insets.bottom;
+            for (int i = 0; i < title.getComponentCount(); i++) {
+                Component c = title.getComponent(i);
+                if (c instanceof JButton) {
+                    Icon icon = ((JButton) c).getIcon();
+                    if( icon.getIconHeight() > height - 4 ||
```
turbanoff, line 102:
```suggestion
                    if (icon.getIconHeight() > height - 4 ||
```

src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/WindowsInternalFrameTitlePane.java
```diff
@@ -85,6 +85,7 @@ protected void installDefaults() {
                 buttonWidth = (int) ((float) buttonWidth * d.width / d.height);
             }
         } else {
+            buttonWidth = buttonHeight;
```
aivanov-jdk, line 88: This means line 71 can be removed because `buttonWidth` is never read from `UIManager`.

aivanov-jdk (commented):
Does Windows classic theme has its own set UIManager settings?

src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/WindowsInternalFrameTitlePane.java
prsadhuk: It can be removed in the strict sense but we need something to initialise it to so kept it as it is, as it will do no harm

src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/WindowsInternalFrameTitlePane.java
```diff
@@ -85,6 +85,7 @@ protected void installDefaults() {
...
             buttonWidth += 2;
```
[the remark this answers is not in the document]
prsadhuk: If we make it exactly square, it results in some loss of aesthetics as can be seen here where icon is too close to right edge, so kept it as it is

image

prsadhuk:
> I wonder if the values stored in the UI Manager should have this trick applied rather tweaking the values on the fly. If a programmer gets the values from `UIManager` directly, they will get the incorrect value.

OK. UIManager is also updated with updated width so values will be identical

src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/WindowsIconFactory.java
```diff
@@ -294,8 +294,9 @@ public int getIconWidth() {
                     width = (int) ((float) width * d.width / d.height);
                 }
             } else {
-                width = UIManager.getInt("InternalFrame.titleButtonWidth") -2;
+                width = UIManager.getInt("InternalFrame.titleButtonHeight") - 2;
```
aivanov-jdk, line 297:
```suggestion
                width = UIManager.getInt("InternalFrame.titleButtonHeight") -2;
```
I agree it looks better with the space on either side but it's consistently used without the space after `-`. So I'd rather keep the original formatting.

src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/WindowsInternalFrameTitlePane.java
aivanov-jdk: > It can be removed in the strict sense but we need something to initialise it to so kept it as it is, as it will do no harm

Why does it need anything to initialise? The value is now set in either `if` or `else` branch.

It does not harm, yet it's never used. Not initialising it with `"InternalFrame.titleButtonWidth"` conveys it clearer: the value is computed later.

aivanov-jdk (commented):
We have exactly the same code in both `WindowsIconFactory.java` and `WindowsInternalFrameTitlePane.java`. Can it moved into a helper method to avoid duplication?

Would be enough to have it in `WindowsIconFactory.java` only?

Can `WindowsInternalFrameTitlePane.java` rely on the value set to UI Manager?

aivanov-jdk:
> > I wonder if the values stored in the UI Manager should have this trick applied rather tweaking the values on the fly. If a programmer gets the values from `UIManager` directly, they will get the incorrect value.
> 
> OK. UIManager is also updated with updated width so values will be identical

Can it be modified so that the value for `"InternalFrame.titleButtonWidth"` gets computed based on `"InternalFrame.titleButtonHeight"` using the algorithm in `WindowsIconFactory.java` and `WindowsInternalFrameTitlePane.java` when its value is requested?

This way UIManager automatically returns the value we want; the adjustment logic can be dropped from both classes above.

src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/WindowsInternalFrameTitlePane.java
prsadhuk: ok. I thought buttonWidth was declared in the method but it is class variable

prsadhuk:
> > > I wonder if the values stored in the UI Manager should have this trick applied rather tweaking the values on the fly. If a programmer gets the values from `UIManager` directly, they will get the incorrect value.
> > 
> > 
> > OK. UIManager is also updated with updated width so values will be identical
> 
> Can it be modified so that the value for `"InternalFrame.titleButtonWidth"` gets computed based on `"InternalFrame.titleButtonHeight"` using the algorithm in `WindowsIconFactory.java` and `WindowsInternalFrameTitlePane.java` when its value is requested?
> 
> This way UIManager automatically returns the value we want; the adjustment logic can be dropped from both classes above.

We probably can change in [WindowsLookAndFeel](https://github.com/openjdk/jdk/blob/5a82fa3bb278b5b80b9ede0619f364fe13cdbede/src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/WindowsLookAndFeel.java#L813-L814) 
but for XPtheme `width = height-4` and for ClassicTheme `width = height-2` so we need `getXP ` check so I will prefer to keep it as it is as otherwise it might create regression and also I could not find anyway to access WIndowsInternalFrameTitlePane methods in WindowsIconFactory to avoid duplication..

prsadhuk:
> > > > I wonder if the values stored in the UI Manager should have this trick applied rather tweaking the values on the fly. If a programmer gets the values from `UIManager` directly, they will get the incorrect value.
> > > 
> > > 
> > > OK. UIManager is also updated with updated width so values will be identical
> > 
> > 
> > Can it be modified so that the value for `"InternalFrame.titleButtonWidth"` gets computed based on `"InternalFrame.titleButtonHeight"` using the algorithm in `WindowsIconFactory.java` and `WindowsInternalFrameTitlePane.java` when its value is requested?
> > This way UIManager automatically returns the value we want; the adjustment logic can be dropped from both classes above.
> 
> We probably can change in [WindowsLookAndFeel](https://github.com/openjdk/jdk/blob/5a82fa3bb278b5b80b9ede0619f364fe13cdbede/src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/WindowsLookAndFeel.java#L813-L814) but for XPtheme `width = height-4` and for ClassicTheme `width = height-2` so we need `getXP ` check so I will prefer to keep it as it is as otherwise it might create regression and also I could not find anyway to access WIndowsInternalFrameTitlePane methods in WindowsIconFactory to avoid duplication..

Having said that, I guess XP theme also padding is not correct as can be seen in this bug [JDK-8139392](https://bugs.openjdk.org/browse/JDK-8139392) which I have rectified

After fix
image

src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/WindowsInternalFrameTitlePane.java
aivanov-jdk: Nothing would've changed if it had been a local variable. Both branches initialise the local variable, the compiler is happy.

test/jdk/javax/swing/JInternalFrame/InternalFrameTitleButtonTest.java
```diff
@@ -0,0 +1,116 @@
...
+import javax.swing.SwingUtilities;
+import javax.swing.UIManager;
+
+public class InternalFrameTitleButtonTest {
+
+    private static JFrame frame;
+    private static JInternalFrame iframe;
+
+    public static void main(String[] args) throws Exception {
+        if (OSInfo.getOSType() == OSInfo.OSType.WINDOWS) {
```
aivanov-jdk, line 55: If you remove this line (OS check), the test would be easier to run as a stand-alone app. It's up to you though.

test/jdk/javax/swing/JInternalFrame/InternalFrameTitleButtonTest.java
```diff
@@ -0,0 +1,116 @@
...
+            int height = title.getHeight();
+            Insets insets = title.getInsets();
+            height = height - insets.top - insets.bottom;
+            for (int i = 0; i < title.getComponentCount(); i++) {
+                Component c = title.getComponent(i);
+                if (c instanceof JButton) {
+                    Icon icon = ((JButton) c).getIcon();
+                    if (icon.getIconHeight() > height - 4 ||
+                        icon.getIconWidth() > height - 2) {
+                        throw new RuntimeException("Wrong title icon size");
```
aivanov-jdk, line 104:
```suggestion
                    if (icon.getIconHeight() > height - 4
                        || icon.getIconWidth() > height - 2) {
                        throw new RuntimeException("Wrong title icon size");
```
Wrapping the operator prevents ambiguity: continuation line or statement.

src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/WindowsIconFactory.java
```diff
@@ -286,19 +286,12 @@ public int getIconWidth() {
...
-                    width = (int) ((float) width * d.width / d.height);
-                }
+                // Assume for now that height is correct and derive width from height
+                width = UIManager.getInt("InternalFrame.titleButtonHeight") + 2;
             } else {
-                width = UIManager.getInt("InternalFrame.titleButtonWidth") -2;
-            }
-            if (XPStyle.getXP() != null) {
-                width -= 2;
+                width = UIManager.getInt("InternalFrame.titleButtonHeight") -2;
```
aivanov-jdk, line 292: Perhaps, add a space between `-2` (to be consistent with `+ 2`). You've updated the entire method `getIconWidth`, so it will look better.

src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/WindowsInternalFrameTitlePane.java
```diff
@@ -77,19 +76,16 @@ protected void installDefaults() {
...
             // Fix for XP bug where sometimes these sizes aren't updated properly
-            // Assume for now that height is correct and derive width using the
-            // ratio from the uxtheme part
-            buttonWidth = buttonHeight;
-            Dimension d = XPStyle.getPartSize(Part.WP_CLOSEBUTTON, State.NORMAL);
-            if (d != null && d.width != 0 && d.height != 0) {
-                buttonWidth = (int) ((float) buttonWidth * d.width / d.height);
-            }
+            // Assume for now that height is correct and derive width from height
+            buttonWidth = buttonHeight+6;
```
aivanov-jdk, line 80: Perhaps?
```suggestion
            buttonWidth = buttonHeight + 6;
```
For `height` above, it's `- 4`; no other instances of `-` or `+` inside the method, so spaces around `+` will be consistent.

aivanov-jdk (approved):
Looks good to me.

The only concern is that UI Manager has a wrong width until `JInternalFrame` is shown or initialised. But it has been the case before the fix too.

aivanov-jdk:
> Having said that, I guess XP theme also padding is not correct as can be seen in this bug [JDK-8139392](https://bugs.openjdk.org/browse/JDK-8139392) which I have rectified
> 
> After fix image

Your latest changes also fix [JDK-8139392](https://bugs.openjdk.org/browse/JDK-8139392), do I get it right? If so, add it to the list of fixed issues.

test/jdk/javax/swing/JInternalFrame/InternalFrameTitleButtonTest.java
```diff
@@ -52,18 +49,21 @@ public class InternalFrameTitleButtonTest {
...
-                });
-            }
+        String osName = System.getProperty("os.name");
+        if(!osName.toLowerCase().contains("win")) {
+            System.out.println("The test is applicable only for Windows.");
+            return;
+        }
+        UIManager.setLookAndFeel(
+               "com.sun.java.swing.plaf.windows.WindowsClassicLookAndFeel");
+        try {
```
aivanov-jdk, line 59: Actually, I meant that you can skip checking for Windows at all: the test has `@requires (os.family == "windows")` tag. Then setting *Windows Classic Look and Feel* will fail on other platforms but Windows, that is `UIManager.setLookAndFeel` will throw an exception which will eventually fail the test.

Whatever way you prefer.

aivanov-jdk (approved):
Would you mind updating the copyright years?

test/jdk/javax/swing/JInternalFrame/InternalFrameTitleButtonTest.java
```diff
@@ -0,0 +1,116 @@
...
+import javax.swing.UIManager;
+
+public class InternalFrameTitleButtonTest {
+
+    private static JFrame frame;
+    private static JInternalFrame iframe;
+
+    public static void main(String[] args) throws Exception {
+        String osName = System.getProperty("os.name");
+        if(!osName.toLowerCase().contains("win")) {
```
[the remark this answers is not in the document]
honkar-jdk: @prsadhuk The string compare can be replaced with `OSInfo.getOSType() == OSInfo.OSType.WINDOWS`  and you want to retain the OSCheck along with `@requires (os.family == "windows")` in case the it is run as standalone test.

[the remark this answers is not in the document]
aivanov-jdk: I think [it's redundant](https://github.com/openjdk/jdk/pull/14555#discussion_r1242206200) in this case because setting Windows Classic L&F will throw an exception on other platforms.

It was [`OSInfo`-check initially](https://github.com/openjdk/jdk/pull/14555#discussion_r1240155464).

I'm fine with either way. However, I prefer *no check* in this particular case.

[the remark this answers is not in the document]
prsadhuk: Actually I will prefer gentle nudge stating it's not applicable other than windows as compared to harsh failing via throwing exception, if you dont mind..

aivanov-jdk:
> > Having said that, I guess XP theme also padding is not correct as can be seen in this bug [JDK-8139392](https://bugs.openjdk.org/browse/JDK-8139392) which I have rectified
> > 
> > After fix
> > image
> 
> Your latest changes also fix [JDK-8139392](https://bugs.openjdk.org/browse/JDK-8139392), do I get it right? If so, add it to the list of fixed issues.

This change does not address JDK-8139392 fully. The title bar buttons in Vista theme, which are still used for MDI windows, are not square. The dimensions of the native buttons are 32×18. The dimensions of the JInternalFrame title bar buttons after your fix are 24×18; before your fix they were 28×18. That is the width of the buttons is smaller by 8 and 4 pixels. Aren't these the values subtracted from the width value fetched from Windows?

As for the classic theme… It's not available in Windows anymore, there's only Visual-style enabled theme now. So, you can't get the dimensions of the title bar buttons of an MDI window. I guess the only way is to use heuristics. According to old screenshots I have, the main title bar buttons in Windows 98 were 16×14 and in Windows 2000 the dimensions were 17×15. That is the buttons were not square, the width was 2 pixels greater than the height.

The dimensions of the buttons in the Windows classic theme after your fix are 20×18. The buttons are still larger than they were (the height comes from the visual-style theme) but proportions are correct. Before the fix, the dimensions were 34×18 — the buttons were even wider than in the visual-style enabled theme.

prsadhuk:
> > > Having said that, I guess XP theme also padding is not correct as can be seen in this bug [JDK-8139392](https://bugs.openjdk.org/browse/JDK-8139392) which I have rectified
> > > After fix
> > > image
> > 
> > 
> > Your latest changes also fix [JDK-8139392](https://bugs.openjdk.org/browse/JDK-8139392), do I get it right? If so, add it to the list of fixed issues.
> 
> This change does not address JDK-8139392 fully. The title bar buttons in Vista theme, which are still used for MDI windows, are not square. The dimensions of the native buttons are 32×18. The dimensions of the JInternalFrame title bar buttons after your fix are 24×18; before your fix they were 28×18. That is the width of the buttons is smaller by 8 and 4 pixels. Aren't these the values subtracted from the width value fetched from Windows?
> 

Adjusted to native button dimensions of 32x18

aivanov-jdk:
> > This change does not address JDK-8139392 fully. The title bar buttons in Vista theme, which are still used for MDI windows, are not square. The dimensions of the native buttons are 32×18. The dimensions of the JInternalFrame title bar buttons after your fix are 24×18; before your fix they were 28×18. That is the width of the buttons is smaller by 8 and 4 pixels. Aren't these the values subtracted from the width value fetched from Windows?
> 
> Adjusted to native button dimensions of 32x18

It looks good now.

In scale of 150%, the bottom and right borders of the internal frame buttons disappear and re-appear as you drag an internal frame. But it's a completely different story. At scale of 100% or 200%, the buttons look as they should.

src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/WindowsIconFactory.java
```diff
@@ -286,19 +286,12 @@ public int getIconWidth() {
             int width;
             if (XPStyle.getXP() != null) {
                 // Fix for XP bug where sometimes these sizes aren't updated properly
-                // Assume for now that height is correct and derive width using the
-                // ratio from the uxtheme part
-                width = UIManager.getInt("InternalFrame.titleButtonHeight") -2;
-                Dimension d = XPStyle.getPartSize(Part.WP_CLOSEBUTTON, State.NORMAL);
```
aivanov-jdk, line 292: The part should've been `Part.WP_MDICLOSEBUTTON`.

Yet the returned size is wrong either way. On my Windows 11 system, the dimensions are 14×9 and 3×5 for `WP_CLOSEBUTTON` and `WP_MDICLOSEBUTTON`. Either is far from the real values.

src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/WindowsInternalFrameTitlePane.java
```diff
@@ -77,19 +76,16 @@ protected void installDefaults() {
...
-            buttonWidth = buttonHeight;
-            Dimension d = XPStyle.getPartSize(Part.WP_CLOSEBUTTON, State.NORMAL);
-            if (d != null && d.width != 0 && d.height != 0) {
-                buttonWidth = (int) ((float) buttonWidth * d.width / d.height);
-            }
+            // Assume for now that height is correct and derive width from height
+            buttonWidth = buttonHeight + 14;
         } else {
+            buttonWidth = buttonHeight;
             buttonWidth += 2;
```
aivanov-jdk, line 83:
```suggestion
            buttonWidth = buttonHeight + 2;
```

src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/WindowsInternalFrameTitlePane.java
```diff
@@ -77,19 +76,16 @@ protected void installDefaults() {
...
             Color activeBorderColor =
                     UIManager.getColor("InternalFrame.activeBorderColor");
             setBorder(BorderFactory.createLineBorder(activeBorderColor, 1));
         }
+        UIManager.put("InternalFrame.titleButtonWidth", buttonWidth);
```
aivanov-jdk, line 88: > The only concern is that UI Manager has a wrong width until `JInternalFrame` is shown or initialised. But it has been the case before the fix too.

I was probably wrong in making you save the width into the `UIManager`. First of all, it creates an inconsistency between the width and the height. We don't use the value of height (`InternalFrame.titleButtonHeight`) either, we use the adjusted value only; neither do we store it back — it will break our calculations.

For the sake of backward compatibility, *we shouldn't update `InternalFrame.titleButtonWidth`.*

test/jdk/javax/swing/JInternalFrame/InternalFrameTitleButtonTest.java
```diff
@@ -0,0 +1,116 @@
...
+    private static JInternalFrame iframe;
+
+    public static void main(String[] args) throws Exception {
+        String osName = System.getProperty("os.name");
+        if (!osName.toLowerCase().contains("win")) {
+            System.out.println("The test is applicable only for Windows.");
+            return;
+        }
+        UIManager.setLookAndFeel(
+               "com.sun.java.swing.plaf.windows.WindowsClassicLookAndFeel");
```
aivanov-jdk, line 58: Shall `WindowsLookAndFeel` be tested as well?

test/jdk/javax/swing/JInternalFrame/InternalFrameTitleButtonTest.java
```diff
@@ -0,0 +1,116 @@
...
+                            iframe.getUI()).getNorthPane());
+            int height = title.getHeight();
+            Insets insets = title.getInsets();
+            height = height - insets.top - insets.bottom;
+            for (int i = 0; i < title.getComponentCount(); i++) {
+                Component c = title.getComponent(i);
+                if (c instanceof JButton) {
+                    Icon icon = ((JButton) c).getIcon();
+                    if (icon.getIconHeight() > height - 4
+                        || icon.getIconWidth() > height - 2) {
```
aivanov-jdk, line 103: Should we make the test stricter? According to the code, `width = height + 2` for the classic theme and `width = height + 14` for visual styles. In both cases, `height = UIManager.getInt("InternalFrame.titleButtonHeight") - 4`.

aivanov-jdk (changes requested):
**`WindowsIconFactory.java`**

Could you also remove `Dimension` import? It's unused.

Could you please add `final` modifier to the `part` field in `FrameButtonIcon` (line 178)?

**`WindowsInternalFrameTitlePane.java`**

Could you please organise imports? This will expand all the wildcard imports.

aivanov-jdk:
The values in `UIManager` — `InternalFrame.titleButtonWidth` and `InternalFrame.titleButtonHeight` — come from Windows desktop properties, which are gotten from [`SystemParametersInfo`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-systemparametersinfow) called with `SPI_GETNONCLIENTMETRICS`, they are `iCaptionWidth` and `iCaptionHeight` members of the [`NONCLIENTMETRICS`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-nonclientmetricsw) structure.

https://github.com/openjdk/jdk/blob/292ee630ae32c3b50363b10ffa6090e57ffef1e8/src/java.desktop/windows/native/libawt/windows/awt_DesktopProperties.cpp#L299-L300

Interestingly, the returned values are 32×22 which don't match the width and height of any buttons. They may be correct for the Windows classic theme if it could be enabled.

I wonder if [`GetSystemMetrics`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getsystemmetrics) with `SM_CXSIZE` and `SM_CYSIZE` gives the correct result. Yet it may be too late to change…

aivanov-jdk (changes requested):
In general, looks good except for the couple of comments.

I also think the test can be simplified, more on this below.

aivanov-jdk:
May I suggest a simplified version of the test [`InternalFrameTitleButtonTest.java`](https://github.com/aivanov-jdk/jdk/blob/6decd8ae82ed95f0c4bda9089d1e05eb084567b9/test/jdk/javax/swing/JInternalFrame/InternalFrameTitleButtonTest.java)?

Its `test` method is shorter; I moved the condition to its own method `testButtonSize` where an internal frame title button is passed as well as an offset for width that is different for classic and visual-style-enabled L&F. This also avoids checking which L&F is selected.

Here's [the diff](https://github.com/aivanov-jdk/jdk/compare/ec2c9ab..6decd8a#diff-f941d2d497179f4da7530480b24522ccd05da02ca24c6e3fa55824d4199529e1) on top of your latest update.

What do you think?

src/java.desktop/windows/classes/com/sun/java/swing/plaf/windows/WindowsInternalFrameTitlePane.java
```diff
@@ -27,13 +27,36 @@
...
+import javax.swing.JPopupMenu;
+import javax.swing.JSeparator;
+import javax.swing.LookAndFeel;
+import javax.swing.UIDefaults;
 import javax.swing.UIManager;
-import javax.swing.plaf.*;
 import javax.swing.plaf.basic.BasicInternalFrameTitlePane;
-import java.awt.*;
-import java.awt.event.*;
+import javax.swing.plaf.UIResource;
```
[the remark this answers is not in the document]
prsadhuk: ok..updated..

prsadhuk:
> May I suggest a simplified version of the test [`InternalFrameTitleButtonTest.java`](https://github.com/aivanov-jdk/jdk/blob/6decd8ae82ed95f0c4bda9089d1e05eb084567b9/test/jdk/javax/swing/JInternalFrame/InternalFrameTitleButtonTest.java)?
> 
> Its `test` method is shorter; I moved the condition to its own method `testButtonSize` where an internal frame title button is passed as well as an offset for width that is different for classic and visual-style-enabled L&F. This also avoids checking which L&F is selected.
> 
> Here's [the diff](https://github.com/aivanov-jdk/jdk/compare/ec2c9ab..6decd8a#diff-f941d2d497179f4da7530480b24522ccd05da02ca24c6e3fa55824d4199529e1) on top of your latest update.
> 
> What do you think?

OK. updated to your version..seems precise..thanks...

TejeshR13 (approved):
Looks good to me.

PR #14562, opened 2023-06-20
JDK-8310380: Handle problems in core-related tests on macOS when codesign tool does not work

MBaesken:
Currently, a number of tests fail on macOS because they miss the core file (e.g. serviceability/sa/TestJmapCore.java).
The reason is that configure detects on some setups that codesign does not work ("checking if debug mode codesign is possible... no) .
So adding the needed entitlement to generate cores is not done in the build. This is currently not checked later in the tests.
But without the entitlement, a core is not generated.

---------
### Progress

### Issue
 * [JDK-8310380](https://bugs.openjdk.org/browse/JDK-8310380): Handle problems in core-related tests on macOS when codesign tool does not work (**Bug** - P4)

### Reviewers
 * [Lutz Schmidt](https://openjdk.org/census#lucy) (@RealLucy - **Reviewer**) ⚠️ Review applies to [0d8c68c2](https://git.openjdk.org/jdk/pull/14562/files/0d8c68c2e554abc94542053802d45cba95fb2fc6)
 * [Christoph Langer](https://openjdk.org/census#clanger) (@RealCLanger - **Reviewer**) ⚠️ Review applies to [619b3578](https://git.openjdk.org/jdk/pull/14562/files/619b3578defdcecdf45ff75cb7a8c84ccc1b7b73)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14562/head:pull/14562` \
`$ git checkout pull/14562`

`$ git checkout pull/14562` \
`$ git pull https://git.openjdk.org/jdk.git pull/14562/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14562`

`$ git pr show -t 14562`

</details>

<a href="https://git.openjdk.org/jdk/pull/14562.diff">https://git.openjdk.org/jdk/pull/14562.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14562#issuecomment-1598798179)

### DIFF
--- a/test/lib/jdk/test/lib/Platform.java
+++ b/test/lib/jdk/test/lib/Platform.java
@@ -260,6 +260,34 @@ public class Platform {
         return true;
     }
 
+
+    public static boolean hasPlistEntriesOSX() throws IOException {
+        // Find the path to the java binary.
+        String jdkPath = System.getProperty("java.home");
+        Path javaPath = Paths.get(jdkPath + "/bin/java");
+        String javaFileName = javaPath.toAbsolutePath().toString();
+        if (Files.notExists(javaPath)) {
+            throw new FileNotFoundException("Could not find file " + javaFileName);
+        }
+
+        // Run codesign on the java binary.
+        ProcessBuilder pb = new ProcessBuilder("codesign", "--display", "--verbose", javaFileName);
+        pb.redirectErrorStream(true); // redirect stderr to stdout
+        Process codesignProcess = pb.start();
+        BufferedReader is = new BufferedReader(new InputStreamReader(codesignProcess.getInputStream()));
+        String line;
+        while ((line = is.readLine()) != null) {
+            System.out.println("STDOUT: " + line);
+            if (line.indexOf("Info.plist=not bound") != -1) {
+                return false;
+            }
+            if (line.indexOf("Info.plist entries=") != -1) {
+                return true;
+            }
+        }
+        return false;
+    }
+
     /**
      * Return true if the test JDK is hardened, otherwise false. Only valid on OSX.
      */
--- a/test/lib/jdk/test/lib/Platform.java
+++ b/test/lib/jdk/test/lib/Platform.java
@@ -273,7 +273,7 @@ public class Platform {
         return codesignProcess;
     }
 
-    public static boolean hasPlistEntriesOSX() throws IOException {
+    public static boolean hasOSXPlistEntries() throws IOException {
         Process codesignProcess = launchCodesignOnJavaBinary();
         BufferedReader is = new BufferedReader(new InputStreamReader(codesignProcess.getInputStream()));
         String line;
--- a/test/lib/jdk/test/lib/Platform.java
+++ b/test/lib/jdk/test/lib/Platform.java
@@ -260,7 +260,13 @@ public class Platform {
         return true;
     }
 
-    private static codesignProcess launchCodesignOnJavaBinary(String javaFileName) {
+    private static Process launchCodesignOnJavaBinary() throws IOException {
+        String jdkPath = System.getProperty("java.home");
+        Path javaPath = Paths.get(jdkPath + "/bin/java");
+        String javaFileName = javaPath.toAbsolutePath().toString();
+        if (Files.notExists(javaPath)) {
+            throw new FileNotFoundException("Could not find file " + javaFileName);
+        }
         ProcessBuilder pb = new ProcessBuilder("codesign", "--display", "--verbose", javaFileName);
         pb.redirectErrorStream(true); // redirect stderr to stdout
         Process codesignProcess = pb.start();
@@ -268,13 +274,7 @@ public class Platform {
     }
 
     public static boolean hasPlistEntriesOSX() throws IOException {
-        Path javaPath = Utils.javaPath();
-        String javaFileName = javaPath.toAbsolutePath().toString();
-        if (Files.notExists(javaPath)) {
-            throw new FileNotFoundException("Could not find file " + javaFileName);
-        }
-
-        Process codesignProcess = launchCodesignOnJavaBinary(javaFileName);
+        Process codesignProcess = launchCodesignOnJavaBinary();
         BufferedReader is = new BufferedReader(new InputStreamReader(codesignProcess.getInputStream()));
         String line;
         while ((line = is.readLine()) != null) {
@@ -299,14 +299,7 @@ public class Platform {
         if (getOsVersionMajor() == 10 && getOsVersionMinor() < 14) {
             return false; // assume not hardened
         }
-
-        Path javaPath = Utils.javaPath();
-        String javaFileName = javaPath.toAbsolutePath().toString();
-        if (Files.notExists(javaPath)) {
-            throw new FileNotFoundException("Could not find file " + javaFileName);
-        }
-
-        Process codesignProcess = launchCodesignOnJavaBinary(javaFileName);
+        Process codesignProcess = launchCodesignOnJavaBinary();
         BufferedReader is = new BufferedReader(new InputStreamReader(codesignProcess.getInputStream()));
         String line;
         boolean isHardened = false;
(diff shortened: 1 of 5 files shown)


RealLucy (approved):
LGTM.
Thanks for enhancing test analysis.

test/lib/jdk/test/lib/util/CoreUtils.java
```diff
@@ -148,6 +150,11 @@ public static String getCoreFileLocation(String crashOutputString, long pid) thr
                     // We can't generate cores files with hardened binaries on OSX 10.15 and later.
                     throw new SkippedException("Cannot produce core file with hardened binary on OSX 10.15 and later");
                 }
+            } else {
```
RealCLanger, line 153: Maybe you could do just one case here:
`else if (!Platform.hasPlistEntriesOSX()) {`...

test/lib/jdk/test/lib/Platform.java
```diff
@@ -260,6 +260,34 @@ public static boolean hasSA() {
...
+        // Find the path to the java binary.
+        String jdkPath = System.getProperty("java.home");
+        Path javaPath = Paths.get(jdkPath + "/bin/java");
+        String javaFileName = javaPath.toAbsolutePath().toString();
+        if (Files.notExists(javaPath)) {
+            throw new FileNotFoundException("Could not find file " + javaFileName);
+        }
+
+        // Run codesign on the java binary.
+        ProcessBuilder pb = new ProcessBuilder("codesign", "--display", "--verbose", javaFileName);
```
RealCLanger, line 274: And then here 'ProcessBuilder pb = new ProcessBuilder("codesign", "--display", "--verbose", javaPath.toAbsolutePath().toString());`

test/lib/jdk/test/lib/Platform.java
```diff
@@ -260,6 +260,34 @@ public static boolean hasSA() {
...
+        pb.redirectErrorStream(true); // redirect stderr to stdout
+        Process codesignProcess = pb.start();
+        BufferedReader is = new BufferedReader(new InputStreamReader(codesignProcess.getInputStream()));
```
RealCLanger, line 277: Maybe do the BufferedReader stuff in a try-with-resources and then return false instead of potentially throwing an IOException?

test/lib/jdk/test/lib/Platform.java
```diff
@@ -260,6 +260,34 @@ public static boolean hasSA() {
...
+        String line;
+        while ((line = is.readLine()) != null) {
+            System.out.println("STDOUT: " + line);
```
RealCLanger, line 280: This output is for every line seems too much. Maybe just print the lines where you find "Info.plist=not bound" or "Info.plist entries="?

RealCLanger (commented):
I made a few review suggestions. Does the symptom happen on both, arm64 and x64?

test/lib/jdk/test/lib/Platform.java
```diff
@@ -260,6 +260,34 @@ public static boolean hasSA() {
...
+            if (line.indexOf("Info.plist=not bound") != -1) {
+                return false;
+            }
+            if (line.indexOf("Info.plist entries=") != -1) {
+                return true;
+            }
+        }
+        return false;
```
plummercj, line 288: Probably would be good to log that no Info.plist entry was found.

test/lib/jdk/test/lib/util/CoreUtils.java
```diff
@@ -127,6 +127,8 @@ public static String getCoreFileLocation(String crashOutputString, long pid) thr
             }
 
             return coreFileLocation; // success!
+        } else {
+            System.out.println("Core file not found, try to find a reason for this");
```
plummercj, line 131:
```suggestion
            System.out.println("Core file not found. Trying to find a reason why...");
```

test/lib/jdk/test/lib/Platform.java
```diff
@@ -260,6 +260,34 @@ public static boolean hasSA() {
         return true;
     }
 
+
+    public static boolean hasPlistEntriesOSX() throws IOException {
```
plummercj, line 264: Almost all of this is replicated from `isHardenedOSX()`. I wonder if there is a way to do some sharing while still maintaining separate APIs. Combining them into one API might make it harder to understand the code. Maybe a `launchCodesign()` API that returns the BufferedReader would help.

plummercj (commented):
Copyrights need updating.

plummercj (changes requested):
Copyrights need updating.

plummercj:
`jdk/test/lib/TestMutuallyExclusivePlatformPredicates.java` is failing due to the new API in Platform.java.

test/lib/jdk/test/lib/Platform.java
MBaesken: I found the output useful when analyzing the issue (and it is just a few lines).

MBaesken:
I adjusted COPYRIGHT info, adjusted TestMutuallyExclusivePlatformPredicates.java .
Also simplified the coding in Platform.java a bit.

test/lib/jdk/test/lib/util/CoreUtils.java
MBaesken: I changed the message.

test/lib/jdk/test/lib/Platform.java
```diff
@@ -260,6 +260,36 @@ public static boolean hasSA() {
         return true;
     }
 
+    private static codesignProcess launchCodesignOnJavaBinary(String javaFileName) {
```
RealCLanger, line 263: That can't work... should be `private static Process` 😉

RealCLanger (changes requested):
Looks a bit better. But I think instead of adding a Utils.javaPath() method, you can do all this path handling in Platform.launchCodesignOnJavaBinary(). Then even more code would be shared.

MBaesken:
Yes it must be 'private static Process' . And I moved the javaPath method into the caller ,

MBaesken:
> Looks a bit better. But I think instead of adding a Utils.javaPath() method, you can do all this path handling in Platform.launchCodesignOnJavaBinary(). Then even more code would be shared.

Okay, had had the idea  to later use  the method javaPath  at some more places, but let's do this separately.

RealCLanger (approved):
Looks good to me now.

MBaesken:
Hi Christoph, thanks for the review !
Chris are you okay with the latest rev.  of this change?

plummercj:
> Chris are you okay with the latest rev. of this change?

Sorry, I've been on vacation the past few days. I will have a look at it tomorrow.

test/lib/jdk/test/lib/Platform.java
```diff
@@ -260,28 +260,46 @@ public static boolean hasSA() {
...
         }
-
-        // Run codesign on the java binary.
         ProcessBuilder pb = new ProcessBuilder("codesign", "--display", "--verbose", javaFileName);
         pb.redirectErrorStream(true); // redirect stderr to stdout
         Process codesignProcess = pb.start();
+        return codesignProcess;
+    }
+
+    public static boolean hasPlistEntriesOSX() throws IOException {
```
plummercj, line 276: Although I understand why you chose this API name (it mimics the form used for `isHardenedOSX`), I don't think it is a good choice. `isHardenedOSX` is short for "is this a hardened OSX system". That mapping does not work with `hasPlistEntriesOSX`, which is more like "does this OSX system have plist entries". I would suggest `hasOSXPlistEntries`.

test/lib/jdk/test/lib/util/CoreUtils.java
```diff
@@ -148,6 +150,11 @@ public static String getCoreFileLocation(String crashOutputString, long pid) thr
                     // We can't generate cores files with hardened binaries on OSX 10.15 and later.
                     throw new SkippedException("Cannot produce core file with hardened binary on OSX 10.15 and later");
                 }
+            } else {
+                // codesign has to add entitlements using the plist, if this is not present we might not generate a core file
```
plummercj, line 154:
```suggestion
                // codesign has to add entitlements using the plist. If this is not present we might not generate a core file.
```

plummercj (changes requested):

MBaesken:
Hi Chris,  I renamed the method and adjusted the comment.

plummercj (approved):

MBaesken:
Thanks for the review !

/integrate

PR #14565, opened 2023-06-20
8310405: Linker.Option.firstVariadicArg should specify which index values are valid

JornVernee:
Improve the specification of `Linker.Option.firstVariadicArg`, by specifying more clearly which index values are valid.

---------
### Progress

### Issues
 * [JDK-8310405](https://bugs.openjdk.org/browse/JDK-8310405): Linker.Option.firstVariadicArg should specify which index values are valid (**Enhancement** - P3)
 * [JDK-8310836](https://bugs.openjdk.org/browse/JDK-8310836): Linker.Option.firstVariadicArg should specify which index values are valid (**CSR**) (Withdrawn)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14565/head:pull/14565` \
`$ git checkout pull/14565`

`$ git checkout pull/14565` \
`$ git pull https://git.openjdk.org/jdk.git pull/14565/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14565`

`$ git pr show -t 14565`

</details>

<a href="https://git.openjdk.org/jdk/pull/14565.diff">https://git.openjdk.org/jdk/pull/14565.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14565#issuecomment-1599040933)

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/Linker.java
+++ b/src/java.base/share/classes/java/lang/foreign/Linker.java
@@ -416,8 +416,8 @@ import java.util.stream.Stream;
  * The native linker only supports linking the specialized form of a variadic function. A variadic function in its specialized
  * form can be linked using a function descriptor describing the specialized form. Additionally, the
  * {@link Linker.Option#firstVariadicArg(int)} linker option must be provided to indicate the first variadic parameter in
- * the parameter list. The corresponding argument layout, and all following argument layouts in the specialized function
- * descriptor, are called <em>variadic argument layouts</em>. For a prototype-less function, the index passed to
+ * the parameter list. The corresponding argument layout (if any), and all following argument layouts in the specialized
+ * function descriptor, are called <em>variadic argument layouts</em>. For a prototype-less function, the index passed to
  * {@link Linker.Option#firstVariadicArg(int)} should always be {@code 0}.
  * <p>
  * The native linker will reject an attempt to link a specialized function descriptor with any variadic argument layouts
@@ -644,20 +644,25 @@ public sealed interface Linker permits AbstractLinker {
             permits LinkerOptions.LinkerOptionImpl {
 
         /**
-         * {@return a linker option used to denote the index of the first variadic argument layout in the
-         *          function descriptor associated with a downcall linkage request}
+         * {@return a linker option used to denote the index indicating the start of the variadic arguments passed to the
+         *          function described by the function descriptor associated with a downcall linkage request}
          * <p>
-         * The {@code index} value must be greater than zero, and less than or equal to the number of argument layouts
-         * of the function descriptor that is used in the same linkage request as this option. When the {@code index} is
-         * equal to the number of argument layouts in the descriptor, it indicates a variadic function to which zero
-         * variadic arguments are passed. It is important to always use this linker option when linking a variadic function,
-         * as this affects the linking process on certain platforms (even if no variadic arguments are passed).
+         * The {@code index} value must conform to {@code 0 <= index <= N}, where {@code N} is the number of argument
+         * layouts of the function descriptor used in conjunction with this linker option. When the {@code index} is:
+         * <ul>
+         * <li>{@code 0}, all arguments passed to the function are passed as variadic arguments</li>
+         * <li>{@code N}, none of the arguments passed to the function are passed as variadic argument</li>
+         * <li>{@code n}, where {@code 0 < m < N}, the arguments {@code m..N} are passed as variadic arguments</li>
+         * </ul>
+         * It is important to always use this linker option when linking a <a href=Linker.html#variadic-funcs>variadic
+         * function</a>, even when none of the arguments are passed as variadic arguments (the second case in the list
+         * above), as this affects the linking process on certain platforms.
          *
          * @implNote The index value is validated when making a linkage request, which is when the function descriptor
-         * against which the index is validated is available.
+         *           against which the index is validated is available.
          *
-         * @param index the index of the first variadic argument layout in the function descriptor associated
-         *              with a downcall linkage request.
+         * @param index the index indicating the start of the variadic arguments passed to the function described by
+         *              the function descriptor used in conjunction with this linker option.
          */
         static Option firstVariadicArg(int index) {
             return new LinkerOptions.FirstVariadicArg(index);
--- a/src/java.base/share/classes/java/lang/foreign/Linker.java
+++ b/src/java.base/share/classes/java/lang/foreign/Linker.java
@@ -655,7 +655,7 @@ public sealed interface Linker permits AbstractLinker {
          * <li>{@code n}, where {@code 0 < m < N}, the arguments {@code m..N} are passed as variadic arguments</li>
          * </ul>
          * It is important to always use this linker option when linking a <a href=Linker.html#variadic-funcs>variadic
-         * function</a>, even when none of the arguments are passed as variadic arguments (the second case in the list
+         * function</a>, even if no variadic argument is passed (the second case in the list
          * above), as this affects the linking process on certain platforms.
          *
          * @implNote The index value is validated when making a linkage request, which is when the function descriptor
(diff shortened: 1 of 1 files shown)


src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -646,6 +646,16 @@ static Linker nativeLinker() {
         /**
          * {@return a linker option used to denote the index of the first variadic argument layout in the
          *          function descriptor associated with a downcall linkage request}
+         * <p>
+         * The {@code index} value must be greater than zero, and less than or equal to the number of argument layouts
```
mcimadamore, line 650: Why > 0 and not >= 0? If the list of argument layout is empty, wouldn't the first vararg be in position 0? E.g. a call to a function `m(...)` with _no_ actual arguments?

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -646,6 +646,16 @@ static Linker nativeLinker() {
...
+         * of the function descriptor that is used in the same linkage request as this option. When the {@code index} is
+         * equal to the number of argument layouts in the descriptor, it indicates a variadic function to which zero
```
mcimadamore, line 652: This sentence is not clear. If index == num layouts then it means *all* arguments are variadic?

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -646,6 +646,16 @@ static Linker nativeLinker() {
...
+         * variadic arguments are passed. It is important to always use this linker option when linking a variadic function,
+         * as this affects the linking process on certain platforms (even if no variadic arguments are passed).
+         *
+         * @implNote The index value is validated when making a linkage request, which is when the function descriptor
+         * against which the index is validated is available.
+         *
          * @param index the index of the first variadic argument layout in the function descriptor associated
```
mcimadamore, line 659: IMHO this description is still confusing because it speaks about an index into the function descriptor argument layouts - which this thing is not.

src/java.base/share/classes/java/lang/foreign/Linker.java
mcimadamore: E.g. are we making an assumption here that if no arguments are passed, then the calling sequence won't be affected by variadic-ness - is that assumption always true?

JornVernee: Ah, sorry, this is just a  mistake I made. It should be >= 0

src/java.base/share/classes/java/lang/foreign/Linker.java
JornVernee: index == 0 would mean all arguments are variadic. index == num layouts means none. (just like `"abcd".substring(4)` returns an empty string).

src/java.base/share/classes/java/lang/foreign/Linker.java
JornVernee: Right, will polish this some more as well.

src/java.base/share/classes/java/lang/foreign/Linker.java
mcimadamore: Yes, that's what I meant and got confused. But I missed the all-important `zero` in the javadoc - so I think that this bit of text is ok as it is. Apologize for the confusion.

src/java.base/share/classes/java/lang/foreign/Linker.java
mcimadamore: Possible suggestion/thing to try: use a bullet list to spell out all cases for `index`. E.g. we know there's index == 0 (all variadic). Then we know there's index = N (no variadic). Then there's index == m, 0 < m < N - which means layouts 0..m are non-variadic and m..N are variadic (where n..m denotes an interval with n included and m excluded).

minborg: > Possible suggestion/thing to try: use a bullet list to spell out all cases for `index`. E.g. we know there's index == 0 (all variadic). Then we know there's index = N (no variadic). Then there's index == m, 0 < m < N - which means layouts 0..m are non-variadic and m..N are variadic (where n..m denotes an interval with n included and m excluded).

I think this is a good suggestion. It makes it much easier to understand.

JornVernee: Thanks for the review.

I've added a bullet list, and switch same of the language to refer to the 'start of the variadic arguments passed to the function described by the function descriptor'. I think the latter avoids implying the index is an index into the argument layouts. (I've also added a small note to the global variadic function doc to indicate that the index might not necessarily have a corresponding argument layout).

How does the new version look?

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -644,20 +644,25 @@ sealed interface Option
...
-         * The {@code index} value must be greater than zero, and less than or equal to the number of argument layouts
-         * of the function descriptor that is used in the same linkage request as this option. When the {@code index} is
-         * equal to the number of argument layouts in the descriptor, it indicates a variadic function to which zero
-         * variadic arguments are passed. It is important to always use this linker option when linking a variadic function,
-         * as this affects the linking process on certain platforms (even if no variadic arguments are passed).
+         * The {@code index} value must conform to {@code 0 <= index <= N}, where {@code N} is the number of argument
+         * layouts of the function descriptor used in conjunction with this linker option. When the {@code index} is:
+         * <ul>
+         * <li>{@code 0}, all arguments passed to the function are passed as variadic arguments</li>
+         * <li>{@code N}, none of the arguments passed to the function are passed as variadic argument</li>
```
mcimadamore, line 654:
```suggestion
         * <li>{@code N}, none of the arguments passed to the function are passed as variadic argumenst</li>
```

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -644,20 +644,25 @@ sealed interface Option
...
+         * <li>{@code n}, where {@code 0 < m < N}, the arguments {@code m..N} are passed as variadic arguments</li>
+         * </ul>
+         * It is important to always use this linker option when linking a <a href=Linker.html#variadic-funcs>variadic
+         * function</a>, even when none of the arguments are passed as variadic arguments (the second case in the list
```
mcimadamore, line 658:
```suggestion
         * function</a>, even if no variadic argument is passed (the second case in the list
```

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -644,20 +644,25 @@ sealed interface Option
...
+         * above), as this affects the linking process on certain platforms.
```
mcimadamore, line 659:
```suggestion
         * above), as this might still affect the calling convention on certain platforms.
```

mcimadamore: Note: we use the term calling convention in Linker javadoc - I think it's slightly more precise than "linking process".

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -644,20 +644,25 @@ sealed interface Option
...
          *
          * @implNote The index value is validated when making a linkage request, which is when the function descriptor
-         * against which the index is validated is available.
+         *           against which the index is validated is available.
          *
-         * @param index the index of the first variadic argument layout in the function descriptor associated
-         *              with a downcall linkage request.
+         * @param index the index indicating the start of the variadic arguments passed to the function described by
```
mcimadamore, line 664: My general feeling is that the previous text was better here. We describe in plenty details what this "index" is in the method javadoc. I think I'm more for reverting the changes here.

JornVernee:
I've addressed the review comments. I think we're converging so also went ahead and created the CSR for this.

mcimadamore (approved):

PR #14566, opened 2023-06-20
8310426: (ch) Channels.newInputStream transferTo cleanup

AlanBateman:
ChannelInputStream, the InputStream returned by Channels.newInputStream, implements the transferTo method with support for the cross product of many interesting underlying channels. It's getting a bit too unwieldy and can be simplified by introducing a FileChannelInputStream to wrap a FileChannel.

---------
### Progress

### Issue
 * [JDK-8310426](https://bugs.openjdk.org/browse/JDK-8310426): (ch) Channels.newInputStream transferTo cleanup (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14566/head:pull/14566` \
`$ git checkout pull/14566`

`$ git checkout pull/14566` \
`$ git pull https://git.openjdk.org/jdk.git pull/14566/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14566`

`$ git pr show -t 14566`

</details>

<a href="https://git.openjdk.org/jdk/pull/14566.diff">https://git.openjdk.org/jdk/pull/14566.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14566#issuecomment-1599252845)

### DIFF
--- a/test/jdk/java/nio/channels/Channels/TransferTo.java
+++ b/test/jdk/java/nio/channels/Channels/TransferTo.java
@@ -22,6 +22,7 @@
  */
 
 import java.io.ByteArrayOutputStream;
+import java.io.FileOutputStream;
 import java.io.InputStream;
 import java.io.IOException;
 import java.io.OutputStream;
@@ -49,8 +50,7 @@ import static org.testng.Assert.assertThrows;
  * @build jdk.test.lib.RandomFactory
  * @run testng/othervm/timeout=180 TransferTo
  * @bug 8265891
- * @summary Tests whether sun.nio.ChannelInputStream.transferTo conforms to the
- *          InputStream.transferTo specification
+ * @summary Tests Channels.newInputStream.transferTo
  * @key randomness
  */
 public class TransferTo extends TransferToBase {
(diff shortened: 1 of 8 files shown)


test/jdk/java/nio/channels/Channels/TransferTo.java
```diff
@@ -49,8 +50,7 @@
  * @build jdk.test.lib.RandomFactory
  * @run testng/othervm/timeout=180 TransferTo
  * @bug 8265891
```
bplb, line 52: Add 8310426?

mkarg (commented):
The undoubted benefits (pros) of this PR is:
* A cleaner ChannelInputStream::transferTo implementation.

The obvious drawbacks (contras) of this PR are:
* The overall optimization solution ("the sum of all optimizations") is much harder to grasp now (it took me half an hour just to check if we still cover the full cross product), as the solution is scattered over lots of files now. In contrast, a central place was much simpler to grasp before this PR.
* IIUC the PR needed to duplicate algorithms just for the sole sake of code cleanup. In contrast, a central place was the sole source of truth for all transfer algorithms before this PR.
* The new solution is much more complex and comes with additional classes just for the sole sake of a method's cleanup; those classes apparently have no other benefit, at least so far.

IMHO the single benefit does not outweigh the multiple drawbacks.

A counter proposal for a compromise between this PR (code duplication) and the original code (single source of truth) could be:
* Let's refactor the cross product (the original trigger for this PR) on the channel level (not on the stream level) into a new static utility method Channels::transfer(ReadableByteChannel, WritableByteChannel), and let's call that single method from all code locations where "some channels" need to transfer bytes.

The benefit of the counter proposal would be:
* Keeping a single source of truth, which does not need code duplication, and which is easy to grasp (all optimization rules in one place).
* Can be used with either the original code and/or the new PR.
* Can simply get centrally extended in case a future OS comes up with a new type of optimization API eventually, e. g. hardware-implemented Socket-to-Socket transfers.
* Could even get used by other JDK locations, and/or custom application code that wants to perform optimized non-stream channel-to-channel transfers.

AlanBateman:
> IMHO the single benefit does not outweigh the multiple drawbacks.

SocketInputStream is already tightly coupled to the underlying SocketChannelImpl, it's not a ChannelInputStream. More is required here to avoid needing the use the blockingLock, make SIS.transferTo friendly to virtual thread usages, and also to coordinate with async close of a target FileChannel. With FileInputStream, the idea is the same, it will be able to coordinate with FileChannelImpl internals.

I think ChannelInputStream should be a simple input stream based on a readable channel. It can delegate to FileChannel.transferFrom when the target is an output stream that is based on a FileChannel but beyond that, I don't think we want it to have a close relationship with SocketChannelImpl or FileChannel internals.

mkarg:
I am planning to provide optimizations for Channel-based Reader::transferTo in future, just like I did for InputStream::transferTo in the past. If we would have Channels::transfer(RBW, MBW) then I could simply reuse proven and trusted code. If we separate the code, then I have to provide (possibly several) new classes. I better understand and appreciate the merits of your new classes now thanks to you explanation, but it would be beneficial if the actual channel-to-channel algorithm could get reused.

PR #14567, opened 2023-06-20
JDK-8310453: Update javadoc of java.lang.Object

jddarcy:
I had occasion to read over the javadoc sources in java.lang.Object recently and noticed a few items that could be updated.

There are some new or expanded API notes referring to methods in java.util.Objects. I added these references as apiNote items rather than, say, `@see` tags since  `@see` tag changes would propagate into classes that overrode the methods in questions.

Changing toString to use an inline `@return` tag has the consequence of omitting a trailing period, "." in the "Returns" section of its javadoc. This also omits  a trailing period in subclasses that use the `@return` statement of Object.toString in their own toString method. Likewise for hashCode.

---------
### Progress

### Issue
 * [JDK-8310453](https://bugs.openjdk.org/browse/JDK-8310453): Update javadoc of java.lang.Object (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14567/head:pull/14567` \
`$ git checkout pull/14567`

`$ git checkout pull/14567` \
`$ git pull https://git.openjdk.org/jdk.git pull/14567/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14567`

`$ git pr show -t 14567`

</details>

<a href="https://git.openjdk.org/jdk/pull/14567.diff">https://git.openjdk.org/jdk/pull/14567.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14567#issuecomment-1599205551)

### DIFF
--- a/src/java.base/share/classes/java/lang/Object.java
+++ b/src/java.base/share/classes/java/lang/Object.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1994, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1994, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -67,7 +67,7 @@ public class Object {
     public final native Class<?> getClass();
 
     /**
-     * Returns a hash code value for the object. This method is
+     * {@return a hash code value for this object} This method is
      * supported for the benefit of hash tables such as those provided by
      * {@link java.util.HashMap}.
      * <p>
@@ -95,7 +95,11 @@ public class Object {
      * As far as is reasonably practical, the {@code hashCode} method defined
      * by class {@code Object} returns distinct integers for distinct objects.
      *
-     * @return  a hash code value for this object.
+     * @apiNote
+     * The {@link java.util.Objects#hash(Object...) hash} and {@link
+     * java.util.Objects#hashCode(Object) hashCode} methods of {@link
+     * java.util.Objects} can be used to help construct simple hash codes.
+     *
      * @see     java.lang.Object#equals(java.lang.Object)
      * @see     java.lang.System#identityHashCode
      */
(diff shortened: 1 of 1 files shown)


RogerRiggs (approved):
Looks good.

src/java.base/share/classes/java/lang/Object.java
```diff
@@ -246,12 +253,14 @@ public boolean equals(Object obj) {
...
-     * <blockquote>
-     * <pre>
+     * {@snippet lang=java :
      * getClass().getName() + '@' + Integer.toHexString(hashCode())
-     * </pre></blockquote>
-     *
-     * @return  a string representation of the object.
+     * }
+     * The {@link java.util.Objects#toIdentityString(Object)
+     * Objects.toIdentityString} method returns the string for an
```
pavelrappo, line 260: That last sentence doesn't feel like it needs to be part of `@implSpec`, although there's definitely a connection.

pavelrappo, line 260: Separately, is it just me or this sentence could be rephrased to not read like `toIdentityString(Object)` accepts an object equal to the string (that would ...)?

pavelrappo: That is, a clear connection to what's being described by that `@implSpec`.

pavelrappo:
While it's out of scope of this PR, I note that we could also use the inline variant of `@return` in those methods of `Objects` that the `Object` refers to: `hash(Object)`, `hashCode(Object...)`, and `equals(Object, Object)`. In fact, I think they are crying for it.

src/java.base/share/classes/java/lang/Object.java
AlanBateman: This method has an existing apiNote and implSpec. I suspect Joe meant to add this sentence to the apiNote, not the implSpec.

jddarcy: I did mean to add the cross-reference to Objects.toIdentityString to the implSpec section since earlier in the implSpec section the expression that is the basis for Objects.toIdentityString is listed.

pavelrappo:
> While it's out of scope of this PR, I note that we could also use the inline variant of `@return` in those methods of `Objects` that the `Object` refers to: `hash(Object)`, `hashCode(Object...)`, and `equals(Object, Object)`. In fact, I think they are crying for it.

@jddarcy, I see you have since published a separate PR that includes those methods: #14608; thanks!

src/java.base/share/classes/java/lang/Object.java
AlanBateman: okay

PR #6213, opened 2021-11-02
8276348: Use blessed modifier order in java.base

pavelrappo:
This PR follows up one of the recent PRs, where I used a non-canonical modifier order. Since the problem was noticed [^1], why not to address it en masse?

As far as I remember, the first mass-canonicalization of modifiers took place in JDK-8136583 in 2015 [^2]. That change affected 1780 lines spanning 453 files. Since then modifiers have become a bit loose, and it makes sense to re-bless (using the JDK-8136583 terminology) them.

This change was produced by running the below command followed by updating the copyright years on the affected files where necessary:

    $ sh ./bin/blessed-modifier-order.sh src/java.base

The resulting change is much smaller than that of 2015: 39 lines spanning 21 files.

[^1]: https://mail.openjdk.java.net/pipermail/core-libs-dev/2021-November/082987.html (or https://github.com/openjdk/jdk/pull/6191#pullrequestreview-794333365)
[^2]: http://mail.openjdk.java.net/pipermail/core-libs-dev/2015-September/035217.html

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed

### Issue
 * [JDK-8276348](https://bugs.openjdk.java.net/browse/JDK-8276348): Use blessed modifier order in java.base

### Reviewers
 * [Daniel Fuchs](https://openjdk.java.net/census#dfuchs) (@dfuch - **Reviewer**)
 * [Joe Darcy](https://openjdk.java.net/census#darcy) (@jddarcy - **Reviewer**)
 * [Iris Clark](https://openjdk.java.net/census#iris) (@irisclark - **Reviewer**)
 * [Roger Riggs](https://openjdk.java.net/census#rriggs) (@RogerRiggs - **Reviewer**)
 * [Martin Buchholz](https://openjdk.java.net/census#martin) (@Martin-Buchholz - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk pull/6213/head:pull/6213` \
`$ git checkout pull/6213`

Update a local copy of the PR: \
`$ git checkout pull/6213` \
`$ git pull https://git.openjdk.java.net/jdk pull/6213/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 6213`

View PR using the GUI difftool: \
`$ git pr show -t 6213`

[...]

PR #14574, opened 2023-06-21
JDK-8310267: Javadoc for Class#isPrimitive() is incorrect regarding Class objects for primitives

jddarcy:
Correct misstatement that the Class object for a primitive type can only be be access via fields like java.lang.Integer.TYPE.

---------
### Progress

### Issue
 * [JDK-8310267](https://bugs.openjdk.org/browse/JDK-8310267): Javadoc for Class#isPrimitive() is incorrect regarding Class objects for primitives (**Bug** - P4)

### Reviewers
 * [Pavel Rappo](https://openjdk.org/census#prappo) (@pavelrappo - **Reviewer**) ⚠️ Review applies to [9f4a5555](https://git.openjdk.org/jdk/pull/14574/files/9f4a5555e6698c2b4685e98cc986a57a1eab55dc)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14574/head:pull/14574` \
`$ git checkout pull/14574`

`$ git checkout pull/14574` \
`$ git pull https://git.openjdk.org/jdk.git pull/14574/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14574`

`$ git pr show -t 14574`

</details>

<a href="https://git.openjdk.org/jdk/pull/14574.diff">https://git.openjdk.org/jdk/pull/14574.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14574#issuecomment-1599757996)


src/java.base/share/classes/java/lang/Class.java
```diff
@@ -812,14 +812,19 @@ public Void run() {
      *
      * <p> There are nine predefined {@code Class} objects to represent
      * the eight primitive types and void.  These are created by the Java
-     * Virtual Machine, and have the same names as the primitive types that
+     * Virtual Machine, and have the same {@linkplain #getName() names} as the primitive types that
      * they represent, namely {@code boolean}, {@code byte},
      * {@code char}, {@code short}, {@code int},
      * {@code long}, {@code float}, and {@code double}.
```
liach, line 818: Should we add `{@code void}` to the list here, as this is one of the primitive type names?

jddarcy: Hmm. I'll consider that. The javadoc in java.lang.Class is inconsistent in the formatting of "void" as a type name, some instances are in code markup while others are not.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -812,14 +812,19 @@ public Void run() {
...
      *
-     * <p> These objects may only be accessed via the following public static
-     * final variables, and are the only {@code Class} objects for which
-     * this method returns {@code true}.
+     * <p>No other class objects are considered primitive.
+     *
+     * @apiNote
+     * These {@code Class} objects can be accessed via the {@code
```
AlanBateman, line 823: It might be clearer if the API note starts with something like "A Class object representing a primitive type" rather than "These Class objects".

src/java.base/share/classes/java/lang/Class.java
pavelrappo: @liach please have a look at this intensive exchange between David Holmes (@dholmes-ora) and me: https://github.com/openjdk/jdk/pull/6213#issuecomment-960420076 That exchange relates to other keywords that might appear in plain (i.e. prose) font.

Speaking of which, I might still owe David a follow-up PR.

sbrannen: @jddarcy and @pavelrappo, as I understood it, @liach is not suggesting that the term "void" in "primitive types and void" should be replaced by `{@code void}`.

Rather, `{@code void}` should be included in the enumeration of primitive types -- for example, by replacing `and {@code double}` with `{@code double}, and {@code void}`.

pavelrappo: > @jddarcy and @pavelrappo, as I understood it, @liach is not suggesting that the term "void" in "primitive types and void" should be replaced by `{@code void}`.
> 
> Rather, `{@code void}` should be included in the enumeration of primitive types -- for example, by replacing `and {@code double}` with `{@code double}, and {@code void}`.

AFAIK, that sentence enumerates primitive types, which according to the Java Language Specification [^1][^2], do not include `void`.

[^1]: https://docs.oracle.com/javase/specs/jls/se20/html/jls-4.html#jls-4.2
[^2]: https://docs.oracle.com/javase/specs/jls/se20/html/jls-8.html#jls-8.4.5

sbrannen: Thanks for the links, @pavelrappo.

I understand the point you're making; however, there is still ambiguous wording in the documentation for this method.

The method is called `isPrimitive`, and it returns `true` for `isPrimitive(void.class)`. In addition, the first sentence states: "Determines if the specified `Class` object represents a primitive type."

Note as well that the documentation states "There are nine predefined ..." but then goes on to only list 8 names.

So, in line with the intent of the method with regard to `void`, it should be clear to the user that `void` is the name for the type `void.class` (and `Void.TYPE`).

Perhaps the easiest way to include `void.class` and match the 9 predefined types with their names is to remove "primitive" from "as the primitive types that" and replace `and {@code double}` with `{@code double}, and {@code void}`.

Along that line of thinking, it might be best to change the first sentence to "Determines if the specified `Class` object represents a primitive type or void."

pavelrappo: > Along that line of thinking, it might be best to change the first sentence to "Determines if the specified Class object represents a primitive type or void."

I was just about to suggest exactly that, but from the superinterface method's perspective. Here's what I've been writing when received your message:

At the same, time java.lang.invoke.TypeDescriptor.OfField#isPrimitive:

        /**
         * Does this field descriptor describe a primitive type (including void.)
         *
         * @return whether this field descriptor describes a primitive type
         */
        boolean isPrimitive();

I suggest rephrasing that and similar text elsewhere; for example:

    /** {@return whether this field descriptor describes a primitive type or void} */
    boolean isPrimitive();

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -808,18 +808,26 @@ public Void run() {
...
      * {@code char}, {@code short}, {@code int},
      * {@code long}, {@code float}, and {@code double}.
      *
-     * <p> These objects may only be accessed via the following public static
-     * final variables, and are the only {@code Class} objects for which
-     * this method returns {@code true}.
+     * <p>No other class objects are considered primitive.
+     *
+     * @apiNote
+     * A {@code Class} object represented a primitive type can be
```
pavelrappo, line 823: Not a native speaker, but "representED" not "representING"?

pavelrappo (approved):
Looks good to me.

pavelrappo:
Only after I have approved the change, I re-read the whole thing again. Would you be able to fix this in this PR too?
```
    /**
     * Determines if the specified {@code Class} object represents a
     * primitive type.
...
```
There's no "specified class object",  it's **this** class object. That method has zero parameters.

sbrannen:
Thanks for improving the documentation, @jddarcy. 👍

PR #14576, opened 2023-06-21
8308340: C2: Idealize Fma nodes

fg1417:
Some platforms, like aarch64, ppc, and riscv, support fusing `Math.fma(-a, b, c)` or `Math.fma(a, -b, c)` by generating partially symmetric match rules like:

```
  match(Set dst (FmaF src3 (Binary (NegF src1) src2)));
  match(Set dst (FmaF src3 (Binary src1 (NegF src2))));
```

Since `Fma` is partially commutative, the patch is to convert `Math.fma(-a, b, c)` to `Math.fma(b, -a, c)` in gvn phase, making node patterns canonical. Then we can remove redundant rules.

Also, we should guarantee that C2 generates `Fma` nodes only on platforms supporting `Fma` instructions before matcher, so we can remove all `predicate(UseFMA)` for all `Fma` rules.

After the patch, the code size of libjvm.so on aarch64 platform decreased by 63.4k.

The patch passed all tier 1 - 3 on aarch64 and x86 platforms.

---------
### Progress

### Issue
 * [JDK-8308340](https://bugs.openjdk.org/browse/JDK-8308340): C2: Idealize Fma nodes (**Enhancement** - P4)

### Reviewers
 * [Vladimir Kozlov](https://openjdk.org/census#kvn) (@vnkozlov - **Reviewer**) ⚠️ Review applies to [06162d88](https://git.openjdk.org/jdk/pull/14576/files/06162d882df009d5227d70521b27c03d9a4bc6b8)
 * [Emanuel Peter](https://openjdk.org/census#epeter) (@eme64 - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14576/head:pull/14576` \
`$ git checkout pull/14576`

`$ git checkout pull/14576` \
`$ git pull https://git.openjdk.org/jdk.git pull/14576/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14576`

`$ git pr show -t 14576`

</details>

<a href="https://git.openjdk.org/jdk/pull/14576.diff">https://git.openjdk.org/jdk/pull/14576.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14576#issuecomment-1600022791)


fg1417:
> Check for `UseFMA` should be moved from `c2compiler.cpp` to `Matcher::match_rule_supported` in `.ad` files. I see we have such check for Fma vectors in `x86.ad` but not for scalars. Similar issue exist for other platforms.

@vnkozlov Thanks for your review! I updated it in the latest commit.

vnkozlov (approved):
Looks good to me.
You need second review.

fg1417:
Thanks for your review @vnkozlov .

I would appreciate it very much if some expert on ppc or riscv could help review it! Perhaps  @RealFYang @reinrich

fg1417:
Can I get a second review? Thanks.

src/hotspot/share/opto/vectornode.cpp
```diff
@@ -1875,6 +1875,17 @@ Node* VectorLongToMaskNode::Ideal(PhaseGVN* phase, bool can_reshape) {
   return nullptr;
 }
 
+Node* FmaVNode::Ideal(PhaseGVN* phase, bool can_reshape) {
+  // We canonicalize the node by converting "(-a)*b+c" into "b*(-a)+c"
```
reinrich, line 1879: Could you explain a little bit more please?

reinrich (commented):
I've added the patch to our nightly testing about 10 days ago and forgot about it (sorry).
So it passed several iterations of tier1-4 of hotspot and jdk, all of langtools and jaxp, renaissance benchmarks as functional tests. All testing with fastdebug and release builds on the main platforms and also on Linux/PPC64le.

PPC changes do look good to me. I'm not the greatest C2 expert though. So I'd suggest to get another review.

src/hotspot/share/opto/vectornode.cpp
fg1417: Thanks for your review!

For vectorapi masked operations, like `av.neg().lanewise(VectorOperators.FMA, bv, cv, mask)`, the inactive lanes of the output should save the first input of the node, so the inactive lanes of the output should be equal to lane values in `av.neg()`. If we exchange the inputs, the inactive lanes will be equal to `bv`, which is incorrect. So we shouldn't swap edges for masked nodes. The newly added testcases in ` jdk/test/hotspot/jtreg/compiler/vectorapi/VectorFusedMultiplyAddSubTest.java` can cover this. Fortunately, there is no such constraint for non-masked vector nodes.

fg1417:
> I've added the patch to our nightly testing about 10 days ago and forgot about it (sorry). So it passed several iterations of tier1-4 of hotspot and jdk, all of langtools and jaxp, renaissance benchmarks as functional tests. All testing with fastdebug and release builds on the main platforms and also on Linux/PPC64le.
> 
> PPC changes do look good to me. I'm not the greatest C2 expert though. So I'd suggest to get another review.

Thanks a lot for your review and test work @reinrich!

src/hotspot/share/opto/vectornode.cpp
reinrich: Thanks for the explanation. I think I understood it to some degree.
What happens with the subgraphs that are not canonicalized? They will have extra vector operations, right?

fg1417: Yes. For `av.neg().lanewise(VectorOperators.FMA, bv, cv, mask)`, the subgraph is like:
`match (Set dst (FmaV (Binary (NegV src1) src2) (Binary src3 pg)));`, almost all platforms don't support fuse it directly, so it should be split into two vector operations: `NegV` + `FmaV`. I suppose the `NegV` is what you called as "the extra vector operation", right?

reinrich: Yes that's what I meant. Thanks. Now on PPC, my understanding would be that with the symmetrical match-rules (removed with this pr) the `NegV` wouldn't be generated. Is my understanding correct?

fg1417: Hi @reinrich, I'm sorry that I didn't explain the transformation here clearly enough to lead to your misunderstanding.  Let's revisit the comment here.

The example `av.neg().lanewise(VectorOperators.FMA, bv, cv, mask)` I gave here was used to explain the latter part
"// except vectorapi masked nodes, since the inactive lanes should
// save the first input of the masked node."

It means the pr has no real impact on subgraph or codegen for `FmaV` nodes **with mask**, certainly including `PPC`. Since it doesn't apply the change to vector nodes with mask (`is_predicated_vector()`), and the pr doesn't remove any rules **with mask**. 

> Now on PPC, my understanding would be that with the symmetrical match-rules (removed with this pr) the NegV wouldn't be generated. Is my understanding correct?

The symmetrical match-rules removed with this pr works only for `FmaV` nodes **without mask**. The reason why we can remove them is that here we apply the transformation to these non-masked vector nodes. For example, on `PPC` backend, we removed ` match(Set dst (FmaVF dst (Binary (NegVF src1) src2)));` and kept `match(Set dst (FmaVF dst (Binary src1 (NegVF src2))));`, because all `(-a)*b+c` should be converted into `b*(-a)+c` here. Therefore, even without these symmetrical match-rules, whether `(-a)*b+c` or `a*(-b)+c` can be fused and the `NegV` won't be generated, given that the backend supports it.

Thanks.

reinrich: Thanks. How are `FmaV` nodes with mask handled then? Are they transformed into equivalent nodes without mask?

fg1417: Actually, there is no handling on `FmaV` nodes **with mask** in this patch, whether in the C2 mid-end or codegen backend. The gvn transformation just skips them. And I suppose `FmaV` nodes with mask can't be transformed into nodes **without mask**, except that C2 can guarantee that the mask is all true (this transformation has not been supported by current C2). Thanks.

RealFYang:
> Thanks for your review @vnkozlov .
> 
> I would appreciate it very much if some expert on ppc or riscv could help review it! Perhaps @RealFYang @reinrich

Hello, the RISC-V part looks fine. And this has passed tier1-3 tests on linux-riscv64 platform. I didn't check the shared code changes.

fg1417:
> Hello, the RISC-V part looks fine from what this PR is supposed to do. And this has passed tier1-3 tests on linux-riscv64 platform. I didn't check the shared code changes.

@RealFYang Thanks for your review and test work!

src/hotspot/share/opto/vectornode.cpp
reinrich: I guess I would have to dig deeper into the vector api implementation to really understand how it works.
Thanks for your patience.

eme64: @fg1417 I only understood the comment with the help of your explanations in this thread. I think you should improve the comment. I would not mention the vectorapi. We may generate `FmaV` through an auto-vectorizer. Though I guess that is unlikely, since the scalar version `Fma::Ideal` would already reshape things.

Suggestion:
```
// We canonicalize the node by converting "(-a)*b+c" into "b*(-a)+c"
// This reduces the number of rules in the matcher, as we only need to check
// for negations on the second argument, and not the symmetric case where
// the first argument is negated.
// We cannot do this if he FmaV is masked. the inactive lanes have to return
// the first input (ie "-a"). If we were to swap the inputs, the inactive lanes would
// incorrectly return "b".
```

src/hotspot/share/opto/mulnode.cpp
```diff
@@ -1711,6 +1711,17 @@ const Type* URShiftLNode::Value(PhaseGVN* phase) const {
   return TypeLong::LONG;                // Give up
 }
 
+//=============================================================================
+//------------------------------Ideal------------------------------------------
+Node* FmaNode::Ideal(PhaseGVN* phase, bool can_reshape) {
+  // We canonicalize the node by converting "(-a)*b+c" into "b*(-a)+c"
```
eme64, line 1717: Add motivation to comment

```
// This reduces the number of rules in the matcher, as we only need to check
// for negations on the second argument, and not the symmetric case where
// the first argument is negated.
```

src/hotspot/cpu/x86/x86.ad
```diff
@@ -3972,7 +3973,6 @@ instruct fmaD_reg(regD a, regD b, regD c) %{
 
 // a * b + c
 instruct fmaF_reg(regF a, regF b, regF c) %{
-  predicate(UseFMA);
```
eme64, line 3975: You could add an assert to the encoding code. Just to ensure that we do not generate bad code, even if it is never executed during testing.

eme64 (commented):
@fg1417 This looks like a reasonable refactoring.

We should probably do the verification that the canonicalization happened, if the normal `fma` matcher rule is chosen. We should add asserts that the first argument is not a negation (you could check the second argument also, just in case). What do you think?

test/hotspot/jtreg/compiler/vectorapi/VectorFusedMultiplyAddSubTest.java
```diff
@@ -60,7 +60,7 @@ public class VectorFusedMultiplyAddSubTest {
     private static final VectorSpecies<Long> L_SPECIES = LongVector.SPECIES_MAX;
     private static final VectorSpecies<Short> S_SPECIES = ShortVector.SPECIES_MAX;
 
-    private static int LENGTH = 1024;
+    private static int LENGTH = 128;
```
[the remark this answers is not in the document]
fg1417: Yes, it's for speeding up.

src/hotspot/share/opto/vectornode.cpp
fg1417: Thanks! Done.

src/hotspot/cpu/x86/x86.ad
fg1417: Yes, thanks for your suggestion! Updated in the new commit.

fg1417:
> We should probably do the verification that the canonicalization happened, if the normal `fma` matcher rule is chosen. We should add asserts that the first argument is not a negation (you could check the second argument also, just in case). What do you think?

Hi @eme64, thanks for your review!

The check may be more complex than expected. Matcher can fuse two instructions into one, only when there is no other use for the inputs. It means that we can do the fusion for the case like:
```
return Math.fma(-a, b, c);
```
But we can't fuse them for the case like:
```
float tmp = -a;
return Math.fma(tmp, b, c) + (-a);
```
For the second case, we still match normal `neg` + `fma` rules separately, instead of these combined rules. So, we can't simply guarantee that the first argument is not a negation when the normal `fma` matcher rule is chosen. If considering the def-use while doing the instruction selection, the check may be complex. WDYT? Thanks.

eme64 (approved):
@fg1417 fair enough. the checks may be too complex. I'll approve it now :)

fg1417:
Thanks a lot for all your kind reviews and test work, @RealFYang @vnkozlov @eme64 @reinrich.

I'll integrate it.

PR #11840, opened 2023-01-04
8299576: Reimplement java.io.Bits using VarHandle access

minborg:
Currently, `java.io.Bits` is using explicit logic to read/write various primitive types to/from byte arrays. Switching to the use of `VarHandle` access would provide better performance and less code. 

Also, using a standard API for these conversions means future `VarHandle` improvements will benefit `Bits` too. 

Improvements in `Bits` will propagate to `ObjectInputStream`, `ObjectOutputStream` and `RandomAccessFile`.

Initial benchmarks and performance discussions can be found here: https://github.com/openjdk/panama-foreign/pull/762

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8299576](https://bugs.openjdk.org/browse/JDK-8299576): Reimplement java.io.Bits using VarHandle access

### Reviewers
 * [Uwe Schindler](https://openjdk.org/census#uschindler) (@uschindler - Author) ⚠️ Review applies to [ec2e7c99](https://git.openjdk.org/jdk/pull/11840/files/ec2e7c99181453c4c8086b17594b765005403683)
 * [Alan Bateman](https://openjdk.org/census#alanb) (@AlanBateman - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk pull/11840/head:pull/11840` \
`$ git checkout pull/11840`

Update a local copy of the PR: \
`$ git checkout pull/11840` \
`$ git pull https://git.openjdk.org/jdk pull/11840/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 11840`

View PR using the GUI difftool: \
`$ git pr show -t 11840`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jdk/pull/11840.diff">https://git.openjdk.org/jdk/pull/11840.diff</a>

</details>

PR #14578, opened 2023-06-21
JDK-8310502 : Optimization for j.l.Long.fastUUID()

wenshao:
By optimizing the implementation of java.lang.Long#fastUUID, the performance of the java.util.UUID#toString method can be significantly improved.

The following are the test results of JMH: 
```
Benchmark                     Mode  Cnt      Score      Error   Units
UUIDUtilsBenchmark.new       thrpt    5  92676.550 ±  292.213  ops/ms
UUIDUtilsBenchmark.original  thrpt    5  37040.165 ± 1023.532  ops/ms
```

---------
### Progress

### Issue
 * [JDK-8310502](https://bugs.openjdk.org/browse/JDK-8310502): Optimization for j.l.Long.fastUUID() (**Enhancement** - P4)

### Reviewers
 * [Chen Liang](https://openjdk.org/census#liach) (@liach - Author) ⚠️ Review applies to [93e74b6c](https://git.openjdk.org/jdk/pull/14578/files/93e74b6cac32156f03896ca5d40d432473e2f34a)
 * [Alan Bateman](https://openjdk.org/census#alanb) (@AlanBateman - **Reviewer**) ⚠️ Review applies to [93e74b6c](https://git.openjdk.org/jdk/pull/14578/files/93e74b6cac32156f03896ca5d40d432473e2f34a)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14578/head:pull/14578` \
`$ git checkout pull/14578`

`$ git checkout pull/14578` \
`$ git pull https://git.openjdk.org/jdk.git pull/14578/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14578`

`$ git pr show -t 14578`

</details>

<a href="https://git.openjdk.org/jdk/pull/14578.diff">https://git.openjdk.org/jdk/pull/14578.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14578#issuecomment-1600333873)


src/java.base/share/classes/java/lang/Long.java
```diff
@@ -85,6 +85,17 @@ public final class Long extends Number
...
+    static final char[] HEX256;
+    static {
+        HEX256 = new char[256];
+        for (int i = 0; i < 256; i++) {
+            int hi = (i >> 4) & 15;
+            int lo = i & 15;
+            HEX256[i] = (char) (((hi < 10 ? '0' + hi : 'a' + hi - 10) << 8)
+                    + (lo < 10 ? '0' + lo : 'a' + lo - 10));
+        }
+    }
```
AlanBateman, line 97: Are you checking the impact on startup? I guess I would have expected to see it loaded from the CDS archive. Given that the usage is just UUID::toString then maybe you could move this to a holder class containing a Stable array, at least until there is a better way to have lazily computed constants. Another thing to try is moving fastUUID out of Long - moving to an array of precomputed hex values means it is not tied to Long internals anymore.

wenshao: 1. HEX256 can be used to optimize all integers to hex strings, such as Integer#toHexString and Long#toHexString and byte [] to hexstring.

2. i am a newcomer, and i don't understand what 'loaded from the CDS archive' means, do you mean to use constants instead? such as
```java
HEX256 = new char[]{
        12336, 12337, 12338, 12339, 12340, 12341, 12342, 12343, 12344, 12345,
        12385, 12386, 12387, 12388, 12389, 12390, 12592, 12593, 12594, 12595,
        12596, 12597, 12598, 12599, 12600, 12601, 12641, 12642, 12643, 12644,
        12645, 12646, 12848, 12849, 12850, 12851, 12852, 12853, 12854, 12855,
        12856, 12857, 12897, 12898, 12899, 12900, 12901, 12902, 13104, 13105,
        13106, 13107, 13108, 13109, 13110, 13111, 13112, 13113, 13153, 13154,
        13155, 13156, 13157, 13158, 13360, 13361, 13362, 13363, 13364, 13365,
        13366, 13367, 13368, 13369, 13409, 13410, 13411, 13412, 13413, 13414,
        13616, 13617, 13618, 13619, 13620, 13621, 13622, 13623, 13624, 13625,
        13665, 13666, 13667, 13668, 13669, 13670, 13872, 13873, 13874, 13875,
        13876, 13877, 13878, 13879, 13880, 13881, 13921, 13922, 13923, 13924,
        13925, 13926, 14128, 14129, 14130, 14131, 14132, 14133, 14134, 14135,
        14136, 14137, 14177, 14178, 14179, 14180, 14181, 14182, 14384, 14385,
        14386, 14387, 14388, 14389, 14390, 14391, 14392, 14393, 14433, 14434,
        14435, 14436, 14437, 14438, 14640, 14641, 14642, 14643, 14644, 14645,
        14646, 14647, 14648, 14649, 14689, 14690, 14691, 14692, 14693, 14694,
        24880, 24881, 24882, 24883, 24884, 24885, 24886, 24887, 24888, 24889,
        24929, 24930, 24931, 24932, 24933, 24934, 25136, 25137, 25138, 25139,
        25140, 25141, 25142, 25143, 25144, 25145, 25185, 25186, 25187, 25188,
        25189, 25190, 25392, 25393, 25394, 25395, 25396, 25397, 25398, 25399,
        25400, 25401, 25441, 25442, 25443, 25444, 25445, 25446, 25648, 25649,
        25650, 25651, 25652, 25653, 25654, 25655, 25656, 25657, 25697, 25698,
        25699, 25700, 25701, 25702, 25904, 25905, 25906, 25907, 25908, 25909,
        25910, 25911, 25912, 25913, 25953, 25954, 25955, 25956, 25957, 25958,
        26160, 26161, 26162, 26163, 26164, 26165, 26166, 26167, 26168, 26169,
        26209, 26210, 26211, 26212, 26213, 26214
};
```

liach: > Another thing to try is moving fastUUID out of Long - moving to an array of precomputed hex values means it is not tied to Long internals anymore.

A note about `@Stable`: `Integer.digits` and `HEX256` are not, and they might see performance improvements and change the benchmark results if they are declared so.

wenshao: > > Another thing to try is moving fastUUID out of Long - moving to an array of precomputed hex values means it is not tied to Long internals anymore.
> 
> A note about `@Stable`: `Integer.digits` and `HEX256` are not, and they might see performance improvements and change the benchmark results if they are declared so.

use HEX256 optimize Integer.toHexString
https://github.com/wenshao/jdk_8310502_test/blob/main/src/main/java/com/alibaba/openjdk/HexUtils.java
```java
       char[] hex256 = UUIDUtils.HEX256;

        int i0 = (i >> 24) & 255;
        int i1 = (i >> 16) & 255;
        int i2 = (i >> 8) & 255;
        int i3 = i & 255;

        char c0 = hex256[i0];
        char c1 = hex256[i1];
        char c2 = hex256[i2];
        char c3 = hex256[i3];

        byte[] bytes;
        if (COMPACT_STRINGS) {
            if ((i >> 4) == 0) {
                bytes = new byte[1];
                bytes[0] = (byte) c3;
            } else if ((i >> 8) == 0) {
                bytes = new byte[2];
                bytes[0] = (byte) (c3 >> 8);
                bytes[1] = (byte) c3;
            } else if ((i >> 12) == 0) {
                bytes = new byte[3];
                bytes[0] = (byte) c2;
                bytes[1] = (byte) (c3 >> 8);
                bytes[2] = (byte) c3;
            } else if ((i >> 16) == 0) {
                bytes = new byte[4];
                bytes[0] = (byte) (c2 >> 8);
                bytes[1] = (byte) c2;
                bytes[2] = (byte) (c3 >> 8);
                bytes[3] = (byte) c3;
            } else if ((i >> 20) == 0) {
                bytes = new byte[5];
                bytes[0] = (byte) c1;
                bytes[1] = (byte) (c2 >> 8);
                bytes[2] = (byte) c2;
                bytes[3] = (byte) (c3 >> 8);
                bytes[4] = (byte) c3;
            } else if ((i >> 24) == 0) {
                bytes = new byte[6];
                bytes[0] = (byte) (c1 >> 8);
                bytes[1] = (byte) c1;
                bytes[2] = (byte) (c2 >> 8);
                bytes[3] = (byte) c2;
                bytes[4] = (byte) (c3 >> 8);
                bytes[5] = (byte) c3;
            } else if ((i >> 28) == 0) {
                bytes = new byte[7];
                bytes[0] = (byte) c0;
                bytes[1] = (byte) (c1 >> 8);
                bytes[2] = (byte) c1;
                bytes[3] = (byte) (c2 >> 8);
                bytes[4] = (byte) c2;
                bytes[5] = (byte) (c3 >> 8);
                bytes[6] = (byte) c3;
            } else {
                bytes = new byte[8];
                bytes[0] = (byte) (c0 >> 8);
                bytes[1] = (byte) c0;
                bytes[2] = (byte) (c1 >> 8);
                bytes[3] = (byte) c1;
                bytes[4] = (byte) (c2 >> 8);
                bytes[5] = (byte) c2;
                bytes[6] = (byte) (c3 >> 8);
                bytes[7] = (byte) c3;
            }

            return new Stringbytes, LATIN1);
      }
      // ....
```

* benchmark code
https://github.com/wenshao/jdk_8310502_test/blob/main/src/main/java/com/alibaba/openjdk/HexUtilsBenchmark.java

* benchmark result
```
Benchmark                Mode  Cnt     Score    Error   Units
HexUtilsBenchmark.fast  thrpt    5  2910.332 ± 20.051  ops/ms
HexUtilsBenchmark.jdk   thrpt    5  1966.712 ± 48.141  ops/ms
```

src/java.base/share/classes/java/lang/Long.java
```diff
@@ -85,6 +85,47 @@ public final class Long extends Number
...
+        25140, 25141, 25142, 25143, 25144, 25145, 25185, 25186, 25187, 25188,
+        25189, 25190, 25392, 25393, 25394, 25395, 25396, 25397, 25398, 25399,
+        25400, 25401, 25441, 25442, 25443, 25444, 25445, 25446, 25648, 25649,
+        25650, 25651, 25652, 25653, 25654, 25655, 25656, 25657, 25697, 25698,
+        25699, 25700, 25701, 25702, 25904, 25905, 25906, 25907, 25908, 25909,
+        25910, 25911, 25912, 25913, 25953, 25954, 25955, 25956, 25957, 25958,
+        26160, 26161, 26162, 26163, 26164, 26165, 26166, 26167, 26168, 26169,
+        26209, 26210, 26211, 26212, 26213, 26214
+    };
+
```
rgiulietti, line 128: Can we perhaps have hex literals here? I think they are easier to visually check than decimals.
```
static final char[] HEX256 = new char[]{
        0x3030, 0x3031, 0x3032, 0x3033, 0x3034, 0x3035, 0x3036, 0x3037, ...
```
or
```
static final char[] HEX256 = new char[]{
        0x30_30, 0x30_31, 0x30_32, 0x30_33, 0x30_34, 0x30_35, 0x30_36, 0x30_37, ...
```

src/java.base/share/classes/java/lang/Long.java
```diff
@@ -446,33 +487,105 @@ private static void formatUnsignedLong0UTF16(long val, int shift, byte[] buf, in
     }
 
     static String fastUUID(long lsb, long msb) {
+        final char[] hex256 = HEX256;
+
+        char i = hex256[((int) (msb >> 56)) & 255];
```
rgiulietti, line 492: Masks are easier to read when expressed as hex literals.
```
char i = hex256[((int) (msb >> 56)) & 0xFF];
...
```

wenshao:
* benchmark code
https://github.com/wenshao/jdk_8310502_test/blob/main/src/main/java/com/alibaba/openjdk/UUIDUtilsBenchmark.java

* benchmark result
```
Benchmark                 Mode  Cnt     Score    Error   Units
HexUtilsBenchmark.fast   thrpt    5  2971.864 ± 17.277  ops/ms
HexUtilsBenchmark.jdk    thrpt    5  1984.653 ± 11.733  ops/ms
```

src/java.base/share/classes/java/lang/Long.java
minborg: Not sure if this can be applied but some months ago, I optimized Bits to use VarHandles rather than explicitly shifting bits around. This gave us a significant performance increase:

https://github.com/openjdk/jdk/pull/11840/files

src/java.base/share/classes/java/lang/Long.java
```diff
@@ -446,33 +488,105 @@ private static void formatUnsignedLong0UTF16(long val, int shift, byte[] buf, in
...
+            StringUTF16.putChar(buf, 0, (byte) (i0 >> 8));
+            StringUTF16.putChar(buf, 1, (byte) i0);
+            StringUTF16.putChar(buf, 2, (byte) (i1 >> 8));
+            StringUTF16.putChar(buf, 3, (byte) i1);
+            StringUTF16.putChar(buf, 4, (byte) (i2 >> 8));
+            StringUTF16.putChar(buf, 5, (byte) i2);
+            StringUTF16.putChar(buf, 6, (byte) (i3 >> 8));
+            StringUTF16.putChar(buf, 7, (byte) i3);
+            StringUTF16.putChar(buf, 8, '-');
+            StringUTF16.putChar(buf, 9, (byte) (i4 >> 8));
```
RogerRiggs, line 563: This might be cheating but you could avoid a store of 0 to the high byte (its already zero) by inlining the code from StringUTF16.putChar();  just double the index on the store of the byte.
```
            buf[0] = (byte) (i0 >> 8);
            buf[2] = (byte) i0;
            buf[4] = (byte) (i1 >> 8);
            buf[6] = (byte) i1;
            buf[8] = (byte) (i2 >> 8);
...
```

src/java.base/share/classes/java/lang/Long.java
RogerRiggs: CDS can initialize arrays efficiently.  The Long class already uses CDS to initialize the cache of Long values.
See the LongCache code for an example of initializing from CDS with a fallback to direct initialization.
You could move the HEX256 array to the LongCache holder class and avoid any performance hit on startup.

src/java.base/share/classes/java/lang/Long.java
ExE-Boss: Note that `StringUTF16::putChar` takes endian‑ness into account, so the above code would only be correct on half of the supported endian types.

src/java.base/share/classes/java/lang/Long.java
```diff
@@ -446,33 +488,105 @@ private static void formatUnsignedLong0UTF16(long val, int shift, byte[] buf, in
...
+            buf[26] = (byte) (i11 >> 8);
+            buf[27] = (byte) i11;
+            buf[28] = (byte) (i12 >> 8);
+            buf[29] = (byte) i12;
+            buf[30] = (byte) (i13 >> 8);
+            buf[31] = (byte) i13;
+            buf[32] = (byte) (i14 >> 8);
+            buf[33] = (byte) i14;
+            buf[34] = (byte) (i15 >> 8);
+            buf[35] = (byte) i15;
```
Glavo, line 548: You may be able to use `jdk.internal.util.ByteArray` to simplify code and improve performance:

```suggestion
            ByteArray.setChar(buf, 0, i0);
            ByteArray.setChar(buf, 2, i1);
            ByteArray.setChar(buf, 4, i2);
            ByteArray.setChar(buf, 6, i3);
            buf[8] = '-';
            ByteArray.setChar(buf, 9, i4);
            ByteArray.setChar(buf, 11, i5);
            buf[13] = '-';
            ByteArray.setChar(buf, 14, i6);
            ByteArray.setChar(buf, 16, i7);
            buf[18] = '-';
            ByteArray.setChar(buf, 19, i8);
            ByteArray.setChar(buf, 21, i9);
            buf[23] = '-';
            ByteArray.setChar(buf, 24, i10);
            ByteArray.setChar(buf, 26, i11);
            ByteArray.setChar(buf, 28, i12);
            ByteArray.setChar(buf, 30, i13);
            ByteArray.setChar(buf, 32, i14);
            ByteArray.setChar(buf, 34, i15);
```

I measured the throughput of `UUID::toString()` using JMH on my linux server (CPU: R7-5800X):

```diff
  Benchmark                 Mode  Cnt      Score     Error   Units
- UUIDUtilsBenchmark.test  thrpt    5  76117.237 ± 461.679  ops/ms
+ UUIDUtilsBenchmark.test  thrpt    5  87291.496 ± 729.527  ops/ms
```

The result of this benchmark test is that using `ByteArray` can improve peak throughput by nearly 15%.

Note: `ByteArray` uses `VarHandle` internally, and the impact on startup time is unknown, so more testing is needed.

src/java.base/share/classes/java/lang/Long.java
wenshao: i have simplified the code using ENDIAN and COMPACT_STRINGS

src/java.base/share/classes/java/lang/Long.java
wenshao: > CDS can initialize arrays efficiently. The Long class already uses CDS to initialize the cache of Long values. See the LongCache code for an example of initializing from CDS with a fallback to direct initialization. You could move the HEX256 array to the LongCache holder class and avoid any performance hit on startup.

done

wenshao:
> @wenshao Please do not rebase or force-push to an active PR as it invalidates existing review comments. Note for future reference, the bots always squash all changes into a single commit automatically as part of the integration. See [OpenJDK Developers’ Guide](https://openjdk.org/guide/#working-with-pull-requests) for more information.

got it

RogerRiggs:
fyi, the title of this PR need to match exactly the title of the bug [JDK-8310502](https://bugs.openjdk.org/browse/JDK-8310502).
The mismatch is an Integration blocker. (See the comment in the description).

wenshao:
> fyi, the title of this PR need to match exactly the title of the bug [JDK-8310502](https://bugs.openjdk.org/browse/JDK-8310502). The mismatch is an Integration blocker. (See the comment in the description).

i have made the changes. is it ok now?

RogerRiggs:
The Bots have removed the warning so the titles match.
Please wait 24hrs to integrate to give anyone who has commented a chance to review and approve.

src/java.base/share/classes/java/lang/Long.java
```diff
@@ -446,36 +447,77 @@ private static void formatUnsignedLong0UTF16(long val, int shift, byte[] buf, in
...
-
-            return new String(buf, UTF16);
+            coder = UTF16;
+            charSize = 2;
+            off = ByteOrder.nativeOrder() == ByteOrder.BIG_ENDIAN ? 1 : 0;
         }
+        final byte[] buf = new byte[36 * charSize];
+
+        buf[off] = (byte) (i >> 8);
+        buf[1 * charSize + off] = (byte) i;
```
liach, line 484:
```suggestion
        buf[off] = (byte) (i0 >> 8);
        buf[1 * charSize + off] = (byte) i0;
```
Doesn't compile on my end

liach (changes requested):

liach:
About @Glavo's VH suggestion: I think it is feasible, since the VH field is not initialized until the method is used, so there should be no startup issue. 

On a side note, JDK itself has a `UUIDBench` that benchmarks toString as well: can run it with `make test TEST="micro:java.util.UUIDBench.toString"` once the configuration has JMH set up, which I will be using (against master and this patch)

liach:
I've tested the benchmarks and the patch and baseline (with extra stable annotation) with a slightly varied version suitable for gradle run:

```java
package com.alibaba.openjdk;

import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.BenchmarkMode;
import org.openjdk.jmh.annotations.Fork;
import org.openjdk.jmh.annotations.Measurement;
import org.openjdk.jmh.annotations.Mode;
import org.openjdk.jmh.annotations.OutputTimeUnit;
import org.openjdk.jmh.annotations.Warmup;
import org.openjdk.jmh.infra.Blackhole;

import java.util.UUID;
import java.util.concurrent.TimeUnit;

@OutputTimeUnit(TimeUnit.MILLISECONDS)
@Warmup(iterations = 3, time = 10)
@Measurement(iterations = 6, time = 5)
@Fork(1)
public class UUIDUtilsBenchmark {
    public static UUID uuid = UUID.randomUUID();

    @Benchmark
    public void jdk(Blackhole bh) {
        bh.consume(uuid.toString());
    }

    @Benchmark
    public void fast(Blackhole bh) {
        bh.consume(UUIDUtils.fastUUID(uuid));
    }
}
```
The throughput varies a lot between iterations somehow; the patch and baseline with stable has no significant difference (i.e. within the error range, about 10%)

src/java.base/share/classes/java/lang/Long.java
```diff
@@ -1178,6 +1223,14 @@ private LongCache() {}
...
             }
             cache = archivedCache;
+
+            HEX256 = new char[256];
+            for (int i = 0; i < 256; i++) {
+                int hi = (i >> 4) & 15;
+                int lo = i & 15;
+                HEX256[i] = (char) (((hi < 10 ? '0' + hi : 'a' + hi - 10) << 8)
+                        + (lo < 10 ? '0' + lo : 'a' + lo - 10));
+            }
```
AlanBateman, line 1233: Did you mean to put this in LongCache and is the intention it be archived or are you putting this into its own holder class? Right now it's confusing as HEX256 is not read from the archive.

Glavo:
> About @Glavo's VH suggestion: I think it is feasible, since the VH field is not initialized until the method is used, so there should be no startup issue.

@liach This sounds good.

I'm thinking about refactoring `ByteArray` and `ByteArrayLittleEndian` to use `Unsafe` internally instead of `VarHandle` so we can use it in more places without worrying about affecting startup time.

I'm running some tests to determine the benefit of doing this.

wenshao:
> I've tested the benchmarks and the patch and baseline (with extra stable annotation) with a slightly varied version suitable for gradle run:
> 
> ```java
> package com.alibaba.openjdk;
> 
> import org.openjdk.jmh.annotations.Benchmark;
> import org.openjdk.jmh.annotations.BenchmarkMode;
> import org.openjdk.jmh.annotations.Fork;
> import org.openjdk.jmh.annotations.Measurement;
> import org.openjdk.jmh.annotations.Mode;
> import org.openjdk.jmh.annotations.OutputTimeUnit;
> import org.openjdk.jmh.annotations.Warmup;
> import org.openjdk.jmh.infra.Blackhole;
> 
> import java.util.UUID;
> import java.util.concurrent.TimeUnit;
> 
> @BenchmarkMode(Mode.Throughput)
> @OutputTimeUnit(TimeUnit.MILLISECONDS)
> @Warmup(iterations = 3, time = 10)
> @Measurement(iterations = 6, time = 5)
> @Fork(1)
> public class UUIDUtilsBenchmark {
>     public static UUID uuid = UUID.randomUUID();
> 
>     @Benchmark
>     public void jdk(Blackhole bh) {
>         bh.consume(uuid.toString());
>     }
> 
>     @Benchmark
>     public void fast(Blackhole bh) {
>         bh.consume(UUIDUtils.fastUUID(uuid));
>     }
> }
> ```
> 
> The throughput varies a lot between iterations somehow; the patch and baseline with stable has no significant difference (i.e. within the error range, about 10%)

the performance of this improvement varies among different environments, with a significant improvement on the MacBookPro M1 Max chip (my laptop) and only about 10% improvement on Linux servers (Intel Xeon(Sapphire Rapids) Platinum and Aliyun Yitian 710 arm CPU and Orange Pi 5 Plus).

src/java.base/share/classes/java/lang/Long.java
wenshao: i have already used a separate holder class "DigitCache", and modifiers do not use private because I will submit a new optimization using HEX256 for Integer # toHexString and Long # toHexString

liach:
@wenshao Can you try to just add `@Stable` on `Integer.digits` array and check its benchmark results?

src/java.base/share/classes/java/lang/Long.java
```diff
@@ -1164,6 +1206,9 @@ private LongCache() {}
         static final Long[] cache;
         static Long[] archivedCache;
 
+        @Stable
+        static final char[] HEX256;
+
```
Glavo, line 1211: Builds are broken.

```suggestion
```

liach: Also the stable annotation is missing in the copy of this field in the cache class.

Glavo:
@wenshao I rerun the benchmark test:

```
Baseline:
Benchmark   Mode  Cnt      Score     Error   Units
UUID.test  thrpt    5  84224.961 ± 556.817  ops/ms

Your PR:
Benchmark   Mode  Cnt      Score    Error   Units
UUID.test  thrpt    5  76205.978 ± 93.024  ops/ms

Using ByteArray (https://github.com/openjdk/jdk/pull/14578#discussion_r1237573456)
Benchmark   Mode  Cnt      Score     Error   Units
UUID.test  thrpt    5  87015.810 ± 329.174  ops/ms
```

Your current implementation has lower performance on my server than the baseline.

src/java.base/share/classes/java/lang/Long.java
```diff
@@ -469,123 +470,69 @@ static String fastUUID(long lsb, long msb) {
...
-            buf[32] = (byte) (i14 >> 8);
-            buf[33] = (byte) i14;
-            buf[34] = (byte) (i15 >> 8);
-            buf[35] = (byte) i15;
+            ByteArray.setChar(buf, 24, i10);
+            ByteArray.setChar(buf, 26, i11);
+            ByteArray.setChar(buf, 28, i12);
+            ByteArray.setChar(buf, 30, i13);
+            ByteArray.setChar(buf, 32, i14);
+            ByteArray.setChar(buf, 34, i15);
```
Glavo:
```suggestion
            ByteArray.setLong(buf, 0, ((long) i0 << 48) | ((long) i1 << 32) | ((long) i2 << 16) | i3);
            buf[8] = '-';
            ByteArray.setInt(buf, 9, (i4 << 16) | i5);
            buf[13] = '-';
            ByteArray.setInt(buf, 14, (i6 << 16) | i7);
            buf[18] = '-';
            ByteArray.setInt(buf, 19, (i8 << 16) | i9);
            buf[23] = '-';
            ByteArray.setLong(buf, 24, ((long) i10 << 48) | ((long) i11 << 32) | ((long) i12 << 16) | i13);
            ByteArray.setInt(buf, 32, (i14 << 16) | i15);
```

Benchmark:

```diff
  Benchmark   Mode  Cnt      Score     Error   Units
- UUID.test  thrpt    5  87196.091 ± 492.129  ops/ms
+ UUID.test  thrpt    5  99197.287 ± 925.266  ops/ms
```

src/java.base/share/classes/java/lang/Long.java
```diff
@@ -446,36 +448,93 @@ private static void formatUnsignedLong0UTF16(long val, int shift, byte[] buf, in
...
+            buf[18] = '-';
+            ByteArray.setChar(buf, 19, i8);
+            ByteArray.setChar(buf, 21, i9);
+            buf[23] = '-';
+            ByteArray.setChar(buf, 24, i10);
+            ByteArray.setChar(buf, 26, i11);
+            ByteArray.setChar(buf, 28, i12);
+            ByteArray.setChar(buf, 30, i13);
+            ByteArray.setChar(buf, 32, i14);
+            ByteArray.setChar(buf, 34, i15);
```
Glavo, line 492:
```suggestion
            ByteArray.setLong(buf, 0, ((long) i0 << 48) | ((long) i1 << 32) | ((long) i2 << 16) | i3);
            buf[8] = '-';
            ByteArray.setInt(buf, 9, (i4 << 16) | i5);
            buf[13] = '-';
            ByteArray.setInt(buf, 14, (i6 << 16) | i7);
            buf[18] = '-';
            ByteArray.setInt(buf, 19, (i8 << 16) | i9);
            buf[23] = '-';
            ByteArray.setLong(buf, 24, ((long) i10 << 48) | ((long) i11 << 32) | ((long) i12 << 16) | i13);
            ByteArray.setInt(buf, 32, (i14 << 16) | i15);
```

JMH Result:

```diff
  Benchmark   Mode  Cnt      Score     Error   Units
- UUID.test  thrpt    5  87196.091 ± 492.129  ops/ms
+ UUID.test  thrpt    5  99197.287 ± 925.266  ops/ms
```

src/java.base/share/classes/java/lang/Long.java
```diff
@@ -446,36 +448,83 @@ private static void formatUnsignedLong0UTF16(long val, int shift, byte[] buf, in
...
+        buf[56 + off] = (byte) (i12 >> 8);
+        buf[58 + off] = (byte) i12;
+        buf[60 + off] = (byte) (i13 >> 8);
+        buf[62 + off] = (byte) i13;
+        buf[64 + off] = (byte) (i14 >> 8);
+        buf[66 + off] = (byte) i14;
+        buf[68 + off] = (byte) (i15 >> 8);
+        buf[70 + off] = (byte) i15;
+
+        return new String(buf, UTF16);
```
ExE-Boss, line 527: Maybe restore the `else` block?
```suggestion
        } else {
            byte[] buf = new byte[72];
            int off = ByteOrder.nativeOrder() == ByteOrder.BIG_ENDIAN ? 1 : 0;

            buf[0 + off] = (byte) (i0 >> 8);
            buf[2 + off] = (byte) i0;
            buf[4 + off] = (byte) (i1 >> 8);
            buf[6 + off] = (byte) i1;
            buf[8 + off] = (byte) (i2 >> 8);
            buf[10 + off] = (byte) i2;
            buf[12 + off] = (byte) (i3 >> 8);
            buf[14 + off] = (byte) i3;
            buf[16 + off] = '-';
            buf[18 + off] = (byte) (i4 >> 8);
            buf[20 + off] = (byte) i4;
            buf[22 + off] = (byte) (i5 >> 8);
            buf[24 + off] = (byte) i5;
            buf[26 + off] = '-';
            buf[28 + off] = (byte) (i6 >> 8);
            buf[30 + off] = (byte) i6;
            buf[32 + off] = (byte) (i7 >> 8);
            buf[34 + off] = (byte) i7;
            buf[36 + off] = '-';
            buf[38 + off] = (byte) (i8 >> 8);
            buf[40 + off] = (byte) i8;
            buf[42 + off] = (byte) (i9 >> 8);
            buf[44 + off] = (byte) i9;
            buf[46 + off] = '-';
            buf[48 + off] = (byte) (i10 >> 8);
            buf[50 + off] = (byte) i10;
            buf[52 + off] = (byte) (i11 >> 8);
            buf[54 + off] = (byte) i11;
            buf[56 + off] = (byte) (i12 >> 8);
            buf[58 + off] = (byte) i12;
            buf[60 + off] = (byte) (i13 >> 8);
            buf[62 + off] = (byte) i13;
            buf[64 + off] = (byte) (i14 >> 8);
            buf[66 + off] = (byte) i14;
            buf[68 + off] = (byte) (i15 >> 8);
            buf[70 + off] = (byte) i15;

            return new String(buf, UTF16);
        }
```

liach: We should avoid extra indentation when possible.

RogerRiggs:
@wenshao Integration is only possible after Reviewers have approved; so far there it has not been "Approved".
It may be useful to re-read sections on developing a PR of the [OpenJDK developers guide](https://openjdk.org/guide/#life-of-a-pr).

wenshao:
Latest benchmark results:

* aliyun_c8i.xlarge (+23.25%)
cpu intel xeon sapphire rapids (x64)
``` diff
 Benchmark           (size)   Mode  Cnt   Score   Error   Units
-UUIDBench.toString   20000  thrpt   15  62.457 ± 0.226  ops/us
+UUIDBench.toString   20000  thrpt   15  76.979 ± 1.011  ops/us
```

* aliyun_c8a.xlarge (+29.93%)
cpu amd epc genoa (x64)
``` diff
 Benchmark           (size)   Mode  Cnt   Score   Error   Units
-UUIDBench.toString   20000  thrpt   15  65.673 ± 0.110  ops/us
+UUIDBench.toString   20000  thrpt   15  85.331 ± 0.181  ops/us
```

* aliyun_c8y.xlarge (+29.06%)
cpu aliyun yitian 710 (aarch64)
``` diff
 Benchmark           (size)   Mode  Cnt   Score   Error   Units
-UUIDBench.toString   20000  thrpt   15  32.198 ± 0.907  ops/us
+UUIDBench.toString   20000  thrpt   15  41.555 ± 0.963  ops/us
```

* aws_c5.xlarge (+35.29%)
``` diff
 Benchmark           (size)   Mode  Cnt   Score   Error   Units
-UUIDBench.toString   20000  thrpt   15  46.003 ± 0.291  ops/us
+UUIDBench.toString   20000  thrpt   15  62.239 ± 0.072  ops/us
```

wenshao:
Latest benchmark results:

###  aliyun_c8i.xlarge (+23.25% )
cpu : intel xeon sapphire rapids (x64)
``` diff
 Benchmark           (size)   Mode  Cnt   Score   Error   Units
-UUIDBench.toString   20000  thrpt   15  62.457 ± 0.226  ops/us
+UUIDBench.toString   20000  thrpt   15  76.979 ± 1.011  ops/us
```

### aliyun_c8a.xlarge (+23.09%)
cpu : amd epc genoa (x64)
``` diff
 Benchmark           (size)   Mode  Cnt   Score   Error   Units
-UUIDBench.toString   20000  thrpt   15  65.673 ± 0.110  ops/us
+UUIDBench.toString   20000  thrpt   15  80.839 ± 0.278  ops/us
```

### aliyun_c8y.xlarge (+71.92%)
cpu : aliyun yitian 710 (aarch64)
``` diff
 Benchmark           (size)   Mode  Cnt   Score   Error   Units
-UUIDBench.toString   20000  thrpt   15  32.198 ± 0.907  ops/us
+UUIDBench.toString   20000  thrpt   15  55.356 ± 1.882  ops/us
```

### aws_c5.xlarge (+36.23% )
aws x64
``` diff
 Benchmark           (size)   Mode  Cnt   Score   Error   Units
-UUIDBench.toString   20000  thrpt   15  46.003 ± 0.291  ops/us
+UUIDBench.toString   20000  thrpt   15  62.673 ± 0.259  ops/us
```

### MacBook Pro M1 Pro (+179.94%)
``` diff
 Benchmark           (size)   Mode  Cnt   Score   Error   Units
-UUIDBench.toString   20000  thrpt   15  36.875 ± 0.902  ops/us
+UUIDBench.toString   20000  thrpt   15  103.229 ± 1.063  ops/us
```

### Orange Pi 5 Plus (+71.16%)
cpu : RK3588 (aarch64)
``` diff
 Benchmark           (size)   Mode  Cnt   Score   Error   Units
-UUIDBench.toString   20000  thrpt   15  21.691 ± 0.204  ops/us
+UUIDBench.toString   20000  thrpt   15  38.211 ± 0.344  ops/us
```

src/java.base/share/classes/java/lang/Long.java
```diff
@@ -476,23 +475,20 @@ static String fastUUID(long lsb, long msb) {
...
+                        | H256[(((int) lsb) >> 16) & 0xff]);
         ByteArray.setInt(
                 buf,
                 32,
                 (H256[(((int) lsb) >> 8) & 0xff] << 16)
-                        | H256[((int) lsb) & 0xff]
-        );
+                        | H256[((int) lsb) & 0xff]);
 
         return new String(buf, LATIN1);
```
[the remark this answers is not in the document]
wenshao: where COMPACT_STRINGS is false,  jdk.internal.access.JavaLangAccess#fastUUID will use fastUUIDUTF16

```
 public String fastUUID(long lsb, long msb) {
         return String.COMPACT_STRINGS ? Long.fastUUID(lsb, msb) : Long.fastUUIDUTF16(lsb, msb);
}
```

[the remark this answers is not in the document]
ExE-Boss: There should probably be a check or something at the top of this method that `COMPACT_STRINGS` is `true` then, e.g.:
```java
if (!COMPACT_STRINGS) {
	throw new InternalError("...");
}
```

AlanBateman:
@wenshao The patch has changed 19 times in the last 3 days, is it ready for review now?

wenshao:
> @wenshao The patch has changed 19 times in the last 3 days, is it ready for review now?

@AlanBateman 
it's ready for review,. thank you.

wenshao:
@AlanBateman so sorry, this is the first time i submit code to openjdk, i am not familiar with compiling and testing jdk, so i have modified it many times. i've noticed the problem and finally i'd a full benchmark. i am waiting for your review.

wenshao:
i found a class with the same functionality called 'java.util.HexDigits', but is non-public. 

should i rename it to 'jdk.util.HexDigits' and make it public so that multiple locations can be shared it?

wenshao:
@AlanBateman 

i found an implementation with minor changes, but slower when COMPACT_STRINGS is false.

https://github.com/wenshao/jdk/commit/4a3da4b20fe0f7244c2c7bde3d9a0e2c8b91455a

do you suggest using a new implementation? if yes, i commit new changes.

liach:
@wenshao I don't think this is feasible, for this uses a new feature and hampers backport to jdk11. I suggest adding such a cache array in jdk.internal packages that digits class can use as well

wenshao:
@AlanBateman i have modified according to liach's suggestion, it's ready for review also.

AlanBateman:
> @wenshao I don't think this is feasible, for this uses a new feature and hampers backport to jdk11.

I don't think backporting should be in the discussion here. The important thing is that we converge on agreement for a maintainable patch for the main line.

src/java.base/share/classes/java/lang/Long.java
```diff
@@ -446,36 +449,111 @@ private static void formatUnsignedLong0UTF16(long val, int shift, byte[] buf, in
     }
 
     static String fastUUID(long lsb, long msb) {
```
ExE-Boss, line 451: This method should probably have an assert or something that `COMPACT_STRINGS` is `true`, e.g.:
```suggestion
    static String fastUUID(long lsb, long msb) {
        if (!COMPACT_STRINGS) {
            throw new InternalError("Long::fastUUID requires compact strings");
        }

```

wenshao: it's none-public method, only for UUID.toString

Glavo:
> I don't think this is feasible, for this uses a new feature and hampers backport to jdk11.

I don't think affecting backports in general should be a reason to discourage new features from being used.

https://github.com/wenshao/jdk/commit/4a3da4b20fe0f7244c2c7bde3d9a0e2c8b91455a is great for me.

Implementing some functions of UUID in `Long` is inherently illogical. Now that we can implement it efficiently in the `UUID` class, there's no reason why we shouldn't.

src/java.base/share/classes/java/lang/Long.java
wenshao: now the fastUUID method codeSize is 325, which is equal to the maximum value of FreqInlineSize under aarch64 and x64. if additional assertion is added, the method codeSize > 325, cannot be inlined, resulting in performance degradation.

wenshao:
i agree that backport is not the goal, but in the two packages java.util and java.lang share a cache array, it is more appropriate to put it in jdk.util.

wenshao:
> > I don't think this is feasible, for this uses a new feature and hampers backport to jdk11.
> 
> I don't think affecting backports in general should be a reason to discourage new features from being used.
> 
> [wenshao@4a3da4b](https://github.com/wenshao/jdk/commit/4a3da4b20fe0f7244c2c7bde3d9a0e2c8b91455a) is great for me.
> 
> Implementing some functions of UUID in `Long` is inherently illogical. Now that we can implement it efficiently in the `UUID` class, there's no reason why we shouldn't.

wenshao:
> 



> > I don't think this is feasible, for this uses a new feature and hampers backport to jdk11.
> 
> I don't think affecting backports in general should be a reason to discourage new features from being used.
> 
> [wenshao@4a3da4b](https://github.com/wenshao/jdk/commit/4a3da4b20fe0f7244c2c7bde3d9a0e2c8b91455a) is great for me.
> 
> Implementing some functions of UUID in `Long` is inherently illogical. Now that we can implement it efficiently in the `UUID` class, there's no reason why we shouldn't.

this implementation is cleaner and is my favorite.

the implementation of changing JavaLangAccess and Long for UUID.toString in the baseline is too intrusive.

wenshao:
Should we consider that when COMPACT_STRINGS is false, should it only be guaranteed to work correctly, not the best performance. If based on this principle, the implementation code will be much simpler.

Glavo:
> Should we consider that when COMPACT_STRINGS is false, should it only be guaranteed to work correctly, not the best performance. If based on this principle, the implementation code will be much simpler.

I'm also curious about this question.

To be honest, I have never seen anyone actually use the `-XX:-CompactStrings`. Perhaps at some point in the future, we can investigate its usage and consider deprecate it to reduce the maintenance cost of JDK?

wenshao:
@AlanBateman i have modified according to @Glavo and @ExE-Boss 's suggestion, it's ready for review also.

src/java.base/share/classes/java/util/UUID.java
```diff
@@ -463,7 +468,54 @@ public long node() {
      */
     @Override
     public String toString() {
-        return jla.fastUUID(leastSigBits, mostSigBits);
```
turbanoff, line 466: `JavaLangAccess.fastUUID` method now can be removed

wenshao: it fixed.

wenshao:
@AlanBateman i have modified according to @Glavo and @ExE-Boss 's suggestion, it's ready for review also.

src/java.base/share/classes/java/util/UUID.java
```diff
@@ -463,7 +468,54 @@ public long node() {
...
+                        | ((long) digits[(((int) lsb) >> 24) & 0xff] << 16)
+                        | digits[(((int) lsb) >> 16) & 0xff]);
+        ByteArray.setInt(
+                buf,
+                32,
+                (digits[(((int) lsb) >> 8) & 0xff] << 16)
+                        | digits[((int) lsb) & 0xff]);
+
+        try {
+            return jla.newStringNoRepl(buf, StandardCharsets.ISO_8859_1);
```
turbanoff, line 515: in java.base it's better to use `ISO_8859_1.INSTANCE` directly instead of `StandardCharsets.ISO_8859_1`.
See [JDK-8234147](https://bugs.openjdk.org/browse/JDK-8234147)

wenshao: done

RogerRiggs:
> To be honest, I have never seen anyone actually use the `-XX:-CompactStrings`. Perhaps at some point in the future, we can investigate its usage and consider deprecate it to reduce the maintenance cost of JDK?

The implementation should do its best whether or not CompactStrings is true or false. (Until the time it is deprecated and remove).

AlanBateman:
> To be honest, I have never seen anyone actually use the `-XX:-CompactStrings`. Perhaps at some point in the future, we can investigate its usage and consider deprecate it to reduce the maintenance cost of JDK?

It's come up a few times. I don't think we have any data on usage. It's possible that there are down stream projects (like the J9 VM) that use.

liach:
I think we can work towards only optimizing the (compat string) case. Many other optimization patches improve the performance of certain scenarios without affecting, or at a slight cost for, other scenarios. I think this patch would be acceptable as long as the non-compat case doesn't get a significant overhead compared to the baseline.

RogerRiggs:
With respect to performance, it can't be worse than it is today, all optimization is about tradeoffs; in this case I would agree it does not need to be better.

wenshao:
should i revert to this version?
https://github.com/openjdk/jdk/pull/14578/files/17a9f16976488889907ec383ab76e2048dbc576b

wenshao:
with JVMOpts '-XX:-CompactStrings', i tested on the MacBookPro M1 Max machine and found that the current version is faster than the baseline version.

```
Benchmark                Mode  Cnt      Score      Error   Units (PR current version 6b61a715) 
UUIDUtilsBenchmark.jdk  thrpt    5  89601.097 ± 1879.659  ops/ms

Benchmark                Mode  Cnt      Score     Error   Units (baseline)
UUIDUtilsBenchmark.jdk  thrpt    5  71783.920 ± 170.914  ops/ms
```

wenshao:
the last implementation (commit id 6b61a715) is the best performant whether CompactStrings is true or false.

when CompactStrings is false, the improvement brought by using ByteArray is greater than the overhead of StringLatin1.inflate. the benchmark result above also confirms this.

because the implemented code is in UUID.toString, there is no need to intrude into the code of java/lang/Long.java, which is clearer and easier to maintain.

src/java.base/share/classes/java/util/UUID.java
liach, line 1: You should update the copyright year in the license header of UUID.java from 2022 to 2023.

liach:
For your info, the reviews required are approving GitHub reviews.

This patch looks great in shape; currently, I think the startup impact of ByteArray. If `UUID.toString` is used in early startup, this patch will increase the startup time for VarHandle initialization. #14636 replaces VarHandle with direct unsafe usage, which significantly reduces startup cost.

A search online says that `-Xlog:class+init=info:file=trace.log` can be added to VM options to print class initialization order. I think we might use it to check the startup impact by seeing if certain classes are loaded earlier now. Disclaimer: I personally haven't tried this; please correct me immediately if there is a better way to measure the startup impact.

AlanBateman:
I'm looking through [6b61a715], update 24.

Moving fastUUID out of Long is good.

UUID::toString can use StandardChars.ISO_8859_1, no need to use sun.nio.cs code here.

The list of imports in java.util.UUID is a bit messy now, can you clean this up.

It's confusing to have java.util.HexDigits and jdk.internal.util.HexDigits. It also creates an inconsistency with DecimalDigits and OctalDigits as the latter encapsulate their arrays. I assume you've done this to allow sharing of the 256 element array. I don't object to move the array to a supporting class but it needs to be renamed and the class description replaced as it is not a Digits implementation. Adding a private constructor would help make it clear that it should not be instantiated.

src/java.base/share/classes/java/util/UUID.java
wenshao: do I need to change the copyright information? I am not sure

Glavo: > do I need to change the copyright information? I am not sure

Yes, you should update the copyright information of each file you have modified.

src/java.base/share/classes/java/util/UUID.java
```diff
@@ -463,7 +467,54 @@ public long node() {
...
+        long lsb = leastSigBits;
+        long msb = mostSigBits;
+        short[] hex256 = Hex256.DIGITS;
+        byte[] buf = new byte[36];
+        ByteArray.setLong(
+                buf,
+                0,
+                ((long) hex256[((int) (msb >> 56)) & 0xff] << 48)
+                        | ((long) hex256[((int) (msb >> 48)) & 0xff] << 32)
+                        | ((long) hex256[((int) (msb >> 40)) & 0xff] << 16)
```
liach, line 479: The long cast seems redundant here.

src/java.base/share/classes/jdk/internal/util/Hex256.java
```diff
@@ -0,0 +1,52 @@
...
+
+package jdk.internal.util;
+
+import jdk.internal.vm.annotation.Stable;
+
+/**
+ * Provides a hexadecimal cache array of values from 0 to 255
+ */
+public final class Hex256 {
+    private Hex256(){
```
liach, line 34:
```suggestion
    private Hex256() {
```

src/java.base/share/classes/java/util/UUID.java
liach: The years in the header are in the form of "Copyright creationYear, lastModifiedYear, xxx". All other files already have their last modified year as 2023; only this one doesn't.

src/java.base/share/classes/jdk/internal/util/Hex256.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Stable
+    public static final short[] DIGITS;
```
dfuch, line 38: It's a bit smelly to have a public static field of type array.
Static analysers are likely to flag this.
It would be better to make the field private and have a public static method that returns DIGITS.clone() - then each class that needs it (and if I'm not mistaken there are only two) could encapsulate its own private copy.

liach: This array takes some time to prepare. We don't want each class to copy a large array when we can just prepare it once and use it everywhere within the JDK implementation. This array is, under no circumstances, leaked to users so it's safe.

dfuch: What I suggest is to prepare the array only once (in the static block as it is now), but have each class that use it encapsulate is own copy - obtained from clone(). Surely 256 shorts is not so large that we can't have two arrays?

Glavo: > What I suggest is to prepare the array only once (in the static block as it is now), but have each class that use it encapsulate is own copy - obtained from clone(). Surely 256 shorts is not so large that we can't have two arrays?

The point is that it doesn't make any sense, so I don't feel like we should pay anything for it.

RogerRiggs, line 38: If UUID shows up in startup class loading this array would be a candidate for loading by CDS, as mentioned in earlier review comments. Please create a separate issue to evaluate and add the code to load the array using CDS.

liach: > What I suggest is to prepare the array only once (in the static block as it is now), but have each class that use it encapsulate is own copy - obtained from clone(). Surely 256 shorts is not so large that we can't have two arrays?

I really wish we have frozen arrays, which are safely sharable like records. The closest we have right now is a final array with `@Stable`, which indicates array elements are lazy (any non-default value read can be treated as a constant by JIT); since this array is explicitly `@Stable final`, we can share it within the JDK like any other immutable objects.

RogerRiggs:
in general, do not rush to integrate, given the number and variety of comments and commenters, it is good practice to wait 24 hours after the last comment to see that comments have stabilized. 
Given the global scope of OpenJDK, reviewers are in most/every timezone.
And formally, every PR must have the required number of approvals from official Reviewers.
We're not there yet.

src/java.base/share/classes/jdk/internal/util/Hex256.java
liach: On second thought, we should be able to create a getter like
```java
@ForceInline
static short hex256(int byteValue) {
    return HEX256[byteValue];
}
```
and switch usages to that getter instead. This is a better approach than cloning the array and makes the array safe from modification by modules that get jdk.internal.util exports.

Glavo: > Sadly, jdk.internal.util is exported to other modules so it does need to be looked at from an integrity perspective.

This is indeed a problem to consider.

Maybe we can move this array back into `HexDigits`. Then we create a new package `jdk.internal.digits` and move `Digits`, `DecimalDigits`, `HexDigits`, `OctalDigits` all into it.

wenshao: Hex256 has been moved to jdk.internal.digits, jdk.internal.digits is not exported.

src/java.base/share/classes/jdk/internal/util/Hex256.java
```diff
@@ -0,0 +1,52 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package jdk.internal.util;
+
+import jdk.internal.vm.annotation.Stable;
+
+/**
+ * Provides a hexadecimal cache array of values from 0 to 255
```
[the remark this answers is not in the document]
wenshao: comments have been added, including usage examples.

AlanBateman:
@wenshao - adding a jdk.internal.digits package for this tiny class is overkill. For this PR, the simplest thing is to revert the changes to HexDigit, remove jdk.internal.digits.Hex256 and just put a small holder class in UUID for use by UUID.toString. That would be my preference to avoid splitting up HexDigits and creating inconsistencies in code used for string templates.

Glavo:
> @wenshao - adding a jdk.internal.digits package for this tiny class is overkill. For this PR, the simplest thing is to revert the changes to HexDigit, remove jdk.internal.digits.Hex256 and just put a small holder class in UUID for use by UUID.toString. That would be my preference to avoid splitting up HexDigits and creating inconsistencies in code used for string templates.

Wenshao hopes to use this array for other optimizations in `java.lang` in the future.

But as for this PR, I suggest wenshao not to do so much. I suggest moving `DIGITS` back to `HexDigits` first to make the modifications in this PR smaller. Reducing modifications helps with faster review.

When you need to optimize the classes in `java.lang`, we will continue to discuss how to share these arrays more safely and efficiently.

AlanBateman:
> Wenshao hopes to use this array for other optimizations in `java.lang` in the future.
> 
> But as for this PR, I suggest wenshao not to do so much. I suggest moving `DIGITS` back to `HexDigits` first to make the modifications in this PR smaller. Reducing modifications helps with faster review.
> 
> When you need to optimize the classes in `java.lang`, we will continue to discuss how to share these arrays more safely and efficiently.

Let's keep this PR as simple as possible. Right now, it's up to 100 comments and 27 iterations, it's annoying to be sending new contributor Wenshao in different directions. Future work that would allow safer sharing of arrays is future work, let's keep this PR as simple as possible. Not all applications use UUID.

src/java.base/share/classes/java/util/HexDigits.java
```diff
@@ -29,14 +29,61 @@
 
 import jdk.internal.vm.annotation.Stable;
 
+import static jdk.internal.digits.Hex256.hex256;
+
```
liach, line 33:
```suggestion
```
This is now redundant and causes build failure.

src/java.base/share/classes/jdk/internal/util/Hex256.java
wenshao: now it's reuse the field of HexDigits.DIGITS

wenshao: now it's reuse the field of java.util.HexDigits#DIGITS

AlanBateman:
I've read through 26f3a6a5 (iteration 29) and it mostly looks okay.

It's good to see the UUID code removed from Long.

It's unfortunate that UUID is directly accessing HexDigits.DIGITS but at least its in the same package and the array is accessed more widely. If there were frozen arrays then this would be better. Have you measured using an accessor method to access the array elements rather than accessing the array directly?

Somewhat subjective but I find the new comment in Digits a bit confusing, particularly "In the byte[] encoded in LATIN1" as a byte[] is just bytes and I think you want to say a String encoding as LATIN1 in a byte[]. Maybe drop this and the reference to jdk.internal.util.ByteArray.

wenshao:
> HexDigits.DIGITS



> I've read through [26f3a6a](https://github.com/openjdk/jdk/commit/26f3a6a5317718235bc3506f0a3e851001f43404) (iteration 29) and it mostly looks okay.
> 
> It's good to see the UUID code removed from Long.
> 
> It's unfortunate that UUID is directly accessing HexDigits.DIGITS but at least its in the same package and the array is accessed more widely. If there were frozen arrays then this would be better. Have you measured using an accessor method to access the array elements rather than accessing the array directly?
> 
> Somewhat subjective but I find the new comment in Digits a bit confusing, particularly "In the byte[] encoded in LATIN1" as a byte[] is just bytes and I think you want to say a String encoding as LATIN1 in a byte[]. Maybe drop this and the reference to jdk.internal.util.ByteArray.

Good suggestion, it has been modified and submitted. now UUID#toString codeSize is 203, less than default FreqInlineSize 325,  It can be inline, and the performance is better than the previous version.

AlanBateman (approved):
I think you've got this to a good place with version 93e74b6c. It might be helpful if the Digit.digts methods mentioned the endianness in the comment otherwise okay.

liach (approved):
I think ByteArray can be safely used here: UUID is not used by anything in java.base and shouldn't be initialized on VM startup.

src/java.base/share/classes/java/util/HexDigits.java
```diff
@@ -103,18 +103,18 @@ private HexDigits() {
     }
 
     /**
-     * Combine two hex shorts into one int
+     * Combine two hex shorts into a int based on BigEndian
```
DamonFool, line 106: 'a int' --> 'an int' ?

src/java.base/share/classes/java/util/HexDigits.java
```diff
@@ -64,6 +102,23 @@ final class HexDigits implements Digits {
     private HexDigits() {
     }
 
+    /**
+     * Combine two hex shorts into one int based on big endian
+     */
+    static int digit(int b0, int b1) {
```
RogerRiggs, line 108: The method name `digit` is not very informative.
As it is package private it should be more descriptive and describe the parameters and result.
Perhaps `packDigits` or `encodeAndPackHexDigits`.

AlanBateman:
@DamonFool  In the future, please wait at least 24 hours before sponsoring changes like this. There have been 33 revisions of this change, many of the people that have commented, potential reviewers are in many time zones and need to have time to see a chance to look at changes before they are integrated.

RogerRiggs (changes requested):

DamonFool:
> @DamonFool In the future, please wait at least 24 hours before sponsoring changes like this. There have been 33 revisions of this change, many of the people that have commented, potential reviewers are in many time zones and need to have time to see a chance to look at changes before they are integrated.

Okay, got it and will do so in the future.

src/java.base/share/classes/java/util/HexDigits.java
wenshao: can you create a new issue and i continue improving?

DamonFool: > can you create a new issue and i continue improving?

Filed: https://bugs.openjdk.org/browse/JDK-8311207

wenshao: @liach @RogerRiggs continue discussion in [PR 14745](https://github.com/openjdk/jdk/pull/14745)

PR #14585, opened 2023-06-21
8310537: Fix -Wconversion warnings in gcUtil.hpp

coleenp:
Please review small changes to fix -Wconversion warnings in gc code.  See CR for details.
Tested with tier1 on Oracle platforms.

---------
### Progress

### Issue
 * [JDK-8310537](https://bugs.openjdk.org/browse/JDK-8310537): Fix -Wconversion warnings in gcUtil.hpp (**Enhancement** - P4)

### Reviewers
 * [Albert Mingkun Yang](https://openjdk.org/census#ayang) (@albertnetymk - **Reviewer**) ⚠️ Review applies to [cf267363](https://git.openjdk.org/jdk/pull/14585/files/cf267363a80d452f68626d54cb11033ce5c44b92)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14585/head:pull/14585` \
`$ git checkout pull/14585`

`$ git checkout pull/14585` \
`$ git pull https://git.openjdk.org/jdk.git pull/14585/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14585`

`$ git pr show -t 14585`

</details>

<a href="https://git.openjdk.org/jdk/pull/14585.diff">https://git.openjdk.org/jdk/pull/14585.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14585#issuecomment-1600737119)

### DIFF
--- a/src/hotspot/share/gc/shared/gcUtil.cpp
+++ b/src/hotspot/share/gc/shared/gcUtil.cpp
@@ -118,8 +118,8 @@ void LinearLeastSquareFit::update(double x, double y) {
   _sum_x_squared = _sum_x_squared + x * x;
   _sum_y = _sum_y + y;
   _sum_xy = _sum_xy + x * y;
-  _mean_x.sample((float)x);
-  _mean_y.sample((float)y);
+  _mean_x.sample((float)x);  // Used to track generation sizes so casting to float should
+  _mean_y.sample((float)y);  // not lose precision for valid samples.
   assert(_mean_x.count() == _mean_y.count(), "Incorrect count");
   if ( _mean_x.count() > 1 ) {
     double slope_denominator;
--- a/src/hotspot/share/gc/shared/gcUtil.cpp
+++ b/src/hotspot/share/gc/shared/gcUtil.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2002, 2015, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2002, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -86,10 +86,10 @@ void AdaptivePaddedAverage::sample(float new_sample) {
 
   // Now update the deviation and the padded average.
   float new_avg = average();
-  float new_dev = compute_adaptive_average(fabsd(new_sample - new_avg),
+  float new_dev = compute_adaptive_average((float)fabsd(new_sample - new_avg),
                                            deviation());
   set_deviation(new_dev);
-  set_padded_average(new_avg + padding() * new_dev);
+  set_padded_average(new_avg + (float)padding() * new_dev);
   _last_sample = new_sample;
 }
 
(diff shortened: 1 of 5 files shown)


coleenp:
Thanks Albert.

src/hotspot/share/gc/shared/gcUtil.cpp
```diff
@@ -118,8 +118,8 @@ void LinearLeastSquareFit::update(double x, double y) {
   _sum_x_squared = _sum_x_squared + x * x;
   _sum_y = _sum_y + y;
   _sum_xy = _sum_xy + x * y;
-  _mean_x.sample(x);
-  _mean_y.sample(y);
+  _mean_x.sample((float)x);
+  _mean_y.sample((float)y);
```
xmas92, line 122: If this narrowed more than some small epsilon the behaviour would be strange. Seems like the actual interface here should be to sample float precision `x` and `y`.  (`void LinearLeastSquareFit::update(float x, float y)`)

I have no objection for doing it this way, just an observation of `LinearLeastSquareFit`

I know from my own experiments with `-Wconversions` that we have many areas where it is the interface that seems wrong. (Sometimes moving from int -> long -> int -> long, just to fit different subcomponent interfaces)

src/hotspot/share/gc/shared/gcUtil.cpp
```diff
@@ -86,10 +86,10 @@ void AdaptivePaddedAverage::sample(float new_sample) {
 
   // Now update the deviation and the padded average.
   float new_avg = average();
-  float new_dev = compute_adaptive_average(fabsd(new_sample - new_avg),
+  float new_dev = compute_adaptive_average((float)fabsd(new_sample - new_avg),
```
xmas92, line 89: Is JDK-6302670 still relevant, and/or are there other reasons we avoid `cmath`? This could just be `fabsf`

src/hotspot/share/gc/shared/gcUtil.cpp
coleenp: Oh you've had experiences with -Wconversions? Thank you, I can ask you questions. I've tried to fix the interface when possible to be more precise with int types. In this case, it was float so not as sure if this is the right thing to do.  I'm not sure I understand your comment.  LinearLeastSquare fit seems determined to take double, and AdaptiveWeightedAverage uses float. I'm not sure how to change this to make it safer.  My usual strategy would be to make the AdaptiveWeightedAverage take double but then the fall out from that would be more than I can deal with.  What do you think?

src/hotspot/share/gc/shared/gcUtil.cpp
coleenp: JDK-630267 does not seem relevant.  We should clean that up (or maybe at least the comment).  I changed these occurrences to fabs.

src/hotspot/share/gc/shared/gcUtil.cpp
xmas92: What I mean is that we are doing a least square fit on a bunch of datapoints. The algorithm requires the mean, sum and sum of squares. If the datapoints truly required double precision or magnitude then storing the sum and sum of square in a double will overflow and/or be inaccurate. 

AdaptiveWeightedAverage having a float interface strengthens the case that the samples should be floats, as the running mean should never overflow or underflow the samples `min sample ≤ mean sample ≤ max sample`, (may lose precision, but if the samples are float precision, it seems resonable that the mean also have this precision)

It was not about safety but about having the interface express the capabilities of the class. It can sample float x,y points and provide a least square fit linear function`y(x)`

xmas92: At least that is my interpretation when I ever see intermediate sums use a type with more bits of information than the samples/parts. But maybe that was never the intent.

coleenp: I tried changing update to take floats but there are callers that cast unsigned ints to double to call this function, so they get -Wconversion messages.  Since this seems to not lose precision, I added a comment and suggest further cleanup in this area.

xmas92 (approved):

coleenp:
Thanks Axel and for your help.
/integrate

PR #14586, opened 2023-06-21
8275202: C2: optimize out more redundant conditions

rwestrel:
This change adds a new loop opts pass to optimize redundant conditions
such as the second one in:

```
if (i < 10) {
  if (i < 42) {
```

In the branch of the first if, the type of i can be narrowed down to
[min_jint, 9] which can then be used to constant fold the second
condition.
  
The compiler already keeps track of type[n] for every node in the
current compilation unit. That's not sufficient to optimize the
snippet above though because the type of i can only be narrowed in
some sections of the control flow (that is a subset of all
controls). The solution is to build a new table that tracks the type
of n at every control c

```
type'[n, root] = type[n] // initialized from igvn's type table
type'[n, c] = type[n, idom(c)]
```

This pass iterates over the CFG looking for conditions such as:

```
if (i < 10) {
```

that allows narrowing the type of i and updates the type' table
accordingly.

At a region r:

```
type'[n, r] = meet(type'[n, r->in(1)], type'[n, r->in(2)]...)
```

For a Phi phi at a region r:

```
type'[phi, r] = meet(type'[phi->in(1), r->in(1)], type'[phi->in(2), r->in(2)]...)
```

Once a type is narrowed, uses are enqueued and their types are
computed by calling the Value() methods. If a use's type is narrowed,
it's recorded at c in the type' table. Value() methods retrieve types
from the type table, not the type' table. To address that issue while
leaving Value() methods unchanged, before calling Value() at c, the
type table is updated so:

```
type[n] = type'[n, c]
```

An exception is for Phi::Value which needs to retrieve the type of
nodes are various controls: there, a new type(Node* n, Node* c)
method is used.

For most n and c, type'[n, c] is likely the same as type[n], the type
recorded in the global igvn table (that is there shouldn't be many
nodes at only a few control for which we can narrow the type down). As
a consequence, the types'[n, c] table is implemented with:

- At c, narrowed down types are stored in a GrowableArray. Each entry
  records the previous type at idom(c) and the narrowed down type at
  c.

- The GrowableArray of type updates is recorded in a hash table
  indexed by c. If there's no update at c, there's no entry in the
  hash table.

This pass operates in 2 steps:

- it first iterates over the graph looking for conditions that narrow
  the types of some nodes and propagate type updates to uses until a
  fix point.

- it transforms the graph so newly found constant nodes are folded.

The new pass is run on every loop opts. There are a couple reasons for
that:

1- one of the goals is to avoid the many bugs we've been hitting where
  a CastII nodes capture a type that, at some point, conflicts with
  its input type. The CastII becomes top, that control path is dead
  but the compiler fails to prove it. By running the new pass often,
  there's a better chance that these inconsistencies can be caught and
  fixed.
  
2- I tried running it less often and ran into inconsistencies where:
  when the new pass is run, it eliminates a range check. At a later
  point, the range check CastII becomes top. If the pass is not run
  then, we hit the problem of 1- above. If it is run, then it can
  prove the path that leads to the CastII is dead.
  
I looked at compilation time by running ctw on java.base and looking
at times reported by CITime. Overall, the new pass adds +50% to the
time spent in loop opts and about 5% to total compilation time. I
already spent quite a bit of time trying to decrease the compilation
time overhead and I don't see any obvious bottleneck at this
point. The pass runs until a fixed point is reached and can go over
the cfg several times. If it goes only once over the cfg, the
compilation time overhead goes down to ~40%. That's a fairly small
improvement in terms of compilation time in my opinion and, I would
say, it's better to let it run until a fixed point and find all
constants that it can find.

There are several changes that I had to make outside the new pass
itself. Some of them could be pushed separately eventhough they are
solving issues that may not exist without the new pass. In particular,
I had to add assert predicates for all eliminated conditions during
range check eliminations.

---------
### Progress

### Issue
 * [JDK-8275202](https://bugs.openjdk.org/browse/JDK-8275202): C2: optimize out more redundant conditions (**Enhancement** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14586/head:pull/14586` \
`$ git checkout pull/14586`

`$ git checkout pull/14586` \
`$ git pull https://git.openjdk.org/jdk.git pull/14586/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14586`

`$ git pr show -t 14586`

</details>

<a href="https://git.openjdk.org/jdk/pull/14586.diff">https://git.openjdk.org/jdk/pull/14586.diff</a>

</details>
<details><summary>Using Webrev</summary>

[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14586#issuecomment-1600845371)
</details>


rwestrel:
I realize this is complicated to review and that it's a risky change.  Maybe first steps could be to:

- agree on whether this is worth pursuing
- decide whether to integrate c2 changes that are required for this to be stable but not directly related to this change separately. I could start preparing PRs for them then.

vnkozlov:
1. Yes, it is worth pursuing. This is obvious missing optimization opportunity.
2. Yes, not directly related changes could be pushed separately.

Do you know why CCP phase can't help for simple cases (not in loop)? Can we fix CCP to handle simple cases?

rwestrel:
> Do you know why CCP phase can't help for simple cases (not in loop)? Can we fix CCP to handle simple cases?

CCP and IGVN keep track of types in a global table. In the example above:
```
if (i < 10) {
  if (i < 42) {
```
`i`'s type is `[min, 9]` only in the if branch so that `[min,9]` can't be stored in a global table that contains types that must be valid everywhere in the IR graph. The way c2 handles control dependent types is with cast nodes. The change I propose is equivalent to having cast nodes on every projection of every `CmpI`/`CmpU`/`CmpL`/`CmpL` for both inputs of the cmp node and then have logic in igvn to narrow the type of the cast. Instead of using cast nodes, that change keeps track of types in side tables. I think it's much better than trying to make things work with cast nodes because it's a lot less invasive and disruptive to the rest of c2. It's a single new standalone pass contained in a couple files that runs during loop opts.

What's your opinion on the compile time overhead for this? Do you find the number I mention too high?

vnkozlov:
Thank you for explanation. Yes, time increase is concerning. But I agree that it should run to fixed point.
May be look on the case which consume a lot of time and investigate why. May be you have false positives when you look for narrowing type conditions.

rwestrel:
Comment to keep alive.

test/hotspot/jtreg/compiler/c2/irTests/TestLoopConditionalPropagation.java
```diff
@@ -0,0 +1,730 @@
...
+    // @IR(failOn = { IRNode.IF })
+    // private static int test14(int i) {
+    //     if (i < 0) {
+    //         i = 0;
+    //     }
+    //     if (i - 1 < -1) {
+    //         throw new RuntimeException("never taken");
+    //     }
+    //     return i;
+    // }
```
eme64, line 678: Dead code? You are probably aware of this.

rwestrel:
> @rwestrel this looks like an awesome idea. I gave the code a quick glance, and have a few questions:

Thanks for looking at this. I have an updated patch coming but I'll make sure to look at your comments before I send it.

> **How powerful is this?** Can we make inferences from `expr(i) > 0` about `i` itself? I guess not really, that would require us to propagate type information "upward" through `expr` to `i`.
> 
> Assume we have`expr1(i) > 0` and `expr2(i) < 0`:
> 
>     * And let's assume that together this implied that `i` was a constant at c. Your algorithm would not detect this, right?

No
It can do:

```
if (2 * i < 0) {
  if (i >= 0 && i <= 10) {
    // i in [0, 10] => 2 * i in [0, 20] which is impossible so this branch is unreachable
  }
}
```

>     * Alternatively, let's assume this meant that `i` was TOP at c. We would not directly detect this, but we would somehow detect that one of the two expressions is not TOP at c, and therefore we could collapse c? This is very powerful for dead code elimination.

Yes, I think that's correct.

> **How expensive is this?** Worst case, you cannot get around having a basically quadratic runtime and memory requirement, right? So basically `O(C * D)`, where C is number of control nodes, and D is number of data nodes. I'm trying to think of such a worst-case example: Can there be a program where we have lots of expressions of `i`, and lots of comparisons on `i`, and therefore have to update all expressions of `i` at every comparison of `i`? This worst-case is probably far off of the "average" program, right? But still, one wonders what the distribution is - even the occasional program/method that hits quadratic runtime could be problematically expensive.
> 
> If you don't think the runtime is indeed worst-case quadratic, then I'd be interested in hearing the argument ;)

I think you're right about this.
I've seen a case where the type of a single node is repeatedly narrowed and that node has many uses. That method has no loop so the compiler spends almost no time in loop opts without the patch and quite a bit of time with the patch. I found that method while looking at compilation times for CTW on java.base. Working around the issue improves compilation time of that method but made no difference on the total time spent on java.base. So it's true that there could be pathological cases but they seem to be uncommon.

> **Do you think this really solves all the other bugs?** If we put this behind a product flag, then we do need to fix all other bugs that are linked as duplicate separately. But that may not be easy, because they basically happen because separate type info about expressions of `i` are suddenly combined at some data node, but not visible to the control -> inconsistent graph.

So maybe this should be behind a diagnostic flag. The reason for having a flag is to diagnose issues that this could cause: new crashes, incorrect execution or maybe compilation time becoming to expensive and maybe work around those issues temporarily.

src/hotspot/share/opto/loopConditionalPropagation.cpp
```diff
@@ -0,0 +1,1213 @@
...
+
+/*
+ With the following code snippet:
+ if (i - 1) > 0) {
+    // i - 1 in [1, max]
+   if (i == 0) {
+     // i - 1 is both -1 and [1, max] so top
+
+ The second if is redundant but first if updates the type of i-1, not i alone, we can't tell i != 0.
+ Because i-1 becomes top in the second if branch, we can tell that branch is dead
```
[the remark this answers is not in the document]
rwestrel: `i + 1 > 0` and` i - 1 < 0` doesn't work indeed.

src/hotspot/share/opto/loopConditionalPropagation.cpp
```diff
@@ -0,0 +1,1213 @@
...
+    // mark as needing processing either in the pass over the CFG (control not yet processed) or on the next one
+    assert(!_visited.test(u->in(0)->_idx), "");
+    _control_dependent_node[_iterations%2].set(u->in(0)->_idx);
+    _control_dependent_node[_iterations%2].set(u->_idx);
+    return false;
+  }
+  return true;
+}
+
+void PhaseConditionalPropagation::enqueue_uses(const Node* n, Node* c) {
```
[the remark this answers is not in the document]
rwestrel: You're right. Having shared code to push uses that's shared across CCP, IGVN and this would be nice. There are 2 kind of issues, right: 1) missed opportunities that are otherwise harmless 2) inconsistencies that cause something to break. I would be happy if this avoided 2) initially. There's verification code that checks that no further progress can be made once we reached a fixed point.

dean-long:
I'm probably missing something, but isn't there a simpler way to accomplish this?  If we create a new node, then we can attach the constrained type to that node.  So instead of relying on dominator information for the constrained type, we just grab the current value of the local:
```
  if (i < 10) {
    i = ConstraintCast("i < 10"); // injected
    if (i < 42) {  // sees new constrained type above
    }
  } else {
    i = ConstraintCast("i >= 10");
    if (i < 42) {
    }
  }
```
In case of nested ifs, we end of up with ConstraintCast(ConstraintCast(...)) to flatten appropriately.

dean-long:
I see that `Parse::sharpen_type_after_if` is already doing what I suggested, so I guess I don't understand why that isn't enough.

dean-long:
For some reason the new tests under test/hotspot/jtreg/compiler/loopconditionalpropagation are passing, even without these new changes.

rwestrel:
> For some reason the new tests under test/hotspot/jtreg/compiler/loopconditionalpropagation are passing, even without these new changes.

I added those tests for issues I ran into while working on the patch. They are bugs that were introduced by the patch (and that I then fixed) but that I found to be tricky enough to get right that I thought it was best to have a test for them.
`compiler/c2/irTests/TestLoopConditionalPropagation.java`, the one that verifies that the patch optimizes code better, fails without the patch.

rwestrel:
> I see that `Parse::sharpen_type_after_if` is already doing what I suggested, so I guess I don't understand why that isn't enough.

It's true but for integer types, it's only for `BoolTest::eq`, that is:
```
if (i == 10) {
  // use 10 instead of i here
```
but not:
```
if (i < 10) {
  // use (CastII i [min, 9]) here
}

**Comment by rwestrel** [comment]
> I'm probably missing something, but isn't there a simpler way to accomplish this? If we create a new node, then we can attach the constrained type to that node. So instead of relying on dominator information for the constrained type, we just grab the current value of the local:
> 
> ```
>   if (i < 10) {
>     i = ConstraintCast("i < 10", i); // injected
>     if (i < 42) {  // sees new constrained type above
>     }
>   } else {
>     i = ConstraintCast("i >= 10", i);
>     if (i < 42) {
>     }
>   }
> ```
> 
> In case of nested ifs, we end of up with ConstraintCast(ConstraintCast(...)) to flatten appropriately.

That's another way to achieve the same result but I don't think it's simpler:

- there would be a lot more casts in the graph. So pattern matching that c2 relies on almost everywhere including `Ideal` or `Identity` calls would have to be updated to take the extra cast nodes into account. That would maybe be ok if we had tests for every transformation in the compiler. We don't but making extensive use of casts would essentially require writing them.
- casts are a source of bugs. A change to `PhaseIdealLoop::try_sink_out_of_loop()` caused more casts to be added to pin nodes out of loops. They exposed inconsistencies in types between the control and data subgraphs and we've been struggling to fix all bugs that the fuzzer has found since.
- the way casts are added at parse time doesn't cover all cases. So:
```
if (field >= 0) {
  if (field <= 10) {
    // use field
```
Cast are propagated to uses by replacing a local's value with a cast in the c2 parse time map. If `field` is loaded from memory twice, then replacing the local for the first load with a narrowed type doesn't propagate tho the second load.
- casts introduce extra dependencies. So for instance:
```
if (i >= 0) {
  v = array[i];
  if (i <= 10) {
    v += array[i];
```
The 2 loads should common but if we add cast nodes after each if, then the address computed for each array load is different because it depends on a different cast node and the loads won't common.

**Comment by dean-long** [comment]
OK, thanks for the explanation about the problem with casts.

**Comment by dean-long** [comment]
Does your solution handle code like this?
```
    if (i >= 1) {
      if (i <= 2) {
        if (i != 2) {
          // do we know i == 1 here?
```
because the type system doesn't seem to handle != very well, unless it is != min or max.

**Comment by dean-long** [comment]
BTW, the IR tests don't run with -XX:-UseLoopConditionalPropagation because it's not a recognized flag.

**Comment by dean-long** [comment]
test13 seems to pass even with UseLoopConditionalPropagation set to false.

**Comment by merykitty** [comment]
Can the inspection be made in retrospect. In other word, can it infer this:

    if (i < j) {
        if (i > 4) {
            // then j >= 6 here
        }
    }

Thanks.

**Comment by rwestrel** [comment]
> Does your solution handle code like this?
> 
> ```
>     if (i >= 1) {
>       if (i <= 2) {
>         if (i != 2) {
>           // do we know i == 1 here?
> ```
> 
> because the type system doesn't seem to handle != very well, unless it is != min or max.

I haven't tried it but I think it should. The new code relies on existing logic to compute the narrowed type:

https://github.com/openjdk/jdk/blob/1eb88623fa550f5fca9bb3e4f5f6c05d32d69384/src/hotspot/share/opto/subnode.cpp#L1899

And that logic does the right thing AFAICT.

**Comment by rwestrel** [comment]
> BTW, the IR tests don't run with -XX:-UseLoopConditionalPropagation because it's not a recognized flag.
> test13 seems to pass even with UseLoopConditionalPropagation set to false.

Thanks. I'll have a look at those issues.

**Comment by rwestrel** [comment]
A comment to keep this open.

**Comment by robcasloz** [comment]
> except for a potential regression in SPECjvm2008-Compress which I am trying to confirm

I did a second run of this benchmark with more repetitions (20, compared to 5 in the original run) and the regression went down from around 4% to 2%. I guess it might be due to some unfortunate interaction and does not deserve much attention. [Here](https://github.com/openjdk/jdk/files/13487199/SPECjvm2008-Compress.log) is the standard output of running the benchmark with `-Xbatch -XX:+PrintCompilation -XX:+PrintLoopConditionalPropagation` for whatever it's worth.

**Comment by rwestrel** [comment]
Hi Roberto,

Thanks for the beedback.

> Do you know of other benchmarks / applications where the proposed optimization gives measurable overall improvements?

I'm not. Out of curiosity, have you tried anything from Renaissance?

> I also compared the overall C2 speed (as reported by `-XX:+CITime` in bytes/s) with and without `UseLoopConditionalPropagation` on DaCapo and SPECjvm2008 (x64 only), additionally using `-Xbatch -XX:-TieredCompilation` to reduce variability. Overall, in this context the proposed pass slows down C2 by more than 10%, see results here: [C2-median-speed-baseline-vs-JDK-8275202.pdf](https://github.com/openjdk/jdk/files/13476248/C2-median-speed-baseline-vs-JDK-8275202.pdf) and normalized version here: [C2-median-speed-baseline-vs-JDK-8275202-normalized-to-baseline.pdf](https://github.com/openjdk/jdk/files/13476257/C2-median-speed-baseline-vs-JDK-8275202-normalized-to-baseline.pdf).

Thanks for the details. I'm aware of the impact on compilation time and I will propose a new change where the slow down is not as big (the new pass is run less often).

**Comment by rwestrel** [comment]
@robcasloz thanks!

**Comment by caojoshua** [comment]
Would it make sense to handle this case in `IfNode::ideal()`? `IfNode::fold_compares` already handles more complicated cases of integer comparisons. We can match `cmp(i, constant)`, build integer range data, and eliminate IfNode's based on the data.

It won't be as powerful as the fixed point optimization that is proposed in this PR, but it is sufficient to cover the case mentioned in the JBS issue and would have less compile time overhead.

**Comment by rwestrel** [comment]
> It won't be as powerful as the fixed point optimization that is proposed in this PR, but it is sufficient to cover the case mentioned in the JBS issue and would have less compile time overhead.

I think it would do a lot less so in my opinion, it's worth pursuing the fixed point optimization. The main reason it has such an overhead in the current patch is that it's run too often (every time there's a loop opts pass). There are corner cases that were hard to handled otherwise (and cause the compiler to crash). I've been working on a new version of the patch where the pass is run only once or twice (and found ways to handle the corner cases that were problematic). It's not ready yet.

**Comment by caojoshua** [comment]
@rwestrel Do you plan to cover cases like this?
```
if (arr.length - 1 < arr.length)
```

We can fold this to `true` because we know `arr.length >= 0`.

I'm looking at https://bugs.openjdk.org/browse/JDK-8325151. I don't believe we can cover this with integer range data. If you do not plan on covering this, I will work on it separately.

**Comment by rwestrel** [comment]
> @rwestrel Do you plan to cover cases like this?
> 
> ```
> if (arr.length - 1 < arr.length)
> ```

I don't expect this to be covered.

**Comment by rwestrel** [comment]
comment to keep alive

**Comment by rwestrel** [comment]
Another comment to keep alive

**Comment by rwestrel** [comment]
I still intend to propose an updated patch for this

File: src/hotspot/share/opto/loopConditionalPropagation.cpp [hunk 4]
```diff
@@ -0,0 +1,1213 @@
...
+        bol = bol->in(1);
+      }
+      if (bol->is_Bool()) {
+        Node* cmp = bol->in(1);
+        if (cmp->Opcode() == Op_CmpI || cmp->Opcode() == Op_CmpU ||
+            cmp->Opcode() == Op_CmpL || cmp->Opcode() == Op_CmpUL) {
+          Node* cmp1 = cmp->in(1);
+          Node* cmp2 = cmp->in(2);
+          sync(iff);
+          // narrowing the type of a LoadRange could cause a range check to optimize out and a Load to be hoisted above
```
merykitty, line 519: Not sure if you are still working on this one. Here you can remove the range check if you pin all the nodes that depend on this check. Similar logic can be found in range check smearing.

rwestrel: Yes, I'm still working on this. I'll update this PR soon, hopefully. I reworked it quite a bit.

merykitty: Some more observations: When removing an `IfNode`, not only for `LoadNode`, you will need to pin all nodes that `depends_only_on_test` at that point (e.g. `ConstraintCast`), since the node does not only depend on the immediate dominating test (if any) but on the whole sequence of control nodes leading to that position. This can be done by e.g upgrading a `RegularDependency` `ConstraintCast` to one with `StrongDependency`. However, this action can lead to the node not able to float as freely. So I think before completing all loop opts, you should refrain from removing any `IfNode` of which the taken path has at least 1 node that `depends_only_on_test`. Pruning the untaken branch should still be done to simplify the graph. I think you have probably thought about this but just in case it can help.

rwestrel:
I finally updated this PR. The main push back from the previous version of this change was excessive compile time. The updated change addresses this issue. In the previous version of the change, the new optimization pass was run on every pass of loop optimizations. The reason for that was issues similar to the one explained in https://github.com/openjdk/jdk/pull/23468 . Running the new pass often was a way to mitigate the issue. Since the first version of this PR, I actually found code patterns where running the pass often was not even sufficient to prevent a crash. The change from 8349479 solves all those issues and running the pass only once or a few times doesn't cause any problem. This helps compilation time quite a bit. According to my rough measurement (running `CompileTheWorld` on java.base and looking at times reported by `CITime`), the overhead of the new pass is now around 20% for `IdealLoop` (from 4.7s to 5.7s) and around 2.5% on total compilation time (1 extra second on around 40s of total compilation time).

I also refactored the code quite a bit, worked on integrating changes that are non specific to the new pass and is in other parts of the compiler so there are a lot fewer unrelated changes now, added more tests and comments.

src/hotspot/share/opto/loopConditionalPropagation.cpp
rwestrel: Thanks for the observations.
The updated patch does pin a `LoadNode` when a range check it depends on is optimized out. It doesn't pin it at the control that immediately dominates the eliminated range check. Instead, it goes over the dominating controls until it finds the earliest control at which the type of the range check test constant folds and sets the `LoadNode` control to that control. So it pins it at the earliest possible control.
The new constant propagation pass is also run after all loop opts by default so, nodes get pinned late in the compilation process.

src/hotspot/share/opto/loopConditionalPropagation.cpp
```diff
@@ -0,0 +1,1815 @@
...
+      Node* con = nullptr;
+      for (DUIterator i = node->outs(); node->has_out(i); i++) {
+        Node* use = node->out(i);
+        if (use->is_Phi()) {
+          Node* r = use->in(0);
+          int nb_deleted = 0;
+          for (uint j = 1; j < use->req(); ++j) {
+            if (use->in(j) == node && !r->in(j)->is_top() && _conditional_propagation.is_dominator(c, r->in(j)) &&
+                is_safe_for_replacement_at_phi(node, use, r, j)) {
+              if (con == NULL) {
```
robcasloz:
```suggestion
              if (con == nullptr) {
```

robcasloz:
> According to my rough measurement (...), the overhead of the new pass is now (...) around 2.5% on total compilation time

Thanks for working on this. This matches the results I see on DaCapo 23.11-chopin on x64 (mean 3% slowdown). The results seem a bit worse on aarch64 (Apple M1), there I see a mean slowdown of around 5% (from 3% for `jython` up to 10% for `eclipse`). Still a great improvement compared to the initial version!

robcasloz:
I tested this changeset applied on top of  jdk-25+26 (Oracle CI tier1-5) and found the following issues (besides the trivial `NULL` occurrence reported above):

- `assert(c->_idx >= _unique || _type_table->find_type_between(c, c, _phase->C->root()) != Type::TOP) failed: for If we don't follow dead projections`  in multiple tests, e.g. `compiler/predicates/TestHoistedPredicateForNonRangeCheck.java` and 	`compiler/predicates/assertion/TestOpaqueInitializedAssertionPredicateNode.java`.

- `assert(!failure) failed: Missed optimization opportunity in PhaseIterGVN` in multiple tests, e.g. `compiler/predicates/TestHoistedPredicateForNonRangeCheck.java` and `compiler/codegen/aes/TestCipherBlockChainingEncrypt.java`.

- `There were one or multiple IR rule failures` in `compiler/loopconditionalpropagation/TestLoopConditionalPropagation.java`.

Let me know if you need more details to reproduce these issues.

rwestrel:
> Let me know if you need more details to reproduce these issues.

Thanks for the reports but I can't reproduce any of them. Are there extra options that need to be passed to the JVM?

robcasloz:
> Thanks for the reports but I can't reproduce any of them. Are there extra options that need to be passed to the JVM?

You are right, here is a more detailed report:

```
make run-test CONF=linux-x64-debug TEST="compiler/predicates/assertion/TestOpaqueInitializedAssertionPredicateNode.java" TEST_VM_OPTS="-XX:StressLongCountedLoop=2000000"
make run-test CONF=linux-x64-debug TEST="compiler/predicates/TestHoistedPredicateForNonRangeCheck.java" TEST_VM_OPTS="-XX:-TieredCompilation -XX:VerifyIterativeGVN=10"
make run-test CONF=linux-x64-debug TEST="compiler/loopconditionalpropagation/TestLoopConditionalPropagation.java" TEST_VM_OPTS="-XX:-TieredCompilation"
make run-test CONF=linux-x64-debug TEST="compiler/codegen/aes/TestCipherBlockChainingEncrypt.java" TEST_VM_OPTS="-XX:-TieredCompilation -XX:VerifyIterativeGVN=10"
make run-test CONF=linux-x64-debug TEST="compiler/predicates/TestHoistedPredicateForNonRangeCheck.java" TEST_VM_OPTS="-XX:-TieredCompilation -XX:VerifyIterativeGVN=10"
```

Hope that helps!

rwestrel:
> I tested this changeset applied on top of jdk-25+26 (Oracle CI tier1-5) and found the following issues (besides the trivial `NULL` occurrence reported above):

I pushed new commits that should address those failures. I added a test case for that one (a tricky issue):

>     * `assert(c->_idx >= _unique || _type_table->find_type_between(c, c, _phase->C->root()) != Type::TOP) failed: for If we don't follow dead projections`  in multiple tests, e.g. `compiler/predicates/TestHoistedPredicateForNonRangeCheck.java` and 	`compiler/predicates/assertion/TestOpaqueInitializedAssertionPredicateNode.java`.

New commits also include some tweaks and cleanup.

@robcasloz would you mind running tests again?

PR #14591, opened 2023-06-21
8306922: IR verification fails because IR dump is chopped up

eme64:
Some IR tests were failing because the logs were messed up (badly interleaved), due to multiple threads writing at the same time.

The issue was in `Compile::print_ideal_ir`. While we did lock the `tty`, this lock was broken because the node dumping sometimes calls into the VM, which can trigger Safepoints, which in turn can explicitly break the locks, with `ttyLocker::break_tty_lock_for_safepoint`.

What good is a lock if it can be broken? Not much. Though it is only broken if there is a Safepoint, so if we can avoid safepointing while holding the lock we should be safe. To verify that, we can write:

```
  NoSafepointVerifier nsv;
  ttyLocker ttyl;
```

The verifier triggered immediately, as expected.
And I now gather the whole dump into a separate `stringStream` first, so that all the Safepoints can happen before I even acquire the lock. Then, I acquire the log and just print the `stringStream` to `tty / xtty`, without triggering any Safepoint.

We still need to acquire a lock, else other threads may be printing also, and we get a bad interleaving in the locks.

I had to enable `dump_bfs` to print to an arbitrary stream, instead of just `tty`.

I un-problemlisted `compiler/c2/irTests/TestVectorConditionalMove.java`, and fixed a CompileCommand issue for it.

**Follow-up work**

[JDK-8310712](https://bugs.openjdk.org/browse/JDK-8310712) C2: check for broken tty locks due to SafePoint
We should in general go through all uses of `ttyLocker`, and add a `NoSafepointVerifier` to ensure no such lock is broken. Or maybe we just build the verifier into the ttyLocker?

[JDK-8310711](https://bugs.openjdk.org/browse/JDK-8310711) IR Framework: remove safepoint while printing handling

**Testing**

I ran `TestVectorConditionalMove.java` 1000x on master, just with the CompileCommand issue fixed. It triggered an IR issue 5 times (hit rate `0.005`). With the fix, it never triggers (`0.995^1000 = 0.007`). I also ran up to tier6 and stress-testing.

---------
### Progress

### Issue
 * [JDK-8306922](https://bugs.openjdk.org/browse/JDK-8306922): IR verification fails because IR dump is chopped up (**Bug** - P4)

### Reviewers
 * [Christian Hagedorn](https://openjdk.org/census#chagedorn) (@chhagedorn - **Reviewer**) ⚠️ Review applies to [bca07373](https://git.openjdk.org/jdk/pull/14591/files/bca07373576431c515f26f83fb0e8611a812a7bb)
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - **Reviewer**) ⚠️ Review applies to [bca07373](https://git.openjdk.org/jdk/pull/14591/files/bca07373576431c515f26f83fb0e8611a812a7bb)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14591/head:pull/14591` \
`$ git checkout pull/14591`

`$ git checkout pull/14591` \
`$ git pull https://git.openjdk.org/jdk.git pull/14591/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14591`

`$ git pr show -t 14591`

</details>

<a href="https://git.openjdk.org/jdk/pull/14591.diff">https://git.openjdk.org/jdk/pull/14591.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14591#issuecomment-1603861295)

### DIFF
--- a/src/hotspot/share/opto/compile.cpp
+++ b/src/hotspot/share/opto/compile.cpp
@@ -558,6 +558,7 @@ void Compile::print_ideal_ir(const char* phase_name) {
 
   // Node dumping can cause a safepoint, which can break the tty lock.
   // Buffer all node dumps, so that all safepoints happen before we lock.
+  ResourceMark rm;
   stringStream ss;
 
   if (_output == nullptr) {
--- a/src/hotspot/share/opto/compile.cpp
+++ b/src/hotspot/share/opto/compile.cpp
@@ -556,16 +556,9 @@ void Compile::print_ideal_ir(const char* phase_name) {
   // To enable tools to match it up with the compilation activity,
   // be sure to tag this tty output with the compile ID.
 
-  // Buffer first, and dump all at once to avoid output splitting.
+  // Node dumping can cause a safepoint, which can break the ttyLocker.
+  // Buffer all node dumps, so that all safepoints happen before we lock.
   stringStream ss;
-  xmlStream x(&ss);
-
-  if (xtty != nullptr) {
-    x.head("ideal compile_id='%d'%s compile_phase='%s'",
-           compile_id(),
-           is_osr_compilation() ? " compile_kind='osr'" : "",
-           phase_name);
-  }
 
   if (_output == nullptr) {
     ss.print_cr("AFTER: %s", phase_name);
@@ -576,13 +569,16 @@ void Compile::print_ideal_ir(const char* phase_name) {
     _output->print_scheduling(&ss);
   }
 
+  // Check that the lock is not broken by a safepoint.
+  NoSafepointVerifier nsv;
+  ttyLocker ttyl;
   if (xtty != nullptr) {
-    x.tail("ideal");
-  }
-
-  // Print and flush all at once
-  if (xtty != nullptr) {
-    xtty->print("%s", ss.as_string());
+    xtty->head("ideal compile_id='%d'%s compile_phase='%s'",
+               compile_id(),
+               is_osr_compilation() ? " compile_kind='osr'" : "",
+               phase_name);
+    tty->print("%s", ss.as_string());
+    xtty->tail("ideal");
   } else {
     tty->print("%s", ss.as_string());
   }
--- a/src/hotspot/share/opto/compile.cpp
+++ b/src/hotspot/share/opto/compile.cpp
@@ -581,6 +581,7 @@ void Compile::print_ideal_ir(const char* phase_name) {
   }
 
   // Print and flush all at once
+  ttyLocker ttyl;
   tty->print("%s", ss.as_string());
 }
 #endif
--- a/src/hotspot/share/opto/compile.cpp
+++ b/src/hotspot/share/opto/compile.cpp
@@ -565,7 +565,11 @@ void Compile::print_ideal_ir(const char* phase_name) {
   if (_output == nullptr) {
     tty->print_cr("AFTER: %s", phase_name);
     // Print out all nodes in ascending order of index.
-    root()->dump_bfs(MaxNodeLimit, nullptr, "+S$");
+    // Buffer the dump first, and print it at once to avoid
+    // having to lock on tty or risk having the dump split.
+    stringStream ss;
+    root()->dump_bfs(MaxNodeLimit, nullptr, "+S$", &ss);
+    tty->print("%s", ss.as_string());
   } else {
     // Dump the node blockwise if we have a scheduling
     _output->print_scheduling();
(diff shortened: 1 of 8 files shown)


src/hotspot/share/opto/compile.cpp
```diff
@@ -551,28 +551,36 @@ void Compile::print_compile_messages() {
...
   // To enable tools to match it up with the compilation activity,
   // be sure to tag this tty output with the compile ID.
-  if (xtty != nullptr) {
-    xtty->head("ideal compile_id='%d'%s compile_phase='%s'",
-               compile_id(),
-               is_osr_compilation() ? " compile_kind='osr'" : "",
-               phase_name);
-  }
+
+  // Node dumping can cause a safepoint, which can break the ttyLocker.
```
chhagedorn, line 559:
```suggestion
  // Node dumping can cause a safepoint, which can break the tty lock.
```

src/hotspot/share/opto/output.cpp
```diff
@@ -2068,23 +2068,29 @@ void PhaseOutput::ScheduleAndBundle() {
...
+    tty->print("%s", ss.as_string());
   }
 #endif
 }
 
 #ifndef PRODUCT
 // Separated out so that it can be called directly from debugger
 void PhaseOutput::print_scheduling() {
+  print_scheduling(tty);
+}
```
chhagedorn, line 2084: New line:
```suggestion
}

```

test/hotspot/jtreg/compiler/c2/irTests/TestVectorConditionalMove.java
```diff
@@ -46,7 +46,7 @@ public static void main(String[] args) {
         TestFramework.runWithFlags("-XX:-TieredCompilation",
                                    "-XX:+UseCMoveUnconditionally",
                                    "-XX:+UseVectorCmov",
-                                   "-XX:CompileCommand=compileonly,*.TestVectorConditionalMove.test*");
+                                   "-XX:CompileCommand=compileonly,compiler.c2.irTests.TestVectorConditionalMove::test*");
```
chhagedorn, line 49: `compileonly` is probably not necessary but you can leave it as such.

chhagedorn (approved):
Looks good, thanks for cleaning this up!

I think it's a good idea to also check the other usages of `ttyLocker` separately.

Can you also file an RFE to remove the "safepoint while printing handling" in the IR framework? This is no longer needed with this patch but would exceed the scope of this fix. You can assign that to me.

test/hotspot/jtreg/compiler/c2/irTests/TestVectorConditionalMove.java
eme64: Also removed the `-XX:-TieredCompilation` :)

chhagedorn (approved):
Thanks for the update, looks good!

src/hotspot/share/opto/compile.cpp
```diff
@@ -551,28 +551,36 @@ void Compile::print_compile_messages() {
...
-  if (xtty != nullptr) {
-    xtty->head("ideal compile_id='%d'%s compile_phase='%s'",
-               compile_id(),
-               is_osr_compilation() ? " compile_kind='osr'" : "",
-               phase_name);
-  }
+
+  // Node dumping can cause a safepoint, which can break the tty lock.
+  // Buffer all node dumps, so that all safepoints happen before we lock.
+  stringStream ss;
```
TobiHartmann, line 561: Should we add a `ResourceMark rm;`?

src/hotspot/share/opto/output.cpp
```diff
@@ -2068,23 +2068,30 @@ void PhaseOutput::ScheduleAndBundle() {
 
 #ifndef PRODUCT
   if (C->trace_opto_output()) {
-    tty->print("\n---- After ScheduleAndBundle ----\n");
-    print_scheduling();
+    // Buffer and print all at once
+    stringStream ss;
```
TobiHartmann, line 2072: Should we add a `ResourceMark rm;`?

TobiHartmann (approved):
Looks good to me otherwise.

> Maybe we should in general go through all uses of ttyLocker, and add a NoSafepointVerifier to ensure no such lock is broken? Or maybe we just build the verifier into the ttyLocker? Should I file an RFE?

Yes, I think it would be good to file an RFE to investigate the impact of adding a NoSafepointVerifier to the ttyLocker.

eme64:
Thanks @chhagedorn for all the discussions and help! And thanks @TobiHartmann for the review!
/integrate

PR #14593, opened 2023-06-21
JDK-8310550: Adjust references to rt.jar

MBaesken:
There are a few references to rt.jar in comments and in the codebase itself. Some of them might be removed or adjusted.

---------
### Progress

### Issue
 * [JDK-8310550](https://bugs.openjdk.org/browse/JDK-8310550): Adjust references to rt.jar (**Bug** - P4)

### Reviewers
 * [Erik Joelsson](https://openjdk.org/census#erikj) (@erikj79 - **Reviewer**) ⚠️ Review applies to [5d52b4cb](https://git.openjdk.org/jdk/pull/14593/files/5d52b4cb1aad2f3e40c20654037605cd86c42ec7)
 * [Christoph Langer](https://openjdk.org/census#clanger) (@RealCLanger - **Reviewer**) ⚠️ Review applies to [3a7b057a](https://git.openjdk.org/jdk/pull/14593/files/3a7b057a5569e62d56f5fd637fd0e69c879a9aa0)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14593/head:pull/14593` \
`$ git checkout pull/14593`

`$ git checkout pull/14593` \
`$ git pull https://git.openjdk.org/jdk.git pull/14593/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14593`

`$ git pr show -t 14593`

</details>

<a href="https://git.openjdk.org/jdk/pull/14593.diff">https://git.openjdk.org/jdk/pull/14593.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14593#issuecomment-1601054002)

### DIFF
--- a/src/java.sql/share/classes/java/sql/DriverManager.java
+++ b/src/java.sql/share/classes/java/sql/DriverManager.java
@@ -36,10 +36,10 @@ import java.security.PrivilegedAction;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.stream.Stream;
 
+import jdk.internal.misc.VM;
 import jdk.internal.reflect.CallerSensitive;
 import jdk.internal.reflect.CallerSensitiveAdapter;
 import jdk.internal.reflect.Reflection;
-
 
 /**
  * The basic service for managing a set of JDBC drivers.
@@ -652,10 +652,9 @@ public class DriverManager {
     private static Connection getConnection(
         String url, java.util.Properties info, Class<?> caller) throws SQLException {
         /*
-         * When callerCl is null, we should check the application's
-         * (which is invoking this class indirectly)
-         * classloader, so that the JDBC driver class outside the image
-         * can be loaded from here.
+         * If the caller is defined to the bootstrap or platform class loader then use
+         * the Thread CCL as the initiating class loader so that a JDBC on the class path,
+         * or bundled with an application, is found.
          */
         ClassLoader callerCL = caller != null ? caller.getClassLoader() : null;
         if (callerCL == null || callerCL == ClassLoader.getPlatformClassLoader()) {
--- a/src/java.sql/share/classes/java/sql/DriverManager.java
+++ b/src/java.sql/share/classes/java/sql/DriverManager.java
@@ -36,10 +36,10 @@ import java.security.PrivilegedAction;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.stream.Stream;
 
-import jdk.internal.misc.VM;
 import jdk.internal.reflect.CallerSensitive;
 import jdk.internal.reflect.CallerSensitiveAdapter;
 import jdk.internal.reflect.Reflection;
+
 
 /**
  * The basic service for managing a set of JDBC drivers.
--- a/src/java.sql/share/classes/java/sql/DriverManager.java
+++ b/src/java.sql/share/classes/java/sql/DriverManager.java
@@ -654,7 +654,7 @@ public class DriverManager {
         /*
          * When callerCl is null, we should check the application's
          * (which is invoking this class indirectly)
-         * classloader, so that the JDBC driver class outside rt.jar
+         * classloader, so that the JDBC driver class outside the image
          * can be loaded from here.
          */
         ClassLoader callerCL = caller != null ? caller.getClassLoader() : null;
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/file/JavacFileManager.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/file/JavacFileManager.java
@@ -193,7 +193,7 @@ public class JavacFileManager extends BaseFileManager implements StandardJavaFil
     }
 
     /**
-     * Set whether or not to use ct.sym as an alternate
+     * Set whether or not to use ct.sym as an alternate to the current runtime
      */
     public void setSymbolFileEnabled(boolean b) {
         symbolFileEnabled = b;
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/file/JavacFileManager.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/file/JavacFileManager.java
@@ -193,7 +193,7 @@ public class JavacFileManager extends BaseFileManager implements StandardJavaFil
     }
 
     /**
-     * Set whether or not to use ct.sym as an alternate to the current runtime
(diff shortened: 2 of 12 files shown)


erikj79 (approved):
The update to Java.gmk is good.

/reviewers 2

src/jdk.compiler/share/classes/com/sun/tools/javac/file/JavacFileManager.java
```diff
@@ -193,7 +193,7 @@ private Path getPath(String first, String... more) {
     }
 
     /**
-     * Set whether or not to use ct.sym as an alternate to rt.jar.
+     * Set whether or not to use ct.sym as an alternate
```
dholmes-ora, line 196: As an alternate to what? This needs something else.

test/langtools/tools/javap/4798312/JavapShouldLoadClassesFromRTJarTest.java
dholmes-ora, line 1: The name of this test includes RTJar. It needs to be changed too I think. Does this test actually still test something?

dholmes-ora (changes requested):
Mostly seems okay - a couple of things need further adjusting I think.

Thanks.

src/jdk.compiler/share/classes/com/sun/tools/javac/file/JavacFileManager.java
MBaesken: should "to the image modules files"  be used instead ?

test/langtools/tools/javap/4798312/JavapShouldLoadClassesFromRTJarTest.java
MBaesken: It seems to start a javap. So I think it tests something, how important this is and what other tests might cover similar stuff, I do not know unfortunately .

src/java.sql/share/classes/java/sql/DriverManager.java
```diff
@@ -654,7 +654,7 @@ private static Connection getConnection(
         /*
          * When callerCl is null, we should check the application's
          * (which is invoking this class indirectly)
-         * classloader, so that the JDBC driver class outside rt.jar
+         * classloader, so that the JDBC driver class outside the image
          * can be loaded from here.
```
AlanBateman, line 658: This code should probably be changed to use VM.isSystemDomainLoader(callerCL).

I think the comment should be replaced because it doesn't match what it actually does and it's nothing to do with the whether the JDBC driver is in the run-time image or not. How about:

"If the caller is defined to the bootstrap or platform class loader then use the Thread CCL as the initiating class loader so that a JDBC on the class path, or bundled with an application, is found."

MBaesken: Hi Alan, regarding usage of class VM I get 
'package jdk.internal.misc is declared in module java.base, which does not export it to module java.sql'
Is there any concern to export it as well to module java.sql  ?
And btw did you mean to use it like this, in the if  ?

`        if (callerCL == null || VM.isSystemDomainLoader(callerCL)) {
            callerCL = Thread.currentThread().getContextClassLoader();
        }`

AlanBateman: > Hi Alan, regarding usage of class VM I get 'package jdk.internal.misc is declared in module java.base, which does not export it to module java.sql' Is there any concern to export it as well to module java.sql ? And btw did you mean to use it like this, in the if ?
> 
> `if (callerCL == null || VM.isSystemDomainLoader(callerCL)) { callerCL = Thread.currentThread().getContextClassLoader(); }`

It was just a passing comment, I didn't meant to suggest changing it as part of this PR. We have always think twice before adding qualified exports from java.base and this is case where java.sql is very "non-core", we don't want to give it any access to java.base internals.

MBaesken: Hi Alan, thanks for clarifying.  So I should only adjust the comment, correct ?

test/hotspot/jtreg/vmTestbase/nsk/jvmti/AttachOnDemand/attach024/TestDescription.java
```diff
@@ -37,7 +37,7 @@
  *     Test checks that agent's JAR file is appended at the end of the system class path.
  *     Agent's JAR file contains modified class java.util.TooManyListenersException (it is assumed
  *     that this class isn't loaded before agent is loaded), agent instantiates TooManyListenersException
- *     and checks that non-modified version of this class was loaded from rt.jar (not from agent's JAR).
+ *     and checks that non-modified version of this class was loaded from jdk image (not from agent's JAR).
```
RealCLanger, line 40: "from the jdk image"

test/langtools/tools/javap/4798312/JavapShouldLoadClassesFromRTJarTest.java
```diff
@@ -24,7 +24,7 @@
 /*
  * @test
  * @bug 4798312
- * @summary In Windows, javap doesn't load classes from rt.jar
+ * @summary In Windows, javap doesn't load classes from image
```
RealCLanger, line 27: "... from the runtime image"

RealCLanger (changes requested):
Looks good overall. I made a few suggestions.

MBaesken:
Hi  Christoph, thanks for the suggestions, I added some changes.

RealCLanger (approved):
Fine from my end now. Just one minor nit left. 😄

MBaesken:
Hi Christoph, thanks for the review !
I added the '.' as suggested.

Any objections to the latest revision?

test/langtools/tools/javap/4798312/JavapShouldLoadClassesFromRTJarTest.java
dholmes-ora: This is a trivial test for a trivial issue. javap will be tested much more thoroughly by other tests. I think this test can be deleted without any loss of coverage. Or it can just be left.

src/java.sql/share/classes/java/sql/DriverManager.java
AlanBateman: > Hi Alan, I adjusted the comment in DriverManager.java .

Thanks, the update looks okay.

PR #14595, opened 2023-06-21
8310428: Add CollectedHeap::reserved_range() API to return reserved heap memory range

ashu-mehra:
Please review this patch to add a new API `CollectedHeap::reserved_range()`. The changes are extracted out from https://github.com/openjdk/jdk/pull/14520 as per the suggestion [here](https://github.com/openjdk/jdk/pull/14520#discussion_r1234991165). Support for ZGC is not implemented yet as it can have discontiguous heap regions and unlike other collectors, it does not set `CollectedHeap::_reserved`.

---------
### Progress

### Issue
 * [JDK-8310428](https://bugs.openjdk.org/browse/JDK-8310428): Add CollectedHeap::reserved_range() API to return reserved heap memory range (**Sub-task** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14595/head:pull/14595` \
`$ git checkout pull/14595`

`$ git checkout pull/14595` \
`$ git pull https://git.openjdk.org/jdk.git pull/14595/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14595`

`$ git pr show -t 14595`

</details>

<a href="https://git.openjdk.org/jdk/pull/14595.diff">https://git.openjdk.org/jdk/pull/14595.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14595#issuecomment-1601350272)

### DIFF
--- a/src/hotspot/share/gc/shared/collectedHeap.hpp
+++ b/src/hotspot/share/gc/shared/collectedHeap.hpp
@@ -25,6 +25,7 @@
 #ifndef SHARE_GC_SHARED_COLLECTEDHEAP_HPP
 #define SHARE_GC_SHARED_COLLECTEDHEAP_HPP
 
+#include "gc/shared/gc_globals.hpp"
 #include "gc/shared/gcCause.hpp"
 #include "gc/shared/gcWhen.hpp"
 #include "gc/shared/verifyOption.hpp"
@@ -32,6 +33,7 @@
 #include "memory/metaspace.hpp"
 #include "memory/universe.hpp"
 #include "oops/stackChunkOop.hpp"
+#include "runtime/globals.hpp"
 #include "runtime/handles.hpp"
 #include "runtime/perfDataTypes.hpp"
 #include "runtime/safepoint.hpp"
@@ -207,6 +209,10 @@ class CollectedHeap : public CHeapObj<mtGC> {
   }
 
  public:
+  MemRegion reserved_region() const {
+    assert(!UseZGC, "Not supported for ZGC");
+    return _reserved;
+  }
 
   static inline size_t filler_array_max_size() {
     return _filler_array_max_size;
--- a/src/hotspot/share/cds/archiveHeapWriter.cpp
+++ b/src/hotspot/share/cds/archiveHeapWriter.cpp
@@ -333,7 +333,7 @@ size_t ArchiveHeapWriter::copy_one_source_obj_to_buffer(oop src_obj) {
 void ArchiveHeapWriter::set_requested_address(ArchiveHeapInfo* info) {
   assert(!info->is_used(), "only set once");
   assert(UseG1GC, "must be");
-  address heap_end = (address)G1CollectedHeap::heap()->reserved().end();
+  address heap_end = (address)G1CollectedHeap::heap()->reserved_region().end();
   log_info(cds, heap)("Heap end = %p", heap_end);
 
   size_t heap_region_byte_size = _buffer_used;
--- a/src/hotspot/share/cds/filemap.cpp
+++ b/src/hotspot/share/cds/filemap.cpp
@@ -212,8 +212,8 @@ void FileMapHeader::populate(FileMapInfo *info, size_t core_region_alignment,
       _heap_end = CompressedOops::end();
     } else {
 #if INCLUDE_G1GC
-      address start = (address)G1CollectedHeap::heap()->reserved().start();
-      address end = (address)G1CollectedHeap::heap()->reserved().end();
+      address start = (address)G1CollectedHeap::heap()->reserved_region().start();
+      address end = (address)G1CollectedHeap::heap()->reserved_region().end();
       _heap_begin = HeapShared::to_requested_address(start);
       _heap_end = HeapShared::to_requested_address(end);
 #endif
@@ -1570,7 +1570,7 @@ void FileMapInfo::write_region(int region, char* base, size_t size,
       assert((mapping_offset >> CompressedOops::shift()) << CompressedOops::shift() == mapping_offset, "must be");
     } else {
 #if INCLUDE_G1GC
-      mapping_offset = requested_base - (char*)G1CollectedHeap::heap()->reserved().start();
+      mapping_offset = requested_base - (char*)G1CollectedHeap::heap()->reserved_region().start();
 #endif
     }
 #endif // INCLUDE_CDS_JAVA_HEAP
@@ -2033,9 +2033,9 @@ bool FileMapInfo::can_use_heap_region() {
                 CompressedOops::mode(), p2i(CompressedOops::base()), CompressedOops::shift());
   log_info(cds)("    heap range = [" PTR_FORMAT " - "  PTR_FORMAT "]",
                 UseCompressedOops ? p2i(CompressedOops::begin()) :
-                                    UseG1GC ? p2i((address)G1CollectedHeap::heap()->reserved().start()) : 0L,
+                                    UseG1GC ? p2i((address)G1CollectedHeap::heap()->reserved_region().start()) : 0L,
                 UseCompressedOops ? p2i(CompressedOops::end()) :
-                                    UseG1GC ? p2i((address)G1CollectedHeap::heap()->reserved().end()) : 0L);
+                                    UseG1GC ? p2i((address)G1CollectedHeap::heap()->reserved_region().end()) : 0L);
 
   if (narrow_klass_base() != CompressedKlassPointers::base() ||
       narrow_klass_shift() != CompressedKlassPointers::shift()) {
@@ -2094,7 +2094,7 @@ bool FileMapInfo::map_heap_region() {
 
     // Make sure we map at the very top of the heap - see comments in
     // init_heap_region_relocation().
-    MemRegion heap_range = G1CollectedHeap::heap()->reserved();
+    MemRegion heap_range = G1CollectedHeap::heap()->reserved_region();
     assert(heap_range.contains(_mapped_heap_memregion), "must be");
 
     address heap_end = (address)heap_range.end();
--- a/src/hotspot/share/cds/heapShared.cpp
+++ b/src/hotspot/share/cds/heapShared.cpp
@@ -532,9 +532,9 @@ void HeapShared::archive_objects(ArchiveHeapInfo *heap_info) {
 
     log_info(cds)("Heap range = [" PTR_FORMAT " - "  PTR_FORMAT "]",
                    UseCompressedOops ? p2i(CompressedOops::begin()) :
-                                       p2i((address)G1CollectedHeap::heap()->reserved().start()),
+                                       p2i((address)G1CollectedHeap::heap()->reserved_region().start()),
                    UseCompressedOops ? p2i(CompressedOops::end()) :
-                                       p2i((address)G1CollectedHeap::heap()->reserved().end()));
+                                       p2i((address)G1CollectedHeap::heap()->reserved_region().end()));
     copy_objects();
 
     CDSHeapVerifier::verify();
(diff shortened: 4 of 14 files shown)


src/hotspot/share/gc/shared/collectedHeap.hpp
```diff
@@ -207,6 +209,10 @@ class CollectedHeap : public CHeapObj<mtGC> {
   }
 
  public:
+  MemRegion reserved_region() const {
+    assert(!UseZGC, "Not supported for ZGC");
+    return _reserved;
+  }
```
tstuefe, line 215: Instead of asserting, would it be better to return an empty null range, and document such that "if null range is returned, this method is unsupported"? That way we don't need ZGC specific conditionals in shared code.

tstuefe (commented):
Thank you for doing this. 

Would it be possible to add a simple gtest for this API (something simple, as in the result of this function should correspond to MaxHeapSize)?

ashu-mehra:
Sorry, I had to force push a commit to update the commit message to include the issue number.

stefank (changes requested):
This change completely undoes the removal of CollectedHeap::reserved_region() we did in:
8224815: Remove non-GC uses of CollectedHeap::is_in_reserved()

We removed it because it is a false abstraction that makes it easy for people to accidentally start to use this API without realizing that it will not work for ZGC.

I'd prefer to see a plan on how to implement heap archiving for ZGC before we decide on making abstractions in CollectedHeap that doesn't work for ZGC. I know that @fisk has had earlier discussions with  @iklam about other ways to make the heap archiving GC-agnostic. In #14520 we saw that another approach was taken and @fisk stepped in and voiced again that we have concerns about this approach. Given that we haven't completed that discussion I'd like to pause this PR until we have come to an agreement on how to proceed.

fisk:
Yeah I strongly don't think shared code should assume heap memory is contiguous. Even less so when that assumption isn't true for all collectors. It seems like taking a shortcut, instead of building a more general solution.

tstuefe:
Hi @stefank and @fisk,

I still opt for an abstraction like this, couples with an added "heap_is_contiguous()". Its just a fact that for the majority of collectors this abstraction holds.

Let me give you a specific example. 

I am in the process of reworking metaspace and nKlass encoding for Lilliput. One of the things I always disliked a lot was the mechanism by which we "allocate class space adjacent to heap, but only if UseCompressedOops and heap is allocated in the lower n GB range". So, while I'm on it, I'll change that.

What we really wanted here was zero-based nKlass mode, which has nothing to do whatsoever with the heap. But if someone went to the trouble of allocating the heap in lower address ranges, we just piggyback on that and allocate class space adjacent to it.

That approach is flawed for a few reasons I won't detail here. But it illuminates one problem: any improved process I am thinking about (for example, scanning `/proc/self/maps` or `VirtualQuery()` for address space holes or the typical sequential-mapping-attempts approach) in practice benefits from knowing where the heap is located. Simply because it happens to be the largest known contiguous reserved address range at that point in time. Since scanning procfs or mmaping are expensive operations, it pays to avoid this address range since we know mapping attempt will fail.

In this example, the current coding mis-uses CompressedOops::encoding_range() as a (flawed) proxy for heap range. So, in practice, we already use the concept of heap range.

ZGC is a great collector, but a bit like our exotic platforms; there is a tension between keeping code super-portable and getting the job done on other platforms. With platforms, we were always good at finding compromises. For me, a "reserved_region()" coupled with "is_contiguous()" would be such a reasonable workaround.

iklam:
For CDS, we actually don't need to know the exact range of the heap.

Currently, archive heap dumping is implemented only for G1. In the future, we want to be able to dump with any collector.

The eventual goal is - if you dump with collector X, then the layout of the archive heap should be optimal for this collector (e.g., avoid runtime relocation). The archive heap created by X should still be loadable by other collectors (but relocation is more likely).

So, for people who are really sensitive to start-up, they can create a custom archive with the specific GC that their app uses.

Currently, when dumping the archive heap with G1, we put it at the very top of the heap, which is the best location for G1 as this can usually avoid relocation. We determine the range by getting the entire G1 heap range, and take a portion of that from the top end.

To support all collectors, I think we can move the range calculation into each collector, and have a new API like

```
CollectorHeap::get_archive_heap_preferred_base(size_t required_bytes, size_t alignment);
```
The collector should return the range according to its policy. E.g., for SerialGC, it should probably return the bottom of the heap. The "reserve at heap top" logic can be moved into G1.

fisk:
> For CDS, we actually don't need to know the exact range of the heap.
> 
> 
> 
> Currently, archive heap dumping is implemented only for G1. In the future, we want to be able to dump with any collector.
> 
> 
> 
> The eventual goal is - if you dump with collector X, then the layout of the archive heap should be optimal for this collector (e.g., avoid runtime relocation). The archive heap created by X should still be loadable by other collectors (but relocation is more likely).
> 
> 
> 
> So, for people who are really sensitive to start-up, they can create a custom archive with the specific GC that their app uses.
> 
> 
> 
> Currently, when dumping the archive heap with G1, we put it at the very top of the heap, which is the best location for G1 as this can usually avoid relocation. We determine the range by getting the entire G1 heap range, and take a portion of that from the top end.
> 
> 
> 
> To support all collectors, I think we can move the range calculation into each collector, and have a new API like
> 
> 
> 
> ```
> 
> CollectorHeap::get_archive_heap_preferred_base(size_t required_bytes, size_t alignment);
> 
> ```
> 
> The collector should return the range according to its policy. E.g., for SerialGC, it should probably return the bottom of the heap. The "reserve at heap top" logic can be moved into G1.
> 

Right - each GC might have a specialized strategy that is optimal for that GC. A good ZGC strategy would be to materialize objects in the heap lazily and concurrently using load barriers. That way materializing objects wouldn't stall the program.

However, I believe we can have a common baseline strategy that is completely GC agnostic, which just materializes objects in the archive using normal allocations, and translates the in-archive pointers to real pointers using normal stores, based on the oops we got from the materializing allocations, using a GC agnostic translation mechanism.

I think there are interesting ways of optimizing such a shared translation strategy by exploiting some "mostly contiguous" property that all collectors have, combined with locality properties that tend to be quite universal (most pointers don't point very far away from its object). In fact, I suspect an optimized GC agnostic mechanism could be fast enough, such that making specialized solutions for collector X might not be worth it. 

I think starting with a GC agnostic solution and then checking if specific collectors want to optimize further by handing their guts out to CDS, makes more sense than the opposite order. So maybe we could try doing that first?

iklam:
> > For CDS, we actually don't need to know the exact range of the heap.
> > Currently, archive heap dumping is implemented only for G1. In the future, we want to be able to dump with any collector.
> > The eventual goal is - if you dump with collector X, then the layout of the archive heap should be optimal for this collector (e.g., avoid runtime relocation). The archive heap created by X should still be loadable by other collectors (but relocation is more likely).
> > So, for people who are really sensitive to start-up, they can create a custom archive with the specific GC that their app uses.
> > Currently, when dumping the archive heap with G1, we put it at the very top of the heap, which is the best location for G1 as this can usually avoid relocation. We determine the range by getting the entire G1 heap range, and take a portion of that from the top end.
> > To support all collectors, I think we can move the range calculation into each collector, and have a new API like
> > ```
> > 
> > CollectorHeap::get_archive_heap_preferred_base(size_t required_bytes, size_t alignment);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The collector should return the range according to its policy. E.g., for SerialGC, it should probably return the bottom of the heap. The "reserve at heap top" logic can be moved into G1.
> 
> Right - each GC might have a specialized strategy that is optimal for that GC. A good ZGC strategy would be to materialize objects in the heap lazily and concurrently using load barriers. That way materializing objects wouldn't stall the program.
> 
> However, I believe we can have a common baseline strategy that is completely GC agnostic, which just materializes objects in the archive using normal allocations, and translates the in-archive pointers to real pointers using normal stores, based on the oops we got from the materializing allocations, using a GC agnostic translation mechanism.
> 
> I think there are interesting ways of optimizing such a shared translation strategy by exploiting some "mostly contiguous" property that all collectors have, combined with locality properties that tend to be quite universal (most pointers don't point very far away from its object). In fact, I suspect an optimized GC agnostic mechanism could be fast enough, such that making specialized solutions for collector X might not be worth it.
> 
> I think starting with a GC agnostic solution and then checking if specific collectors want to optimize further by handing their guts out to CDS, makes more sense than the opposite order. So maybe we could try doing that first?

I've had an off-line discussion with @fisk and created an investigative RFE based on his ideas. Please see https://github.com/openjdk/jdk/pull/14520 and use that PR for further discussion on this topic.

tstuefe:
> I've had an off-line discussion with @fisk and created an investigative RFE based on his ideas. Please see #14520 and use that PR for further discussion on this topic.

Please let's have the discussion in the open.

Nobody answered me, but I repeat, there is merit in knowing the contiguous region for collectors that have a contiguous reserved region. I don't understand why we have to treat developers like children and force them to use really bad workarounds.

Just an example of the weird complexity this API omission causes. See class space reservation, where we try to reserve class space adjacent to the java heap: https://github.com/openjdk/jdk/blob/bfcca5eff96ac3cd72996b6c4865872c2da4de53/src/hotspot/share/memory/metaspace.cpp#L794-L796.

1) we don't know the address range of the heap
2) so we have to misuse CompressedOops::encoding_range, which is a bad fit. Encoding range is not heap range. It can (often does) start before the heap, so we don't know the heap start address. It only tells us the heap range end, but that is an accident of implementation: semantically, it is incorrect since the encoding range end is not the end of the heap.
3) Since we don't get the start address, we only know heap range end. That forces us to reserve the class space adjacent - following - the heap range.
4) So, for zero-based heap reservation, we have to leave a (potentially very large - 1G) gap between zero-based encoding range end and the heap end: https://github.com/openjdk/jdk/blob/bfcca5eff96ac3cd72996b6c4865872c2da4de53/src/hotspot/share/memory/virtualspace.cpp#L548-L558 This is bad, since it reduces the chance of getting zero-based heap. It also means we need to code intimate knowledge about class space (e.g. class space size) at heap reservation. Heap reservation should not have to know these things.
5) if, instead, we had known the start address of the heap at (4), we could just reserve the class space in front to it. There would have to be no need for leaving a large gap between zero-based compressed oops encoding range end and heap end. Also, we would not need to have class-space knowledge at heap reservation.

And all this complexity for a scenario that is not even supported on ZGC (CompressedOops) !

This is suboptimal. With a simple "CollectedHeap::reserved_range()" in combination with "CollectedHeap::is_contiguous()", this could be solved in about 10 lines of code.

So,no, I don't think this discussion has come to a satisfying close.

ashu-mehra:
> Such an API would not need to know anything about the
heap but would benefit from knowing large reserved areas *like* the heap to
prevent unnecessary work. 

How about the api returning array of MemRange to handle discontiguous heap layouts? 
@tstuefe  would that work for your case?
@stefank is that possible to do with ZGC?

iklam:
> > I've had an off-line discussion with @fisk and created an investigative RFE based on his ideas. Please see #14520 and use that PR for further discussion on this topic.
> 
> Please let's have the discussion in the open.

Hi Thomas,

I just want to clarify that for the purpose of CDS, there's no need for the proposed `CollectedHeap::reserved_range()` API. If CDS needs to find out the optimal location for the archived objects, it should be done using a more specific API. Any CDS-related discussions should be moved to #14520.

If we need the `reserved_range()` API for other purposes, then of course we can continue the discussion here.

tstuefe:
> > Such an API would not need to know anything about the
> > heap but would benefit from knowing large reserved areas _like_ the heap to
> > prevent unnecessary work.
> 
> How about the api returning array of MemRange to handle discontiguous heap layouts? @tstuefe would that work for your case? @stefank is that possible to do with ZGC?

Hi @ashu-mehra, I'm afraid not. For collectors that maintain disjunct regions this could be costly (they would have to collect all ranges and return it in a data structure), while it is not clear the caller even needs it.

I'm not sure I can contribute much more here; sorry for causing you that work. I'll wait for Oracle's decision.

ashu-mehra:
ok, closing this based on the discussions so far.

PR #14597, opened 2023-06-21
8310494: Do not include constantPool.hpp from instanceKlass.hpp

iklam:
instanceKlass.hpp is needed quite frequently for HotSpot C++ sources, but constantPool.hpp is much less frequently.

This PR is the first step of reducing the inclusion of constantPool.hpp, by removing it from instanceKlass.hpp

I moved a bunch of inline functions into instanceKlass.cpp. These don't seem to be performance related.

---------
### Progress

### Issue
 * [JDK-8310494](https://bugs.openjdk.org/browse/JDK-8310494): Do not include constantPool.hpp from instanceKlass.hpp (**Enhancement** - P4)

### Reviewers
 * [Stefan Karlsson](https://openjdk.org/census#stefank) (@stefank - **Reviewer**) ⚠️ Review applies to [68697195](https://git.openjdk.org/jdk/pull/14597/files/686971955997b0b15b1cde60d6d69b85d321e126)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14597/head:pull/14597` \
`$ git checkout pull/14597`

`$ git checkout pull/14597` \
`$ git pull https://git.openjdk.org/jdk.git pull/14597/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14597`

`$ git pr show -t 14597`

</details>

<a href="https://git.openjdk.org/jdk/pull/14597.diff">https://git.openjdk.org/jdk/pull/14597.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14597#issuecomment-1601098899)


src/hotspot/share/classfile/javaClasses.hpp
```diff
@@ -28,6 +28,8 @@
 #include "classfile/vmClasses.hpp"
 #include "oops/instanceKlass.hpp"
 #include "oops/oopsHierarchy.hpp"
+#include "oops/symbol.hpp"
+#include "runtime/os.hpp"
```
stefank, line 32: Sort order issue

src/hotspot/share/oops/instanceKlass.hpp
```diff
@@ -997,7 +993,7 @@ class InstanceKlass: public Klass {
 
   // The constant pool is on stack if any of the methods are executing or
   // referenced by handles.
-  bool on_stack() const { return _constants->on_stack(); }
+  bool on_stack() const; // virtual
```
stefank, line 996:
```suggestion
  virtual bool on_stack() const;
```

stefank (commented):
I think this change looks fine (minus a couple of nits).

I do wonder if the code wouldn't be nicer if we simply removed these delegating functions in InstanceKlass and let the callers operate on the ConstantPool directly instead?

iklam:
> I do wonder if the code wouldn't be nicer if we simply removed these delegating functions in InstanceKlass and let the callers operate on the ConstantPool directly instead?

The version and signature logically belong to the class, so I think the functions should be part of InstanceKlass. The fact that they are stored inside the ConstantPool is just an implementation detail that can change. IIRC it was done this way because the ConstantPool can be stored in the read-only section of the CDS archive.

src/hotspot/share/oops/instanceKlass.hpp
```diff
@@ -997,7 +993,7 @@ class InstanceKlass: public Klass {
 
   // The constant pool is on stack if any of the methods are executing or
   // referenced by handles.
-  bool on_stack() const { return _constants->on_stack(); }
+  virtual bool on_stack() const;
```
stefank, line 996: The comment looks weird now that the implementation, which delegates to the constant pool, has been moved from the header. I think the comment needs to be updated.

stefank (approved):
Reviewed. Left some nits that you could consider.

src/hotspot/share/oops/instanceKlass.hpp
coleenp: Suggest move the comment to the implementation.

coleenp (approved):

iklam:
Thanks @coleenp and @stefank for the review.
/integrate

PR #14604, opened 2023-06-21
8310577: Fix -Wconversion warnings in interpreter and oops

coleenp:
Please review this change that's mostly casts and a few int type narrowing changes.  The casts are for size calculations between two addresses within the same area so can be int and are never going to be 64 bit.
I didn't change CompileThreshold to be 'int' because SA and JVMCI code expects it to be intx.  Maybe that should change though.
Tested with tier1 on Oracle platforms, and tier1-4 linux and windows x64.

---------
### Progress

### Issue
 * [JDK-8310577](https://bugs.openjdk.org/browse/JDK-8310577): Fix -Wconversion warnings in interpreter and oops (**Enhancement** - P4)

### Reviewers
 * [Frederic Parain](https://openjdk.org/census#fparain) (@fparain - **Reviewer**)
 * [Matias Saavedra Silva](https://openjdk.org/census#matsaave) (@matias9927 - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14604/head:pull/14604` \
`$ git checkout pull/14604`

`$ git checkout pull/14604` \
`$ git pull https://git.openjdk.org/jdk.git pull/14604/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14604`

`$ git pr show -t 14604`

</details>

<a href="https://git.openjdk.org/jdk/pull/14604.diff">https://git.openjdk.org/jdk/pull/14604.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14604#issuecomment-1601594087)

### DIFF
--- a/src/hotspot/share/interpreter/bytecodeUtils.cpp
+++ b/src/hotspot/share/interpreter/bytecodeUtils.cpp
@@ -353,7 +353,7 @@ static void print_local_var(outputStream *os, unsigned int bci, Method* method, 
 
 StackSlotAnalysisData::StackSlotAnalysisData(BasicType type) : _bci(INVALID), _type(type) {}
 
-StackSlotAnalysisData::StackSlotAnalysisData(int bci, BasicType type) : _bci(bci), _type(type) {
+StackSlotAnalysisData::StackSlotAnalysisData(int bci, BasicType type) : _bci((u2)bci), _type(type) {
   assert(bci >= 0, "BCI must be >= 0");
   assert(bci < 65536, "BCI must be < 65536");
 }
@@ -1081,7 +1081,7 @@ int ExceptionMessageBuilder::do_instruction(int bci) {
   }
 
   // If we have more than one branch target, process these too.
-  for (int64_t i = 0; i < dests.length(); ++i) {
+  for (int i = 0; i < dests.length(); ++i) {
     if (_stacks->at(dests.at(i)) == nullptr) {
       _added_one = true;
     }
--- a/src/hotspot/share/interpreter/bytecodes.cpp
+++ b/src/hotspot/share/interpreter/bytecodes.cpp
@@ -385,9 +385,9 @@ int Bytecodes::special_length_at(Bytecodes::Code code, address bcp, address end)
       if (end != nullptr && aligned_bcp + 3*jintSize >= end) {
         return -1; // don't read past end of code buffer
       }
-      jlong lo = (jint)Bytes::get_Java_u4(aligned_bcp + 1*jintSize);
-      jlong hi = (jint)Bytes::get_Java_u4(aligned_bcp + 2*jintSize);
-      jlong len = (aligned_bcp - bcp) + (3 + hi - lo + 1)*jintSize;
+      int lo = Bytes::get_Java_u4(aligned_bcp + 1*jintSize);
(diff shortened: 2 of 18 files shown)


src/hotspot/share/interpreter/invocationCounter.hpp
```diff
@@ -68,7 +68,7 @@ class InvocationCounter {
   // Accessors
   bool carry() const               { return (_counter & carry_mask) != 0; }
   uint count() const               { return _counter >> number_of_noncount_bits; }
-  uint limit() const               { return CompileThreshold; }
+  uint limit() const               { return (uint)CompileThreshold; }
```
fparain, line 71: CompileThreshold is an intx, and can be set by users on the command line. Casting it to uint looks wrong. This method seems to be used in only one place (invocationCounter.cpp:69), so changing the return type of the method to intx and fixing the print statement where it is used could be a better fix.

fparain (changes requested):
Overall, looks good. One fix could be better in invocationCounter.hpp.

src/hotspot/share/interpreter/invocationCounter.hpp
coleenp: That was a good suggestion.  Thank you!

fparain (approved):

src/hotspot/share/interpreter/bytecodeUtils.cpp
```diff
@@ -353,7 +353,7 @@ static void print_local_var(outputStream *os, unsigned int bci, Method* method,
 
 StackSlotAnalysisData::StackSlotAnalysisData(BasicType type) : _bci(INVALID), _type(type) {}
 
-StackSlotAnalysisData::StackSlotAnalysisData(int bci, BasicType type) : _bci(bci), _type(type) {
+StackSlotAnalysisData::StackSlotAnalysisData(int bci, BasicType type) : _bci((u2)bci), _type(type) {
```
matias9927, line 356: Looking at this more closely, it seems like bci is meant to be a u2 but it needs one extra bit for the "INVALID" value. Maybe it's possible to represent invalid bci's differently? That way we can just set _bci to be a u2.

src/hotspot/share/interpreter/bytecodes.cpp
```diff
@@ -439,7 +439,7 @@ void Bytecodes::def_flags(Code code, const char* format, const char* wide_format
   int len  = (format      != nullptr ? (int) strlen(format)      : 0);
   int wlen = (wide_format != nullptr ? (int) strlen(wide_format) : 0);
 #endif
-  int bc_flags = 0;
+  jchar bc_flags = 0;
   if (can_trap)           bc_flags |= _bc_can_trap;
```
matias9927, line 443: I believe that bitwise-or will promote the operands to int. Will this cause any issues here?

src/hotspot/share/interpreter/bytecodeUtils.cpp
coleenp: If I pass u2 bci, it'll fit in the field.  The INVALID bit is added after bci - ie the bci is never INVALID.

coleenp:
Thanks Fred and Matias.
/integrate

RealLucy:
Breaks the build, unfortunately. See JDK-8310735.

PR #14605, opened 2023-06-21
8309733: [macOS, Accessibility] VoiceOver: Incorrect announcements of JRadioButton

azuev-java:
Only pass value changed event to the native accessibility peer when the value is actually changed. While at fixing it for RadioButton also fixing it for checkbox and toggle button so there is no need to re-visit this file later.
Also updating the copyright year.

### DIFF
--- a/src/java.desktop/macosx/classes/sun/lwawt/macosx/CAccessible.java
+++ b/src/java.desktop/macosx/classes/sun/lwawt/macosx/CAccessible.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


kumarabhi006:
Verified the fix and it behaves as expected.

`Expected announcement: "Radio Two, radio button, 2 of 3. ..." `
As mentioned in JBS, should VO announce radiobutton's position also?

There is an another bug [JDK-8309732](https://bugs.openjdk.org/browse/JDK-8309732) for position issue.

kumarabhi006:
> > Should VO announce radiobutton's position also?
> 
> That is beyond the scope of this bug but overall i would look at the native applications with the same type of controls. If native controls do not demonstrate the same behavior then i would say that it is not a problem on our side.

I checked in native application (i.e. system settings -> appearance) where radio buttons are announced with it's position along with label.

kumarabhi006 (approved):

src/java.desktop/macosx/classes/sun/lwawt/macosx/CAccessible.java
```diff
@@ -182,17 +182,23 @@ public void propertyChange(PropertyChangeEvent e) {
 
                     // Do send check box state changes to native side
                     if (thisRole == AccessibleRole.CHECK_BOX) {
-                        valueChanged(ptr);
+                        if (newValue != null && !newValue.equals(oldValue)) {
```
mrserb, line 185: Is it possible to have "newValue == null" and the "oldValue != null"? Do we need to update the native in this case?

azuev-java: No. Null value is not a correct value to be set for these types of control so if null is passed then it is not a real value update event.

savoptik (commented):
Is it possible to add a regression test on these components?
to ```test/jdk/java/awt/a11y```

---

@NikitkoCent please review

prrace (approved):

azuev-java:
> Is it possible to add a regression test on these components? to `test/jdk/java/awt/a11y`

Not within this regression type fix unless i find a way to write an automatic regression test which i don't see a way of. As of creating a comprehensive test or set of tests that covers all of these components - that might be a separate task. But again it is outside of this fix scope.

src/java.desktop/macosx/classes/sun/lwawt/macosx/CAccessible.java
aivanov-jdk: [`Objects.equals`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Objects.html#equals(java.lang.Object,java.lang.Object)) can be used to avoid null-check, it handles `null` values gracefully.

mrserb (approved):

savoptik (approved):
LGTM.

PR #14607, opened 2023-06-21
8293972: runtime/NMT/NMTInitializationTest.java#default_long-off failed with "Suspiciously long bucket chains in lookup table."

gerard-ziemski:
We improve the hash function, by using Mersenne prime (2^x - 1) i.e. 8191 , as the table size, which when combined with "modulo" operation, provides a uniform bit distribution.

To help with testing this change to make sure we improve the hash function, I collected `NMTPreInitAllocation` data on 3 platforms: macOS, Linux and Windows, using 100 runs of `test/hotspot/jtreg/runtime/NMT/NMTInitializationTest.java` and computed `stdev` as well as average `max chain`:

```
mac OS data:
 BEFORE total: max chain: 676, empties: 210892, ranges: 676, std_devs: 112.771
  AFTER total: max chain: 622, empties: 230762, ranges: 622, std_devs: 111.094
 BEFORE avg: max chain: 5.061, empties: 18.695, ranges: 5.061, std_devs: 0.010
  AFTER avg: max chain: 4.181, empties: 16.970, ranges: 4.161, std_devs: 0.009
	
Linux data:
 BEFORE total: max chain: 523, empties: 193174, ranges: 523, std_devs: 103.243
  AFTER total: max chain: 432, empties: 175347, ranges: 430, std_devs: 90.670
 BEFORE avg: max chain: 6.540, empties: 20.402, ranges: 6.540, std_devs: 0.011
  AFTER avg: max chain: 6.017, empties: 22.324, ranges: 6.017, std_devs: 0.011

Windows data:
 BEFORE total: max chain: 578, empties: 201224, ranges: 578, std_devs: 108.988
  AFTER total: max chain: 491, empties: 218653, ranges: 491, std_devs: 104.879
 BEFORE avg: max chain: 5.602, empties: 19.504, ranges: 5.602, std_devs: 0.011
  AFTER avg: max chain: 4.759, empties: 21.193, ranges: 4.759, std_devs: 0.010
```

Note: the lower the number, the better.
Note: the `stdedev` and `max chain` are the most important numbers here.

Tested manually with `jtreg -nr -va -jdk:./build/macosx-x86_64-server-fastdebug/images/jdk test/hotspot/jtreg/runtime/NMT` and `gtestLauncher -jdk ./build/macosx-x86_64-server-fastdebug/images/jdk --gtest_output=xml:test_result.xml --gtest_catch_exceptions=0 --gtest_filter="NMT*:os*" -XX:NativeMemoryTracking=summary` and `Mach5 hs-tier1,2,3,4,5`

---------
### Progress

### Issue
 * [JDK-8293972](https://bugs.openjdk.org/browse/JDK-8293972): runtime/NMT/NMTInitializationTest.java#default_long-off failed with "Suspiciously long bucket chains in lookup table." (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14607/head:pull/14607` \
`$ git checkout pull/14607`

`$ git checkout pull/14607` \
`$ git pull https://git.openjdk.org/jdk.git pull/14607/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14607`

`$ git pr show -t 14607`

</details>

<a href="https://git.openjdk.org/jdk/pull/14607.diff">https://git.openjdk.org/jdk/pull/14607.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14607#issuecomment-1601798684)


dholmes-ora:
> We also, mark output with PID to help identify which test process it is coming from as the attached log was very confusing to read.

Isn't there only a single process created per test?

dholmes-ora:
This is failing in GHA:
```
 assert(false) failed: NMT preinit lookup table degenerated (7/14 chains longer than 30)
```

gerard-ziemski:
> This is failing in GHA:
> 
> ```
>  assert(false) failed: NMT preinit lookup table degenerated (7/14 chains longer than 30)
> ```

Investigating...

gerard-ziemski:
> > This is failing in GHA:
> > ```
> >  assert(false) failed: NMT preinit lookup table degenerated (7/14 chains longer than 30)
> > ```
> 
> Investigating...

I used `(2^13 - 1)` notation in the code to be more explicit on how the value is constructed, but `^` is XOR operator, not `pow`. I will fix this:

`  static const int table_size = 8191; // i.e. 8191==(2^13 - 1);
`

src/hotspot/share/services/nmtPreInit.hpp
```diff
@@ -143,22 +143,25 @@ class NMTPreInitAllocationTable {
...
   const index_t invalid_index = -1;
 
   static unsigned calculate_hash(const void* p) {
-    uintptr_t tmp = p2i(p);
-    unsigned hash = (unsigned)tmp
-                     LP64_ONLY( ^ (unsigned)(tmp >> 32));
-    return hash;
+    // Keep hash function simple, the modulo
+    // operation in index function will do the "heavy lifting".
+    return (unsigned)p2i(p);
```
tstuefe, line 159: How can the modulo do the "heavy lifting" if you shear off half of the value space? (well, about a quarter to a third, since most kernels have 48..52 bit addresses.

In fact, I would go the other direction and improve this hash like this:
return (tmp >> 3) ^ (tmp >> 35)

since the lower 3 bits are always 0 for malloc pointers.

tstuefe (commented):
Looking back at this, I wonder whether we should reduce the table size. 64k for this lookup table is pretty big. The MallocSizeTable itself, that arguably holds a lot more allocations, is just have as wide.

src/hotspot/share/services/nmtPreInit.hpp
gerard-ziemski: > How can the modulo do the "heavy lifting" if you shear off half of the value space? (well, about a quarter to a third, since most kernels have 48..52 bit addresses.
> 
> In fact, I would go the other direction and improve this hash like this: return (tmp >> 3) ^ (tmp >> 35)
> 
> since the lower 3 bits are always 0 for malloc pointers.

I ported the code wrong from my sandbox. The hash function should be dealing with 64 bits, not 32 bits:

```
  static unsigned calculate_hash(const void* p) {
    // Keep hash function simple, the modulo
    // operation in index function will do the "heavy lifting".
    return (unsigned)p2i(p);
  }

  static index_t index_for_key(const void* p) {
    const uint64_t hash = calculate_hash(p);
    // "table_size" is a Mersenne prime, so "modulo" is all we need here.
    return hash % table_size;
  }
```

The problem with your approach is that you are trying to extract "good" bits, by hand. I ran multiple experiments trying to find "uniform" bits by hand and it was hard. I never gotten a better hash compared to using full 64 bits, and using modulo operation (with Mersenne prime table size) to mix those 64 bits for us.

I will try your proposed hash function and report bak here.

Good catch about the 64 to 32 bit clipping - thank you.

gerard-ziemski:
Here are the results with Thomas proposed hash function `(tmp >> 3) ^ (tmp >> 35)`:

```
Mac OS data:

original total: max chain: 523, empties: 193174, ranges: 523, std_devs: 103.243
           avg: max chain: 5.061, empties: 18.695, ranges: 5.061, std_devs: 0.010

  gerard total: max chain: 432, empties: 175347, ranges: 430, std_devs: 90.670
           avg: max chain: 4.181, empties: 16.970, ranges: 4.161, std_devs: 0.009

 Thomas1 total: max chain: 485, empties: 207085, ranges: 484, std_devs: 99.429
           avg: max chain: 4.694, empties: 20.041, ranges: 4.684, std_devs: 0.010

 Thomas2 total: max chain: 520, empties: 192194, ranges: 520, std_devs: 102.801
           avg: max chain: 5.032, empties: 18.600, ranges: 5.032, std_devs: 0.010


Linux OS data:

original total: max chain: 676, empties: 210892, ranges: 676, std_devs: 112.771
           avg: max chain: 6.540, empties: 20.402, ranges: 6.540, std_devs: 0.011

  gerard total: max chain: 622, empties: 230762, ranges: 622, std_devs: 111.094
           avg: max chain: 6.017, empties: 22.324, ranges: 6.017, std_devs: 0.011

 Thomas1 total: max chain: 656, empties: 225850, ranges: 656, std_devs: 110.384
           avg: max chain: 6.346, empties: 21.849, ranges: 6.346, std_devs: 0.011

 Thomas2 total: max chain: 680, empties: 210918, ranges: 680, std_devs: 112.780
           avg: max chain: 6.578, empties: 20.404, ranges: 6.578, std_devs: 0.011


Windows OS data:

original total: max chain: 578, empties: 201224, ranges: 578, std_devs: 108.988
           avg: max chain: 5.602, empties: 19.504, ranges: 5.602, std_devs: 0.011

  gerard total: max chain: 491, empties: 218653, ranges: 491, std_devs: 104.879
           avg: max chain: 4.759, empties: 21.193, ranges: 4.759, std_devs: 0.010

 Thomas1 total: max chain: 565, empties: 218421, ranges: 565, std_devs: 107.261
           avg: max chain: 5.476, empties: 21.171, ranges: 5.476, std_devs: 0.010

 Thomas2 total: max chain: 578, empties: 201224, ranges: 578, std_devs: 108.987
           avg: max chain: 5.602, empties: 19.504, ranges: 5.602, std_devs: 0.011

Note:
 original table_size 7919
 gerard table_size 8191
 Thomas1 table_size 8191
 Thomas2 table_size 7919
```

Summary:

- at the table size of 8191 which is Mersenne Prime (Thomas1), it is less than 1% better in stddev for Linux pointers, but quite a bit worse everywhere else (including max chain length for Linux)
- at the table size of 7919 (Thomas2), it is virtually the same as the original

So we don't gain anything by using clever bit shifts and xors (I know I tried many myself) compared to the "naked" pointer - the opposite in fact.

gerard-ziemski:
Uploaded the src code to the sandbox + data sets, which I used for comparing hash functions to the bug issue https://bugs.openjdk.org/browse/JDK-8293972

gerard-ziemski:
> Looking back at this, I wonder whether we should reduce the table size. 64k for this lookup table is pretty big. The MallocSizeTable itself, that arguably holds a lot more allocations, is just have as wide.

Tested with a smaller table size, about half of 8191:

```
Using original table size and Marsenne prime:

macOS data set:
  avg1 - max chain: 5.129, empties: 1897.891, ranges: 4.139, std_devs: 0.813
  avg2 - max chain: 4.228, empties: 1712.020, ranges: 3.238, std_devs: 0.686
  avg3 - max chain: 4.743, empties: 2030.426, ranges: 3.752, std_devs: 0.762

Linux data set:
  avg1 - max chain: 6.624, empties: 2067.396, ranges: 5.634, std_devs: 0.921
  avg2 - max chain: 6.109, empties: 2263.485, ranges: 5.119, std_devs: 0.899
  avg3 - max chain: 6.436, empties: 2213.792, ranges: 5.446, std_devs: 0.897

Windows data set:
  avg1 - max chain: 5.653, empties: 1970.208, ranges: 4.663, std_devs: 0.883
  avg2 - max chain: 4.802, empties: 2143.931, ranges: 3.812, std_devs: 0.822
  avg3 - max chain: 5.525, empties: 2136.158, ranges: 4.535, std_devs: 0.862

Note1: avg1 - the original code, table size of 7919
Note2: avg2 - gerard's proposed hash function, table size of 8191
Note3: avg3 - Thomas'es proposed hash function, table size of 8191
Note4: elapsed time: 46.5 sec
Note5: the smaller the numbers the better


Using smaller table size of 4093 for all cases:

macOS data set:
  avg1 - max chain: 6.871, empties: 318.000, ranges: 5.871, std_devs: 1.207
  avg2 - max chain: 5.881, empties: 219.634, ranges: 4.851, std_devs: 1.002
  avg3 - max chain: 6.891, empties: 308.396, ranges: 5.891, std_devs: 1.194

Linux data set:
  avg1 - max chain: 8.980, empties: 310.990, ranges: 7.990, std_devs: 1.420
  avg2 - max chain: 8.545, empties: 326.772, ranges: 7.554, std_devs: 1.401
  avg3 - max chain: 8.970, empties: 311.000, ranges: 7.980, std_devs: 1.420

Windows data set:
  avg1 - max chain: 7.703, empties: 277.386, ranges: 6.713, std_devs: 1.339
  avg2 - max chain: 6.901, empties: 314.089, ranges: 5.911, std_devs: 1.353
  avg3 - max chain: 7.703, empties: 277.366, ranges: 6.713, std_devs: 1.340

Note1: table_size == 4093 (about half of the Marsenne prime of 8191)
Note2: avg1 - the original code
Note3: avg2 - gerard's proposed hash function
Note4: avg3 - Thomas'es proposed hash function
Note5: elapsed time: 25.4 sec
Note6: the smaller the numbers the better
```

I think we could go smaller as you asked. My proposed hash function still is the best 2 out of 3 for stddev and all max chain and ranges.

gerard-ziemski:
Once I started thinking of making the `NMTPreInit` table smaller I kept thinking and decided to try moving NMT initialization earlier, before processing of VM arguments. This allows us to keep the table smaller, code simpler (no need to worry about realloc) and delete bunch of testing code, which used some hardcoded limits whose values seemed to be based on lots of heuristic analysis.

Please take another look and tell me what you think.

After this fix goes in, I would like to continue work alongside the idea of less reliance on `NMTPreInit` and try to eliminate it completely, by turning "summary" NMT ON by default in a follow-up issue.

tstuefe:
Sorry, I appreciate your work, but I'm not a big fan. I had similar thoughts when building NMTPreinit, but decided to implement realloc fully because it is just a lot safer and flexible.

You bank on no reallocs ever happening in early initialization code. The amount of code that could run may surprise you. I know it surprised me. And those can be code paths very rarely taken. Realloc is such a basic function. Automatic C++ objects may realloc, for instance. I think UL has a ton of those, as has the Compiler. Or, you may crash and execute the crash handler - are we sure no realloc ever runs then?

Your patch introduces a new dependency: now we will never be able to run code beforehand that uses realloc. And we will not be able to precede NMT argument parsing with another argument. That makes the code brittle.

To be sure this works, we'd need at least add a guarantee - not an assert - at os::realloc(). Because the code paths taken can vary wildly, and you can never be sure we test everything.

Bottomline, you trade in potential crashes (guarantees) and more inflexibility for a complexity reduction in code that makes no real problem. I'd rather just dumb down the test.

src/hotspot/share/runtime/arguments.cpp
```diff
@@ -333,6 +334,29 @@ bool Arguments::is_internal_module_property(const char* property) {
   return false;
 }
 
+ccstr Arguments::process_nmt_property(JavaVMInitArgs* args) {
```
dholmes-ora, line 337: Please don't use "property" here as "property" has a very specific meaning when it comes to command-line arguments. I think "setting" would suffice here.

src/hotspot/share/runtime/threads.cpp
```diff
@@ -450,8 +455,8 @@ jint Threads::create_vm(JavaVMInitArgs* args, bool* canTryAgain) {
   jint parse_result = Arguments::parse(args);
   if (parse_result != JNI_OK) return parse_result;
 
-  // Initialize NMT right after argument parsing to keep the pre-NMT-init window small.
-  MemTracker::initialize();
+  // Verify NMT flag(s) and log the initial state to the ouput
```
dholmes-ora, line 458: Nit: So far in this PR you have used the terms "NMT property", "NMT setting", "NMT arguments", and now "NMT flag(s)". Please pick one (not "property") and use consistently.

dholmes-ora: Typo: ouput

src/hotspot/share/services/memTracker.cpp
```diff
@@ -52,14 +52,14 @@ NMT_TrackingLevel MemTracker::_tracking_level = NMT_unknown;
...
 MemBaseline MemTracker::_baseline;
 
-void MemTracker::initialize() {
+void MemTracker::initialize(ccstr value) {
   bool rc = true;
   assert(_tracking_level == NMT_unknown, "only call once");
 
-  NMT_TrackingLevel level = NMTUtil::parse_tracking_level(NativeMemoryTracking);
+  NMT_TrackingLevel level = NMTUtil::parse_tracking_level(value);
   // Should have been validated before in arguments.cpp
```
dholmes-ora, line 60: Why don't we save the value parsed in arguments.cpp so we don't need to repeat the parsing?

dholmes-ora (commented):
Can someone remind me what the realloc issue is please. On the surface this looks like a big simplification.

Thanks.

tstuefe:
One more note of why I hesitate to remove "realloc"-support:

realloc is often used for buffers that grow on demand. These growth points are difficult to predict since they depend on dynamic buffer content. Therefore this code is very difficult to fully cover with tests.

We have seen these problems with stringStream back when it was still using RA as backing store. All would go well in tests, but when the customer turned on logging in a particular way, and we logged a particular sequence of log strings, the reallocation would hit just right at the wrong spot - in this case, we would reallocate in a leaf function under a temporary ResourceMark - and then later crash in the caller when the RM was removed.

For our case, it means if someone creates a global C++ object that uses e.g. a stringStream (now uses C-Heap) or a GrowableHeapArray, and uses that before CJVM, it may trigger reallocs before CJVM at a customer, but not in our tests.

gerard-ziemski:
Thank you David and Thomas for your feedback.

I will do David's suggestions shortly, but Thomas `realloc` will require bit more time of  thinking it through. And I may need to add `realloc` back...

gerard-ziemski:
I'm leaning towards adding back the realloc to `NMTPreInit` for now.

I will tackle replacing `NMTPreInit` with `NMT` in **summary** mode in a different issue.

src/hotspot/share/runtime/arguments.cpp
gerard-ziemski: Fixed.

src/hotspot/share/runtime/threads.cpp
gerard-ziemski: Fixed.

src/hotspot/share/services/memTracker.cpp
gerard-ziemski: Fixed.

I moved the relevant NMT related argument parsing into one place, hence we only need to parse the setting once.

tstuefe:
> I really do not like the complexity of `NMTPreInit`

Yeah, I noticed. 

IMHO There are tons of better targets for cleanups and reviewer cycles. Just from the top of my head the NMT backend, the virtual memory abstraction layer, UL, ReservedSpace, Hotspot arena handling... all these are areas are quite complex and often very messy and would benefit a lot from cleanup attention.

NMTPreinit was a large improvement over the way things were. It has made little problems over the years, barring over-strict tests like this one (my own darn fault for writing them, ironically - untested code does not draw attention).

> So I can add `realloc` back to `NMTPreInit` if required, after I consider this issue more and am unable to accommodate your useå cases.
> 
> What about the other change, where we move `NMT` initialization before we process the arguments, which makes the `NMTPreInit` table **much** smaller? Is that part OK?

I like the change to a Mersenne prime, that's a good idea.

But the initialization change not so much. It adds a lot of complexity and brittleness to NMT initialization, for not much benefit.

Right now, NMT initialization is *simple*. Because it is time-independent. All VM subsystems are already started, and we are in normal VM land. We can log, we can use Thread::current, we can use ostream, we can crash and have useful hs_err files... That was a big simplification that NMTPreInit brought. 

With your patch, we are back at having two NMT initialization phases. And the early phase happens before UL or Thread::current() or ostream or reasonable hs_err file printing are available.

And we don't get that much for it. `NMTInitializationTest.java` is a very artificial test, not sure we should optimize for that. A normal debug VM started without arguments has 331 pre-init malloc calls. That drops to 47 with your patch. So, with arguments, I convert maybe 300..400 preinit mallocs to normal os::malloc calls.

I think instead our time would be well spent on tracking down those mallocs and removing them. IIRC A lot of that stuff comes from UL. And UL would benefit *so much* from simplification. It is so darn complex for a simple logging framework. I don't see why it needs to dynamically allocate at all before argument parsing, since it has no way of knowing yet what's what. I suspect whatever is generated there could just as well live as automatic variable, or inlined, or hard-coded as constants.

For this test, I think you could make the failure threshold much higher. When I originally wrote the test, I intended to weed out stupid problems like "oh, hashcode is constant", or "oh, we only use every 16th table bucket because of ptr alignment". But in this case, the hash distribution is not that bad.

gerard-ziemski:
OK, I will keep the hash table changes and revert anything else.

I'm not giving up on tackling `NMTPreInit` yet though, but instead of simplifying it, I will see if I can drop all of it, by turning `NMT` in summary mode ON instead (if I can figure out how to color memory pointers)

gerard-ziemski:
Removed anything even remotely controversial. We only change the heuristic function and tweak the hash table size to be 8191, i.e. Mersenne prime.

Thank you David and Thomas for your patience!

gerard-ziemski:
I'm going on vacation for a week this Monday, so if I get your OKs I will integrate after I come back, just in case this will affect anything and I need to be available...

tstuefe (approved):
Good!

dholmes-ora (approved):
I'm also fine with this very simple fix.

Thanks.

gerard-ziemski:
Thank you David and Thomas for the reviews!

PR #14608, opened 2023-06-22
JDK-8310571: Use inline @return tag on java.util.Objects

jddarcy:
Small cleanup, minor differences in the wording of portions of toString(Object, String), nonNull(Object), requireNonNullElse, and requireNonNullElseGet.

---------
### Progress

### Issue
 * [JDK-8310571](https://bugs.openjdk.org/browse/JDK-8310571): Use inline @return tag on java.util.Objects (**Enhancement** - P4)

### Reviewers
 * [Stuart Marks](https://openjdk.org/census#smarks) (@stuart-marks - **Reviewer**) ⚠️ Review applies to [026b3caf](https://git.openjdk.org/jdk/pull/14608/files/026b3cafc24452167ccff720d6bc2f78fcf5c4e6)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14608/head:pull/14608` \
`$ git checkout pull/14608`

`$ git checkout pull/14608` \
`$ git pull https://git.openjdk.org/jdk.git pull/14608/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14608`

`$ git pr show -t 14608`

</details>

<a href="https://git.openjdk.org/jdk/pull/14608.diff">https://git.openjdk.org/jdk/pull/14608.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14608#issuecomment-1601852978)


stuart-marks (approved):
Looks good. I had to refresh my understanding of the exact behavior of the inline `@return` tag:

https://docs.oracle.com/en/java/javase/17/docs/specs/javadoc/doc-comment-spec.html

(search for `{@return`)

It adds a leading "Returns " and a trailing "." to the block's text, and inserts a "Returns" section in the proper place. This seems oddly specific, but it's tailored for this exact use case. Anyway, good cleanup.

pavelrappo:
I can see that `Objects.hash()` was not changed in this PR. Although I think I can see why, I wonder if any distinctions between "Generates" and "Returns" are not important in that context:
```
   /**
    * Generates a hash code for a sequence of input values. The hash
...
    * @return a hash value of the sequence of input values
...
    */
    public static int hash(Object... values) {
```

src/java.base/share/classes/java/util/Objects.java
```diff
@@ -261,15 +248,13 @@ public static <T> T requireNonNull(T obj, String message) {
     }
 
     /**
-     * Returns {@code true} if the provided reference is {@code null} otherwise
-     * returns {@code false}.
+     * {@return {@code true} if the provided reference is {@code
```
pavelrappo, line 251: See my previous comment on the "returns" verb in the doc comment for toString(Object, String). Here (IMO), you picked the better of the two options.

src/java.base/share/classes/java/util/Objects.java
```diff
@@ -297,15 +280,13 @@ public static boolean nonNull(Object obj) {
     }
 
     /**
-     * Returns the first argument if it is non-{@code null} and
-     * otherwise returns the non-{@code null} second argument.
+     * {@return the first argument if it is non-{@code null} and
+     * otherwise returns the non-{@code null} second argument}
```
pavelrappo, line 284: Same comment for "returns".

src/java.base/share/classes/java/util/Objects.java
```diff
@@ -315,15 +296,13 @@ public static <T> T requireNonNullElse(T obj, T defaultObj) {
     }
 
     /**
-     * Returns the first argument if it is non-{@code null} and otherwise
-     * returns the non-{@code null} value of {@code supplier.get()}.
+     * {@return the first argument if it is non-{@code null} and otherwise
+     * returns the non-{@code null} value of {@code supplier.get()}}
```
pavelrappo, line 300: Same.

pavelrappo (commented):
Thanks for doing this, Joe. It seems that I have made multiple comments on the same issue. It makes me wonder if that isn't an issue at all, but is a deliberate choice made in this PR.

PR #14616, opened 2023-06-22
8222329: Readable read(CharBuffer) does not specify that 0 is returned when there is no remaining space in buffer

bplb:
Clarify the behavior of `java.lang.Readable` when the specified `java.nio.CharBuffer` parameter is empty but read-only, and when it is full.

---------
### Progress
- [x] Change requires CSR request [JDK-8310677](https://bugs.openjdk.org/browse/JDK-8310677) to be approved

### Issues
 * [JDK-8222329](https://bugs.openjdk.org/browse/JDK-8222329): Readable read(CharBuffer) does not specify that 0 is returned when there is no remaining space in buffer (**Enhancement** - P3)
 * [JDK-8310677](https://bugs.openjdk.org/browse/JDK-8310677): Readable read(CharBuffer) does not specify that 0 is returned when there is no remaining space in buffer (**CSR**)

### Reviewers
 * [Roger Riggs](https://openjdk.org/census#rriggs) (@RogerRiggs - **Reviewer**) ⚠️ Review applies to [8bcf4674](https://git.openjdk.org/jdk/pull/14616/files/8bcf4674c24f72401ab80564541d724336a60d79)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14616/head:pull/14616` \
`$ git checkout pull/14616`

`$ git checkout pull/14616` \
`$ git pull https://git.openjdk.org/jdk.git pull/14616/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14616`

`$ git pr show -t 14616`

</details>

<a href="https://git.openjdk.org/jdk/pull/14616.diff">https://git.openjdk.org/jdk/pull/14616.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14616#issuecomment-1603131006)

### DIFF
--- a/src/java.base/share/classes/java/io/Reader.java
+++ b/src/java.base/share/classes/java/io/Reader.java
@@ -189,14 +189,17 @@ public abstract class Reader implements Readable, Closeable {
      * Attempts to read characters into the specified character buffer.
      * The buffer is used as a repository of characters as-is: the only
      * changes made are the results of a put operation. No flipping or
-     * rewinding of the buffer is performed.
+     * rewinding of the buffer is performed.  If the specified character
+     * buffer is full, then no characters will be read and zero will be
+     * returned.
      *
      * @param target the buffer to read characters into
      * @return The number of characters added to the buffer, or
      *         -1 if this source of characters is at its end
      * @throws IOException if an I/O error occurs
      * @throws NullPointerException if target is null
-     * @throws java.nio.ReadOnlyBufferException if target is a read only buffer
+     * @throws java.nio.ReadOnlyBufferException if target is a read only buffer,
+     *         even if it is empty
      * @since 1.5
      */
     public int read(CharBuffer target) throws IOException {
--- a/src/java.base/share/classes/java/io/Reader.java
+++ b/src/java.base/share/classes/java/io/Reader.java
@@ -191,7 +191,8 @@ public abstract class Reader implements Readable, Closeable {
      * changes made are the results of a put operation. No flipping or
      * rewinding of the buffer is performed.  If the specified character
      * buffer has no space {@linkplain java.nio.Buffer#hasRemaining
-     * remaining}, then no characters will be read and zero will be returned.
+     * remaining} or or its {@linkplain java.nio.CharBuffer#length length}
+     * is zero, then no characters will be read and zero will be returned.
      *
      * @param target the buffer to read characters into
      * @return The number of characters added to the buffer,
@@ -199,7 +200,7 @@ public abstract class Reader implements Readable, Closeable {
      * @throws IOException if an I/O error occurs
      * @throws NullPointerException if target is null
      * @throws java.nio.ReadOnlyBufferException if target is a read only buffer,
-     *         even if it is empty
+     *         even if its length is zero
      * @since 1.5
      */
     public int read(CharBuffer target) throws IOException {
--- a/src/java.base/share/classes/java/io/Reader.java
+++ b/src/java.base/share/classes/java/io/Reader.java
@@ -191,7 +191,7 @@ public abstract class Reader implements Readable, Closeable {
      * changes made are the results of a put operation. No flipping or
      * rewinding of the buffer is performed.  If the specified character
      * buffer has no space {@linkplain java.nio.Buffer#hasRemaining
-     * remaining} or or its {@linkplain java.nio.CharBuffer#length length}
+     * remaining} or its {@linkplain java.nio.CharBuffer#length length}
      * is zero, then no characters will be read and zero will be returned.
      *
      * @param target the buffer to read characters into
--- a/src/java.base/share/classes/java/io/Reader.java
+++ b/src/java.base/share/classes/java/io/Reader.java
@@ -190,7 +190,7 @@ public abstract class Reader implements Readable, Closeable {
      * The buffer is used as a repository of characters as-is: the only
      * changes made are the results of a put operation. No flipping or
      * rewinding of the buffer is performed.  If the specified character
-     * buffer has no elements {@linkplain java.nio.Buffer#hasRemaining
+     * buffer has no space {@linkplain java.nio.Buffer#hasRemaining
      * remaining}, then no characters will be read and zero will be returned.
      *
      * @param target the buffer to read characters into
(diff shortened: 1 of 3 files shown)


bplb:
CSR will be filed if consensus is reached here.

src/java.base/share/classes/java/lang/Readable.java
```diff
@@ -40,14 +40,17 @@ public interface Readable {
...
      * changes made are the results of a put operation. No flipping or
-     * rewinding of the buffer is performed.
+     * rewinding of the buffer is performed. If the specified character
+     * buffer is full, then no characters will be read and zero will be
+     * returned.
      *
      * @param cb the buffer to read characters into
      * @return The number of {@code char} values added to the buffer,
-     *                 or -1 if this source of characters is at its end
+     *         possibly zero, or -1 if this source of characters is at its end
```
LanceAndersen, line 49: Would it be clearer to specifically indicate 0 is returned when no characters will be read.

RogerRiggs (approved):
Looks good.

src/java.base/share/classes/java/lang/Readable.java
bplb: Might be. I was mimicking the return specification from [FileChannel::read](https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/nio/channels/FileChannel.html#read(java.nio.ByteBuffer[],int,int)).

LanceAndersen: Consistency is probably good I guess in this case, though specifying `zero` vs `0` seems inconsistent when we specify `-1` as I believe this varies throughout the javadoc.  But it matches the other javadoc so we are at least consistent (somewhat )

OK for now by me

LanceAndersen (approved):

src/java.base/share/classes/java/io/Reader.java
```diff
@@ -189,14 +189,17 @@ protected Reader(Object lock) {
      * Attempts to read characters into the specified character buffer.
      * The buffer is used as a repository of characters as-is: the only
      * changes made are the results of a put operation. No flipping or
-     * rewinding of the buffer is performed.
+     * rewinding of the buffer is performed.  If the specified character
+     * buffer is full, then no characters will be read and zero will be
```
bplb, line 193: Instead of "is full" here, it would be more consistent to state "has no remaining elements".

RogerRiggs: For a buffer that is being written into, it sounds a bit odd to say "no remaining elements"; that phrasing is ok for reading/getting bytes but doesn't sound like there's free space to put anything.

Would "no remaining free elements" or "no remaining available elements" be better?

bplb: Or "no remaining space" / "no space remaining"?

RogerRiggs: I have a slight preference for "no space remaining"; remaining can still be a linkplain to the method if that helps it be specific.

bplb: I concur. Thanks.

src/java.base/share/classes/java/nio/X-Buffer.java.template
```diff
@@ -471,13 +471,16 @@ public abstract sealed class $Type$Buffer
...
      * @param target the buffer to read characters into
-     * @return The number of characters added to the buffer, or
-     *         -1 if this source of characters is at its end
+     * @return The number of characters added to the buffer,
+     *         possibly zero, or -1 if this source of characters is at its end
      * @throws IOException if an I/O error occurs
-     * @throws ReadOnlyBufferException if target is a read only buffer
+     * @throws ReadOnlyBufferException if target is a read only buffer,
+     *         even if it is empty
      * @since 1.5
```
AlanBateman, line 484: It might be clearer if this method linked to the length method. That way it could be specified to return 0 when the length of the target buffer is 0.

For the @throws of ROBE, probably need to avoid the word "empty", only because this method writes to the target buffer, it doesn't read from the target buffer.

bplb: Addressed in 94505a45ce099faf3e31a07b8500ac456345a3b1; CSR not yet updated.

src/java.base/share/classes/java/io/Reader.java
bplb: Addressed in 055107bddcc38a331ee93e178402473ab9f91740.

src/java.base/share/classes/java/nio/X-Buffer.java.template
AlanBateman: I checked the update but it now has "no source remaining or its length is zero". In CB, length == remaining so I think you just want one in the description.

bplb: Ah, good catch!

bplb: Fixed by 7281b482c24a3c74da15dbf637f7627c5dc5407b.

src/java.base/share/classes/java/io/Reader.java
```diff
@@ -189,14 +189,18 @@ protected Reader(Object lock) {
      * Attempts to read characters into the specified character buffer.
      * The buffer is used as a repository of characters as-is: the only
      * changes made are the results of a put operation. No flipping or
-     * rewinding of the buffer is performed.
+     * rewinding of the buffer is performed.  If the specified character
+     * buffer has no space {@linkplain java.nio.Buffer#hasRemaining
+     * remaining} or or its {@linkplain java.nio.CharBuffer#length length}
```
RogerRiggs, line 194: typo: "or or" -> "or"

RogerRiggs (approved):
Looks good but for one typo.

src/java.base/share/classes/java/io/Reader.java
bplb: Bummer. Thanks.

bplb: Fixed by 8bcf4674c24f72401ab80564541d724336a60d79.

RogerRiggs (approved):

LanceAndersen (approved):

AlanBateman (approved):

PR #14623, opened 2023-06-23
8309811: BytecodePrinter cannot handle unlinked classes

iklam:
`BytecodePrinter` assumes that it's printing a method from a linked class (e.g., it assumes that the two bytes following an `invokestatic` bytecode have been rewritten to be a `ConstantPoolCache` index.)

This works fine for `-XX:+TraceBytecodes`, which prints only code from linked classes. However, it doesn't work `Method::print_codes_on()` for a method from an unlinked class.

The old code in `BytecodePrinter` uses a very convoluted logic to go from the rewritten index back to the `cp_index`. It's also difficult to understand the type of the indices from the badly named variables like `i` and `orig_i`.

I had to rewrite all the logic for fetching the indices. Now they are in this form:

```
    case Bytecodes::_putfield:
    case Bytecodes::_getfield:
      {
        int cp_index;
        if (is_linked()) {
          int cpcache_index = get_native_index_u2();
          cp_index = cpcache()->entry_at(cpcache_index)->constant_pool_index();
        } else {
          cp_index = get_Java_index_u2();
        }
        print_field_or_method(cp_index, st);
      }
```

The indices have names like `cp_index`, `cpcache_index`, `indy_index` so we know exactly what we are dealing with.

Other changes:
- Added test cases to cover both linked/unlinked classes for the bytecodes modified by this PR
- Fixed the printing of strings to escape unprintable characters (E.g., "\u0001" is very common used by indified string concat)
- Added a test case for `-XX:+TraceBytecodes`
- Added a develop flag `-XX:TraceBytecodesStopAt=` so the test cases won't produce excessive output.

---------
### Progress

### Issue
 * [JDK-8309811](https://bugs.openjdk.org/browse/JDK-8309811): BytecodePrinter cannot handle unlinked classes (**Bug** - P4)

### Reviewers
 * [Frederic Parain](https://openjdk.org/census#fparain) (@fparain - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14623/head:pull/14623` \
`$ git checkout pull/14623`

`$ git checkout pull/14623` \
`$ git pull https://git.openjdk.org/jdk.git pull/14623/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14623`

`$ git pr show -t 14623`

</details>

<a href="https://git.openjdk.org/jdk/pull/14623.diff">https://git.openjdk.org/jdk/pull/14623.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14623#issuecomment-1603698335)


src/hotspot/share/interpreter/bytecodeTracer.cpp
```diff
@@ -55,34 +55,32 @@ class BytecodePrinter {
...
-
-
-  bool      check_index(int i, int& cp_index, outputStream* st);
-  bool      check_cp_cache_index(int i, int& cp_index, outputStream* st);
-  bool      check_obj_index(int i, int& cp_index, outputStream* st);
-  bool      check_invokedynamic_index(int i, int& cp_index, outputStream* st);
+  int       get_native_index_u2()    { int i=Bytes::get_native_u2(_next_pc); _next_pc+=2; return i; }
+  int       get_native_index_u4()    { int i=Bytes::get_native_u4(_next_pc); _next_pc+=4; return i; }
+  int       get_Java_index_u2()      { int i=Bytes::get_Java_u2(_next_pc); _next_pc+=2; return i; }
+  int       get_Java_index_u4()      { int i=Bytes::get_Java_u4(_next_pc); _next_pc+=4; return i; }
```
coleenp, line 70: Since you've done the reformatting, can you add spaces around i=Bytes ? I don't really care if you add spaces around _next_pc+=x.  That looks ok to me.

src/hotspot/share/interpreter/bytecodeTracer.cpp
```diff
@@ -472,19 +353,29 @@ void BytecodePrinter::print_attributes(int bci, outputStream* st) {
...
 
     case Bytecodes::_ldc_w:
     case Bytecodes::_ldc2_w:
-      if (Bytecodes::uses_cp_cache(raw_code())) {
-        print_constant(get_index_u2_cpcache(), st);
-      } else {
-        print_constant(get_index_u2(), st);
+      {
+        int cp_index;
+        if (Bytecodes::uses_cp_cache(raw_code())) {
```
coleenp, line 372: Should this be uses_cp_cache() and is_linked()?

coleenp (changes requested):

src/hotspot/share/interpreter/bytecodeTracer.cpp
iklam: I reformatted the code and reordered the functions a little.

src/hotspot/share/interpreter/bytecodeTracer.cpp
iklam: I added an assert. Such bytecodes can appear only in linked classes.

fparain (approved):
Looks good to me.

iklam:
Thanks @coleenp and @fparain for the review.
/integrate

PR #14625, opened 2023-06-23
8310645: CancelledResponse.java does not use HTTP/2 when testting the  HttpClient

c-cleary:


test/jdk/java/net/httpclient/CancelledResponse2.java
```diff
@@ -0,0 +1,258 @@
...
+                    .GET()
+                    .build();
+            AtomicBoolean cancelled = new AtomicBoolean();
+            BodyHandler<String> bh = ofString(response, cancelled);
+            CompletableFuture<HttpResponse<String>> cf = httpClient.sendAsync(httpRequest, bh);
+            try {
+                cf.get();
+            } catch (Exception e) {
+                e.printStackTrace();
+                assertTrue(e.getCause() instanceof IOException, "HTTP/2 should cancel with an IOException when the Subscription is cancelled.");
```
dfuch, line 107: I believe we want to check that the exception we get at the cause is the expected CancelException, or at least that it can be found somewhere while following the cause chain. Otherwise looks good!

c-cleary: There is a check that the cancelling BodyHandler sets the cancelled field which may be sufficient

c-cleary: Could also just check that the returned a call to `getMessage()` returns a string that contains cancelled but its not a very nice solution

c-cleary: Update: can verify the subscription is properly cancelled by adding a reference to the BodySubscriber used to the CancellingHandler (client) in the test. When the test is done, all thats needed is a call to `SubscriptionWrapper.cancelled()` to verify that the connection ended correctly.

test/jdk/java/net/httpclient/CancelledResponse2.java
```diff
@@ -0,0 +1,258 @@
...
+                }
+                System.err.println();
+            }
+            if (index.get() > 0 && !result.isDone()) {
+                // we should complete the result here, but let's
+                // see if we get something back...
+                out.println("Cancelling subscription after reading " + index.get());
+                cancelled.set(true);
+                subscription.cancel();
+                result.completeExceptionally(new CancelException());
```
dfuch, line 228: I believe that's the exception we expect to find in the `HttpResponse`. IIRC the original test was checking for that.

c-cleary: Right, but calling `subscription.cancel()` causes `Stream.cancelImpl()` to be called. That in turn causes the `HttpResponse` (or the variable result in your snippet above) to complete exceptionally with an IOException which has the message "Stream x cancelled". I think that means that the call to `completeExceptionally(new CancelException())` has no effect because `subscription.cancel()` triggers a call to `cancelImpl()`

PR #14627, opened 2023-06-23
8310735: Build failure after JDK-8310577 with GCC8

DamonFool:
Since `enum BasicType` had been changed to `enum BasicType : u1`, `basic_type >= 0` would be always true.
So it's safe to remove such conditions in these asserts.
Thanks.


DamonFool:
> Changes look good and obviously do fix the build failure. Thanks for taking care.

Thanks @RealLucy for your quick review.

Do you think it's trivial?

RealLucy:
> Do you think it's trivial?

I would guess so. Go ahead and integrate.

DamonFool:
> I would guess so. Go ahead and integrate.

Thanks @RealLucy .
/integrate

coleenp (commented):
Thanks for fixing this.

PR #14632, opened 2023-06-23
8310837: Use ByteArrayLittleEndian in java.util.zip

Glavo:
Using `ByteArrayLittleEndian` is simpler and faster.

`make test TEST="micro:java.util.zip.ZipFileOpen"`:

```diff
  Benchmark                     (size)  Mode  Cnt      Score      Error  Units
- ZipFileOpen.openCloseZipFile     512  avgt   15  39052.832 ±  107.496  ns/op
+ ZipFileOpen.openCloseZipFile     512  avgt   15  36275.539 ±  663.193  ns/op
- ZipFileOpen.openCloseZipFile    1024  avgt   15  77106.494 ± 4159.300  ns/op
+ ZipFileOpen.openCloseZipFile    1024  avgt   15  71955.013 ± 2296.050  ns/op
```

---------
### Progress

### Issue
 * [JDK-8310837](https://bugs.openjdk.org/browse/JDK-8310837): Use ByteArrayLittleEndian in java.util.zip (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14632/head:pull/14632` \
`$ git checkout pull/14632`

`$ git checkout pull/14632` \
`$ git pull https://git.openjdk.org/jdk.git pull/14632/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14632`

`$ git pr show -t 14632`

</details>

<a href="https://git.openjdk.org/jdk/pull/14632.diff">https://git.openjdk.org/jdk/pull/14632.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14632#issuecomment-1605239587)


liach:
Looks great. Created https://bugs.openjdk.org/browse/JDK-8310837 for your patch.

AlanBateman:
@LanceAndersen This one is going to require checking that startup isn't impacted.

src/java.base/share/classes/jdk/internal/util/ByteArrayLittleEndian.java
```diff
@@ -132,6 +132,23 @@ public static int getInt(byte[] array, int offset) {
...
+     * @param array  to get a value from.
+     * @param offset where extraction in the array should begin
+     * @return an {@code int} representing an unsigned short from the array
+     * @throws IndexOutOfBoundsException if the provided {@code offset} is outside
+     *                                   the range [0, array.length - 4]
+     * @see #setUnsignedInt(byte[], int, long)
+     */
+    public static long getUnsignedInt(byte[] array, int offset) {
+        return Integer.toUnsignedLong((int) INT.get(array, offset));
+    }
```
jaikiran, line 150: Hello Glavo, I was going to recommend adding a test method to existing jtreg tests to test these new methods. But it looks like there's no jtreg test for this `ByteArrayLittleEndian` class. Would you mind creating a new test class to (at least) test these methods?

jaikiran:
On an existing Java runtime, I had a quick look at the classes loaded at application startup to see if `MethodHandles` already gets loaded before `ZipUtils` (and related jar/zip classes). From what I see in the output, `MethodHandles` indeed is already loaded before `ZipUtils`. So I think this change doesn't bring in any new/unexpected usage of `MethodHandles` that could potentially impact startup performance.

Glavo:
@AlanBateman @jaikiran Sorry, I thought about its possible impact on startup time, but I don't know which tests can be used to test JVM startup time. Can you tell me some relevant tests?

In fact, I now have a branch(https://github.com/openjdk/jdk/commit/30289cc48c36bd7a59d6c707118d74383bf5f378) that rewrites `ByteArray` and `ByteArrayLittleEndian` to avoid using `VarHandle` because I hope to be able to use them more inside the JDK later.

I tested its performance with the `-Xint` option on and it is indeed faster than `VarHandle` in interpreter mode. But I think I'm missing some benchmark results of the impact on startup time to justify its necessity.

src/java.base/share/classes/jdk/internal/util/ByteArrayLittleEndian.java
Glavo: > Hello Glavo, I was going to recommend adding a test method to existing jtreg tests to test these new methods. But it looks like there's no jtreg test for this `ByteArrayLittleEndian` class. Would you mind creating a new test class to (at least) test these methods?

I think I could modify `test/jdk/jdk/internal/util/ByteArray/ReadWriteValues.java` to also test `ByteArrayLittleEndian`.

Glavo: I added a test for `ByteArrayLittleEndian` in #14636

jaikiran:
Hello Glavo,

> @AlanBateman @jaikiran Sorry, I thought about its possible impact on startup time, but I don't know which tests can be used to test JVM startup time. Can you tell me some relevant tests?
> 
> In fact, I now have a branch(#14636) that rewrites `ByteArray` and `ByteArrayLittleEndian` to avoid using `VarHandle` because I hope to be able to use them more inside the JDK later.
> 

I think this PR should then wait for a bit to see what comes out of https://github.com/openjdk/jdk/pull/14636. That will then help us understand what kind of additional testing might have to be considered for this current PR.

LanceAndersen:
> @LanceAndersen This one is going to require checking that startup isn't impacted.



> @LanceAndersen This one is going to require checking that startup isn't impacted.

ZipFS has a slightly different macro for `SH`:

  ```
 static final int SH(byte[] b, int n) {
        return Byte.toUnsignedInt(b[n]) | (Byte.toUnsignedInt(b[n + 1]) << 8);
    }
```

I think as we compare performance, we also should strive for consistency between Zip and ZipFS implementation as well

**Comment by Glavo** [comment]
> I think as we compare performance, we also should strive for consistency between Zip and ZipFS implementation as well

Can we move `ZipUtils`, `ZipCoder`, and `ZipConstants` to other packages (such as `jdk.internal.zip`), then we can share them between `java.util.zip` and zipfs?

**Comment by AlanBateman** [comment]
> If there's no reason to stop us from doing this, I can open a new PR to do this after this PR is merged.

We shouldn't rush into that as it it creates coupling between java.base and jdk.zipfs that is problematic for maintenance, integrity and security. Need to be careful with every qualified exports that is added.

**Comment by Glavo** [comment]
> We shouldn't rush into that as it it creates coupling between java.base and jdk.zipfs that is problematic for maintenance, integrity and security. Need to be careful with every qualified exports that is added.

Oh well. I'm just confused as to why these few classes are being duplicated, is it for historical reasons?

Another question that confuses me is why `ZipUtils` needs to use `Unsafe` to get the `hb` and `offset` fields of `ByteBuffer`. Is it because of performance?

**Comment by liach** [comment]
> @LanceAndersen This one is going to require checking that startup isn't impacted.

Now that `ByteArrayLittleEndian` is used in Integer toString, don't think this patch will affect the startup class loading sequence.

**Comment by jaikiran** [comment]
Hello Glavo, I see that you are interested in pursuing this change further. Would you mind getting the latest micro benchmark numbers which this proposed change? I see that your PR description has a run from some time back, getting a latest one would be useful.

Additionally, I see that https://github.com/openjdk/jdk/pull/14636 where you had proposed a test case for the `ByteArrayLittleEndian` class (in addition to other things) got closed without being integrated. Would you mind adding a new test case for that class as part of this current PR since you have a few more new methods being added to that class?

**Comment by Glavo** [comment]
> Hello Glavo, I see that you are interested in pursuing this change further. Would you mind getting the latest micro benchmark numbers which this proposed change? I see that your PR description has a run from some time back, getting a latest one would be useful.
> 
> Additionally, I see that #14636 where you had proposed a test case for the `ByteArrayLittleEndian` class (in addition to other things) got closed without being integrated. Would you mind adding a new test case for that class as part of this current PR since you have a few more new methods being added to that class?

I've moved those changes into this PR and am running tests. I'll push these changes once the tests are finished running.

**Comment by Glavo** [comment]
I ran the tier1 tests with no failures.

**Comment by Glavo** [comment]
@jaikiran Can you review this PR again?

**Comment by jaikiran** [comment]
Hello Glavo, I'll need some more time on this to review this. In the meantime, could you update the micro benchmark latest numbers with this latest state?

**Comment by minborg** [comment]
Nice with better testing for the utility classes.

**Comment by Glavo** [comment]
> Hello Glavo, I'll need some more time on this to review this. In the meantime, could you update the micro benchmark latest numbers with this latest state?

Latest results:

```diff
  Benchmark                       (size)  Mode  Cnt      Score     Error  Units
+ ZipFileOpen.openCloseZipFile       512  avgt   15  53247.244 ± 602.826  ns/op
- ZipFileOpen.openCloseZipFile       512  avgt   15   55411.637 ± 2010.322  ns/op
+ ZipFileOpen.openCloseZipFile      1024  avgt   15  96463.939 ± 273.324  ns/op
- ZipFileOpen.openCloseZipFile      1024  avgt   15   99983.208 ±  374.463  ns/op
+ ZipFileOpen.openCloseZipFilex2     512  avgt   15  54735.296 ± 241.473  ns/op
- ZipFileOpen.openCloseZipFilex2     512  avgt   15   57308.739 ± 1627.708  ns/op
+ ZipFileOpen.openCloseZipFilex2    1024  avgt   15  98095.240 ± 319.524  ns/op
- ZipFileOpen.openCloseZipFilex2    1024  avgt   15  102068.887 ±  318.778  ns/op
```

liach:
@cl4es Can you take a look at this performance improvement patch?

Glavo:
I've been busy working on other projects lately, so I've put these PRs on hold. If anyone cares to take a look at them I'll be right back.

PR #14636, opened 2023-06-24
8310843: Reimplement ByteArray and ByteArrayLittleEndian with Unsafe

Glavo:
`ByteArray` and `ByteArrayLittleEndian` are very useful tool classes that can be used in many places to performance tuning.

Currently they are implemented by `VarHandle`, so using them may have some impact on startup time.

This PR reimplements them using `Unsafe`, which reduces the impact on startup time.

---------
### Progress

### Issue
 * [JDK-8310843](https://bugs.openjdk.org/browse/JDK-8310843): Reimplement ByteArray and ByteArrayLittleEndian with Unsafe (**Enhancement** - P4)

### Reviewers
 * [Chen Liang](https://openjdk.org/census#liach) (@liach - Author)
 * [Jim Laskey](https://openjdk.org/census#jlaskey) (@JimLaskey - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14636/head:pull/14636` \
`$ git checkout pull/14636`

`$ git checkout pull/14636` \
`$ git pull https://git.openjdk.org/jdk.git pull/14636/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14636`

`$ git pr show -t 14636`

</details>

<a href="https://git.openjdk.org/jdk/pull/14636.diff">https://git.openjdk.org/jdk/pull/14636.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14636#issuecomment-1605327569)


Glavo:
I removed `getFloatRaw`/`getDoubleRaw` because they are exactly the same as `getFloat`/`getDouble`, so they have no meaning.

liach:
The reimplementation allows parts that invoke package depend on to utilize faster byte array access, namely bytecode generation and Classfile API; which IMO is more important than the reduction on startup time.

src/java.base/share/classes/jdk/internal/util/ByteArray.java
```diff
@@ -41,12 +43,16 @@ public final class ByteArray {
...
     }
 
-    private static final VarHandle SHORT = create(short[].class);
-    private static final VarHandle CHAR = create(char[].class);
-    private static final VarHandle INT = create(int[].class);
-    private static final VarHandle FLOAT = create(float[].class);
-    private static final VarHandle LONG = create(long[].class);
-    private static final VarHandle DOUBLE = create(double[].class);
+    private static final BiFunction<String, List<Number>, ArrayIndexOutOfBoundsException>
+            OOBEF = Preconditions.outOfBoundsExceptionFormatter(ArrayIndexOutOfBoundsException::new);
```
liach, line 47: Can't this be replaced by `Preconditions.AIOBBE_FORMATTER`? And this lambda expression prevents this class to be used before java.lang.invoke initialization.

liach (changes requested):

liach:
Created an issue at https://bugs.openjdk.org/browse/JDK-8310843.
I look forward to this improvement, for it can potentially be used by the Classfile API to speed up Classfile parsing and writing.

src/java.base/share/classes/jdk/internal/util/ByteArray.java
```diff
@@ -41,12 +40,13 @@ public final class ByteArray {
     private ByteArray() {
     }
 
-    private static final VarHandle SHORT = create(short[].class);
-    private static final VarHandle CHAR = create(char[].class);
-    private static final VarHandle INT = create(int[].class);
-    private static final VarHandle FLOAT = create(float[].class);
-    private static final VarHandle LONG = create(long[].class);
-    private static final VarHandle DOUBLE = create(double[].class);
+    static final Unsafe UNSAFE = Unsafe.getUnsafe();
```
liach, line 43: I recommend documenting the reason to use Unsafe (early startup) and comment that regular users should use VarHandle instead.

src/java.base/share/classes/jdk/internal/util/ByteArray.java
```diff
@@ -25,13 +25,12 @@
...
-import java.nio.ByteOrder;
+import jdk.internal.misc.Unsafe;
+import jdk.internal.vm.annotation.ForceInline;
 
 /**
  * Utility methods for packing/unpacking primitive values in/out of byte arrays
- * using {@linkplain ByteOrder#BIG_ENDIAN big endian order} (aka. "network order").
+ * using {@linkplain java.nio.ByteOrder#BIG_ENDIAN big endian order} (aka. "network order").
  * <p>
  * All methods in this class will throw an {@linkplain NullPointerException} if {@code null} is
```
liach, line 35: Do the Unsafe methods throw NPE on null array argument?

liach (commented):
I recommend documenting that this class is intended to be usable at early startup so we don't accidentally introduce features like lambda into the code.

src/java.base/share/classes/jdk/internal/util/ByteArray.java
Glavo: > Do the Unsafe methods throw NPE on null array argument?

No, but `arrayOffset` will throw.

Glavo:
I deleted some incorrect comments.

The original author of these two classes misunderstood the behavior of `intBitsToFloat` and `longBitsToDouble`. These two methods never collapse NaN values to a "canonical" NaN value, so the comments is incorrect.

I deleted those comments because conversions from `int`/`long` to `float`/`double` don't need to differentiate that behavior, their difference only exists when converting from `float`/`double` to `int`/`long`.

The conversion methods in the `Float`/`Double` class also exhibits this: there is only one method for converting from `int`/`long` to `float`/`double` , but there are two methods for converting from `float`/`double` to `int`/`long`.

Glavo:
Can someone review this PR?

liach:
@minborg Sorry for pinging, but I think this might be of your interest since you've introduced the ByteArray classes.

AlanBateman:
> I deleted some incorrect comments.
> 
> The original author of these two classes misunderstood the behavior of `intBitsToFloat` and `longBitsToDouble`. These two methods never collapse NaN values to a "canonical" NaN value, so the comments is incorrect.

You'll need to go back through the discussion in the original PRs (there was more than one) as there was lengthy discussion this already.

AlanBateman:
> Can someone review this PR?

What issue are you running into at startup that require this change?

liach:
> > Can someone review this PR?
> 
> What issue are you running into at startup that require this change?

I wish to use this too before java.lang.invoke is ready, such as reading unaligned in the Classfile API.

Glavo:
> You'll need to go back through the discussion in the original PRs (there was more than one) as there was lengthy discussion this already.

https://github.com/openjdk/jdk/pull/11840#discussion_r1064489163

Are you referring to this discussion?

I understand the original reason for retaining it before, but this discussion is outdated for this PR.

`Unsafe` does not provide `getFloatUnaligned`/`getDoubleUnaligned` and `putFloatUnaligned`/`putDoubleUnaligned`, so we must convert floating point numbers to integers here. This problem is no longer hidden behind `VarHanle`, so we have to face it.

In addition, these comments are indeed wrong and may mislead users. I saw this mentioned in the previous discussion, but I don't know why the document was not cleaned up. I think it's time to clean it up.

Glavo:
Can any reviewers take a look at this PR?

JimLaskey (approved):

RogerRiggs (approved):
LGTM

liach (approved):

AlanBateman:
I don't think this change should be integrated without further discussion on what problems you are running into. Code that runs is initPhase1 needs to have as few dependences as possible and can use Unsafe directly when needed.

liach:
This patch merely allows `ByteArray` and `ByteArrayLittleEndian` to be used before `VarHandle` is ready. An example is the Classfile API (which will be used to generate `LambdaForm`s for `VarHandle`), as shown in #14637.

src/java.base/share/classes/jdk/internal/util/ByteArray.java
```diff
@@ -41,12 +40,21 @@ public final class ByteArray {
...
-    private static final VarHandle FLOAT = create(float[].class);
-    private static final VarHandle LONG = create(long[].class);
-    private static final VarHandle DOUBLE = create(double[].class);
+    /**
+     * The {@code Unsafe} can be functionality replaced by
+     * {@linkplain java.lang.invoke.MethodHandles#byteArrayViewVarHandle byteArrayViewVarHandle},
+     * but it's not feasible in practices, because {@code ByteArray} and {@code ByteArrayLittleEndian}
+     * can be used in fundamental classes, {@code VarHandle} exercise many other
+     * code at VM startup, this could lead a recursive calls when fundamental
+     * classes is used in {@code VarHandle}.
```
AlanBateman, line 49: This comment is confusing. If this code is changed then the comment can be very simple to say that it uses Unsafe to allow it be used in early startup and the VarHandle implementation.

src/java.base/share/classes/jdk/internal/util/ByteArray.java
```diff
@@ -41,12 +40,21 @@ public final class ByteArray {
...
+     */
+    static final Unsafe UNSAFE = Unsafe.getUnsafe();
+
+    @ForceInline
+    static long arrayOffset(byte[] array, int typeBytes, int offset) {
```
mcimadamore, line 54: IMHO, this is the really interesting thing that this class does - e.g. it introduces a way to translate a (logical) offset into a byte array into a physical offset that can be used for unsafe. After you have an helper method like this, it seems like the client can just do what it wants by using Unsafe directly (which would remove the need for having this class) ? Was some experiment of that kind done (e.g. replacing usage of ByteArray with Unsafe + helpers) - or does it lead to code that is too cumbersome on the client?

Also, were ByteBuffers considered as an alternative? (I'm not suggesting MemorySegment as those depend on VarHandle again, but a heap ByteBuffer is just a thin wrapper around an array which uses Unsafe). ByteBuffer will have a bound check, but so does your code (which call checkIndex). I believe that, at least in hot code, wrapping a ByteBuffer around a byte array should be routinely scalarized, as there's no control flow inside these little methods.

mcimadamore: Actually, a byte buffer is big endian, so some extra code would be required. But maybe that's another helper function:

```
@ForceInline
ByteBuffer asBuffer(byte[] array) { return ByteBuffer.wrap(array).order(ByteOrder.nativeOrder()); }
```

And then replace:

```
ByteArray.getChar(array, 42)
```
With
```
asBuffer(array).getChar(42);
```

mcimadamore: Also... in a lot of cases where ByteArray is used (DataXYZStream, ObjectXYZStream) the array being used is a field in the class. So the byte buffer creation can definitively be amortized (or the code changed to work on buffers instead of arrays).

mcimadamore:
> The reimplementation allows parts that invoke package depend on to utilize faster byte array access, namely bytecode generation and Classfile API; which IMO is more important than the reduction on startup time.

Is `BufWriterImpl` what you are looking to refactor using this class? Any reason why that implementation doesn't internally rely on a ByteBuffer, rather than a plain array?

src/java.base/share/classes/jdk/internal/util/ByteArray.java
liach: The Unsafe-based writing will be used by `Integer.toString` and `Long.toString` as well; in those cases, will creating a ByteBuffer wrapper be overkill?

liach:
> Is `BufWriterImpl` what you are looking to refactor using this class? Any reason why that implementation doesn't internally rely on a ByteBuffer, rather than a plain array?

Good point. Not only that, `ClassReaderImpl` should be migrated to read U2/S2 and U4/S4 via `ByteBuffer` as well.

src/java.base/share/classes/jdk/internal/util/ByteArray.java
AlanBateman: > The Unsafe-based writing will be used by `Integer.toString` and `Long.toString` as well; in those cases, will creating a ByteBuffer wrapper be overkill?

Integer/Long are very core classes so I assume they can use Unsafe if needed, they probably want as few dependences as possible.

mcimadamore: > in those cases, will creating a ByteBuffer wrapper be overkill?

Perhaps - but do we have any benchmark to back up that claim (or backing up the fact that Integer.toString would benefit from using ByteArray in the first place) ? It is likely that in code that is compiled and inlined, this will make no difference at all, as an object that is created, and then disposed of will be routinely scalarized (since there's no meaningful control flow).

If in some cases we have reason to believe that we need more performance, we could always fall back and use Unsafe directly.

I guess in general I'm a bit skeptical re. creating a class that wraps an array and exposes get/setXYZ that are implemented by Unsafe, adding some bound checks on top. That's 99.99% of what an heap buffer does. I've benchmarked BB quite a extensively given my work on Panama and I can assure that, if accessed using the right API (e.g. absolute addressing scheme, which takes an offset) and the right endianness (ByteOrder::NATIVE_ORDER), the performance of BB access is very very very close to Unsafe (esp. here, given that ByteArray also introduces some kind of bound check).

mcimadamore: Also... `Integer::toString` seems to be `@IntrinsicCandidate` ?

RogerRiggs (changes requested):
Til the comments settle down; don't integrate or sponsor.

src/java.base/share/classes/jdk/internal/util/ByteArray.java
Glavo: > Also... `Integer::toString` seems to be `@IntrinsicCandidate` ?

It's just a bytecode intrinsics, it is only replaced when used in a fluent-chain of StringBuilder/Buffer.

RogerRiggs:
These classes are used where numbers are being formatted into existing byte arrays and may be assembled with other strings.
For example, StringConcatHelper and AbstractStringBuilder, both optimized for assembling LATIN1 and UTF16 strings.
Switching to use a different implementation would result in more allocation and/or copying the strings.

src/java.base/share/classes/jdk/internal/util/ByteArray.java
Glavo: I tried a few more. It looks like the JIT is able to optimize the ByteBuffer away pretty well by keeping it only as a local variable without escaping.

Glavo: It seems that as long as the `ByteBuffer` is stored in a field (even if it is `static final`), the JIT compiler cannot completely eliminate the overhead of the `ByteBuffer`.

@mcimadamore I think your suggested changes for `DataInputStream` is dubious, it's likely to introduce non-trivial additional overhead. The correct change may be like this:

```diff
public final double readDouble() throws IOException {
    readFully(readBuffer, 0, 8);
-   return ByteArray.getDouble(readBuffer, 0);
+   return ByteBuffer.wrap(readBuffer).getDouble(0);
}
```

However this change can also increase the warmup time and allocate many small objects before C2 compiles it.

mcimadamore: > it's likely to introduce non-trivial additional overhead

What do you mean? E.g. where would the overhead come from?

Glavo: > > it's likely to introduce non-trivial additional overhead
> 
> What do you mean? E.g. where would the overhead come from?

You suggested changes stored the ByteBuffer into a field. I measured throughput with JMH, and it shows that using ByteBuffer like this results in over 50% drop in throughput.

mcimadamore: > In this result, ByteBuffer is much slower than VarHandle. Am I doing something wrong? What conditions are needed to make the performance of ByteBuffer close to that of Unsafe?

These are some benchmarks we have in the JDK:

Off-heap access:

```
Benchmark                       Mode  Cnt  Score   Error  Units
LoopOverNonConstant.BB_get      avgt   30  3.837 ± 0.045  ns/op
LoopOverNonConstant.unsafe_get  avgt   30  3.425 ± 0.032  ns/op
```

On-heap access:

```
Benchmark                           (polluteProfile)  Mode  Cnt  Score   Error  Units
LoopOverNonConstantHeap.BB_get                 false  avgt   30  3.897 ± 0.055  ns/op
LoopOverNonConstantHeap.unsafe_get             false  avgt   30  3.446 ± 0.027  ns/op
```

BB is almost on par with unsafe. There's of course an extra bound check which you pay for. These benchmark are almost idential to the one you shared, so not sure what's going on.

Glavo: @mcimadamore I ran a series of experiments. When `ByteBuffer` is only used as a local variable without escaping, it is very close to `VarHandle`(`Unsafe`). However, any attempt to put the `ByteBuffer` into a field (even a static final field) resulted in a throughput drop of more than 50%.

mcimadamore: I just ran your benchmark above - which is similar as I said to the ones we already have in the JDK. Results:
```
Benchmark                     Mode  Cnt  Score   Error  Units
ByteArray.readByte            avgt   30  3.619 ± 0.051  ns/op
ByteArray.readByteFromBuffer  avgt   30  3.936 ± 0.071  ns/op
ByteArray.readInt             avgt   30  3.653 ± 0.054  ns/op
ByteArray.readIntFromBuffer   avgt   30  3.798 ± 0.019  ns/op
ByteArray.readLong            avgt   30  3.637 ± 0.053  ns/op
ByteArray.readLongFromBuffer  avgt   30  3.922 ± 0.012  ns/op
```

I don't see 2x slowdown.

Glavo: > I don't see 2x slowdown. Which JDK are you using? Which platform?

I tried running the benchmarks with OpenJDK 20.0.1 and my own jdk built from master and the results were similar. The test platform is based on Ubuntu 23.04, and the CPU is AMD Ryzen 7 3700X.

Here's the benchmark configuration I use:

```
@Warmup(iterations = 5, time = 3)
@Measurement(iterations = 5, time = 2)
@Fork(value = 1, jvmArgsAppend = {"-XX:+UseG1GC", "-Xms4g", "-Xmx4g"})
@OutputTimeUnit(TimeUnit.MILLISECONDS)
@State(Scope.Benchmark)
```

Also, I just ran the benchmarks on my ARM machine (Ubuntu 22.04, RK3588) and here are the results:

```
Benchmark                      Mode  Cnt       Score       Error   Units
ByteArray.readByte            thrpt    5  637843.606 ± 23657.623  ops/ms
ByteArray.readByteFromBuffer  thrpt    5  350328.152 ±  3635.769  ops/ms
ByteArray.readInt             thrpt    5  633518.784 ±  2558.021  ops/ms
ByteArray.readIntFromBuffer   thrpt    5  327456.830 ±  4588.002  ops/ms
ByteArray.readLong            thrpt    5  627409.162 ± 29301.498  ops/ms
ByteArray.readLongFromBuffer  thrpt    5  328154.906 ±  1067.439  ops/ms
```

Glavo: > Here's with the same parameters as the one you are using:

I don't understand why there is such a difference. I have replicated similar results on several other devices:

Apple M1:
```
Benchmark                      Mode  Cnt        Score       Error   Units
ByteArray.readByte            thrpt    5  1418606.665 ±  2308.711  ops/ms
ByteArray.readByteFromBuffer  thrpt    5   879084.678 ± 20981.303  ops/ms
ByteArray.readInt             thrpt    5  1404164.063 ±  4845.412  ops/ms
ByteArray.readIntFromBuffer   thrpt    5   829768.843 ±  8730.467  ops/ms
ByteArray.readLong            thrpt    5  1400713.660 ±  1924.752  ops/ms
ByteArray.readLongFromBuffer  thrpt    5   825721.508 ±  4146.961  ops/ms
```


Intel Celeron N5105:
```
Benchmark                      Mode  Cnt       Score        Error   Units
ByteArray.readByte            thrpt    5  927248.807 ±  29630.442  ops/ms
ByteArray.readByteFromBuffer  thrpt    5  448374.359 ±   3210.571  ops/ms
ByteArray.readInt             thrpt    5  719143.935 ± 183584.747  ops/ms
ByteArray.readIntFromBuffer   thrpt    5  398578.488 ±    100.058  ops/ms
ByteArray.readLong            thrpt    5  742555.147 ±  13006.710  ops/ms
ByteArray.readLongFromBuffer  thrpt    5  398771.660 ±   1673.524  ops/ms
```

AMD Ryzen 7 5800X:
```
Benchmark                      Mode  Cnt        Score       Error   Units
ByteArray.readByte            thrpt    5  1784096.905 ± 47611.535  ops/ms
ByteArray.readByteFromBuffer  thrpt    5   992974.747 ± 19389.011  ops/ms
ByteArray.readInt             thrpt    5  1719368.129 ± 70606.284  ops/ms
ByteArray.readIntFromBuffer   thrpt    5   919654.075 ±  9292.283  ops/ms
ByteArray.readLong            thrpt    5  1719873.280 ± 18441.122  ops/ms
ByteArray.readLongFromBuffer  thrpt    5   921992.992 ±  6439.747  ops/ms
```

mcimadamore: On a newer processor/OS (Alder Lake/Ubuntu 22.04) I see a bit more difference:

```
Benchmark                      Mode  Cnt       Score       Error   Units
ByteArray.readByte            thrpt    5  680397.046 ± 27504.022  ops/ms
ByteArray.readByteFromBuffer  thrpt    5  576449.569 ±  3633.135  ops/ms
ByteArray.readInt             thrpt    5  685419.089 ±  6740.268  ops/ms
ByteArray.readIntFromBuffer   thrpt    5  542887.418 ±  2863.907  ops/ms
ByteArray.readLong            thrpt    5  687949.037 ±  3510.613  ops/ms
ByteArray.readLongFromBuffer  thrpt    5  548120.950 ±  5461.145  ops/ms
```

But still far from 2x.

mcimadamore: For full disclosure, I'm running the benchmark using the JDK microbenchmark support. My JMH version is a bit behind. I've updated it to 1.35 (which is the latest I see being used around here) and getting similar results.

Glavo: By the way, I ran `LoopOverNonConstantHeap` on the 3700x platform, and the performance of ByteBuffer was also poor:

```
Benchmark                                                (polluteProfile)  Mode  Cnt  Score    Error  Units
LoopOverNonConstantHeap.BB_get                                      false  avgt   30  1.855 ±  0.092  ns/op
LoopOverNonConstantHeap.BB_get                                       true  avgt   30  1.655 ±  0.010  ns/op
LoopOverNonConstantHeap.BB_loop                                     false  avgt   30  0.254 ±  0.001  ms/op
LoopOverNonConstantHeap.BB_loop                                      true  avgt   30  0.254 ±  0.001  ms/op
LoopOverNonConstantHeap.segment_get                                 false  avgt   30  2.333 ±  0.043  ns/op
LoopOverNonConstantHeap.segment_get                                  true  avgt   30  2.362 ±  0.006  ns/op
LoopOverNonConstantHeap.segment_loop                                false  avgt   30  0.251 ±  0.001  ms/op
LoopOverNonConstantHeap.segment_loop                                 true  avgt   30  0.251 ±  0.001  ms/op
LoopOverNonConstantHeap.segment_loop_instance                       false  avgt   30  0.254 ±  0.001  ms/op
LoopOverNonConstantHeap.segment_loop_instance                        true  avgt   30  0.254 ±  0.001  ms/op
LoopOverNonConstantHeap.segment_loop_instance_unaligned             false  avgt   30  0.254 ±  0.001  ms/op
LoopOverNonConstantHeap.segment_loop_instance_unaligned              true  avgt   30  0.254 ±  0.001  ms/op
LoopOverNonConstantHeap.segment_loop_readonly                       false  avgt   30  0.252 ±  0.001  ms/op
LoopOverNonConstantHeap.segment_loop_readonly                        true  avgt   30  0.251 ±  0.001  ms/op
LoopOverNonConstantHeap.segment_loop_slice                          false  avgt   30  0.251 ±  0.001  ms/op
LoopOverNonConstantHeap.segment_loop_slice                           true  avgt   30  0.252 ±  0.001  ms/op
LoopOverNonConstantHeap.segment_loop_unaligned                      false  avgt   30  0.251 ±  0.001  ms/op
LoopOverNonConstantHeap.segment_loop_unaligned                       true  avgt   30  0.252 ±  0.001  ms/op
LoopOverNonConstantHeap.unsafe_get                                  false  avgt   30  0.628 ±  0.004  ns/op
LoopOverNonConstantHeap.unsafe_get                                   true  avgt   30  0.628 ±  0.003  ns/op
LoopOverNonConstantHeap.unsafe_loop                                 false  avgt   30  0.255 ±  0.001  ms/op
LoopOverNonConstantHeap.unsafe_loop                                  true  avgt   30  0.256 ±  0.001  ms/op
```

mcimadamore: > By the way, I ran `LoopOverNonConstantHeap` on the 3700x platform, and the performance of ByteBuffer was also poor:

I finally see it.

```
Benchmark                           (polluteProfile)  Mode  Cnt  Score   Error  Units
LoopOverNonConstantHeap.BB_get                 false  avgt   30  1.801 ± 0.020  ns/op
LoopOverNonConstantHeap.unsafe_get             false  avgt   30  0.567 ± 0.007  
```

It seems that, between updating JMH and rebuilding the JDK from scratch, *something* did the trick.

While I knew that random access on a BB is slower than Unsafe (as there's an extra check), whereas looped access is as fast (as C2 is good at hoisting the checks outside the loop, as shown in the benchmark). Note also that we are in the nanosecond realm, so each instruction here counts.

Is there any benchmark for DataInput/Output stream that can be used? I mean, it would be interesting to understand how these numbers translate when running the stuff that is built on top.

mcimadamore: > Is there any benchmark for DataInput/Output stream that can be used? I mean, it would be interesting to understand how these numbers translate when running the stuff that is built on top.

I've tried to run the benchmark in test/micro/java/io/DataInputStream.java. This is the baseline:

```
Benchmark                     Mode  Cnt  Score   Error  Units
DataInputStreamTest.readChar  avgt   20  7.583 ± 0.026  us/op
DataInputStreamTest.readInt   avgt   20  3.804 ± 0.045  us/op
```

And this is with a patch similar to the one I shared above, to use ByteBuffer internally:

```
Benchmark                     Mode  Cnt  Score   Error  Units
DataInputStreamTest.readChar  avgt   20  7.594 ± 0.106  us/op
DataInputStreamTest.readInt   avgt   20  3.795 ± 0.030  us/op
```

There does not seem to be any extra overhead. That said, access occurs in a counted loop, and in these cases we know buffer/segment access is optimized quite well.

I believe the question here is: do we have benchmark which are representative of the kind of gain that would be introduced by micro-optimizing ByteArray? It can be quite tricky to estimate real benefits from synthetic benchmark on the ByteArray class, especially when fetching a single element outside of a loop - as those are not representative of how the clients will use this. I note that the original benchmark made by Per used a loop with two iterations to assess the cost of the ByteArray operations:

http://minborgsjavapot.blogspot.com/2023/01/java-21-performance-improvements.html

If I change the benchmark to do 2 iterations, I see this:

```
Benchmark                      Mode  Cnt       Score       Error   Units
ByteArray.readByte            thrpt    5  704199.172 ± 34101.508  ops/ms
ByteArray.readByteFromBuffer  thrpt    5  474321.828 ±  6588.471  ops/ms
ByteArray.readInt             thrpt    5  662411.181 ±  4470.951  ops/ms
ByteArray.readIntFromBuffer   thrpt    5  496900.429 ±  3705.737  ops/ms
ByteArray.readLong            thrpt    5  665138.063 ±  5944.814  ops/ms
ByteArray.readLongFromBuffer  thrpt    5  517781.548 ± 27106.331  ops/ms
```
The more the iterations, the less the cost (and you don't need many iterations to break even). This probably explains why the DataInputStream benchmark doesn't change - there's 1024 iterations in there.

I guess all this is to say that excessively focussing on microbenchmark of a simple class such as ByteArray in conditions that are likely unrealistic (e.g. single access) is IMHO the wrong way to look at things, as ByteArray is mostly used by classes that most definitively will read more than one value at a time (including classfile API). 

So, also IMHO, we should try to measure the use cases we care about of the higher-level API we care about (I/O streams, classfile) and then see if adding Unsafe/VarHandle/ByteBuffer access in here is going to lead to any benefit at all.

uschindler: Just some feedback about this discussion:
- I agree that the DataInput/OutputStreams should maybe use `ByteBuffer` directly as they use buffering already. So the patch above looks fine. In my project Apache Lucene (which has many performance critical methods like this), we have already implemented ByteBuffer based access like this for all IO-stream-based classes (we call then `DataInput/DataOutput`). I don't know why you have seen differences in using a `ByteBuffer` as final field in the class. That's common and used in most frameworks out there (NETTY,...) and is bullet proof (unless theres a bug in optimizer which sometimes happened in the past).
- We noticed that wrapping a byte array on each access by ByteBuffer causes a lot of overhead and GC activity if used in hot loops. In addition we have seen cases where it is not optimized anymore (not sure why). @mcimadamore: You remember the similar discussions about the `MemorySegment` slices and copying them around between heap/foreign? Maybe inside the JDK you can do better by using `@ForceInline`. Our code can't do this, so we try to avoid creating instances of classes in such low-level code.
- The original VarHandle approach is now used in Lucene's code at all places (basically the idea to use VarHandles for this class was suggested by me a while back). We often have byte arrays and can't wrap them as ByteBuffer on each call (because its not always inlined). For code outside of the JDK this looks like the best approach to have fast access to short/int/long values at specific (not necessarily aligned) positions. We have seen LZ4 compression getting much faster after changing the code from manually constructing logs/floats from bytes like in the reference code. With ByteBuffer it was often getting slower (depending on how it was called, I think because we can't do `@ForceInline` in code outside the JDK.

Generally: A class like this is very nice and also very much needed in code outside the JDK. A lot of code like encoding/decoding network bytes or compression algorithms often has the pattern that they want to read primitive types from byte arrays from. The overhead with wrapping looks bad in code and also causes long startup times and sometimes also OOM (if used multithreaded from different threads hammering the byte array accessors). Also you don not want to write a LZ4 decompressor using ByteBuffer as its only source of data... :-(

So have you thought of making this low-level classes public so we outside users no longer need to deal with VarHandles?

Maybe `java.util.ByteArrays` with solely static methods. The internal implementation of such a useful basic utility class could definitly be using `Unsafe` internally, so I would leave out the discussion here. If you use Unsafe there are no surprises! Personally I have no problem with the current implementation in this PR! I would just put little/big endian impl in the same class and move it to java.util (this is just my comment about this, coming from a library which does this low level stuff all the time).

mcimadamore: > 
> So have you thought of making this low-level classes public so we outside users no longer need to deal with VarHandles?
> 
I believe this is beyond the scope of this PR.

As for what do we do in the JDK, I can see few options:

1. We keep things as they are in current mainline.
2. We keep changes in this PR.
3. We rewrite most uses of ByteArray in java.io to use BB and remove ByteArray
4. We remove ByteArray and provide some static helper function to generate an unsafe offset from an array

I agree with @uschindler that wrapping stuff in ByteBuffer "on the fly" might be problematic for code that is not inlined, so I don't think we should do that.

I have to admit that I'm a little unclear as to what the goal of this PR is. Initially, it started as an "improve startup" effort, which then morphed into a "let's make ByteArray" more usable, even for other clients (like classfile API), or Long::toString. I'm unsure about the latter use cases, because (a) Long/Integer are core classes and should probably use Unsafe directly, where needed and (b) for classfile API, using ByteBuffer seems a good candidate on paper (of course there is the unknown of how well the byte buffer access will optimize in the classfile API code - but if there's more than one access on the same buffer, we should be more than ok).

I'd like to add some more words of caution against the synthetic benchmarks that we tried above. These benchmarks are quite peculiar, for at least two reasons:

* we only ever access one element
* the accessed offset is always zero

No general API can equal Unsafe under this set of conditions. When playing with the benchmark I realize that every little thing mattered (we're really measuring the number of instructions emitted by C2)  - for instance, the fact that when access occurs with a byte buffer, the underlying array and limit have to be fetched from their fields has a cost. Also, the fact that ByteBuffer has a hierarchy has an even bigger cost (as C2 has to make sure you are really invoking HeapByteBuffer). The mutable endianness state in byte buffer also adds up to the noise. The above is what ends up in a big fat "2x slower" label.

That said, all these "factors" are only relevant because we're looking at a _single_ buffer operation. In fact, all such costs can be easily be amortized as soon as there more than one access. Or as soon as you start accessing offsets that are not known statically (unlike in the benchmark).

So, there's a question of what's the code idiom that leads to the absolute fastest code (and I agree that Unsafe + static wrappers seems the best here). And then there's the question of "but, what do we need to get the performance number/startup behavior we want". I feel the important question is the second, but we keep arguing about the former.

And, to assess that second question, we need to understand better what the goals are (which, so far, seems a bit fuzzy).

uschindler: >> So have you thought of making this low-level classes public so we outside users no longer need to deal with VarHandles?
> I believe this is beyond the scope of this PR.

Sure, I brought this up here but yes, it is not really the scope of this PR. It is just another idea that this class could be of more wide use although outside of this PR and also outside of Lucene. Actually it would be nice to have it public, but I know this involves creating a JEP and so on. If there's interest I could start on proposing something like this on mailing list, later creating a JEP or whatever else is needed.

P.S.: Actually for a 3rd party user the whole thing is not much complicated. You only need a class to allocate the VarHandles and then use them from code, you don't even need the wrapper methods (although the ymake it nicer to read and you don't need the cast of return value). As there is no security involved, one can have those VarHandles as public static fields in some utility class: https://lucene.apache.org/core/9_7_0/core/org/apache/lucene/util/BitUtil.html; Usage of them is quite simple then: https://github.com/apache/lucene/blob/59c56a0aed9a43d24c676376b5d50c5c6518e3bc/lucene/core/src/java/org/apache/lucene/store/ByteArrayDataInput.java#L96 (there are many of those throughout Lucene's code)

So I agree with your ideas, we have to decide what is best for this PR. I tend to think that those 2 options are good:
- Use ByteBuffer in classfile API
- commit the PR as proposed here (looks fine to me).

minborg:
In my opinion, reducing startup time can be achieved in other ways, both short-term (sharing VHs) and long-term (pre-generate VHs at build time using condensers). I think having a supported API carries some advantages rather than using Unsafe directly.

PR #14640, opened 2023-06-25
8310848: Convert ClassDesc and MethodTypeDesc to be stored in static final fields

liach:
This would encourage Classfile API users to use the descriptors as constants, which can improve performance by avoiding repeated validation and reusing cached descriptor strings for MethodTypeDesc. This patch updates usages in the main codebase and benchmarks; tests are left untouched.

### DIFF
--- a/src/java.base/share/classes/jdk/internal/classfile/components/snippet-files/PackageSnippets.java
+++ b/src/java.base/share/classes/jdk/internal/classfile/components/snippet-files/PackageSnippets.java
@@ -25,6 +25,7 @@
 import java.lang.constant.ClassDesc;
 import java.lang.constant.ConstantDesc;
 
+import java.lang.constant.ConstantDescs;
 import java.lang.reflect.AccessFlag;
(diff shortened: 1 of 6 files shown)


liach:
@asotona Can you review this patch that migrates MethodTypeDesc creation to static final field access whenever possible? This allows to reuse cached descriptor strings and avoid revalidation. In addition, I dropped the jdkTreePrimitive benchmark for jdk.classfile.Write, which is identical to jdkTree as far as I see.

asotona:
This code replacing repeated symbols instantiation with static fields looks good.

However the patch also contains significant refactoring of jdk.jlink SystemModulesPlugin
and significant refactoring of the JMH benchmarks Write class.

I suggest to split this PR or link additional issues to it.

There is missing description of the problems so the proposed solution cannot be reviewed.

liach:
I have reverted the changes to SystemModulesPlugin and Write. Can you review this patch again?

asotona (approved):
It looks good 👍

PR #14641, opened 2023-06-25
8310829: guarantee(!HAS_PENDING_EXCEPTION) failed in ExceptionTranslation::doit

dougxc:
The VMSupport class is required for translating an exception between the HotSpot and libgraal heaps.
Loading it lazily can result in a loading exception, obscuring the exception being translated.
To avoid this, VMSupport is loaded eagerly along with the other vmClasses.

---------
### Progress

### Issue
 * [JDK-8310829](https://bugs.openjdk.org/browse/JDK-8310829): guarantee(!HAS_PENDING_EXCEPTION) failed in ExceptionTranslation::doit (**Bug** - P2)

### Reviewers
 * [Tom Rodriguez](https://openjdk.org/census#never) (@tkrodriguez - **Reviewer**) ⚠️ Review applies to [e46a6a17](https://git.openjdk.org/jdk/pull/14641/files/e46a6a17d4220d49c77208449ee3a00751dacdb9)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14641/head:pull/14641` \
`$ git checkout pull/14641`

`$ git checkout pull/14641` \
`$ git pull https://git.openjdk.org/jdk.git pull/14641/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14641`

`$ git pr show -t 14641`

</details>

<a href="https://git.openjdk.org/jdk/pull/14641.diff">https://git.openjdk.org/jdk/pull/14641.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14641#issuecomment-1605903368)


src/hotspot/share/jvmci/jvmciCompilerToVM.cpp
```diff
@@ -582,7 +582,7 @@ C2V_VMENTRY_NULL(jobject, lookupType, (JNIEnv* env, jobject, jstring jname, ARGU
   TempNewSymbol class_name = SymbolTable::new_symbol(str);
 
   if (class_name->utf8_length() <= 1) {
-    JVMCI_THROW_MSG_0(InternalError, err_msg("Primitive type %s should be handled in Java code", class_name->as_C_string()));
+    JVMCI_THROW_MSG_0(InternalError, err_msg("Primitive type %s should be handled in Java code", str));
```
dholmes-ora, line 585: Seems unrelated to the fix at hand.

dholmes-ora (commented):
The eager loading seems reasonable, but I do not understand the details here. In what way was loading failing? You still have to initialize `VMSupport` before you can call methods on it, so that could also fail - though the code does not seem to notice/handle this. ??

src/hotspot/share/jvmci/jvmciCompilerToVM.cpp
dougxc: Yes, it's a minor fix up I noticed while making changes a few lines below. It just avoids a conversion of a `Symbol` back to a C string when the original C string is available.

dougxc:
> The eager loading seems reasonable, but I do not understand the details here. In what way was loading failing? You still have to initialize `VMSupport` before you can call methods on it, so that could also fail - though the code does not seem to notice/handle this. ??

The usages of `vmSupport` below all use `JavaCalls:call_static` which takes care of linking and initializing the class.

dholmes-ora:
> The usages of `vmSupport` below all use `JavaCalls:call_static` which takes care of linking and initializing the class.

Sure but my point is that initialization can fail and the code does not seem to directly take that into account - if the first `decode` encounters a class initialization error then it will just return with the exception pending and no decoding will actually have occurred. If we get to the `encode` first and it encounters an initialization error it will still attempt to do a `decode` that must in turn fail and again we return with an exception pending and no encoding/decoding occurring. This code does not seem to be written in a  way that allows for an error initializing `VMSupport`. So while the fix side-steps the problematic guarantee, I think it is just highlighting that this code is prepared for "system" errors like this.

dholmes-ora:
Edited above to correct: I think it is just highlighting that this code **not** is prepared for "system" errors like this.

dougxc:
> if the first decode encounters a class initialization error then it will just return with the exception pending and no decoding will actually have occurred

That's fine - if VMSupport fails class initialization, then no "rich" decoding can occur (by definition) and so throwing the clinit error is the best we can do.

> If we get to the encode first and it encounters an initialization error it will still attempt to do a decode that must in turn fail 

You have to keep in mind that `encode` and `decode` are called in different runtimes. If encoding an exception in the HotSpot runtime fails (e.g. due to an OOME calling `VMSupport.<clinit>` in the HotSpot heap), then it's still fine to try call `VMSupport.decode` in the libgraal runtime. If `VMSupport.decode` in the libgraal runtime also fails, then it throw the exception describing the secondary failure. There's simply no way to guarantee all info is shown when secondary issues occur during exception handling.

dougxc:
BTW, I've manually tested this with libgraal. It's not possible to add a jtreg test until libgraal in part of OpenJDK.

tkrodriguez (approved):
looks ok to me.

dholmes-ora:
> That's fine - if VMSupport fails class initialization, then no "rich" decoding can occur (by definition) and so throwing the clinit error is the best we can do.

Then why did you not simply handle the exception from the loading of VMSupport the same way?

dougxc:
> Then why did you not simply handle the exception from the loading of VMSupport the same way?

The only actual case seen of the original way failing is due to OOME in GC stress tests. If loading of VMSupport is done during VM startup, then no stress test code can cause an OOME while loading VMSupport. I doubt `VMSupport.<clinit>` would ever fail in practice.

dholmes-ora:
>  I doubt VMSupport.<clinit> would ever fail in practice.

I would think it is likely to fail with OOME under the same GC stress test conditions.

But if this is just a stress test issue then bailing out when the loading fails would be far simpler than pre-loading the class. You've added to the VM startup cost just to avoid a stress test failure.

dougxc:
> I would think it is likely to fail with OOME under the same GC stress test conditions.
> 
> But if this is just a stress test issue then bailing out when the loading fails would be far simpler than pre-loading the class. You've added to the VM startup cost just to avoid a stress test failure.

Any app running near the GC limit can cause this and we've seen this with libgraal in practice. When the VM is near the GC limit, any compilation can cause an OOME and then when that OOME wants to be translated back to libgraal, it can be the first time VMSupport is used.
The time for loading `VMSupport` eagerly is in the noise.
For example, `-verbose:class` shows that about 30 classes (including `VMSupport`) are loaded in 1ms:
```
[0.011s][info][class,load] java.lang.reflect.Executable source: shared objects file
[0.012s][info][class,load] java.lang.reflect.Method source: shared objects file
[0.012s][info][class,load] java.lang.reflect.Constructor source: shared objects file
[0.012s][info][class,load] jdk.internal.vm.ContinuationScope source: shared objects file
[0.012s][info][class,load] jdk.internal.vm.Continuation source: shared objects file
[0.012s][info][class,load] jdk.internal.vm.StackChunk source: shared objects file
[0.012s][info][class,load] jdk.internal.vm.VMSupport source: shared objects file
[0.012s][info][class,load] jdk.internal.reflect.MagicAccessorImpl source: shared objects file
[0.012s][info][class,load] jdk.internal.reflect.MethodAccessor source: shared objects file
[0.012s][info][class,load] jdk.internal.reflect.MethodAccessorImpl source: shared objects file
[0.012s][info][class,load] jdk.internal.reflect.DelegatingClassLoader source: shared objects file
[0.012s][info][class,load] jdk.internal.reflect.ConstantPool source: shared objects file
[0.012s][info][class,load] java.lang.annotation.Annotation source: shared objects file
[0.012s][info][class,load] jdk.internal.reflect.CallerSensitive source: shared objects file
[0.012s][info][class,load] jdk.internal.reflect.ConstructorAccessor source: shared objects file
[0.012s][info][class,load] jdk.internal.reflect.ConstructorAccessorImpl source: shared objects file
[0.012s][info][class,load] jdk.internal.reflect.DirectConstructorHandleAccessor$NativeAccessor source: shared objects file
[0.012s][info][class,load] jdk.internal.reflect.SerializationConstructorAccessorImpl source: shared objects file
[0.012s][info][class,load] java.lang.invoke.MethodHandle source: shared objects file
[0.012s][info][class,load] java.lang.invoke.DirectMethodHandle source: shared objects file
[0.012s][info][class,load] java.lang.invoke.VarHandle source: shared objects file
[0.012s][info][class,load] java.lang.invoke.MemberName source: shared objects file
[0.012s][info][class,load] java.lang.invoke.ResolvedMethodName source: shared objects file
[0.012s][info][class,load] java.lang.invoke.MethodHandleNatives source: shared objects file
[0.012s][info][class,load] java.lang.invoke.LambdaForm source: shared objects file
[0.012s][info][class,load] java.lang.invoke.TypeDescriptor$OfMethod source: shared objects file
[0.012s][info][class,load] java.lang.invoke.MethodType source: shared objects file
[0.012s][info][class,load] java.lang.BootstrapMethodError source: shared objects file
[0.012s][info][class,load] java.lang.invoke.CallSite source: shared objects file
[0.012s][info][class,load] jdk.internal.foreign.abi.NativeEntryPoint source: shared objects file
[0.012s][info][class,load] jdk.internal.foreign.abi.ABIDescriptor source: shared objects file
[0.012s][info][class,load] jdk.internal.foreign.abi.VMStorage source: shared objects file
[0.013s][info][class,load] jdk.internal.foreign.abi.UpcallLinker$CallRegs source: shared objects file
```

dholmes-ora:
It may be in the noise but noise adds up over time.

It just seems to me that the simplest fix here would have been to convert
```
Klass* vmSupport = SystemDictionary::resolve_or_fail(vmSymbols::jdk_internal_vm_VMSupport(), true, THREAD);
guarantee(!HAS_PENDING_EXCEPTION, "");
```
to
```
Klass* vmSupport = SystemDictionary::resolve_or_fail(vmSymbols::jdk_internal_vm_VMSupport(), true, CHECK);
```
and just return on exception. A very isolated change with zero impact on anything else.

dougxc:
I've pushed the loading of `VMSupport` down into the 2 cases where it's actually needed and made it lazy again which should address all concerns about extra noise at VM startup. Thanks for the push back on this - I agree that it's the best solution.

tkrodriguez:
I don't think pushing it down that far improves things.  encode always precedes decode so why not resolve it before the encode call.
```
      // Resolve VMSupport class explicitly as HotSpotJVMCI::compute_offsets
      // may not have been called.
      Klass* vmSupport = SystemDictionary::resolve_or_fail(vmSymbols::jdk_internal_vm_VMSupport(), true, THREAD);
      int res = 0;
      if (!HAS_PENDING_EXCEPTION) {
        res = encode(THREAD, vmSupport, buffer, buffer_size);
      }
```

src/hotspot/share/jvmci/jvmciEnv.cpp
```diff
@@ -382,40 +382,45 @@ class ExceptionTranslation: public StackObj {
...
       jlong buffer = (jlong) NEW_RESOURCE_ARRAY_IN_THREAD_RETURN_NULL(THREAD, jbyte, buffer_size);
       if (buffer == 0L) {
-        decode(THREAD, vmSupport, 0L);
+        JVMCI_event_1("error translating exception: translation buffer allocation failed");
+        decode(THREAD, -1, 0L);
         return;
       }
-      int res = encode(THREAD, vmSupport, buffer, buffer_size);
+      int res = encode(THREAD, buffer, buffer_size);
       if (_from_env != nullptr && !_from_env->is_hotspot() && _from_env->has_pending_exception()) {
```
tkrodriguez, line 402: Why is this check before the `HAS_PENDING_EXCEPTION` check?  Couldn't you end up returning with a pending exception in this path?

dougxc: If `encode` is `SharedLibraryToHotSpotExceptionTranslation::encode` there is no possibility of a pending HotSpot exception upon it returning. If it's `HotSpotToSharedLibraryExceptionTranslation::encode` then `_from_env->is_hotspot()` is `true` and so this `if` branch is not taken.

dougxc:
> I don't think pushing it down that far improves things. encode always precedes decode so why not resolve it before the encode call.

Because the `VMSupport` klass is only needed if calling into HotSpot so it's better to push it down into `HotSpotToSharedLibraryExceptionTranslation::encode`. Also, if the `VMSupport` klass is used for encoding, it's *not* needed for decoding (the libgraal `VMSupport` jclass value is used instead).

tkrodriguez:
Right, I was forgetting about the virtual nature of this code.  Why can't the pending exception handling be in the respective virtual?  The partition between virtual and shared is kind of ugly.

dougxc:
> Why can't the pending exception handling be in the respective virtual?

Done: 9236128ad1b7297c8c4e9d3022b88c3ab3278022

tkrodriguez (approved):
That's more clear to me.  Thanks!

dholmes-ora (commented):
This is now isolated to JVMCI code that I'm not familiar with, so don't feel competent to review. The general idea seems okay but I'm not familiar enough with the details. Someone from the compiler team should review this now.

Thanks.

dougxc:
> Someone from the compiler team should review this now.

@vnkozlov could you please review this or nominate someone else from the compiler team to look at it. Thanks.

vnkozlov:
I am fins with idea of changes. But, please, activate GHA testing for this branch.
And there is build error on Windows:
```
c:\workspace\open\src\hotspot\share\jvmci\jvmciEnv.cpp(449): error C2220: the following warning is treated as an error
c:\workspace\open\src\hotspot\share\jvmci\jvmciEnv.cpp(449): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
```

dougxc:
I have fixed the warning on Windows: 5bb3b529d36c906ac861e5ebf1b861dbb35bfe2c

> But, please, activate GHA testing for this branch.

Isn't GHA a strict subset of or equal to tier1 mach5 testing? If so, what's the point of doing redundant testing?

vnkozlov:
> > But, please, activate GHA testing for this branch.
> 
> Isn't GHA a strict subset of or equal to tier1 mach5 testing? If so, what's the point of doing redundant testing?

It builds and tests configurations (32-bit) we don't have in our testing.

vnkozlov (approved):
Good.

dougxc:
> > > But, please, activate GHA testing for this branch.
> > 
> > 
> > Isn't GHA a strict subset of or equal to tier1 mach5 testing? If so, what's the point of doing redundant testing?
> 
> It builds and tests configurations (32-bit) we don't have in our testing.

Good to know - thanks!

dougxc:
Thanks for the reviews.

/integrate

PR #14642, opened 2023-06-25
8310849: Pattern matching for instanceof and arrayType cleanup in j.l.invoke and j.l.reflect

liach:
This patch touches java.lang.reflect and java.lang.invoke packages. It replaces instanceof + cast with pattern matching and updates  Array.newInstance().getClass() patterns with arrayType() for obtaining array types of a class.

### DIFF
--- a/src/java.base/share/classes/java/lang/invoke/MemberName.java
+++ b/src/java.base/share/classes/java/lang/invoke/MemberName.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -135,8 +135,8 @@ final class MemberName implements Member, Cloneable {
         {
             // Get a snapshot of type which doesn't get changed by racing threads.
             final Object type = this.type;
-            if (type instanceof MethodType) {
-                return (MethodType) type;
+            if (type instanceof MethodType mt) {
+                return mt;
             }
         }
 
@@ -173,8 +173,8 @@ final class MemberName implements Member, Cloneable {
 
         // Get a snapshot of type which doesn't get changed by racing threads.
         final Object type = this.type;
-        if (type instanceof String) {
-            return (String) type;
+        if (type instanceof String str) {
+            return str;
         } else {
             return getMethodType().toMethodDescriptorString();
         }
@@ -211,8 +211,8 @@ final class MemberName implements Member, Cloneable {
         {
             // Get a snapshot of type which doesn't get changed by racing threads.
             final Object type = this.type;
-            if (type instanceof Class<?>) {
-                return (Class<?>) type;
+            if (type instanceof Class<?> cl) {
+                return cl;
             }
         }
 
@@ -725,7 +725,7 @@ final class MemberName implements Member, Cloneable {
 
     @Override
     public boolean equals(Object that) {
-        return (that instanceof MemberName && this.equals((MemberName)that));
+        return that instanceof MemberName mn && this.equals(mn);
     }
 
     /** Decide if two member names have exactly the same symbolic content.
@@ -808,8 +808,8 @@ final class MemberName implements Member, Cloneable {
     void checkForTypeAlias(Class<?> refc) {
         if (isInvocable()) {
             MethodType type;
-            if (this.type instanceof MethodType)
-                type = (MethodType) this.type;
+            if (this.type instanceof MethodType mt)
+                type = mt;
             else
                 this.type = type = getMethodType();
             if (type.erase() == type)  return;
@@ -817,8 +817,8 @@ final class MemberName implements Member, Cloneable {
             throw new LinkageError("bad method type alias: "+type+" not visible from "+refc);
         } else {
             Class<?> type;
-            if (this.type instanceof Class<?>)
-                type = (Class<?>) this.type;
+            if (this.type instanceof Class<?> cl)
+                type = cl;
             else
                 this.type = type = getFieldType();
             if (VerifyAccess.isTypeVisible(type, refc))  return;
@@ -863,8 +863,8 @@ final class MemberName implements Member, Cloneable {
         return buf.toString();
     }
     private static String getName(Object obj) {
-        if (obj instanceof Class<?>)
-            return ((Class<?>)obj).getName();
+        if (obj instanceof Class<?> cl)
+            return cl.getName();
         return String.valueOf(obj);
     }
 
@@ -915,8 +915,8 @@ final class MemberName implements Member, Cloneable {
             ex = new NoSuchMethodException(message);
         else
             ex = new NoSuchFieldException(message);
-        if (resolution instanceof Throwable)
-            ex.initCause((Throwable) resolution);
+        if (resolution instanceof Throwable res)
+            ex.initCause(res);
         return ex;
     }
 
@@ -992,7 +992,7 @@ final class MemberName implements Member, Cloneable {
             if (result.isResolved())
                 return result;
             ReflectiveOperationException ex = result.makeAccessException();
-            if (ex instanceof IllegalAccessException)  throw (IllegalAccessException) ex;
+            if (ex instanceof IllegalAccessException iae)  throw iae;
             throw nsmClass.cast(ex);
         }
         /** Produce a resolved version of the given member.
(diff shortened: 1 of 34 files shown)


src/java.base/share/classes/sun/reflect/generics/reflectiveObjects/GenericArrayTypeImpl.java
```diff
@@ -70,9 +70,7 @@ public String toString() {
 
     @Override
     public boolean equals(Object o) {
-        if (o instanceof GenericArrayType) {
-            GenericArrayType that = (GenericArrayType) o;
-
+        if (o instanceof GenericArrayType that) {
```
Glavo, line 73:
```java
return o instanceof GenericArrayType that 
    && Objects.equals(genericComponentType, that.getGenericComponentType());
```

src/java.base/share/classes/sun/reflect/generics/reflectiveObjects/WildcardTypeImpl.java
```diff
@@ -168,8 +168,7 @@ public String toString() {
 
     @Override
     public boolean equals(Object o) {
-        if (o instanceof WildcardType) {
-            WildcardType that = (WildcardType) o;
+        if (o instanceof WildcardType that) {
```
Glavo, line 171:
```java
return o instanceof WildcardType that
        && Arrays.equals(this.getLowerBounds(), that.getLowerBounds()) 
        && Arrays.equals(this.getUpperBounds(), that.getUpperBounds());
```

src/java.base/share/classes/sun/reflect/annotation/AnnotatedTypeFactory.java
```diff
@@ -250,12 +249,11 @@ int baseHashCode() {
 
         @Override
         public boolean equals(Object o) {
-            if (o instanceof AnnotatedType &&
+            if (o instanceof AnnotatedType that &&
```
Glavo, line 252:
```java
return o instanceof AnnotatedType that &&
     !(o instanceof AnnotatedArrayType) &&
     !(o instanceof AnnotatedTypeVariable) &&
     !(o instanceof AnnotatedParameterizedType) &&
     !(o instanceof AnnotatedWildcardType) &&
     equalsTypeAndAnnotations(that);
```

src/java.base/share/classes/sun/invoke/util/Wrapper.java
```diff
@@ -580,9 +580,9 @@ public Object wrap(int x) {
     }
 
     private static Number numberValue(Object x) {
-        if (x instanceof Number)     return (Number)x;
-        if (x instanceof Character)  return (int)(Character)x;
-        if (x instanceof Boolean)    return (Boolean)x ? 1 : 0;
+        if (x instanceof Number n)     return n;
```
Glavo, line 583:
```java
return switch (x) {
    case Number n -> n;
    case Character c -> (int) c;
    case Boolean b -> b ? 1 : 0;
    // Remaining allowed case of void: Must be a null reference.
    case null, default -> (Number) x;
};
```

Glavo (changes requested):

src/java.base/share/classes/java/lang/invoke/MethodHandleNatives.java
```diff
@@ -350,10 +350,10 @@ static Object linkDynamicConstantImpl(Class<?> caller,
...
-            return "BSA="+java.util.Arrays.asList((Object[]) staticArguments);
-        if (staticArguments instanceof int[])
-            return "BSA@"+java.util.Arrays.toString((int[]) staticArguments);
+        if (staticArguments instanceof Object[] array)
+            return "BSA="+java.util.Arrays.asList(array);
+        if (staticArguments instanceof int[] array)
+            return "BSA@"+java.util.Arrays.toString(array);
         if (staticArguments == null)
             return "BSA0=null";
         return "BSA1="+staticArguments;
```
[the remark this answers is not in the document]
liach: Nice observation and suggestions that I've totally missed, but unfortunately these code is used to set up boostrap methods, so pattern matching switches, which are implemented with bootstrap method [`SwitchBootstraps::typeSwitch`](https://github.com/openjdk/jdk/blob/4f5829e3433270d21f75ee949fbc056ae309a494/src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java#L144), cannot be used or it will lead to infinite loops like using lambdas in early startup code.

Meanwhile, even though Supplier, Consumer etc. are loaded as collections are loaded very early in startup, lambdas in their implementation methods are not called until j.l.invoke is ready, so they are safe. It's just some early startup shenanigans. Same story for early-startup records: https://github.com/openjdk/jdk/blob/4f5829e3433270d21f75ee949fbc056ae309a494/src/java.base/share/classes/jdk/internal/reflect/ReflectionFactory.java#L574-L578

src/java.base/share/classes/sun/invoke/util/ValueConversions.java
```diff
@@ -236,12 +236,12 @@ public static Number primitiveConversion(Wrapper wrap, Object x, boolean cast) {
             if (!cast)  return null;
             return ZERO_INT;
         }
-        if (x instanceof Number) {
-            res = (Number) x;
-        } else if (x instanceof Boolean) {
-            res = ((boolean)x ? ONE_INT : ZERO_INT);
-        } else if (x instanceof Character) {
-            res = (int)(char)x;
+        if (x instanceof Number n) {
```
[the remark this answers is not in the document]
liach: Unfortunately, this is used by MethodHandleImpl, so can't use indy switches.

src/java.base/share/classes/sun/invoke/util/Wrapper.java
liach: This is used by Wrapper.convert and MethodHandles.insertArgumentPrimitive, which I don't think is a safe candidate either.

src/java.base/share/classes/sun/invoke/util/ValueConversions.java
```diff
@@ -259,16 +259,16 @@ public static Number primitiveConversion(Wrapper wrap, Object x, boolean cast) {
...
-            return (int) x;
-        else if (x instanceof Boolean)
-            return fromBoolean((boolean) x);
-        else if (x instanceof Character)
-            return (char) x;
-        else if (x instanceof Short)
-            return (short) x;
-        else if (x instanceof Byte)
-            return (byte) x;
+        if (x instanceof Integer i)
```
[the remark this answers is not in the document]
liach: Also used by MethodHandles.insertArgumentPrimitive, likely not safe.

mlchung (approved):
Looks okay to me.

jddarcy (approved):

src/java.base/share/classes/java/lang/invoke/MemberName.java
```diff
@@ -992,7 +992,7 @@ MemberName resolveOrFail(byte refKind, MemberName m,
             if (result.isResolved())
                 return result;
             ReflectiveOperationException ex = result.makeAccessException();
-            if (ex instanceof IllegalAccessException)  throw (IllegalAccessException) ex;
+            if (ex instanceof IllegalAccessException iae)  throw iae;
```
turbanoff, line 995:
```suggestion
            if (ex instanceof IllegalAccessException iae) throw iae;
```

src/java.base/share/classes/java/lang/invoke/MethodHandleStatics.java
```diff
@@ -185,8 +185,8 @@ static RuntimeException newIllegalArgumentException(String message, Object obj,
     /** Propagate unchecked exceptions and errors, but wrap anything checked and throw that instead. */
     /*non-public*/
     static Error uncaughtException(Throwable ex) {
-        if (ex instanceof Error)  throw (Error) ex;
-        if (ex instanceof RuntimeException)  throw (RuntimeException) ex;
+        if (ex instanceof Error error)  throw error;
+        if (ex instanceof RuntimeException re)  throw re;
```
turbanoff, line 189:
```suggestion
        if (ex instanceof Error error) throw error;
        if (ex instanceof RuntimeException re) throw re;
```

src/java.base/share/classes/java/lang/invoke/MemberName.java
liach: Don't know why, but sometimes JDK devs leaves 2 spaces for certain breaks (like Joe Darcy tends to leave 2 for sentence breaks in Javadocs). Can anyone explain why? In MemberName in particular, I see a mix of 2 and 1 spaces for single-line if statements, so I will take your suggestion.

src/java.base/share/classes/java/lang/invoke/MethodType.java
```diff
@@ -893,14 +893,11 @@ public boolean equals(Object x) {
...
-            return equals((MethodType)x);
+        if (x instanceof MethodType mt) {
+            return equals(mt);
         }
-        if (x instanceof ConcurrentWeakInternSet.WeakEntry) {
-            Object o = ((ConcurrentWeakInternSet.WeakEntry)x).get();
-            if (o instanceof MethodType) {
-                return equals((MethodType)o);
-            }
+        if (x instanceof ConcurrentWeakInternSet.WeakEntry<?> e && e.get() instanceof MethodType mt) {
```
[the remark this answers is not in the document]
liach: Done. Thanks for the review and suggestion.

liach:
Tests pass, integrating. Thank you for your reviews!

/integrate

PR #14644, opened 2023-06-26
JDK-8310861: Improve location reporting for javac serial lint warnings

jddarcy:
For a subset of the expanded serial lint warnings ([JDK-8202056](https://bugs.openjdk.org/browse/JDK-8202056), [JDK-8160675](https://bugs.openjdk.org/browse/JDK-8160675)), javac uses the location of the enclosing type rather than the method or field in question.

This PR addresses this by using the idiom of finding the method/field's tree within the type tree being checked.

All langtools regression tests pass when run locally; the intention is to backport this fix to JDK 21 as well.


jddarcy:
PS To expand "negative" testing, I could add more `@compile` lines to verify no warnings are generated without serial warnings being enabled -- or revert the few such changes like that already added.

lahodaj (approved):
Looks good to me.

PR #14653, opened 2023-06-26
8310890: Normalize identifier names

pavelrappo:
Please review this cleanup PR to normalize names of identifiers which are Java variables/fields or tokens in text files. Those names either contain a pronoun that is very rarely used in code, or seem like they contain such a pronoun, which, in fact, they don't. Either way, the goal is to improve readability and clarity.

Also, this PR fixes a few related typos.

---------
### Progress

### Issue
 * [JDK-8310890](https://bugs.openjdk.org/browse/JDK-8310890): Normalize identifier names (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14653/head:pull/14653` \
`$ git checkout pull/14653`

`$ git checkout pull/14653` \
`$ git pull https://git.openjdk.org/jdk.git pull/14653/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14653`

`$ git pr show -t 14653`

</details>

<a href="https://git.openjdk.org/jdk/pull/14653.diff">https://git.openjdk.org/jdk/pull/14653.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14653#issuecomment-1607966307)

### DIFF
--- a/make/data/charsetmapping/charsets
+++ b/make/data/charsetmapping/charsets
@@ -34,7 +34,7 @@
 # be packaged into the charsets provider class "StandardCharsets"
 # which is initialized at startup time by java.nio.charset.Charset,
 # compared to the charsets packaged in "ExtendedCharsets" provider,
-# which is lazy initialized.
+# which is lazily initialized.
 ########################################################
 
 charset US-ASCII US_ASCII
@@ -146,7 +146,7 @@ charset ISO-8859-1 ISO_8859_1
 charset ISO-8859-2 ISO_8859_2
     package sun.nio.cs
     type    sbcs
-    hisname ISO8859_2
+    histname ISO8859_2
     ascii   true
     alias   iso8859_2			# JDK historical
     alias   8859_2
@@ -165,7 +165,7 @@ charset ISO-8859-2 ISO_8859_2
 charset ISO-8859-4 ISO_8859_4
     package sun.nio.cs
     type    sbcs
-    hisname ISO8859_4
+    histname ISO8859_4
     ascii   true
     alias   iso8859_4			# JDK historical
     alias   iso8859-4
@@ -184,7 +184,7 @@ charset ISO-8859-4 ISO_8859_4
 charset ISO-8859-5 ISO_8859_5
     package sun.nio.cs
     type    sbcs
-    hisname ISO8859_5
+    histname ISO8859_5
     ascii   true
     alias   iso8859_5			# JDK historical
     alias   8859_5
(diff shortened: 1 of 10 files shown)


naotoj (approved):
Looks fine to me

src/java.base/share/classes/java/util/EnumMap.java
```diff
@@ -685,9 +685,9 @@ private boolean equals(EnumMap<?,?> em) {
         // Key types match, compare each value
         for (int i = 0; i < keyUniverse.length; i++) {
             Object ourValue =    vals[i];
-            Object hisValue = em.vals[i];
-            if (hisValue != ourValue &&
-                (hisValue == null || !hisValue.equals(ourValue)))
+            Object otherValue = em.vals[i];
+            if (otherValue != ourValue &&
+                (otherValue == null || !otherValue.equals(ourValue)))
```
RogerRiggs, line 690: Is this the same as java.util.Objects:  
  `!Objects.equals(vals[i], em.vals[i]);`

RogerRiggs (approved):
Looks good, with or without the suggestion.

make/data/charsetmapping/charsets
```diff
@@ -146,7 +146,7 @@ charset ISO-8859-1 ISO_8859_1
 charset ISO-8859-2 ISO_8859_2
     package sun.nio.cs
     type    sbcs
-    hisname ISO8859_2
+    histname ISO8859_2
```
jensli, line 149: Should this column be re-aligned with the longer name?

src/java.base/share/classes/java/util/EnumMap.java
pavelrappo: You are right: we can fold this if-else construct into that utility method. That said, it in *this* PR, I'd prefer not to.

The reason is that I've been preparing a much bigger PR that uses Objects.equals as well as other utility methods and modern language features to improve some equals, hashCode, and compareTo implementations across java.base. I'm planning to publish that PR soon. In fact, this change to Java variable names was cherry-picked from that bigger PR, to separate trivial renames from code changes.

make/data/charsetmapping/charsets
pavelrappo: I thought about it before publishing the PR. I decided not to re-align anything because (i) the change would be bigger and (ii) the fact that there was already a property that is similarly misaligned; search for:

    internal true

pavelrappo: If you are concerned with functionality rather than looks, then I can tell you this:

1. The build succeeds and tier1 tests pass.
2. The code that parses that file expect one or more whitespace characters as a separator:

        String[] tokens = line.split("\\s+");

PR #14654, opened 2023-06-26
JDK-8311009: Long.toUnsignedString(long, int) doesn't have to create a BigInteger

yuantj:
This PR changes the implementation of `Long.toUnsignedString(long, int)` for "default" radices, which avoids creating a `BigInteger` and makes use of `Long.divideUnsigned` and `Long.remainderUnsigned`.

I've run the test on `test/jdk/java/lang/Long/Unsigned.java` and it works correctly. I believe that there is no need to add more test cases.

I've added a benchmark case to `test/micro/org/openjdk/bench/java/lang/Longs.java` for this method. Here is the benchmark result tested on my machine:

Before (JDK 20.0.1):

```plain
Benchmark                       (size)  Mode  Cnt     Score    Error  Units
Longs.toUnsignedStringNegative     500  avgt   15  6428.711 ± 63.142  us/op
```

After:

```plain
Benchmark                       (size)  Mode  Cnt     Score     Error  Units
Longs.toUnsignedStringNegative     500  avgt   15  3823.655 ± 146.171  us/op
```

No CSR needed since the behavior is not changed.

---------
### Progress

### Issue
 * [JDK-8311009](https://bugs.openjdk.org/browse/JDK-8311009): Long.toUnsignedString(long, int) doesn't have to create a BigInteger (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14654/head:pull/14654` \
`$ git checkout pull/14654`

`$ git checkout pull/14654` \
`$ git pull https://git.openjdk.org/jdk.git pull/14654/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14654`

`$ git pr show -t 14654`

</details>

<a href="https://git.openjdk.org/jdk/pull/14654.diff">https://git.openjdk.org/jdk/pull/14654.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14654#issuecomment-1610640728)

### DIFF
--- a/src/java.base/share/classes/java/lang/Long.java
+++ b/src/java.base/share/classes/java/lang/Long.java
@@ -215,6 +215,9 @@ public final class Long extends Number
         if (i >= 0)
             return toString(i, radix);
         else {
+            if (radix < Character.MIN_RADIX || radix > Character.MAX_RADIX) {
+                radix = 10;
+            }
             return switch (radix) {
                 case 2  -> toBinaryString(i);
                 case 4  -> toUnsignedString0(i, 2);
@@ -234,7 +237,11 @@ public final class Long extends Number
                 }
                 case 16 -> toHexString(i);
                 case 32 -> toUnsignedString0(i, 5);
-                default -> toUnsignedBigInteger(i).toString(radix);
+                default -> {
+                    long leadingDigits = divideUnsigned(i, radix);  // always positive
+                    int lastDigit = (int)remainderUnsigned(i, radix);
+                    yield toString(leadingDigits, radix) + Integer.digits[lastDigit];
+                }
             };
         }
     }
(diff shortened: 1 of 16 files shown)


yuantj:
Please ignore the merge warning from the bot.

rgiulietti:
The change looks good (but I'm not an official reviewer).
Please update the copyright year in the micro benchmark.

yuantj:
BTW, do I need to remove `toUnsignedBigInteger` method (which is private and no longer used)?

rgiulietti:
I checked whether the method is ever searched for reflectively: this seems _not_ be the case in the JDK codebase. So I guess it can be removed from `Long.java`.

src/java.base/share/classes/java/lang/Long.java
```diff
@@ -234,7 +237,11 @@ public static String toUnsignedString(long i, int radix) {
                 }
                 case 16 -> toHexString(i);
                 case 32 -> toUnsignedString0(i, 5);
-                default -> toUnsignedBigInteger(i).toString(radix);
+                default -> {
+                    long leadingDigits = divideUnsigned(i, radix);  // always positive
+                    int lastDigit = (int)remainderUnsigned(i, radix);
```
rgiulietti, line 242: You can avoid the additional division from `remainderUnsigned()` like so
```suggestion
                    int lastDigit = (int)(i - leadingDigits * radix);
```

PR #14655, opened 2023-06-26
8310901: Convert String::newStringNoRepl with Latin-1 to String::newStringLatin1NoRepl

Glavo:
Added a new method `newStringLatin1NoRepl` to the `JavaLangAccess`.

Reasons:

* Most use cases of `newStringNoRepl` use `ISO_8859_1` as the charset, creating a new shortcut can make writing shorter;
* Since all possible values of `byte` are legal Latin-1 characters, `newStringLatin1NoRepl` **will not throw `CharacterCodingException`**, so users can make the compiler happy without using useless try-catch statements.

---------
### Progress

### Issue
 * [JDK-8310901](https://bugs.openjdk.org/browse/JDK-8310901): Convert String::newStringNoRepl with Latin-1 to String::newStringLatin1NoRepl (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14655/head:pull/14655` \
`$ git checkout pull/14655`

`$ git checkout pull/14655` \
`$ git pull https://git.openjdk.org/jdk.git pull/14655/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14655`

`$ git pr show -t 14655`

</details>

<a href="https://git.openjdk.org/jdk/pull/14655.diff">https://git.openjdk.org/jdk/pull/14655.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14655#issuecomment-1607781991)


liach:
Since it appears you missed my edit to my comment, this part can be replaced by the newStringLatin1NoRepl as well:
https://github.com/openjdk/jdk/blob/5ff42d14294199eb3bf10b66530f9249fb68810d/src/java.base/share/classes/java/lang/String.java#L812-L814

after this block:
```java
        if (cd instanceof ArrayDecoder ad &&
                ad.isASCIICompatible() &&
                !StringCoding.hasNegatives(src, 0, src.length)) {
```

RogerRiggs:
I don't think this is a productive change. 
It takes a local inconvenience try/catch and spreads the impact over multiple files and packages as well as bulking up JLA.

Glavo:
> It takes a local inconvenience try/catch and spreads the impact over multiple files and packages as well as bulking up JLA.

I don't quite understand what you mean.

The main goal of this PR is to get rid of those annoying try-catch, but it looks like you misunderstood this PR?

My English is poor, please forgive me if I misunderstood this sentence of yours.

liach:
> I don't think this is a productive change. It takes a local inconvenience try/catch and spreads the impact over multiple files and packages as well as bulking up JLA.

By this rationale, newStringUTF8NoRepl should be removed as well.

RogerRiggs:
The goal of removing the try/catch from HexFormat can be solved locally by creating a private method in HexFormat that swallows the exception. There is no need to add a method to Java Lang Access or change 3 other files.
The method `newStringUTF8NoRepl` is different, it can/does throw runtime exceptions but there is no need for try/catch in that case, no change is needed there.

Glavo:
> The goal of removing the try/catch from HexFormat can be solved locally by creating a private method in HexFormat that swallows the exception. There is no need to add a method to Java Lang Access or change 3 other files. The method `newStringUTF8NoRepl` is different, it can/does throw runtime exceptions but there is no need for try/catch in that case, no change is needed there.

The different functions of `newStringNoRepl` from public APIs are: 1) It will try not to copy the byte array as much as possible; 2) It will throw an exception when there are  malformed or unmappable byte sequence.

Currently, `newStringNoRepl` has two use cases: 1) `HexFormat`; 2) `Files.readString`. If #14578 is merged, it will have a third use case.

If we review these use cases, we will realize that `newStringNoRepl` did not make any use cases happy:

* For `HexFormat` and [#14578](https://github.com/openjdk/jdk/pull/14578), `newStringNoRepl` is used to access `String(byte[],byte)` outside of `java.lang`.
  In this usage, `newStringNoRepl` is inconvenient to use. Users not only need to pass the charset, but also need to catch `CharacterCodingException`, even though it is actually impossible to occur.
  The purpose of this PR is to provide a better way of doing this.
* For `Files.readString`: Although `newStringNoRepl` was originally created for `Files.readString`, `Files.readString` has changed.
  Currently `Files.readString` is implemented as follows:
  
  https://github.com/openjdk/jdk/blob/7f094353673f5047643a2d7b512d0de8c665f215/src/java.base/share/classes/java/nio/file/Files.java#L3342-L3350

Obviously, for security reasons, additional check `path.getClass().getModule() != Object.class.getModule()` has been added here.

This makes `newStringNoRepl` awkward here. When reading a file containing non-ASCII characters from a non-default filesystem, an unnecessary copy of the byte array occurs here.

---

Since this method is so lame, I'm planning to do some retrofits:

1. Create a new method `newStringLatin1NoRepl`, use it to access `String(byte[],byte)` outside of `java.lang`;
2. Delete `newStringUTF8NoRepl` and add a new parameter `boolean noShare` to `newStringNoRepl`.

The new signature for `newStringNoRepl` looks like this:

```java
String newStringNoRepl(byte[] bytes, Charset cs, boolean noShare) throws CharacterCodingException;
```

`newStringUTF8NoRepl(bytes)` can be replaced with `newStringNoRepl(bytes, UTF_8, true)`.

`Files.readString` can be optimized to:

```java
return JLA.newStringNoRepl(readAllBytes(path), cs, path.getClass().getModule() != Object.class.getModule());
```

This is my idea. Can you take a look?

liach:
I've investigated the few xxNoRepl methods and think they are best coalesced together into newStringNoRepl and getBytesNoRepl taking byte arrays.

1. The exceptions to the NoRepl methods was added in [8205058](https://bugs.openjdk.org/browse/JDK-8205058) when the access method only catered to Files use cases; these are applicable nowhere else.
2. The UTF8 noRepl methods are sometimes implemented incorrectly, especially getBytesUTF8NoRepl, which always copies the array. They are added specifically for Zip handling as well.

Thus, I recommend the following changes:
1. Make the NoRepl methods throw IllegalArgument (like NoRepl1 do) by default; move the exception handling logic into Files implementation, the only place where it's actually needed.
2. Coalesce all charset-specific NoRepl methods in JLA into newStringNoRepl/getBytesNoRepl; the ones in String itself can stay, for they simplify code much.

liach:
On a side note, `NoRepl` means "no replication", implying the passed array is already trusted. I think you should do something like this instead:
```java
public static String readString(Path path, Charset cs) throws IOException { 
    Objects.requireNonNull(path); 
    Objects.requireNonNull(cs); 

    byte[] ba = readAllBytes(path); 
    if (path.getClass().getModule() != Object.class.getModule())
        return new String(ba, 0, ba.length, cs);
    return JLA.newStringNoRepl(ba, cs); 
} 
```

Glavo:
> On a side note, `NoRepl` means "no replication", implying the passed array is already trusted. I think you should do something like this instead:

I think `NoRepl` means "no replace", it's an ambiguous acronym.

It cannot be replaced by `new String(byte[], int, int, Charset)` because the constructor will replaces malformed-input and unmappable-character sequences with this charset's default replacement string.

I think throwing an exception when malformed or unmappable bytes is encountered is a useful behavior, and maybe a new public API should be provided to users.

liach:
See #16260: `NoRepl` means to report encoding error as CCE. The API you are creating should be renamed to something like `newTrustedLatin1String` as we don't check encoding errors and this API solely focuses on sharing the array.

PR #14656, opened 2023-06-26
8310902: (fc) FileChannel.transferXXX async close and interrupt issues

AlanBateman:
This change fixes a number of bugs with FileChannel.transferTo/transerFrom that arise when the source or target channel is closed, or the thread is interrupted, during the transfer. More specifically:

- The direct implementations operate on the raw file descriptor and don't prevent the target/source channel from being closed during the transfer. For this PR, the direct transferTo is limited to cases where the target is a FileChannel or SocketChannel. It could be extended to other SelChImpl implementations (SinkChannel, DatagramChannel) if needed, but they are less interesting at this time.

- The transferTo/transferFrom methods are specified to throw ClosedByInterruptException after closing both channels. This is implemented inconsistently, and leaves one channel open in several cases.

---------
### Progress

### Issue
 * [JDK-8310902](https://bugs.openjdk.org/browse/JDK-8310902): (fc) FileChannel.transferXXX async close and interrupt issues (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14656/head:pull/14656` \
`$ git checkout pull/14656`

`$ git checkout pull/14656` \
`$ git pull https://git.openjdk.org/jdk.git pull/14656/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14656`

`$ git pr show -t 14656`

</details>

<a href="https://git.openjdk.org/jdk/pull/14656.diff">https://git.openjdk.org/jdk/pull/14656.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14656#issuecomment-1608856083)


test/jdk/java/nio/channels/FileChannel/CloseDuringTransfer.java
```diff
@@ -0,0 +1,418 @@
...
+            assertTrue(src.isOpen());
+        }
+    }
+
+    /**
+     * Interrupt thread during transferFrom.
+     */
+    @ParameterizedTest
+    @MethodSource("sources")
+    void testInterruptTransferDuring(ReadableByteChannel src) throws Exception {
```
bplb, line 194: I think this method must've been intended to be named `testInterruptDuringTransferFrom`.

src/java.base/share/classes/sun/nio/ch/FileChannelImpl.java
```diff
@@ -537,106 +547,206 @@ public void force(boolean metaData) throws IOException {
     // Assume at first that the underlying kernel supports sendfile();
```
bplb, line 547: It's not just sendfile() any more so maybe this comment should be broadened.

src/java.base/share/classes/sun/nio/ch/FileChannelImpl.java
```diff
@@ -537,106 +547,206 @@ public void force(boolean metaData) throws IOException {
...
     // set this to true if we find out later that it doesn't
     //
-    private static volatile boolean transferToNotSupported;
+    private static volatile boolean transferToDirectNotSupported;
 
     // Assume that the underlying kernel sendfile() will work if the target
```
bplb, line 552: This comment could be broadened as well.

src/java.base/share/classes/sun/nio/ch/FileChannelImpl.java
AlanBateman: This PR doesn't change which syscalls are used but I agree that this comment is a bit outdated, we can make it sendfile/equivalent to avoid enumerating the different syscalls and implementations that can be used.

bplb: Sounds good.

test/jdk/java/nio/channels/FileChannel/CloseDuringTransfer.java
```diff
@@ -0,0 +1,416 @@
...
+            assertTrue(src.isOpen());
+        }
+    }
+
+    /**
+     * Interrupt thread during transferFrom.
+     */
+    @ParameterizedTest
+    @MethodSource("sources")
+    void testInterruptTransferDuringTransferFrom(ReadableByteChannel src) throws Exception {
```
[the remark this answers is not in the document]
AlanBateman: Is there a typo here, I can't quite see it.

src/java.base/share/classes/sun/nio/ch/FileChannelImpl.java
```diff
@@ -534,109 +544,215 @@ public void force(boolean metaData) throws IOException {
...
 
-    // Assume that the underlying kernel sendfile() will work if the target
-    // fd is a file; set this to false if we find out later that it doesn't
-    //
-    private static volatile boolean fileSupported = true;
+    /**
+     * Marks the beginning of a transfer to or from this channel.
+     * @throws ClosedChannelException if channel is closed
+     */
+    private int beforeTransfer() throws ClosedChannelException {
```
bplb, line 561: Should this method and `afterTransfer()` be moved up before the `*NotSupported` fields as they cover transferring both to and from?
```
    // beforeTransfer()
    // afterTransfer()
    // transferTo fields and methods
    // transferFrom fields and methods
```

AlanBateman: Alternatively, transferFromNotSupported; can be moved up so that the transfer flags are in one place.

src/java.base/share/classes/sun/nio/ch/FileChannelImpl.java
```diff
@@ -534,109 +544,215 @@ public void force(boolean metaData) throws IOException {
...
-            do {
-                long comp = Blocker.begin();
+            int sourceIndex = source.beforeTransfer();
+            try {
+                int targetIndex = target.beforeTransfer();
                 try {
-                    n = nd.transferTo(fd, position, icount, targetFD, append);
+                    long n = transferToFileDescriptor(position, count, target.fd);
+                    if (n == IOStatus.UNSUPPORTED_CASE) {
+                        transferToFileDirectNotSupported = true;
```
bplb, line 657: On Linux, `transferToFileDescriptor` will fail with `UNSUPPORTED_CASE` if the target is open for appending. I don't think that this should mean that all future fast transfers to a file should be suppressed.

src/java.base/share/classes/sun/nio/ch/FileChannelImpl.java
bplb: Either way.

src/java.base/share/classes/sun/nio/ch/FileChannelImpl.java
AlanBateman: I'll check it but all I've really done here is rename fileSupported to transferToFileDirectNotSupported and inverted the values.

bplb: In that case the situation did not change and whether it poses a problem could be investigated under another issue.

AlanBateman: I think the simplest thing is to remove fileSupported for now.  Also if we limit the get/set of transferToDirectNotSupported to transferToDirect, and limit the get/set of transfeFromDirectNotSupported to transferFromDirect then it removes any confusion as to when the direct cases are disabled. I've done that now and updated the PR

bplb (approved):
Last changes look good. Approved.

PR #14659, opened 2023-06-26
8310906: Fix -Wconversion warnings in runtime, oops and some code header files.

coleenp:
More casts for size of address differences in same code space, checked_casts<> and type changes to fix -Wconversion warnings.  See CR for details.
Tested with tier1-4.

---------
### Progress

### Issue
 * [JDK-8310906](https://bugs.openjdk.org/browse/JDK-8310906): Fix -Wconversion warnings in runtime, oops and some code header files. (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14659/head:pull/14659` \
`$ git checkout pull/14659`

`$ git checkout pull/14659` \
`$ git pull https://git.openjdk.org/jdk.git pull/14659/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14659`

`$ git pr show -t 14659`

</details>

<a href="https://git.openjdk.org/jdk/pull/14659.diff">https://git.openjdk.org/jdk/pull/14659.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14659#issuecomment-1607955175)

### DIFF
--- a/src/hotspot/cpu/aarch64/continuationFreezeThaw_aarch64.inline.hpp
+++ b/src/hotspot/cpu/aarch64/continuationFreezeThaw_aarch64.inline.hpp
@@ -217,7 +217,7 @@ template<typename FKind> frame ThawBase::new_stack_frame(const frame& hf, frame&
     intptr_t* heap_sp = hf.unextended_sp();
     // If caller is interpreted it already made room for the callee arguments
     int overlap = caller.is_interpreted_frame() ? ContinuationHelper::InterpretedFrame::stack_argsize(hf) : 0;
-    const int fsize = ContinuationHelper::InterpretedFrame::frame_bottom(hf) - hf.unextended_sp() - overlap;
+    const int fsize = checked_cast<int>(ContinuationHelper::InterpretedFrame::frame_bottom(hf) - hf.unextended_sp() - overlap);
     const int locals = hf.interpreter_frame_method()->max_locals();
     intptr_t* frame_sp = caller.unextended_sp() - fsize;
     intptr_t* fp = frame_sp + (hf.fp() - heap_sp);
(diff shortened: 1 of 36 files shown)


src/hotspot/cpu/aarch64/continuationFreezeThaw_aarch64.inline.hpp
```diff
@@ -217,7 +217,7 @@ template<typename FKind> frame ThawBase::new_stack_frame(const frame& hf, frame&
     intptr_t* heap_sp = hf.unextended_sp();
     // If caller is interpreted it already made room for the callee arguments
     int overlap = caller.is_interpreted_frame() ? ContinuationHelper::InterpretedFrame::stack_argsize(hf) : 0;
-    const int fsize = ContinuationHelper::InterpretedFrame::frame_bottom(hf) - hf.unextended_sp() - overlap;
+    const int fsize = checked_cast<int>(ContinuationHelper::InterpretedFrame::frame_bottom(hf) - hf.unextended_sp() - overlap);
```
fparain, line 220: This line computes the size of the unextended frame minus the overlap that exists if the callee is using arguments in the caller’s frame. So this value is lower or equal to the frame’ size which is encoded with an int. I don't think a checked_cast is required here. This applies to the x86 version too.

fparain (changes requested):

src/hotspot/cpu/aarch64/continuationFreezeThaw_aarch64.inline.hpp
coleenp: I'll make this change in the new PR.  I'm going to redo the address calculations to call a function rather than the casts.

coleenp:
After some offline conversation with Ioi, we thought of a better way to express these pointer subtractions returning int, so I'm going to close this PR and open a new one with at least these changed with that new function.

```// delta usually called with pointers or intptr_t that returns an int
// used for code buffer and other nearby address calculations.
// This allows the left side to be less than the right side.
template <typename T>
inline int delta_as_int(const volatile T left, const volatile T right) {
  return checked_cast<int>(left - right);
}       ```

PR #14669, opened 2023-06-27
8309209: C2 failed "assert(_stack_guard_state == stack_guard_reserved_disabled) failed: inconsistent state"

dholmes-ora:
This appears to be the same kind of issue as reported in [JDK-8146697](https://bugs.openjdk.org/browse/JDK-8146697) way back in Java 9, which was only "fixed" on x86.  The current failure was seen on Aarch64. It seems prudent to apply the same changes to all the other platforms. I've done Aarch64, and took a guess at RISC-V but do not know PPC or S390, so I am looking to others to provide the appropriate equivalent code changes there.

Testing so far is Aarch64 only:
- Tiers 1-3
- 50x the closed stackoverflow test that failed previously
- 25x vmTestbase/nsk/stress/stack/*

As these failures are so rare, passing tests don't really tell us much. This is more an attempt at additional robustness.

Thanks.

---------
### Progress

### Issue
 * [JDK-8309209](https://bugs.openjdk.org/browse/JDK-8309209): C2 failed "assert(_stack_guard_state == stack_guard_reserved_disabled) failed: inconsistent state" (**Bug** - P3)

### Reviewers
 * [Andrew Haley](https://openjdk.org/census#aph) (@theRealAph - **Reviewer**) ⚠️ Review applies to [1d001c3b](https://git.openjdk.org/jdk/pull/14669/files/1d001c3b91864263c9d1a8539077a232fc34535e)
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - **Reviewer**) ⚠️ Review applies to [1d001c3b](https://git.openjdk.org/jdk/pull/14669/files/1d001c3b91864263c9d1a8539077a232fc34535e)

### Contributors
 * Fei Yang `<fyang@openjdk.org>`
 * Martin Doerr `<mdoerr@openjdk.org>`
 * Amit Kumar `<amitkumar@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14669/head:pull/14669` \
`$ git checkout pull/14669`

`$ git checkout pull/14669` \
`$ git pull https://git.openjdk.org/jdk.git pull/14669/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14669`

`$ git pr show -t 14669`

</details>

<a href="https://git.openjdk.org/jdk/pull/14669.diff">https://git.openjdk.org/jdk/pull/14669.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14669#issuecomment-1608866044)

### DIFF
--- a/src/hotspot/cpu/aarch64/interp_masm_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/interp_masm_aarch64.cpp
@@ -694,8 +694,9 @@ void InterpreterMacroAssembler::remove_activation(
     Label no_reserved_zone_enabling;
 
     // check if already enabled - if so no re-enabling needed
-    ldr(rscratch1, Address(rthread, JavaThread::stack_guard_state_offset()));
-    cmp(rscratch1, (u1)StackOverflow::stack_guard_enabled);
+    assert(sizeof(StackOverflow::StackGuardState) == 4, "unexpected size");
+    ldrw(rscratch1, Address(rthread, JavaThread::stack_guard_state_offset()));
+    cmpw(rscratch1, (u1)StackOverflow::stack_guard_enabled);
     br(Assembler::EQ, no_reserved_zone_enabling);
 
     // look for an overflow into the stack reserved zone, i.e.
--- a/src/hotspot/cpu/aarch64/interp_masm_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/interp_masm_aarch64.cpp
@@ -693,6 +693,11 @@ void InterpreterMacroAssembler::remove_activation(
     // testing if reserved zone needs to be re-enabled
     Label no_reserved_zone_enabling;
 
+    // check if already enabled - if so no re-enabling needed
+    ldr(rscratch1, Address(rthread, JavaThread::stack_guard_state_offset()));
+    cmp(rscratch1, (u1)StackOverflow::stack_guard_enabled);
+    br(Assembler::EQ, no_reserved_zone_enabling);
+
     // look for an overflow into the stack reserved zone, i.e.
     // interpreter_frame_sender_sp <= JavaThread::reserved_stack_activation
     ldr(rscratch1, Address(rthread, JavaThread::reserved_stack_activation_offset()));
--- a/src/hotspot/cpu/riscv/interp_masm_riscv.cpp
+++ b/src/hotspot/cpu/riscv/interp_masm_riscv.cpp
@@ -765,7 +765,7 @@ void InterpreterMacroAssembler::remove_activation(
     Label no_reserved_zone_enabling;
 
     // check if already enabled - if so no re-enabling needed
-    ldw(t0, Address(xthread, JavaThread::stack_guard_state_offset()));
+    lw(t0, Address(xthread, JavaThread::stack_guard_state_offset()));
     subw(t0, t0, StackOverflow::stack_guard_enabled);
     beqz(t0, no_reserved_zone_enabling);
 
--- a/src/hotspot/cpu/riscv/interp_masm_riscv.cpp
+++ b/src/hotspot/cpu/riscv/interp_masm_riscv.cpp
@@ -765,8 +765,8 @@ void InterpreterMacroAssembler::remove_activation(
     Label no_reserved_zone_enabling;
 
     // check if already enabled - if so no re-enabling needed
-    ld(t0, Address(xthread, JavaThread::stack_guard_state_offset()));
-    sub(t0, t0, StackOverflow::stack_guard_enabled);
+    ldw(t0, Address(xthread, JavaThread::stack_guard_state_offset()));
+    subw(t0, t0, StackOverflow::stack_guard_enabled);
     beqz(t0, no_reserved_zone_enabling);
 
     ld(t0, Address(xthread, JavaThread::reserved_stack_activation_offset()));
--- a/src/hotspot/cpu/riscv/interp_masm_riscv.cpp
+++ b/src/hotspot/cpu/riscv/interp_masm_riscv.cpp
@@ -765,8 +765,9 @@ void InterpreterMacroAssembler::remove_activation(
     Label no_reserved_zone_enabling;
 
     // check if already enabled - if so no re-enabling needed
-    ld(t0, Address(rthread, JavaThread::stack_guard_state_offset()));
-    beq(t0, StackOverflow::stack_guard_enabled, no_reserved_zone_enabling);
+    ld(t0, Address(xthread, JavaThread::stack_guard_state_offset()));
+    sub(t0, t0, StackOverflow::stack_guard_enabled);
+    beqz(t0, no_reserved_zone_enabling);
 
     ld(t0, Address(xthread, JavaThread::reserved_stack_activation_offset()));
     ble(t1, t0, no_reserved_zone_enabling);
--- a/src/hotspot/cpu/riscv/interp_masm_riscv.cpp
+++ b/src/hotspot/cpu/riscv/interp_masm_riscv.cpp
@@ -765,6 +765,7 @@ void InterpreterMacroAssembler::remove_activation(
     Label no_reserved_zone_enabling;
 
     // check if already enabled - if so no re-enabling needed
+    assert(sizeof(StackOverflow::StackGuardState) == 4, "unexpected size");
     lw(t0, Address(xthread, JavaThread::stack_guard_state_offset()));
     subw(t0, t0, StackOverflow::stack_guard_enabled);
     beqz(t0, no_reserved_zone_enabling);
--- a/src/hotspot/cpu/riscv/interp_masm_riscv.cpp
+++ b/src/hotspot/cpu/riscv/interp_masm_riscv.cpp
@@ -764,6 +764,10 @@ void InterpreterMacroAssembler::remove_activation(
     // testing if reserved zone needs to be re-enabled
     Label no_reserved_zone_enabling;
 
+    // check if already enabled - if so no re-enabling needed
+    ld(t0, Address(rthread, JavaThread::stack_guard_state_offset()));
+    beq(t0, StackOverflow::stack_guard_enabled, no_reserved_zone_enabling);
+
     ld(t0, Address(xthread, JavaThread::reserved_stack_activation_offset()));
     ble(t1, t0, no_reserved_zone_enabling);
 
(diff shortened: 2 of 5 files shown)


theRealAph (approved):

dholmes-ora:
Thanks for the review @theRealAph !

src/hotspot/cpu/riscv/interp_masm_riscv.cpp
```diff
@@ -764,6 +764,10 @@ void InterpreterMacroAssembler::remove_activation(
     // testing if reserved zone needs to be re-enabled
     Label no_reserved_zone_enabling;
 
+    // check if already enabled - if so no re-enabling needed
+    ld(t0, Address(rthread, JavaThread::stack_guard_state_offset()));
+    beq(t0, StackOverflow::stack_guard_enabled, no_reserved_zone_enabling);
+
```
RealFYang, line 770: Hi, Thanks for handling RISC-V. But this won't build on RISC-V. We would need something like:

```
diff --git a/src/hotspot/cpu/riscv/interp_masm_riscv.cpp b/src/hotspot/cpu/riscv/interp_masm_riscv.cpp
index edec2e08c83..99906c0fea6 100644
--- a/src/hotspot/cpu/riscv/interp_masm_riscv.cpp
+++ b/src/hotspot/cpu/riscv/interp_masm_riscv.cpp
@@ -764,6 +764,11 @@ void InterpreterMacroAssembler::remove_activation(
     // testing if reserved zone needs to be re-enabled
     Label no_reserved_zone_enabling;

+    // check if already enabled - if so no re-enabling needed
+    ld(t0, Address(xthread, JavaThread::stack_guard_state_offset()));
+    sub(t0, t0, StackOverflow::stack_guard_enabled);
+    beqz(t0, no_reserved_zone_enabling);
+
     ld(t0, Address(xthread, JavaThread::reserved_stack_activation_offset()));
     ble(t1, t0, no_reserved_zone_enabling);
```

RealFYang (changes requested):

TheRealMDoerr:
PPC64 implementation:
```diff
--- a/src/hotspot/cpu/ppc/interp_masm_ppc_64.cpp
+++ b/src/hotspot/cpu/ppc/interp_masm_ppc_64.cpp
@@ -888,6 +888,11 @@ void InterpreterMacroAssembler::remove_activation(TosState state,
     // Test if reserved zone needs to be enabled.
     Label no_reserved_zone_enabling;
 
+    // check if already enabled - if so no re-enabling needed
+    lwz(R0, in_bytes(JavaThread::stack_guard_state_offset()), R16_thread);
+    cmpwi(CCR0, R0, StackOverflow::stack_guard_enabled);
+    beq_predict_taken(CCR0, no_reserved_zone_enabling);
+
     // Compare frame pointers. There is no good stack pointer, as with stack
     // frame compression we can get different SPs when we do calls. A subsequent
     // call could have a smaller SP, so that this compare succeeds for an
```

Is the RISCV version correct? `StackGuardState` is an `enum` and should typically have 4 Bytes.

dholmes-ora:
Thanks @TobiHartmann for the review. Note there is no arm32 version here as for some reason it does not have the reserved stack access support, at least in this area.

@RealFYang - thanks for that I will apply your change.

@TheRealMDoerr - thanks for PPC code. I will look into the issue with using the enum as for x64.

dholmes-ora:
> The other implementation don't look correct to me. StackGuardState is an enum and should typically have 4 Bytes.

@TheRealMDoerr could you elaborate please. We have the following:
- x86: we use `void cmpl(Address dst, int32_t imm32);`
- Aarch64: we use `cmp(Register Rd, unsigned char imm8)` and cast to `u1`
- RISC-V: we use `sub (Register Rd, Register Rn, int64_t decrement, Register temp = t0);` (perhaps should be `subw` for 32-bit? @RealFYang ?)

RealFYang:
@dholmes-ora : Yes, that makes sense to me. And I will also need to change to use 32-bit load (lw) to get the guard state.

```
diff --git a/src/hotspot/cpu/riscv/interp_masm_riscv.cpp b/src/hotspot/cpu/riscv/interp_masm_riscv.cpp
index edec2e08c83..1e981498fcc 100644
--- a/src/hotspot/cpu/riscv/interp_masm_riscv.cpp
+++ b/src/hotspot/cpu/riscv/interp_masm_riscv.cpp
@@ -764,6 +764,11 @@ void InterpreterMacroAssembler::remove_activation(
     // testing if reserved zone needs to be re-enabled
     Label no_reserved_zone_enabling;

+    // check if already enabled - if so no re-enabling needed
+    lw(t0, Address(xthread, JavaThread::stack_guard_state_offset()));
+    subw(t0, t0, StackOverflow::stack_guard_enabled);
+    beqz(t0, no_reserved_zone_enabling);
+
     ld(t0, Address(xthread, JavaThread::reserved_stack_activation_offset()));
     ble(t1, t0, no_reserved_zone_enabling);
```

dholmes-ora:
All modified platforms are building correctly - see GHA results.

src/hotspot/cpu/riscv/interp_masm_riscv.cpp
```diff
@@ -764,6 +764,11 @@ void InterpreterMacroAssembler::remove_activation(
     // testing if reserved zone needs to be re-enabled
     Label no_reserved_zone_enabling;
 
+    // check if already enabled - if so no re-enabling needed
+    ldw(t0, Address(xthread, JavaThread::stack_guard_state_offset()));
```
RealFYang, line 768: It's `lw` instead of `ldw`. See my previous comment. Thanks.

offamitkumar:
Here is s390x port:
```
diff --git a/src/hotspot/cpu/s390/interp_masm_s390.cpp b/src/hotspot/cpu/s390/interp_masm_s390.cpp
index 79c758f8b11..7555eb05b1f 100644
--- a/src/hotspot/cpu/s390/interp_masm_s390.cpp
+++ b/src/hotspot/cpu/s390/interp_masm_s390.cpp
@@ -952,6 +952,11 @@ void InterpreterMacroAssembler::remove_activation(TosState state,
     // Test if reserved zone needs to be enabled.
     Label no_reserved_zone_enabling;
 
+    // check if already enabled - if so no re-enabling needed
+    guarantee(sizeof(StackOverflow::StackGuardState) == 4, "unexptected size");
+    z_ly(Z_R0, Address(Z_thread, JavaThread::stack_guard_state_offset()));
+    compare32_and_branch(Z_R0, StackOverflow::stack_guard_enabled, bcondEqual, no_reserved_zone_enabling);
+
     // Compare frame pointers. There is no good stack pointer, as with stack
     // frame compression we can get different SPs when we do calls. A subsequent
     // call could have a smaller SP, so that this compare succeeds for an
```


@TheRealMDoerr I have a 2nd patch as well, Would you please confirm which will be better:

```
diff --git a/src/hotspot/cpu/s390/interp_masm_s390.cpp b/src/hotspot/cpu/s390/interp_masm_s390.cpp
index 79c758f8b11..a6774326286 100644
--- a/src/hotspot/cpu/s390/interp_masm_s390.cpp
+++ b/src/hotspot/cpu/s390/interp_masm_s390.cpp
@@ -952,6 +952,12 @@ void InterpreterMacroAssembler::remove_activation(TosState state,
     // Test if reserved zone needs to be enabled.
     Label no_reserved_zone_enabling;
 
+    // check if already enabled - if so no re-enabling needed
+    guarantee(sizeof(StackOverflow::StackGuardState) == 4, "unexptected size");
+    z_cli(Address(Z_thread, JavaThread::stack_guard_state_offset() + in_ByteSize(sizeof(StackOverflow::StackGuardState) - 1)),
+            StackOverflow::stack_guard_enabled);
+    z_bre(no_reserved_zone_enabling);
+
     // Compare frame pointers. There is no good stack pointer, as with stack
     // frame compression we can get different SPs when we do calls. A subsequent
     // call could have a smaller SP, so that this compare succeeds for an
```

src/hotspot/cpu/riscv/interp_masm_riscv.cpp
dholmes-ora: Fixed

dholmes-ora:
Thanks @offamitkumar ! I've applied the first variant for now.

/contributor add @offamitkumar

src/hotspot/cpu/aarch64/interp_masm_aarch64.cpp
```diff
@@ -693,6 +693,11 @@ void InterpreterMacroAssembler::remove_activation(
     // testing if reserved zone needs to be re-enabled
     Label no_reserved_zone_enabling;
 
+    // check if already enabled - if so no re-enabling needed
+    ldr(rscratch1, Address(rthread, JavaThread::stack_guard_state_offset()));
+    cmp(rscratch1, (u1)StackOverflow::stack_guard_enabled);
```
TheRealMDoerr, line 698: Not ldrw + cmpw?

dcubed-ojdk:
S390 typo: s/unexptected/unexpected/

src/hotspot/cpu/aarch64/interp_masm_aarch64.cpp
dholmes-ora: I've no idea, I used the code below as the pattern here. @theRealAph reviewed this.

dholmes-ora:
> S390 typo: s/unexptected/unexpected/

Thanks @dcubed-ojdk , I spotted that before comitting.

src/hotspot/cpu/aarch64/interp_masm_aarch64.cpp
dean-long: ldrw + cmpw does seem more correct, plus an asset that sizeof _stack_guard_state == 4.  ldr+cmp is only going to work for little-endian, and only as long as the alignment padding between _stack_guard_state and _stack_overflow_limit is all zeroes.

dholmes-ora: Changed

bulasevich:
> Note there is no arm32 version here as for some reason it does not have the reserved stack access support, at least in this area.

Yes, 'JEP 270: Reserved Stack Areas for Critical Sections' is not implemented for arm32, so the current change is not applicable to the platform.

dholmes-ora:
If there are no further comments I will integrate on Monday.

TheRealMDoerr:
Thanks for adding the assertions. I think they would be good to have for all platforms.

dholmes-ora:
> Thanks for adding the assertions. I think they would be good to have for all platforms. I have tested it on PPC64, too.

Thanks @TheRealMDoerr . I have made the comment and assertion the same on all platforms.

dholmes-ora:
Re-testing has passed okay.

/integrate

PR #14670, opened 2023-06-27
8310656: RISC-V: __builtin___clear_cache can fail silently.

robehn:
Hi, please consider.

We recently had a bug where user were missing permissions to use this syscall.
Which caused crashing on, according to hs_err on things like "addi x11, x24, 0" with SIGILL.
If it fails it is even possible to execute valid but 'old' instruction which may not lead to a crash, instead the program misbehaves.

To avoid this mess I suggest that we first test the syscall during vm init and we use it directly.
This way we can make sure it never fails.

Tested failing syscall with qemu, tested t1 in qemu, t1 on jh7110 in-progress.

---------
### Progress

### Issue
 * [JDK-8310656](https://bugs.openjdk.org/browse/JDK-8310656): RISC-V: __builtin___clear_cache can fail silently. (**Bug** - P2)

### Reviewers
 * [Ludovic Henry](https://openjdk.org/census#luhenry) (@luhenry - Committer) ⚠️ Review applies to [fe139d58](https://git.openjdk.org/jdk/pull/14670/files/fe139d58271e6396781f57b624e8b8607e6aa4ce)
 * [Thomas Stuefe](https://openjdk.org/census#stuefe) (@tstuefe - **Reviewer**) ⚠️ Review applies to [fe139d58](https://git.openjdk.org/jdk/pull/14670/files/fe139d58271e6396781f57b624e8b8607e6aa4ce)
 * [Fei Yang](https://openjdk.org/census#fyang) (@RealFYang - **Reviewer**) ⚠️ Review applies to [6b6e2dd9](https://git.openjdk.org/jdk/pull/14670/files/6b6e2dd92551f7e1a07eac5561a2807afcd57313)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14670/head:pull/14670` \
`$ git checkout pull/14670`

`$ git checkout pull/14670` \
`$ git pull https://git.openjdk.org/jdk.git pull/14670/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14670`

`$ git pr show -t 14670`

</details>

<a href="https://git.openjdk.org/jdk/pull/14670.diff">https://git.openjdk.org/jdk/pull/14670.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14670#issuecomment-1609032759)


luhenry:
For context, this was tracked down for https://github.com/adoptium/adoptium-support/issues/697

src/hotspot/os_cpu/linux_riscv/riscv_flush_icache.cpp
```diff
@@ -0,0 +1,76 @@
...
+#endif
+
+#define SYS_RISCV_FLUSH_ICACHE_LOCAL 1UL
+#define SYS_RISCV_FLUSH_ICACHE_ALL   0UL
+
+static long sys_flush_icache(uintptr_t start, uintptr_t end , uintptr_t flags) {
+  return syscall(NR_riscv_flush_icache, start, end, flags);
+}
+
+bool RiscvFlushIcache::test() {
```
luhenry, line 60: Why not having all this code in src/hotspot/cpu/riscv/icache_riscv.cpp directly? It seems it's going to be (and it should be) the only user of it.

robehn: During boot we want to give the user an informative message of what is wrong and not crash.
Calling RiscvFlushIcache::flush() brings the vm down hard, so we can't call that.
So the only options would be to expose the syscall and it's defines, which I don't want to.
Hence a separate method was done.

robehn: Note that reason why I bring down the VM is because if we are executing old instruction, anything can happen.
And it can go unnoticed for very long, by bring the VM down directly the user will not get unwanted consequences and have a chance to do a failover before anything bad happens.

robehn:
Thanks!

src/hotspot/os_cpu/linux_riscv/riscv_flush_icache.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+#define check_with_errno(check_type, cond, msg)                             \
+  do {                                                                      \
+    int err = errno;                                                        \
+    check_type(cond, "%s; error='%s' (errno=%s)", msg, os::strerror(err),   \
+               os::errno_name(err));                                        \
+} while (false)
+
+#define assert_with_errno(cond, msg)    check_with_errno(assert, cond, msg)
+#define guarantee_with_errno(cond, msg) check_with_errno(guarantee, cond, msg)
```
tstuefe, line 44: useful; potentially in debug.hpp?

tstuefe (approved):
Got curious. Looks good.

I was not even aware that __builtin___clear_cache had an errno. Is this documented somewhere? We should probably check arm64 too.

Question, would cacheflush(2) have worked too? Instead of the syscall?

RealFYang:
Hi, I don't quite understand how this issue triggers. Could it happen on real hardware platforms like Unmatched or others?
Since `__builtin___clear_cache` is a GCC built-in function which doesn't return an error code, my previous understanding is that it will never fails to do. And the GCC manual doesn't mention the cases where this could fail. Also I see uses of this built-in funtions on other platforms like AArch64.

```
void __builtin___clear_cache (void *begin, void *end) [Built-in Function]
This function is used to flush the processor’s instruction cache for the region of memory
between begin inclusive and end exclusive. Some targets require that the instruction
cache be flushed, after modifying memory containing code, in order to obtain
deterministic behavior.
If the target does not require instruction cache flushes, __builtin___clear_cache
has no effect. Otherwise either instructions are emitted in-line to clear the instruction
cache or a call to the __clear_cache function in libgcc is made.
```

luhenry:
@RealFYang the issue arises when you run Java _inside_ Docker. In that case, Docker will block certain syscalls (like riscv_flush_icache before Docker v23) unless you disable this security feature. `__builtin___clear_cache` will then silently fail (the syscall will fail with EPERM), and no icache flush will actually happen.

That doesn't happen on QEMU because the `riscv_flush_icache` syscall never actually gets called (it's "interpreted" by qemu inside the docker container), but it reproduces very reliably on HiFive Unmatched. To reproduce, simply run `java -version` on an Unmatched board inside a Docker container on stock Ubuntu or Debian (these still ship Docker v20).

robehn:
> Got curious. Looks good.
> 
> I was not even aware that __builtin___clear_cache had an errno. Is this documented somewhere? We should probably check arm64 too.
> 
> Question, would cacheflush(2) have worked too? Instead of the syscall?

Hi Thomas, thanks for looking, and thanks for being curios!

And sorry in advanced for this long answer, I felt that I needed to add some context.

Cross modifying code/CMODX is not specified on RV.
RV allows both coherent and non-coherent I/D cache.
Right now we have to assume non-coherent I/D cache (I have suggest hwprobe should return this).

But there is two pieces we can use, ifence and full IPI shootdown.
ifence is a local hart instruction read fence.
With non-coherent I/D cache ifence is not sufficient.
And ifence is an extension, not on all RV cpus.

Kernel 4.15 got the syscall riscv_flush_icache added with the IPI shootdown.
There is also a hart local version SYS_RISCV_FLUSH_ICACHE_LOCAL, which can be used instead of ifence.

If we ignore the case of having an UP (as you know we removed UP from the VM),
to make sure all threads/harts have a fresh I-cache our only option, at the moment,
is to emit that syscall => IPI.

There is still some cases in the VM I believe is not entirely correct.
The IPI is very expensive and system-wide intrusive, so if we are running on
I/D cache coherent system we should be able to utilize the ifence instruction and avoid (some?) IPI's.

Regarding the cacheflush wrapper it does not seem to implemented and if I tested it I get undef ref when linking. I only see the syscall in that header.

It looks like glibc/gcc don't consider failing syscalls, maybe this was uncommon, but now when everyone is running in different containers it's certainly more common. E.g. cacheflush do not list EPERM as a return value.

Note that there is a vdso wrapper for this syscall, which just do the ecall (syscall instruction), so I fail to see the point in trying to use that. And since we are doing an a full system IPI the speed of the syscall is not essential.

So I feel more comfortable with the VM explicit emitting this syscall, so everyone can see exactly what we are doing.

src/hotspot/os_cpu/linux_riscv/riscv_flush_icache.cpp
robehn: I'll create an jira ticker, I have spread this around to other places which also needs to be cleanup.

src/hotspot/cpu/riscv/icache_riscv.cpp
```diff
@@ -36,13 +39,20 @@ static int icache_flush(address addr, int lines, int magic) {
   //
   // No sush assurance is defined at the interface level of the builtin
   // method, and so we should make sure it works.
-  __asm__ volatile("fence rw, rw" : : : "memory");
```
RealFYang, line 39: I think we should keep this inline assembly code as explained by the preceding code comment.
If I remembered correctly, such a data fence is not specified at the syscall interface level.
```
__asm__ volatile("fence rw, rw" : : : "memory");
```
Also you might want to update code comment to reflect the new code.

RealFYang (changes requested):

src/hotspot/cpu/riscv/icache_riscv.cpp
robehn: Yes, ok, I'll add it back and fix comment.

src/hotspot/os_cpu/linux_riscv/riscv_flush_icache.cpp
robehn: https://bugs.openjdk.org/browse/JDK-8311145

robehn:
Pushed small update.

src/hotspot/cpu/riscv/icache_riscv.cpp
```diff
@@ -33,16 +36,22 @@ static int icache_flush(address addr, int lines, int magic) {
...
   __asm__ volatile("fence rw, rw" : : : "memory");
 
-  __builtin___clear_cache(addr, addr + (lines << ICache::log2_line_size));
+  RiscvFlushIcache::flush((uintptr_t)addr, ((uintptr_t)lines) << ICache::log2_line_size);
+
   return magic;
 }
 
 void ICacheStubGenerator::generate_icache_flush(ICache::flush_icache_stub_t* flush_icache_stub) {
+  // Only riscv_flush_icache is supported as I-cache synhronization.
```
RealFYang, line 49: Typo: s/synhronization/synchronization/

src/hotspot/os_cpu/linux_riscv/riscv_flush_icache.cpp
```diff
@@ -0,0 +1,76 @@
...
+#define SYS_RISCV_FLUSH_ICACHE_LOCAL 1UL
+#define SYS_RISCV_FLUSH_ICACHE_ALL   0UL
+
+static long sys_flush_icache(uintptr_t start, uintptr_t end , uintptr_t flags) {
+  return syscall(NR_riscv_flush_icache, start, end, flags);
+}
+
+bool RiscvFlushIcache::test() {
+  long ret;
+  intptr_t end = ((intptr_t)&ret) + 64;
```
RealFYang, line 62: It looks a little bit odd to me here as `ret` is only 8 bytes. Maybe define a local array of 64 bytes and operate on that?

RealFYang (approved):
Overall LGTM. Two nits remain.

tstuefe:
@robehn Thanks for the thorough explanation!
 
> So I feel more comfortable with the VM explicit emitting this syscall, so everyone can see exactly what we are doing.

I agree, rather do the real thing explicitly than use - and then have to second-guess - a wrapper.

src/hotspot/os_cpu/linux_riscv/riscv_flush_icache.cpp
robehn: We are not flushing ret, but the one/two cachelines around ret, since we know they are valid. (we just created the current frame(s) on to it)
Ret is only used to get an stack address in current frame, i.e.. it could have been char just as well.

But I can change since this was obviously not clear.

src/hotspot/cpu/riscv/icache_riscv.cpp
robehn: Fixed, thanks!

src/hotspot/os_cpu/linux_riscv/riscv_flush_icache.cpp
robehn: Fixed, thanks!

robehn:
Merge and updated for "[8311145](https://bugs.openjdk.org/browse/JDK-8311145)
Remove check_with_errno duplicates". and nits.

Thanks @RealFYang @tstuefe !

@RealFYang Let me know if that flush address update is better!

RealFYang:
@robehn : Yes, looks better to me now. Thank you for the update.

VladimirKempik:
We had an issue with __builtin___clear_cache and clang-compiled jvm linked with libstdc++ ( instead of libc++) where calling __builtin___clear_cache was resulting in the NOP. 
This effectively fixes that case as well.

VladimirKempik:
would you mind backporting it to 17u-dev as well ?

robehn:
> would you mind backporting it to 17u-dev as well ?

Sorry missed this.

@luhenry it will not apply clean due to "8311145: Remove check_with_errno duplicates"

luhenry:
Backport to jdk17u-dev in progress at https://github.com/openjdk/jdk17u-dev/pull/1968

PR #14672, opened 2023-06-27
8309902: C2: assert(false) failed: Bad graph detected in build_loop_late after JDK-8305189

rwestrel:
The crash happens after the following steps:

1- pre/main/post loops are created with assert predicates above the main loop.

2- the main loop is peeled

3- as a consequence, the `OpaqueZeroTripGuard` for the main loop is removed

4- That allows narrowing of the type of the CastII that was added right
  after the zero trip guard during pre/main/post loops creation
  
5- The CastII feeds into a range check CastII for the peeled iteration
  that becomes top because the narrowed type of the first CastII
  conflicts with the type recorded in the range check CastII.
  
6- The assert predicate that should fold to protect the range check
  CastII doesn't because of the fix for JDK-8282592: on assert
  predicate updates, the CastII at the zero trip guard is skipped. So
  the range check CastII sees the narrowing of the type of the CastII
  at the zero trip guard but the assert predicate doesn't.
  
The fix I propose is to revert that part of the change from
JDK-8282592 so both the range check CastII and the assert predicate
have the CastII at the zero trip guard as input and observe its type
updates. I went back to that bug and tried to reproduce the failure
again but couldn't. Reverting JDK-8281429 causes the bug to reproduce
again. I tried tweaking the test so the crash reproduces with
JDK-8281429 applied but couldn't.

This is caused by JDK-8305189 because step 3- happens because of
it. Before JDK-8305189, 3- happened after loop opts are over. I think
what happened then was that a template assertion predicate that was in
the process of having its `OpaqueLoopInit` and `OpaqueLoopStride`
removed constant folded so the crash wouldn't reproduce.

---------
### Progress

### Issue
 * [JDK-8309902](https://bugs.openjdk.org/browse/JDK-8309902): C2: assert(false) failed: Bad graph detected in build_loop_late after JDK-8305189 (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14672/head:pull/14672` \
`$ git checkout pull/14672`

`$ git checkout pull/14672` \
`$ git pull https://git.openjdk.org/jdk.git pull/14672/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14672`

`$ git pr show -t 14672`

</details>

<a href="https://git.openjdk.org/jdk/pull/14672.diff">https://git.openjdk.org/jdk/pull/14672.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14672#issuecomment-1609204650)

### DIFF
--- a/src/hotspot/share/opto/loopTransform.cpp
+++ b/src/hotspot/share/opto/loopTransform.cpp
@@ -2043,12 +2043,6 @@ bool IdealLoopTree::is_invariant(Node* n) const {
 // to the new stride.
 void PhaseIdealLoop::update_main_loop_assertion_predicates(Node* ctrl, CountedLoopNode* loop_head, Node* init,
                                                            const int stride_con) {
-  if (init->Opcode() == Op_CastII) {
-    // skip over the cast added by PhaseIdealLoop::cast_incr_before_loop() when pre/post/main loops are created because
-    // it can get in the way of type propagation
-    assert(((CastIINode*)init)->carry_dependency() && loop_head->skip_predicates() == init->in(0), "casted iv phi from pre loop expected");
-    init = init->in(1);
-  }
   Node* entry = ctrl;
   Node* prev_proj = ctrl;
   LoopNode* outer_loop_head = loop_head->skip_strip_mined();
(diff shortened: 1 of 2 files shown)


vnkozlov (approved):
Looks good.

chhagedorn (approved):
Looks reasonable. I will also remove any skipping of cast nodes in my patch for JDK-8288981.

PR #14674, opened 2023-06-27
8310948: Fix ignored-qualifiers warning in Hotspot

djelinski:
Please review this attempt to fix ignored-qualifiers warning.

Example warnings:
```
src/hotspot/share/oops/method.hpp:413:19: warning: 'volatile' type qualifier on return type has no effect [-Wignored-qualifiers]
   CompiledMethod* volatile code() const;
                   ^~~~~~~~~
```
```
src/hotspot/share/jfr/periodic/jfrModuleEvent.cpp:65:20: warning: type qualifiers ignored on cast result type [-Wignored-qualifiers]
    65 |   event.set_source((const ModuleEntry* const)from_module);
       |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The proposed fix removes the ignored qualifiers.
In a few AD files I replaced `const` with `constexpr` where I noticed that the method is returning a compile-time constant, and other platforms use `constexpr` on the same method.

Release, debug and slowdebug builds on Aarch64 / x64 and Mac / Linux complete without errors. Cross-compile GHA builds also pass.

---------
### Progress

### Issue
 * [JDK-8310948](https://bugs.openjdk.org/browse/JDK-8310948): Fix ignored-qualifiers warning in Hotspot (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14674/head:pull/14674` \
`$ git checkout pull/14674`

`$ git checkout pull/14674` \
`$ git pull https://git.openjdk.org/jdk.git pull/14674/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14674`

`$ git pr show -t 14674`

</details>

<a href="https://git.openjdk.org/jdk/pull/14674.diff">https://git.openjdk.org/jdk/pull/14674.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14674#issuecomment-1609675270)

### DIFF
--- a/src/hotspot/cpu/aarch64/aarch64.ad
+++ b/src/hotspot/cpu/aarch64/aarch64.ad
@@ -2285,7 +2285,7 @@ int HandlerImpl::emit_deopt_handler(CodeBuffer& cbuf)
 
 //=============================================================================
 
-const bool Matcher::match_rule_supported(int opcode) {
+bool Matcher::match_rule_supported(int opcode) {
   if (!has_match_rule(opcode))
     return false;
 
@@ -2320,7 +2320,7 @@ const TypeVectMask* Matcher::predicate_reg_type(const Type* elemTy, int length) 
 }
 
 // Vector calling convention not yet implemented.
-const bool Matcher::supports_vector_calling_convention(void) {
+bool Matcher::supports_vector_calling_convention(void) {
   return false;
 }
 
@@ -2340,7 +2340,7 @@ bool Matcher::is_short_branch_offset(int rule, int br_size, int offset) {
 }
 
 // Vector width in bytes.
-const int Matcher::vector_width_in_bytes(BasicType bt) {
+int Matcher::vector_width_in_bytes(BasicType bt) {
   // The MaxVectorSize should have been set by detecting SVE max vector register size.
   int size = MIN2((UseSVE > 0) ? 256 : 16, (int)MaxVectorSize);
   // Minimum 2 values in vector
@@ -2351,11 +2351,11 @@ const int Matcher::vector_width_in_bytes(BasicType bt) {
 }
 
 // Limits on vector size (number of elements) loaded into vector.
-const int Matcher::max_vector_size(const BasicType bt) {
+int Matcher::max_vector_size(const BasicType bt) {
   return vector_width_in_bytes(bt)/type2aelembytes(bt);
 }
 
-const int Matcher::min_vector_size(const BasicType bt) {
+int Matcher::min_vector_size(const BasicType bt) {
   int max_size = max_vector_size(bt);
   // Limit the min vector size to 8 bytes.
   int size = 8 / type2aelembytes(bt);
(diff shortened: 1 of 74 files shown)


dholmes-ora:
> warning: 'volatile' type qualifier on return type has no effect

Can't say that I understand that warning. If I pass in a volatile pointer and return the same pointer then I would not expect to lose the volatile aspect of it. @kbarrett can you explain this one to me?

kimbarrett (approved):
Looks good.  Though mind-numbingly large; breaking something like this up into easier
to digest chunks can make reviewing easier.

djelinski:
David: I think this part of the spec is relevant here:
> A non-class non-array prvalue cannot be [cv-qualified](https://en.cppreference.com/w/cpp/language/cv), [...]. (Note: a function call or cast expression may result in a prvalue of non-class cv-qualified type, but the cv-qualifier is generally immediately stripped out.)

[source](https://en.cppreference.com/w/cpp/language/value_category)
given that the cv qualifiers are immediately stripped by the compiler, there's no point in providing them.

In the particular volatile pointer case: the function performs a volatile read to get the pointer value (address). That address can then be used in a non-volatile manner.

Kim: I realize that it's a big change, so thank you very much for reviewing it anyway! I was prepared to split it up, just wanted to know if this warning is something we want to fix.

dean-long:
In the example, was the intention perhaps `volatile CompiledMethod*` instead of `CompiledMethod* volatile`?

djelinski:
I don't think that was the intention. There's a comment on `Method::_code` stating that the pointer can change at any point (so the pointer was supposed to be volatile), and `class CompiledMethod` does not have any `volatile` - qualified methods.

There are similar cases where `const` pointer to a class was returned involving `JavaThread`, `ReferenceProcessor`, `PSCardTable` and `ShenandoahHeapRegion`; I suppose we could review if `const` could be added to the returned class, but that's outside of the scope of this PR.

djelinski:
Can I get another +1 on this? or should I proceed with splitting?

src/hotspot/cpu/aarch64/aarch64.ad
```diff
@@ -2285,7 +2285,7 @@ int HandlerImpl::emit_deopt_handler(CodeBuffer& cbuf)
 
 //=============================================================================
 
-const bool Matcher::match_rule_supported(int opcode) {
```
dholmes-ora, line 2288: Have to wonder if these were all meant to be `bool Match:xxx() const {`?

dholmes-ora (approved):
I will approve this as-is but have to wonder whether many of these cases of const return types were intending to declare const functions?

dholmes-ora:
P.S. Forgot to say thanks for dealing with this!

src/hotspot/cpu/aarch64/aarch64.ad
djelinski: Yes, I think that may have been the original intent. I'll add const on these functions.

dholmes-ora: Ah okay :)

djelinski:
Thanks for the reviews!

/integrate

PR #14676, opened 2023-06-27
8310949: RISC-V: Initialize UseUnalignedAccesses

luhenry:
---------
### Progress

### Issue
 * [JDK-8310949](https://bugs.openjdk.org/browse/JDK-8310949): RISC-V: Initialize UseUnalignedAccesses (**Enhancement** - P4)

### Reviewers
 * [Robbin Ehn](https://openjdk.org/census#rehn) (@robehn - **Reviewer**) ⚠️ Review applies to [ca1d1240](https://git.openjdk.org/jdk/pull/14676/files/ca1d12404865d79aaa1be7e2146b08fb1808caa4)
 * [Vladimir Kempik](https://openjdk.org/census#vkempik) (@VladimirKempik - Committer)
 * [Fei Yang](https://openjdk.org/census#fyang) (@RealFYang - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14676/head:pull/14676` \
`$ git checkout pull/14676`

`$ git checkout pull/14676` \
`$ git pull https://git.openjdk.org/jdk.git pull/14676/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14676`

`$ git pr show -t 14676`

</details>

<a href="https://git.openjdk.org/jdk/pull/14676.diff">https://git.openjdk.org/jdk/pull/14676.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14676#issuecomment-1609444448)


robehn (approved):
Looks good to me!

RealFYang:
We will need some performance numbers to validate this change. We witnessed some obvious regressions of Unsafe.Put*Unaligned methods when enabling UseUnalignedAccesses on some platforms like HiFive Unmatched with slow unaligned accesses. My guess is that we might want enable UseUnalignedAccesses for platforms like T-Head with fast unaligned accesses (RISCV_HWPROBE_MISALIGNED_FAST). But that will still depend on the measured JMH numbers.

shipilev:
The unaligned accesses paths were originally introduced to support faster word-sized accesses, where possible: https://bugs.openjdk.org/browse/JDK-8026049. So `UseUnalignedAccesses` means "unaligned accesses are supported and faster than groups of byte-sized accesses". If this is not true on some RISC-V implementations, we should not enable `+UseUnalignedAccesses` there, I think.

It would not be unusual to default to `+UseUnalignedAccesses` by default, and then make a few exceptions for known-slow platforms. Assuming we can sense them, of course. See how x86 code sometimes excepts configurations based on CPU family or microarchitecture.

shipilev:
Hey, there is a block before this in the same file:

```
 if (FLAG_IS_DEFAULT(AvoidUnalignedAccesses)) {
    if (unaligned_access.value() != MISALIGNED_FAST) {
      FLAG_SET_DEFAULT(AvoidUnalignedAccesses, true);
    } else {
      FLAG_SET_DEFAULT(AvoidUnalignedAccesses, false);
    }
}
```

...which needs to be reconciled with this one? At very least, it shows that it is possible to sense `MISALIGNED_FAST`.

luhenry:
@shipilev sounds good! Let's only enable `UseUnalignedAccesses` on platform that support fast unaligned accesses. We should use `unaligned_access.value() == MISALIGNED_FAST` to set it up IMO.

shipilev:
Yes, that would be my suggestion. But @RealFYang's point still stands: we would still need to confirm that actually works well on Unmatched and other known RISC-V boards.

src/hotspot/cpu/riscv/vm_version_riscv.cpp
```diff
@@ -224,6 +231,7 @@ void VM_Version::initialize() {
     FLAG_SET_DEFAULT(UseBlockZeroing, false);
   }
 
+
```
VladimirKempik, line 234: Please remove this extra line, otherwise looks good

shipilev:
> > Yes, that would be my suggestion. But @RealFYang's point still stands: we would still need to confirm that actually works well on Unmatched and other known RISC-V boards.
> 
> unmatched has misaligned emulator and slow thead c910 core has fast misaligned access but it's stuck on linux 5.10 from vendor which don't have new hw_prober syscall

Is there a board/image that has the hw_prober implemented, and it returns `MISALIGNED_FAST` there?

VladimirKempik:
> > > Yes, that would be my suggestion. But @RealFYang's point still stands: we would still need to confirm that actually works well on Unmatched and other known RISC-V boards.
> > 
> > 
> > unmatched has misaligned emulator and slow thead c910 core has fast misaligned access but it's stuck on linux 5.10 from vendor which don't have new hw_prober syscall
> 
> Is there a board/image that has the hw_prober implemented, and it returns `MISALIGNED_FAST` there?

Doubt so, my fpga board with custom scr cores ( both with misaligned_emu and misglined_fast) might get kernel 6.4 ( with hw_prober) later this summer

VladimirKempik:
One case to consider:
lets say I have a system with X big cores ( which support MISALIGNED_FAST) and X small cores ( with MISALIGNED_EMU)

if I run some java workload on all cores, what should hw_prober return ?  obvious result here is to use +AvoidUnallignedAccesses.

If I run same java workload but use taskset to run it only on big cores, how will jdk's hw_prober code work ? should it work properly and disable AvoidUnallignedAccesses or it's too much and one need to manually set -XX:-AvoidUnallignedAccesses ?

robehn:
> One case to consider: lets say I have a system with X big cores ( which support MISALIGNED_FAST) and X small cores ( with MISALIGNED_EMU)
> 
> if I run some java workload on all cores, what should hw_prober return ? obvious result here is to use +AvoidUnallignedAccesses.
> 
> If I run same java workload but use taskset to run it only on big cores, how will jdk's hw_prober code work ? should it work properly and disable AvoidUnallignedAccesses or it's too much and one need to manually set -XX:-AvoidUnallignedAccesses ?

FYI:
The only way today using hwprobe on in example above is to query each cpu individually to find the cpu set that have fast and the cpu set have emulated. (or vector or any other extension which may differ)
I have proposed that hwprobe should be able to also return a cpu set for some set of features.
To either inform user of what affinity they should use, or if we want change the affinity of the VM automagically.

palmer-dabbelt:
> > One case to consider: lets say I have a system with X big cores ( which support MISALIGNED_FAST) and X small cores ( with MISALIGNED_EMU)
> > if I run some java workload on all cores, what should hw_prober return ? obvious result here is to use +AvoidUnallignedAccesses.
> > If I run same java workload but use taskset to run it only on big cores, how will jdk's hw_prober code work ? should it work properly and disable AvoidUnallignedAccesses or it's too much and one need to manually set -XX:-AvoidUnallignedAccesses ?
> 
> FYI: The only way today using hwprobe in example above is to query each cpu individually to find the cpu set that have fast and the cpu set have emulated. (or vector or any other extension which may differ) I have proposed that hwprobe should be able to also return a cpu set for some set of features. To either inform user of what affinity they should use, or if we want change the affinity of the VM automagically.

I think Robbin brought this up at some meeting type thing, but IMO that's a pretty reasonable ask.  We hadn't thought of it when writing the syscall, but we've got some flags for extensibility so I think we could make it work.

Another option might be to tie this to some hueristics in userspace, maybe probing along the CPU topology or something.  There's been some vague discussions about having a hwprobe userspace library to handle things like bit->string mappings, maybe we should just have it do this too?

VladimirKempik:
Hello @palmer-dabbelt, I meant I don't know how this thing should work "properly" when we have cpus with different capabilities (regarding misaligned access) and affinity manually set to misaligned_fast cores.

palmer-dabbelt:
> Hello @palmer-dabbelt, I meant I don't know how this thing should work "properly" when we have cpus with different capabilities (regarding misaligned access) and affinity manually set to misaligned_fast cores.

Sorry, I'm kind of lost here.  There's some kernel docs for how `riscv_hwprobe()` works, it's got a CPU set argument to control which cores are being probed.  That argument is meant to line up with how the other scheduling controls work, so it's semi-transparent when userspace doesn't care that much.

I guess I'd need to go read the docs, but we might have a grey area for the misaligned access speed: if we say it's slow, is that slow on one core or slow on all cores.  For the other features it's for all cores, but not sure if we were explicit enough in the docs for the performance stuff.  Probably worth a read of the docs to see if we can improve them, this is all pretty new so I wouldn't be surprised if that's the case.

luhenry:
@VladimirKempik the `riscv_hwprobe` is going to return the intersection of the features supported by all cores. In the case of a big.LITTLE set of cores, if the LITTLE cores don't support fast unaligned accesses, then `riscv_hwprobe` will return `SLOW` or `UNSUPPORTED` and `UseUnalignedAccesses` will then be false.

That keeps us in the safer case where we will not generate unaligned accesses if we are not guaranteed that all cores support fast unaligned accesses.

VladimirKempik:
> @VladimirKempik the `riscv_hwprobe` is going to return the intersection of the features supported by all cores. In the case of a big.LITTLE set of cores, if the LITTLE cores don't support fast unaligned accesses, then `riscv_hwprobe` will return `SLOW` or `UNSUPPORTED` and `UseUnalignedAccesses` will then be false.
> 
> That keeps us in the safer case where we will not generate unaligned accesses if we are not guaranteed that all cores support fast unaligned accesses.
> 
> 

Hello Ludovic, that sounds good, thanks.

luhenry:
@robehn @VladimirKempik @RealFYang could I please get another review? Thanks!

VladimirKempik (approved):

RealFYang:
Hello, could you please provide some JMH numbers on platforms like T-Head? (-XX:+UseUnalignedAccesses vs -XX:-UseUnalignedAccesses)? I don't see such numbers yet. We only tested on Unmatched before: https://cr.openjdk.org/~dzhang/TestUseUnalignedAccesses/

VladimirKempik:
Can we deduct UseUnalignedAccesses based on the value of AvoidUnalignedAccess or vice-versa?, We currently have two flags which is confusing

Ilyagavrilin:
Results on T-Head board
-XX:-UseUnalignedAccesses :
```
Benchmark                                       (size)   Mode  Cnt     Score   Error   Units
TestUseUnalignedAccesses.testPutIntUnaligned       100  thrpt   15  1907.338 ± 6.943  ops/ms
TestUseUnalignedAccesses.testPutLongUnaligned      100  thrpt   15  1200.164 ± 2.577  ops/ms
TestUseUnalignedAccesses.testPutShortUnaligned     100  thrpt   15  2005.762 ± 8.693  ops/ms

```
XX:+UseUnalignedAccesses:
```
Benchmark                                       (size)   Mode  Cnt     Score    Error   Units
TestUseUnalignedAccesses.testPutIntUnaligned       100  thrpt   15  2997.842 ± 11.228  ops/ms
TestUseUnalignedAccesses.testPutLongUnaligned      100  thrpt   15  3058.505 ± 11.970  ops/ms
TestUseUnalignedAccesses.testPutShortUnaligned     100  thrpt   15  3157.681 ± 39.818  ops/ms

```

RealFYang (approved):

luhenry:
@RealFYang I expect the T-Head to return that it supports fast unaligned memory accesses, as that's clearly the case.

@VladimirKempik agreed that 1 flag could be used instead of 2. However, these 2 flags are there on all other platforms already. I would then do a follow-up change to merge them across all platforms and not just RISC-V.

VladimirKempik:
> @RealFYang I expect the T-Head to return that it supports fast unaligned memory accesses, as that's clearly the case.

Don't expect thead to be equipped with 6.4+ kernel any time soon

luhenry:
> Don't expect thead to be equipped with 6.4+ kernel any time soon

Absolutely, and until that's the case, the behaviour stays the same as it currently is, so no regressions.

PR #14677, opened 2023-06-27
8310939: [c1] The visibility of write-volatile requires membar_volatile instead of membar

sunny868:
For c1 now, a volatile write case:

membar_release // LoadStore | StoreStore
write volatile
membar

Just like c2, here `membar` should be defined `membar_volatile` clearly, then for risc-v, ppc and loongarch can use StoreLoad for `membar_volatile` for better performance.

Testing:
GHA testing
jtreg tier1-3 for loongarch64

---------
### Progress

### Issue
 * [JDK-8310939](https://bugs.openjdk.org/browse/JDK-8310939): [c1] The visibility of write-volatile requires membar_volatile instead of membar (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14677/head:pull/14677` \
`$ git checkout pull/14677`

`$ git checkout pull/14677` \
`$ git pull https://git.openjdk.org/jdk.git pull/14677/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14677`

`$ git pr show -t 14677`

</details>

<a href="https://git.openjdk.org/jdk/pull/14677.diff">https://git.openjdk.org/jdk/pull/14677.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14677#issuecomment-1609444638)

### DIFF
--- a/src/hotspot/cpu/aarch64/c1_LIRAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/c1_LIRAssembler_aarch64.cpp
@@ -2976,6 +2976,10 @@ void LIR_Assembler::membar_release() {
   __ membar(Assembler::LoadStore|Assembler::StoreStore);
 }
 
+void LIR_Assembler::membar_volatile() {
+  __ membar(Assembler::StoreLoad);
+}
+
 void LIR_Assembler::membar_loadload() {
   __ membar(Assembler::LoadLoad);
 }
--- a/src/hotspot/cpu/arm/c1_LIRAssembler_arm.cpp
+++ b/src/hotspot/cpu/arm/c1_LIRAssembler_arm.cpp
@@ -2697,6 +2697,10 @@ void LIR_Assembler::membar_release() {
   __ membar(MacroAssembler::Membar_mask_bits(MacroAssembler::StoreStore | MacroAssembler::LoadStore), Rtemp);
 }
 
+void LIR_Assembler::membar_volatile() {
+  __ membar(MacroAssembler::StoreLoad, Rtemp);
+}
+
 void LIR_Assembler::membar_loadload() {
   __ membar(MacroAssembler::LoadLoad, Rtemp);
 }
--- a/src/hotspot/cpu/ppc/c1_LIRAssembler_ppc.cpp
+++ b/src/hotspot/cpu/ppc/c1_LIRAssembler_ppc.cpp
@@ -2910,6 +2910,10 @@ void LIR_Assembler::membar_release() {
   __ release();
 }
 
+void LIR_Assembler::membar_volatile() {
+  __ membar(Assembler::StoreLoad);
+}
+
 void LIR_Assembler::membar_loadload() {
   __ membar(Assembler::LoadLoad);
 }
--- a/src/hotspot/cpu/riscv/c1_LIRAssembler_riscv.cpp
+++ b/src/hotspot/cpu/riscv/c1_LIRAssembler_riscv.cpp
@@ -1918,6 +1918,10 @@ void LIR_Assembler::membar_release() {
   __ membar(MacroAssembler::LoadStore | MacroAssembler::StoreStore);
 }
 
+void LIR_Assembler::membar_volatile() {
+  __ membar(MacroAssembler::StoreLoad);
+}
+
 void LIR_Assembler::membar_loadload() {
   __ membar(MacroAssembler::LoadLoad);
 }
--- a/src/hotspot/cpu/s390/c1_LIRAssembler_s390.cpp
+++ b/src/hotspot/cpu/s390/c1_LIRAssembler_s390.cpp
@@ -2886,6 +2886,10 @@ void LIR_Assembler::membar_release() {
   __ z_release();
 }
 
+void LIR_Assembler::membar_volatile() {
+  __ z_fence();
+}
+
 void LIR_Assembler::membar_loadload() {
   __ z_acquire();
 }
--- a/src/hotspot/cpu/x86/c1_LIRAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/c1_LIRAssembler_x86.cpp
@@ -4000,6 +4000,10 @@ void LIR_Assembler::membar_release() {
   // No x86 machines currently require store fences
 }
 
+void LIR_Assembler::membar_volatile() {
+  __ membar( Assembler::Membar_mask_bits(Assembler::StoreLoad));
+}
+
 void LIR_Assembler::membar_loadload() {
   // no-op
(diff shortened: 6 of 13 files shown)


shipilev:
We need to be really careful about this, because we had the issues with inconsistent barrier schemes in C1 and C2:
https://bugs.openjdk.org/browse/JDK-8179954

Modern jcstress is able to run tests with one side compiled by C1 and another one compiled by C2, exactly to catch these things. So I suggest you get at least one clean jcstress run with this patch, at least on AArch64 (with both `-UseLSE` and `+UseLSE`).

sunny868:
@shipilev thanks for the reminder.   After the patch for AArch64, we also can see

  STLR #1 -> [x] | STLR #1 -> [y]

  DMB | DMB
  LDR [y] -> r1 | LDR [x] -> r1
  DMB LD | DMB LD
  
 On AArch64, Assembler::AnyAny and Assembler::StoreLoad also be defined ISH, so not found regression. And this patch can not cause regression on X86 and S390, because also used same instruction in membar() and membar_volatile().

I had run jcstress `org.openjdk.jcstress.samples.jmm.basic.BasicJMM_07_Consensus` on AArch64 and LoongArch64, and result is

`(Results: 102 planned; 102 passed, 0 failed, 0 soft errs, 0 hard errs)`

But I am not sure for risc-v, ppc. Maybe please @RealFYang @TheRealMDoerr help to confirm this patch.

TheRealMDoerr:
This doesn't change anything for PPC64 for the following reasons:
1. `support_IRIW_for_not_multiple_copy_atomic_cpu` is true
2. StoreLoad = fence = sync = membar_volatile
I wonder if a performance improvement can be measured on any other platform. I guess StoreLoad is typically so slow that other orderings don't matter?

sunny868:
OK, thanks @TheRealIMDoerr for the confirm. This PR has performance advantages on LoongArch, because there are special StoreLoad instructions and full barrier instructions on LA, and the performance is different.

dbar 0x10 // Wait for earlier read-write (RW) operations complete before subsequent read-write (RW) operations
dbar 0x19 // Wait for earlier write (W) operations to complete before subsequent read (R) operations

Maybe I should write a performance test case for this PR.

xiangzhai:
PING

theRealAph:
`membar_volatile()` is a really bad name for this operation. It isn't any such thing: on some processors a volatile access requires more than just a storeLoad, and on some processors a storeLoad isn't needed for a volatile store. If you want a storeLoad you should call it storeLoad, please. Calling it `volatile` is just wrong.

theRealAph:
> I wonder if a performance improvement can be measured on any other platform. I guess StoreLoad is typically so slow that other orderings don't matter?

I know of no other processor on which StoreLoad is not also a full barrier.

theRealAph:
> Maybe I should write a performance test case for this PR.

Seriously? You never measured the speed problem before writing a patch for it? How did you know it was a problem?

dean-long:
It's confusing to me to rename the post-barrier but not the pre-barrier.  Instead of changing membar_release + membar to 
membar_release + membar_volatile, something like volatile_write_pre_barrier + volatile_write_post_barrier in the shared code makes more sense to me.

sunny868:
> > Maybe I should write a performance test case for this PR.
> 
> Seriously? You never measured the speed problem before writing a patch for it? How did you know it was a problem?

I've tested specjbb2015 and some related JMH tests, really haven't found a performance improvement.
But for the LoongArch architecture, StoreLoad is not equal to full barrier. And compared to C2's volatile write `MembarRelease-StoreNode-MembarVolatile`implementation, it is argued that c1 here membar() should be membar_storeload().

I previously thought that PPC, RISC-V's StoreLoad was not equivalent to full barrier, so I submitted the PR. Now it seems that only LoongArch has this problem, so should I make the following changes or abandon this PR?

--- a/src/hotspot/share/gc/shared/c1/barrierSetC1.cpp
+++ b/src/hotspot/share/gc/shared/c1/barrierSetC1.cpp
@@ -161,7 +161,7 @@ void BarrierSetC1::store_at_resolved(LIRAccess& access, LIR_Opr value) {
   }
 
   if (is_volatile && !support_IRIW_for_not_multiple_copy_atomic_cpu) {
-    __ membar();
+    __ membar_storeload();
   }
 }

TheRealMDoerr:
I'd abandon it unless it is really beneficial for LoongArch which I can't tell. (Probably not.)

sunny868:
> I'd abandon it unless it is really beneficial for LoongArch which I can't tell. (Probably not.)

ok

theRealAph:
> _Mailing list message from [David Holmes](mailto:david.holmes@oracle.com) on [hotspot-dev](mailto:hotspot-dev@mail.openjdk.org):_
> > 
> > I know that this patch is now withdrawn, but I'd like to make a point.
> > Please don't use terms like volatile_write_post_barrier for StoreLoad
> > if you can possibly avoid it, because that's only one way to implement Java's volatile.
> > The assumption that volatile write is implemented by release; write; StoreLoad
> > or similar permeates HotSpot, and is a pain for architectures like AArch64
> > which that don't do volatile that way.
> 
> My long term understanding of the issue is that the requirements for volatile accesses is determined by the nature of the surrounding accesses - as per Doug Lea's JMM Cookbook [1].

That is to say, if you have to implement volatile using plain memory accesses and fences, then you are going to need a StoreLoad between a plain store and a plain load in the same thread. But the JMM Cookbook is of historical interest only, because there are now alternative ways to do it. 

> But neither the interpreter nor C1 consider this and only apply barriers to the actual volatile read/write in isolation - hence they (have to) assume the worst-case and use the strongest pre/post barriers that may be needed. And of course the kinds of barriers available depend on architecture.

They don't have to do that: they could simply have punted volatile-store/volatile-load to the back end.

theRealAph:
For a little more clarity: let's imagine that you wanted to have a sequentially-consistent set of stores and loads, using plain memory accesses and fences.

This would work:

str r1, [x1] ; fullFence; ldr r1, [x1]; fullFence; str r1, [x2]; fullFence; ldr r1, [x2]

but this would not:

str r1, [x1] ; storeLoad; ldr r1, [x1]; storeLoad; str r1, [x2]; storeLoad; ldr r1, [x2]

This is why I wouldn't call storeLoad "membar_volatile".

dean-long:
Just to clarify, I was suggesting that the shared code should not try to impose which barriers, if any, are used by the cpu-specific implementation.  So a better name for `volatile_write_post_barrier` might have been `volatile_write_post_hook`, where the "volatile_write" part is because that's the logical operation the shared code is doing.  The implementation of the hook could be a barrier or even a no-op.

PR #14678, opened 2023-06-27
8303279: C2: crash in SubTypeCheckNode::sub() at IGVN split if

rwestrel:
The crash occurs because at split if during IGVN, a `SubTypeCheck` is
created with null as input. That happens because the control path the
`SubTypeCheck` is cloned for is dead. To fix that I propose delaying
split if until dead paths are collapsed.

I added an assert to check a nullable first input to `SubTypeCheck`
nodes (which should be impossible because it should be null
checked). When I ran testing, a number of cases showed up with known
non null values non properly marked as non null. I fixed them.

---------
### Progress

### Issue
 * [JDK-8303279](https://bugs.openjdk.org/browse/JDK-8303279): C2: crash in SubTypeCheckNode::sub() at IGVN split if (**Bug** - P3)

### Reviewers
 * [Vladimir Kozlov](https://openjdk.org/census#kvn) (@vnkozlov - **Reviewer**) ⚠️ Review applies to [cb3a04f0](https://git.openjdk.org/jdk/pull/14678/files/cb3a04f07ba6d2de8bdd22491c1fc3b7d8581d54)
 * [Vladimir Ivanov](https://openjdk.org/census#vlivanov) (@iwanowww - **Reviewer**) ⚠️ Review applies to [017d60b1](https://git.openjdk.org/jdk/pull/14678/files/017d60b13790b3eb9861c32b2a9b7cb00c8cc877)
 * [Volker Simonis](https://openjdk.org/census#simonis) (@simonis - **Reviewer**) ⚠️ Review applies to [017d60b1](https://git.openjdk.org/jdk/pull/14678/files/017d60b13790b3eb9861c32b2a9b7cb00c8cc877)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14678/head:pull/14678` \
`$ git checkout pull/14678`

`$ git checkout pull/14678` \
`$ git pull https://git.openjdk.org/jdk.git pull/14678/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14678`

`$ git pr show -t 14678`

</details>

<a href="https://git.openjdk.org/jdk/pull/14678.diff">https://git.openjdk.org/jdk/pull/14678.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14678#issuecomment-1609660162)

### DIFF
--- a/src/hotspot/share/opto/ifnode.cpp
+++ b/src/hotspot/share/opto/ifnode.cpp
@@ -92,7 +92,7 @@ static Node* split_if(IfNode *iff, PhaseIterGVN *igvn) {
   Node *con1=nullptr;
   uint i4;
   RegionNode* phi_region = phi->region();
-  for(i4 = 1; i4 < phi->req(); i4++ ) {
+  for (i4 = 1; i4 < phi->req(); i4++ ) {
     con1 = phi->in(i4);
     // Do not optimize partially collapsed merges
     if (con1 == nullptr || phi_region->in(i4) == nullptr || igvn->type(phi_region->in(i4)) == Type::TOP) {
--- a/src/hotspot/share/opto/ifnode.cpp
+++ b/src/hotspot/share/opto/ifnode.cpp
@@ -91,9 +91,14 @@ static Node* split_if(IfNode *iff, PhaseIterGVN *igvn) {
   // See that the merge point contains some constants
   Node *con1=nullptr;
   uint i4;
-  for( i4 = 1; i4 < phi->req(); i4++ ) {
+  RegionNode* phi_region = phi->region();
+  for(i4 = 1; i4 < phi->req(); i4++ ) {
     con1 = phi->in(i4);
-    if( !con1 ) return nullptr;    // Do not optimize partially collapsed merges
+    // Do not optimize partially collapsed merges
+    if (con1 == nullptr || phi_region->in(i4) == nullptr || igvn->type(phi_region->in(i4)) == Type::TOP) {
+      igvn->_worklist.push(iff);
+      return nullptr;
+    }
     if( con1->is_Con() ) break; // Found a constant
     // Also allow null-vs-not-null checks
     const TypePtr *tp = igvn->type(con1)->isa_ptr();
@@ -115,7 +120,7 @@ static Node* split_if(IfNode *iff, PhaseIterGVN *igvn) {
 
   // No intervening control, like a simple Call
   Node* r = iff->in(0);
-  if (!r->is_Region() || r->is_Loop() || phi->region() != r || r->as_Region()->is_copy()) {
+  if (!r->is_Region() || r->is_Loop() || phi_region != r || r->as_Region()->is_copy()) {
     return nullptr;
   }
 
--- a/src/hotspot/share/opto/library_call.hpp
+++ b/src/hotspot/share/opto/library_call.hpp
@@ -180,9 +180,6 @@ class LibraryCallKit : public GraphKit {
   CallJavaNode* generate_method_call_static(vmIntrinsicID method_id, bool res_not_null) {
     return generate_method_call(method_id, false, true, res_not_null);
   }
-  CallJavaNode* generate_method_call_virtual(vmIntrinsics::ID method_id) {
-    return generate_method_call(method_id, true, false, false);
-  }
   Node* load_field_from_object(Node* fromObj, const char* fieldName, const char* fieldTypeString, DecoratorSet decorators = IN_HEAP, bool is_static = false, ciInstanceKlass* fromKls = nullptr);
   Node* field_address_from_object(Node* fromObj, const char* fieldName, const char* fieldTypeString, bool is_exact = true, bool is_static = false, ciInstanceKlass* fromKls = nullptr);
 
--- a/src/hotspot/share/opto/library_call.hpp
+++ b/src/hotspot/share/opto/library_call.hpp
@@ -176,13 +176,12 @@ class LibraryCallKit : public GraphKit {
   Node* generate_array_guard_common(Node* kls, RegionNode* region,
                                     bool obj_array, bool not_array);
   Node* generate_virtual_guard(Node* obj_klass, RegionNode* slow_region);
-  CallJavaNode* generate_method_call(vmIntrinsics::ID method_id,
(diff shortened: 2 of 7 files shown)


vnkozlov (approved):
Looks reasonable.

src/hotspot/share/opto/ifnode.cpp
```diff
@@ -91,9 +91,14 @@ static Node* split_if(IfNode *iff, PhaseIterGVN *igvn) {
   // See that the merge point contains some constants
   Node *con1=nullptr;
   uint i4;
-  for( i4 = 1; i4 < phi->req(); i4++ ) {
+  RegionNode* phi_region = phi->region();
+  for(i4 = 1; i4 < phi->req(); i4++ ) {
```
iwanowww, line 95: Missing space: `for (i4`.

test/hotspot/jtreg/compiler/splitif/TestCrashAtIGVNSplitIfSubType.java
```diff
@@ -0,0 +1,73 @@
...
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/*
+ * @test
+ * @bug 8303279
+ * @summary C2 Compiler crash (triggered by Kotlin 1.8.10)
+ * @run main/othervm -XX:-TieredCompilation -XX:-BackgroundCompilation -XX:-UseOnStackReplacement -XX:+PrintCompilation -XX:CompileOnly=TestCrashAtIGVNSplitIfSubType::test -XX:CompileCommand=quiet -XX:+StressIGVN -XX:StressSeed=598200189 TestCrashAtIGVNSplitIfSubType
```
iwanowww, line 28: Missing flag: `-XX:+StressIGVN` requires `-XX:+UnlockDiagnosticVMOptions`.

src/hotspot/share/opto/subtypenode.cpp
```diff
@@ -34,6 +34,7 @@
 
 const Type* SubTypeCheckNode::sub(const Type* sub_t, const Type* super_t) const {
   const TypeKlassPtr* superk = super_t->isa_klassptr();
+  assert(sub_t != Type::TOP && !TypePtr::NULL_PTR->higher_equal(sub_t), "should be not null");
```
iwanowww, line 37: There's a failure observed during testing. I attached logs to the bug.

iwanowww (commented):
Proposed fix looks good.

The testing revealed a failure in the newly introduced assertion (attached logs to the bug).

Also, the bug summary is way too generic and lacks any details about the actual problem. Please, update it.

src/hotspot/share/opto/library_call.hpp
```diff
@@ -176,13 +176,12 @@ class LibraryCallKit : public GraphKit {
...
   Node* generate_virtual_guard(Node* obj_klass, RegionNode* slow_region);
-  CallJavaNode* generate_method_call(vmIntrinsics::ID method_id,
-                                     bool is_virtual = false, bool is_static = false);
-  CallJavaNode* generate_method_call_static(vmIntrinsics::ID method_id) {
-    return generate_method_call(method_id, false, true);
+  CallJavaNode* generate_method_call(vmIntrinsicID method_id, bool is_virtual, bool is_static, bool res_not_null);
+  CallJavaNode* generate_method_call_static(vmIntrinsicID method_id, bool res_not_null) {
+    return generate_method_call(method_id, false, true, res_not_null);
   }
   CallJavaNode* generate_method_call_virtual(vmIntrinsics::ID method_id) {
```
simonis, line 183: `generate_method_call_virtual()` doesn't seem to be used anywhere in the code base so maybe we can drop it instead of updating it?

rwestrel: Thanks for looking at this. Removed.

src/hotspot/share/opto/ifnode.cpp
rwestrel: Done.

test/hotspot/jtreg/compiler/splitif/TestCrashAtIGVNSplitIfSubType.java
rwestrel: Done.

rwestrel:
Thanks for reviewing this and running tests.

> The testing revealed a failure in the newly introduced assertion (attached logs to the bug).

I pushed a fix. The problem is that the return type of a boxing method is marked not null but if the method is inlined late, the result from inlining may not  have a non null type.

> Also, the bug summary is way too generic and lacks any details about the actual problem. Please, update it.

Does the new one look ok?

iwanowww (approved):
(Bug summary looks much clearer now, thank.)

The fix looks good.

Test results are clean. (Tobias, thanks for submitting the patch for testing.)

test/hotspot/jtreg/compiler/splitif/TestCrashAtIGVNSplitIfSubType.java
```diff
@@ -0,0 +1,73 @@
...
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/*
+ * @test
+ * @bug 8303279
+ * @summary C2: crash in SubTypeCheckNode::sub() at IGVN split if
+ * @run main/othervm -XX:-TieredCompilation -XX:-BackgroundCompilation -XX:+UnlockDiagnosticVMOptions -XX:+StressIGVN -XX:StressSeed=598200189 TestCrashAtIGVNSplitIfSubType
```
TobiHartmann, line 28: Maybe add a `@run` without a fixed seed to give this a chance to still trigger in the future.

TobiHartmann (approved):
Looks good to me too.

simonis (approved):
Looks good to me. Thanks for the quick fix!

rwestrel:
@vnkozlov @iwanowww @simonis @TobiHartmann thanks for the reviews.

TobiHartmann (approved):

PR #14684, opened 2023-06-27
JDK-8310913 Move ReferencedKeyMap to jdk.internal so it may be shared

JimLaskey:
java.lang.runtime.ReferencedKeyMap was introduced to provide a concurrent caching scheme for Carrier objects. The technique used is generally useful for a variety of caching schemes and is being moved to be shared in other parts of the jdk. The MethodType interning case is one example.

---------
### Progress

### Issue
 * [JDK-8310913](https://bugs.openjdk.org/browse/JDK-8310913): Move ReferencedKeyMap to jdk.internal so it may be shared (**Enhancement** - P3)

### Reviewers
 * [Naoto Sato](https://openjdk.org/census#naoto) (@naotoj - **Reviewer**) ⚠️ Review applies to [a7217b66](https://git.openjdk.org/jdk/pull/14684/files/a7217b66ee01a84381c766d47f2418f084c2b108)
 * [Roger Riggs](https://openjdk.org/census#rriggs) (@RogerRiggs - **Reviewer**) ⚠️ Review applies to [fbb78778](https://git.openjdk.org/jdk/pull/14684/files/fbb7877884e6f8ef0d4313fffc935968aee72a63)
 * [Mandy Chung](https://openjdk.org/census#mchung) (@mlchung - **Reviewer**) ⚠️ Review applies to [8eff9183](https://git.openjdk.org/jdk/pull/14684/files/8eff91836f98cb33164046f2304185f11df12609)
 * [Chen Liang](https://openjdk.org/census#liach) (@liach - Author) ⚠️ Review applies to [fbb78778](https://git.openjdk.org/jdk/pull/14684/files/fbb7877884e6f8ef0d4313fffc935968aee72a63)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14684/head:pull/14684` \
`$ git checkout pull/14684`

`$ git checkout pull/14684` \
`$ git pull https://git.openjdk.org/jdk.git pull/14684/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14684`

`$ git pr show -t 14684`

</details>

<a href="https://git.openjdk.org/jdk/pull/14684.diff">https://git.openjdk.org/jdk/pull/14684.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14684#issuecomment-1613388144)


naotoj (approved):
Looks fine to me. Good to see this class is being shared, as this can be used for refactoring the locale-related caches (https://github.com/openjdk/jdk/pull/14404).

src/java.base/share/classes/jdk/internal/util/ReferencedKeyMap.java
```diff
@@ -82,7 +82,7 @@
  * Warning: This class is part of PreviewFeature.Feature.STRING_TEMPLATES.
  *          Do not rely on its availability.
  */
```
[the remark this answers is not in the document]
JimLaskey: True that.

src/java.base/share/classes/jdk/internal/util/ReferencedKeySet.java
```diff
@@ -0,0 +1,184 @@
...
+ * set.add(40_000_001L);
+ * set.add(40_000_002L);
+ * set.add(40_000_003L);
+ * set.add(40_000_004L);
+ * }
+ *
+ * @implNote Care must be given that the backing map does replacement by
+ * replacing the value in the map entry instead of deleting the old entry and
+ * adding a new entry, otherwise replaced entries may end up with a strongly
+ * referenced key. {@link HashMap} and {@link ConcurrentHashMap} are known
```
RogerRiggs, line 65: Can this requirement be backed up with an assert checking for known ok HashMaps?
It could catch mistakes (in debug builds).

RogerRiggs (approved):
LGTM

src/java.base/share/classes/jdk/internal/util/ReferencedKeySet.java
JimLaskey: It's known to be safe for HashMap and ConcurrentHashMap. Most others should fine but need some thought experiments to work it out.

src/java.base/share/classes/java/lang/invoke/MethodType.java
```diff
@@ -896,10 +906,6 @@ public boolean equals(Object x) {
         if (x instanceof MethodType mt) {
             return equals(mt);
         }
-        if (x instanceof ConcurrentWeakInternSet.WeakEntry<?> e
```
mlchung, line 899: The comment above about `WeakEntry` should also be removed.

mlchung (commented):
MethodType's `ConcurrentWeakInternSet` uses `NativeReferenceQueue` which uses native monitors (that was added for virtual threads).   This change uses the reference queue using j.u.c locks.   Has you confirmed that this is not an issue?     Or `ReferencedKeySet` can take a parameter to specify using the native reference queue?

In `ReferencedKey::equals` method, can use `Reference::refersTo` instead of `Objects.equals`.

JimLaskey:
@mlchung I missed the use of newNativeReferenceQueue. This change has been sitting in my queue for a few years, I will adapt to the new reality. Thank you for pointing it out.

src/java.base/share/classes/java/lang/invoke/MethodType.java
JimLaskey: Will update

RogerRiggs (approved):

mlchung (commented):
`WeakReferenceKey::equals` and `SoftReferenceKey::equals`:

Instead of `return Objects.equals(get(), obj);`, suggest to do:

```
    return refersTo(obj);
```

JimLaskey:
Not sure I catch the nuance. `refersTo` doesn't work if the search key is constructed. For example, if the key reference is a `record Pair(int a, int b) {}`, then `map.containsKey(new Pair(10, 20))` would locate with equals but not with `refersTo`.

RogerRiggs:
> @mlchung Not sure I catch the nuance. `refersTo` doesn't work if the search key is constructed. For example, if the key reference is a `record Pair(int a, int b) {}`, then `map.containsKey(new Pair(10, 20))` would locate with equals but not with `refersTo`.

Using `get()` on a reference can momentarily create a strong reference to the item. `refersTo()` does not.
But since your map is holding proxies to the object,  `refersTo()` would be to the proxy, not the object; so that's not an option.

src/java.base/share/classes/jdk/internal/util/ReferencedKeyMap.java
```diff
@@ -109,7 +108,7 @@ final class ReferencedKeyMap<K, V> implements Map<K, V> {
     private ReferencedKeyMap(boolean isSoft, Map<ReferenceKey<K>, V> map) {
         this.isSoft = isSoft;
         this.map = map;
-        this.stale = new ReferenceQueue<>();
+        this.stale = SharedSecrets.getJavaLangRefAccess().newNativeReferenceQueue();
```
mlchung, line 111: I don't think you want all `ReferencedKeyMap` instances use native reference queue which may cause the virtual thread to pin on the carrier thread if it synchronizes the monitor lock.

Perhaps add a new factory method that takes a `ReferenceQueue` and `MethodType` will create it with `NativeReferenceQueue`.

RogerRiggs:
> @RogerRiggs `WeakReferenceKey` is a `WeakReference` with the key as the referent. I might be missing something?

The comparison of the unwrapped key using `refersTo` is an identity comparison; but the usual comparison for Maps uses `equals`.  Using `refersTo` would limit the use of ReferencedKeyMap to cases where the original key requires identity comparision. (I missed Jim's earlier comment that pointed this out.)

mlchung:
thanks.  It may worth adding a comment about equality comparison and can't use `refersTo`

src/java.base/share/classes/jdk/internal/util/ReferencedKeyMap.java
```diff
@@ -94,54 +94,46 @@ public final class ReferencedKeyMap<K, V> implements Map<K, V> {
...
     /**
      * Private constructor.
      *
-     * @param isSoft  true if {@link SoftReference} keys are to
-     *                be used, {@link WeakReference} otherwise.
-     * @param map     backing map
+     * @param isSoft          true if {@link SoftReference} keys are to
+     *                        be used, {@link WeakReference} otherwise.
+     * @param useNativeQueue  true if uses NativeReferenceQueue
+     *                        otherwise use {@link ReferenceQueue}.
```
mlchung, line 107: no such param for this constructor.

src/java.base/share/classes/jdk/internal/util/ReferencedKeyMap.java
```diff
@@ -94,54 +94,46 @@ public final class ReferencedKeyMap<K, V> implements Map<K, V> {
...
-     *
-     * @return a new map with {@link Reference} keys
-     *
-     * @param <K> the type of keys maintained by the new map
-     * @param <V> the type of mapped values
-     */
-    public static <K, V> ReferencedKeyMap<K, V>
-    create(Supplier<Map<ReferenceKey<K>, V>> supplier) {
-        return new ReferencedKeyMap<K, V>(false, supplier.get());
+    create(boolean isSoft, boolean useNativeQueue, Supplier<Map<ReferenceKey<K>, V>> supplier) {
```
mlchung, line 132: I suggest to keep the previous `create(boolean isSoft, Supplier supplier)`  factory method that defaults to non-native reference queue, which is commonly used.   Only `MethodType` needs to use native reference queue.

test/jdk/jdk/internal/util/ReferencedKeyTest.java
```diff
@@ -53,15 +59,24 @@ static void assertTrue(boolean test, String message) {
         }
     }
 
-    static void  mapTest(boolean isSoft, Supplier<Map<ReferenceKey<Long>, String>> supplier) {
-        Map<Long, String> map = ReferencedKeyMap.create(isSoft, supplier);
+    static void mapTest(boolean isSoft, Supplier<Map<ReferenceKey<Long>, String>> supplier) {
+        Map<Long, String> map = ReferencedKeyMap.create(isSoft, false, supplier);
```
mlchung, line 63:
```suggestion
        Map<Long, String> map = ReferencedKeyMap.create(isSoft, supplier);
```

test/jdk/jdk/internal/util/ReferencedKeyTest.java
```diff
@@ -53,15 +59,24 @@ static void assertTrue(boolean test, String message) {
...
         populate(map);
         collect();
         // assertTrue(map.isEmpty() || isSoft, "Weak not collecting");
         populate(map);
         methods(map);
     }
 
+    static void setTest(boolean isSoft, Supplier<Map<ReferenceKey<Long>, ReferenceKey<Long>>> supplier) {
+        ReferencedKeySet<Long> set = ReferencedKeySet.create(isSoft, false, supplier);
```
mlchung, line 72:
```suggestion
        ReferencedKeySet<Long> set = ReferencedKeySet.create(isSoft, supplier);
```

mlchung (approved):
Looks good.  Thanks for the update.   Nit:  the test can use the 2-arg factory method.

RogerRiggs (approved):

src/java.base/share/classes/jdk/internal/util/ReferencedKeyMap.java
liach: Already implemented above.

src/java.base/share/classes/jdk/internal/util/ReferencedKeySet.java
```diff
@@ -0,0 +1,188 @@
...
+ * to be safe.
+ *
+ * @param <T> the type of elements maintained by this set
+ */
+public class ReferencedKeySet<T> extends AbstractSet<T> {
```
liach, line 71:
```suggestion
public final class ReferencedKeySet<T> extends AbstractSet<T> {
```
Make this class final.

src/java.base/share/classes/jdk/internal/util/ReferencedKeySet.java
```diff
@@ -0,0 +1,188 @@
...
+
+    /**
+     * Intern an element to the set, returning the element if newly added or the
+     * old element if previously added.
+     *
+     * @param e  element to add
+     *
+     * @return the old element if present, otherwise, the new element
+     */
+    public T intern(T e) {
```
liach, line 185: I would like to see an additional `T intern(T e, UnaryOperator<T> internFunction)`, where the `internFunction` behaves like `LocalObjectCache::normalizeKey`.

src/java.base/share/classes/jdk/internal/util/ReferencedKeyMap.java
mlchung: Yes Jim made the change per this review feedback - see [8eff918](https://github.com/openjdk/jdk/pull/14684/commits/8eff91836f98cb33164046f2304185f11df12609)

src/java.base/share/classes/jdk/internal/util/ReferencedKeySet.java
```diff
@@ -0,0 +1,188 @@
...
+     *
+     * @param isSoft          true if {@link SoftReference} elements are to
+     *                        be used, {@link WeakReference} otherwise.
+     * @param supplier        {@link Supplier} of the backing map
+     *
+     * @return a new set with {@link Reference} elements
+     *
+     * @param <E> the type of elements maintained by this set
+     */
+    public static <E> jdk.internal.util.ReferencedKeySet<E>
```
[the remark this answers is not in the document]
JimLaskey: IntelliJ has this nasty habit... Changing.

src/java.base/share/classes/jdk/internal/util/ReferencedKeySet.java
JimLaskey: Changing.

src/java.base/share/classes/jdk/internal/util/ReferencedKeyMap.java
```diff
@@ -320,15 +326,59 @@ public String toString() {
...
+     * be thrown.
+     *
+     * @param key  key to add
+     *
+     * @return the old key instance unless null then the new key instance
+     *
+     * @throws ClassCastException if {@code V} is not {@code EntryKey<K>}
+     */
+    @SuppressWarnings("unchecked")
+    K intern(K key) {
```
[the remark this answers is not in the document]
JimLaskey: Changing.

src/java.base/share/classes/jdk/internal/util/ReferencedKeySet.java
JimLaskey: This was a bit tricky. Lambdas can't be used in phase1 so messed when MethodType kicked in. So I created two paths, one for normal and one for the UnaryOperator.

liach (approved):
Thank you for the updates.

src/java.base/share/classes/jdk/internal/util/ReferencedKeyMap.java
```diff
@@ -338,34 +339,91 @@ public void removeStaleReferences() {
 
     /**
      * Puts an entry where the key and the value are the same. Used for
-     * interning values.
+     * interning values in a set.
      *
      * @implNote Requires a {@link ReferencedKeyMap} whose {@code V} type
```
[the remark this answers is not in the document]
JimLaskey: I think it still serves a warning to not use this form of the `intern` method during phase1. I will add a note that they should use an UnaryOperator instance instead of a lambda.

src/java.base/share/classes/jdk/internal/util/ReferencedKeySet.java
```diff
@@ -0,0 +1,205 @@
...
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public boolean contains(Object o) {
+        return map.containsKey((T)o);
+    }
+
+    @Override
+    public boolean add(T e) {
+        return intern(e) == null;
```
ExE-Boss, line 151: This line is wrong, as `intern(…)` will never return `null`.

--------------------------------------------------------------------------------

This is the closest to the correct implementation,
```suggestion
        return !contains(e) & intern(e) == e;
```

but may incorrectly return `true` in case of the following data race, assuming `t1e == t2e`:

1. **Thread 1** calls `contains(t1e)` and gets `false`.
2. **Thread 2** calls `intern(t2e)` and successfully adds `t2e`.
3. **Thread 1** calls `intern(t1e)` and gets back `t2e`, which is `==` to `t1e`.
4. **Thread 1** returns `true`, even though it was **Thread 2** which modified the `ReferencedKeySet`.

JimLaskey: Good catch. Your solution might be correct but I think `!contains(e)` is redundant since that is how intern starts out.

```
   static <T> T intern(ReferencedKeyMap<T, ReferenceKey<T>> setMap, T key, UnaryOperator<T> interner) {
        T value = existingKey(setMap, key);
        if (value != null) {
            return value;
        }
        key = interner.apply(key);
        return internKey(setMap, key);
    }
```

Agree? So changing to `return intern(e) == e;` should be sufficient.

The other aspect of this is that `internKey` uses `putIfAbsent` which should prevent the race (assuming `ConcurrentHashMap`).

```
    /**
     * Attempt to add key to map.
     *
     * @param setMap    {@link ReferencedKeyMap} where interning takes place
     * @param key       key to add
     *
     * @param <T> type of key
     *
     * @return the old key instance if found otherwise the new key instance
     */
    private static <T> T internKey(ReferencedKeyMap<T, ReferenceKey<T>> setMap, T key) {
        ReferenceKey<T> entryKey = setMap.entryKey(key);
        T interned;
        do {
            setMap.removeStaleReferences();
            ReferenceKey<T> existing = setMap.map.putIfAbsent(entryKey, entryKey);
            if (existing == null) {
                return key;
            } else {
                // If {@code putIfAbsent} returns non-null then was actually a
                // {@code replace} and older key was used. In that case the new
                // key was not used and the reference marked stale.
                interned = existing.get();
                if (interned != null) {
                    entryKey.unused();
                }
            }
        } while (interned == null);
        return interned;
    }
```

Agree? Anyone else want to pipe in?

RogerRiggs: ok, `intern(e) == e` is sufficient.

JimLaskey: Created https://bugs.openjdk.org/browse/JDK-8325255

ExE-Boss: While `intern(e) == e` is (mostly) sufficient, it will return a false positive when `add(e)` is called and `e` is already present in the map, the correctest implementation would be some internal API in `ReferencedKeyMap` for implementing `ReferencedKeySet::add`.

PR #14688, opened 2023-06-28
JDK-8311035: CDS should not use dump time JVM narrow Klass encoding to pre-compute Klass ids

tstuefe:
I recently spent too much time trying to understand the interleaving of narrow Klass decoding with CDS. Thanks to @iklam for clarifying some details. This patch aims to make the code easier to understand and more correct.

(*2023-07-05 Updated to fit patch description to the agreed final form*)

----

CDS narrow Klass handling plays a role for archived heap objects. 

When dumping heap objects, we must recompute their narrow Klass ids since the relative positions of archived Klass instances change compared to their live counterparts in the dump time JVM.
We recompute those narrow Klass ids using the following encoding scheme:
- base = future assumed mapping start address
- shift = dump time (!) JVMs encoding shift  (A)
see `ArchiveHeapWriter::update_header_for_requested_obj` https://github.com/openjdk/jdk/blob/c3f10e847999ec254893de5a1a5de32fd07f715a/src/hotspot/share/cds/archiveHeapWriter.cpp#L419-L425

At CDS runtime, we load the CDS archive, then place the class space behind it. We then initialize narrow Klass encoding for the resulting combined Klass range such that:
- encoding base is the range start address (mapping base)
- encoding shift is always zero
see `CompressedKlassPointers::initialize` : https://github.com/openjdk/jdk/blob/c3f10e847999ec254893de5a1a5de32fd07f715a/src/hotspot/share/oops/compressedOops.cpp#L195-L231

The lengthy comment there is misleading and partly wrong (regrettable since it was mine :)

At dump time, when initializing the JVM, we also set the encoding base to klass range start and shift to zero (also `CompressedKlassPointers::initialize`). That is the shift we later use for (A); hence, that shift is zero.

-------------------

There are minor things wrong with the current code. Mainly, the *dump time* VM's narrow Klass encoding scheme is irrelevant for the encoding employed on the future runtime archive since we recalculate the narrow Klass ids for archived heap objects. That means:

- In `CompressedKlassPointers::initialize`, there is no need to fix the encoding base and shift for the *dump time* JVM. Dump time JVM can use whatever base+shift it pleases; it can use the same code path as when CDS is off (e.g. use zero-based encoding).

- In `ArchiveHeapWriter::update_header_for_requested_obj`, we should not use the dump time JVM shift for pre-computing the narrow Klass ids. Instead, we should use the *minimal shift needed for the maximal possible future Klass range size*. The comment in ArchiveHeapWriter explains this in greater detail. Thankfully, that minimal shift is zero in today's hotspot (but Lilliput may change that).

Similarly, we should not save the dump time JVMs encoding shift as the "narrow Klass shift" into the archive but the one we use for pre-computing the narrow Klass ids.

-------------------

The patch makes the encoding scheme used for precomputing klass ids explicit and obvious. It introduces a new constant, `ArchiveHeapWriter::precomputed_narrow_klass_shift`, that holds the shift used to encode nKlass Ids (for the current hotspot, it is null). (A companion constant for the precomputed base makes little sense since the archive can be mapped everywhere; the only condition for using the precomputed narrow Klass Ids is to set the runtime base to the mapping start; I considered encoding this in a "delta" constant but Ioi convinced me otherwise).

I removed the "narrow_klass_" data holders from FileMap.

The only functional change is the fact that at dumptime, the JVM now is free to choose its base+shift for its own JVM. Encoding at runtime and for CDS=off does not change.

---------
### Progress

### Issue
 * [JDK-8311035](https://bugs.openjdk.org/browse/JDK-8311035): CDS should not use dump time JVM narrow Klass encoding to pre-compute Klass ids (**Enhancement** - P4)

### Reviewers
 * [Ioi Lam](https://openjdk.org/census#iklam) (@iklam - **Reviewer**) ⚠️ Review applies to [7ed8cad0](https://git.openjdk.org/jdk/pull/14688/files/7ed8cad0508d0b8ca9d3d1c23ecb025f9cc9f3cf)
 * @ashu-mehra (no known openjdk.org user name / role)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14688/head:pull/14688` \
`$ git checkout pull/14688`

`$ git checkout pull/14688` \
`$ git pull https://git.openjdk.org/jdk.git pull/14688/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14688`

`$ git pr show -t 14688`

</details>

<a href="https://git.openjdk.org/jdk/pull/14688.diff">https://git.openjdk.org/jdk/pull/14688.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14688#issuecomment-1611259204)


iklam:
Hi Thomas, thanks for doing the clean up.

I am not quite sure about the meaning of `precomputed_narrow_klass_base_delta`. Is it intended to be used for Lilliput, where a non-zero value may be used?

Since it's hard coded-to zero now, it's hard to observe how it would affect the calculation.

Before Liliput:

At dump time, all narrowKlass values are updated to be relative to `_requested_static_archive_bottom`, with `ArchiveHeapWriter::precomputed_narrow_klass_shift` (which is hard coded to zero)

At runtime, we always set `CompressedKlassPointers::base()` to `header()->mapped_base_address()` in here:

https://github.com/openjdk/jdk/blob/46e4ee1e80652203bd59d968ea72b27681bdf312/src/hotspot/share/cds/metaspaceShared.cpp#L1149

Currently it's impossible to set an alternative `CompressedClassSpaceBaseAddress` when using CDS:

```
  develop(size_t, CompressedClassSpaceBaseAddress, 0,                       \
          "Force the class space to be allocated at this address or "       \
          "fails VM initialization (requires -Xshare=off.")                 \
```

Also we will always have zero shifts, as we don't allow CompressedClassSpaceSize to be larger than 3G:

```
  product(size_t, CompressedClassSpaceSize, 1*G,                            \
          "Maximum size of class area in Metaspace when compressed "        \
          "class pointers are used")                                        \
          range(1*M, 3*G)                                                   \
                                                                            \
```

This means we can actually always use the archived Java objects, and this "if" block can be turned into an assert

```
  if (archive_narrow_klass_base != CompressedKlassPointers::base() ||
      narrow_klass_shift() != CompressedKlassPointers::shift()) {
    log_info(cds)("CDS heap data cannot be used because the archive was created with an incompatible narrow klass encoding mode.");
    return false;
```

I think for the current implementation, we might as well get rid of these fields in filemap.hpp. 

```
  size_t  _narrow_klass_base_delta;               // narrow Klass base and shift used to pre-calculate nKlass ids
  int     _narrow_klass_shift;                    //  in archived objects (see comment in archiveHeapWriter.hpp)
```

Having them there will create more confusion. They may be useful for the future, but it's hard to understand how these fields would be used without implementing the "future" :-)

tstuefe:
Hi @iklam,

thanks a lot for looking at this!

> Hi Thomas, thanks for doing the clean up.
> 
> I am not quite sure about the meaning of `precomputed_narrow_klass_base_delta`. Is it intended to be used for Lilliput, where a non-zero value may be used?
> 
> Since it's hard-coded to zero now, it's hard to observe how it would affect the calculation.
> 
> Before Liliput:
> 
> At dump time, all narrowKlass values are updated to be relative to `_requested_static_archive_bottom`, with `ArchiveHeapWriter::precomputed_narrow_klass_shift` (which is hard coded to zero)
> 
> At runtime, we always set `CompressedKlassPointers::base()` to `header()->mapped_base_address()` in here:
> 
> https://github.com/openjdk/jdk/blob/46e4ee1e80652203bd59d968ea72b27681bdf312/src/hotspot/share/cds/metaspaceShared.cpp#L1149
> 

You missed a hunk without this patch makes no sense :)

https://github.com/openjdk/jdk/pull/14688/files#diff-ca37be6e6608d8db9e461dffd6574bc8a2dc3d1e0e88d9ba66735d4b37f8c226

I added a new function, `CompressedKlassPointers::initialize_for_given_encoding(klass range, encoding)`, that takes both the Klass range and encoding settings, and then sets up the encoding from these settings. This function complements the pre-existing `CompressedKlassPointers::initialize(klass range)`, that only gets a Klass range and then is free to think up any encoding it wants.

We call `CompressedKlassPointers::initialize_for_given_encoding` at CDS runtime now to re-instate base and shift. We call the old `CompressedKlassPointers::initialize()` when CDS is off, or (with this patch) at CDS dumptime.

I introduce this new function to remove CDS-specific paths from `CompressedKlassPointers`. The new function is a basically a replacement for the `if (UseSharedSpaces || DumpSharedSpaces) { ... }` path from the old `CompressedKlassPointers::initialize`.

About the "delta" constant: Its point was to codify the knowledge that the pre-computed encoding is calculated such that the future encoding base must be the same as the future mapping address (hence the 0). I'm fine with removing the constant though and replacing it with clear comments.

> Currently it's impossible to set an alternative `CompressedClassSpaceBaseAddress` when using CDS:
> 
> ```
>   develop(size_t, CompressedClassSpaceBaseAddress, 0,                       \
>           "Force the class space to be allocated at this address or "       \
>           "fails VM initialization (requires -Xshare=off.")                 \
> ```
> 
> Also we will always have zero shifts, as we don't allow CompressedClassSpaceSize to be larger than 3G:
> 
> ```
>   product(size_t, CompressedClassSpaceSize, 1*G,                            \
>           "Maximum size of class area in Metaspace when compressed "        \
>           "class pointers are used")                                        \
>           range(1*M, 3*G)                                                   \
>                                                                             \
> ```
> 
> This means we can actually always use the archived Java objects, and this "if" block can be turned into an assert
> 
> ```
>   if (archive_narrow_klass_base != CompressedKlassPointers::base() ||
>       narrow_klass_shift() != CompressedKlassPointers::shift()) {
>     log_info(cds)("CDS heap data cannot be used because the archive was created with an incompatible narrow klass encoding mode.");
>     return false;
> ```
> 

I agree. We still could fail, but not here, and here we cannot deal with it.

We could still fail later if we mapped somewhere inconvenient due to ASLR, and the resulting Klass range start could not be used as an encoding base. That only can happen on Arm64 today. Today, we then assert, and this patch carries that assert over (at the start of the new `CompressedKlassPointers::initialize_for_given_encoding`).

As part of my Lilliput work, I also tackle that one. The JVM will then fall back to try a different base address and be smarter about it too. Then, as you have suggested offline, we could go and recompute the narrow Klass ids.

> I think for the current implementation, we might as well get rid of these fields in filemap.hpp.
> 
> ```
>   size_t  _narrow_klass_base_delta;               // narrow Klass base and shift used to pre-calculate nKlass ids
>   int     _narrow_klass_shift;                    //  in archived objects (see comment in archiveHeapWriter.hpp)
> ```
> 
> Having them there will create more confusion. They may be useful for the future, but it's hard to understand how these fields would be used without implementing the "future" :-)

I agree.

Cheers, Thomas

tstuefe:
@iklam : I removed the "delta" constant and replaced it with hopefully clear comments. I also removed the narrow_klass_xxx fields from FileMap. And translated runtime fail conditions to asserts in FileMapInfo::can_use_heap_region().

tstuefe:
x86 test error unrelated.

src/hotspot/share/cds/archiveHeapWriter.hpp
```diff
@@ -192,6 +192,18 @@ class ArchiveHeapWriter : AllStatic {
...
+  // Archived heap object headers carry pre-computed narrow Klass ids calculated with the
+  // following scheme:
+  // 1) the encoding base must be the mapping start address.
+  // 2) shift must be large enough to result in an encoding range that covers the future Klass range.
+  //    That Klass range is defined by CDS archive size and future class space size. Luckily, the maximum
+  //    size can be predicted: archive size is assumed to be <1G, class space size capped at 3G, and at
+  //    runtime we put both regions adjacent to each other. Therefore, future Klass range size < 4G.
+  //    Since nKlass itself is 32 bit, our encoding range len is 4G, and since we set the base directly
+  //    at mapping start, these 4G are enough. Therefore, we don't need to shift at all (shift=0).
+  static constexpr int precomputed_narrow_klass_shift = 0;
```
iklam, line 205: "future" and "runtime" seem to mean the same thing. Maybe we should consistently use "runtime"?

src/hotspot/share/oops/compressedOops.cpp
```diff
@@ -186,6 +186,28 @@ NarrowPtrStruct CompressedKlassPointers::_narrow_klass = { nullptr, 0, true };
 //  are compiled for 32bit to LP64_ONLY).
 size_t CompressedKlassPointers::_range = 0;
 
+// Given a klass range [addr, addr+len) and a given encoding scheme, assert that this scheme covers the range, then
+// set this encoding scheme. Used by CDS at runtime to re-instate the scheme used to pre-compute klass ids for
+// archived heap objects.
+void CompressedKlassPointers::initialize_for_given_encoding(address addr, size_t len, address requested_base, int requested_shift) {
```
iklam, line 192: This entire function should be inside _LP64. It should not be referenced by a 32-bit build.

iklam (approved):
Looks reasonable to me. This is a nice clean up. Just a few nits.

src/hotspot/share/oops/compressedOops.cpp
tstuefe: I'll fix this. 32-bit would be worth a whole other cleanup. Unfortunately, a lot of that coding (e.g. the `decode_xxx` functions) must still be there for 32-bit since they are only guarded by constant UseCompressedClassPointers, not by ifdefs.

src/hotspot/share/cds/archiveHeapWriter.hpp
tstuefe: Done

tstuefe:
> Looks reasonable to me. This is a nice clean up. Just a few nits.

Thanks @iklam !

src/hotspot/share/oops/compressedOops.cpp
```diff
@@ -186,83 +186,60 @@ NarrowPtrStruct CompressedKlassPointers::_narrow_klass = { nullptr, 0, true };
...
+  assert(is_valid_base(requested_base), "Address must be a valid encoding base");
+  address const end = addr + len;
+
+  const int narrow_klasspointer_bits = sizeof(narrowKlass) * 8;
+  const size_t encoding_range_size = nth_bit(narrow_klasspointer_bits + requested_shift);
+  address encoding_range_end = requested_base + encoding_range_size;
+
+  assert(requested_base <= addr && encoding_range_end >= end, "Encoding does not cover the full Klass range");
```
ashu-mehra, line 202: Is it possible for `requested_base` to ever be less than `addr`?

tstuefe: Not at the moment, no. Probably never. I separated those two mainly for cleanliness, since Klass range and encoding range are different things.

ashu-mehra: I am fine having them separately but that assert condition forced me to think if requested_base can ever be less than addr.  If not, then I guess assert for `requested_base == addr` ?

tstuefe: Sure, makes sense.

ashu-mehra (approved):

ashu-mehra:
Thanks for the changes. lgtm!

tstuefe:
Thanks @iklam and @ashu-mehra !

/integrate

PR #14692, opened 2023-06-28
8311030: ResourceBundle.handleKeySet() is racy

stsypanov:
Double-checked locking should rely on local variable to avoid racy reads from volatile field.

---------
### Progress

### Issue
 * [JDK-8311030](https://bugs.openjdk.org/browse/JDK-8311030): ResourceBundle.handleKeySet() is racy (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14692/head:pull/14692` \
`$ git checkout pull/14692`

`$ git checkout pull/14692` \
`$ git pull https://git.openjdk.org/jdk.git pull/14692/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14692`

`$ git pr show -t 14692`

</details>

<a href="https://git.openjdk.org/jdk/pull/14692.diff">https://git.openjdk.org/jdk/pull/14692.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14692#issuecomment-1611214372)

### DIFF
--- a/src/java.base/share/classes/java/util/ResourceBundle.java
+++ b/src/java.base/share/classes/java/util/ResourceBundle.java
@@ -2346,9 +2346,10 @@ public abstract class ResourceBundle {
      * @since 1.6
      */
     protected Set<String> handleKeySet() {
-        if (keySet == null) {
+      Set<String> keySet = this.keySet;
+      if (keySet == null) {
             synchronized (this) {
-                if (keySet == null) {
+                if ((keySet = this.keySet) == null) {
                     Set<String> keys = new HashSet<>();
                     Enumeration<String> enumKeys = getKeys();
                     while (enumKeys.hasMoreElements()) {
@@ -2357,7 +2358,7 @@ public abstract class ResourceBundle {
                             keys.add(key);
                         }
                     }
-                    keySet = keys;
+                    this.keySet = keySet = keys;
                 }
             }
         }


pavelrappo:
It seems like a benign race, so that the whole PR qualifies as performance improvement, not a bug fix. Do I understand it correctly?

pavelrappo:
To put this into perspective, the race you are trying to fix seems benign, so the PR qualifies as a performance improvement, not a bug fix. Is my understanding correct?

stsypanov:
It'd be a benign race in case all members of `HashSet` are `final`, but they aren't, so there are no safe publication guarantees.

RogerRiggs (approved):
Looks ok; was this a tool detected race?

naotoj (approved):
+1

stsypanov:
@RogerRiggs no tool, just fell into the sources while debugging my application

dholmes-ora:
I don't see any race here. The addition of the local serves no purpose AFAICS - we don't even save any volatile reads.

rgiulietti:
The role of the local `keySet` seems pretty useless. It doesn't save neither volatile reads nor writes.

viktorklang-ora:
@stsypanov I don't see that there's any issue with the existing code as the volatile field is only written at most once as guarded by the synchronized-block. Best case this patch avoids an extra volatile read (presuming it isn't optimized away).

dfuch:
The new code makes it more visible that 1. keySet is expected to be volatile and 2. that there is no path were code could be ordered leading to the 'wrong' value of keySet being returned. So maybe there's no race, but the new code make it more obvious.

RogerRiggs (changes requested):
Given all the comments, I don't think this is ready to integrate.

shipilev:
Yeah, current code looks like a correct volatile-bearing-DCL. `keySet` is safely published already. Introducing local variable gains us nothing here performance-wise. Arguably, readability is worse. So I would avoid doing this change.

We could have explored if we could drop `volatile` from `keySet`, but it does not look safe, given `HashMap` is not safely constructed.

naotoj (changes requested):
Retrieving the approval, per the discussion.

stsypanov:
@rgiulietti `keySet` is volatile, so providing it has already been initialized, in original code we read from it twice when doing null-check and returning the value. With local variable we read from the field only once, don't we?

rgiulietti:
True. I was mislead because the return statement does not appear in the changeset (the line is textually the same, although it has a different semantics: `keySet` now refers to a local), so I didn't count for it.

In the end, this PR is not about fixing a race, as the title seems to suggest (the original code is correct), but to avoid a volatile read, right?

stsypanov:
> In the end, this PR is not about fixing a race, as the title seems to suggest (the original code is correct), but to avoid a volatile read, right?

Yeah, probably I was wrong in my conclusion. Should I rename the ticket?

dholmes-ora:
I don't think saving one volatile read is a reasonable trade-off for the loss of readability of this code change.

PR #14699, opened 2023-06-28
8310929: Optimization for Integer.toString

wenshao:
Optimization for:
Integer.toString
Long.toString
StringBuilder#append(int)

# Benchmark Result

```
sh make/devkit/createJMHBundle.sh
bash configure --with-jmh=build/jmh/jars
make test TEST="micro:java.lang.Integers.toString*" 
make test TEST="micro:java.lang.Longs.toString*" 
make test TEST="micro:java.lang.StringBuilders.toStringCharWithInt8"
```

## 1. [aliyun_ecs_c8i.xlarge](https://help.aliyun.com/document_detail/25378.html#c8i)
* cpu : intel xeon sapphire rapids (x64)

``` diff
-Benchmark               (size)  Mode  Cnt  Score   Error  Units (baseline)
-Integers.toStringBig       500  avgt   15  6.825 ± 0.023  us/op
-Integers.toStringSmall     500  avgt   15  4.823 ± 0.023  us/op
-Integers.toStringTiny      500  avgt   15  3.878 ± 0.101  us/op

+Benchmark               (size)  Mode  Cnt  Score   Error  Units (PR Update 04 f4aa1989)
+Integers.toStringBig       500  avgt   15  6.002 ± 0.054  us/op (+13.71%)
+Integers.toStringSmall     500  avgt   15  4.025 ± 0.020  us/op (+19.82%)
+Integers.toStringTiny      500  avgt   15  3.874 ± 0.067  us/op (+0.10%)

-Benchmark            (size)  Mode  Cnt  Score   Error  Units (baseline)
-Longs.toStringBig       500  avgt   15  9.224 ± 0.021  us/op
-Longs.toStringSmall     500  avgt   15  4.621 ± 0.087  us/op

+Benchmark            (size)  Mode  Cnt  Score   Error  Units (PR Update 04 f4aa1989)
+Longs.toStringBig       500  avgt   15  7.483 ± 0.018  us/op (+23.26%)
+Longs.toStringSmall     500  avgt   15  4.020 ± 0.016  us/op (+14.95%)

-Benchmark                           Mode  Cnt     Score    Error  Units (baseline)
-StringBuilders.toStringCharWithInt8 avgt   15    89.327 ±  0.733  ns/op

+Benchmark                            Mode  Cnt   Score   Error  Units (PR Update 04 f4aa1989)
+StringBuilders.toStringCharWithInt8  avgt   15  36.639 ± 0.422  ns/op (+143.80%)

```

## 2. [aliyun_ecs_c8a.xlarge](https://help.aliyun.com/document_detail/25378.html#c8a)
* cpu : amd epc genoa (x64)

``` diff
-Benchmark               (size)  Mode  Cnt  Score   Error  Units (baseline)
-Integers.toStringBig       500  avgt   15  6.753 ± 0.007  us/op
-Integers.toStringSmall     500  avgt   15  4.470 ± 0.005  us/op
-Integers.toStringTiny      500  avgt   15  2.764 ± 0.020  us/op

+Benchmark               (size)  Mode  Cnt  Score   Error  Units (PR Update 04 f4aa1989)
+Integers.toStringBig       500  avgt   15  5.036 ± 0.005  us/op (+34.09%)
+Integers.toStringSmall     500  avgt   15  3.491 ± 0.025  us/op (+28.04%)
+Integers.toStringTiny      500  avgt   15  2.627 ± 0.552  us/op (+5.21%)

-Benchmark            (size)  Mode  Cnt   Score   Error  Units (baseline)
-Longs.toStringBig       500  avgt   15  10.087 ± 0.016  us/op
-Longs.toStringSmall     500  avgt   15   4.231 ± 0.068  us/op

+Benchmark            (size)  Mode  Cnt  Score   Error  Units (PR Update 04 f4aa1989)
+Longs.toStringBig       500  avgt   15  8.120 ± 0.010  us/op (+24.22%)
+Longs.toStringSmall     500  avgt   15  3.352 ± 0.006  us/op (+26.22%)

-Benchmark                           Mode  Cnt     Score    Error  Units (baseline)
-StringBuilders.toStringCharWithInt8 avgt   15    94.228 ±  0.150  ns/op

+Benchmark                            Mode  Cnt   Score   Error  Units (PR Update 04 f4aa1989)
+StringBuilders.toStringCharWithInt8  avgt   15  34.766 ± 0.283  ns/op (+171.03%)
```

## 3. [aliyun_ecs_c8y.xlarge](https://help.aliyun.com/document_detail/25378.html#c8y)
* cpu : aliyun yitian 710 (aarch64)
``` diff
-Benchmark               (size)  Mode  Cnt   Score   Error  Units (baseline)
-Integers.toStringBig       500  avgt   15  12.224 ± 0.488  us/op
-Integers.toStringSmall     500  avgt   15   7.243 ± 0.189  us/op
-Integers.toStringTiny      500  avgt   15   6.131 ± 0.158  us/op

+Benchmark               (size)  Mode  Cnt  Score   Error  Units (PR Update 04 f4aa1989)
+Integers.toStringBig       500  avgt   15  9.488 ± 0.404  us/op (+28.83%)
+Integers.toStringSmall     500  avgt   15  6.535 ± 0.342  us/op (+10.83%)
+Integers.toStringTiny      500  avgt   15  6.121 ± 0.266  us/op (+0.16%)

-Benchmark            (size)  Mode  Cnt   Score   Error  Units (baseline)
-Longs.toStringBig       500  avgt   15  15.301 ± 0.523  us/op
-Longs.toStringSmall     500  avgt   15   7.797 ± 0.396  us/op

+Benchmark            (size)  Mode  Cnt   Score   Error  Units (PR Update 04 f4aa1989)
+Longs.toStringBig       500  avgt   15  14.367 ± 0.437  us/op (+6.50%)
+Longs.toStringSmall     500  avgt   15   6.408 ± 0.271  us/op (+21.67%)

-Benchmark                           Mode  Cnt    Score    Error  Units (baseline)
-StringBuilders.toStringCharWithInt8 avgt   15   52.980 ±  0.786  ns/op

+Benchmark                            Mode  Cnt   Score   Error  Units (PR Update 04 f4aa1989)
+StringBuilders.toStringCharWithInt8  avgt   15  41.639 ± 1.412  ns/op (+27.23%)
```

## 4. MacBookPro M1 Pro
``` diff
-Benchmark               (size)  Mode  Cnt   Score   Error  Units (baseline)
-Integers.toStringBig       500  avgt   15  12.023 ± 0.100  us/op
-Integers.toStringSmall     500  avgt   15   4.631 ± 0.095  us/op
-Integers.toStringTiny      500  avgt   15   2.512 ± 0.036  us/op

+Benchmark               (size)  Mode  Cnt  Score   Error  Units (PR Update 04 f4aa1989)
+Integers.toStringBig       500  avgt   15  5.170 ± 0.055  us/op (+132.55%)
+Integers.toStringSmall     500  avgt   15  3.149 ± 0.019  us/op (+47.06%)
+Integers.toStringTiny      500  avgt   15  2.685 ± 0.025  us/op (-6.44%)

-Benchmark            (size)  Mode  Cnt  Score   Error  Units (baseline)
-Longs.toStringBig       500  avgt   15  8.429 ± 0.154  us/op
-Longs.toStringSmall     500  avgt   15  4.587 ± 0.054  us/op

+Benchmark            (size)  Mode  Cnt  Score   Error  Units (PR Update 04 f4aa1989)
+Longs.toStringBig       500  avgt   15  7.587 ± 0.140  us/op (+11.09%)
+Longs.toStringSmall     500  avgt   15  2.980 ± 0.026  us/op (+53.92%)

-Benchmark                            Mode  Cnt    Score    Error  Units (baseline)
-StringBuilders.toStringCharWithInt8  avgt   15  126.624 ± 58.659  ns/op

+Benchmark                            Mode  Cnt   Score    Error  Units (PR Update 04 f4aa1989)
+StringBuilders.toStringCharWithInt8  avgt   15  41.752 ± 20.337  ns/op (+203.27%)
```

## 5. Orange Pi 5 Plus

``` diff
-Benchmark               (size)  Mode  Cnt   Score   Error  Units (baseline)
-Integers.toStringBig       500  avgt   15  20.895 ± 0.245  us/op
-Integers.toStringSmall     500  avgt   15  13.042 ± 0.187  us/op
-Integers.toStringTiny      500  avgt   15   8.017 ± 0.153  us/op

+Benchmark               (size)  Mode  Cnt   Score   Error  Units (PR Update 04 f4aa1989)
+Integers.toStringBig       500  avgt   15  16.814 ± 0.276  us/op (+24.27%)
+Integers.toStringSmall     500  avgt   15  10.028 ± 0.191  us/op (+30.05%)
+Integers.toStringTiny      500  avgt   15   9.392 ± 1.476  us/op (-14.65%)

-Benchmark            (size)  Mode  Cnt   Score   Error  Units 
-Longs.toStringBig       500  avgt   15  27.222 ± 0.386  us/op
-Longs.toStringSmall     500  avgt   15  12.366 ± 0.156  us/op

+Benchmark            (size)  Mode  Cnt   Score   Error  Units (PR Update 04 f4aa1989)
+Longs.toStringBig       500  avgt   15  23.967 ± 0.336  us/op (+13.58%)
+Longs.toStringSmall     500  avgt   15   9.787 ± 0.143  us/op (+26.35%)

-Benchmark                            Mode  Cnt   Score   Error  Units (baseline)
-StringBuilders.toStringCharWithInt8  avgt   15  95.690 ± 1.582  ns/op

+Benchmark                            Mode  Cnt   Score   Error  Units (PR Update 04 f4aa1989)
+StringBuilders.toStringCharWithInt8  avgt   15  70.560 ± 1.486  ns/op (+35.61)
```

---------
### Progress

### Issue
 * [JDK-8310929](https://bugs.openjdk.org/browse/JDK-8310929): Optimization for Integer.toString (**Enhancement** - P4)

### Reviewers
 * [Claes Redestad](https://openjdk.org/census#redestad) (@cl4es - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14699/head:pull/14699` \
`$ git checkout pull/14699`

`$ git checkout pull/14699` \
`$ git pull https://git.openjdk.org/jdk.git pull/14699/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14699`

`$ git pr show -t 14699`

</details>

<a href="https://git.openjdk.org/jdk/pull/14699.diff">https://git.openjdk.org/jdk/pull/14699.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14699#issuecomment-1611798144)


src/java.base/share/classes/java/lang/Integer.java
```diff
@@ -453,16 +454,104 @@ private static void formatUnsignedIntUTF16(int val, int shift, byte[] buf, int l
...
+        if (negative) {
+            i = -i;
+        }
+        final int[] digits = DecimalDigits.DIGITS;
+
+        int off = 0;
+        final int q1 = i / 1000;
+
+        byte[] buf;
+        if (q1 == 0) {
```
rgiulietti, line 478: You could postpone the division to compute `q1` later, and here test whether `i < 1_000`.
I've no statistics, but I guess it makes sense to optimize for tiny `int`s by avoiding a division.

Glavo:
Although this method is `IntrinsicCandidate`, in reality it will only be optimized in the fluent-chain of `StringBuilder`/`StringBuffer`, so optimizing its Java implementation is meaningful.

src/java.base/share/classes/java/lang/Integer.java
wenshao: I have submitted the changes and updated the benchmark results, which shows an improvement.

Glavo:
Can we cache results for small integers?

src/java.base/share/classes/java/lang/Integer.java
```diff
@@ -453,16 +454,102 @@ private static void formatUnsignedIntUTF16(int val, int shift, byte[] buf, int l
...
-            return new String(buf, LATIN1);
-        } else {
+        if (!COMPACT_STRINGS) {
+            int size = stringSize(i);
             byte[] buf = new byte[size * 2];
             StringUTF16.getChars(i, size, buf);
             return new String(buf, UTF16);
         }
+
+        if (i == Integer.MIN_VALUE) {
```
Glavo, line 464: The more common practice is to negate positive numbers. In this way, we can remove this branch and maintain the same style as the other methods in this file.

src/java.base/share/classes/java/lang/Integer.java
```diff
@@ -514,8 +548,8 @@ static int getChars(int i, int index, byte[] buf) {
             q = i / 100;
             r = (q * 100) - i;
             i = q;
-            buf[--charPos] = DigitOnes[r];
-            buf[--charPos] = DigitTens[r];
+            charPos -= 2;
+            Unsafe.getUnsafe().putShortUnaligned(buf, Unsafe.ARRAY_BYTE_BASE_OFFSET + charPos, DigitPacks[r]);
```
liach, line 552:
```suggestion
            Unsafe.getUnsafe().putShortUnaligned(buf, Unsafe.ARRAY_BYTE_BASE_OFFSET + charPos, DigitPacks[r], false);
```
You should explicitly specify the endianness, which is platform dependent.

liach (changes requested):

wenshao:
benchmark data is update, it's ready for review @rgiulietti

src/java.base/share/classes/java/lang/StringUTF16.java
```diff
@@ -1513,6 +1515,34 @@ public static int lastIndexOfLatin1(byte[] src, int srcCount,
...
+                0x300032, 0x310032, 0x320032, 0x330032, 0x340032, 0x350032, 0x360032, 0x370032, 0x380032, 0x390032,
+                0x300033, 0x310033, 0x320033, 0x330033, 0x340033, 0x350033, 0x360033, 0x370033, 0x380033, 0x390033,
+                0x300034, 0x310034, 0x320034, 0x330034, 0x340034, 0x350034, 0x360034, 0x370034, 0x380034, 0x390034,
+                0x300035, 0x310035, 0x320035, 0x330035, 0x340035, 0x350035, 0x360035, 0x370035, 0x380035, 0x390035,
+                0x300036, 0x310036, 0x320036, 0x330036, 0x340036, 0x350036, 0x360036, 0x370036, 0x380036, 0x390036,
+                0x300037, 0x310037, 0x320037, 0x330037, 0x340037, 0x350037, 0x360037, 0x370037, 0x380037, 0x390037,
+                0x300038, 0x310038, 0x320038, 0x330038, 0x340038, 0x350038, 0x360038, 0x370038, 0x380038, 0x390038,
+                0x300039, 0x310039, 0x320039, 0x330039, 0x340039, 0x350039, 0x360039, 0x370039, 0x380039, 0x390039
+        };
+        if (isBigEndian()) {
```
liach, line 1538: If the digits always have to adapt to the endianness on initialization, should we just initialize this packed digits array to endianness and remove the endianness boolean in putInt/LongUnaligned?

wenshao: here is the endianness of each char, and putIntUnaligned is the endianness of the entire int

src/java.base/share/classes/java/lang/StringUTF16.java
```diff
@@ -1513,6 +1515,34 @@ public static int lastIndexOfLatin1(byte[] src, int srcCount,
...
+            for (int i = 0; i < digits.length; i++) {
+                digits[i] <<= 8;
```
liach, line 1540: I know this changes the char-based endian (so putInt puts the non-0 to right bytes), what I suggest is something like:
```java
digits[i] = Integer.reverseBytes(digits[i] << 8);
```

And you then remove the trailing `false` (little-endian) argument to `UNSAFE.putIntUnaligned`. 

This is safe because you don't use `PACKED_DIGITS_UTF16` with bitwise operations, unlike in UUID.

wenshao: in the getChars method, it is written from the back to the front. I run benchmark and found that the performance of using little endian is better.

src/java.base/share/classes/java/lang/Integer.java
```diff
@@ -514,14 +524,16 @@ static int getChars(int i, int index, byte[] buf) {
             q = i / 100;
             r = (q * 100) - i;
             i = q;
-            buf[--charPos] = DigitOnes[r];
-            buf[--charPos] = DigitTens[r];
+            charPos -= 2;
+            UNSAFE.putShortUnaligned(buf, Unsafe.ARRAY_BYTE_BASE_OFFSET + charPos, PACKED_DIGITS[r], false);
```
RogerRiggs, line 528: When switching to use Unsafe, `getChars` should do the array bounds check in the loop of the store index.

src/java.base/share/classes/java/lang/Long.java
```diff
@@ -549,27 +551,39 @@ static int getChars(long i, int index, byte[] buf) {
         // Get 2 digits/iteration using longs until quotient fits into an int
         while (i <= Integer.MIN_VALUE) {
             q = i / 100;
-            r = (int)((q * 100) - i);
+            charPos -= 2;
+            UNSAFE.putShortUnaligned(
+                    buf,
+                    Unsafe.ARRAY_BYTE_BASE_OFFSET + charPos,
+                    Integer.PACKED_DIGITS[(int)((q * 100) - i)],
+                    false);
```
RogerRiggs, line 559: Add the array bound check for the store of the characters.

src/java.base/share/classes/java/lang/Long.java
```diff
@@ -549,27 +551,39 @@ static int getChars(long i, int index, byte[] buf) {
...
         // We know there are at most two digits left at this point.
-        buf[--charPos] = Integer.DigitOnes[-i2];
         if (i2 < -9) {
-            buf[--charPos] = Integer.DigitTens[-i2];
+            charPos -= 2;
+            UNSAFE.putShortUnaligned(
+                    buf,
+                    Unsafe.ARRAY_BYTE_BASE_OFFSET + charPos,
+                    Integer.PACKED_DIGITS[-i2],
+                    false);
```
RogerRiggs, line 584: Replace the implicit array bounds check with an explicit array index check if using Unsafe.

src/java.base/share/classes/java/lang/StringUTF16.java
```diff
@@ -1541,14 +1576,23 @@ static int getChars(int i, int index, byte[] buf) {
...
-        putChar(buf, --charPos, Integer.DigitOnes[-i]);
         if (i < -9) {
-            putChar(buf, --charPos, Integer.DigitTens[-i]);
+            charPos -= 2;
+            UNSAFE.putIntUnaligned(
+                    buf,
+                    Unsafe.ARRAY_BYTE_BASE_OFFSET + (charPos << 1),
+                    PACKED_DIGITS_UTF16[-i]);
+        } else {
+            putChar(buf, --charPos, '0' - i);
```
RogerRiggs, line 1595: Ditto add explicit array bounds check when using Unsafe, especially since the method is used outside of the source file.  Here and in the uses of Unsafe below.

src/java.base/share/classes/java/lang/Integer.java
wenshao: The value range of the r variable is 0-99, and the length of PACKED_DIGITS is 100, There is no need to check the array boundary here.

liach: I think he means to check the `charPos` to ensure it is not out of bounds.

wenshao: Oh, I see, charPos needs to do bound checks,  I added assert with reference to the implementation of StringUTF16#putChar, is this safe enough?

liach: I think we might need to look into the code around to ensure user input cannot cause charPos to go out of bounds. If charPos is not touched by user code at all, I think assert suffices (which are enabled via -esa and is thus enabled in the jtreg test suite)

src/java.base/share/classes/java/lang/Long.java
wenshao: I added assert with reference to the implementation of StringUTF16#putChar, is this safe enough?

src/java.base/share/classes/java/lang/Long.java
wenshao: I added assert with reference to the implementation of StringUTF16#putChar, is this safe enough?

src/java.base/share/classes/java/lang/StringUTF16.java
wenshao: I added assert with reference to the implementation of StringUTF16#putChar, is this safe enough?

merykitty:
It could be worth it to have a cache for small integers to skip the calculations altogether.

wenshao:
> It could be worth it to have a cache for small integers to skip the calculations altogether.

this PR is only for calculation optimization. caching small values should be a separate PR

wenshao:
@AlanBateman can you help me to review this PR?

wenshao:
@cl4es can you help me to review this PR?

theRealAph:
> I'm wondering if a micro benchmark like this is very realistic. 

Exactly so! This is almost the canonical example of the "JMH considered harmful" talk I gave recently.

plokhotnyuk:
@wenshao How about of approach used in James Anhalt's algorithm: https://jk-jeon.github.io/posts/2022/02/jeaiii-algorithm/

It reduces number of multiplications (and store operation in case of writing by 4-8 byte words) but increases the total number of instructions for the routine.

cl4es (commented):
While I share the wariness against lookup tables (and the microbenchmarks that lend them too much credence), I think the approach in this PR is mostly an incremental improvement on an existing design and should be evaluated as such. By packing two lookup tables into one we reduce the opportunity for cache misses. The new lookup table is no larger as the two existing ones, so we're not wasting memory or sacrificing density. Having it in a single table makes it less likely that the arrays will end up in different locations on the heap or in compiled code. I see mostly positives here.

The main drawback is the proliferation of `Unsafe` usage.

Switching out the lookup table-based algorithm for something clever without a lookup table is laudable, but comes with a new set of challenges and should be done as a follow up. Since these tables can be aggressively constant-folded into the code section by our JITs it might even turn out to be a wash. 

If @RogerRiggs is happy with asserts in lieu of explicit bounds checking then I move to approve this.

src/java.base/share/classes/java/lang/StringUTF16.java
```diff
@@ -1541,14 +1576,25 @@ static int getChars(int i, int index, byte[] buf) {
...
             i = q;
-            putChar(buf, --charPos, Integer.DigitOnes[r]);
-            putChar(buf, --charPos, Integer.DigitTens[r]);
+
+            charPos -= 2;
+            assert charPos >= 0 && charPos < buf.length : "Trusted caller missed bounds check";
+            UNSAFE.putIntUnaligned(
+                    buf,
+                    Unsafe.ARRAY_BYTE_BASE_OFFSET + (charPos << 1),
+                    PACKED_DIGITS_UTF16[r]);
```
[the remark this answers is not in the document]
wenshao: Good suggestion, using the same lookup table, we can get similar performance.

```java
  int packed = (int) Integer.PACKED_DIGITS[-i];
  int inflated = ((packed & 0xFF00) << 8) | (packed & 0xFF);

  charPos -= 2;
  assert charPos >= 0 && charPos < buf.length : "Trusted caller missed bounds check";
  UNSAFE.putIntUnaligned(
          buf,
          Unsafe.ARRAY_BYTE_BASE_OFFSET + (charPos << 1),
          inflated, 
          false);
```

The performance comparison data is as follows:

```diff
-Benchmark                                 Mode  Cnt   Score   Error  Units
-StringBuilders.toStringCharWithInt8UTF16  avgt   15  26.812 ± 0.095  ns/op

+Benchmark                                 Mode  Cnt   Score   Error  Units  (use same lookup table)
+StringBuilders.toStringCharWithInt8UTF16  avgt   15  27.807 ± 0.046  ns/op
```

wenshao:
> @wenshao How about of approach used in [James Anhalt's algorithm](https://jk-jeon.github.io/posts/2022/02/jeaiii-algorithm/)?
> 
> It reduces number of multiplications ([and store operations in case of writing by 4-8 byte words](https://github.com/plokhotnyuk/jsoniter-scala/blob/b1020bafa7e2e5b7e8dd87b86a9e860975f4695e/jsoniter-scala-core/jvm/src/main/scala/com/github/plokhotnyuk/jsoniter_scala/core/JsonWriter.scala#L2004-L2023)) but increases the total number of instructions for the routine.

If the compiled code size is greater than 325 (FreqInlineSize), it will not be inlined and performance will slow down. This algorithm obviously greater than 325.

different algorithms perform differently on different types of tiny/small/big values.

The [first commit](https://github.com/openjdk/jdk/pull/14699/files/8bdda26b8bb7f6162592a07445ecd2285e4fabaa) of this PR performs better under big values. But I still use the current algorithm, with few changes, and performance can be improved in all scenarios.

wenshao:
> @wenshao Please do not rebase or force-push to an active PR as it invalidates existing review comments. Note for future reference, the bots always squash all changes into a single commit automatically as part of the integration. See [OpenJDK Developers’ Guide](https://openjdk.org/guide/#working-with-pull-requests) for more information.

to solve the problem of build failure, i did git rebase & push force.

src/java.base/share/classes/java/lang/Long.java
merykitty: Can you use `VarHandle` here?

wenshao:
@RogerRiggs Is it okay to use asserts to check boundaries?

cl4es:
A comment on bounds checks. Before `StringUTF16::getChars(int, int, byte[])` it's explicitly spelled out that the important bounds checks must have been done already:
```
    // Used by trusted callers.  Assumes all necessary bounds checks have
    // been done by the caller.
```

This is not spelled out for `Integer/Long.getChars(int/long, int, byte[])` but might as well have been since all the usage thereof are from trusted sources that calculate the indexes to fit the string inside the byte buffer.

Perhaps `Integer/Long.getChars` and `PACKED_DIGITS` should all be moved to `java.lang.StringLatin1` and be annotated with the same comment as the equivalent methods in `java.lang.StringUTF16`?

src/java.base/share/classes/java/lang/StringLatin1.java
```diff
@@ -79,6 +117,130 @@ public static byte[] inflate(byte[] value, int off, int len) {
         return ret;
     }
```
cl4es, line 119: I think this is an improvement since it better groups these latin1 formatting methods together. Add a copy of the comment from `StringUTF16` about bound-checking here?

src/java.base/share/classes/java/lang/StringUTF16.java
```diff
@@ -1531,6 +1531,7 @@ public static int lastIndexOfLatin1(byte[] src, int srcCount,
      * @return index of the most significant digit or minus sign, if present
      */
     static int getChars(int i, int index, byte[] buf) {
+        // Used by trusted callers.  Assumes all necessary bounds checks have been done by the caller.
```
cl4es, line 1534: Now these are a bit redundant since these methods are surrounded by a comment to the same effect. I prefer inline comments like these, but we should avoid having both.

cl4es (approved):
Looks good to me. Please be patient and wait for @RogerRiggs to approve this, too (today is a holiday in the US)

src/java.base/share/classes/java/lang/StringUTF16.java
wenshao: Comments are redundant, but I can't find a solution

merykitty:
We can consolidate the implementation of `Integer::toString` and `Integer::toUnsignedString` by taking the absolute value of the signed integer and move the later operation to unsigned domain. This helps remove the need of relying on much more expensive `BigInteger` when invoking `Long::toUnsignedString` and may improve the performance of `Integer::toUnsignedString` a little bit. Computing in unsigned domain is also faster (`x s/ 100 = (x * 1374389535) >> 37 - (x >> 31)` while `x u/ 100 = (x * 1374389535) >>> 37`). Folding of unsigned division is still in review but we can use `(int)(((long)x * 1374389535) >>> 37)` for `int` and `mulhi(x >>> 2, 2951479051793528259) >>> 2` for `long` directly.

Thanks.

wenshao:
> We can consolidate the implementation of `Integer::toString` and `Integer::toUnsignedString` by taking the absolute value of the signed integer and move the later operation to unsigned domain. This helps remove the need of relying on much more expensive `BigInteger` when invoking `Long::toUnsignedString` and may improve the performance of `Integer::toUnsignedString` a little bit. Computing in unsigned domain is also faster (`x s/ 100 = (x * 1374389535) >> 37 - (x >> 31)` while `x u/ 100 = (x * 1374389535) >>> 37`). Folding of unsigned division is still in review but we can use `(int)(((long)x * 1374389535) >>> 37)` for `int` and `mulhi(x >>> 2, 2951479051793528259) >>> 2` for `long` directly.
> 
> Thanks.

According to your suggestion, I tested the following code, and the performance has not improved. I'm not sure this implementation is what you think.

```java
static final byte[] INTEGER_MIN_BYTES = new byte[] {45, 50, 49, 52, 55, 52, 56, 51, 54, 52, 56};

static int getChars(int i, int index, byte[] buf) {
        // Used by trusted callers.  Assumes all necessary bounds checks have been done by the caller.
        int q, r;
        int charPos = index;

        if (i == Integer.MIN_VALUE) {
            int len = INTEGER_MIN_BYTES.length;
            System.arraycopy(INTEGER_MIN_BYTES, 0, buf, index - len, len);
            return index - len;
        }

        boolean negative = i < 0;
        if (negative) {
            i = -i;
        }

        // Generate two digits per iteration
        while (i >= 100) {
            // really: q = i / 100
            q = (int) ((((long) i) * 1374389535) >>> 37);
            r = i - (q * 100);
            i = q;
            charPos -= 2;
            assert charPos >= 0 && charPos < buf.length : "Trusted caller missed bounds check";
            UNSAFE.putShortUnaligned(buf, Unsafe.ARRAY_BYTE_BASE_OFFSET + charPos, PACKED_DIGITS[r], false);
        }

        // We know there are at most two digits left at this point.
        if (i > 9) {
            charPos -= 2;
            assert charPos >= 0 && charPos < buf.length : "Trusted caller missed bounds check";
            UNSAFE.putShortUnaligned(buf, Unsafe.ARRAY_BYTE_BASE_OFFSET + charPos, PACKED_DIGITS[i], false);
        } else {
            buf[--charPos] = (byte)('0' + i);
        }

        if (negative) {
            buf[--charPos] = (byte)'-';
        }
        return charPos;
    }
```

RogerRiggs (changes requested):
I'd be more comfortable replacing the use of Unsafe with either the ByteArray functions or VarHandles.
Using VarHandles will enable future optimizations, whereas Unsafe is a primitive tool and is brittle.

wenshao:
> I'd be more comfortable replacing the use of Unsafe with either the ByteArray functions or VarHandles. Using VarHandles will enable future optimizations, whereas Unsafe is a primitive tool and is brittle.

I also agree that using VarHandler is better, but using VarHandler in StringLatin1 causes exception, as follows:

```java
final class StringLatin1 {
    private static final VarHandle SHORT = MethodHandles.byteArrayViewVarHandle(short[].class, ByteOrder.LITTLE_ENDIAN);
}
```

```
make images
Building target 'images' in configuration 'macosx-aarch64-server-release'
Compiling up to 3452 files for java.base
Updating support/src.zip
Updating images/sec-bin.zip
Optimizing the exploded image
Error occurred during initialization of VM
java.lang.ExceptionInInitializerError
	at java.lang.invoke.VarHandle.<clinit>(java.base/VarHandle.java:2246)
	at java.lang.invoke.VarHandles.byteArrayViewHandle(java.base/VarHandles.java:258)
	at java.lang.invoke.MethodHandles.byteArrayViewVarHandle(java.base/MethodHandles.java:4553)
	at java.lang.StringLatin1.<clinit>(java.base/StringLatin1.java:84)
	at java.lang.String.equals(java.base/String.java:1863)
	at java.util.ImmutableCollections$Set12.<init>(java.base/ImmutableCollections.java:797)
	at java.util.Set.of(java.base/Set.java:487)
	at jdk.internal.reflect.Reflection.<clinit>(java.base/Reflection.java:58)
	at java.security.AccessController.doPrivileged(java.base/AccessController.java:319)
	at java.lang.reflect.AccessibleObject.<clinit>(java.base/AccessibleObject.java:524)
Caused by: java.lang.NullPointerException
	at java.lang.invoke.MethodHandleStatics.<clinit>(java.base/MethodHandleStatics.java:70)
	at java.lang.invoke.VarHandle.<clinit>(java.base/VarHandle.java:2246)
	at java.lang.invoke.VarHandles.byteArrayViewHandle(java.base/VarHandles.java:258)
	at java.lang.invoke.MethodHandles.byteArrayViewVarHandle(java.base/MethodHandles.java:4553)
	at java.lang.StringLatin1.<clinit>(java.base/StringLatin1.java:84)
	at java.lang.String.equals(java.base/String.java:1863)
	at java.util.ImmutableCollections$Set12.<init>(java.base/ImmutableCollections.java:797)
	at java.util.Set.of(java.base/Set.java:487)
	at jdk.internal.reflect.Reflection.<clinit>(java.base/Reflection.java:58)
	at java.security.AccessController.doPrivileged(java.base/AccessController.java:319)
	at java.lang.reflect.AccessibleObject.<clinit>(java.base/AccessibleObject.java:524)
```

wenshao:
In the test of Integers.toStringBig, using ByteArrayLittleEndian is about 5% slower than using Unsafe.

* Integer.getChars
```java
    static int getChars(int i, int index, byte[] buf) {
        // Used by trusted callers.  Assumes all necessary bounds checks have been done by the caller.
        int q, r;
        int charPos = index;

        boolean negative = i < 0;
        if (!negative) {
            i = -i;
        }

        // Generate two digits per iteration
        while (i <= -100) {
            q = i / 100;
            r = (q * 100) - i;
            i = q;
            charPos -= 2;
            assert charPos >= 0 && charPos < buf.length : "Trusted caller missed bounds check";
            ByteArrayLittleEndian.setShort(buf, charPos, PACKED_DIGITS[r]);
        }

        // We know there are at most two digits left at this point.
        if (i < -9) {
            charPos -= 2;
            assert charPos >= 0 && charPos < buf.length : "Trusted caller missed bounds check";
            ByteArrayLittleEndian.setShort(buf, charPos, PACKED_DIGITS[-i]);
        } else {
            buf[--charPos] = (byte)('0' - i);
        }

        if (negative) {
            buf[--charPos] = (byte)'-';
        }
        return charPos;
    }
```

* jmh test
```shell
make test TEST="micro:java.lang.Integers.toStringBig" 
```

* result
```diff
-Benchmark             (size)  Mode  Cnt  Score   Error  Units (use Unsafe)
-Integers.toStringBig     500  avgt   15  5.152 ? 0.052  us/op

+Benchmark             (size)  Mode  Cnt  Score   Error  Units (use ByteArrayLittleEndian)
+Integers.toStringBig     500  avgt   15  5.386 ? 0.020  us/op (slower 4.5%)
```

If you must not use Unsafe, I will replace it with ByteArrayLittleEndian

RogerRiggs:
The bulk of the improvement comes from the algorithmic change; using ByteArrayLittleEndian is more robust.
Thanks for the performance numbers, though hard to compare with the originals.

cl4es:
Assuming this is the M1 numbers we're still looking at a ~120% improvement. And if there's some systemic overhead to using `ByteArrayLittleEndian` compared to `Unsafe` then that's something we might improve over time. Rewiring this to use `ByteArrayLittleEndian` may be the way to go for now.

wenshao:
The performance test results of the latest version (PR Update 20 [c0f42a7c](https://github.com/openjdk/jdk/pull/14699/files/c0f42a7cfcb3783423689bb05c21b3313d2644f6) ) are as follows:

## 1. [aliyun_ecs_c8i.xlarge](https://help.aliyun.com/document_detail/25378.html#c8i)
* cpu : intel xeon sapphire rapids (x64)

```diff
-Benchmark               (size)  Mode  Cnt  Score   Error  Units (baseline)
-Integers.toStringBig       500  avgt   15  6.800 ? 0.022  us/op
-Integers.toStringSmall     500  avgt   15  4.792 ? 0.021  us/op
-Integers.toStringTiny      500  avgt   15  3.757 ? 0.081  us/op

+Benchmark               (size)  Mode  Cnt  Score   Error  Units (PR Update 20 c0f42a7c)
+Integers.toStringBig       500  avgt   15  5.894 ? 0.046  us/op (+15.37%)
+Integers.toStringSmall     500  avgt   15  4.027 ? 0.012  us/op (+18.99%)
+Integers.toStringTiny      500  avgt   15  3.491 ? 0.090  us/op (+7.61%)

-Benchmark            (size)  Mode  Cnt  Score   Error  Units (baseline)
-Longs.toStringBig       500  avgt   15  9.213 ? 0.019  us/op
-Longs.toStringSmall     500  avgt   15  4.550 ? 0.016  us/op

+Benchmark            (size)  Mode  Cnt  Score   Error  Units (PR Update 20 c0f42a7c)
+Longs.toStringBig       500  avgt   15  7.507 ? 0.011  us/op (+22.72%)
+Longs.toStringSmall     500  avgt   15  3.967 ? 0.021  us/op (+14.69%)

-Benchmark                            Mode  Cnt   Score   Error  Units (baseline)
-StringBuilders.toStringCharWithInt8  avgt   15  89.187 ? 0.236  ns/op

+Benchmark                            Mode  Cnt   Score   Error  Units (PR Update 20 c0f42a7c)
+StringBuilders.toStringCharWithInt8  avgt   15  36.125 ? 0.309  ns/op (+146.88%)
Finished running test 'micro:java.lang.StringBuilders.toStringCharWithInt8'
```



## 2. [aliyun_ecs_c8y.xlarge](https://help.aliyun.com/document_detail/25378.html#c8y)
* cpu : aliyun yitian 710 (aarch64)

```diff
-Benchmark               (size)  Mode  Cnt   Score   Error  Units (baseline)
-Integers.toStringBig       500  avgt   15  11.649 ? 0.011  us/op
-Integers.toStringSmall     500  avgt   15   6.985 ? 0.018  us/op
-Integers.toStringTiny      500  avgt   15   5.972 ? 0.013  us/op

+Benchmark               (size)  Mode  Cnt  Score   Error  Units (PR Update 20 c0f42a7c)
+Integers.toStringBig       500  avgt   15  8.957 ? 0.026  us/op (+30.05%)
+Integers.toStringSmall     500  avgt   15  6.136 ? 0.018  us/op (+13.83%)
+Integers.toStringTiny      500  avgt   15  5.753 ? 0.026  us/op (+3.80%)

-Benchmark            (size)  Mode  Cnt   Score   Error  Units (baseline)
-Longs.toStringBig       500  avgt   15  14.568 ? 0.021  us/op
-Longs.toStringSmall     500  avgt   15   7.250 ? 0.023  us/op

+Benchmark            (size)  Mode  Cnt   Score   Error  Units (PR Update 20 c0f42a7c)
+Longs.toStringBig       500  avgt   15  13.401 ? 0.012  us/op (+8.70%)
+Longs.toStringSmall     500  avgt   15   6.031 ? 0.018  us/op (+20.21%)

-Benchmark                            Mode  Cnt   Score   Error  Units (baseline)
-StringBuilders.toStringCharWithInt8  avgt   15  52.484 ? 0.534  ns/op

+Benchmark                            Mode  Cnt   Score   Error  Units (PR Update 20 c0f42a7c)
+StringBuilders.toStringCharWithInt8  avgt   15  40.410 ? 0.348  ns/op (+29.87%)
```

## 3. MacBookPro M1 Pro

```diff
-Benchmark               (size)  Mode  Cnt   Score   Error  Units (baseline)
-Integers.toStringBig       500  avgt   15  18.483 ± 2.771  us/op
-Integers.toStringSmall     500  avgt   15   4.435 ± 0.067  us/op
-Integers.toStringTiny      500  avgt   15   2.382 ± 0.063  us/op

+Benchmark               (size)  Mode  Cnt  Score   Error  Units (PR Update 20 c0f42a7c)
+Integers.toStringBig       500  avgt   15  5.392 ? 0.016  us/op (+242.78%)
+Integers.toStringSmall     500  avgt   15  3.201 ? 0.024  us/op (+38.55%)
+Integers.toStringTiny      500  avgt   15  2.141 ? 0.021  us/op (+11.25%)

-Benchmark            (size)  Mode  Cnt  Score   Error  Units (baseline)
-Longs.toStringBig       500  avgt   15  8.336 ± 0.025  us/op
-Longs.toStringSmall     500  avgt   15  4.389 ± 0.018  us/op

+Benchmark            (size)  Mode  Cnt  Score   Error  Units (PR Update 20 c0f42a7c)
+Longs.toStringBig       500  avgt   15  7.706 ? 0.015  us/op (+8.17%)
+Longs.toStringSmall     500  avgt   15  3.094 ? 0.021  us/op (+41.85%)

-Benchmark                            Mode  Cnt    Score    Error  Units (baseline)
-StringBuilders.toStringCharWithInt8  avgt   15  124.316 ± 61.017  ns/op

+Benchmark                            Mode  Cnt   Score    Error  Units (PR Update 20 c0f42a7c)
+StringBuilders.toStringCharWithInt8  avgt   15  44.497 ? 29.741  ns/op (+179.38%)
```

cl4es (approved):
Using `ByteArrayLittleEndian` cleaned up the code nicely. The overhead compared to `Unsafe` doesn't seem too concerning (maybe even some wins?)

wenshao:
Can it be merged now?

RogerRiggs (approved):
Looks good, thank you for your patience.

For future performance PRs,  please put the detailed performance data in a separate comment, not the description. The description is included in every email and the perf data gets out of date quickly.

RogerRiggs:
> /backport jdk21u

fyi, the backport command is issued on the final commit (not the PR). 
So that https://github.com/openjdk/jdk/commit/4b43c25fe382b5ee805a2d1b173fdd32d8da7fad from the bot message above.

PR #14701, opened 2023-06-28
8311064: Windows builds fail without precompiled headers after JDK-8310728

djelinski:
Please review this change to fix Windows build without precompiled headers.

After JDK-8310728:
> the compiler enforces the C++11 requirement that all functions declared inline must have a definition available in the same translation unit if they're used

When precompiled headers are used, more headers are automatically included, which allowed the build to pass without this change.

This PR adds 2 includes and un-inlines `PlatformMutex` destructor, which was needed by 3 files and not easily exposed.

Windows build passes with this change, both with and without precompiled headers. Full tier1-5 build in progress.

---------
### Progress

### Issue
 * [JDK-8311064](https://bugs.openjdk.org/browse/JDK-8311064): Windows builds fail without precompiled headers after JDK-8310728 (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14701/head:pull/14701` \
`$ git checkout pull/14701`

`$ git checkout pull/14701` \
`$ git pull https://git.openjdk.org/jdk.git pull/14701/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14701`

`$ git pr show -t 14701`

</details>

<a href="https://git.openjdk.org/jdk/pull/14701.diff">https://git.openjdk.org/jdk/pull/14701.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14701#issuecomment-1612128517)

### DIFF
--- a/src/hotspot/os/windows/gc/z/zVirtualMemory_windows.cpp
+++ b/src/hotspot/os/windows/gc/z/zVirtualMemory_windows.cpp
@@ -27,7 +27,7 @@
 #include "gc/z/zLargePages.inline.hpp"
 #include "gc/z/zMapper_windows.hpp"
 #include "gc/z/zSyscall_windows.hpp"
-#include "gc/z/zVirtualMemory.hpp"
+#include "gc/z/zVirtualMemory.inline.hpp"
 #include "utilities/align.hpp"
 #include "utilities/debug.hpp"
 
--- a/src/hotspot/os/windows/os_windows.cpp
+++ b/src/hotspot/os/windows/os_windows.cpp
@@ -5451,6 +5451,10 @@ void Parker::unpark() {
   SetEvent(_ParkHandle);
 }
 
+PlatformMutex::~PlatformMutex() {
+  DeleteCriticalSection(&_mutex);
+}
+
 // Platform Monitor implementation
 
 // Must already be locked
--- a/src/hotspot/os/windows/os_windows.inline.hpp
+++ b/src/hotspot/os/windows/os_windows.inline.hpp
@@ -65,10 +65,6 @@ inline PlatformMutex::PlatformMutex() {
   InitializeCriticalSection(&_mutex);
 }
 
-inline PlatformMutex::~PlatformMutex() {
-  DeleteCriticalSection(&_mutex);
-}
-
 inline PlatformMonitor::PlatformMonitor() {
   InitializeConditionVariable(&_cond);
 }
--- a/src/hotspot/share/gc/z/zPhysicalMemory.cpp
+++ b/src/hotspot/share/gc/z/zPhysicalMemory.cpp
@@ -27,6 +27,7 @@
 #include "gc/z/zArray.inline.hpp"
 #include "gc/z/zGlobals.hpp"
 #include "gc/z/zLargePages.inline.hpp"
+#include "gc/z/zList.inline.hpp"
 #include "gc/z/zNUMA.inline.hpp"
 #include "gc/z/zPhysicalMemory.inline.hpp"
 #include "logging/log.hpp"


dholmes-ora (commented):
Why couldn't you include `os_windows.inline.hpp` where the missing decl was needed?

djelinski:
Mostly because the same destructor is not inline in os_posix.cpp.
I could include "runtime/os.inline.hpp" in xPhysicalMemory.cpp, xVirtualMemory.cpp and zVirtualMemory.cpp, but that would add weight to non-Windows builds for no good reason.

dholmes-ora (approved):
Okay. The changes seem harmless in that regard. We should have more consistency with the os specific inline definitions.

Thanks.

stefank (commented):
@djelinski @dholmes-ora It's not obvious to me that this was the right fix for ZGC. Could you explain where in zPhysicalMemory.cpp we uses a ZListNode? If it is not and it is instead used in one of the included files the added include line should go there instead.

I also wonder where in ZVirtualMemory_windows.cpp we use zVirtualMemory.inline.hpp? The error message states that we miss `ZMemory::end` etc. So, it seems like the appropriate fix would have been to include zMemory.inline.hpp instead.

dholmes-ora:
@djelinski  hotspot changes require two reviews before integration.

@stefank  I took it on faith that the ZGC changes were appropriate, or that a ZGC reviewer would report otherwise. If a hpp file is trying to use an inline function that is a problem in itself!

dholmes-ora:
This error:
```
zPhysicalMemory.obj : error LNK2001: unresolved external symbol "public: __cdecl ZListNode<class ZMemory>::~ZListNode<class ZMemory>(void)" (??1?$[ZListNode@VZMemory](mailto:ZListNode@VZMemory)@@@@[QEAA@XZ](mailto:QEAA@XZ))
```
certainly suggests to me that zPhysicalMemory.cpp needs the ZListNode header file.

stefank:
> If a hpp file is trying to use an inline function that is a problem in itself!

I didn't say that it was used in an hpp file. I said "included files" and implicitly meant an  .inline.hpp file.

stefank:
> certainly suggests to me that zPhysicalMemory.cpp needs the ZListNode header file.

How do you know that it isn't one of the include .inline.hpp files that needs the ZListNode destructor?

djelinski:
Apologies for integrating this early, I wanted to get the CI off my back as quickly as possible, and the changes seemed innocuous enough.

The compiler/linker output does not make it easy to figure out where the offending ZListNode destructor was called. I would guess that it could be called from the implicit ZMemory destructor, will try to confirm that.

Good point about including zMemory instead of zVirtualMemory, I'll look into that.

On a side note, I assume that the split into `hpp` and `inline.hpp` was done to speed up compilation; has anyone checked recently if the split still makes sense? If not, I could do some experiments here.

stefank:
> Apologies for integrating this early, I wanted to get the CI off my back as quickly as possible, and the changes seemed innocuous enough.

Understood. Thanks. We will take care of updating the ZGC code.

> The compiler/linker output does not make it easy to figure out where the offending ZListNode destructor was called. I would guess that it could be called from the implicit ZMemory destructor, will try to confirm that.

That's exactly what happens. There is an implicit usage of the inline function from our hpp files. This propagates from ZList -> ZMemory -> ZMemoryManager -> ZPhysicalMemoryManager. This is problematic and we will think about what we think about a solution that fits the ZGC code.

> Good point about including zMemory instead of zVirtualMemory, I'll look into that.

I tested it and it works.

> On a side note, I assume that the split into hpp and inline.hpp was done to speed up compilation; has anyone checked recently if the split still makes sense? If not, I could do some experiments here.

The compile times are one reason. Another is to minimize issues with circular dependencies.

djelinski:
Thanks. In case that helps, I verified that creating explicit inline destructors for ZMemory and ZMemoryManager in zMemory.inline.hpp also fixes the zPhysicalMemory linking problem.

dholmes-ora:
> > certainly suggests to me that zPhysicalMemory.cpp needs the ZListNode header file.
> 
> How do you know that it isn't one of the include .inline.hpp files that needs the ZListNode destructor?

I don't know for certain but in the context of reviewing these changes the code change was not obviously incorrect in relation to the error message.

stefank:
> > > certainly suggests to me that zPhysicalMemory.cpp needs the ZListNode header file.
> > 
> > 
> > How do you know that it isn't one of the include .inline.hpp files that needs the ZListNode destructor?
> 
> I don't know for certain but in the context of reviewing these changes the code change was not obviously incorrect in relation to the error message.

Right, the change works, but it adds an include in a file that doesn't explicitly use that included functions. That shows that there is an issues somewhere else that we will likely show up to be a problem in the future. We've now identified that the issues is that zMemory.hpp has a dependency against the destructor in zList.inline.hpp. We'll deal with that a new RFE.

PR #14705, opened 2023-06-29
8311085: Remove implementation detail writeTo from LocalVariable(Type)

liach:
`LocalVariable` and `LocalVariableType` includes `writeTo(BufWriter)`, which should be implementation details.

See https://mail.openjdk.org/pipermail/classfile-api-dev/2023-June/000381.html for context.

This patch moves the implementation to `DirectCodeBuilder`'s original use sites; the old `b.canWriteDirect` branch   is redundant, as `writeIndex`'s implementation already performs such an optimization.

### DIFF
--- a/src/java.base/share/classes/jdk/internal/classfile/impl/DirectCodeBuilder.java
+++ b/src/java.base/share/classes/jdk/internal/classfile/impl/DirectCodeBuilder.java
@@ -24,7 +24,6 @@
  */
 package jdk.internal.classfile.impl;
 
-import java.lang.constant.MethodTypeDesc;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


src/java.base/share/classes/jdk/internal/classfile/impl/DirectCodeBuilder.java
```diff
@@ -263,14 +260,24 @@ public void writeBody(BufWriter b) {
...
                                 } else {
                                     throw new IllegalArgumentException("Unbound label in local variable type");
                                 }
                             }
+                            int length = endBci - startBci;
+                            b.writeU2(startBci);
+                            b.writeU2(length);
+                            b.writeIndex(l.name());
+                            b.writeIndex(l.type());
+                            b.writeU2(l.slot());
```
briangoetz, line 280: This idiom is repeated four times; seems like it can be factored somewhere into a helper?

briangoetz (approved):

liach:
I will merge this one. Also during development, I found a few other issues:
1. CharacterRangeTable entry writing can be shared, and I wish to move them to this new `AttributeHelpers` as well.
2. CharacterRangeInfo has documentation errors: the ending pc is inclusive instead of exclusive, as seen in Classfile and javac implementations.
/integrate

liach:
Please hold off on sponsoring this; after another look, I find that LocalVariable and LocalVariableType can probably implement `WritableElement`.

Per our previous discussion, `WritableElement` should probably not be exposed; it is currently only exposed as a type parameter upper bound in `<T extends WritableElement<?>> void writeList(List<T> list);`

I think we might convert the `WritableElement` to an internal interface: is it possible to declare the method as `void writeTo(BufWriterImpl)`, and the public method is yet hidden from the API because `BufWriterImpl` is not in an exported package? For the `writeList`, I would turn it into `<T> void writeList(List<T> list, BiConsumer<T, BufWriter> writer)` or simply remove it.

TheShermanTanker:
@liach You can turn this back to draft or run the reviewers 2 command to reduce this risk of accidental sponsorship

PR #14710, opened 2023-06-29
8311077: Fix -Wconversion warnings in jvmti code.

coleenp:
Please review change for mostly fixing return types in the constant pool and metadata to fix -Wconversion warnings in JVMTI code.
Tested with tier1-4.

### DIFF
--- a/src/hotspot/share/prims/jvmtiClassFileReconstituter.cpp
+++ b/src/hotspot/share/prims/jvmtiClassFileReconstituter.cpp
@@ -64,15 +64,15 @@ void JvmtiClassFileReconstituter::write_field_infos() {
   // Compute the real number of Java fields
   int java_fields = ik()->java_fields_count();
 
-  write_u2(java_fields);
+  write_u2(checked_cast<u2>(java_fields));
   for (JavaFieldStream fs(ik()); !fs.done(); fs.next()) {
     AccessFlags access_flags = fs.access_flags();
-    int name_index = fs.name_index();
-    int signature_index = fs.signature_index();
-    int initial_value_index = fs.initval_index();
+    u2 name_index = fs.name_index();
+    u2 signature_index = fs.signature_index();
+    u2 initial_value_index = fs.initval_index();
     guarantee(name_index != 0 && signature_index != 0, "bad constant pool index for field");
     // int offset = ik()->field_offset( index );
-    int generic_signature_index = fs.generic_signature_index();
+    u2 generic_signature_index = fs.generic_signature_index();
     AnnotationArray* anno = fields_anno == nullptr ? nullptr : fields_anno->at(fs.index());
     AnnotationArray* type_anno = fields_type_anno == nullptr ? nullptr : fields_type_anno->at(fs.index());
 
@@ -84,10 +84,10 @@ void JvmtiClassFileReconstituter::write_field_infos() {
     // JVMSpec|         attribute_info attributes[attributes_count];
     // JVMSpec|   }
 
-    write_u2(access_flags.as_int() & JVM_RECOGNIZED_FIELD_MODIFIERS);
+    write_u2(access_flags.get_flags() & JVM_RECOGNIZED_FIELD_MODIFIERS);
     write_u2(name_index);
     write_u2(signature_index);
-    int attr_count = 0;
+    u2 attr_count = 0;
     if (initial_value_index != 0) {
       ++attr_count;
     }
@@ -147,11 +147,11 @@ void JvmtiClassFileReconstituter::write_code_attribute(const methodHandle& metho
   ConstMethod* const_method = method->constMethod();
   u2 line_num_cnt = 0;
   int stackmap_len = 0;
-  int local_variable_table_length = 0;
-  int local_variable_type_table_length = 0;
+  u2 local_variable_table_length = 0;
+  u2 local_variable_type_table_length = 0;
 
   // compute number and length of attributes
-  int attr_count = 0;
+  u2 attr_count = 0;
   int attr_size = 0;
   if (const_method->has_linenumber_table()) {
     line_num_cnt = line_number_table_entries(method);
@@ -229,7 +229,7 @@ void JvmtiClassFileReconstituter::write_code_attribute(const methodHandle& metho
   }
 
   ExceptionTable exception_table(method());
-  int exception_table_length = exception_table.length();
+  u2 exception_table_length = exception_table.length();
   int code_size = const_method->code_size();
   int size =
     2+2+4 +                                // max_stack, max_locals, code_length
@@ -276,7 +276,7 @@ void JvmtiClassFileReconstituter::write_code_attribute(const methodHandle& metho
 // JVMSpec|   }
 void JvmtiClassFileReconstituter::write_exceptions_attribute(ConstMethod* const_method) {
   CheckedExceptionElement* checked_exceptions = const_method->checked_exceptions_start();
-  int checked_exceptions_length = const_method->checked_exceptions_length();
+  u2 checked_exceptions_length = const_method->checked_exceptions_length();
   int size =
     2 +                                    // number_of_exceptions
     2 * checked_exceptions_length;         // exception_index_table
@@ -307,7 +307,7 @@ void JvmtiClassFileReconstituter::write_method_parameter_attribute(const ConstMe
 
   write_attribute_name_index("MethodParameters");
   write_u4(size);
-  write_u1(length);
+  write_u1((u1)length);
   for (int index = 0; index < length; index++) {
     write_u2(parameters[index].name_cp_index);
     write_u2(parameters[index].flags);
@@ -361,7 +361,7 @@ void JvmtiClassFileReconstituter::write_signature_attribute(u2 generic_signature
 // Compute the number of entries in the InnerClasses attribute
 u2 JvmtiClassFileReconstituter::inner_classes_attribute_length() {
   InnerClassesIterator iter(ik());
-  return iter.length();
+  return checked_cast<u2>(iter.length());
 }
 
 // Write an annotation attribute.  The VM stores them in raw form, so all we need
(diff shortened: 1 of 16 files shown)


fparain (approved):

src/hotspot/share/prims/jvmtiClassFileReconstituter.cpp
```diff
@@ -396,15 +396,15 @@ void JvmtiClassFileReconstituter::write_bootstrapmethod_attribute() {
   // calculate length of attribute
   int length = sizeof(u2); // num_bootstrap_methods
   for (int n = 0; n < num_bootstrap_methods; n++) {
-    u2 num_bootstrap_arguments = cpool()->operand_argument_count_at(n);
+    int num_bootstrap_arguments = cpool()->operand_argument_count_at(n);
```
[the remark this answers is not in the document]
coleenp: I fixed both things you noticed.  Thank you!

src/hotspot/share/prims/methodComparator.cpp
```diff
@@ -75,8 +75,8 @@ bool MethodComparator::args_same(Bytecodes::Code const c_old,  Bytecodes::Code c
   case Bytecodes::_multianewarray : // fall through
   case Bytecodes::_checkcast      : // fall through
   case Bytecodes::_instanceof     : {
-    u2 cpi_old = s_old->get_index_u2();
-    u2 cpi_new = s_new->get_index_u2();
+    int cpi_old = s_old->get_index_u2();
+    int cpi_new = s_new->get_index_u2();
```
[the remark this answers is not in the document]
coleenp: I had to change these two lines because BytecodeStream::get_index_u2 returns an int, so got the warning and this didn't need to be declared with u2.  get_index_u2() could be fixed to return a u2 but I didn't want to go that far as no casts were involved in this change.

src/hotspot/share/prims/jvmtiRedefineClasses.cpp
```diff
@@ -2262,7 +2266,7 @@ void VM_RedefineClasses::rewrite_cp_refs_in_method(methodHandle method,
       {
         address p = bcp + 1;
         int cp_index = Bytes::get_Java_u2(p);
```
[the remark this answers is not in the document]
coleenp: It could be a u2, but since find_new_index's parameter is int and didn't need a cast further down, I didn't change it to one.

src/hotspot/share/prims/jvmtiRedefineClasses.cpp
```diff
@@ -2188,8 +2192,8 @@ void VM_RedefineClasses::rewrite_cp_refs_in_method(methodHandle method,
     switch (c) {
       case Bytecodes::_ldc:
       {
-        int cp_index = *(bcp + 1);
-        int new_index = find_new_index(cp_index);
+        u1 cp_index = *(bcp + 1);
```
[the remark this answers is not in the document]
coleenp: This could be a u2 to avoid confusion.  Since it's ldc, the cp_index in the ldc bytecode is only a u1 but this didn't get a Wconversion error so I should probably keep it as int.

coleenp (commented):
Thanks for your comments Matias.  Some of the changes that I didn't make were because Wconversion didn't complain and where I could avoid casting.  I think more precise types for constant pool indices and cpCache indices might be more pervasive than we want and might cause more warnings down the line.  So I avoided them.

src/hotspot/share/prims/jvmtiRedefineClasses.cpp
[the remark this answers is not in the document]
dholmes-ora: I agree - using u1 to match the spec is a good thing here.

src/hotspot/share/prims/jvmtiClassFileReconstituter.cpp
```diff
@@ -307,7 +307,7 @@ void JvmtiClassFileReconstituter::write_method_parameter_attribute(const ConstMe
 
   write_attribute_name_index("MethodParameters");
   write_u4(size);
-  write_u1(length);
+  write_u1((u1)length);
```
dholmes-ora, line 310: What is the rule for using a plain cast versus a checked_cast?

src/hotspot/share/prims/jvmtiClassFileReconstituter.cpp
```diff
@@ -394,17 +394,17 @@ void JvmtiClassFileReconstituter::write_bootstrapmethod_attribute() {
...
 
   // calculate length of attribute
-  int length = sizeof(u2); // num_bootstrap_methods
+  u4 length = sizeof(u2); // num_bootstrap_methods
   for (int n = 0; n < num_bootstrap_methods; n++) {
     u2 num_bootstrap_arguments = cpool()->operand_argument_count_at(n);
     length += sizeof(u2); // bootstrap_method_ref
     length += sizeof(u2); // num_bootstrap_arguments
-    length += sizeof(u2) * num_bootstrap_arguments; // bootstrap_arguments[num_bootstrap_arguments]
+    length += (int)sizeof(u2) * num_bootstrap_arguments; // bootstrap_arguments[num_bootstrap_arguments]
```
dholmes-ora, line 402: Not clear why the int cast is used given length is a u4?

src/hotspot/share/prims/jvmtiRawMonitor.cpp
```diff
@@ -80,7 +80,7 @@ JvmtiRawMonitor::~JvmtiRawMonitor() {
 
 bool
 JvmtiRawMonitor::is_valid() {
-  int value = 0;
+  uint64_t value = 0;
```
dholmes-ora, line 83: Why `uint64_t` here when `JvmtiEnvBase::is_valid` uses `jlong`?

src/hotspot/share/prims/jvmtiRawMonitor.cpp
```diff
@@ -382,7 +382,7 @@ int JvmtiRawMonitor::raw_wait(jlong millis, Thread* self) {
   self->_ParkEvent->reset();
   OrderAccess::fence();
 
-  intptr_t save = _recursions;
+  int save = _recursions;
```
dholmes-ora, line 385: `_recursions` is `intx`

src/hotspot/share/prims/jvmtiTrace.cpp
```diff
@@ -202,7 +202,7 @@ void JvmtiTrace::initialize() {
             if (op == '+') {
               _trace_flags[i] |= bits;
             } else {
-              _trace_flags[i] &= ~bits;
+              _trace_flags[i] &= (jbyte)~bits;
```
dholmes-ora, line 205: Looks strange that only this use of bits needs the cast?

src/hotspot/share/prims/jvmtiTrace.cpp
```diff
@@ -231,7 +231,7 @@ void JvmtiTrace::initialize() {
           if (op == '+') {
             _event_trace_flags[i] |= bits;
           } else {
-            _event_trace_flags[i] &= ~bits;
+            _event_trace_flags[i] &= (jbyte)~bits;
```
dholmes-ora, line 234: Looks strange that only this use of bits needs the cast?

src/hotspot/share/prims/methodComparator.cpp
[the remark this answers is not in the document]
dholmes-ora: I think this change looks "wrong" - the indices are supposed to be u2's, if the function returns an int that seems an error.

src/hotspot/share/runtime/jfieldIDWorkaround.hpp
```diff
@@ -84,12 +84,12 @@ class jfieldIDWorkaround: AllStatic {
     uintptr_t as_uint = (uintptr_t) id;
     return ((as_uint & checked_mask_in_place) != 0);
   }
-  static intptr_t raw_instance_offset(jfieldID id) {
+  static int raw_instance_offset(jfieldID id) {
```
dholmes-ora, line 87: This doesn't look right as we've gone from returning a 64-bit value to a 32-bit value. If it is meant to be only 32-bit I expect the callers need some adjustment too.

dholmes-ora (commented):
Sorry Coleen but this raises a lot of questions for me. I expect there are a few ways to silence these conversion warnings but many of the proposed changes don't look semantically right to me. I realize that we have a lot of inconsistencies in the way we declare and use types and that the proposed changes may be the most minimal way to fix things, but it is better to type them correctly from the start IMO and only cast at the "boundaries" if it requires a change to propagate too far.

coleenp (commented):
The task is NOT to just silence the conversion warnings but to correct the types to be more precise to not get conversion warnings.  That's what this change does. I think I've answered your questions about the types chosen.

coleenp:
@dholmes-ora The reason that these changes are broken up like this is because there are a huge amount of -Wconversion warnings and to do it right, we need to examine where the warnings are to see if there are bugs, and to correct the code to either use the right types or allow the cast with an assertion check if necessary.  The changes are not designed to be completely minimal but balanced so that people can review them and that changes that aren't necessary aren't made.
We pass around 'int' as a parameter type for values that are u2, for example.  It really matters most when we're storing that value or using it in a context that requires the narrow type.  If we're using it as an index which most constant pool and constant pool cache indices are used as, we can keep it as an int.  A u2 parameter is promoted to int so it's the same thing.
I hope I've answered your questions about these specific types in this change.

src/hotspot/share/runtime/jfieldIDWorkaround.hpp
```diff
@@ -84,12 +84,12 @@ class jfieldIDWorkaround: AllStatic {
...
     return ((as_uint & checked_mask_in_place) != 0);
   }
-  static int raw_instance_offset(jfieldID id) {
+  static intptr_t raw_instance_offset(jfieldID id) {
     uintptr_t result = (uintptr_t) id >> address_shift;
     if (VerifyJNIFields && is_checked_jfieldID(id)) {
       result &= small_offset_mask;  // cut off the hash bits
     }
-    return (int)result;
+    return result;
```
dholmes-ora, line 92: Doesn't this trigger a warning as we go from unsigned to signed?

dholmes-ora (approved):
As a general rule I'd prefer to see the lowest-level functions use the types of the actual thing they are exposing - so for anything CP related we return u1/u2/u4 - and then have higher-level code do any convenience conversions. But I realize this is tricky to deal with and there are alternatives and trade-offs in where these warnings get silenced.

Thanks.

src/hotspot/share/prims/jvmtiRawMonitor.cpp
dholmes-ora: Sorry, yes was looking at the wrong `_recursions`. `int` is fine here, `intx` is odd as the max expected recursions should not depend on 32-bit versus 64-bit.

coleenp:
Thanks David for the review.  I do change the types as low in the call stack as practical with these changes.  I think your comment above disappeared because the last change reverted that code.
/integrate

PR #14726, opened 2023-06-30
8311162: Simplify and modernize equals and hashCode for java.net

pavelrappo:
Please review this PR to use modern APIs and language features to simplify `equals` and `hashCode` for the java.net package.

Note, in NetworkInterface we could do even more and change this:

        for (InetAddress thisAddr : this.addrs) {
            boolean found = false;
            for (InetAddress thatAddr : that.addrs) {
                if (thisAddr.equals(thatAddr)) {
                    found = true;
                    break;
                }
            }
            if (!found) {
                return false;
            }
        }
        return true;

to this:

        return Set.of(that.addrs).containsAll(Set.of(this.addrs));

But arguably, the first option is already better enough than what was there before.

---------
### Progress

### Issue
 * [JDK-8311162](https://bugs.openjdk.org/browse/JDK-8311162): Simplify and modernize equals and hashCode for java.net (**Bug** - P4)

### Reviewers
 * [Michael McMahon](https://openjdk.org/census#michaelm) (@Michael-Mc-Mahon - **Reviewer**)
 * [Mark Sheppard](https://openjdk.org/census#msheppar) (@msheppar - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14726/head:pull/14726` \
`$ git checkout pull/14726`

`$ git checkout pull/14726` \
`$ git pull https://git.openjdk.org/jdk.git pull/14726/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14726`

`$ git pr show -t 14726`

</details>

<a href="https://git.openjdk.org/jdk/pull/14726.diff">https://git.openjdk.org/jdk/pull/14726.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14726#issuecomment-1614394168)

### DIFF
--- a/src/java.base/share/classes/java/net/NetworkInterface.java
+++ b/src/java.base/share/classes/java/net/NetworkInterface.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2000, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2000, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -28,6 +28,7 @@ package java.net;
 import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.NoSuchElementException;
+import java.util.Objects;
 import java.util.Spliterator;
 import java.util.Spliterators;
 import java.util.stream.Stream;
@@ -586,18 +587,13 @@ public final class NetworkInterface {
      *          {@code false} otherwise.
      * @see     java.net.InetAddress#getAddress()
      */
+    @Override
     public boolean equals(Object obj) {
         if (!(obj instanceof NetworkInterface that)) {
             return false;
         }
-        if (this.name != null ) {
-            if (!this.name.equals(that.name)) {
-                return false;
-            }
-        } else {
-            if (that.name != null) {
-                return false;
-            }
+        if (!Objects.equals(this.name, that.name)) {
+            return false;
         }
 
         if (this.addrs == null) {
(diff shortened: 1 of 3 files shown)


dfuch (approved):
LGTM - please run either `tier2` or `:jdk_net` before integrating.

dfuch:
> ```return Set.of(that.addrs).containsAll(Set.of(this.addrs));```

Since it's been checked before that the arrays have the same length it could even be:
```
    return Set.of(that.addrs).equals(Set.of(this.addrs));
```

but your proposed changes are already good enough.

Michael-Mc-Mahon (approved):
The change looks good to me, per the webrev

msheppar:
For NetworkInterface the non Set version is more straight forward and clearer, Set intersection is a little arcane
Not sure it is possible to assign the same IP address multiple times to an interface, maybe an IPv6 with different scope ids? but I think not.

pavelrappo:
> For NetworkInterface the non Set version is more straight forward and clearer, Set intersection is a little arcane
> Not sure it is possible to assign the same IP address multiple times to an interface, maybe an IPv6 with different scope ids? but I think not.

After eliminating indexes from the loops, they certainly look cleaner: prior to this change, the loops might have given an impression that there was some sort of element matching (correlating by index), which there wasn't. My personal rule of thumb is this: avoid indexing if it's only used to access the *next* element in the iteration.

That said, I'm a bit surprised that you find set equality (not a subset test, but an equality test, as we discussed with @dfuch in this PR) to be more arcane than nested loops.

Anyway, the testing has shown no issues, so unless there is anything else, I'll integrate it shortly.

msheppar:
right so, well that aside, one could consider the overhead in instantiating the Sets !!
What also stands out wrt NetworkInterface::equals is a perceived anomaly that when available the MAC address is not considered part of the "equation"

msheppar (approved):
good job!

PR #14730, opened 2023-06-30
8310743: assert(reserved_rgn != nullptr) failed: Add committed region, No reserved region found

xmas92:
JDK-8309675 tried to solve the issue of NMT reporting to much committed memory by faking the virtual addresses registered when committing physical memory. This mapping was simply the `virtual address  = physical offset in backing file + heap base`. This works when ZGC gets a contiguous heap which starts at the heap base. However fails if it does not start at the heap base or is discontiguous.

This proposal fixes this by keeping track of all the reserved virtual address ranges and by creating a mapping from the physical offsets to the actual reserved virtual address ranges. If the physical offsets committed maps to multiple reserved ranges, it is processed as multiple commits.

The develop flag `ZForceDiscontiguousHeapReservations` is introduced to test the discontiguous heap feature and interactions.

Testing: 
 * Added test on all Oracle platforms
 * GHA
 * Running tier1-3

---------
### Progress

### Issue
 * [JDK-8310743](https://bugs.openjdk.org/browse/JDK-8310743): assert(reserved_rgn != nullptr) failed: Add committed region, No reserved region found (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14730/head:pull/14730` \
`$ git checkout pull/14730`

`$ git checkout pull/14730` \
`$ git pull https://git.openjdk.org/jdk.git pull/14730/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14730`

`$ git pr show -t 14730`

</details>

<a href="https://git.openjdk.org/jdk/pull/14730.diff">https://git.openjdk.org/jdk/pull/14730.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14730#issuecomment-1614642326)

### DIFF
--- a/src/hotspot/share/gc/z/zNMT.cpp
+++ b/src/hotspot/share/gc/z/zNMT.cpp
@@ -1,0 +1,121 @@
+/*
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+#include "precompiled.hpp"
+#include "gc/z/zAddress.inline.hpp"
+#include "gc/z/zGlobals.hpp"
+#include "gc/z/zNMT.hpp"
+#include "gc/z/zVirtualMemory.hpp"
+#include "memory/allocation.hpp"
+#include "services/memTracker.hpp"
+#include "utilities/nativeCallStack.hpp"
+
+ZNMT::Reservation ZNMT::_reservations[ZMaxVirtualReservations] = {};
+size_t ZNMT::_num_reservations = 0;
+
+size_t ZNMT::reservation_index(zoffset offset, size_t* offset_in_reservation) {
+  assert(_num_reservations > 0, "at least one reservation must exist");
+
+  size_t index = 0;
+  *offset_in_reservation = untype(offset);
+  for (; index < _num_reservations; ++index) {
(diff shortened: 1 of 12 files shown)


stefank (approved):

src/hotspot/share/gc/z/zNMT.cpp
```diff
@@ -0,0 +1,121 @@
...
+    left_to_process -= sub_range_size;
+    if (left_to_process == 0) {
+      // Processed all nmt registrations
+      return;
+    }
+
+    offset_in_reservation = 0;
+  }
+
+  assert(left_to_process == 0, "everything was not commited");
```
albertnetymk, line 91: When will this be reachable? I'd expect the if-check inside the for-loop to handle exit-condition.

src/hotspot/share/gc/z/zNMT.cpp
```diff
@@ -0,0 +1,121 @@
...
+    const uintptr_t sub_range_addr = untype(reservation_start) + offset_in_reservation;
+
+    // commit / uncommit memory
+    if (commit) {
+      MemTracker::record_virtual_memory_commit((void*)sub_range_addr, sub_range_size, CALLER_PC);
+    } else {
+      if (MemTracker::enabled()) {
+        Tracker tracker(Tracker::uncommit);
+        tracker.record((address)sub_range_addr, sub_range_size);
+      }
```
albertnetymk, line 79: Sth like `record_virtual_memory_uncommit` would make this more symmetric. (Better be done in a followup PR though.)

src/hotspot/share/gc/z/zNMT.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+size_t ZNMT::reservation_index(zoffset offset, size_t* offset_in_reservation) {
+  assert(_num_reservations > 0, "at least one reservation must exist");
+
+  size_t index = 0;
+  *offset_in_reservation = untype(offset);
+  for (; index < _num_reservations; ++index) {
+    const size_t reservation_size = _reservations[index]._size;
+    if (*offset_in_reservation < reservation_size) {
+      break;
```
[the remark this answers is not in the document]
xmas92: It was purposely written like this avoiding a `fatal`/`guarantee`/`ShouldNotReachHere`, the intent is to have bugs only be caught in debug. If we return `_num_reservations` as an index all that happens is that we do not register some commit with nmt. Which seems more favourable than crashing if some bug were to slip through into production.

src/hotspot/share/gc/z/zNMT.cpp
xmas92: If we run out of reservations. That is the virtual memory reservations range is less than the physical memory offset we are committing. Such a thing should not be reachable and is a bug. But if it does happen it should be caught here.

src/hotspot/share/gc/z/zNMT.cpp
xmas92: I agree. Purposely kept this minimal as it needs to be backported to 21.

albertnetymk (approved):
I think it's better to fail-early when sth unexpected occurs, so both places warrant a should-not-reach-here; subjective though.

xmas92:
I generally agree with the fail early strategy. My personal ILW assessment of the two approaches made me favour this one. But maybe I'll reassess this in the future. 

Thanks for all reviews.

/integrate

PR #14732, opened 2023-06-30
8308103: Massive (up to ~30x) increase in C2 compilation time since JDK 17

rwestrel:
A long chain of nodes are sunk out of a loop. Every time a node is
moved out of the loop, a cast is created to pin the node out of the
loop. When its input is next sunk, the cast is removed (the cast is
replaced by its input) and a new cast is created. Some nodes on the
chain have 2 other nodes in the chain as uses. When such a node is
sunk, 2 cast nodes are created, one for each use. So as the compiler
moves forward in the chain, the number of cast to remove grows. From
some profiling, removing those casts is what takes a lot of time.

The fix I propose is, when a node is processed, to check whether a
cast at the out of loop control was already created for that node and
to reuse it.

The test case takes 6 minutes when I run it without the fix and 3
seconds with it.

---------
### Progress

### Issue
 * [JDK-8308103](https://bugs.openjdk.org/browse/JDK-8308103): Massive (up to ~30x) increase in C2 compilation time since JDK 17 (**Bug** - P3)

### Reviewers
 * [Vladimir Kozlov](https://openjdk.org/census#kvn) (@vnkozlov - **Reviewer**) ⚠️ Review applies to [cdb24902](https://git.openjdk.org/jdk/pull/14732/files/cdb24902d72bde2aae7113b9dd8d24441fbf7051)
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - **Reviewer**) ⚠️ Review applies to [cdb24902](https://git.openjdk.org/jdk/pull/14732/files/cdb24902d72bde2aae7113b9dd8d24441fbf7051)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14732/head:pull/14732` \
`$ git checkout pull/14732`

`$ git checkout pull/14732` \
`$ git pull https://git.openjdk.org/jdk.git pull/14732/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14732`

`$ git pr show -t 14732`

</details>

<a href="https://git.openjdk.org/jdk/pull/14732.diff">https://git.openjdk.org/jdk/pull/14732.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14732#issuecomment-1614658172)


src/hotspot/share/opto/loopopts.cpp
```diff
@@ -1696,7 +1696,13 @@ void PhaseIdealLoop::try_sink_out_of_loop(Node* n) {
                 cast = ConstraintCastNode::make_cast_for_type(x_ctrl, in, in_t, ConstraintCastNode::UnconditionalDependency);
               }
               if (cast != nullptr) {
-                register_new_node(cast, x_ctrl);
+                Node* prev = _igvn.hash_find_insert(cast);
+                if (prev != nullptr) {
+                  cast->destruct(&_igvn);
+                  cast = prev;
+                } else {
+                  register_new_node(cast, x_ctrl);
```
vnkozlov, line 1704: Can you move creation of `cast` here so you don't need to destroy it in case of previous cast existance?
Or it is possible that `ConstraintCastNode::make_cast_for_type() can return `null`?

test/hotspot/jtreg/compiler/loopopts/TestSinkingNodesCausesLongCompilation.java
```diff
@@ -0,0 +1,62 @@
...
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/*
+ * @test
+ * @bug 8308103
+ * @summary Massive (up to ~30x) increase in C2 compilation time since JDK 17
+ * @run main/othervm -Xcomp -XX:CompileOnly=TestSinkingNodesCausesLongCompilation::mainTest -XX:RepeatCompilation=30 TestSinkingNodesCausesLongCompilation
```
chhagedorn, line 28: You should add `-XX:+UnlockDiagnosticVMOptions` since `RepeatCompilation` is diagnostic.

test/hotspot/jtreg/compiler/loopopts/TestSinkingNodesCausesLongCompilation.java
```diff
@@ -0,0 +1,62 @@
...
+                i10 = iFld;
+                iArr2[i18] |= i11;
+            }
+        }
+        System.out.println("i9 = " + i9);
+    }
+
+    public static void main(String[] strArr) {
+        TestSinkingNodesCausesLongCompilation _instance = new TestSinkingNodesCausesLongCompilation();
+        for (int i = 0; i < 10; i++ ) {
```
chhagedorn, line 58:
```suggestion
        for (int i = 0; i < 10; i++) {
```

src/hotspot/share/opto/loopopts.cpp
rwestrel: Thanks for looking at this, Vladimir.
I'm not sure I understand what you're suggesting. Is it to not allocate a new node so it doesn't have to be destroyed if an identical node exist? But without a node it's not possible to rely on IGVN hashing?

vnkozlov: You are right. Using hash to look for existing node is smart.

vnkozlov (approved):
Good.

TobiHartmann (approved):
The fix looks good to me but I still have a hard time to comprehend that this leads to a 30x increase in compilation time. And I'm worried that we have similar issues in other code. As a follow-up, could we have `PhaseIdealLoop::register_new_node` check the hash and return an existing node?

rwestrel:
> The fix looks good to me but I still have a hard time to comprehend that this leads to a 30x increase in compilation time. And I'm worried that we have similar issues in other code. As a follow-up, could we have `PhaseIdealLoop::register_new_node` check the hash and return an existing node?

There's an exponential increase of the number of casts that are created. 

First node to be sunk is:
```
  386  RShiftI  === _ 385 316  [[ 441 417 429 ]]  !orig=[615] !jvms: TestSinkingNodesCausesLongCompilation::mainTest @ bci:104 (line 46)                                                                                                                                          
 ```
It has 3 uses so 3 casts are created:
```
 1137  CastII  === 426 385  [[ ]]  #int unconditional dependency                                                                                                                                                                                                                  
 1139  CastII  === 414 385  [[ ]]  #int unconditional dependency                                                                                                                                                                                                                  
 1141  CastII  === 438 385  [[ ]]  #int unconditional dependency     
```

Next:
```  
385  AddI  === _ 384 605  [[ 1141 1137 1139 ]]  !jvms: TestSinkingNodesCausesLongCompilation::mainTest @ bci:99 (line 45)
```
(input of previous one)
The 3 previous casts are removed and new ones are created:
```
 1143  CastII  === 414 384  [[ ]]  #int unconditional dependency                                                                                                                                                                                                                  
 1145  CastII  === 426 384  [[ ]]  #int unconditional dependency                                                                                                                                                                                                                  
 1147  CastII  === 438 384  [[ ]]  #int unconditional dependency              
```

next:
```
  384  LShiftI  === _ 605 383  [[ 1147 1143 1145 ]]  !jvms: TestSinkingNodesCausesLongCompilation::mainTest @ bci:99 (line 45) 
```
input of previous one. Same as step before, 3 just created casts are moved and new ones created:
```
 1149  CastII  === 426 605  [[ ]]  #int unconditional dependency                                                                                                                                                                                                                  
 1151  CastII  === 414 605  [[ ]]  #int unconditional dependency                                                                                                                                                                                                                  
 1153  CastII  === 438 605  [[ ]]  #int unconditional dependency                
```

next:
```
  605  RShiftI  === _ 606 316  [[ 1146 1153 1142 1144 1149 1151 ]]  !orig=[386],[615] !jvms: TestSinkingNodesCausesLongCompilation::mainTest @ bci:104 (line 46)
```
which was input to both 384 and 385 just sunk. It has 6 uses. The 3 casts above and 3 clones of 385.
So 3 casts are removed and 6 casts are created:
```
 1155  CastII  === 414 606  [[ ]]  #int unconditional dependency                                                                                                                                                                                                                  
 1157  CastII  === 426 606  [[ ]]  #int unconditional dependency                                                                                                                                                                                                                  
 1159  CastII  === 426 606  [[ ]]  #int unconditional dependency                                                                                                                                                                                                                  
 1161  CastII  === 414 606  [[ ]]  #int unconditional dependency                                                                                                                                                                                                                  
 1163  CastII  === 438 606  [[ ]]  #int unconditional dependency                                                                                                                                                                                                                  
 1165  CastII  === 438 606  [[ ]]  #int unconditional dependency                                                                                                                                                                                                                  
```
The same sequence of nodes repeats and every 3 nodes there's a RShiftI and the number of clones double. At the last step, the number of casts is 12288. That happens after 39 nodes are sunk.

All of this seems pretty specific to that transformation so a local fix seems good enough to me.

TobiHartmann:
Makes sense. Thanks for the details, Roland.

test/hotspot/jtreg/compiler/loopopts/TestSinkingNodesCausesLongCompilation.java
rwestrel: Thanks for the comments. I updated the change.

chhagedorn (approved):
Looks good to me, too!

rwestrel:
@vnkozlov @TobiHartmann @chhagedorn thanks for the reviews.

PR #14735, opened 2023-06-30
8311102: Write annotations in the classfile dumped by SA

ashu-mehra:
Please review this PR that enables ClassWriter to write annotations to the class file being dumped.

The fields annotations are stored in `Annotations::_fields_annotations` which is of type `Array<Array<u1>*>`. There is no class in SA that can represent it. I have added ArrayOfU1Array to correspond to the type `Array<Array<u1>*>` and it works. I believe there are better approaches but that would require a bit more refactoring of the classes representing Array types. I will leave that for future work for now.

Testing: `test/hotspot/jtreg/serviceability/sa` and `test/jdk/sun/tools/jhsdb`
Tested it manually by dumping j.l.String class and comparing the annotations with the original class using javap.
The test case mentioned in [JDK-8311101](https://bugs.openjdk.org/browse/JDK-8311101) would provide better coverage.

---------
### Progress

### Issue
 * [JDK-8311102](https://bugs.openjdk.org/browse/JDK-8311102): Write annotations in the classfile dumped by SA (**Sub-task** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14735/head:pull/14735` \
`$ git checkout pull/14735`

`$ git checkout pull/14735` \
`$ git pull https://git.openjdk.org/jdk.git pull/14735/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14735`

`$ git pr show -t 14735`

</details>

<a href="https://git.openjdk.org/jdk/pull/14735.diff">https://git.openjdk.org/jdk/pull/14735.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14735#issuecomment-1614753258)

### DIFF
--- a/src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/oops/Annotations.java
+++ b/src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/oops/Annotations.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ *
+ */
+
+package sun.jvm.hotspot.oops;
+
+import java.io.*;
+import java.util.*;
+import sun.jvm.hotspot.debugger.*;
+import sun.jvm.hotspot.interpreter.Bytecodes;
+import sun.jvm.hotspot.runtime.*;
+import sun.jvm.hotspot.types.*;
+import sun.jvm.hotspot.utilities.*;
+import sun.jvm.hotspot.utilities.Observable;
+import sun.jvm.hotspot.utilities.Observer;
+
+// An Annotation is an oop containing annotations as described in the class file
+
+public class Annotations extends Metadata {
+  static {
+    VM.registerVMInitializedObserver(new Observer() {
+        public void update(Observable o, Object data) {
+          initialize(VM.getVM().getTypeDataBase());
+        }
+      });
+  }
+
+  public Annotations(Address addr) {
+    super(addr);
+  }
+
+  public boolean isAnnotations()            { return true; }
+
+  private static synchronized void initialize(TypeDataBase db) throws WrongTypeException {
+    Type type = db.lookupType("Annotations");
+    classAnnotations = type.getAddressField("_class_annotations");
+    fieldsAnnotations = type.getAddressField("_fields_annotations");
+    classTypeAnnotations = type.getAddressField("_class_type_annotations");
+    fieldsTypeAnnotations = type.getAddressField("_fields_type_annotations");
+  }
+
+  private static AddressField classAnnotations;
+  private static AddressField fieldsAnnotations;
+  private static AddressField classTypeAnnotations;
+  private static AddressField fieldsTypeAnnotations;
+
+  public U1Array getClassAnnotations() {
+    Address addr = classAnnotations.getValue(getAddress());
+    return VMObjectFactory.newObject(U1Array.class, addr);
+  }
+
+  public U1Array getFieldAnnotations(int fieldIndex) {
+    Address addr = fieldsAnnotations.getValue(getAddress());
+    ArrayOfU1Array annotationsArray = VMObjectFactory.newObject(ArrayOfU1Array.class, addr);
+    if (annotationsArray != null) {
+      addr = annotationsArray.at(fieldIndex);
+      return VMObjectFactory.newObject(U1Array.class, addr);
+    } else {
+      return null;
+    }
+  }
+
+  public U1Array getClassTypeAnnotations() {
+    Address addr = classTypeAnnotations.getValue(getAddress());
+    return VMObjectFactory.newObject(U1Array.class, addr);
+  }
+
+  public U1Array getFieldTypeAnnotations(int fieldIndex) {
+    Address addr = fieldsTypeAnnotations.getValue(getAddress());
+    ArrayOfU1Array annotationsArray = VMObjectFactory.newObject(ArrayOfU1Array.class, addr);
+    if (annotationsArray != null) {
+      addr = annotationsArray.at(fieldIndex);
+      return VMObjectFactory.newObject(U1Array.class, addr);
+    } else {
+      return null;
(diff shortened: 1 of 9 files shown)


ashu-mehra:
Looking for reviewers for this. TIA.

src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/oops/Annotations.java
```diff
@@ -0,0 +1,102 @@
...
+  private static AddressField fieldsTypeAnnotations;
+
+  public U1Array getClassAnnotations() {
+    Address addr = classAnnotations.getValue(getAddress());
+    return VMObjectFactory.newObject(U1Array.class, addr);
+  }
+
+  public U1Array getFieldAnnotations(int fieldIndex) {
+    Address addr = fieldsAnnotations.getValue(getAddress());
+    ArrayOfU1Array annotationsArray = VMObjectFactory.newObject(ArrayOfU1Array.class, addr);
```
plummercj, line 74: How about caching this result so you don't need to allocate a new object every time this API is called. Same thing in `getFieldTypeAnnotations()`.

src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/oops/ConstMethod.java
```diff
@@ -429,18 +429,74 @@ private boolean hasMethodAnnotations() {
...
 
-  private boolean hasDefaultAnnotations() {
-    return (getFlags() & HAS_DEFAULT_ANNOTATIONS) != 0;
+  public U1Array getParameterAnnotations() {
+    if (hasParameterAnnotations()) {
+      int offset = 1;
+      if (hasMethodAnnotations()) {
+        offset += 1;
+      }
+      Address addr = getAddress().getAddressAt((getSize() - offset) * VM.getVM().getAddressSize());
```
plummercj, line 451: A comment on the address computation would be useful here and in the changes below.

plummercj (changes requested):
Overall the changes look good. Just a few minor suggestions.

tstuefe (commented):
Hi @ashu-mehra,

looks mostly good. Some nits inline.

I tried it out and it works fine.

What would be needed to make the Annotations appear in the "printall" command? I was somehow expecting to see at least something like "Annotation@xxxx".

Cheers, Thomas

ashu-mehra:
> What would be needed to make the Annotations appear in the "printall" command? I was somehow expecting to see at least something like "Annotation@xxxx".

I am not sure what all details `printall` is expected to emit out. Looking at the code, printall doesn't seem to use ClassWriter. It uses HTMLGenerator to format the method data. I can emit something like "Annotation@xxxx" but it would be more useful if it can display the contents of the annotations. Unfortunately HTMLGenerator doesn't understand Annotations at all. Probably it is better left for another task.

src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/oops/ConstMethod.java
ashu-mehra: Added a comment about the layout of the annotation pointers.

src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/oops/ConstMethod.java
```diff
@@ -429,18 +429,74 @@ private boolean hasMethodAnnotations() {
...
 
+  public U1Array getTypeAnnotations() {
+    if (hasTypeAnnotations()) {
+      int offset = 1;
+      if (hasMethodAnnotations()) {
+        offset += 1;
+      }
+      if (hasParameterAnnotations()) {
+        offset += 1;
+      }
```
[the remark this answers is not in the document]
ashu-mehra: Done.

src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/oops/Annotations.java
ashu-mehra: I think VMObjectFactory is a better place to implement the caching behavior so that all such patterns can benefit from it. I think it is better addressed in another task.

src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/oops/ConstMethod.java
```diff
@@ -429,18 +429,74 @@ private boolean hasMethodAnnotations() {
...
+      if (hasTypeAnnotations()) {
+        offset += 1;
+      }
+      Address addr = getAddress().getAddressAt((getSize() - offset) * VM.getVM().getAddressSize());
```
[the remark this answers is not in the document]
ashu-mehra: Done.

src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/oops/InstanceKlass.java
```diff
@@ -859,6 +864,42 @@ public U2Array getInnerClasses() {
...
   }
 
+  public U1Array getClassAnnotations() {
+    Annotations annotations = getAnnotations();
+    if (annotations != null) {
+      return annotations.getClassAnnotations();
+    } else {
+      return null;
+    }
+  }
```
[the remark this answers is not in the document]
ashu-mehra: I would keep it as is following existing code pattern.

src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/oops/Annotations.java
plummercj: I think maybe you misunderstood what I meant by "cache". I'm not talking about a hashmap of weak references or anything like that. Just add a `ArrayOfU1Array annotationsArray` field to the Annotations object and store the result there.

src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/oops/ConstMethod.java
```diff
@@ -666,4 +621,67 @@ private long offsetOfExceptionTable() {
...
+  private Address getAddressAtOffset(int offsetInWords) {
+    return getAddress().getAddressAt(offsetInWords * VM.getVM().getAddressSize());
+  }
+
+  // Pointers to annotations are stored towards the end of the ConstMethod in following format:
+  //
+  //   start of ConstMethod                          end of ConstMethod
+  //            |                                           |
+  //            V                                           V
+  //            | ... | default | type | parameter | method |
```
plummercj, line 641: So the `...` part is represented by `getSize()`? It would be good to call that out. Also point out that each of the annotations pointers is optional.

tstuefe:
> > What would be needed to make the Annotations appear in the "printall" command? I was somehow expecting to see at least something like "Annotation@xxxx".
> 
> I am not sure what all details `printall` is expected to emit out. Looking at the code, printall doesn't seem to use ClassWriter. It uses HTMLGenerator to format the method data. I can emit something like "Annotation@xxxx" but it would be more useful if it can display the contents of the annotations. Unfortunately HTMLGenerator doesn't understand Annotations at all. Probably it is better left for another task.

No problem at all. I only thought about this because it would have making a regression test very easy (there is a jtreg test that calls printall and then parses the output).

src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/oops/ConstMethod.java
ashu-mehra: Updated the comment.

src/jdk.hotspot.agent/share/classes/sun/jvm/hotspot/oops/Annotations.java
ashu-mehra: Got it. Updated the code as suggested.

plummercj (approved):

tstuefe (approved):
Looks good, thanks!

ashu-mehra:
@plummercj @tstuefe thank you for reviewing this.

PR #14738, opened 2023-06-30
8311170: Simplify and modernize equals and hashCode in security area

pavelrappo:
Please review this PR to use modern APIs and language features to simplify `equals` and `hashCode` in security area.

I understand that security area is sensitive and a non-expert, such as myself, should tread carefully; so below are my notes to assist the review.

* Unlike `hashCode`, non-secure `equals` implementations are typically short-circuit. But because of "timing attacks", we seem to have specialized implementations, such as `java.security.MessageDigest.isEqual(byte[], byte[])` and a more general `sun.security.util.ByteArrays.isEqual(byte[], int, int, byte[], int, int)`. So while reviewing this PR, take an opportunity to audit the affected `equals` implementations: perhaps some of them need to become secure, not modern. I have no domain knowledge to tell those cases apart, I only note that those cases exist.

* This PR sacrifices compatibility for pragmatism: it changes some `hashCode` implementations to produce different values than before to allow more utilization of methods from `Objects` and `Arrays`. To my mind, those changes are **benign**. If you disagree, I'd be happy to discuss that and/or retract the concerning part of the change.

* BitArray could be a topic of its own, but I'll do my best to be concise.

    * Truth to be told, BitArray's `equals` and `hashCode` are not used anywhere in source, and `equals` is only used in one test. For that reason, I refrained from reimplementing internals of `BitArray` using more general `java.util.BitSet`: too much effort and risk for almost nothing.
    * Speaking of `BitSet`-powered `BitArray`. Such an implementation is not for the faint of heart: there's too much impedance mismatch between data structures that those classes use to store bits. That said, for the sake of testing that it is possible and that I understand the `BitArray` correctly, I actually implemented it using `BitSet`. While that implementation is **NOT** part of this PR, you can have a look at it [here](https://cr.openjdk.org/~prappo/8311170/BitArray.java).

* One suggestion to consider is to change this somewhat arcane piece in java.security.UnresolvedPermission.equals:

          // check certs
          if (this.certs == null && that.certs != null ||
              this.certs != null && that.certs == null ||
              this.certs != null &&
                 this.certs.length != that.certs.length) {
              return false;
          }
  
          int i,j;
          boolean match;
  
          for (i = 0; this.certs != null && i < this.certs.length; i++) {
              match = false;
              for (j = 0; j < that.certs.length; j++) {
                  if (this.certs[i].equals(that.certs[j])) {
                      match = true;
                      break;
                  }
              }
              if (!match) return false;
          }
  
          for (i = 0; that.certs != null && i < that.certs.length; i++) {
              match = false;
              for (j = 0; j < this.certs.length; j++) {
                  if (that.certs[i].equals(this.certs[j])) {
                      match = true;
                      break;
                  }
              }
              if (!match) return false;
          }
          return true;

  to this:

          // check certs
          if (this.certs == null && that.certs != null ||
              this.certs != null && that.certs == null)
              return false;
          
          if (this.certs == null) {
              assert that.certs == null;
              return false;
          }
          
          return Set.of(this.certs).equals(Set.of(that.certs));

### DIFF
--- a/src/java.base/share/classes/javax/crypto/CryptoPolicyParser.java
+++ b/src/java.base/share/classes/javax/crypto/CryptoPolicyParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1999, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1999, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -29,6 +29,7 @@ import java.io.*;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.Hashtable;
+import java.util.Objects;
 import java.util.Vector;
 import static java.util.Locale.ENGLISH;
 
@@ -616,20 +617,18 @@ final class CryptoPolicyParser {
          * Calculates a hash code value for the object.  Objects
          * which are equal will also have the same hashcode.
          */
+        @Override
         public int hashCode() {
             int retval = cryptoPermission.hashCode();
-            if (alg != null) retval ^= alg.hashCode();
-            if (exemptionMechanism != null) {
-                retval ^= exemptionMechanism.hashCode();
-            }
+            retval ^= Objects.hashCode(alg);
+            retval ^= Objects.hashCode(exemptionMechanism);
             retval ^= maxKeySize;
-            if (checkParam) retval ^= 100;
-            if (algParamSpec != null) {
-                retval ^= algParamSpec.hashCode();
-            }
+            retval ^= (checkParam ? 100 : 0);
+            retval ^= Objects.hashCode(algParamSpec);
             return retval;
         }
 
+        @Override
         public boolean equals(Object obj) {
             if (obj == this)
                 return true;
@@ -637,12 +636,8 @@ final class CryptoPolicyParser {
             if (!(obj instanceof CryptoPermissionEntry that))
                 return false;
 
-            if (this.cryptoPermission == null) {
-                if (that.cryptoPermission != null) return false;
-            } else {
-                if (!this.cryptoPermission.equals(
-                                                 that.cryptoPermission))
-                    return false;
+            if (Objects.equals(this.cryptoPermission, that.cryptoPermission)) {
+                return false;
             }
 
             if (this.alg == null) {
@@ -652,15 +647,11 @@ final class CryptoPolicyParser {
                     return false;
             }
 
-            if (!(this.maxKeySize == that.maxKeySize)) return false;
+            if (this.maxKeySize != that.maxKeySize) return false;
 
             if (this.checkParam != that.checkParam) return false;
 
-            if (this.algParamSpec == null) {
-                return that.algParamSpec == null;
-            } else {
-                return this.algParamSpec.equals(that.algParamSpec);
-            }
+            return Objects.equals(this.algParamSpec, that.algParamSpec);
         }
     }
 
--- a/src/java.base/share/classes/javax/security/cert/Certificate.java
+++ b/src/java.base/share/classes/javax/security/cert/Certificate.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1997, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -31,6 +31,7 @@ import java.security.NoSuchAlgorithmException;
 import java.security.NoSuchProviderException;
 import java.security.InvalidKeyException;
 import java.security.SignatureException;
+import java.util.Arrays;
 
 /**
  * <p>Abstract class for managing a variety of identity certificates.
@@ -81,42 +82,30 @@ public abstract class Certificate {
      * @return true if the encoded forms of the two certificates
      *         match, false otherwise.
      */
+    @Override
     public boolean equals(Object other) {
         if (this == other)
             return true;
         if (!(other instanceof Certificate))
             return false;
         try {
-            byte[] thisCert = this.getEncoded();
-            byte[] otherCert = ((Certificate)other).getEncoded();
-
-            if (thisCert.length != otherCert.length)
-                return false;
-            for (int i = 0; i < thisCert.length; i++)
-                 if (thisCert[i] != otherCert[i])
-                     return false;
-            return true;
+            return Arrays.equals(this.getEncoded(),
+                    ((Certificate)other).getEncoded());
         } catch (CertificateException e) {
             return false;
         }
     }
 
     /**
-     * Returns a hashcode value for this certificate from its
-     * encoded form.
-     *
-     * @return the hashcode value.
+     * {@return a hashcode value for this certificate from its
+     * encoded form}
      */
+    @Override
     public int hashCode() {
-        int     retval = 0;
         try {
-            byte[] certData = this.getEncoded();
-            for (int i = 1; i < certData.length; i++) {
-                 retval += certData[i] * i;
-            }
-            return (retval);
+            return Arrays.hashCode(this.getEncoded());
         } catch (CertificateException e) {
-            return (retval);
+            return 0;
         }
     }
 
--- a/src/java.base/share/classes/sun/security/util/BitArray.java
+++ b/src/java.base/share/classes/sun/security/util/BitArray.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1997, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -26,6 +26,7 @@
 package sun.security.util;
 
 import java.io.ByteArrayOutputStream;
+import java.util.Arrays;
 
 import jdk.internal.util.Preconditions;
 
@@ -68,7 +69,7 @@ public class BitArray {
      * Creates a BitArray of the specified size, initialized from the
      * specified byte array. The most significant bit of {@code a[0]} gets
      * index zero in the BitArray. The array must be large enough to specify
-     * a value for every bit of the BitArray. i.e. {@code 8*a.length <= length}.
+     * a value for every bit of the BitArray, i.e. {@code 8*a.length >= length}.
      */
     public BitArray(int length, byte[] a) throws IllegalArgumentException {
         this(length, a, 0);
@@ -79,7 +80,7 @@ public class BitArray {
      * specified byte array starting at the specified offset.  The most
      * significant bit of {@code a[ofs]} gets index zero in the BitArray.
      * The array must be large enough to specify a value for every bit of
-     * the BitArray, i.e. {@code 8*(a.length - ofs) <= length}.
+     * the BitArray, i.e. {@code 8*(a.length - ofs) >= length}.
      */
     public BitArray(int length, byte[] a, int ofs)
             throws IllegalArgumentException {
@@ -177,16 +178,12 @@ public class BitArray {
         return repn.clone();
     }
 
+    @Override
     public boolean equals(Object obj) {
         if (obj == this) return true;
-        if (!(obj instanceof BitArray ba)) return false;
-
-        if (ba.length != length) return false;
-
-        for (int i = 0; i < repn.length; i += 1) {
-            if (repn[i] != ba.repn[i]) return false;
-        }
-        return true;
+        if (!(obj instanceof BitArray other)) return false;
+        if (length != other.length) return false;
+        return Arrays.equals(repn, other.repn);
     }
 
     /**
@@ -202,17 +199,11 @@ public class BitArray {
     }
 
     /**
-     * Returns a hash code value for this bit array.
-     *
-     * @return  a hash code value for this bit array.
+     * {@return a hash code value for this bit array}
      */
+    @Override
     public int hashCode() {
-        int hashCode = 0;
-
-        for (int i = 0; i < repn.length; i++)
-            hashCode = 31*hashCode + repn[i];
-
-        return hashCode ^ length;
+        return Arrays.hashCode(repn) ^ length;
     }
 
 
--- a/src/java.base/share/classes/java/security/CodeSigner.java
+++ b/src/java.base/share/classes/java/security/CodeSigner.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2003, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2003, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -27,6 +27,7 @@ package java.security;
 
 import java.io.*;
 import java.security.cert.CertPath;
+import java.util.Objects;
 
 /**
  * This class encapsulates information about a code signer.
@@ -98,12 +99,11 @@ public final class CodeSigner implements Serializable {
     }
 
     /**
-     * Returns the hash code value for this code signer.
+     * {@return the hash code value for this code signer}
      * The hash code is generated using the signer's certificate path and the
      * timestamp, if present.
-     *
-     * @return a hash code value for this code signer.
      */
+    @Override
     public int hashCode() {
         if (myhash == -1) {
             if (timestamp == null) {
@@ -126,24 +126,18 @@ public final class CodeSigner implements Serializable {
      * @return {@code true} if the objects are considered equal,
      * {@code false} otherwise.
      */
+    @Override
     public boolean equals(Object obj) {
-        if ((!(obj instanceof CodeSigner that))) {
+        if (this == obj) {
+            return true;
+        }
+
+        if (!(obj instanceof CodeSigner that)) {
             return false;
         }
 
-        if (this == that) {
-            return true;
-        }
-        Timestamp thatTimestamp = that.getTimestamp();
-        if (timestamp == null) {
-            if (thatTimestamp != null) {
-                return false;
-            }
-        } else {
-            if ((!timestamp.equals(thatTimestamp))) {
-                return false;
-            }
-        }
+        if (!Objects.equals(timestamp, that.getTimestamp()))
+            return false;
         return signerCertPath.equals(that.getSignerCertPath());
     }
 
--- a/src/java.base/share/classes/java/security/UnresolvedPermission.java
+++ b/src/java.base/share/classes/java/security/UnresolvedPermission.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1997, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -35,6 +35,7 @@ import java.util.Hashtable;
 import java.lang.reflect.*;
 import java.security.cert.*;
 import java.util.List;
+import java.util.Objects;
 
 /**
  * The {@code UnresolvedPermission} class is used to hold Permissions that
@@ -349,23 +350,13 @@ implements java.io.Serializable
         }
 
         // check name
-        if (this.name == null) {
-            if (that.name != null) {
-                return false;
-            }
-        } else if (!this.name.equals(that.name)) {
+        if (!Objects.equals(this.name, that.name)) {
             return false;
         }
 
         // check actions
-        if (this.actions == null) {
-            if (that.actions != null) {
-                return false;
-            }
-        } else {
-            if (!this.actions.equals(that.actions)) {
-                return false;
-            }
+        if (!Objects.equals(this.actions, that.actions)) {
+            return false;
         }
 
         // check certs
@@ -404,18 +395,11 @@ implements java.io.Serializable
     }
 
     /**
-     * Returns the hash code value for this object.
-     *
-     * @return a hash code value for this object.
+     * {@return the hash code value for this object}
      */
     @Override
     public int hashCode() {
-        int hash = type.hashCode();
-        if (name != null)
-            hash ^= name.hashCode();
-        if (actions != null)
-            hash ^= actions.hashCode();
-        return hash;
+        return Objects.hash(type, name, actions);
     }
 
     /**
--- a/src/java.base/share/classes/javax/security/auth/Subject.java
+++ b/src/java.base/share/classes/javax/security/auth/Subject.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -907,10 +907,6 @@ public final class Subject implements java.io.Serializable {
     @Override
     public boolean equals(Object o) {
 
-        if (o == null) {
-            return false;
-        }
-
         if (this == o) {
             return true;
         }
@@ -1003,9 +999,7 @@ public final class Subject implements java.io.Serializable {
     }
 
     /**
-     * Returns a hashcode for this {@code Subject}.
-     *
-     * @return a hashcode for this {@code Subject}.
+     * {@return a hashcode for this {@code Subject}}
      *
      * @throws SecurityException if a security manager is installed and the
      *         caller does not have a {@link PrivateCredentialPermission}


src/java.base/share/classes/javax/crypto/CryptoPolicyParser.java
```diff
@@ -616,33 +617,27 @@ private static class CryptoPermissionEntry {
...
             if (!(obj instanceof CryptoPermissionEntry that))
                 return false;
 
-            if (this.cryptoPermission == null) {
-                if (that.cryptoPermission != null) return false;
-            } else {
-                if (!this.cryptoPermission.equals(
-                                                 that.cryptoPermission))
-                    return false;
+            if (Objects.equals(this.cryptoPermission, that.cryptoPermission)) {
```
djelinski, line 639:
```suggestion
            if (!Objects.equals(this.cryptoPermission, that.cryptoPermission)) {
```

src/java.base/share/classes/sun/security/provider/PolicyParser.java
```diff
@@ -1119,43 +1119,18 @@ public PermissionEntry(String permission,
...
-
-            if (this.signedBy == null) {
-                return that.signedBy == null;
-            } else {
-                return this.signedBy.equals(that.signedBy);
-            }
+            return obj instanceof PermissionEntry that
+                    && Objects.equals(this.permission, that.permission)
+                    && Objects.equals(this.name, that.name)
+                    && Objects.equals(this.signedBy, that.signedBy);
```
djelinski, line 1133: The original code also compared `action`

src/java.base/share/classes/java/security/spec/ECFieldF2m.java
```diff
@@ -226,12 +227,13 @@ public boolean equals(Object obj) {
...
     /**
-     * Returns a hash code value for this characteristic 2
-     * finite field.
-     * @return a hash code value.
+     * {@return the hash code value for this characteristic 2 finite field}
      */
+    @Override
     public int hashCode() {
         int value = m << 5;
+        // consider simplifying using Objects.hashCode(rp) after JDK-8015417
```
djelinski, line 235: Could you explain this comment? Why does it apply here, and not to other methods like CodeSigner.hashCode?

pavelrappo: You are absolutely right: adding that comment, while actively disregarding concerns described in it elsewhere, warrants some explanation.

While working on this and the related PRs, I discovered JDK-8015417. hashCode and equals in ECFieldF2m seemed lean and performance-sensitive so as to apply the smarts of JDK-8015417. Frankly, performance sensitivity was my guess; anyway, I kept the original behaviour for that site.

Since then, I learned that JDK-8015417 is quite complicated and well beyond my level of understanding of how JVM works. Until I have clearer understanding of the effects of that issue on this and similar refactoring efforts, I won't integrate this or any other PRs where using the java.util.Objects static methods is a primary goal.

That said, I'll remove the comment and refactor that `?:` conditional to use `Objects.equals(Object, Object)`, to be consistent with the rest of this PR.

FWIW, we've accidentally started discussion on those effects in another PR: https://github.com/openjdk/jdk/pull/14752#pullrequestreview-1511190453. It might as well have been this PR, but that PR was first. While either PR is good, let's keep the it confined and organized. So, please follow that discussion, if you're interested.

pavelrappo: Refactored in 3f316b49f26.

djelinski (approved):

src/java.base/share/classes/javax/security/cert/Certificate.java
```diff
@@ -81,42 +82,30 @@ public Certificate() {}
...
     public int hashCode() {
-        int     retval = 0;
         try {
-            byte[] certData = this.getEncoded();
-            for (int i = 1; i < certData.length; i++) {
-                 retval += certData[i] * i;
-            }
-            return (retval);
+            return Arrays.hashCode(this.getEncoded());
         } catch (CertificateException e) {
```
RogerRiggs, line 107: Ditto, what value at `certData[0]` is now included in the hash?

src/java.base/share/classes/sun/security/x509/X500Name.java
```diff
@@ -416,17 +418,9 @@ public boolean equals(Object obj) {
             return this.canonicalDn.equals(other.canonicalDn);
         }
         // quick check that number of RDNs and AVAs match before canonicalizing
-        int n = this.names.length;
-        if (n != other.names.length) {
+        if (!Arrays.equals(this.names, other.names,
+                Comparator.comparingInt(n -> n.assertion.length)))
```
RogerRiggs, line 422: I'd keep the original comparison of the lengths; its a lot less magical than `Comparator.comparingInt(n -> n.assertion.length))`

src/java.base/share/classes/sun/security/util/BitArray.java
```diff
@@ -68,7 +69,7 @@ public BitArray(int length) throws IllegalArgumentException {
      * Creates a BitArray of the specified size, initialized from the
      * specified byte array. The most significant bit of {@code a[0]} gets
      * index zero in the BitArray. The array must be large enough to specify
-     * a value for every bit of the BitArray. i.e. {@code 8*a.length <= length}.
+     * a value for every bit of the BitArray, i.e. {@code 8*a.length >= length}.
```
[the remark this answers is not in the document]
djelinski: Are you sure? I just checked lines 91-92 and I'd say the change looks correct.

src/java.base/share/classes/javax/security/cert/Certificate.java
pavelrappo: Those are correct observations. Unspecified hashCode methods changed in this proposal might return different values than the ones they currently return in the mainline; hence my disclaimer in this PR description:

> This PR sacrifices compatibility for pragmatism: it changes some hashCode implementations to produce different values than before to allow more utilization of methods from Objects and Arrays. To my mind, those changes are **benign**. If you disagree, I'd be happy to discuss that and/or retract the concerning part of the change.

I'm waiting from someone from security-dev to tell me if that's okay to skip zeroth elements in both hashCode. It might've been accidental to skip it in the first place, or not. FWIW, their equals counterparts do not skip zeroth elements. On the other hand, javax.security.cert.Certificate is deprecated for removal, so I might as well exclude it.

src/java.base/share/classes/sun/security/util/BitArray.java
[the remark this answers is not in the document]
pavelrappo: Hm... My reading is that those "i.e." parts state preconditions for the constructors to return successfully, not preconditions for them to throw an exception.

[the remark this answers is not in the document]
pavelrappo: > The original `<=` was correct, the number of bits in the input array must be less than the requested length of the BitArray. The constructors also describe the length using `<=`; they all should be consistent.

Hm... My reading is that those "i.e." parts state preconditions for the constructors to return successfully, not preconditions for them to throw an exception.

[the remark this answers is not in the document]
RogerRiggs: I stand corrected.

src/java.base/share/classes/java/security/BasicPermission.java
```diff
@@ -217,12 +217,10 @@ public boolean equals(Object obj) {
 
 
     /**
-     * Returns the hash code value for this object.
+     * {@return the hash code value for this object}
```
seanjmullan, line 220: Where is this new javadoc syntax documented/specified? I recall seeing something about it, but not sure where and now I would like to read up on it.

pavelrappo: https://docs.oracle.com/en/java/javase/20/docs/specs/javadoc/doc-comment-spec.html#return

src/java.base/share/classes/java/security/cert/CertPath.java
```diff
@@ -177,18 +177,19 @@ public String getType() {
...
         return other instanceof CertPath that
-                && that.getType().equals(this.type)
+                && this.type.equals(that.getType())
                 && this.getCertificates().equals(that.getCertificates());
     }
 
     /**
-     * Returns the hashcode for this certification path. The hash code of
-     * a certification path is defined to be the result of the following
+     * {@return the hashcode value for this certification path} The hash code
```
seanjmullan, line 191: This looks a bit odd w/o a period at the end of the `}`. Is it optional? [The javadoc spec](https://docs.oracle.com/en/java/javase/20/docs/specs/javadoc/doc-comment-spec.html#return) isn't very clear on that - perhaps you could clarify that in the spec.

If it is optional, my general preference would be to include the period, as it makes the code more readable.

jonathan-gibbons: While I agree it looks "a bit odd", the problem is having to address the two ways in which the tag is rendered:

1. As the first sentence. In this case, the text in the braces is preceded by `Returns` and followed by a period.
2. As the equivalent for an `@return` tag, where `Returns:` is a bold heading, and the content is a phrase, and (generally) not followed by a period.

The current guideline is that the period should not be present. Elsewhere, the "oddness" is reduced by putting
a newline after the tag, so that the `{@return}` tag stands on a line by itself.  To forestall any possible suggestion,
I would not be in favor of looking outside the tag to see if there is a period following the tag.

We can clarify the spec.

If you are uncomfortable with the usage here, it is always possible to avoid using the short inline form, and to "duplicate" the text and use an old-fashioned plain first sentence, and an old-fashioned `@return` "block" tag

src/java.base/share/classes/sun/security/x509/X500Name.java
seanjmullan: I tend to agree with Roger. I have to think more about what is being compared with the new code.

src/java.base/share/classes/java/security/cert/X509CRLEntry.java
```diff
@@ -92,34 +94,23 @@ public boolean equals(Object other) {
...
      */
+    @Override
     public int hashCode() {
-        int     retval = 0;
         try {
-            byte[] entryData = this.getEncoded();
-            for (int i = 1; i < entryData.length; i++)
-                 retval += entryData[i] * i;
-
+            return Arrays.hashCode(this.getEncoded());
```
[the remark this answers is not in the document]
seanjmullan: I would have to dump out an encoded CRL to check, but it probably doesn't matter, as long as 2 different CRLs will always have different encodings.

src/java.base/share/classes/sun/security/x509/X500Name.java
pavelrappo: Yeah, I assume no matter how one shapes it, it doesn't seem a readability win:

	if (!Arrays.equals(this.names, other.names,
	        (n1, n2) -> Integer.compare(n1.assertion.length, n2.assertion.length))) {
	    return false;
	}

Reverted in 42d9543b769.

RogerRiggs (approved):

seanjmullan:
Took another pass at this, looks good, but I would like to take another last look and make sure that changing the hash code for some of the classes like X509CRL is a benign change.

pavelrappo:
> Took another pass at this, looks good, but I would like to take another last look and make sure that changing the hash code for some of the classes like X509CRL is a benign change.

Thanks, Sean. Take your time, you're an expert in this area. Meanwhile, I'll reflow other similar `{@return ... }` constructs that I missed before, for readability.

pavelrappo:
> > Took another pass at this, looks good, but I would like to take another last look and make sure that changing the hash code for some of the classes like X509CRL is a benign change.
> 
> Thanks, Sean. Take your time, you're an expert in this area. Meanwhile, I'll reflow other similar `{@return ... }` constructs that I missed before, for readability.

This comment is NOT to rush the review.

I once again note, that those classes in this PR that skip the first array element in `hashCode`, do not seem to skip that same element in `equals`. Although that does not breach equals-hashCode contract, it puzzles the reader and theoretically makes `hashCode` more blunt (i.e. more instances may share a hash code value).

This observation is an investigation opportunity. But equally, I'm okay with skipping the investigation and reverting those particular changes, if security-dev wants to be on the safe side. We might soon have a more flexible way to compute hashCode [^*]. If we have it, we could then both refactor those hashCode implementations and preserve their behaviour.

[^*]: https://github.com/openjdk/jdk/pull/14831

pavelrappo:
> > > Took another pass at this, looks good, but I would like to take another last look and make sure that changing the hash code for some of the classes like X509CRL is a benign change.
> > 
> > 
> > Thanks, Sean. Take your time, you're an expert in this area. Meanwhile, I'll reflow other similar `{@return ... }` constructs that I missed before, for readability.
> 
> This comment is NOT to rush the review.
> 
> I once again note, that those classes in this PR that skip the first array element in `hashCode`, do not seem to skip that same element in `equals`. Although that does not breach equals-hashCode contract, it puzzles the reader and theoretically makes `hashCode` more blunt (i.e. more instances may share a hash code value).
> 
> This observation is an investigation opportunity. But equally, I'm okay with skipping the investigation and reverting those particular changes, if security-dev wants to be on the safe side. We might soon have a more flexible way to compute hashCode [1](#user-content-fn-*-d08640fefec8cc5b91d743888afc0a5d). If we have it, we could then both refactor those hashCode implementations and preserve their behaviour.
> 
> ## Footnotes
> 1. https://github.com/openjdk/jdk/pull/14831 [↩](#user-content-fnref-*-d08640fefec8cc5b91d743888afc0a5d)

To add to my previous comment. Don't know if it helps, but I found this bug, that applies similar (to my non-expert eye) refactoring to java.security.cert.Certificate.hashCode: 

8011402: Move blacklisting certificate logic from hard code to data

pavelrappo:
I pushed a new commit, 457b9c5, that changes almost 90 files; I realise it's huge review cost. Perhaps at this stage, reviewers could glance at that commit, and we could collectively then cherry-pick the most useful changes.

FWIW, the tests are green.

src/java.base/share/classes/javax/crypto/CryptoPolicyParser.java
```diff
@@ -616,33 +617,27 @@ private static class CryptoPermissionEntry {
...
-            }
+            retval ^= Objects.hashCode(alg);
+            retval ^= Objects.hashCode(exemptionMechanism);
             retval ^= maxKeySize;
-            if (checkParam) retval ^= 100;
-            if (algParamSpec != null) {
-                retval ^= algParamSpec.hashCode();
-            }
+            retval ^= (checkParam ? 100 : 0);
+            retval ^= Objects.hashCode(algParamSpec);
```
RogerRiggs, line 627: After this change, someone will wonder why the code does all the intermediate assignments.

```
    return Objects.hashCode(alg)
        ^ Objects.hashCode(exemptionMechanism) 
        ^ (checkParam ? 100 : 0)  
        ^ Objects.hashCode(algParamSpec);
}

Here and elsewhere...

File: src/java.base/share/classes/sun/security/x509/DistributionPoint.java [hunk 11]
```diff
@@ -339,26 +340,15 @@ public boolean equals(Object obj) {
...
-                for (int i = 0; i < reasonFlags.length; i++) {
-                    if (reasonFlags[i]) {
-                        hash += i;
-                    }
-                }
-            }
+            hash += Objects.hashCode(fullName);
+            hash += Objects.hashCode(relativeName);
+            hash += Objects.hash(crlIssuer);
+            hash += Arrays.hashCode(reasonFlags);
```
RogerRiggs, line 351: Another case where the intermediate assignments look unnecessary.

RogerRiggs (approved):
Look good, lots of nice cleanup.

pavelrappo:
> I pushed a new commit, [457b9c5](https://github.com/openjdk/jdk/commit/457b9c56a937b98d64201ce378f6960f769dcf65)
> 
> FWIW, the tests are green.

Our CI disagrees with me: testing at 457b9c56a93 revealed three failures. Apologies for misinforming my reviewers.

At 3caa774d007 all three failures are fixed, more previously missed use cases are addressed, [feedback](https://github.com/openjdk/jdk/pull/14738#discussion_r1262696059) from @RogerRiggs is incorporated, and the tests are run. This time, the tests are completely green, not just mostly green; I double-checked that.

A few comments on new commits.

* 6d339fce adds a must-have null-check to equals here (https://github.com/openjdk/jdk/blob/3caa774d007d6a133fd1238f3c4459b1a15d6050/src/java.base/share/classes/com/sun/crypto/provider/PBKDF2KeyImpl.java#L167). I believe there's no compatibility risk.

* f83c1578f13 fixes a peculiar failure related to (JFR?) security events.

  PKIXCertPathValidator generates an event that TestX509ValidationLog checks. While the actual and the expected formats have mismatched for some time, the bug reveals itself only now because the new hashCode implementation returns a negative int value, which is not expected.

  I don't know how big can an X.509 key be, but in TestX509ValidationLog it's 294 bytes. That corresponds to a maximum possible old hash code value of 1,392,678, which is a positive int. Now, if my figures are correct, it would require a key of a minimum size of 453,343 (443 kB) for the old hashCode implementation to overflow and return a negative int. And even then, the overflow will only happen if the key consists of all `0xff` bytes, which probably is a bad key. So, in reality, the key would need to be bigger than that for the old hash code to overflow. Hence, the failure hasn't been previously observed.

  With the new hashCode implementation, it takes as few as 1 byte to overflow. Because I think that it's unreasonable to require hashCode to not be negative, I fix the test, not code.

RogerRiggs (approved):
Thanks for the thorough checking and testing.

src/java.base/share/classes/sun/security/x509/OtherName.java
```diff
@@ -203,16 +204,12 @@ public boolean equals(Object other) {
...
+     * {@return the hash code for this OtherName}
      */
+    @Override
     public int hashCode() {
         if (myhash == -1) {
-            myhash = 37 + oid.hashCode();
-            for (int i = 0; i < nameValue.length; i++) {
-                myhash = 37 * myhash + nameValue[i];
-            }
+            myhash = 37 + oid.hashCode() + Arrays.hashCode(nameValue);
```
djelinski, line 212: I think you can drop 37 here. Either that, or use `37 * oid.hashCode()`.

djelinski (approved):

src/java.base/share/classes/sun/security/x509/OtherName.java
pavelrappo: You think it was a typo (+ instead of *) in the original code?

djelinski: The original code calculated `37*(37+oid.hashCode)+Arrays.hashCode`; since you already dropped the multiplication by 37, you can also drop the addition.

pavelrappo: Thanks for noticing this difference!

Yes, the initial value would've been exponentiated (^ or ** in your notation). However, I note that the original code computed this:

if nameValue.length > 0

  (37^nameValue.length) * oid.hashCode() + 37 + Arrays.hashCode(nameValue)

otherwise

  37 + oid.hashCode()

Well, it would've computed that, had the multiplier constant been 37 in Arrays.hashCode, which it is not; it's 31.

Since we cannot achieve absolute fidelity with the old code, I suggest we do this instead, what do you think?

        public int hashCode() {
        if (myhash == -1) {
            myhash = Arrays.deepHashCode(new Object[]{oid, nameValue});
        }
        return myhash;

pavelrappo: Correction. If the multiplier were 31, then it would've been this:

    return ( 30 + oid.hashCode() ) * 31**nameValue.length + Arrays.hashCode(nameValue);
              ^                       ^

@djelinski, do you think we should use this instead?

    return Arrays.deepHashCode(new Object[]{oid, nameValue});

djelinski: I think `oid.hashCode() + Arrays.hashCode(nameValue)` would have been good enough, but deepHashCode is also acceptable.

pavelrappo: > I think `oid.hashCode() + Arrays.hashCode(nameValue)` would have been good enough, but deepHashCode is also acceptable.

Fixed as you suggested, in 3139043370f.

src/java.base/share/classes/sun/security/x509/URIName.java
```diff
@@ -275,10 +275,9 @@ public Object getHostObject() {
     }
 
     /**
-     * Returns the hash code value for this object.
-     *
-     * @return a hash code value for this object.
+     * {@return the hash code value for this object}
      */
+    @Override
```
seanjmullan, line 280: You could also add `@Override` to `equals`.

src/java.base/share/classes/java/security/AccessControlContext.java
seanjmullan, line 1: Missing `@Override` on equals.

src/java.base/share/classes/sun/security/x509/PolicyInformation.java
```diff
@@ -122,28 +122,22 @@ public PolicyInformation(DerValue val) throws IOException {
...
-     *
-     * @return a hash code value.
+     * {@return the hash code for this PolicyInformation}
      */
+    @Override
     public int hashCode() {
-        int myhash = 37 + policyIdentifier.hashCode();
-        myhash = 37 * myhash + policyQualifiers.hashCode();
-        return myhash;
+        return Objects.hash(policyIdentifier, policyQualifiers);
```
valeriepeng, line 140: What is the criteria for changing the hashCode() impl? It seems that you change some and not others. With the many hashCode() impls (e.g. Objects.hash(), Arrays.hashCode()), I wonder why this and not that.

src/java.base/share/classes/com/sun/crypto/provider/PBKDF2KeyImpl.java
```diff
@@ -247,33 +247,29 @@ public String getFormat() {
...
 
-            if (!(obj instanceof SecretKey)) {
+            if (!(obj instanceof SecretKey that)) {
                 return false;
             }
 
-            SecretKey that = (SecretKey) obj;
-
-            if (!(that.getAlgorithm().equalsIgnoreCase(getAlgorithm()))) {
+            if (!(that.getAlgorithm().equalsIgnoreCase(getAlgorithm()))){
```
valeriepeng, line 272: why removing the space before the "{"?

src/java.base/share/classes/sun/security/provider/certpath/CertId.java
```diff
@@ -172,19 +172,13 @@ public void encode(DerOutputStream out) {
...
             myhash = hashAlgId.hashCode();
-            for (int i = 0; i < issuerNameHash.length; i++) {
-                myhash += issuerNameHash[i] * i;
-            }
-            for (int i = 0; i < issuerKeyHash.length; i++) {
-                myhash += issuerKeyHash[i] * i;
-            }
+            myhash += Arrays.hashCode(issuerNameHash);
+            myhash += Arrays.hashCode(issuerKeyHash);
             myhash += certSerialNumber.getNumber().hashCode();
```
valeriepeng, line 182: It seems that for some other classes, you may just use Objects.hash(hashAlgId, issuerNameHash, issuerKeyHash, certSerialNumber.getNumber())? Is there a rule on what to use?

src/java.base/share/classes/com/sun/crypto/provider/PBKDF2KeyImpl.java
pavelrappo: It surprises me as much as it probably surprises you. Looks like a sloppy merge (821f617ffae) on my part.

pavelrappo: I retraced my steps to see how the merge introduced that artefact. This might be useful to others too.

The original merge was done via an IDE, which has this (usually) helpful 1-click feature called "Resolve Simple Conflicts". It's that feature that ate that whitespace character. To its credit, the rest of the resolution seems to have been done correctly. When I examined the original merge before pushing it to the PR, I didn't spot that whitespace; my bad. I'll reintroduce it.

Thanks for your careful review.

src/java.base/share/classes/sun/security/x509/PolicyInformation.java
pavelrappo: Whether I suggest changing hashCode or not depends on multiple criteria, of which these two are the most important:

  * Is the value specified or relied upon?
  * Does it need to be performant?

If the answer to both of these is "NO", I might refactor hashCode. One unintended side effect, is that sometimes such a refactoring may improve hashCode quality (think: perfect hashing) almost for free.

If you have concerns on particular refactorings, I'd be happy to discuss them.

src/java.base/share/classes/com/sun/crypto/provider/DESKey.java
```diff
@@ -108,26 +109,24 @@ public String getFormat() {
      * Calculates a hash code value for the object.
      * Objects that are equal will also have the same hashcode.
      */
+    @Override
     public int hashCode() {
-        int retval = 0;
-        for (int i = 1; i < this.key.length; i++) {
```
seanjmullan, line 113: I don't really know why the hashCode impl skipped the first byte of the array. Source code history doesn't give any insight. I thought maybe this had something to do with the parity bits (8 bits of a DES key are used for parity) but those are every 8th bit, and not all in the first byte. Best I can guess is that this is just an oversight and it should have included all bytes in the hash code calculation.

src/java.base/share/classes/sun/security/x509/PolicyInformation.java
seanjmullan: I would add to that list that I think it is really important any change to the `hashCode` impl doesn't break the equals/hashCode contract, so it's important to look at the `equals` implementation in conjunction with any changes to `hashCode`, mostly to ensure 2 equal objects continue to have the same hash code. It is also important to question any oddity about the `hashCode` impl, like I did for DESKey and try to understand why it may have been implemented that way, such as skipping the first byte.

pavelrappo: True. I skipped that criteria, as I thought it should go without saying.

pavelrappo: > I would add to that list that I think it is really important any change to the `hashCode` impl doesn't break the equals/hashCode contract, so it's important to look at the `equals` implementation in conjunction with any changes to `hashCode`, mostly to ensure 2 equal objects continue to have the same hash code.

True. I skipped that criteria, as I thought it should go without saying.

> It is also important to question any oddity about the `hashCode` impl, like I did for DESKey and try to understand why it may have been implemented that way, such as skipping the first byte.

True again. If you have any other examples where the changes to hashCode look suspicious, please tell me so.

src/java.base/share/classes/com/sun/crypto/provider/DESKey.java
pavelrappo: Thanks for checking.

src/java.security.jgss/share/classes/sun/security/krb5/internal/HostAddress.java
```diff
@@ -78,41 +78,31 @@ public Object clone() {
...
             (address != null && h.address == null) ||
             (address == null && h.address != null))
             return false;
         if (address != null && h.address != null) {
-            if (address.length != h.address.length)
-                return false;
-            for (int i = 0; i < address.length; i++)
-                if (address[i] != h.address[i])
-                    return false;
+            return Arrays.equals(address, h.address);
```
valeriepeng, line 105: Can't this block be combined into:
return` addrType == h.addrType && Arrays.equals(address, h.address); `?

src/java.security.jgss/share/classes/sun/security/krb5/internal/HostAddresses.java
```diff
@@ -130,40 +130,30 @@ public boolean inList(HostAddress addr) {
...
         if ((addresses == null && addrs.addresses != null) ||
             (addresses != null && addrs.addresses == null))
             return false;
         if (addresses != null && addrs.addresses != null) {
-            if (addresses.length != addrs.addresses.length)
-                return false;
-            for (int i = 0; i < addresses.length; i++)
-                if (!addresses[i].equals(addrs.addresses[i]))
-                    return false;
+            return Arrays.equals(addresses, addrs.addresses);
```
valeriepeng, line 156: Same comment as above, i.e. combined into one return condition.

src/java.security.jgss/share/classes/sun/security/krb5/internal/KRBError.java
```diff
@@ -488,39 +490,34 @@ public byte[] asn1Encode() throws Asn1Exception, IOException {
...
         result = 37 * result + msgType;
+        result = 37 * result + errorCode;
         if (cTime != null) result = 37 * result + cTime.hashCode();
         if (cuSec != null) result = 37 * result + cuSec.hashCode();
         if (sTime != null) result = 37 * result + sTime.hashCode();
         if (suSec != null) result = 37 * result + suSec.hashCode();
-        result = 37 * result + errorCode;
         if (crealm != null) result = 37 * result + crealm.hashCode();
         if (cname != null) result = 37 * result + cname.hashCode();
         if (sname != null) result = 37 * result + sname.hashCode();
```
valeriepeng, line 523: Why changing the ordering? I was wondering if you'd change this to Objects.hash(...) with all the fields as arguments.

pavelrappo: > Why changing the ordering?

The order was changed to be parallel with that of `equals`.

> I was wondering if you'd change this to Objects.hash(...) with all the fields as arguments.

Sure, we could use a convenience method, but because `eData` is an array, it should be `Arrays.deepHashCode(..., eData, ...)` or this construct:

    Objects.hash(..., Arrays.hashCode(eData), ...)

Bottom line: as long as we include all eData elements in the computation, we should be fine. If we fail to do that, we'll violate the equals-hashCode contract.

src/java.base/share/classes/sun/security/provider/certpath/CertId.java
valeriepeng: Hmm, I see. Thanks for the clarification.

src/java.security.jgss/share/classes/sun/security/krb5/internal/KRBError.java
valeriepeng: Ok.

src/java.security.jgss/share/classes/sun/security/krb5/internal/HostAddress.java
pavelrappo: Have a look at the end result (commits e71fc42a173 and d9316270976), is it better?

src/java.security.jgss/share/classes/sun/security/krb5/internal/HostAddresses.java
pavelrappo: Same comment as above.

src/java.security.jgss/share/classes/sun/security/krb5/internal/HostAddress.java
pavelrappo: Oops: forgot to commit and push 608258e.

RogerRiggs (approved):

valeriepeng (approved):
Looks good. Thanks

pavelrappo:
Thanks to my patient reviewers. I'm running our CI tests one more time and, if all goes well, I'll integrate this PR.

PR #14739, opened 2023-06-30
JDK-8310233: Fix THP detection on Linux

tstuefe:
Today, if we use UseTransparentHugePages, we assume that the *static* hugepage detection we do is valid for THPs:
- that THPs use the page size (in hotspot used as "default large page size") found in `/proc/memlimit` "Hugepagesize")
- that THPs are enabled if that page size is >0.

Both assumptions are incorrect:

- whether THPs are enabled should be checked at `/sys/kernel/mm/transparent_hugepage/enabled`, which is a tri-state value ("always", "madvise", "never"). THPs are available for the first two states.
- The page size employed by `khugepaged` is set in `/sys/kernel/mm/transparent_hugepage/hpage_pmd_size`. It can differ from the default page size used for static hugepages. For example, we could configure a system such that it uses 1G static hugepages, but the THP page size would still be 2M.

------

About the patch:

This is a limited, minimally invasive patch to fix THP detection. The patch aims to be easy to downport. There is more work to do, which I will do in subsequent RFEs.

Functionally, for *static* (non-THP) pages nothing changes. THP-mode now correctly detects THP support in the OS, and uses the correct page size (see examples below).

-------------

Example 1: System has THPs disabled, but static hugepages (1g, 2m) configured:

```
thomas@starfish $ cat /sys/kernel/mm/transparent_hugepage/enabled
always madvise [never]
thomas@starfish $ cat /proc/meminfo | grep Hugepage
Hugepagesize:    1048576 kB
```

Without patch, we incorrectly assume THPs are enabled, and that THP page size is 1G (!), which we then proceed and use as heap page size, causing the heap size to be rounded up from 512m -> 1G. But - even though it is printed as "1G page backed" in log output - in reality it will still 4K-page-backed: the `madvise(2)` we use to set the THP page size will be ignored by the system, since THPs are disabled.

```
thomas@starfish $ ./images/jdk/bin/java -Xmx512m -XX:+UseLargePages -XX:+UseTransparentHugePages -Xlog:pagesize -version
[0.001s][info][pagesize] Using the default large page size: 1G
[0.001s][info][pagesize] Usable page sizes: 4k, 2M, 1G
...
[0.016s][info][pagesize] Heap:  min=1G max=1G base=0x00000000c0000000 size=1G page_size=1G
```

With patch, we correctly refuse to use large pages (and we log more info):

```
thomas@starfish $ ./images/jdk/bin/java -Xmx512m -XX:+UseLargePages -XX:+UseTransparentHugePages -Xlog:pagesize -version
[0.001s][info][pagesize] Static hugepage support: 2M, 1G (default)
[0.001s][info][pagesize]   default pagesize: 1G
[0.001s][info][pagesize] Transparent hugepage (THP) support:
[0.001s][info][pagesize]   mode: never
[0.001s][warning][pagesize] UseLargePages disabled, no large pages configured and available on the system.
```

----------

Example 2: System has THPs enabled, but THP page size is just *2M*, whereas the system uses a static default hugepage size of *1G*:

```
thomas@starfish $ cat /sys/kernel/mm/transparent_hugepage/enabled
always [madvise] never
thomas@starfish $ cat /sys/kernel/mm/transparent_hugepage/hpage_pmd_size 
2097152
thomas@starfish $ cat /proc/meminfo | grep Hugepage
Hugepagesize:    1048576 kB
```

Without patch, THP page size is not correctly recognized as 2M. Instead, we again use 1G as page size for the heap:

```
thomas@starfish $ ./images/jdk/bin/java -Xmx512m -XX:+UseLargePages -XX:+UseTransparentHugePages -Xlog:pagesize -version
[0.001s][info][pagesize] Using the default large page size: 1G
[0.001s][info][pagesize] Usable page sizes: 4k, 2M, 1G
...
[0.010s][info][pagesize] Heap:  min=1G max=1G base=0x00000000c0000000 size=1G page_size=1G
```

With patch, we correctly identify the THP page size as 2M, and use that for the heap:

```
thomas@starfish $ ./images/jdk/bin/java -Xmx512m -XX:+UseLargePages -XX:+UseTransparentHugePages -Xlog:pagesize -version
[0.001s][info][pagesize] Static hugepage support: 2M, 1G (default)
[0.001s][info][pagesize]   default pagesize: 1G
[0.001s][info][pagesize] Transparent hugepage (THP) support:
[0.001s][info][pagesize]   mode: madvise
[0.001s][info][pagesize]   pagesize: 2M
[0.001s][info][pagesize] Large page support enabled. Usable page sizes: 4k, 2M
[0.001s][info][pagesize]  Default: 2M
...
[0.010s][info][pagesize] Heap:  min=8M max=512M base=0x00000000e0000000 size=512M page_size=2M
```
------------

Tests: GHAs all green. Local experiments on x64 Linux on machines with 1G pages succeeded.

---------
### Progress

### Issue
 * [JDK-8310233](https://bugs.openjdk.org/browse/JDK-8310233): Fix THP detection on Linux (**Bug** - P4)

### Reviewers
 * [Johan Sjölen](https://openjdk.org/census#jsjolen) (@jdksjolen - **Reviewer**) ⚠️ Review applies to [69c39c8a](https://git.openjdk.org/jdk/pull/14739/files/69c39c8aa4411bfd468cfc63faa7559728a10645)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14739/head:pull/14739` \
`$ git checkout pull/14739`

`$ git checkout pull/14739` \
`$ git pull https://git.openjdk.org/jdk.git pull/14739/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14739`

`$ git pr show -t 14739`

</details>

<a href="https://git.openjdk.org/jdk/pull/14739.diff">https://git.openjdk.org/jdk/pull/14739.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14739#issuecomment-1616059461)

### DIFF
--- a/src/hotspot/os/linux/hugepages.cpp
+++ b/src/hotspot/os/linux/hugepages.cpp
@@ -88,9 +88,9 @@ static size_t scan_default_hugepagesize() {
   return pagesize;
 }
 
-// Given a file that contains a single (integral) number, return that number, 0 and false if failed.
+// Given a file that contains a single (integral) number, return that number in (*out) and true;
+// in case of an error, return false.
 static bool read_number_file(const char* file, size_t* out) {
-  (*out) = 0;
   FILE* f = ::fopen(file, "r");
   bool rc = false;
   if (f != nullptr) {
@@ -191,8 +191,9 @@ void THPSupport::scan_os() {
 
   // Scan large page size for THP from hpage_pmd_size
   _pagesize = 0;
-  read_number_file("/sys/kernel/mm/transparent_hugepage/hpage_pmd_size", &_pagesize);
-  assert(_pagesize > 0, "Expected");
+  if (read_number_file("/sys/kernel/mm/transparent_hugepage/hpage_pmd_size", &_pagesize)) {
+    assert(_pagesize > 0, "Expected");
+  }
   _initialized = true;
 
   LogTarget(Info, pagesize) lt;
--- a/src/hotspot/os/linux/hugepages.cpp
+++ b/src/hotspot/os/linux/hugepages.cpp
@@ -131,16 +131,11 @@ static os::PageSizes scan_hugepages() {
 
 void StaticHugePageSupport::print_on(outputStream* os) {
   if (_initialized) {
-    os->print("Static hugepage support: ");
-    bool first = true;
+    os->print_cr("Static hugepage support:");
     for (size_t s = _pagesizes.smallest(); s != 0; s = _pagesizes.next_larger(s)) {
-      os->print("%s" EXACTFMT, (first ? "" : ", "), EXACTFMTARGS(s));
-      if (_default_hugepage_size == s) {
-        os->print_cr(" (default)");
-      }
-      first = false;
+      os->print_cr("  hugepage size: " EXACTFMT, EXACTFMTARGS(s));
     }
-    os->print_cr("  default pagesize: " EXACTFMT, EXACTFMTARGS(_default_hugepage_size));
+    os->print_cr("  default hugepage size: " EXACTFMT, EXACTFMTARGS(_default_hugepage_size));
   } else {
     os->print_cr("  unknown.");
   }
@@ -196,10 +191,8 @@ void THPSupport::scan_os() {
 
   // Scan large page size for THP from hpage_pmd_size
   _pagesize = 0;
-  if (_mode != THPMode::never) {
-    read_number_file("/sys/kernel/mm/transparent_hugepage/hpage_pmd_size", &_pagesize);
-    assert(_pagesize > 0, "Expected");
-  }
+  read_number_file("/sys/kernel/mm/transparent_hugepage/hpage_pmd_size", &_pagesize);
+  assert(_pagesize > 0, "Expected");
   _initialized = true;
 
   LogTarget(Info, pagesize) lt;
@@ -212,11 +205,9 @@ void THPSupport::scan_os() {
 void THPSupport::print_on(outputStream* os) {
   if (_initialized) {
     os->print_cr("Transparent hugepage (THP) support:");
-    os->print_cr("  mode: %s",
+    os->print_cr("  THP mode: %s",
         (_mode == THPMode::always ? "always" : (_mode == THPMode::never ? "never" : "madvise")));
-    if (_mode != THPMode::never) {
-      os->print_cr("  pagesize: " EXACTFMT, EXACTFMTARGS(_pagesize));
-    }
+    os->print_cr("  THP pagesize: " EXACTFMT, EXACTFMTARGS(_pagesize));
   } else {
     os->print_cr("  unknown.");
   }
--- a/src/hotspot/os/linux/hugepages.cpp
+++ b/src/hotspot/os/linux/hugepages.cpp
@@ -1,0 +1,236 @@
+/*
+ * Copyright (c) 2005, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2023, Red Hat Inc. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ *
+ */
+
+#include "precompiled.hpp"
+#include "hugepages.hpp"
+
+#include "logging/log.hpp"
+#include "logging/logStream.hpp"
+#include "runtime/os.hpp"
+#include "utilities/debug.hpp"
+#include "utilities/globalDefinitions.hpp"
+#include "utilities/ostream.hpp"
+
+#include <dirent.h>
+
+StaticHugePageSupport::StaticHugePageSupport() :
+  _initialized(false), _pagesizes(), _default_hugepage_size(SIZE_MAX) {}
+
+os::PageSizes StaticHugePageSupport::pagesizes() const {
+  assert(_initialized, "Not initialized");
+  return _pagesizes;
+}
+
+size_t StaticHugePageSupport::default_hugepage_size() const {
+  assert(_initialized, "Not initialized");
+  return _default_hugepage_size;
+}
+
+// Scan /proc/meminfo and return value of Hugepagesize
+static size_t scan_default_hugepagesize() {
+  size_t pagesize = 0;
+
+  // large_page_size on Linux is used to round up heap size. x86 uses either
+  // 2M or 4M page, depending on whether PAE (Physical Address Extensions)
+  // mode is enabled. AMD64/EM64T uses 2M page in 64bit mode. IA64 can use
+  // page as large as 1G.
+  //
+  // Here we try to figure out page size by parsing /proc/meminfo and looking
+  // for a line with the following format:
+  //    Hugepagesize:     2048 kB
+  //
+  // If we can't determine the value (e.g. /proc is not mounted, or the text
+  // format has been changed), we'll set largest page size to 0
+
+  FILE *fp = os::fopen("/proc/meminfo", "r");
+  if (fp) {
+    while (!feof(fp)) {
+      int x = 0;
+      char buf[16];
+      if (fscanf(fp, "Hugepagesize: %d", &x) == 1) {
+        if (x && fgets(buf, sizeof(buf), fp) && strcmp(buf, " kB\n") == 0) {
+          pagesize = x * K;
+          break;
+        }
+      } else {
+        // skip to next line
+        for (;;) {
+          int ch = fgetc(fp);
+          if (ch == EOF || ch == (int)'\n') break;
+        }
+      }
+    }
+    fclose(fp);
+  }
+
+  return pagesize;
+}
+
+// Given a file that contains a single (integral) number, return that number, 0 and false if failed.
+static bool read_number_file(const char* file, size_t* out) {
+  (*out) = 0;
+  FILE* f = ::fopen(file, "r");
+  bool rc = false;
+  if (f != nullptr) {
+    uint64_t i = 0;
+    if (::fscanf(f, SIZE_FORMAT, out) == 1) {
+      rc = true;
+    }
+    ::fclose(f);
+  }
+  return rc;
+}
+
+static const char* const sys_hugepages = "/sys/kernel/mm/hugepages";
+
+// Scan all directories in /sys/kernel/mm/hugepages/hugepages-xxxx
+// to discover the available page sizes
+static os::PageSizes scan_hugepages() {
+
+  os::PageSizes pagesizes;
+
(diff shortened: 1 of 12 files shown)


src/hotspot/os/linux/hugepages.cpp
```diff
@@ -0,0 +1,236 @@
...
+  _default_hugepage_size = scan_default_hugepagesize();
+  assert(_pagesizes.contains(_default_hugepage_size),
+         "Unexpected configuration: default pagesize (" SIZE_FORMAT ") "
+         "has no associated directory in /sys/kernel/mm/hugepages..", _default_hugepage_size);
+  _initialized = true;
+  LogTarget(Info, pagesize) lt;
+  if (lt.is_enabled()) {
+    LogStream ls(lt);
+    print_on(&ls);
+  }
```
jdksjolen, line 160: As `print_on` prints several lines you want to use `NonInterleavingLogStream`.

Do this:

```c++
LogMessage(pagesize) lm;
NonInterleavingLogStream ls{LogLevelType::Info, lm};
if (ls.is_enabled()) {
  print_on(&ls);
}
```

jdksjolen: I'm just using `{}`-init here, it's the same as using `()` so your choice on which to use.

jdksjolen (commented):
Hi Thomas,

Thanks for these changes. I've some suggestions for the code, but the conceptual change is good.

src/hotspot/os/linux/hugepages.cpp
```diff
@@ -0,0 +1,236 @@
...
+          break;
+        }
+      } else {
+        // skip to next line
+        for (;;) {
+          int ch = fgetc(fp);
+          if (ch == EOF || ch == (int)'\n') break;
+        }
+      }
+    }
```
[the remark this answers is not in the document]
tstuefe: Definitely simpler, but this code moved (almost) verbatim from the old `scan_default_large_page_size` in os_linux.cpp, and I would like to keep code changes to a minimum.

src/hotspot/os/linux/hugepages.cpp
```diff
@@ -0,0 +1,236 @@
...
+  // 2M or 4M page, depending on whether PAE (Physical Address Extensions)
+  // mode is enabled. AMD64/EM64T uses 2M page in 64bit mode. IA64 can use
+  // page as large as 1G.
+  //
+  // Here we try to figure out page size by parsing /proc/meminfo and looking
+  // for a line with the following format:
+  //    Hugepagesize:     2048 kB
+  //
+  // If we can't determine the value (e.g. /proc is not mounted, or the text
+  // format has been changed), we'll set largest page size to 0
```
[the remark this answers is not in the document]
tstuefe: Again, unchanged code from before I'd like to keep unchanged to keep this patch minimally invasive. Mentally earmarked for future cleanups.

src/hotspot/os/linux/hugepages.cpp
```diff
@@ -0,0 +1,236 @@
...
+    }
+    fclose(fp);
+  }
+
+  return pagesize;
+}
+
+// Given a file that contains a single (integral) number, return that number, 0 and false if failed.
+static bool read_number_file(const char* file, size_t* out) {
+  (*out) = 0;
```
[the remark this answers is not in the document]
tstuefe: I'd like to leave it.

While I agree with you, Hotspot style guide says something like "write parentheses if order of ops is unclear". Most of the code in hotspot uses parentheses in assignments like this. I'd like to avoid mixing styles.

src/hotspot/os/linux/hugepages.cpp
```diff
@@ -0,0 +1,236 @@
...
+  //
+  // Here we try to figure out page size by parsing /proc/meminfo and looking
+  // for a line with the following format:
+  //    Hugepagesize:     2048 kB
+  //
+  // If we can't determine the value (e.g. /proc is not mounted, or the text
+  // format has been changed), we'll set largest page size to 0
+
+  FILE *fp = os::fopen("/proc/meminfo", "r");
+  if (fp) {
```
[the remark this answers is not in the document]
tstuefe: Absolutely agree, for a future cleanup, see above.

src/hotspot/os/linux/hugepages.cpp
tstuefe: Okay. We are at initialization time, though; everything should still be single-threaded here. I dislike providing the log level twice, but LogTarget does not expose it.

tstuefe: Wait, if I do this, I need to backport [8286180](https://bugs.openjdk.org/browse/JDK-8286180) before backporting this fix., right? :(

Oh well, chances are this does not integrate cleanly anyway.

tstuefe: The more I think about this, the more doubts I have, sorry. Logging should not cost anything if disabled. But with your proposal, I pay for the construction of both LogMessage and NonInterleavingLogStream every time.

tstuefe:
@jdksjolen Thanks a lot for your review!

See inline remarks. Most of your code suggestions are good, but this patch just moved the static hugepage detection parts out of os_linux.cpp, and left them (mostly) alone otherwise, to avoid adding regressions. I'll keep your input in mind for the next round of cleanups.

Wrt NonInterleavingLogStream, I decided to not do that. It is not needed, since we are single-threaded, and it without it I may still be lucky to get it integrated cleanly to at least 17.

I added a new repro case. Please take a look!

Thank you.

src/hotspot/os/linux/hugepages.cpp
jdksjolen: >But with your proposal, I pay for the construction of both LogMessage and NonInterleavingLogStream every time.

Yeah, I made a mental footnote that `LogTarget` should be able to be used to initialize these inside of an if-block. UL is still a bit painful about these things. My suggestion sounds like a massive pain, I'm happy if you don't do it :).

tstuefe: > > But with your proposal, I pay for the construction of both LogMessage and NonInterleavingLogStream every time.
> 
> Yeah, I made a mental footnote that `LogTarget` should be able to be used to initialize these inside of an if-block. UL is still a bit painful about these things. My suggestion sounds like a massive pain, I'm happy if you don't do it :).

Thank you @jdksjolen !

test/hotspot/jtreg/runtime/os/HugePageConfiguration.java
```diff
@@ -0,0 +1,222 @@
...
+    @java.lang.Override
+    public int hashCode() {
+        return Objects.hash(_staticHugePageSizes, _staticDefaultHugePageSize, _thpMode, _thpPageSize);
+    }
+
+    private static long readDefaultHugePageSizeFromOS() {
+        Pattern pat = Pattern.compile("Hugepagesize: *(\\d+) +kB");
+        Scanner scanner;
+        try {
+            scanner = new Scanner(new File("/proc/meminfo"));
```
jdksjolen, line 92: Nit: You can use try-with-resources here.

test/hotspot/jtreg/runtime/os/HugePageConfiguration.java
```diff
@@ -0,0 +1,222 @@
...
+            }
+        }
+        return pagesizes;
+    }
+
+    private static THPMode readTHPModeFromOS() {
+        THPMode mode;
+        try {
+            String file = "/sys/kernel/mm/transparent_hugepage/enabled";
+            BufferedReader reader = new BufferedReader(new FileReader(file));
```
jdksjolen, line 128: This leaks the file descriptor (never closed), use try-with-resources.

test/hotspot/jtreg/runtime/os/HugePageConfiguration.java
```diff
@@ -0,0 +1,222 @@
...
+            mode = THPMode.never;
+        }
+        return mode;
+    }
+
+    private static long readTHPPageSizeFromOS() {
+        long pagesize = 0;
+        try {
+            String file = "/sys/kernel/mm/transparent_hugepage/hpage_pmd_size";
+            BufferedReader reader = new BufferedReader(new FileReader(file));
```
jdksjolen, line 149: This leaks the file descriptor (never closed), use try-with-resources.

jdksjolen (commented):
Some of the methods seems to leak file descriptors, other than that tests looks good.

test/hotspot/jtreg/runtime/os/HugePageConfiguration.java
tstuefe: Makes sense

tstuefe:
Thanks @jdksjolen! I changed the test accordingly.

jdksjolen (approved):
Thank you Thomas!

These changes look good to me now, I'm approving this.

tstuefe:
Gentle ping, second review needed.

src/hotspot/os/linux/hugepages.cpp
```diff
@@ -0,0 +1,227 @@
...
+  assert(_initialized, "Not initialized");
+  return _pagesizes;
+}
+
+size_t StaticHugePageSupport::default_hugepage_size() const {
+  assert(_initialized, "Not initialized");
+  return _default_hugepage_size;
+}
+
+// Scan /proc/meminfo and return value of Hugepagesize
```
tstuefe, line 51: Reviewer notes: 
`scan_default_hugepagesize` is just a code move, it used to live in `scan_default_large_page_size` in os_linux.cpp. It is responsible for reading the default *static* hugepage size from /proc/meminfo. This code is mostly untouched to reduce chance for regressions (though the code could be tightened and cleaned for sure).

tstuefe:
May I have a second review?

test/hotspot/jtreg/runtime/os/HugePageConfiguration.java
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    public HugePageConfiguration(Set<Long> _staticHugePageSizes, long _staticDefaultHugePageSize, THPMode _thpMode, long _thpPageSize) {
+        this._staticHugePageSizes = _staticHugePageSizes;
+        this._staticDefaultHugePageSize = _staticDefaultHugePageSize;
+        this._thpMode = _thpMode;
+        this._thpPageSize = _thpPageSize;
+    }
+
+    @java.lang.Override
```
dholmes-ora, line 65: The `java.lang` is not necessary and typically not used.

src/hotspot/share/utilities/globalDefinitions.hpp
```diff
@@ -409,6 +409,9 @@ inline size_t byte_size_in_exact_unit(size_t s) {
   return s;
 }
 
+#define EXACTFMT            SIZE_FORMAT "%s"
+#define EXACTFMTARGS(s)     byte_size_in_exact_unit(s), exact_unit_for_byte_size(s)
```
dholmes-ora, line 413: Not sure this is worthy of inclusion in globalDefinitions. And I struggle to understand the use of "exact" in this context (pre-existing).

src/hotspot/os/linux/hugepages.cpp
```diff
@@ -0,0 +1,227 @@
...
+      }
+    }
+    fclose(fp);
+  }
+
+  return pagesize;
+}
+
+// Given a file that contains a single (integral) number, return that number, 0 and false if failed.
+static bool read_number_file(const char* file, size_t* out) {
```
dholmes-ora, line 92: Why `bool` return when you don't check it? Also `*out == 0` also indicates failure. Though why use an out parameter instead of just returning the value?

dholmes-ora (commented):
Not an area I am familiar with but scanning through the code this seems reasonable. A few minor comments below.

Thanks.

tstuefe:
> Not an area I am familiar with but scanning through the code this seems reasonable. A few minor comments below.
> 
> Thanks.

Many thanks, @dholmes-ora! Remarks follow.

src/hotspot/os/linux/hugepages.cpp
tstuefe: I will fix the caller to handle the error. Keeping error state from "0" return separate is just cleaner, and makes it possible to move/reuse this function should the need arise.

tstuefe:
@dholmes-ora : I massaged in your feedback.

dholmes-ora (approved):
Nothing further from me.

Thanks.

tstuefe:
Thanks, @dholmes-ora !

/integrate

PR #14744, opened 2023-06-30
8302618: [macos] Problem typing uppercase letters with java.awt.Robot when moving mouse

honkar-jdk:
**Problem:**

On macOS, Robot erroneously produces lowercase letter when mouse is moved (manually) in unison with Robot's keyEvents. This issue was originally logged by a developer of an on-screen accessibility keyboard  - TouchBoard. Originally reported at https://github.com/adoptium/adoptium-support/issues/710

This issue is reproducible on JDK versions 22 to 11, but works fine on JDK-8. (details below) 

This issue is not restricted to the Shift modifier key and causes problems with other modifier keys as well and in some scenarios without any external mouse movement.

- This works correctly on JDK-8 up to JDK-9+129 when Accessibility APIs (AXUIElementCreateSystemWide/ AXUIElementPostKeyboardEvent) were used. Later on it was changed to CGEvents. 

- With the present code, the issue occurs at [CRobot.m#L295](https://github.com/openjdk/jdk/blob/ac6af6a64099c182e982a0a718bc1b780cef616e/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m#L295.)  The flags gets reset or cleared when mouse is moved physically in unison with Robot's key events.

- The physical mouse movement causes the event flags to be reset. 

**Impact:**

Modifier keys don't work as expected when using Robot with any simultaneous physical mouse movement and in case of TouchBoard, this behavior breaks the usability of the on-screen a11y keyboard. There is no known workaround for this particular use case except for reverting to JDK-8. More details on this use case [here.](https://github.com/adoptium/adoptium-support/issues/710#issuecomment-1594103280)

**Proposed Fix:**

- In order to avoid resetting of the CGEventFlags here [CRobot.m#L295](https://github.com/openjdk/jdk/blob/ac6af6a64099c182e982a0a718bc1b780cef616e/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m#L295.), the CGEvent flag state is obtained in `initRobot` (stored in initFlags) which is later used within `CRobot_keyEvent`.

- The incoming keyCode is used to determine whether it is a modifier key and the corresponding modifierFlagMask is either added or cleared from the initFlags based on whether the modifier key was pressed or released.

- Finally, only the required and known flag bits from initFlag are copied over to local flag which is used in `CGEventSetFlags()`.

**Testing:**

The newly added test - RobotModifierMaskTest.java tests for Shift, Caps, Control, Option and Command keys.
The test runs in 2 modes - as automated and manual test.

CASE 1 :  As automated test. No user interaction needed.
CASE 2 :  When run in manual mode - it is semi-automatic, i.e Typing is automated via Robot and the user is required to manually move the mouse (without clicking/dragging) at the same time (concurrently when Robot is typing). Check if the test passes or fails.

CI Testing (Automated Client tests -  Swing + AWT) works as expected with the fix and does not break any existing Robot related tests in test/jdk/java/awt/Robot folder.

---------
### Progress

### Issue
 * [JDK-8302618](https://bugs.openjdk.org/browse/JDK-8302618): [macos] Problem typing uppercase letters with java.awt.Robot when moving mouse (**Bug** - P4)

### Reviewers
 * [Alexander Zuev](https://openjdk.org/census#kizune) (@azuev-java - **Reviewer**) ⚠️ Review applies to [2b682e7b](https://git.openjdk.org/jdk/pull/14744/files/2b682e7b8fb778ff626a7ce266c368a9628ad82e)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14744/head:pull/14744` \
`$ git checkout pull/14744`

`$ git checkout pull/14744` \
`$ git pull https://git.openjdk.org/jdk.git pull/14744/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14744`

`$ git pr show -t 14744`

</details>

<a href="https://git.openjdk.org/jdk/pull/14744.diff">https://git.openjdk.org/jdk/pull/14744.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14744#issuecomment-1615251518)

### DIFF
--- a/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
+++ b/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
@@ -319,7 +319,7 @@ Java_sun_lwawt_macosx_CRobot_keyEvent
             }
 
             CGEventFlags flags = CGEventSourceFlagsState(kCGEventSourceStateHIDSystemState);
-            flags  = (initFlags & allModifiersMask) | (flags & (~allModifiersMask));
+            flags = (initFlags & allModifiersMask) | (flags & (~allModifiersMask));
             CGEventSetFlags(event, flags);
 
             CGEventPost(kCGHIDEventTap, event);
--- a/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
+++ b/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
@@ -63,6 +63,7 @@ static NSTimeInterval gsLastClickTime;
 static int gsEventNumber;
 static int* gsButtonEventNumber;
 static NSTimeInterval gNextKeyEventTime;
+static CGEventFlags initFlags;
 
 static inline CGKeyCode GetCGKeyCode(jint javaKeyCode);
 
@@ -121,6 +122,13 @@ Java_sun_lwawt_macosx_CRobot_initRobot
             jboolean copy = JNI_FALSE;
 
             setupDone = 1;
+            // initialize CGEventFlags here - which is used in keyEvent
+            initFlags = CGEventSourceFlagsState(kCGEventSourceStateHIDSystemState);
+
+            // Toggle the Function flag bits if they are set
+            if ((initFlags & kCGEventFlagMaskSecondaryFn) != 0) {
+                initFlags ^= kCGEventFlagMaskSecondaryFn;
+            }
             // Don't block local events after posting ours
             CGSetLocalEventsSuppressionInterval(0.0);
 
@@ -288,21 +296,45 @@ Java_sun_lwawt_macosx_CRobot_keyEvent
 {
     autoDelay(NO);
     [ThreadUtilities performOnMainThreadWaiting:YES block:^(){
-        CGEventSourceRef source = CGEventSourceCreate(kCGEventSourceStateHIDSystemState);
-        CGKeyCode keyCode = GetCGKeyCode(javaKeyCode);
-        CGEventRef event = CGEventCreateKeyboardEvent(source, keyCode, keyPressed);
+       CGEventSourceRef source = CGEventSourceCreate(kCGEventSourceStateHIDSystemState);
+       CGKeyCode keyCode = GetCGKeyCode(javaKeyCode);
+       CGEventRef event = CGEventCreateKeyboardEvent(source, keyCode, keyPressed);
+
+       NSLog(@"JAVA KEYCODE: %d", javaKeyCode);
+       NSLog(@"CG KEYCODE: %hu", keyCode);
+
         if (event != NULL) {
-            CGEventFlags flags = CGEventSourceFlagsState(kCGEventSourceStateHIDSystemState);
-            if ((flags & kCGEventFlagMaskSecondaryFn) != 0) {
-                flags ^= kCGEventFlagMaskSecondaryFn;
-                CGEventSetFlags(event, flags);
+            NSLog(@"BEFORE Contents of Flag: %llu", initFlags);
+            switch (keyCode) {
+                case OSX_Shift:
+                    initFlags = keyPressed ? (initFlags | kCGEventFlagMaskShift) : (initFlags & ~kCGEventFlagMaskShift);
+                    break;
+                case OSX_CapsLock:
+                    initFlags ^= kCGEventFlagMaskAlphaShift;
+                    break;
+                case OSX_Control:
+                    initFlags = keyPressed ? (initFlags | kCGEventFlagMaskControl) : (initFlags & ~kCGEventFlagMaskControl);
+                    break;
+                case OSX_Option:
+                    initFlags = keyPressed ? (initFlags | kCGEventFlagMaskAlternate) : (initFlags & ~kCGEventFlagMaskAlternate);
+                    break;
+                case OSX_Command:
+                    initFlags = keyPressed ? (initFlags | kCGEventFlagMaskCommand) : (initFlags & ~kCGEventFlagMaskCommand);
+                    break;
             }
+
+            CGEventFlags flags = CGEventSourceFlagsState(kCGEventSourceStateHIDSystemState);
+            flags |= initFlags;
+            CGEventSetFlags(event, flags);
+
             CGEventPost(kCGHIDEventTap, event);
             CFRelease(event);
         }
         if (source != NULL) {
             CFRelease(source);
         }
+        NSLog(@"AFTER EVENT & SOURCE RELEASE - Contents of Flag: %llu", initFlags);
+        NSLog(@"----------------------------------------");
     }];
 }
 
--- a/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
+++ b/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
@@ -304,17 +304,11 @@ Java_sun_lwawt_macosx_CRobot_keyEvent
         CGKeyCode keyCode = GetCGKeyCode(javaKeyCode);
         CGEventRef event = CGEventCreateKeyboardEvent(source, keyCode, keyPressed);
 
-        NSLog(@"JAVA KEYCODE: %d", javaKeyCode);
-        NSLog(@"CG KEYCODE: %hu", keyCode);
-
         if (event != NULL) {
-            NSLog(@"BEFORE INIT FLAG: %llu", initFlags);
             int flagMaskValue = GetCGKeyMask(keyCode);
             if (OSX_Undefined != flagMaskValue) {
-                NSLog(@"KeyCode %d and Mask Value: %d", keyCode, flagMaskValue);
                 if (keyCode == OSX_CapsLock) {
                     if (keyPressed) {
-                     NSLog(@"IF CapsLock");
                      initFlags ^= flagMaskValue;
                     }
                 } else {
@@ -334,8 +328,6 @@ Java_sun_lwawt_macosx_CRobot_keyEvent
         if (source != NULL) {
             CFRelease(source);
         }
-        NSLog(@"AFTER INIT FLAG - Contents of Flag: %llu", initFlags);
-        NSLog(@"----------------------------------------");
     }];
 }
 
--- a/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
+++ b/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
@@ -305,12 +305,15 @@ Java_sun_lwawt_macosx_CRobot_keyEvent
         NSLog(@"CG KEYCODE: %hu", keyCode);
 
         if (event != NULL) {
-            NSLog(@"BEFORE Contents of Flag: %llu", initFlags);
+            NSLog(@"BEFORE INIT FLAG: %llu", initFlags);
             int flagMaskValue = GetCGKeyMask(keyCode);
             if (OSX_Undefined != flagMaskValue) {
                 NSLog(@"KeyCode %d and Mask Value: %d", keyCode, flagMaskValue);
                 if (keyCode == OSX_CapsLock) {
-                    initFlags ^= flagMaskValue;
+                    if (keyPressed) {
+                     NSLog(@"IF CapsLock");
+                     initFlags ^= flagMaskValue;
+                    }
                 } else {
                     initFlags = keyPressed
                                 ? (initFlags | flagMaskValue)    // add flag bits if modifier key pressed
@@ -319,7 +322,11 @@ Java_sun_lwawt_macosx_CRobot_keyEvent
             }
 
             CGEventFlags flags = CGEventSourceFlagsState(kCGEventSourceStateHIDSystemState);
-            flags |= initFlags;
+            int allModifiersMask = kCGEventFlagMaskShift | kCGEventFlagMaskControl
+                                  | kCGEventFlagMaskAlternate | kCGEventFlagMaskCommand
+                                  | kCGEventFlagMaskAlphaShift;
+
+            flags  = (initFlags & allModifiersMask) | (flags & (!allModifiersMask));
             CGEventSetFlags(event, flags);
 
             CGEventPost(kCGHIDEventTap, event);
@@ -328,7 +335,7 @@ Java_sun_lwawt_macosx_CRobot_keyEvent
         if (source != NULL) {
             CFRelease(source);
         }
-        NSLog(@"AFTER EVENT & SOURCE RELEASE - Contents of Flag: %llu", initFlags);
+        NSLog(@"AFTER INIT FLAG - Contents of Flag: %llu", initFlags);
         NSLog(@"----------------------------------------");
     }];
 }
--- a/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
+++ b/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
@@ -66,7 +66,7 @@ static NSTimeInterval gNextKeyEventTime;
 static CGEventFlags initFlags;
 static int allModifiersMask = kCGEventFlagMaskShift | kCGEventFlagMaskControl
                               | kCGEventFlagMaskAlternate | kCGEventFlagMaskCommand
-                              | kCGEventFlagMaskAlphaShift;
+                              | kCGEventFlagMaskAlphaShift | kCGEventFlagMaskSecondaryFn;
 
 static inline CGKeyCode GetCGKeyCode(jint javaKeyCode);
 static inline int GetCGKeyMask(int cgKeyCode);
--- a/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
+++ b/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
@@ -423,7 +423,8 @@ static inline CGKeyCode GetCGKeyCode(jint javaKeyCode)
     return [keyCodeMapping getOSXKeyCodeForJavaKey:javaKeyCode];
 }
 
-static inline int GetCGKeyMask(int cgKeyCode) {
+static inline int GetCGKeyMask(int cgKeyCode)
+{
     CRobotKeyCodeMapping *keyCodeMapping = [CRobotKeyCodeMapping sharedInstance];
     return [keyCodeMapping getFlagMaskForCGKey:cgKeyCode];
 }
--- a/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
+++ b/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -423,8 +423,7 @@ static inline CGKeyCode GetCGKeyCode(jint javaKeyCode)
     return [keyCodeMapping getOSXKeyCodeForJavaKey:javaKeyCode];
 }
 
-static inline int GetCGKeyMask(int cgKeyCode)
-{
+static inline int GetCGKeyMask(int cgKeyCode) {
     CRobotKeyCodeMapping *keyCodeMapping = [CRobotKeyCodeMapping sharedInstance];
     return [keyCodeMapping getFlagMaskForCGKey:cgKeyCode];
 }
--- a/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
+++ b/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
@@ -64,7 +64,7 @@ static int gsEventNumber;
 static int* gsButtonEventNumber;
 static NSTimeInterval gNextKeyEventTime;
 static CGEventFlags initFlags;
-static int allModifiersMask = kCGEventFlagMaskShift | kCGEventFlagMaskControl
+static CGEventFlags allModifiersMask = kCGEventFlagMaskShift | kCGEventFlagMaskControl
                               | kCGEventFlagMaskAlternate | kCGEventFlagMaskCommand
                               | kCGEventFlagMaskAlphaShift | kCGEventFlagMaskSecondaryFn;
 
--- a/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
+++ b/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
@@ -309,7 +309,7 @@ Java_sun_lwawt_macosx_CRobot_keyEvent
             if (OSX_Undefined != flagMaskValue) {
                 if (keyCode == OSX_CapsLock) {
                     if (keyPressed) {
-                     initFlags ^= flagMaskValue;
+                        initFlags ^= flagMaskValue;
                     }
                 } else {
                     initFlags = keyPressed
@@ -319,7 +319,7 @@ Java_sun_lwawt_macosx_CRobot_keyEvent
             }
 
             CGEventFlags flags = CGEventSourceFlagsState(kCGEventSourceStateHIDSystemState);
-            flags  = (initFlags & allModifiersMask) | (flags & (!allModifiersMask));
+            flags  = (initFlags & allModifiersMask) | (flags & (~allModifiersMask));
             CGEventSetFlags(event, flags);
 
             CGEventPost(kCGHIDEventTap, event);
--- a/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
+++ b/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
@@ -296,12 +296,12 @@ Java_sun_lwawt_macosx_CRobot_keyEvent
 {
     autoDelay(NO);
     [ThreadUtilities performOnMainThreadWaiting:YES block:^(){
-       CGEventSourceRef source = CGEventSourceCreate(kCGEventSourceStateHIDSystemState);
-       CGKeyCode keyCode = GetCGKeyCode(javaKeyCode);
-       CGEventRef event = CGEventCreateKeyboardEvent(source, keyCode, keyPressed);
+        CGEventSourceRef source = CGEventSourceCreate(kCGEventSourceStateHIDSystemState);
+        CGKeyCode keyCode = GetCGKeyCode(javaKeyCode);
+        CGEventRef event = CGEventCreateKeyboardEvent(source, keyCode, keyPressed);
 
-       NSLog(@"JAVA KEYCODE: %d", javaKeyCode);
-       NSLog(@"CG KEYCODE: %hu", keyCode);
+        NSLog(@"JAVA KEYCODE: %d", javaKeyCode);
+        NSLog(@"CG KEYCODE: %hu", keyCode);
 
         if (event != NULL) {
             NSLog(@"BEFORE Contents of Flag: %llu", initFlags);
--- a/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
+++ b/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
@@ -64,6 +64,9 @@ static int gsEventNumber;
 static int* gsButtonEventNumber;
 static NSTimeInterval gNextKeyEventTime;
 static CGEventFlags initFlags;
+static int allModifiersMask = kCGEventFlagMaskShift | kCGEventFlagMaskControl
+                              | kCGEventFlagMaskAlternate | kCGEventFlagMaskCommand
+                              | kCGEventFlagMaskAlphaShift;
 
 static inline CGKeyCode GetCGKeyCode(jint javaKeyCode);
 static inline int GetCGKeyMask(int cgKeyCode);
@@ -322,10 +325,6 @@ Java_sun_lwawt_macosx_CRobot_keyEvent
             }
 
             CGEventFlags flags = CGEventSourceFlagsState(kCGEventSourceStateHIDSystemState);
-            int allModifiersMask = kCGEventFlagMaskShift | kCGEventFlagMaskControl
-                                  | kCGEventFlagMaskAlternate | kCGEventFlagMaskCommand
-                                  | kCGEventFlagMaskAlphaShift;
-
             flags  = (initFlags & allModifiersMask) | (flags & (!allModifiersMask));
             CGEventSetFlags(event, flags);
 
--- a/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
+++ b/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
@@ -66,6 +66,7 @@ static NSTimeInterval gNextKeyEventTime;
 static CGEventFlags initFlags;
 
 static inline CGKeyCode GetCGKeyCode(jint javaKeyCode);
+static inline int GetCGKeyMask(int cgKeyCode);
 
 static void PostMouseEvent(const CGPoint point, CGMouseButton button,
                            CGEventType type, int clickCount, int eventNumber);
@@ -125,7 +126,7 @@ Java_sun_lwawt_macosx_CRobot_initRobot
             // initialize CGEventFlags here - which is used in keyEvent
             initFlags = CGEventSourceFlagsState(kCGEventSourceStateHIDSystemState);
 
-            // Toggle the Function flag bits if they are set
+            // Clear Function flag bits if they are set
             if ((initFlags & kCGEventFlagMaskSecondaryFn) != 0) {
                 initFlags ^= kCGEventFlagMaskSecondaryFn;
             }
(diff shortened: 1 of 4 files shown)


src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
```diff
@@ -63,8 +63,13 @@
 static int gsEventNumber;
 static int* gsButtonEventNumber;
 static NSTimeInterval gNextKeyEventTime;
+static CGEventFlags initFlags;
+static int allModifiersMask = kCGEventFlagMaskShift | kCGEventFlagMaskControl
+                              | kCGEventFlagMaskAlternate | kCGEventFlagMaskCommand
+                              | kCGEventFlagMaskAlphaShift;
```
azuev-java, line 69: You need to add kCGEventFlagMaskSecondaryFn here otherwise the "secondary fn fix" might catch regression.

honkar-jdk: Updated.

test/jdk/java/awt/Robot/RobotModifierMaskTest.java
```diff
@@ -0,0 +1,231 @@
...
+        robot.keyPress(KeyEvent.VK_CONTROL);
+        robot.keyPress(KeyEvent.VK_A);
+        robot.keyRelease(KeyEvent.VK_A);
+        robot.keyRelease(KeyEvent.VK_CONTROL);
+
+        robot.delay(500);
+
+        if (jTextArea.getCaretPosition() != EXPECTED_CARET_POS_CTRL) {
+            errorLog.append("For Control key, Actual and Expected caret position differ \n" +
+                    "Expected Position : " + EXPECTED_CARET_POS_CTRL + " Actual Position : " + jTextArea.getCaretPosition() + "\n");
```
lawrence-andrew, line 217: Most of the errorLog.append("") are more than 80 column length.

src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
```diff
@@ -398,6 +423,12 @@ static inline CGKeyCode GetCGKeyCode(jint javaKeyCode)
     return [keyCodeMapping getOSXKeyCodeForJavaKey:javaKeyCode];
 }
 
+static inline int GetCGKeyMask(int cgKeyCode)
```
kumarabhi006, line 426:
```suggestion
static inline int GetCGKeyMask(int cgKeyCode) {
```

src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
kumarabhi006, line 1: Please update copyright year here and in other files as well.

src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
aivanov-jdk: I disagree: the common style throughout this file is to put the opening brace on the following line. A few lines above, `GetCGKeyCode` follows the same style.

test/jdk/java/awt/Robot/RobotModifierMaskTest.java
```diff
@@ -0,0 +1,231 @@
...
+            CASE 2 : Run the test in semi-automated mode. While the Robot in typing,
+                     manually move the mouse (without clicking/dragging). Check if the test Passes or Fails.\n\n
+
+            NOTE: User doesn't need to compare the actual vs expected result in Case 2.
+            The test compares it.""";
+
+    public static void main(String[] args) throws Exception {
+        if (OSInfo.getOSType() != OSInfo.OSType.MACOSX) {
+            System.out.println("This test is for MacOS platform only");
+            return;
```
aivanov-jdk, line 78: Use `jtreg.SkippedException` to indicate the test isn't applicable in this environment. The jtreg treats this exception in a special: the doesn't fail, it doesn't pass either — it's skipped.

test/jdk/java/awt/Robot/RobotModifierMaskTest.java
```diff
@@ -0,0 +1,231 @@
...
+        robot.delay(500);
+
+        if (!jTextArea.getText().equals(EXPECTED_RESULT_META)) {
+            errorLog.append("For Command key, Actual and Expected results differ \n"+
+                    "Expected Text : " + EXPECTED_RESULT_META + " Actual Text : " + jTextArea.getText() + "\n");
+        }
+    }
+
+    private static void testAltKey() {
+        jTextArea.setText("");
```
aivanov-jdk, line 183: Should be called on EDT.

test/jdk/java/awt/Robot/RobotModifierMaskTest.java
```diff
@@ -0,0 +1,231 @@
...
+            robot.keyPress(KeyEvent.VK_A);
+            robot.keyRelease(KeyEvent.VK_A);
+            robot.keyRelease(KeyEvent.VK_ALT);
+
+            robot.delay(100);
+        }
+
+        robot.delay(500);
+
+        if (!jTextArea.getText().equals(EXPECTED_RESULT_ALT)) {
```
aivanov-jdk, line 196: Should be called on EDT.

test/jdk/java/awt/Robot/RobotModifierMaskTest.java
```diff
@@ -0,0 +1,231 @@
...
+            System.out.println("This test is for MacOS platform only");
+            return;
+        }
+
+        try {
+            robot = new Robot();
+            robot.setAutoWaitForIdle(true);
+            robot.setAutoDelay(200);
+
+            SwingUtilities.invokeAndWait(() -> createTestUI());
```
aivanov-jdk, line 86: I don't insist, yet IDE usually show a warning that such a lambda expression can be replaced with a method reference.

test/jdk/java/awt/Robot/RobotModifierMaskTest.java
```diff
@@ -0,0 +1,231 @@
...
+
+    private static final String INSTRUCTIONS = """
+            This test is a semi-automatic test which checks the effect of typing modifier keys
+            through a Robot.\n
+            It tests the following key modifiers - Shift, Caps, Control, Option and Command.
+            It needs to be checked for following two scenarios \n
+
+            CASE 1 : Run the test as an automated test and let the Robot go through all the test cases.\n
+            CASE 2 : Run the test in semi-automated mode. While the Robot in typing,
+                     manually move the mouse (without clicking/dragging). Check if the test Passes or Fails.\n\n
```
aivanov-jdk: How does a user do this?

src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
honkar-jdk: @aivanov-jdk Yeah, I wasn't sure of this - some methods follow Java-style of having the open brace on the same line and few others have it on the next line.

aivanov-jdk: Follow the style of the surrounding code so that added or modified lines don't stand out.

honkar-jdk:
Test updated to run in automated and manual mode separately. Test instruction frame updated to appear only when executed in manual mode.

src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
honkar-jdk: Updated

test/jdk/java/awt/Robot/RobotModifierMaskTest.java
honkar-jdk: Test updated to run twice (automated & manual), to cover the two cases.

src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
```diff
@@ -291,12 +303,25 @@ static inline void autoDelay(BOOL isMove) {
...
-            CGEventFlags flags = CGEventSourceFlagsState(kCGEventSourceStateHIDSystemState);
-            if ((flags & kCGEventFlagMaskSecondaryFn) != 0) {
-                flags ^= kCGEventFlagMaskSecondaryFn;
-                CGEventSetFlags(event, flags);
+            int flagMaskValue = GetCGKeyMask(keyCode);
+            if (OSX_Undefined != flagMaskValue) {
+                if (keyCode == OSX_CapsLock) {
+                    if (keyPressed) {
+                     initFlags ^= flagMaskValue;
+                    }
```
aivanov-jdk, line 313:
```suggestion
                    if (keyPressed) {
                        initFlags ^= flagMaskValue;
                    }
```
Shouldn't it get be indented?

azuev-java (approved):
Looks good.

src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
```diff
@@ -291,12 +303,25 @@ static inline void autoDelay(BOOL isMove) {
...
+                } else {
+                    initFlags = keyPressed
+                                ? (initFlags | flagMaskValue)    // add flag bits if modifier key pressed
+                                : (initFlags & ~flagMaskValue);  // clear flag bits if modifier key released
+                }
             }
+
+            CGEventFlags flags = CGEventSourceFlagsState(kCGEventSourceStateHIDSystemState);
+            flags  = (initFlags & allModifiersMask) | (flags & (!allModifiersMask));
```
aivanov-jdk, line 322: Should `!allModifiersMask` be `~allModifiersMask`? That is _bitwise_ NOT instead of _logical_ NOT.

```suggestion
            flags = (initFlags & allModifiersMask) | (flags & (!allModifiersMask));
```

test/jdk/java/awt/Robot/RobotModifierMaskTest.java
```diff
@@ -0,0 +1,275 @@
...
+            if (isManual) {
+                try {
+                    invokeAndWait(RobotModifierMaskTest::createInstructionsUI);
+                    robot.waitForIdle();
+                    while (!startTest) {
+                        robot.delay(200);
+                        if (!jFrame.isVisible()) {
+                            throw new RuntimeException("Test instruction frame closed");
+                        }
+                    }
```
aivanov-jdk, line 102: Instead of busy-waiting, you can use `CountDownLatch` to start the test. You should also release the latch if the instruction frame is closed.

test/jdk/java/awt/Robot/RobotModifierMaskTest.java
```diff
@@ -0,0 +1,275 @@
...
+
+    private static final String INSTRUCTIONS = """
+            This test is running in manual mode to check the effect of typing modifier keys
+            through Robot combined with concurrent external manual mouse movement.
+            It tests the following key modifiers - Shift, Caps, Control, Option and Command.
+
+            In this mode when the Robot starts to type, the user is required to concurrently
+            move the mouse (without clicking/dragging).
+            When ready click on the "Start" button to run the test.
+            """;
```
aivanov-jdk, line 78: And start moving the mouse?

test/jdk/java/awt/Robot/RobotModifierMaskTest.java
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    private static void createTestUI() {
+        String mode = isManual ? "MANUAL" : "AUTOMATED";
+        jFrame = new JFrame("RobotModifierMaskTest - Mode: " + mode);
+        jTextArea = new JTextArea("");
+        JScrollPane pane = new JScrollPane(jTextArea);
+        jFrame.getContentPane().add(pane, BorderLayout.CENTER);
+
+        jFrame.setSize(450,100);
```
aivanov-jdk, line 250:
```suggestion
        jFrame.setSize(450, 100);
```

test/jdk/java/awt/Robot/RobotModifierMaskTest.java
```diff
@@ -0,0 +1,275 @@
...
+        jTextArea = new JTextArea(INSTRUCTIONS);
+
+        JScrollPane pane = new JScrollPane(jTextArea);
+        jFrame.getContentPane().add(pane, BorderLayout.CENTER);
+
+        JButton jButton = new JButton("Start");
+        jButton.addActionListener(e -> startTest = true);
+        jFrame.getContentPane().add(jButton, BorderLayout.PAGE_END);
+
+        jFrame.setSize(560,200);
```
aivanov-jdk, line 237:
```suggestion
        jFrame.setSize(560, 200);
```

test/jdk/java/awt/Robot/RobotModifierMaskTest.java
```diff
@@ -0,0 +1,275 @@
...
+        jFrame.setDefaultCloseOperation(WindowConstants.DISPOSE_ON_CLOSE);
+        jFrame.setVisible(true);
+    }
+
+    private static void checkResult(String expectedResult, String prefixText)
+                                                                throws Exception {
+        invokeAndWait(() -> {
+            boolean condition = expectedResult.equals(EXPECTED_RESULT_CTRL)
+                                ? (jTextArea.getCaretPosition()
+                                     != Integer.parseInt(EXPECTED_RESULT_CTRL))
```
aivanov-jdk, line 261: You pass different expected result strings but here you always compare `expectedResult` to `EXPECTED_RESULT_CTRL`. Is it intentional? Do I miss anything?

I haven't run the test.

aivanov-jdk (commented):
I admit I don't understand the Objective-C code, yet it couple of points caught my attention. I mostly looked at the test as I did the previous time.

test/jdk/java/awt/Robot/RobotModifierMaskTest.java
honkar-jdk: Yes, the assertion for CTRL key is different for the rest and involves caret position comparison instead of expected strings (as in other cases), hence I have used `EXPECTED_RESULT_CTRL` to differentiate the cases.

src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
honkar-jdk: That is correct, Thanks for catching it. It should be `~`.

src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
```diff
@@ -121,6 +126,13 @@ static inline void autoDelay(BOOL isMove) {
             jboolean copy = JNI_FALSE;
 
             setupDone = 1;
+            // initialize CGEventFlags here - which is used in keyEvent
+            initFlags = CGEventSourceFlagsState(kCGEventSourceStateHIDSystemState);
+
+            // Clear Function flag bits if they are set
+            if ((initFlags & kCGEventFlagMaskSecondaryFn) != 0) {
+                initFlags ^= kCGEventFlagMaskSecondaryFn;
+            }
```
mrserb, line 135: Can you please double check how the lines below affected the usecase you fix, especially:
> CGEnableEventStateCombining - Pass true to specify that the actual key and mouse state are merged with the application-specified state in a synthetic event; otherwise, pass false.

honkar-jdk: @mrserb I checked by changing `CGEnableEventStateCombining()` to `false` with the updated code and it didn't  make any difference. Works the same when I test it with the reproducer. I will check again to see if there is any difference in the final flag bits that is used to post the event.

Did you also mean trying out different options available for the following lines with the original code intact?

```
// Let our event's modifier key state blend with local hardware events
CGEnableEventStateCombining(TRUE);

// Don't let our events block local hardware events
CGSetLocalEventsFilterDuringSupressionState(
                            kCGEventFilterMaskPermitAllEvents,
                            kCGEventSupressionStateSupressionInterval);

CGSetLocalEventsFilterDuringSupressionState(
                            kCGEventFilterMaskPermitAllEvents,
                            kCGEventSupressionStateRemoteMouseDrag);
```

While going through Apple documentation about [CGEnableEventStateCombining(boolean doCombineState)](https://developer.apple.com/documentation/coregraphics/1541787-cgenableeventstatecombining?language=objc) it states:

_When called with doCombineState equal to false, this function initializes local (per application) state tracking information to a state of all keys, modifiers, and mouse buttons up. When called with doCombineState equal to true, the current global state of keys, modifiers, and mouse buttons are used in generating events._

So I think CGEnableEventStateCombining should be set to TRUE as it is configured now ?

test/jdk/java/awt/Robot/RobotModifierMaskTest.java
honkar-jdk: @aivanov-jdk Will be restructuring test case using Predicate as suggested.

src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
mrserb: I just interested how this change might affect that props.

mrserb:
Just to clarify an idea of this change. We would like to maintain the "flag" keys pressed by the user and maintain the list of current flags. So if the user will press SHIFT key, we will use that modifier for all next keys until shift will be released.

It looks similar to how the mouseMove is implemented in CRobot, where we maintain the mouseLastX/mouseLastY, and we ignore the current location of the mouse for clicks, but use the latest saved value.

Probably we can do the same on the java side in this patch? it will simplify the native code a bit.

BTW do we need to set all this modifiers for the mouse clicks as well?

honkar-jdk:
@mrserb 

> Just to clarify an idea of this change. We would like to maintain the "flag" keys pressed by the user and maintain the list of current flags. So if the user will press SHIFT key, we will use that modifier for all next keys until shift will be released.

That is correct.

> It looks similar to how the mouseMove is implemented in CRobot, where we maintain the mouseLastX/mouseLastY, and we ignore the current location of the mouse for clicks, but use the latest saved value.


> Probably we can do the same on the java side in this patch? it will simplify the native code a bit.

Yes, the code for mouseEvent looks similar to what we are trying to achieve for keyEvent but there are a few differences too-                 
- `mouseEvent(int lastX, int lastY, int buttonsState, boolean isButtonsDownState, boolean isMouseMove)`,  lastX  & lastY are already part of the parameters being sent from Java to native side.

- `keyEvent(int javaKeyCode, boolean keydown)` - Are you suggesting that a modifier flag state be sent from Java to native side? This would involve changing the present native `keyEvent` signature and adding code to native side to evaluate the new modifier flag state from Java and then map it to the corresponding [kCGEventFlagMask](https://developer.apple.com/documentation/coregraphics/cgeventflags/kcgeventflagmaskalphashift?language=objc).

Having everything on native side seems to be easier because with the present fix, the equivalent modifier flag mask is obtained using javaKeyCode + key state and a quick `NSDictionary modifierKeyToMaskMap` lookup without the need of adding or changing  keyEvent signature. 

> BTW do we need to set all this modifiers for the mouse clicks as well?

I don't think so as the issue (details below) occurs only when flagState is obtained using - `CGEventSourceFlagsState(kCGEventSourceStateHIDSystemState)` which was the case for keyEvents.

**Issue:** 
_With the present code, the issue occurs at [CRobot.m#L295](https://github.com/openjdk/jdk/blob/ac6af6a64099c182e982a0a718bc1b780cef616e/src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m#L295.) The flags gets reset or cleared when mouse is moved physically in unison with Robot's key events. The physical mouse movement causes the event flags to be reset._

mrserb:
>I don't think it is applicable for mouse clicks, as the issue (details below) occurs only when flagState is obtained using - CGEventSourceFlagsState(kCGEventSourceStateHIDSystemState) which was the case for keyEvents. I can double-check on this again.

yes, please.
if the physical mouse movement causes the "event flags" to be reset, then what about  "event flags" for mouse events

honkar-jdk:
@mrserb 

> if the physical mouse movement causes the "event flags" to be reset, then what about "event flags" for mouse events

Robot's mouse events are handled differently in comparison to keyEvents , mouse events are processed by going through `CRobot_mouseEvent` > `HandleRobotButton`  > `PostMouseEvent` within CRobot.m. Current flag state of the event source is never queried for mouseEvents, so the issue of resetting of event flag does not occur. Moreover `initFlags` which is queried in initRobot is used within `CRobot_keyEvent` only.

Additionally, I believe concurrent external manual mouse movement is not supported when Robot is performing mouse events. This would produce erroneous result. For example, running MouseLocationOnScreen.java test or any location related test with Robot + simultaneous external mouse movement will fail. So the question of handling physical external mouse movements during Robot's mouse events looks to be invalid or not applicable here.

aivanov-jdk:
> Additionally, I believe concurrent external manual mouse movement is not supported when Robot is performing mouse events. This would produce erroneous result.

I don't think it's correct to say that such a situation is *not supported*. From this point of view, the situation described in the bug is not supported either. Yet you are fixing this scenario.

If the user moves the mouse when Robot generates mouse events, it is expected to produce erroneous results because the mouse ends up in the wrong position when Robot issues further events — essentially it breaks the assumptions in the test.

So, we should check whether the internal flags passed to the generated events remain valid. But it could be not as easy to verify as in this case.

Are mouse events affected if the user types text on the keyboard while Robot is moving the mouse? It's a mirrored scenario to the one described in the bug. The position of mouse remains intact, yet flags may change by concurrent keyboard input.

test/jdk/java/awt/Robot/RobotModifierMaskTest.java
```diff
@@ -0,0 +1,286 @@
...
+
+            if (condition) {
+                errorLog.append(prefixText + "Actual and Expected results differ"
+                        + " Expected : " + expectedResult
+                        + " Actual : " + actualResult + "\n");
+            }
+        });
+    }
+
+    private static class CustomWindowAdapter extends WindowAdapter {
```
aivanov-jdk, line 277:
```suggestion
    private static class CloseWindowHandler extends WindowAdapter {
```

honkar-jdk:
> If the user moves the mouse when Robot generates mouse events, it is expected to produce erroneous results because the mouse ends up in the wrong position when Robot issues further events — essentially it breaks the assumptions in the test.

Correct in case of Robot's mouse events + simultaneous manual/user mouse movement it leads to wrong position which causes tests to fail. But the case being fixed here consists of Robot keyEvents (NOT Robot's mouseEvents) + simultaneous  manual/user mouse movement.

> So, we should check whether the internal flags passed to the generated events remain valid. But it could be not as easy to verify as in this case.
> 
> Are mouse events affected if the user types text on the keyboard while Robot is moving the mouse? It's a mirrored scenario to the one described in the bug. The position of mouse remains intact, yet flags may change by concurrent keyboard input.

I see your point. I think based on what @mrserb and you have discussed, I see 3 cases here 

1. Robot mouse events + external manual mouse move- In this case user mouse events will cause the mouse to be at wrong position thus causing erroneous result as expected. And this is not a valid use case.

2. Robot key events + external manual mouse move - This is the case being fixed for accessibility keyboard use case.

3. Robot mouse events + external manual key events - As you mentioned this is a mirrored scenario and based on initial testing it works as expected - tested by automating mouse moves with Robot (without clicks or drag) and external key presses. I'll have to check the flag state in this case - at the starting of the test and at the end.

aivanov-jdk:
> I see your point. I think based on what @mrserb and you have discussed, I see 3 cases here:
> 
> 1. Robot mouse events + external manual mouse move- In this case user mouse events will cause the mouse to be at wrong position thus causing erroneous result as expected. And this is not a valid use case.

What I mean is to look further whether this situation affects any internal state of Robot so that wrong mouse events are posted.

It could still be a possible scenario. Consider Robot is moving mouse slowly, the user presses a mouse button; the following events generated by the robot should indicate the mouse button is pressed until the user releases the mouse button.

There's also the **fourth possibility**: Robot generates keyboard events while the user presses modifier keys on the keyboard. It's an invalid scenario for tests but not impossible. Should the robot combine its internal state with the global hardware state?

honkar-jdk:
> What I mean is to look further whether this situation affects any internal state of Robot so that wrong mouse events are posted.
> 
> It could still be a possible scenario. Consider Robot is moving mouse slowly, the user presses a mouse button; the following events generated by the robot should indicate the mouse button is pressed until the user releases the mouse button.
> 
> There's also the **fourth possibility**: Robot generates keyboard events while the user presses modifier keys on the keyboard. It's an invalid scenario for tests but not impossible. Should the robot combine its internal state with the global hardware state?

@aivanov-jdk Makes sense. Thanks for clarifying. I'll verify flag state for the cases discussed above and that we are not making any unexpected state changes.

src/java.desktop/macosx/native/libawt_lwawt/awt/CRobot.m
```diff
@@ -291,12 +303,25 @@ static inline void autoDelay(BOOL isMove) {
...
+            int flagMaskValue = GetCGKeyMask(keyCode);
+            if (OSX_Undefined != flagMaskValue) {
+                if (keyCode == OSX_CapsLock) {
+                    if (keyPressed) {
+                        initFlags ^= flagMaskValue;
+                    }
+                } else {
+                    initFlags = keyPressed
+                                ? (initFlags | flagMaskValue)    // add flag bits if modifier key pressed
+                                : (initFlags & ~flagMaskValue);  // clear flag bits if modifier key released
```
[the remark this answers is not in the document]
honkar-jdk: Within keyEvents `initFlags` is appended with or cleared off modifier flag bits, as a result `initFlag` value before and at the end of the test will remain same.

To make sure only necessary flag bits are modified and we are not introducing new flag bits or clearing out something that is already present or known to the system, the `initFlag` is not used directly in `CGEventSetFlags` but it is copied over to a local `flag` variable as follows which is then used in `CGEventSetFlags` API and post the event.

```
 CGEventFlags flags = CGEventSourceFlagsState(kCGEventSourceStateHIDSystemState);
 flags = (initFlags & allModifiersMask) | (flags & (~allModifiersMask));
```
The second line copies over all the flag bits corresponding to modifier flags (if present in initFlags) and all the bits already present.

test/jdk/java/awt/Robot/RobotModifierMaskTest.java
```diff
@@ -0,0 +1,286 @@
...
+
+        jFrame.setSize(560, 200);
+        jFrame.setLocation(200, 200);
+        jFrame.addWindowListener(new CustomWindowAdapter());
+        jFrame.setVisible(true);
+    }
+
+    private static void createTestUI() {
+        String mode = isManual ? "MANUAL" : "AUTOMATED";
+        jFrame = new JFrame("RobotModifierMaskTest - Mode: " + mode);
```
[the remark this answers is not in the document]
honkar-jdk: Since I'm reusing the same `jFrame` static variable , disposing of jFrame is handled in the finally block.

test/jdk/java/awt/Robot/RobotModifierMaskTest.java
```diff
@@ -0,0 +1,286 @@
...
+            if (isManual) {
+                countDownLatch = new CountDownLatch(1);
+                invokeAndWait(RobotModifierMaskTest::createInstructionsUI);
+                robot.waitForIdle();
+                countDownLatch.await(2, TimeUnit.MINUTES);
+            }
+
+            if (!testStarted) {
+                throw new RuntimeException("Test Failed: Manual test timed out!!");
+            }
```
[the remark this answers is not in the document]
honkar-jdk: Updated.

test/jdk/java/awt/Robot/RobotModifierMaskTest.java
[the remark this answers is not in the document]
honkar-jdk: Added try & finally block for instruction frame too.

honkar-jdk:
@mrserb @aivanov-jdk Based on previous suggestions, I have tested out few cases 

Following table list all the cases tested, along with expected & actual result in case of original and proposed solution.

- Case 1-5 are the ones already being tested as part of RobotModifierMaskTest.java
- Case 2 results in wrong output with original code (without mouse move).
- Case 6 is the mirrored case mentioned by @aivanov-jdk and I have verified that we are not setting or introducing new flag states here.
- Case 7-8 involve concurrent manual/ user + Robot interaction. This is bound to produce invalid or erroneous result due to simultaneous manual + Robot key events. With the proposed fix, the state of keyboard at time of initialization of Robot is taken into account and after initialization of Robot any modifier keys pressed by the user are ignored during Robot's activity. Hence we see 'A' for case 7 and 'a' for case 8-9 with proposed fix. I'm not completely sure if these cases are something we need to address since they fall under invalid scenarios for Robot testing/usage. @azuev-java can you please advice on cases 7-9.
 
So to answer previous question -

> There's also the fourth possibility: Robot generates keyboard events while the user presses modifier keys on the keyboard. It's an invalid scenario for tests but not impossible. Should the robot combine its internal state with the global hardware state?

With proposed fix we filter out any external (manual) modifier key events when Robot is typing and consider only the global hardware state only during initialization of Robot. 


| Case | Original Code (No External Mouse move) | With Proposed Fix (Without & With MouseMove) | Expected |
| --- | --- | --- | --- |
| 1. VK_SHIFT + VK_A - 3 times |  AAA |  AAA | AAA |
| 2. VK_CAPS + VK_A - 5 times <br> (caps lock toggled) | **aaaaa** | AaAaA| AaAaA| 
| 3. VK_META + VK_V - Copy clipboard content |  AAA |  AAA | AAA |
| 4. VK_ALT + VK_A - 3 times |  ååå | ååå | ååå |
| 5. VK_CONTROL + VK_A |  Caret position 0 |  Caret position 0 | Moves the caret to the start of the current line. |
| 6. Mirrored case -  Mouse moves triggered by Robot + concurrent external manual key events for above cases 1-5 |  Same result as above except case 2 - AaAaA | Produces same result as above. | Should work the same way as non-mirrored case - Robot key events + concurrent external manual mouse movements   |
|   |   |   |   |
| 7. VK_SHIFT is pressed & released at start of the application manually, before Robot types VK_A |  aaa | AAA | ??? |
| 8. VK_SHIFT is toggled manually after initialization of Robot, while Robot types VK_A - 5 times | AaAaA | aaaaa | ??? |
| 9. VK_CAPS is toggled manually after initialization of Robot, while Robot types VK_A - 5 times |  AaAaA | aaaaa | ??? |

honkar-jdk:
@mrserb 

> BTW do we need to set all this modifiers for the mouse clicks as well?

I wanted to clarify if you mean mouse clicks generated externally or Robot here ?

honkar-jdk:
Issue reproducible with native app. Bug reported to Apple - CGEventSourceFlagsState() API issue. This ongoing PR will be converted to draft for now. Fix in JDK (if any), will be made after hearing back on the filed bug report.

BlakeTNC:
Hello, I am the developer that originally reported this bug to the Adoptium team. I was reviewing the bug discussions on this page and elsewhere, and I have a suggestion and some questions.

### Did Apple fix this bug? (It still does not work for me on the latest macOS.)

I read this comment:
> Link: https://bugs.openjdk.org/browse/JDK-8302618
> Harshitha Onkar added a comment - 2023-12-22 14:45
> [~prr] Revisited this issue. It works fine on the latest macOS 14.2 and still fails on macOS 
> Ventura. This is true for the Java test case and the native reproducer.
> Something must have been fixed on the new version although there wasn't any update on the 
> Apple Bug report that was filed. 

I tested the bug today and I am still able to replicate it on my own computer, using the following set up: An M1 Macbook Pro running Sonoma 14.5, on JDK 19 and JDK 22 (Zulu 22.0.1+8). So, it appears that Apple might not have fixed this issue in the latest MacOS.

### Suggestion, Add unit tests for this bug to the openJDK:
It appears that this pull request includes some test code for this issue. I would suggest, at minimum, that unit testing for this issue be added to the openJDK. That way if the Robot functionality breaks in any future versions of macOS (or other operating systems), it will be caught more quickly by the openJDK developers. Apparently this issue existed for years without anyone noticing. It would be sad for the same issue to resurface in the future, without anyone noticing again. 

Note that unit tests for this issue could be added to the JDK regardless of whether any bug fix is applied in openJDK, using conditional test execution. If there is an express decision to NOT fix this issue in the JDK (and to instead rely on a fix that was or will be supplied by Apple), then if desired, the testing code could be surrounded with a conditional code block to exclude only the past version(s) of macOS that are known to be broken. That way the tests would only execute, 1) On all the existing and past operating system versions where Robot is currently known or expected to work correctly, and 2) On all future operating system releases.

### Option, Apply a bug fix in the openJDK, since Apple has not explicitly responded:

It may be useful to apply a fix for this bug in the openJDK, (if there is no detriment to doing so), rather than relying solely on any fix that Apple may (or may not) release. Based on Harshitha Onkar's comment copied above, I have an impression that Apple has not yet responded to this issue. I also learned today that macOS Ventura is the last macOS system that many older Mac models can run. That is because reportedly[[1]](https://www.techradar.com/computing/mac-os/the-latest-macos-ventura-update-has-left-owners-of-old-macs-stranded-in-a-sea-of-problems-raising-a-chorus-of-complaints) [[2]](https://en.wikipedia.org/wiki/MacOS_Sonoma#Supported_hardware), macOS Sonoma only supports Mac computer models from 2018 onwards. So if this bug is not fixed in the JDK, then it will likely never be fixed for many older Mac models. Also as I mentioned above, this bug still exist for me even though I am currently running macOS Sonoma 14.5. If I understand correctly, that means it's possible that Apple has not fixed the problem for the latest versions of macOS.

### Thanks

Thank you to everyone who worked on this issue. :)

### Just for reference, here are links to some pages discussing this bug:
* This pull request.
https://github.com/openjdk/jdk/pull/14744

* Bug Report at OpenJDK.
https://bugs.openjdk.org/browse/JDK-8302618

* Bug Report at Adoptium.
https://github.com/adoptium/adoptium-support/issues/710

* Apple developer forums post.
https://forums.developer.apple.com/forums/thread/733161 

* An old Stack Overflow report, 2018:
https://stackoverflow.com/questions/49050543/java-robot-type-uppercase-lowercase-and-symbols-from-string

karianna:
I've reopened the bug in OpenJDK.

honkar-jdk:
@BlakeTNC I was able to reproduce this issue again on macOS 14.5. Currently I'm looking into this PR and revisiting all the previous discussions. I'll have an update for all the posted questions. Thanks!

BlakeTNC:
PS, There is an Apple Developer forum post here: (https://forums.developer.apple.com/forums/thread/733161) where someone tried to report this issue to Apple on that forum. In the same thread, Apple suggested reporting the issue using a different channel called the Apple Feedback Assistant. Apparently that is an official channel for reporting Apple bugs.
Feedback Assistant instructions page: https://developer.apple.com/bug-reporting/
Feedback Assistant link: https://feedbackassistant.apple.com/

It is unclear to me if anyone has reported the bug using the Apple Feedback Assistant. I could send a report to Apple myself, but I may not be the best person to file the report. I would have trouble giving them all the needed information because I am not familiar with the topics of Objective-C, the CGEventSourceFlagsState API, or the event flags that have been mentioned as potentially being at the root of the issue.

honkar-jdk:
@BlakeTNC Thank you for your patience. Please find the answers below.

> Did Apple fix this bug? (It still does not work for me on the latest macOS.)

This issue is reproducible on macOS 14.5 although it did seem to work on macOS 14.2.
Along with Feedback Assistant report, a Technical Support Incident was raised. Received a reply from Apple - we are in the discussion stage, so it might take sometime to understand different scenarios and formulate a solution that works well in all cases.

> Option, Apply a bug fix in the openJDK, since Apple has not explicitly responded:
> 
> It may be useful to apply a fix for this bug in the openJDK, (if there is no detriment to doing so), rather than relying solely on any fix that Apple may (or may not) release. Based on Harshitha Onkar's comment copied above, I have an impression that Apple has not yet responded to this issue. I also learned today that macOS Ventura is the last macOS system that many older Mac models can run. That is because reportedly[[1]](https://www.techradar.com/computing/mac-os/the-latest-macos-ventura-update-has-left-owners-of-old-macs-stranded-in-a-sea-of-problems-raising-a-chorus-of-complaints) [[2]](https://en.wikipedia.org/wiki/MacOS_Sonoma#Supported_hardware), macOS Sonoma only supports Mac computer models from 2018 onwards. So if this bug is not fixed in the JDK, then it will likely never be fixed for many older Mac models. Also as I mentioned above, this bug still exist for me even though I am currently running macOS Sonoma 14.5. If I understand correctly, that means it's possible that Apple has not fixed the problem for the latest versions of macOS.

The proposed fix in this PR works well only in some cases. More details documented here https://github.com/openjdk/jdk/pull/14744#issuecomment-1662976048. So adding the proposed fix can cause unexpected results in other cases. Since this scenario is reproducible with a native reproducer too (using Apple APIs), it is important to understand if it needs to be fixed on their end or is this an inherent API behavior before proceeding with a solution.

Now that we have received a reply from Apple, this scenario and native API working can be discussed in depth.

BlakeTNC:
Hello,
Has Apple been able to help with this?

honkar-jdk:
@BlakeTNC 

> Hello,
> Has Apple been able to help with this?

Still in discussion phase as of now.

PR #14745, opened 2023-07-01
8311207: Cleanup for Optimization for UUID.toString

wenshao:
[PR 14578 ](https://github.com/openjdk/jdk/pull/14578) still has unresolved discussions, continue to make improvements.

# Benchmark Result

```
sh make/devkit/createJMHBundle.sh
bash configure --with-jmh=build/jmh/jars
make test TEST="micro:java.util.UUIDBench.toString"
```

## 1. [aliyun_ecs_c8i.xlarge](https://help.aliyun.com/document_detail/25378.html#c8i)
* cpu : intel xeon sapphire rapids (x64)

``` diff
-Benchmark           (size)   Mode  Cnt   Score   Error   Units (baseline)
-UUIDBench.toString   20000  thrpt   15  62.019 ± 0.622  ops/us

+Benchmark           (size)   Mode  Cnt   Score   Error   Units
+UUIDBench.toString   20000  thrpt   15  82.998 ± 0.739  ops/us (+33.82%)
```

## 2. [aliyun_ecs_c8a.xlarge](https://help.aliyun.com/document_detail/25378.html#c8a)
* cpu : amd epc genoa (x64)

``` diff
-Benchmark           (size)   Mode  Cnt   Score   Error   Units (baseline)
-UUIDBench.toString   20000  thrpt   15  88.668 ± 0.672  ops/us

+Benchmark           (size)   Mode  Cnt   Score   Error   Units
+UUIDBench.toString   20000  thrpt   15  89.229 ± 0.271  ops/us (+0.63%)
```

## 3. [aliyun_ecs_c8y.xlarge](https://help.aliyun.com/document_detail/25378.html#c8y)
* cpu : aliyun yitian 710 (aarch64)
``` diff
-Benchmark           (size)   Mode  Cnt   Score   Error   Units (baseline)
-UUIDBench.toString   20000  thrpt   15  49.382 ± 2.160  ops/us

+Benchmark           (size)   Mode  Cnt   Score   Error   Units
+UUIDBench.toString   20000  thrpt   15  49.636 ± 1.974  ops/us (+0.51%)
```

## 4. MacBookPro M1 Pro
``` diff
-Benchmark           (size)   Mode  Cnt    Score   Error   Units (baseline)
-UUIDBench.toString   20000  thrpt   15  103.543 ± 0.963  ops/us

+Benchmark           (size)   Mode  Cnt    Score   Error   Units
+UUIDBench.toString   20000  thrpt   15  110.976 ± 0.685  ops/us (+7.17%)
```

## 5. Orange Pi 5 Plus

``` diff
-Benchmark           (size)   Mode  Cnt   Score   Error   Units (baseline)
-UUIDBench.toString   20000  thrpt   15  33.532 ± 0.396  ops/us

+Benchmark           (size)   Mode  Cnt   Score   Error   Units (PR)
+UUIDBench.toString   20000  thrpt   15  33.054 ± 0.190  ops/us (-4.42%)
```

---------
### Progress

### Issue
 * [JDK-8311207](https://bugs.openjdk.org/browse/JDK-8311207): Cleanup for Optimization for UUID.toString (**Enhancement** - P4)

### Reviewers
 * [Chen Liang](https://openjdk.org/census#liach) (@liach - Author) ⚠️ Review applies to [e5274ddc](https://git.openjdk.org/jdk/pull/14745/files/e5274ddcabc6bfb6b9230e570793664b9dfc2c58)
 * [Roger Riggs](https://openjdk.org/census#rriggs) (@RogerRiggs - **Reviewer**) ⚠️ Review applies to [384354d9](https://git.openjdk.org/jdk/pull/14745/files/384354d9929e3e938c9cf66379cff14e1bef8ef3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14745/head:pull/14745` \
`$ git checkout pull/14745`

`$ git checkout pull/14745` \
`$ git pull https://git.openjdk.org/jdk.git pull/14745/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14745`

`$ git pr show -t 14745`

</details>

<a href="https://git.openjdk.org/jdk/pull/14745.diff">https://git.openjdk.org/jdk/pull/14745.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14745#issuecomment-1615350407)


liach:
Is using `Unsafe` directly consistently faster than using ByteArray? It should have similar performance as ByteArray's VarHandle is simply a wrapper around Unsafe's put/get methods.

src/java.base/share/classes/java/util/HexDigits.java
```diff
@@ -105,14 +105,14 @@ private HexDigits() {
     /**
      * Combine two hex shorts into one int based on big endian
      */
-    static int digit(int b0, int b1) {
+    static int packDigits(int b0, int b1) {
```
liach, line 108: I recommend such a refactor:
```java
/**
 * Return a big-endian packed integer for the 4 ASCII bytes for an input unsigned short value.
 */
static int packDigits16(int b) {
    return (DIGITS[(b >> 8) & 0xff] << 16) | DIGITS[b & 0xff];
}
```
And all your usages can be `HexDigits.packDigits16(((int) msb) >> 16)` and `HexDigits.packDigits16((int) msb)` without the `>> 24` and `>> 8`

src/java.base/share/classes/java/util/HexDigits.java
```diff
@@ -105,14 +105,14 @@ private HexDigits() {
...
         return (DIGITS[b0 & 0xff] << 16) | DIGITS[b1 & 0xff];
     }
 
     /**
      * Combine four hex shorts into one long based on big endian
      */
-    static long digit(int b0, int b1, int b2, int b3) {
+    static long packDigits(int b0, int b1, int b2, int b3) {
```
liach, line 115: Same comment here:

```java
/**
 * Return a big-endian packed long for the 8 ASCII bytes for an input unsigned int value.
 */
static int packDigits32(int b) {
    return (((long) DIGITS[(b >> 24) & 0xff]) << 48)
            | (((long) DIGITS[(b >> 16) & 0xff]) << 32)
            | (DIGITS[(b >> 8) & 0xff] << 16)
            | DIGITS[b & 0xff];
}
```
And use sites become:
```java
// old
HexDigits.packDigits((int) (msb >> 56), (int) (msb >> 48), (int) (msb >> 40), (int) (msb >> 32))
// new
HexDigits.packDigits32((int) (msb >> 32))
```
and `HexDigits.packDigits32((int) (lsb >> 16))`

wenshao:
> Is using `Unsafe` directly consistently faster than using ByteArray? It should have similar performance as ByteArray's VarHandle is simply a wrapper around Unsafe's put/get methods.

Using Unsafe on aliyun_ecs_c8i.xlarge and MacBookPro M1 Pro is faster than ByteArray, and I haven't figured out why

wenshao:
> Is using `Unsafe` directly consistently faster than using ByteArray? It should have similar performance as ByteArray's VarHandle is simply a wrapper around Unsafe's put/get methods.

Using Unsafe on aliyun_ecs_c8i.xlarge and MacBookPro M1 Pro is faster than ByteArray, and I haven't figured out why

src/java.base/share/classes/java/util/HexDigits.java
wenshao: but it's return an int value

liach:
> > Is using `Unsafe` directly consistently faster than using ByteArray? It should have similar performance as ByteArray's VarHandle is simply a wrapper around Unsafe's put/get methods.
> 
> Using Unsafe on aliyun_ecs_c8i.xlarge and MacBookPro M1 Pro is faster than ByteArray, and I haven't figured out why

Then it's probably VarHandle's overhead. No worries; your change to use Unsafe is totally fine.

Meanwhile, can you enable GitHub actions on your fork, so it can detect compile and test errors? Like this:
image

src/java.base/share/classes/java/util/HexDigits.java
wenshao: but it's return a long value

liach: My fault, fixed the return type in my code snippet.

src/java.base/share/classes/java/util/HexDigits.java
liach: The current code snippet still returns a int value... I think I can probably make a patch and demonstrate.

wenshao:
image

it's enabled

liach:
@wenshao I have made my suggestions into a patch for you: https://github.com/wenshao/jdk/pull/1
Feel free to review.

wenshao:
> @wenshao I have made my suggestions into a patch for you: [wenshao#1](https://github.com/wenshao/jdk/pull/1) Feel free to review.

your version performance is a bit worse, if i can't find the reason, I will restore the previous version

liach:
> > @wenshao I have made my suggestions into a patch for you: [wenshao#1](https://github.com/wenshao/jdk/pull/1) Feel free to review.
> 
> your version performance is a bit worse, if i can't find the reason, I will restore the previous version

Can you post the benchmark results? And do you have like a baseline for the benchmarks, as there may be other factors that affect performance from run to run?

Also, my fault for actions: can you go to your actions tab and enable actions for your fork like this?
image

wenshao:
@liach Your version performance is a bit worse, If I can't find the reason, I will revert to the previous version

wenshao:
# Update 02
current version [76dba964](https://git.openjdk.org/jdk/pull/14745/files/76dba9649a530a2f21574f3dff1aa55a6ebe8888)
```
Benchmark           (size)   Mode  Cnt    Score   Error   Units 
UUIDBench.toString   20000  thrpt   15  104.262 ± 2.199  ops/us
```

# Update 01
your version [425559f6](https://git.openjdk.org/jdk/pull/14745/files/425559f65a3e5f72323abdadbbf9df426e196d9b)

```
Benchmark           (size)   Mode  Cnt   Score   Error   Units
UUIDBench.toString   20000  thrpt   15  81.622 ± 0.194  ops/us
```

wenshao:
@liach I guess your version is slower because it doesn't support out-of-order execution.

src/java.base/share/classes/java/util/UUID.java
```diff
@@ -466,37 +466,44 @@ public long node() {
      */
     @Override
     public String toString() {
+        Unsafe unsafe = Unsafe.getUnsafe();
```
ExE-Boss, line 469: This should probably be stored in a `private static final` field.

liach (approved):

liach:
For your information, merging a patch requires a reviewer's approval. They are mostly available during weekdays only. If you wish to have a patch integrated, don't regularly change it so we don't integrate something that's WIP. You can tell everyone that your patch is in shape and won't change unless there's reviews or critical problems.

Also, can you go to Actions tab on your fork repository to enable workflows in your fork repository (green button)? Only allowing actions in settings is not sufficient.

liach:
> /integrate

A "reviewer" means a role in the jdk project in the OpenJDK organization: https://openjdk.org/census#jdk

src/java.base/share/classes/java/util/HexDigits.java
```diff
@@ -103,16 +103,20 @@ private HexDigits() {
...
      */
-    static int digit(int b0, int b1) {
+    static int packDigits(int b0, int b1) {
         return (DIGITS[b0 & 0xff] << 16) | DIGITS[b1 & 0xff];
     }
 
     /**
-     * Combine four hex shorts into one long based on big endian
+     * Return a big-endian packed long for the 8 ASCII bytes for an input unsigned 4-byte integer.
+     * {@code b0} is the most significant byte and {@code b4} is the least significant byte.
```
RogerRiggs, line 116: Typo "b4" -> "b3".

src/java.base/share/classes/java/util/HexDigits.java
```diff
@@ -103,16 +103,20 @@ private HexDigits() {
     }
 
     /**
-     * Combine two hex shorts into one int based on big endian
+     * Return a big-endian packed integer for the 4 ASCII bytes for an input unsigned 2-byte integer.
+     * {@code b0} is the most significant byte and {@code b1} is the least significant byte.
+     * The integer is passed byte-wise to allow reordering of execution.
```
RogerRiggs, line 108: The more natural API would be to pass the 16-bit value as an int (or short). Simplifying the callers.
Is there a significant performance difference?
The extraction of the 8-bit values could be done here instead of the caller and still allow hardware specific instruction optimization.

Also below, pass the 32-bit value as an int.

src/java.base/share/classes/java/util/UUID.java
```diff
@@ -469,34 +472,40 @@ public String toString() {
         long lsb = leastSigBits;
         long msb = mostSigBits;
         byte[] buf = new byte[36];
-        ByteArray.setLong(
+        UNSAFE.putLongUnaligned(
```
RogerRiggs, line 475: I understood that since UUID was not used during startup, that using VarHandles is the preferred abstraction.

src/java.base/share/classes/java/util/HexDigits.java
```diff
@@ -39,27 +39,27 @@ final class HexDigits implements Digits {
      * Each element of the array represents the ascii encoded
      * hex relative to its index, for example:<p>
      * <pre>
-     *       0 -> '00' -> ('0' << 8) | '0' -> 12336
-     *       1 -> '01' -> ('0' << 8) | '1' -> 12337
-     *       2 -> '02' -> ('0' << 8) | '2' -> 12338
+     *       0 -> '00' -> ('0' << 8) | '0' -> 0x3030
+     *       1 -> '01' -> ('0' << 8) | '1' -> 0x3130
+     *       2 -> '02' -> ('0' << 8) | '2' -> 0x3230
```
RogerRiggs, line 44: The description of the array content does not need to be exhaustive; just sufficient to understand how it is indexed and the meaning of the values.

src/java.base/share/classes/java/util/HexDigits.java
liach: > Is there a significant performance difference?

Yes, see https://github.com/openjdk/jdk/pull/14745#issuecomment-1615435719 for the revision at https://github.com/openjdk/jdk/pull/14745/commits/425559f65a3e5f72323abdadbbf9df426e196d9b
On my device, the UUIDBench.toString has a smaller difference of about 8%, but it is still noticeable.

Unfortunately, I don't know architecture well enough to explain this difference.

src/java.base/share/classes/java/util/UUID.java
wenshao: Using VarHandle will be 10% slower than using Unsafe directly, I haven't found the reason yet

wenshao: The comment is already there, I think the value on the right is easier to understand in hexadecimal

src/java.base/share/classes/java/util/HexDigits.java
wenshao: The comment is already there, I think the value on the right is easier to understand in hexadecimal

src/java.base/share/classes/java/util/UUID.java
ExE-Boss: I believe it’s because `VarHandle`s do bounds checking:
https://github.com/openjdk/jdk/blob/581f90e242b8a943215a223189d171b7ede37785/src/java.base/share/classes/java/lang/invoke/X-VarHandleByteArrayView.java.template#L113-L116 https://github.com/openjdk/jdk/blob/581f90e242b8a943215a223189d171b7ede37785/src/java.base/share/classes/java/lang/invoke/X-VarHandleByteArrayView.java.template#L144-L161

wenshao:
Is anyone working on this PR?

wenshao:
@RogerRiggs What changes do I need to make to start the integration?

src/java.base/share/classes/java/util/HexDigits.java
wenshao: comment has been removed

src/java.base/share/classes/java/util/UUID.java
```diff
@@ -469,34 +468,34 @@ public String toString() {
...
-                HexDigits.digit((int) (lsb >> 56), (int) (lsb >> 48)));
+                HexDigits.packDigits((int) (lsb >> 56), (int) (lsb >> 48)));
         buf[23] = '-';
-        ByteArray.setLong(
+        ByteArrayLittleEndian.setLong(
                 buf,
                 24,
-                HexDigits.digit(((int) (lsb >> 40)), (int) (lsb >> 32), ((int) lsb) >> 24, ((int) lsb) >> 16));
-        ByteArray.setInt(
+                HexDigits.packDigits(((int) (lsb >> 40)), (int) (lsb >> 32), ((int) lsb) >> 24, ((int) lsb) >> 16));
```
liach, line 494:
```suggestion
                HexDigits.packDigits((int) (lsb >> 40), (int) (lsb >> 32), ((int) lsb) >> 24, ((int) lsb) >> 16));
```

src/java.base/share/classes/java/util/HexDigits.java
```diff
@@ -103,20 +65,24 @@ private HexDigits() {
...
      */
-    static int digit(int b0, int b1) {
-        return (DIGITS[b0 & 0xff] << 16) | DIGITS[b1 & 0xff];
+    static int packDigits(int b0, int b1) {
+        return DIGITS[b0 & 0xff] | (DIGITS[b1 & 0xff] << 16);
     }
 
     /**
-     * Combine four hex shorts into one long based on big endian
+     * Return a big-endian packed long for the 8 ASCII bytes for an input unsigned 4-byte integer.
```
liach, line 77:
```suggestion
     * Returns a little-endian packed long for the 8 ASCII bytes for an input unsigned 4-byte integer.
```

src/java.base/share/classes/java/util/HexDigits.java
```diff
@@ -103,20 +65,24 @@ private HexDigits() {
     }
 
     /**
-     * Combine two hex shorts into one int based on big endian
+     * Return a big-endian packed integer for the 4 ASCII bytes for an input unsigned 2-byte integer.
```
liach, line 68:
```suggestion
     * Returns a little-endian packed integer for the 4 ASCII bytes for an input unsigned 2-byte integer.
```

wenshao:
@cl4es Can you help me review this PR?

src/java.base/share/classes/java/util/HexDigits.java
```diff
@@ -85,10 +47,10 @@ final class HexDigits implements Digits {
         short[] digits = new short[16 * 16];
 
         for (int i = 0; i < 16; i++) {
-            short hi = (short) ((i < 10 ? i + '0' : i - 10 + 'a') << 8);
+            short hi = (short) (i < 10 ? i + '0' : i - 10 + 'a');
 
             for (int j = 0; j < 16; j++) {
-                short lo = (short) (j < 10 ? j + '0' : j - 10 + 'a');
+                short lo = (short) ((j < 10 ? j + '0' : j - 10 + 'a') << 8);
                 digits[(i << 4) + j] = (short) (hi | lo);
```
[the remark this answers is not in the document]
RogerRiggs: > A description of the DIGITS array layout and indexing is needed; it will help with the comprehension of the masking and shifts in the packDigits methods.

The DIGITS table is missing a description of its indexing.

wenshao:
Is it good enough to integrate now?

RogerRiggs (changes requested):
This request got buried in a "resolved" comment:

> A description of the DIGITS array layout and indexing is needed; it will help with the comprehension of the masking and shifts in the packDigits methods.

The DIGITS table is missing a description of its indexing.

wenshao:
> This request got buried in a "resolved" comment:
> 
> > A description of the DIGITS array layout and indexing is needed; it will help with the comprehension of the masking and shifts in the packDigits methods.
> 
> The DIGITS table is missing a description of its indexing.

Sorry, I have already added the description of DIGITS

src/java.base/share/classes/java/util/HexDigits.java
```diff
@@ -85,11 +74,11 @@ final class HexDigits implements Digits {
...
 
         for (int i = 0; i < 16; i++) {
-            short hi = (short) ((i < 10 ? i + '0' : i - 10 + 'a') << 8);
+            short lo = (short) (i < 10 ? i + '0' : i - 10 + 'a');
 
             for (int j = 0; j < 16; j++) {
-                short lo = (short) (j < 10 ? j + '0' : j - 10 + 'a');
-                digits[(i << 4) + j] = (short) (hi | lo);
+                short hi = (short) ((j < 10 ? j + '0' : j - 10 + 'a') << 8);
+                digits[(i << 4) + j] = (short) (lo | hi);
```
RogerRiggs, line 81: The part that is counter-intuitive is that the bits that are shifted left appear to the right of the bits that are not shifted. (the actual bits are correct, either way, but it reads oddly). In a short, the resulting bits would be `hhll`.
```suggestion
                digits[(i << 4) + j] = (short) (hi | lo);
```
Similarly, the returns of the other `packDigits` methods do not read MSB to LSB.

RogerRiggs (approved):
Thank you improvements.

src/java.base/share/classes/java/util/HexDigits.java
wenshao: You are right, I have changed

src/java.base/share/classes/java/util/HexDigits.java
```diff
@@ -35,6 +35,33 @@
...
+     *
+     *     ...
+     *
+     *     253 -> 'fd' -> 'f' | ('d' << 8) -> 0x6466
+     *     254 -> 'fe' -> 'f' | ('e' << 8) -> 0x6566
+     *     255 -> 'ff' -> 'f' | ('f' << 8) -> 0x6666
+     * </pre>
+     */
     @Stable
     private static final short[] DIGITS;
```
ExE-Boss, line 66: Maybe it should be `char[]` instead since it’s encoded using unsigned 16‑bit integers.

liach: a char would confuse people that each char represents one character, while it represents 2 in fact.

cl4es: @liach has a fair point. Either way it should be performance neutral.

src/java.base/share/classes/java/util/HexDigits.java
```diff
@@ -39,38 +39,27 @@ final class HexDigits implements Digits {
      * Each element of the array represents the ascii encoded
      * hex relative to its index, for example:<p>
      * <pre>
-     *       0 -> '00' -> ('0' << 8) | '0' -> 12336
-     *       1 -> '01' -> ('0' << 8) | '1' -> 12337
-     *       2 -> '02' -> ('0' << 8) | '2' -> 12338
+     *       0 -> '00' -> '0' | ('0' << 8) -> 0x3030
```
cl4es, line 42: AFAICT this is now a copy of `StringLatin1::PACKED_DIGITS`, which might be fine. The comment to `StringLatin1::PACKED_DIGITS` seem wrong, though? I think that needs a re-examination (in this or in a separate PR)

cl4es: Perhaps you could extract the `digitPair` method you're adding in #15555 to a separate PR and use that to remove `HexDigits::DIGITS`?

wenshao: Please help me create an issue, and I will submit a new PR to modify it.

cl4es: Filed https://bugs.openjdk.org/browse/JDK-8315968 - we should also get @JimLaskey to weigh in here (who's recently authored the `java.util.Digits` code). Tl;dr: I think it'd be good to consolidate and share the code to produce these packed digit pairs.

Also note that there might be some issues on big-endian systems with recent changes, so let's figure that out first: https://bugs.openjdk.org/browse/JDK-8310929

wenshao: https://github.com/wenshao/jdk/tree/fix_14699_big_endian

I have a [branch](https://github.com/wenshao/jdk/tree/fix_14699_big_endian) ready to fix the big endian issue and need someone to create an issue

wenshao: https://github.com/openjdk/jdk/pull/15652

I've submitted a PR to fix the big endian problem

RogerRiggs:
Given the endian-ness issues with https://git.openjdk.org/jdk/pull/14699.
I'll need to run a more complete set of tests first.

wenshao:
@TheRealMDoerr Can you help me test this PR on AIX (big endian) ?

TheRealMDoerr:
I have run a couple of tests on linux Big Endian. They have passed. So, it's probably correct. However, I can't tell if it's good to use `ByteArrayLittleEndian`. I don't really like such platform details in the Java classes. Is that necessary for better performance on x86?

wenshao:
@RogerRiggs Can it be merged now?

liach:
@TheRealMDoerr `ByteArrayLittleEndian` only means that the input long/int/short/char will be seen as little-endian when written to a byte array; do you mean that assuming little-endian writes are faster is too platform-specific?

An alternative approach tried before is to pack the digits platform-specifically and use Unsafe (which bypasses platform-endianness reversals) to write directly; I recall it was rejected before, for using unsafe directly seems... unsafe :)

TheRealMDoerr:
I think making sure C2 optimizes it would be a better approach. Java classes shouldn't be optimized for performance on any endianness version IMHO. Rather for readability.
@offamitkumar, @deepa181, @JoKern65, @TOatGithub: You may want to check performance impact on s390x and AIX.

offamitkumar:
>@offamitkumar, @deepa181, @JoKern65, @TOatGithub: You may want to check performance impact on s390x and AIX.

@TheRealMDoerr Testing on s390 is not possible for now, as build is broken due to field resolution changes.

src/java.base/share/classes/jdk/internal/util/HexDigits.java
```diff
@@ -103,20 +92,24 @@ private HexDigits() {
...
-    public static long digit(int b0, int b1, int b2, int b3) {
-        return (((long) DIGITS[b0 & 0xff]) << 48)
-                | (((long) DIGITS[b1 & 0xff]) << 32)
-                | (DIGITS[b2 & 0xff] << 16)
-                | DIGITS[b3 & 0xff];
+    public static long packDigits(int b0, int b1, int b2, int b3) {
+        return DIGITS[b0 & 0xff]
+                | (DIGITS[b1 & 0xff] << 16)
+                | (((long) DIGITS[b2 & 0xff]) << 32)
+                | (((long) DIGITS[b3 & 0xff]) << 48);
```
RogerRiggs, line 112: Can you reverse the order of these source lines to put the shifts of the higher order bits before the lower order bit shifts.  `3333222211110000`. Its easier to understand where the bits end up in the long.
The rest of the change is better focused.

wenshao: if change packDigits order, performance will be slow, I don't know why yet.

The following is the data running on MacBookPro M1 Max :
```
make test TEST="micro:java.util.UUIDBench.toString"

Benchmark           (size)   Mode  Cnt   Score   Error   Units (current order)
UUIDBench.toString   20000  thrpt   15  96.396 ? 0.946  ops/us


Benchmark           (size)   Mode  Cnt   Score   Error   Units (change packDigits order)
UUIDBench.toString   20000  thrpt   15  86.496 ? 0.542  ops/us
```

cl4es: Looks like something that might be an interesting puzzler for JIT compiler folks. Perhaps added implicit casts to long messes something up?

PR #14746, opened 2023-07-01
8311178: JMH tests don't scale well when sharing output buffers

swati-sha:
The below benchmark files have scaling issues due to cache contention and leads to poor scaling when run on multiple threads. The patch sets the scope from benchmark level to thread level to fix the issue:
- org/openjdk/bench/java/io/DataOutputStreamTest.java
- org/openjdk/bench/java/lang/ArrayCopyObject.java
- org/openjdk/bench/java/lang/ArrayFiddle.java
- org/openjdk/bench/java/time/format/DateTimeFormatterBench.java
- org/openjdk/bench/jdk/incubator/vector/IndexInRangeBenchmark.java
- org/openjdk/bench/jdk/incubator/vector/MemorySegmentVectorAccess.java
- org/openjdk/bench/jdk/incubator/vector/StoreMaskedBenchmark.java
- org/openjdk/bench/jdk/incubator/vector/StoreMaskedIOOBEBenchmark.java
- org/openjdk/bench/vm/compiler/ArrayFill.java
- org/openjdk/bench/vm/compiler/IndexVector.java

Also removing the static scope for variables in org/openjdk/bench/jdk/incubator/vector/VectorFPtoIntCastOperations.java for better scaling.

Please review and share your feedback.

Thanks,
Swati

---------
### Progress

### Issue
 * [JDK-8311178](https://bugs.openjdk.org/browse/JDK-8311178): JMH tests don't scale well when sharing output buffers (**Enhancement** - P4)

### Reviewers
 * [Eric Caspole](https://openjdk.org/census#ecaspole) (@ericcaspole - Committer)
 * [Sandhya Viswanathan](https://openjdk.org/census#sviswanathan) (@sviswa7 - **Reviewer**)
 * [Hamlin Li](https://openjdk.org/census#mli) (@Hamlin-Li - **Reviewer**)

### Contributors
 * Vladimir Ivanov `<vaivanov@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14746/head:pull/14746` \
`$ git checkout pull/14746`

`$ git checkout pull/14746` \
`$ git pull https://git.openjdk.org/jdk.git pull/14746/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14746`

`$ git pr show -t 14746`

</details>

<a href="https://git.openjdk.org/jdk/pull/14746.diff">https://git.openjdk.org/jdk/pull/14746.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14746#issuecomment-1615621299)

### DIFF
--- a/test/micro/org/openjdk/bench/java/lang/ArrayCopyObject.java
+++ b/test/micro/org/openjdk/bench/java/lang/ArrayCopyObject.java
@@ -61,7 +61,7 @@ class MyClass {
  }
 }
 
-@State(Scope.Benchmark)
+@State(Scope.Thread)
 @BenchmarkMode(Mode.Throughput)
 @OutputTimeUnit(TimeUnit.MILLISECONDS)
 @Warmup(iterations = 10, time = 1)
--- a/test/micro/org/openjdk/bench/java/io/DataOutputStreamTest.java
+++ b/test/micro/org/openjdk/bench/java/io/DataOutputStreamTest.java
@@ -33,7 +33,7 @@ import java.util.concurrent.TimeUnit;
 @Fork(2)
 @Measurement(iterations = 6, time = 1)
 @Warmup(iterations=4, time = 2)
-@State(Scope.Benchmark)
+@State(Scope.Thread)
 public class DataOutputStreamTest {
 
     public enum BasicType {CHAR, SHORT, INT, STRING}
--- a/test/micro/org/openjdk/bench/java/lang/ArrayFiddle.java
+++ b/test/micro/org/openjdk/bench/java/lang/ArrayFiddle.java
@@ -60,7 +60,7 @@ import java.util.concurrent.TimeUnit;
 @Measurement(iterations = 5, time = 2)
 @Fork(3)
 @OutputTimeUnit(TimeUnit.NANOSECONDS)
-@State(Scope.Benchmark)
+@State(Scope.Thread)
 public class ArrayFiddle {
     @Param("999")
     public int size;
--- a/test/micro/org/openjdk/bench/java/time/format/DateTimeFormatterBench.java
+++ b/test/micro/org/openjdk/bench/java/time/format/DateTimeFormatterBench.java
@@ -53,7 +53,7 @@ import org.openjdk.jmh.infra.Blackhole;
 @Warmup(iterations = 5, time = 1)
 @Measurement(iterations = 5, time = 1)
 @Fork(3)
-@State(Scope.Benchmark)
+@State(Scope.Thread)
 public class DateTimeFormatterBench {
 
     private static final TimeZone TIME_ZONE = TimeZone.getTimeZone("UTC");
--- a/test/micro/org/openjdk/bench/jdk/incubator/vector/IndexInRangeBenchmark.java
+++ b/test/micro/org/openjdk/bench/jdk/incubator/vector/IndexInRangeBenchmark.java
@@ -29,7 +29,7 @@ import org.openjdk.jmh.annotations.*;
 
 @BenchmarkMode(Mode.Throughput)
 @OutputTimeUnit(TimeUnit.MILLISECONDS)
-@State(Scope.Benchmark)
+@State(Scope.Thread)
 @Warmup(iterations = 3, time = 1)
 @Measurement(iterations = 5, time = 1)
 @Fork(value = 1, jvmArgsPrepend = {"--add-modules=jdk.incubator.vector"})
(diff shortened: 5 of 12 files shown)


test/micro/org/openjdk/bench/java/lang/ArrayCopyObject.java
```diff
@@ -61,7 +61,7 @@ public MyClass(int val) {
  }
 }
 
-@State(Scope.Benchmark)
+@State(Scope.Thread)
```
stsypanov, line 64: Are you sure it makes sense as in `main()` method we set `fork(1)` so there's only one thread running the benchmark

swati-sha: AFAIK fork value specifies number of times harness should [fork](https://javadoc.io/doc/org.openjdk.jmh/jmh-core/0.6/org/openjdk/jmh/annotations/Fork.html). Also the change is setting scope to thread level not controlling the number of threads.

ericcaspole (approved):
LGTM.
Eric

Hamlin-Li:
Hi,
I'm not sure if I understand this improvement correctly.
I'm not quite familiar with JMH and it's annotations, but seems to me, the change from `@State(Scope.Benchmark)` to `@State(Scope.Thread)` should not improve the performance by reducing cache contention, as in the jmh doc it says "State objects are usually injected into Benchmark methods as ***arguments***, and JMH takes care of their instantiation and sharing.", this seems mean that @State only matters when the annotated class is used as a parameter of a @Benchmark method, but in the tests you modifed, seems there is no such use case.
Please also check the sample usages at https://github.com/openjdk/jmh/blob/master/jmh-samples/src/main/java/org/openjdk/jmh/samples/JMHSample_03_States.java.

kuksenko:
@Hamlin-Li 
The PR is fully correct.
Don't forget, every Java instance method has a specific argument which called "this".  That is why @State annotation is working.

swati-sha:
> Hi, I'm not sure if I understand this improvement correctly. I'm not quite familiar with JMH and it's annotations, but seems to me, the change from `@State(Scope.Benchmark)` to `@State(Scope.Thread)` should not improve the performance by reducing cache contention, as in the jmh doc it says "State objects are usually injected into Benchmark methods as _**arguments**_, and JMH takes care of their instantiation and sharing.", this seems mean that @State only matters when the annotated class is used as a parameter of a @benchmark method, but in the tests you modifed, seems there is no such use case. Please also check the sample usages at https://github.com/openjdk/jmh/blob/master/jmh-samples/src/main/java/org/openjdk/jmh/samples/JMHSample_03_States.java.

Please check the https://github.com/openjdk/jmh/blob/master/jmh-samples/src/main/java/org/openjdk/jmh/samples/JMHSample_04_DefaultState.java sample case where @State annotation applies to all instance methods. The PR proposal is to change default state from benchmark to thread.

Hamlin-Li:
@kuksenko @swati-sha Thanks for explanation. I can understand what you said.
But I'm still not quite sure, as I remember jmh does some code manipulation or instrumentation at source code (or bytecode level?), so the jmh test code you write or see might not be the exact code to be executed at runtime.
It's better to be reviewed further by some one more familiar with jmh, or could you add some data comparing the performance difference, so we can tell it easily?

kuksenko:
@Hamlin-Li 
I am one of JMH's authors. I know how it works. There is no need for tests.

johnshajiang:
Not review this PR, but just raise a question.
Should a JMH test, at least in JDK repo, always uses `@State(Scope.Thread)`, even though it uses only one thread?

I just looked through those JMH tests, and found all of them, like the bellows, don't specify the number of threads via `@Threads`.

```
org/openjdk/bench/java/io/DataOutputStreamTest.java
org/openjdk/bench/java/lang/ArrayCopyObject.java
```

I suppose the default number of threads is 1.
Maybe the default value will be overridden via the commands when running these JMH tests in bulk (?)

Hamlin-Li:
> @Hamlin-Li I am one of JMH's authors. I know how it works. There is no need for tests.

@kuksenko Thanks for the confirmation.
Then the benchmark modification looks good to me.

Hamlin-Li (approved):

swati-sha:
Thanks all for reviewing this ! @ericcaspole , @sviswa7 , @Hamlin-Li

PR #14758, opened 2023-07-03
8311247: Some cpp files are compiled with -std:c11 flag

djelinski:
Please review this patch that configures C++ arguments on build jobs that involve compiling CPP files. As a result of this change, CPP files are compiled with `-std:c++14` command line argument instead of `-std:c11`, which is used when C++ arguments are not configured.
While at it, I simplified the `java.security.jgss/Lib.gmk` file by moving the additional include directory to `EXTRA_HEADER_DIRS`.

This patch fixes the following clang warning:
```
warning: argument unused during compilation: '-std:c11'
```

Microsoft states that [std:c++14 is the default](https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170), so there shouldn't be any differences in produced code.

Testing:
- verified that after the changes, all CPP files are compiled with `std:c++14` instead of `-std:c11`
- spot-checked a few `cmdline` files that changed after this patch was applied; the `-std` change was the only difference
- tier1-5 builds, tier1-2 tests and client libs tests continue to pass on Windows, Linux & MacOS.

---------
### Progress

### Issue
 * [JDK-8311247](https://bugs.openjdk.org/browse/JDK-8311247): Some cpp files are compiled with -std:c11 flag (**Bug** - P4)

### Reviewers
 * [Julian Waters](https://openjdk.org/census#jwaters) (@TheShermanTanker - Committer)
 * [Erik Joelsson](https://openjdk.org/census#erikj) (@erikj79 - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14758/head:pull/14758` \
`$ git checkout pull/14758`

`$ git checkout pull/14758` \
`$ git pull https://git.openjdk.org/jdk.git pull/14758/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14758`

`$ git pr show -t 14758`

</details>

<a href="https://git.openjdk.org/jdk/pull/14758.diff">https://git.openjdk.org/jdk/pull/14758.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14758#issuecomment-1619000429)

### DIFF
--- a/make/modules/java.desktop/lib/Awt2dLibraries.gmk
+++ b/make/modules/java.desktop/lib/Awt2dLibraries.gmk
@@ -132,6 +132,7 @@ $(eval $(call SetupJdkLibrary, BUILD_LIBAWT, \
     EXCLUDE_FILES := $(LIBAWT_EXFILES), \
     OPTIMIZATION := HIGHEST, \
     CFLAGS := $(CFLAGS_JDKLIB) $(LIBAWT_CFLAGS), \
+    CXXFLAGS := $(CXXFLAGS_JDKLIB) $(LIBAWT_CFLAGS), \
     EXTRA_HEADER_DIRS := $(LIBAWT_EXTRA_HEADER_DIRS), \
     DISABLED_WARNINGS_gcc_awt_LoadLibrary.c := unused-result, \
     DISABLED_WARNINGS_gcc_debug_mem.c := format-nonliteral, \
@@ -609,7 +610,9 @@ ifeq ($(call isTargetOs, windows), true)
   $(eval $(call SetupJdkLibrary, BUILD_LIBJAWT, \
       NAME := jawt, \
       OPTIMIZATION := LOW, \
-      CFLAGS := $(CXXFLAGS_JDKLIB) \
+      CFLAGS := $(CFLAGS_JDKLIB) \
+          $(LIBJAWT_CFLAGS), \
+      CXXFLAGS := $(CXXFLAGS_JDKLIB) \
           $(LIBJAWT_CFLAGS), \
       EXTRA_HEADER_DIRS := $(LIBJAWT_EXTRA_HEADER_DIRS), \
       LDFLAGS := $(LDFLAGS_JDKLIB) $(LDFLAGS_CXX_JDK), \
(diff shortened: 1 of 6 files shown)


make/modules/java.desktop/lib/Awt2dLibraries.gmk
```diff
@@ -132,6 +132,7 @@ $(eval $(call SetupJdkLibrary, BUILD_LIBAWT, \
     EXCLUDE_FILES := $(LIBAWT_EXFILES), \
     OPTIMIZATION := HIGHEST, \
     CFLAGS := $(CFLAGS_JDKLIB) $(LIBAWT_CFLAGS), \
+    CXXFLAGS := $(CXXFLAGS_JDKLIB) $(LIBAWT_CFLAGS), \
```
aivanov-jdk, line 135: `LIBAWT_CFLAGS` applies to both C and C++, right? It contains some common flags; it follows the pattern with other libs below.

djelinski: That's correct. `LIBAWT_CFLAGS contains a bunch of `-D` flags, which are common for C and C++, and `-EHsc` flag, which is probably C++ only, but the compiler doesn't complain.

`jabswitch` below is using some C++-specific flags, but it only compiles one C++ file and no C files. I suppose we could simplify that makefile at some point. All other flags are common between C and C++.

aivanov-jdk: Thank for clarification.

aivanov-jdk:
/reviewers 2

I believe a reviewer from the build team is still required.

aivanov-jdk (approved):

TheShermanTanker:
Looks good to me. I actually have a similar change applied locally on my repo too

TheShermanTanker (approved):

prrace (approved):

mrserb:
Did we document the current standard of C++ and C we can use?

TheShermanTanker:
@mrserb We use C11 and C++14, and we're soon to move to C++17 once the AIX port gets a compiler that can accept it

djelinski:
The C++ standard is only documented for Hotspot, and the C standard is not documented anywhere as far as I could tell.

(Still waiting for a review from build-dev)

erikj79 (approved):
I'm ok with this, but it's unfortunate that we need to repeat so much.

djelinski:
> it's unfortunate that we need to repeat so much.

Right. The default make convention uses `CPPFLAGS` for flags shared between preprocessor and both C and C++ compilers. We don't have any shared variable as far as I could tell.
Thanks for the reviews!

/integrate

PR #13434, opened 2023-04-11
8305905: Deprecate String::toLowerCase() and String::toUpperCase()

Glavo:
This PR proposes to deprecate `toLowerCase()`/`toUpperCase()` of `String`.

---------
### Progress
- [ ] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [ ] Change requires CSR request [JDK-8305907](https://bugs.openjdk.org/browse/JDK-8305907) to be approved
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issues
 * [JDK-8305905](https://bugs.openjdk.org/browse/JDK-8305905): Deprecate String::toLowerCase() and String::toUpperCase() (**Enhancement** - P4)
 * [JDK-8305907](https://bugs.openjdk.org/browse/JDK-8305907): Deprecate String::toLowerCase() and String::toUpperCase() (**CSR**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/13434/head:pull/13434` \
`$ git checkout pull/13434`

Update a local copy of the PR: \
`$ git checkout pull/13434` \
`$ git pull https://git.openjdk.org/jdk.git pull/13434/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 13434`

View PR using the GUI difftool: \
`$ git pr show -t 13434`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jdk/pull/13434.diff">https://git.openjdk.org/jdk/pull/13434.diff</a>

</details>

PR #14763, opened 2023-07-03
8311943: Cleanup usages of toLowerCase() and toUpperCase() in java.base

Glavo:
Clean up misuses of `toLowerCase()`/`toUpperCase()` in java.base.

---------
### Progress

### Issue
 * [JDK-8311943](https://bugs.openjdk.org/browse/JDK-8311943): Cleanup usages of toLowerCase() and toUpperCase() in java.base (**Task** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14763/head:pull/14763` \
`$ git checkout pull/14763`

`$ git checkout pull/14763` \
`$ git pull https://git.openjdk.org/jdk.git pull/14763/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14763`

`$ git pr show -t 14763`

</details>

<a href="https://git.openjdk.org/jdk/pull/14763.diff">https://git.openjdk.org/jdk/pull/14763.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14763#issuecomment-1632480071)


jaikiran:
Hello Glavo, I've created https://bugs.openjdk.org/browse/JDK-8311943 to track this change. Please update the title to this PR to `8311943: Cleanup usages of toLowerCase() and toUpperCase() in java.base` so that it triggers the official RFR.

src/java.base/share/classes/java/io/StreamTokenizer.java
```diff
@@ -628,7 +629,7 @@ else if ('0' <= c && c <= '9') {
             peekc = c;
             sval = String.copyValueOf(buf, 0, i);
             if (forceLower)
-                sval = sval.toLowerCase();
+                sval = sval.toLowerCase(Locale.ROOT);
```
AlanBateman, line 632: I suspect this change to StreamTokenizer needs eyes. I think long standing behavior of the lowerCaseMode(true) has been to use the rules for the default locale so we need to be careful.

Glavo: > I suspect this change to StreamTokenizer needs eyes. I think long standing behavior of the lowerCaseMode(true) has been to use the rules for the default locale so we need to be careful.

I investigated some usage of this method on GitHub:

https://github.com/search?q=%22lowerCaseMode%28true%29%22+language%3AJava&type=code

In the use cases I see, it seems that no one wants to rely on the default locale.

However, while I think this corrects the behavior, this caused a change in the behavior of the API, so a CSR may be required. I don't want to debate this in this PR, so I'll revert this change and open a new PR in the future.

Glavo: > Maybe a small suggestion to make it clear whats wanted here. In other projects I am involved in (Apache Lucene/Solr, Apache TIKA, PostgresSQL JDBC, Checkstyle itsself, Elasticserach/Opensearch), which use the [forbiddenapis Maven/Gradle/Ant plugin](https://github.com/policeman-tools/forbidden-apis/), we forbid all calls to several Java APIs (including toLowerCase/toUpperCase case). All bytecode using this will build failure (FYI, we also disallow other stuff like relying of default timezone or characterset). To make it clear what is really intended, those projects agreed on having `toLowerCase(Locale.getDefault())`, so it is explicit what's wanted. Without that it could be that somebody else starts the discussion again.
> 
> This is just a suggestion to be explicit as it makes maintaining the code easier.

I agree with this.

I'm working on deprecating `toLowerCase()` and `toUpperCase()`, this PR is part of that effort. I wish to convert all use cases of them to `toLowerCase(Locale)` and `toUpperCase(Locale)`.

More backstory is detailed in https://github.com/openjdk/jdk/pull/13434.

AlanBateman: > However, while I think this corrects the behavior, this caused a change in the behavior of the API, so a CSR may be required. I don't want to debate this in this PR, so I'll revert this change and open a new PR in the future.

StreamTokenizer is a very old API and changing long standing behavior may break something or be observable with existing code/usages. I see youve reverted this part (thanks) and looking at it separately is fine. It might be that the conclusion is that it's just too risky to change, in which case Uwe's suggestion is good and would avoid it showing up on someone's else radar in the future.

Glavo: If we can't normalize this use case to be locale-independent, then I even think `lowerCaseMode` should be deprecated, because it's almost impossible for users to get expected behavior with this method.

In order to make it meaningful, I think it is still necessary to consider making it locale insensitive. We can allow users to fall back to the old behavior through new system properties, or introduce new API methods in `StreamTokenizer` to allow users to set the Locale to be used.

AlanBateman:
Now that the incompatible change to StreamTokenizer is dropped from this change then I assume the rest can be reviewed.

Glavo:
I updated this PR to resolve the merge conflict. Now it is waiting to be reviewed again

Glavo:
Can someone review this PR?

naotoj (approved):
LGTM. Thanks for the changes.

PR #14766, opened 2023-07-04
8311031: JTable header border vertical lines are not aligned with data grid lines

TejeshR13:
The header border uses `g.drawLine` whereas the JTable data grid lines uses `SwingUtilities2.drawVLine` and `SwingUtilities2.drawHLine` to draw horizontal and vertical lines. The SwingUtilities2 uses `Graphics.fillRect` which contributes to the difference between the position of these two lines which happens/visible at higher ui scaling (difference in alignment between vertical lines of these two). The fix propose to use the same methods for metal L&F of JTable header border paint. 
CI testing shows green.

image

---------
### Progress

### Issue
 * [JDK-8311031](https://bugs.openjdk.org/browse/JDK-8311031): JTable header border vertical lines are not aligned with data grid lines (**Bug** - P4)

### Reviewers
 * [Abhishek Kumar](https://openjdk.org/census#abhiscxk) (@kumarabhi006 - Committer) ⚠️ Review applies to [bfb3ae8d](https://git.openjdk.org/jdk/pull/14766/files/bfb3ae8d655e2a92adb48b4da50c8e5f3e2af161)
 * [Prasanta Sadhukhan](https://openjdk.org/census#psadhukhan) (@prsadhuk - **Reviewer**) ⚠️ Review applies to [106de984](https://git.openjdk.org/jdk/pull/14766/files/106de98450f595d04a78ba2b0346fedbd6e27b0b)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14766/head:pull/14766` \
`$ git checkout pull/14766`

`$ git checkout pull/14766` \
`$ git pull https://git.openjdk.org/jdk.git pull/14766/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14766`

`$ git pr show -t 14766`

</details>

<a href="https://git.openjdk.org/jdk/pull/14766.diff">https://git.openjdk.org/jdk/pull/14766.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14766#issuecomment-1620045241)


test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,89 @@
...
+    static JFrame f;
+    static JTable j;
+    static PassFailJFrame passFailJFrame;
+    public static void showTable() throws Exception {
+        final String INSTRUCTIONS = """
+                Instructions to Test:
+                1. Ensure the test is running 225% scaling.
+                2. Check if the Table header border vertical lines are
+                aligned with table data grid lines.
+                3. If there is a miss-match between them press FAIL,
```
kumarabhi006, line 51:
```suggestion
                3. If there is a mismatch between them press FAIL,
```

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
kumarabhi006, line 1: Is it possible to make an automated test by comparing pixel colors of images before and after fix?

kumarabhi006:
image

There is a small gap between the horizontal and vertical line in size column (above 1 or Green) whereas grid lines don't have any gap.
 
Is it correct?

src/java.desktop/share/classes/javax/swing/plaf/metal/MetalBorders.java
```diff
@@ -1108,11 +1108,11 @@ public void paintBorder(Component c, Graphics g, int x, int y, int w, int h) {
             g.translate( x, y );
 
             g.setColor( MetalLookAndFeel.getControlDarkShadow() );
-            g.drawLine( w-1, 0, w-1, h-1 );
-            g.drawLine( 1, h-1, w-1, h-1 );
+            SwingUtilities2.drawVLine(g,w-1, 0, h-1);
```
kumarabhi006, line 1111:
```suggestion
            SwingUtilities2.drawVLine(g, w-1, 0, h-1);
```

src/java.desktop/share/classes/javax/swing/plaf/metal/MetalBorders.java
```diff
@@ -1108,11 +1108,11 @@ public void paintBorder(Component c, Graphics g, int x, int y, int w, int h) {
...
+            SwingUtilities2.drawHLine(g,1,  w-1,h-1);
```
kumarabhi006, line 1112:
```suggestion
            SwingUtilities2.drawHLine(g, 1, w-1, h-1);
```

src/java.desktop/share/classes/javax/swing/plaf/metal/MetalBorders.java
```diff
@@ -1108,11 +1108,11 @@ public void paintBorder(Component c, Graphics g, int x, int y, int w, int h) {
...
             g.setColor( MetalLookAndFeel.getControlHighlight() );
-            g.drawLine( 0, 0, w-2, 0 );
-            g.drawLine( 0, 0, 0, h-2 );
+            SwingUtilities2.drawHLine(g,0, w-2,0 );
```
kumarabhi006, line 1114:
```suggestion
            SwingUtilities2.drawHLine(g, 0, w-2, 0 );
```

src/java.desktop/share/classes/javax/swing/plaf/metal/MetalBorders.java
```diff
@@ -1108,11 +1108,11 @@ public void paintBorder(Component c, Graphics g, int x, int y, int w, int h) {
...
+            SwingUtilities2.drawVLine(g,0, 0, h-2);
```
kumarabhi006, line 1115:
```suggestion
            SwingUtilities2.drawVLine(g, 0, 0, h-2);
```

TejeshR13:
> image
> 
> There is a small gap between the horizontal and vertical line in size column (above 1 or Green) whereas grid lines don't have any gap.
> 
> Is it correct?

That gap is caused because of alignment miss-match between vertical lines of header border and data grid lines. To be specific the header vertical line is slightly left (around a pixel) than data grid line.

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
TejeshR13: Tried to check the line alignment, but couldn't do it because of anti aliasing and scaling. And before and after fix means storing the image of before fix and then comparing with after fix?

TejeshR13: Still I feel that would be blind comparison, since we won't be able to exactly test the fix.

kumarabhi006:
> That gap is caused because of alignment miss-match between vertical lines of header border and data grid lines. To be specific the header vertical line is slightly left (around a pixel) than data grid line.

Is it require to address?

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
aivanov-jdk: I strongly believe it should be an automated test. It may even be headless, if possible. There are a couple of examples which compare border rendering.

You don't need to store images, just count the number of pixels of a particular colour. If you detect an inconsistency, fail the test. You may need to tweak the width or height of columns and rows so that old method results in visible inconsistencies in rendering.

TejeshR13: How will counting the number of pixels of a particular color help in validating the fix? I was planning to take a portion of image which includes header and 1st row and then check the column of vertical line from header to data, if all are same color then its a straight line, else fail the test... But I was not able to compare the data, some inconsistencies were there with rgb values.... Will try to work on that again......

TejeshR13:
> > > That gap is caused because of alignment miss-match between vertical lines of header border and data grid lines. To be specific the header vertical line is slightly left (around a pixel) than data grid line.
> > 
> > 
> > Is it require to address?
> 
> I'd rather say ‘yes’. Since you're trying to improve border rendering, it's better to fix this inconsistency too.
> 
> But to get pixel-perfect rendering, you may need to remove the scale transform.

I'm not sure if it is a bug, the gap is because of white lines drawn for each header cell (bottom-left till top-left and then to to-right for each header cell). Kind of shade effect.

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
aivanov-jdk: I wasn't clear enough, however, you understood it. There are three possible colours in the header: background, highlight and shadow. There's also text which we can ignore; if anti-aliasing is disabled, it will be solid colour only; alternatively the headers may have no text at all or a space.

So, the header should start with highlight colour, there should be 2 pixels of it at 225%, right? Then the highlight is followed by the background until you find the shadow colour. It should also be 2 pixels. The shadow should be followed by the highlight colour of the next cell, there should be no background colour between highlight and shadow.

This algorithm works in the [`ScaledEtchedBorderTest.java`](https://github.com/openjdk/jdk/blob/00ac46c06d8a7154a007cd67092bd80226a02fc3/test/jdk/javax/swing/border/EtchedBorder/ScaledEtchedBorderTest.java) test, see the [`checkBorder` method](https://github.com/openjdk/jdk/blob/00ac46c06d8a7154a007cd67092bd80226a02fc3/test/jdk/javax/swing/border/EtchedBorder/ScaledEtchedBorderTest.java#L151-L154).

aivanov-jdk:
> I'm not sure if it is a bug, the gap is because of white lines drawn for each header cell (bottom-left till top-left and then to to-right for each header cell). Kind of shade effect.

I admit I am confused to which part it refers.

I re-checked your after screenshot and it looks fine to me. The shadow is followed by white highlight. There was background colour seen between these two colours in the before-fix screenshot.

There's still a 2-pixel background colour under the highlight. Yet it is present in the before-fix too, so it looks it is designed this way for a smoother 3D shading effect.

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
TejeshR13: I understood your algorithm, hadn't thought this way. What I did is to get the grid line color from data cell, then just compare for the same color at same columns (for straight line/alignment w.r.t vertical grid line and vertical header border). Before my fix, there would be shift of header line to one pixel left, hence the test would fail. I guess this way I can ensure that the alignment will be there after the fix. I tested it in mach5, it is green.

TejeshR13:
@aivanov-jdk @kumarabhi006 Is there anything else w.r.t fix and automated test ? I think have addressed your review suggestions.

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,129 @@
...
+            table.paint(g2d);
+        } finally {
+            g2d.dispose();
+        }
+
+        verticalLineCol = (int)(table.getTableHeader().
+                getColumnModel().getColumn(0).getWidth() * SCALE) - 2;
+        expectedRGB = imgDate.getRGB(verticalLineCol,0);
+
+        for(int i = 0; i < imgHeader.getHeight();i++) {
```
kumarabhi006, line 112:
```suggestion
        for(int i = 0; i < imgHeader.getHeight(); i++) {
```

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,129 @@
...
+            for(int j = verticalLineCol; j < verticalLineCol + 3; j++) {
+                if(expectedRGB != imgHeader.getRGB(j,i)) {
```
kumarabhi006, line 114:
```suggestion
                if (expectedRGB != imgHeader.getRGB(j, i)) {
```

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,129 @@
...
+                    throw new RuntimeException("Test Failed");
+                }
+            }
+        }
+
+        for(int i = 0; i < table.getRowCount() * table.getRowHeight() * SCALE;i++) {
```
kumarabhi006, line 120:
```suggestion
        for(int i = 0; i < table.getRowCount() * table.getRowHeight() * SCALE; i++) {
```

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,129 @@
...
+            for(int j = verticalLineCol; j < verticalLineCol + 3; j++) {
+                if(expectedRGB != imgDate.getRGB(j,i)) {
```
kumarabhi006, line 122:
```suggestion
                if (expectedRGB != imgDate.getRGB(j, i)) {
```

TejeshR13:
@aivanov-jdk @kumarabhi006 Any comments on this PR?

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,129 @@
...
+        double h;
+        String[][] data = {
+                { "1", "1", "Green"},
+                { "2", "2", "Blue"}
+        };
+
+        String[] columnNames = { "Number", "Size", "Color"};
+
+        table = new JTable(data, columnNames);
+        table.setSize(WIDTH,HEIGHT);
```
kumarabhi006, line 76:
```suggestion
        table.setSize(WIDTH, HEIGHT);
```

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,129 @@
...
+        final JTableHeader header = table.getTableHeader();
+        TableCellRenderer renderer = header.getDefaultRenderer();
+        header.setDefaultRenderer(renderer);
+        table.updateUI();
+
+        Dimension size = header.getPreferredSize();
+        header.setSize(size);
+        w = SCALE * size.width;
+        h = SCALE * size.height;
+        imgHeader = new BufferedImage((int)(w),(int)(h),BufferedImage.TYPE_INT_RGB);
```
kumarabhi006, line 87:
```suggestion
        imgHeader = new BufferedImage((int)(w), (int)(h), BufferedImage.TYPE_INT_RGB);
```

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,129 @@
...
+        g2d.scale(SCALE, SCALE);
+        try {
+            table.paint(g2d);
+        } finally {
+            g2d.dispose();
+        }
+
+        verticalLineCol = (int)(table.getTableHeader().
+                getColumnModel().getColumn(0).getWidth() * SCALE) - 2;
+        expectedRGB = imgData.getRGB(verticalLineCol,0);
```
kumarabhi006, line 110:
```suggestion
        expectedRGB = imgData.getRGB(verticalLineCol, 0);
```

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+        for(int i = 0; i < imgHeader.getHeight(); i++) {
+            for(int j = verticalLineCol; j < verticalLineCol + 3; j++) {
+                if(expectedRGB != imgHeader.getRGB(j, i)) {
```
kumarabhi006, line 114:
```suggestion
                if (expectedRGB != imgHeader.getRGB(j, i)) {
```

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,129 @@
...
+            table.paint(g2d);
+        } finally {
+            g2d.dispose();
+        }
+
+        verticalLineCol = (int)(table.getTableHeader().
+                getColumnModel().getColumn(0).getWidth() * SCALE) - 2;
+        expectedRGB = imgData.getRGB(verticalLineCol,0);
+
+        for(int i = 0; i < imgHeader.getHeight(); i++) {
```
kumarabhi006, line 112:
```suggestion
        for (int i = 0; i < imgHeader.getHeight(); i++) {
```

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,129 @@
...
+            for(int j = verticalLineCol; j < verticalLineCol + 3; j++) {
```
kumarabhi006, line 113:
```suggestion
            for (int j = verticalLineCol; j < verticalLineCol + 3; j++) {
```

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,129 @@
...
+                if(expectedRGB != imgHeader.getRGB(j, i)) {
+                    throw new RuntimeException("Test Failed");
+                }
+            }
+        }
+
+        for(int i = 0; i < table.getRowCount() * table.getRowHeight() * SCALE; i++) {
```
kumarabhi006, line 120:
```suggestion
        for (int i = 0; i < table.getRowCount() * table.getRowHeight() * SCALE; i++) {
```

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,129 @@
...
+            for(int j = verticalLineCol; j < verticalLineCol + 3; j++) {
```
kumarabhi006, line 121:
```suggestion
            for (int j = verticalLineCol; j < verticalLineCol + 3; j++) {
```

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,129 @@
...
+                if(expectedRGB != imgData.getRGB(j, i)) {
```
kumarabhi006, line 122:
```suggestion
                if (expectedRGB != imgData.getRGB(j, i)) {
```

kumarabhi006 (commented):
Please check for spaces as per standard.

kumarabhi006 (approved):
Test verified.

alisenchung:
Can this test be run on other L&Fs as well?

TejeshR13:
> Can this test be run on other L&Fs as well?

No, the fix and test is specific to Metal L&F.

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+import java.awt.Dimension;
+import java.awt.Graphics2D;
+import java.awt.image.BufferedImage;
+
+/*
+ * @test
+ * @bug 8311031
+ * @library /java/awt/regtesthelpers
+ * @build PassFailJFrame
```
prsadhuk, line 38: I guess it's no longer manual so this dependency should be removed..

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,129 @@
...
+    private static final int HEIGHT = 150;
+    private static final double SCALE = 2.25;
+
+    public static void main(String[] args) throws Exception {
+        UIManager.setLookAndFeel("javax.swing.plaf.metal.MetalLookAndFeel");
+        SwingUtilities.invokeAndWait(() -> {
+            try {
+                TableHeaderBorderPositionTest.Test();
+            } catch (Exception e) {
+                throw new RuntimeException(e);
```
prsadhuk, line 56: I guess there's no need of `Test()` to have `throws Exception` signature and therefore no need to have this try-catch block..

prsadhuk (commented):
I hope you have run this test on all mach5 platforms

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,129 @@
...
+            for (int j = verticalLineCol; j < verticalLineCol + 3; j++) {
+                if (expectedRGB != imgHeader.getRGB(j, i)) {
+                    throw new RuntimeException("Test Failed");
+                }
+            }
+        }
+
+        for (int i = 0; i < table.getRowCount() * table.getRowHeight() * SCALE; i++) {
+            for (int j = verticalLineCol; j < verticalLineCol + 3; j++) {
+                if (expectedRGB != imgData.getRGB(j, i)) {
```
prsadhuk, line 122: You probably can write the bufferedimage into png file for debugging help during fail case..

TejeshR13:
> I hope you have run this test on all mach5 platforms

I have updated the review comments and mach5 platform runs show green.

prsadhuk (approved):

kumarabhi006 (approved):

src/java.desktop/share/classes/javax/swing/plaf/metal/MetalBorders.java
```diff
@@ -1108,11 +1108,11 @@ public void paintBorder(Component c, Graphics g, int x, int y, int w, int h) {
...
 
             g.setColor( MetalLookAndFeel.getControlDarkShadow() );
-            g.drawLine( w-1, 0, w-1, h-1 );
-            g.drawLine( 1, h-1, w-1, h-1 );
+            SwingUtilities2.drawVLine(g, w-1, 0, h-1);
+            SwingUtilities2.drawHLine(g, 1, w-1, h-1);
             g.setColor( MetalLookAndFeel.getControlHighlight() );
-            g.drawLine( 0, 0, w-2, 0 );
-            g.drawLine( 0, 0, 0, h-2 );
+            SwingUtilities2.drawHLine(g, 0, w-2, 0 );
```
aivanov-jdk, line 1114:
```suggestion
            SwingUtilities2.drawHLine(g, 0, w-2, 0);
```
You removed other spaces but this one's left.

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+        for (int i = 0; i < imgHeader.getHeight(); i++) {
+            for (int j = verticalLineCol; j < verticalLineCol + 3; j++) {
+                if (expectedRGB != imgHeader.getRGB(j, i)) {
+                    try {
+                        ImageIO.write(imgHeader, "png",new File("FailureImageHeader.png"));
+                    } catch (IOException e) {
+                        e.printStackTrace();
+                    }
+                    throw new RuntimeException("Test Failed");
```
aivanov-jdk, line 118: Avoid duplicate code, move saving the image into a separate method.

aivanov-jdk (changes requested):
[The same comments](https://github.com/openjdk/jdk/pull/14464#discussion_r1273475982) I left in #14464 applies here.

Yet this change doesn't make all the borders in the table header consistent: there are header cell borders where shadow or highlight are 1- or 2-pixel wide at 175% scale, however, there are now no cases where background colour shines between shadow and highlight.

TejeshR13:
@aivanov-jdk  Please let me know if there are any more review suggestions.

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,128 @@
...
+                if (expectedRGB != imgHeader.getRGB(x, y)) {
+                    saveImage(imgHeader, "FailureImageHeader.png");
+                    throw new RuntimeException("Header Comparison Test " +
+                            "Failed at <" + x + ", " + y + ">");
+                }
+            }
+        }
+
+        int maxVerticalLine = (int) Math.ceil(
+                table.getRowCount() * table.getRowHeight() * SCALE);
```
aivanov-jdk, line 108:
```suggestion
        int maxVerticalLine = (int)Math.ceil(
                table.getRowCount() * table.getRowHeight() * SCALE);
```
In all the other cases, you didn't add a space after the cast operator.

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,128 @@
...
+        h = (int)Math.ceil(SCALE * HEIGHT);
+        BufferedImage imgData = new BufferedImage(w, h,
+                BufferedImage.TYPE_INT_RGB);
+        g2d = imgData.createGraphics();
+        g2d.scale(SCALE, SCALE);
+        try {
+            table.paint(g2d);
+        } finally {
+            g2d.dispose();
+        }
```
aivanov-jdk, line 91: I wonder if it's possible to make the table paint its header into the same image, this would simplify the test.

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+        int verticalLineCol = (int)(table.getTableHeader().
+                getColumnModel().getColumn(0).getWidth() * SCALE) - 2;
```
aivanov-jdk, line 94: Is it reasonable to use the same formatting as in #14464, in `JTableHeaderLabelRightAlignTest.java`?

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
TejeshR13: I too had this thought initially, but wasn't comfortable in combining the images. Will try now.

TejeshR13: I have updated the test with single buffered image.

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -65,64 +64,40 @@ private static void test() {
...
         System.out.println("Test Passed");
     }
-
-    private static void saveImage(BufferedImage img, String fileName) {
-        try {
-            ImageIO.write(img, "png", new File(fileName));
-        } catch (IOException e) {
-            e.printStackTrace();
-        }
-    }
```
aivanov-jdk: In fact, it looks better with `saveImage` method.

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -65,64 +64,40 @@ private static void test() {
...
 
-        int verticalLineCol = (int)(table.getTableHeader().
-                getColumnModel().getColumn(0).getWidth() * SCALE) - 2;
-        int expectedRGB = imgData.getRGB(verticalLineCol, 0);
+        int verticalLineCol = (int) (table.getTableHeader()
+                                          .getColumnModel()
+                                          .getColumn(0)
+                                          .getWidth() * SCALE) - 2;
+        int expectedRGB = bufferedImage.getRGB(verticalLineCol, 0);
+        int maxHeight = (int)(((double)table.getTableHeader().getHeight() * SCALE )
```
aivanov-jdk, line 85:
```suggestion
        int maxHeight = (int)(((double)table.getTableHeader().getHeight() * SCALE)
```

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -65,64 +64,40 @@ private static void test() {
...
-        h = (int)Math.ceil(SCALE * HEIGHT);
-        BufferedImage imgData = new BufferedImage(w, h,
-                BufferedImage.TYPE_INT_RGB);
-        g2d = imgData.createGraphics();
-        g2d.scale(SCALE, SCALE);
         try {
+            header.paint(g2d);
+            g2d.translate(0, header.getHeight());
             table.paint(g2d);
         } finally {
```
aivanov-jdk: Awesome! I didn't think about it. If it works, it makes verification easier. When you read the code, you clearly see the borders of the header and grid align.

I expected the table would paint both its header and its contents, but it may require validation which can't happen until the table is added to a heavyweight container such as a frame.

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -65,64 +64,40 @@ private static void test() {
         Dimension size = header.getPreferredSize();
         header.setSize(size);
 
-        int w = (int)Math.ceil(SCALE * size.width);
-        int h = (int)Math.ceil(SCALE * size.height);
-
-        BufferedImage imgHeader = new BufferedImage(w, h,
+        BufferedImage bufferedImage = new BufferedImage(WIDTH, HEIGHT,
                 BufferedImage.TYPE_INT_RGB);
```
aivanov-jdk: Hm, you have to create the image large enough to contain the table after scaling it by `SCALE`, so you still need `w` and `h` or you could inline `WIDTH * SCALE` and `HEIGHT * SCALE` correspondingly.

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -65,64 +64,40 @@ private static void test() {
...
             g2d.dispose();
         }
 
-        int verticalLineCol = (int)(table.getTableHeader().
-                getColumnModel().getColumn(0).getWidth() * SCALE) - 2;
-        int expectedRGB = imgData.getRGB(verticalLineCol, 0);
+        int verticalLineCol = (int) (table.getTableHeader()
+                                          .getColumnModel()
+                                          .getColumn(0)
+                                          .getWidth() * SCALE) - 2;
```
aivanov-jdk, line 83:
```suggestion
        int verticalLineCol = (int)(table.getTableHeader()
                                         .getColumnModel()
                                         .getColumn(0)
                                         .getWidth() * SCALE) - 2;
```
If you add a space after the cast operator, add it everywhere; if you don't, add it nowhere.

I could've given you a suggestion based on my formatting, and I have the space after the cast operator because Java Coding Style recommends adding it. Yet there's no consensus about it in client-libs.

aivanov-jdk (changes requested):

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,103 @@
...
+        SwingUtilities.invokeAndWait(TableHeaderBorderPositionTest::test);
+    }
+
+    private static void test() {
+        String[][] data = {
+                {"1", "1"}
+        };
+
+        String[] columnNames = {"Size", "Size"};
+        JTable table = new JTable(data, columnNames);
```
aivanov-jdk, line 60:
```suggestion
        String[][] data = {
                {"1", "1"}
        };
        String[] columnNames = {"Size", "Size"};

        JTable table = new JTable(data, columnNames);
```
Let's separate the data — `data` and `columnNames` from the table.

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,103 @@
...
+        table.setSize(WIDTH, HEIGHT);
+
+        final JTableHeader header = table.getTableHeader();
+        Dimension size = header.getPreferredSize();
+        header.setSize(size);
+
+        BufferedImage bufferedImage = new BufferedImage(WIDTH, HEIGHT,
+                BufferedImage.TYPE_INT_RGB);
```
aivanov-jdk, line 68:
```suggestion
        JTable table = new JTable(data, columnNames);
        Dimension tableSize = table.getPreferredSize();
        table.setSize(tableSize);

        final JTableHeader header = table.getTableHeader();
        Dimension headerSize = header.getPreferredSize();
        header.setSize(headerSize);

        Dimension size = new Dimension(Math.max(headerSize.width, tableSize.width),
                                       headerSize.height + tableSize.height);

        BufferedImage bufferedImage =
                new BufferedImage((int)Math.ceil(size.width * SCALE),
                                  (int)Math.ceil(size.height * SCALE),
                                  BufferedImage.TYPE_INT_RGB);
```

This takes into account both the header size and table size and sets the size of the image appropriately.

The constants `WIDTH` and `HEIGHT` have become redundant.

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
TejeshR13: Since we are saving only once, I don't think separate method is required.

aivanov-jdk: It is not required but it separates the duties: the `test` method *verifies* the rendering, it calls `saveImage` to save the image. Such a design is in the spirit of OOP.

As bonus, it allows one to save the image easily before it is verified or even after each paint is performed. Thus, it has its benefits.

I'm not insisting, even though I think it's a better design.

TejeshR13: Ok, no issues, If it adds something better then I'll update it.

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
aivanov-jdk: I though about not separating `data` and `columnNames` with a blank line…

aivanov-jdk (approved):
Looks good now.

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
aivanov-jdk: > Ok, no issues, If it adds something better then I'll update it.

The test in #14464, `JTableHeaderLabelRightAlignTest.java` will also benefit from it, and you haven't integrated it yet.

TejeshR13:
@prsadhuk @kumarabhi006 Since there are few recent updates to the PR after your approval, I hope it is good to integrate?

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,112 @@
...
+        int expectedRGB = bufferedImage.getRGB(verticalLineCol, 0);
+        int maxHeight = (int)(((double)table.getTableHeader().getHeight() * SCALE)
+                + ((double)table.getRowHeight() * SCALE));
+
+        for (int y = 0; y < maxHeight; y++) {
+            for (int x = verticalLineCol; x < verticalLineCol + 3; x++) {
+                if (expectedRGB != bufferedImage.getRGB(x, y)) {
+                    saveBufferedImage(bufferedImage, "failureImage.png");
+                    throw new RuntimeException("Test Failed at <" + x + ", " + y + ">");
+                }
```
prsadhuk, line 99: Ideally this failure check should be done outside EDT else I think you will get InvocationTargetException instead of RuntimeException

prsadhuk (approved):

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
aivanov-jdk: It's not necessary, especially when it complicates the test logic.

Our CI displays the cause of `InvocationTargetException`, so you get the failure message.

test/jdk/javax/swing/JTableHeader/TableHeaderBorderPositionTest.java
```diff
@@ -0,0 +1,119 @@
...
+    private static final double SCALE = 2.25;
+    private static boolean testStatus = true;
+    private static String failureMsg;
+
+    public static void main(String[] args) throws Exception {
+        UIManager.setLookAndFeel("javax.swing.plaf.metal.MetalLookAndFeel");
+        SwingUtilities.invokeAndWait(TableHeaderBorderPositionTest::test);
+        if (!testStatus) {
+            throw new RuntimeException(failureMsg);
+        }
```
aivanov-jdk, line 54:
```suggestion
        if (failureMsg != null) {
            throw new RuntimeException(failureMsg);
        }
```

`testStatus` is redudant.
`failureMsg` is to be declared `volatile` because it's used on different threads; however, `invokeAndWait` internally synchronises, so I think it's fine but I'm still unsure about it and I couldn't find any confirmation or disproof thereof.

aivanov-jdk (changes requested):

aivanov-jdk:
I'll post [this comment](https://github.com/openjdk/jdk/pull/14464#pullrequestreview-1558834675) here too:

> I still prefer the previous version where the exception was thrown directly.
> 
> In fact, you don't create real UI, the test doesn't depend on processing events — the entire test can be run on main thread *safely*. Get rid of `invokeAndWait` altogether?

TejeshR13:
> I'll post [this comment](https://github.com/openjdk/jdk/pull/14464#pullrequestreview-1558834675) here too:
> 
> > I still prefer the previous version where the exception was thrown directly.
> > In fact, you don't create real UI, the test doesn't depend on processing events — the entire test can be run on main thread _safely_. Get rid of `invokeAndWait` altogether?

Yeah, it's better to run on main without `invokeAndWait` which is not required.

aivanov-jdk (approved):

PR #14768, opened 2023-07-04
8311290: Improve java.lang.ref.Cleaner rendered documentation

pavelrappo:
1. Missing leading asterisk interferes with example indentation; nothing in https://bugs.openjdk.org/browse/JDK-8276700 (or review thereof), which introduced that line, suggests that the asterisk was omitted on purpose. 
2. link and linkplain tags are used suboptimally.


RogerRiggs (approved):
LGTM

PR #14773, opened 2023-07-05
8311500: StackWalker.getCallerClass() throws UOE if invoked reflectively

simonis:
As the included jtreg test demonstrates, `StackWalker.getCallerClass()` can throw an `UnsupportedOperationException` if called reflectively. Currently this only happens if we invoke `StackWalker.getCallerClass()` recursively reflectively, but this issue will become more prominent once we fix [JDK-8285447](https://bugs.openjdk.org/browse/JDK-8285447). The gory details follow below:

The protocol between the Java API and the JVM for `StackWalker.getCallerClass()/walk()` is as follows:
- On the Java side, `StackWalker` calls into `StackStreamFactory` for the real work.
- For `StackWalker.getCallerClass()` `StackStreamFactory` basically creates a `Class[]` which will be passed down and filled in the JVM. For `StackWalker.walk()` it will normally be a `StackFrameInfo[]` (or a `LiveStackFrameInfo[]` if the internal `ExtendedOption.LOCALS_AND_OPERANDS` option was used).
- The default size of this arrays is currently `StackStreamFactory.SMALL_BATCH` which is 8 (but see [JDK-8285447](https://bugs.openjdk.org/browse/JDK-8285447)).
- `StackStreamFactory` than calls `AbstractStackWalker.callStackWalk()` which is a natively implemented in the VM by `JVM_CallStackWalk()`.
-  `JVM_CallStackWalk()` calls `StackWalk::walk()` which calls `StackWalk::fetchFirstBatch()` which calls `StackWalk::fill_in_frames()` which walks the stack and fills in the available class/stackframe slots in the passed in array until the array is full or there are no more stack frames,
- Once  `StackWalk::fill_in_frames()` returns, `StackWalk::fetchFirstBatch()` calls back to Java by invoking `AbstractStackWalker::doStackWalk()` to consume the result.
- `AbstractStackWalker::doStackWalk()` calls `consumeFrames()` (which is overridden depending on whether we initially called `getCallerClass()` or `walk()`) which consumes the frames until it either finishes (e.g. finds the caller class) or until there are no more frames.
- In the latter case `consumeFrames()` will call into the the VM again by calling `AbstractStackWalker.fetchStackFrames()` to fetch additional frames from the stack.
- `AbstractStackWalker.fetchStackFrames()` is implemented by `JVM_MoreStackWalk()` which calls `StackWalk::fetchNextBatch()` which calls `StackWalk::fill_in_frames()` (the same method that already fetched the initial batch of frames).

Following is a stacktrace of what I've explained so far:
```
V  [libjvm.so+0x143a96a]  StackWalk::fill_in_frames(long, BaseFrameStream&, int, int, objArrayHandle, int&, bool, JavaThread*)+0x4c
V  [libjvm.so+0x143c7d4]  StackWalk::fetchNextBatch(Handle, long, long, int, int, objArrayHandle, JavaThread*)+0x234
V  [libjvm.so+0xe52141]  JVM_MoreStackWalk+0x154
C  [libjava.so+0x165a5]  Java_java_lang_StackStreamFactory_00024AbstractStackWalker_fetchStackFrames+0x67
j  java.lang.StackStreamFactory$AbstractStackWalker.fetchStackFrames(JJII[Ljava/lang/Object;)I+0 java.base@22-internal
j  java.lang.StackStreamFactory$AbstractStackWalker.fetchStackFrames(I)I+35 java.base@22-internal
j  java.lang.StackStreamFactory$AbstractStackWalker.getNextBatch()I+107 java.base@22-internal
j  java.lang.StackStreamFactory$AbstractStackWalker.peekFrame()Ljava/lang/Class;+32 java.base@22-internal
j  java.lang.StackStreamFactory$AbstractStackWalker.hasNext()Z+1 java.base@22-internal
j  java.lang.StackStreamFactory$AbstractStackWalker.nextFrame()Ljava/lang/Class;+1 java.base@22-internal
j  java.lang.StackStreamFactory$CallerClassFinder.consumeFrames()Ljava/lang/Integer;+21 java.base@22-internal
j  java.lang.StackStreamFactory$CallerClassFinder.consumeFrames()Ljava/lang/Object;+1 java.base@22-internal
j  java.lang.StackStreamFactory$AbstractStackWalker.doStackWalk(JIIII)Ljava/lang/Object;+78 java.base@22-internal
v  ~StubRoutines::call_stub 0x00007fffdfdebd21
V  [libjvm.so+0xd2ac9d]  JavaCalls::call_helper(JavaValue*, methodHandle const&, JavaCallArguments*, JavaThread*)+0x607
V  [libjvm.so+0x123af32]  os::os_exception_wrapper(void (*)(JavaValue*, methodHandle const&, JavaCallArguments*, JavaThread*), JavaValue*, methodHandle const&, JavaCallArguments*, JavaThread*)+0x3a
V  [libjvm.so+0xd2a692]  JavaCalls::call(JavaValue*, methodHandle const&, JavaCallArguments*, JavaThread*)+0x3e
V  [libjvm.so+0x143c4e9]  StackWalk::fetchFirstBatch(BaseFrameStream&, Handle, long, int, int, int, objArrayHandle, JavaThread*)+0x551
V  [libjvm.so+0x143bf55]  StackWalk::walk(Handle, long, int, Handle, Handle, int, int, objArrayHandle, JavaThread*)+0x3ab
V  [libjvm.so+0xe51f85]  JVM_CallStackWalk+0x1e9
C  [libjava.so+0x16524]  Java_java_lang_StackStreamFactory_00024AbstractStackWalker_callStackWalk+0x74
j  java.lang.StackStreamFactory$AbstractStackWalker.callStackWalk(JILjdk/internal/vm/ContinuationScope;Ljdk/internal/vm/Continuation;II[Ljava/lang/Object;)Ljava/lang/Object;+0 java.base@22-internal
j  java.lang.StackStreamFactory$AbstractStackWalker.beginStackWalk()Ljava/lang/Object;+39 java.base@22-internal
j  java.lang.StackStreamFactory$AbstractStackWalker.walkHelper()Ljava/lang/Object;+1 java.base@22-internal
j  java.lang.StackStreamFactory$AbstractStackWalker.walk()Ljava/lang/Object;+35 java.base@22-internal
j  java.lang.StackStreamFactory$CallerClassFinder.findCaller()Ljava/lang/Class;+1 java.base@22-internal
j  java.lang.StackWalker.getCallerClass()Ljava/lang/Class;+38 java.base@22-internal
j  ReflectiveGetCallerClassTest.getCallerClass()V+6

```

So far so good. But the problem is that if called from `StackWalker.getCallerClass()`, `StackWalk::fill_in_frames()` has to check if the caller of `StackWalker.getCallerClass()` is not a `@CallerSensitive` method and throw a `UnsupportedOperationException` if it is. `StackWalk::fill_in_frames()` currently does this unconditionally, no matter if called from `JVM_CallStackWalk()` for the initial batch or from `JVM_MoreStackWalk()` for a subsequent batch. As long as all the frames required to find the caller class fit into the first batch, this is not a problem. If however finding the caller class requires to fetch additional frames, the check described above might trigger due to a false positive. That's exactly what the included jtreg tests simulates. It recursively, reflectively calls   `StackWalker.getCallerClass()` to provoke a situations where:
 - the first batch doesn't include the "true" caller frame of `StackWalker.getCallerClass()` (because reflective frames are ignored)
 - a subsequent batch has a `@CallerSensitive` annotated method in the first slot.

As usually after so many words, the fix itself is pretty trivial: only do the check for a `@CallerSensitive` caller frame in the very first batch.

---------
### Progress

### Issue
 * [JDK-8311500](https://bugs.openjdk.org/browse/JDK-8311500): StackWalker.getCallerClass() throws UOE if invoked reflectively (**Bug** - P3) ⚠️ Issue is not open.

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14773/head:pull/14773` \
`$ git checkout pull/14773`

`$ git checkout pull/14773` \
`$ git pull https://git.openjdk.org/jdk.git pull/14773/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14773`

`$ git pr show -t 14773`

</details>

<a href="https://git.openjdk.org/jdk/pull/14773.diff">https://git.openjdk.org/jdk/pull/14773.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14773#issuecomment-1621606231)

### DIFF
--- a/src/hotspot/share/prims/stackwalk.cpp
+++ b/src/hotspot/share/prims/stackwalk.cpp
@@ -165,7 +165,7 @@ BaseFrameStream* BaseFrameStream::from_current(JavaThread* thread, jlong magic,
 int StackWalk::fill_in_frames(jlong mode, BaseFrameStream& stream,
                               int max_nframes, int start_index,
                               objArrayHandle  frames_array,
-                              int& end_index, bool firstBatch, TRAPS) {
+                              int& end_index, bool is_first_batch, TRAPS) {
   log_debug(stackwalk)("fill_in_frames limit=%d start=%d frames length=%d",
                        max_nframes, start_index, frames_array->length());
   assert(max_nframes > 0, "invalid max_nframes");
@@ -213,7 +213,7 @@ int StackWalk::fill_in_frames(jlong mode, BaseFrameStream& stream,
     }
 
     if (!need_method_info(mode) && get_caller_class(mode) &&
-          firstBatch && index == start_index && method->caller_sensitive()) {
+          is_first_batch && index == start_index && method->caller_sensitive()) {
       ResourceMark rm(THREAD);
       THROW_MSG_0(vmSymbols::java_lang_UnsupportedOperationException(),
         err_msg("StackWalker::getCallerClass called from @CallerSensitive '%s' method",
--- a/src/hotspot/share/prims/stackwalk.cpp
+++ b/src/hotspot/share/prims/stackwalk.cpp
@@ -43,6 +43,7 @@
 #include "runtime/keepStackGCProcessed.hpp"
 #include "runtime/stackWatermarkSet.hpp"
 #include "runtime/vframe.inline.hpp"
+#include "classfile/systemDictionary.hpp"
 #include "utilities/formatBuffer.hpp"
 #include "utilities/globalDefinitions.hpp"
 
@@ -82,9 +83,10 @@ void BaseFrameStream::set_continuation(Handle cont) {
 
 JavaFrameStream::JavaFrameStream(JavaThread* thread, int mode, Handle cont_scope, Handle cont)
   : BaseFrameStream(thread, cont),
-   _vfst(cont.is_null()
+    _vfst(cont.is_null()
       ? vframeStream(thread, cont_scope)
-      : vframeStream(cont(), cont_scope)) {
+      : vframeStream(cont(), cont_scope)),
+    _needs_caller_sensitive_check(true) {
   _need_method_info = StackWalk::need_method_info(mode);
 }
 
@@ -143,6 +145,25 @@ BaseFrameStream* BaseFrameStream::from_current(JavaThread* thread, jlong magic,
   return stream;
 }
 
+static bool is_reflection_or_methodhandle_frame(Method* method, TRAPS) {
+  static InstanceKlass* method_accessor = nullptr;
+  static InstanceKlass* constructor_accessor = nullptr;
+  if (method_accessor == nullptr) {
+    Klass* k = SystemDictionary::resolve_or_fail(vmSymbols::reflect_MethodAccessor(), true, THREAD);
+    assert(k != nullptr, "Can't load MethodAccessor");
+    method_accessor = InstanceKlass::cast(k);
+    k = SystemDictionary::resolve_or_fail(vmSymbols::reflect_ConstructorAccessor(), true, THREAD);
+    assert(k != nullptr, "Can't load ConstructorAccessor");
+    constructor_accessor = InstanceKlass::cast(k);
+  }
+  return (method->method_holder() == vmClasses::reflect_Method_klass() ||
+    method->method_holder() == vmClasses::reflect_Constructor_klass() ||
+    method->method_holder()->is_subtype_of(method_accessor) ||
+    method->method_holder()->is_subtype_of(constructor_accessor) ||
+    // MethodHandle frames are not hidden and StackWalker::getCallerClass has to filter them out
+    method->method_holder()->name()->starts_with("java/lang/invoke"));
+}
+
 // Unpacks one or more frames into user-supplied buffers.
 // Updates the end index, and returns the number of unpacked frames.
 // Always start with the existing vfst.method and bci.
@@ -165,7 +186,7 @@ BaseFrameStream* BaseFrameStream::from_current(JavaThread* thread, jlong magic,
 int StackWalk::fill_in_frames(jlong mode, BaseFrameStream& stream,
                               int max_nframes, int start_index,
                               objArrayHandle  frames_array,
-                              int& end_index, bool is_first_batch, TRAPS) {
+                              int& end_index, TRAPS) {
   log_debug(stackwalk)("fill_in_frames limit=%d start=%d frames length=%d",
                        max_nframes, start_index, frames_array->length());
   assert(max_nframes > 0, "invalid max_nframes");
@@ -184,9 +205,10 @@ int StackWalk::fill_in_frames(jlong mode, BaseFrameStream& stream,
 
     if (method == nullptr) continue;
 
-    // skip hidden frames for default StackWalker option (i.e. SHOW_HIDDEN_FRAMES
-    // not set) and when StackWalker::getCallerClass is called
-    if (!ShowHiddenFrames && (skip_hidden_frames(mode) || get_caller_class(mode))) {
+    // Always skip hidden frames when StackWalker::getCallerClass is called or for the
+    // default StackWalker option (i.e. SHOW_HIDDEN_FRAMES not set) if ShowHiddenFrames is false.
+    // For StackWalker::getCallerClass also skip reflective/methodhandle frames.
+    if (get_caller_class(mode) || (skip_hidden_frames(mode) && !ShowHiddenFrames)) {
       if (method->is_hidden()) {
         LogTarget(Debug, stackwalk) lt;
         if (lt.is_enabled()) {
@@ -198,6 +220,17 @@ int StackWalk::fill_in_frames(jlong mode, BaseFrameStream& stream,
         }
         // We end a batch on continuation bottom to let the Java side skip top frames of the next one
         if (stream.continuation() != nullptr && method->intrinsic_id() == vmIntrinsics::_Continuation_enter) break;
+        continue;
+      }
+      if (get_caller_class(mode) && is_reflection_or_methodhandle_frame(method, THREAD)) {
+        LogTarget(Debug, stackwalk) lt;
+        if (lt.is_enabled()) {
+          ResourceMark rm(THREAD);
+          LogStream ls(lt);
+          ls.print("  reflective/methodhandle method: ");
+          method->print_short_name(&ls);
+          ls.cr();
+        }
         continue;
       }
     }
@@ -213,12 +246,19 @@ int StackWalk::fill_in_frames(jlong mode, BaseFrameStream& stream,
     }
 
     if (!need_method_info(mode) && get_caller_class(mode) &&
-          is_first_batch && index == start_index && method->caller_sensitive()) {
-      ResourceMark rm(THREAD);
-      THROW_MSG_0(vmSymbols::java_lang_UnsupportedOperationException(),
-        err_msg("StackWalker::getCallerClass called from @CallerSensitive '%s' method",
-                method->external_name()));
+        // if we're in "get_caller_class" mode, 'stream' is a "JavaFrameStream"
+        ((JavaFrameStream&)stream).needs_caller_sensitive_check()) {
+      // reflective and method handle frames have already been skipped in the previous step
+      if (method->caller_sensitive()) {
+        ResourceMark rm(THREAD);
+        THROW_MSG_0(vmSymbols::java_lang_UnsupportedOperationException(),
+          err_msg("StackWalker::getCallerClass called from @CallerSensitive '%s' method",
+                  method->external_name()));
+      } else {
+        ((JavaFrameStream&)stream).caller_sensitive_check_done();
+      }
     }
+
     // fill in StackFrameInfo and initialize MemberName
     stream.fill_frame(index, frames_array, methodHandle(THREAD, method), CHECK_0);
 
@@ -498,7 +538,7 @@ oop StackWalk::fetchFirstBatch(BaseFrameStream& stream, Handle stackStream,
   if (!stream.at_end()) {
     KeepStackGCProcessedMark keep_stack(THREAD);
     numFrames = fill_in_frames(mode, stream, frame_count, start_index,
-                               frames_array, end_index, true, CHECK_NULL);
+                               frames_array, end_index, CHECK_NULL);
     if (numFrames < 1) {
       THROW_MSG_(vmSymbols::java_lang_InternalError(), "stack walk: decode failed", nullptr);
     }
@@ -583,7 +623,7 @@ jint StackWalk::fetchNextBatch(Handle stackStream, jlong mode, jlong magic,
     stream.next(); // advance past the last frame decoded in previous batch
     if (!stream.at_end()) {
       int n = fill_in_frames(mode, stream, frame_count, start_index,
-                             frames_array, end_index, false, CHECK_0);
+                             frames_array, end_index, CHECK_0);
       if (n < 1 && !skip_hidden_frames(mode)) {
         THROW_MSG_(vmSymbols::java_lang_InternalError(), "doStackWalk: later decode failed", 0L);
       }
--- a/src/hotspot/share/prims/stackwalk.cpp
+++ b/src/hotspot/share/prims/stackwalk.cpp
@@ -165,7 +165,7 @@ BaseFrameStream* BaseFrameStream::from_current(JavaThread* thread, jlong magic,
 int StackWalk::fill_in_frames(jlong mode, BaseFrameStream& stream,
                               int max_nframes, int start_index,
                               objArrayHandle  frames_array,
-                              int& end_index, TRAPS) {
+                              int& end_index, bool firstBatch, TRAPS) {
   log_debug(stackwalk)("fill_in_frames limit=%d start=%d frames length=%d",
                        max_nframes, start_index, frames_array->length());
   assert(max_nframes > 0, "invalid max_nframes");
@@ -213,7 +213,7 @@ int StackWalk::fill_in_frames(jlong mode, BaseFrameStream& stream,
     }
 
     if (!need_method_info(mode) && get_caller_class(mode) &&
-          index == start_index && method->caller_sensitive()) {
+          firstBatch && index == start_index && method->caller_sensitive()) {
       ResourceMark rm(THREAD);
       THROW_MSG_0(vmSymbols::java_lang_UnsupportedOperationException(),
         err_msg("StackWalker::getCallerClass called from @CallerSensitive '%s' method",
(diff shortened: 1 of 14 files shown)


src/hotspot/share/prims/stackwalk.cpp
```diff
@@ -165,7 +165,7 @@ BaseFrameStream* BaseFrameStream::from_current(JavaThread* thread, jlong magic,
 int StackWalk::fill_in_frames(jlong mode, BaseFrameStream& stream,
                               int max_nframes, int start_index,
                               objArrayHandle  frames_array,
-                              int& end_index, TRAPS) {
+                              int& end_index, bool firstBatch, TRAPS) {
```
shipilev, line 168: The style is `is_first_batch`.

shipilev (commented):
Neither PR nor the bug describes what the root cause it, so it is very hard to see if patch makes sense. Please describe the problem and solution briefly?

simonis:
> Neither PR nor the bug describes what the root cause it, so it is very hard to see if patch makes sense. Please describe the problem and solution briefly?

Fair enough :)

I've put the details into the initial comment and also renamed the new parameter according to the HS coding conventions.

mlchung:
Thanks for catching this issue.    I agree that `Method::invoke` should be skipped the caller-sensitive test in this case but the fix isn't quite right.    The caller-sensitive test should apply in any batch.    For example, `CSM` calls `getCallerClass` reflectively, I think the stack would look like this:

```
java.lang.StackWalker::getCallerClass
java.lang.invoke.DirectMethodHandle$Holder::invokeStatic
java.lang.invoke.LambdaForm$MH/0x0000000800002c00::invoke
:
:
jdk.internal.reflect.DirectMethodHandleAccessor::invokeImpl
jdk.internal.reflect.DirectMethodHandleAccessor::invoke
java.lang.reflect.Method::invoke
CSM  <--------- caller-sensitive method and UOE should be thrown

```

In this case, UOE should be thrown.

src/hotspot/share/prims/stackwalk.cpp
```diff
@@ -498,7 +498,7 @@ oop StackWalk::fetchFirstBatch(BaseFrameStream& stream, Handle stackStream,
   if (!stream.at_end()) {
     KeepStackGCProcessedMark keep_stack(THREAD);
     numFrames = fill_in_frames(mode, stream, frame_count, start_index,
-                               frames_array, end_index, CHECK_NULL);
+                               frames_array, end_index, true, CHECK_NULL);
```
dholmes-ora, line 501: Could you annotate the new argument please ie. `true /* first batch */` and `false /* not first batch */`. Thanks.

simonis:
> Thanks for catching this issue. I agree that `Method::invoke` should be skipped the caller-sensitive test in this case but the fix isn't quite right. The caller-sensitive test should apply in any batch. For example, `CSM` calls `getCallerClass` reflectively, I think the stack would look like this:
> 
> ```
> java.lang.StackWalker::getCallerClass
> java.lang.invoke.DirectMethodHandle$Holder::invokeStatic
> java.lang.invoke.LambdaForm$MH/0x0000000800002c00::invoke
> :
> :
> jdk.internal.reflect.DirectMethodHandleAccessor::invokeImpl
> jdk.internal.reflect.DirectMethodHandleAccessor::invoke
> java.lang.reflect.Method::invoke
> CSM  <--------- caller-sensitive method and UOE should be thrown
> ```
> 
> In this case, UOE should be thrown.

Hi @mlchung ,
Thanks for your comments. I think you're right. More specifically, the caller-sensitive test should be applied to the first non-reflective, non-methodhandle frame, independent on the batch in which it appears.

I'll add a new test for the case you've brought up and update the fix accordingly.

Best regards,
Volker

simonis:
By the way, great that you raised this issue @mlchung :) While writing a test for it I realized the the current implementation is *already failing* to trow an `UnsupportedOperationException` if `StackWalker::getCallerClass()` is called reflectively from a `@CallerSensitive` method (because the caller sensitive method is not at the first method at `index == start_index`):
```
[0,161s][debug][stackwalk] Start walking: mode 6 skip 0 frames batch size 6
[0,161s][debug][stackwalk] 
[0,161s][debug][stackwalk]   skip  java.lang.StackStreamFactory$AbstractStackWalker::callStackWalk
[0,161s][debug][stackwalk]   skip  java.lang.StackStreamFactory$AbstractStackWalker::beginStackWalk
[0,161s][debug][stackwalk]   skip  java.lang.StackStreamFactory$AbstractStackWalker::walkHelper
[0,161s][debug][stackwalk]   skip  java.lang.StackStreamFactory$AbstractStackWalker::walk
[0,161s][debug][stackwalk]   skip  java.lang.StackStreamFactory$CallerClassFinder::findCaller
[0,161s][debug][stackwalk]   skip  java.lang.StackWalker::getCallerClass
[0,161s][debug][stackwalk] fill_in_frames limit=6 start=2 frames length=8
[0,161s][debug][stackwalk]   hidden method:  java.lang.invoke.DirectMethodHandle$Holder::invokeSpecial
[0,161s][debug][stackwalk]   hidden method:  java.lang.invoke.LambdaForm$MH/0x00000007c0084c00::invoke
[0,161s][debug][stackwalk]   hidden method:  java.lang.invoke.Invokers$Holder::invokeExact_MT
[0,161s][debug][stackwalk]   hidden method:  java.util.CallerSensitiveMethod$$InjectedInvoker/0x00000007c0085000::reflect_invoke_V
[0,161s][debug][stackwalk]   hidden method:  java.lang.invoke.DirectMethodHandle$Holder::invokeStatic
[0,161s][debug][stackwalk]   hidden method:  java.lang.invoke.LambdaForm$MH/0x00000007c0085400::invokeExact_MT
[0,161s][debug][stackwalk]   hidden method:  jdk.internal.reflect.DirectMethodHandleAccessor::invokeImpl
[0,161s][debug][stackwalk]   2: frame method:  jdk.internal.reflect.DirectMethodHandleAccessor::invoke bci=24
[0,161s][debug][stackwalk]   2: done frame method:  jdk.internal.reflect.DirectMethodHandleAccessor::invoke
[0,161s][debug][stackwalk]   3: frame method:  java.lang.reflect.Method::invoke bci=90
[0,161s][debug][stackwalk]   3: done frame method:  java.lang.reflect.Method::invoke
[0,161s][debug][stackwalk]   4: frame method:  java.util.CallerSensitiveMethod::getCallerClass bci=110
[0,161s][debug][stackwalk]   4: done frame method:  java.util.CallerSensitiveMethod::getCallerClass
[0,161s][debug][stackwalk]   5: frame method:  ReflectiveGetCallerClassTest::main bci=38
[0,161s][debug][stackwalk]   5: done frame method:  ReflectiveGetCallerClassTest::main
[0,161s][debug][stackwalk] fill_in_frames done frames_decoded=4 at_end=1
doStackWalk: skip 0 start 2 end 6
  frame 2: class jdk.internal.reflect.DirectMethodHandleAccessor
  skip: frame 2 class jdk.internal.reflect.DirectMethodHandleAccessor
  next frame at 2: class jdk.internal.reflect.DirectMethodHandleAccessor (origin 2 fence 6)
  skip: frame 3 class java.lang.reflect.Method
  next frame at 3: class java.lang.reflect.Method (origin 3 fence 6)
  next frame at 4: class java.util.CallerSensitiveMethod (origin 4 fence 6)
  next frame at 5: class ReflectiveGetCallerClassTest (origin 5 fence 6)
CallerSensitiveMethod::getCallerClass() called from class ReflectiveGetCallerClassTest
Exception in thread "main" java.lang.RuntimeException: Expected UnsupportedOperationException when calling StackWalker::getCallerClass() from @CallerSensitive method reflectively
	at java.base/java.util.CallerSensitiveMethod.getCallerClass(CallerSensitiveMethod.java:79)
	at ReflectiveGetCallerClassTest.main(ReflectiveGetCallerClassTest.java:84)
```

The existing test under `test/jdk/java/lang/StackWalker/CallerSensitiveMethod/src/java.base/java/util/CSM.java` for calls from `@CallerSensitive` methods only handles direct but not reflective calls.

I'll post a patch which will hopefully fix both cases later today.

mlchung:
Yes that's the bug.   `index == start_index` assumes the first frame is the caller of `getCallerClass` as it misses the reflective case.

simonis:
As discussed before, the test for a `@CallerSensitive` method must be applied to the first non-reflective, non-methodhandle frame below `StackWalker::getCallerClass()`, independent from the batch in which it appears. Because we can not predict which batch this will be, I've removed the previous solution which added a special flag to distinguish the first from the follow up batches. Instead I've added a new flag `_needs_caller_sensitive_check` to the `BaseFrameStream` class because every invocation of `StackWalker::getCallerClass()` will always use a single instance of `BaseFrameStream`.

Currently the check for `@CallerSensitive` methods can only be done in the VM because for `StackWalker::getCallerClass()` we only pass the `Class` of each stack frame back to Java but not the corresponding `Method`. On the other hand, we currently skip reflective and methodhandle frames in Java (i.e. with `isReflectionFrame()` and `isMethodHandleFrame()`). In order to avoid the need to pass the stack frame methods back to Java I've therefore implemented (and combined) the the check for reflective and methodhandle frames in the VM (see `is_reflection_or_methodhandle_frame()`). Because `isMethodHandleFrame()` was only used in `ClassBuffer::consumeFrames()` it can be removed. I think we can also skip the remaining test for reflective frames in `AbstractStackWalker::peekFrame()` for the `getCallerClass()` case (because that check is now already done in the VM) but I'll leave that for [JDK-8285447](https://bugs.openjdk.org/browse/JDK-8285447).

I've also moved the `ReflectiveGetCallerClassTest.java` test under `java/lang/StackWalker/CallerSensitiveMethod` such that I can use the `java/util/CSM.java` class which gets injected into the base module for the new `@CallerSensitive` test.

Finally I realized that some `StackWalker` tests fail if we run with `-javaoption:-XX:+ShowHiddenFrames`. Ideally these tests should be fixed to make them agnostic to `ShowHiddenFrames` but for now I've added a `@requires vm.opt.ShowHiddenFrames == false` tag to exclude them if ran with `-XX:+ShowHiddenFrames`.

src/hotspot/share/prims/stackwalk.cpp
simonis: I've removed these arguments in the latest version.

mlchung:
I wonder if we should move the check to throw UOE if called by caller-sensitive method in Java as a general guidance to implement the runtime in Java if desirable.   That means it requires the VM to fill not only the class in the buffer but also need a flag to indicate the method is caller-sensitive or not.   It's a tradeoff of the buffer size.    The common case for `getCallerClass` is being invoked statically and should find the class from the first batch.   Only if it's invoked reflectively and if filtered in the Java,  it'll fetch more batches and hence the performance would not be as fast as filtered in VM but I think that's okay since it's uncommon.

Would you have cycle to implement this alternative and determine any performance impact to common cases?  Then evaluate this further.

The benchmark is at `test/micro/org/openjdk/bench/java/lang/StackWalkBench.java`.

src/hotspot/share/prims/stackwalk.cpp
```diff
@@ -143,6 +145,25 @@ BaseFrameStream* BaseFrameStream::from_current(JavaThread* thread, jlong magic,
...
+    k = SystemDictionary::resolve_or_fail(vmSymbols::reflect_ConstructorAccessor(), true, THREAD);
+    assert(k != nullptr, "Can't load ConstructorAccessor");
+    constructor_accessor = InstanceKlass::cast(k);
+  }
+  return (method->method_holder() == vmClasses::reflect_Method_klass() ||
+    method->method_holder() == vmClasses::reflect_Constructor_klass() ||
+    method->method_holder()->is_subtype_of(method_accessor) ||
+    method->method_holder()->is_subtype_of(constructor_accessor) ||
+    // MethodHandle frames are not hidden and StackWalker::getCallerClass has to filter them out
+    method->method_holder()->name()->starts_with("java/lang/invoke"));
```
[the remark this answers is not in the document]
simonis: Thanks, you're right, this is a copy/paste error. It currently doesn't make any difference, until somebody introduces a new package in `java.lang` that starts with `invoke*` :)

Fixed locally, will be in the next commit.

simonis:
> I wonder if we should move the check to throw UOE if called by caller-sensitive method in Java as a general guidance to implement the runtime in Java if desirable. That means it requires the VM to fill not only the class in the buffer but also need a flag to indicate the method is caller-sensitive or not. It's a tradeoff of the buffer size. The common case for `getCallerClass` is being invoked statically and should find the class from the first batch. Only if it's invoked reflectively and if filtered in the Java, it'll fetch more batches and hence the performance would not be as fast as filtered in VM but I think that's okay since it's uncommon.
> 
> Would you have cycle to implement this alternative and determine any performance impact to common cases? Then evaluate this further.
> 
> The benchmark is at `test/micro/org/openjdk/bench/java/lang/StackWalkBench.java`.

I'm open to investigate other approaches, but this is a real bug and I'd first like to fix it before thinking about implementation improvements. Notice also, that this issue impacts 17 and 21 as well, so the fix should be easily and quickly downportable. We already have [JDK-8285447](https://bugs.openjdk.org/browse/JDK-8285447) for performance improvements of `getCallerClass()` (which I'm also working on).

I'd therefore suggest to first fix this and postpone further refactorings and improvements to [JDK-8285447](https://bugs.openjdk.org/browse/JDK-8285447) or another follow-up issue.

mlchung:
I expect `getCallerClass` is rarely called via reflection.  JDK-8210375 was reported 5 years ago but no customer reports this issue since then.    This bug is not critical to be fixed immediately and so we should look into a proper solution.

I do have a question if filtering in the VM is the right thing to do.   We need to consider the alternatives and understand why this solution is chosen.

simonis:
We actually have two problems here:
1. If called refelctively, `getCallerClass()` can throw an UOE even if it was **not** called from a `@CallerSensitive` method (see inital test case).
2. If called reflectively from a  `@CallerSensitive` method, `getCallerClass()` will currently not throw a UOE as expected (see extended test in this PR).

`getCallerClass()` **is** performance sensitive and we want to improve its performance rather than slow it down (see [JDK-8285447](https://bugs.openjdk.org/browse/JDK-8285447)). I think performance-wise it would be better to do all the filtering in the VM, where we have all the required information at hand and minimize the amount of data that needs to be passed between Java and the VM.

Before starting to implement a new version of the fix which moves all the checks to Java, I'd like to hear some more opinions about whether we agree to move all the filtering and checks from the VM to Java (even at the cost of performance regressions) or if we better want to go the other way and do all the filtering/checks in the JVM. @shipilev, @dholmes-ora, @dfuch, @AlanBateman - what do you think?

mlchung:
@simonis  please see #15370 to show the alternative fix for this issue and do the filtering in Java.  The benchmark shows about 30-60 ns slowdown.  I think it's acceptable with the benefit of keeping the filtering in the libraries rather than in VM.

PR #14776, opened 2023-07-05
JDK-8311264: JavaDoc index comparator is not transitive

hns:
Please review a fix to make the comparator used by the JavaDoc index pages to be transitive (see JBS issue for a description and example of the problem).

I fix the bug by creating method `getIndexElementKey(Element)` to extract the key for comparing elements in `Comparator.makeIndexElementComparator`, and allowing `IndexBuilder` to reuse that extractor method when comparing element index items to search tag index items. The rationale for this approach was to preserve the current order for sorting elements in the index, and to keep the change simple (considering possible backport to 21).

In the process I also simplified some parts of the code a bit (simpler logic to compare names in element comparator, no need for composite comparator for classes-only index). 

For the test, I added various elements and search tags that trigger the condition to the existing `TestIndex` page and replaced `checkOutput` with `checkOrder` to test the order of index page items.

---------
### Progress

### Issue
 * [JDK-8311264](https://bugs.openjdk.org/browse/JDK-8311264): JavaDoc index comparator is not transitive (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14776/head:pull/14776` \
`$ git checkout pull/14776`

`$ git checkout pull/14776` \
`$ git pull https://git.openjdk.org/jdk.git pull/14776/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14776`

`$ git pr show -t 14776`

</details>

<a href="https://git.openjdk.org/jdk/pull/14776.diff">https://git.openjdk.org/jdk/pull/14776.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14776#issuecomment-1622072326)

### DIFF
--- a/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/Comparators.java
+++ b/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/Comparators.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2021, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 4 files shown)


jonathan-gibbons (approved):
Seems reasonable, and cleaner

pavelrappo:
I can see that I'm late to the party. FWIW, the change looks good to me.

There's probably one additional problem that predates this PR. I see that Comparators.makeIndexElementComparator breaks ties by comparing fully qualified names. Although unlikely to happen in practise, in theory it might break when one has a static nested class and a class in the default package (hence not applicable to JDK) that share FQN. Here's an example I used for one of our tests earlier:

https://github.com/openjdk/jdk/blob/0616648c59215d001211423402c6444ce228f01e/test/langtools/jdk/javadoc/doclet/testThrowsInheritanceMatching/TestExceptionTypeMatching.java#L350-L396

PR #14780, opened 2023-07-06
8308762: Metaspace leak with Instrumentation.retransform

jpbempel:
Fix a small leak in constant pool merging during retransformation of a class. If this class has a catch block with `Throwable`, the class `Throwable` is pre-resolved in the constant pool, while all the other classes are in a unresolved state. So the constant pool merging process was considering the entry with pre-resolved class as different compared to the destination and create a new entry. We now try to consider it as equal specially for Methodref/Fieldref.

---------
### Progress

### Issue
 * [JDK-8308762](https://bugs.openjdk.org/browse/JDK-8308762): Metaspace leak with Instrumentation.retransform (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14780/head:pull/14780` \
`$ git checkout pull/14780`

`$ git checkout pull/14780` \
`$ git pull https://git.openjdk.org/jdk.git pull/14780/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14780`

`$ git pr show -t 14780`

</details>

<a href="https://git.openjdk.org/jdk/pull/14780.diff">https://git.openjdk.org/jdk/pull/14780.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14780#issuecomment-1623010432)


src/hotspot/share/oops/constantPool.cpp
```diff
@@ -1276,6 +1276,33 @@ void ConstantPool::unreference_symbols() {
   }
 }
 
+// Returns true if the current mismatch is due to a resolved/unresolved
+// class pair. Otherwise, returns false.
+bool ConstantPool::is_unresolved_class_mismatch(int index1,
+  const constantPoolHandle& cp2, int index2) {
```
dholmes-ora, line 1282: Style nit: please line up parameters "const" under "int"

src/hotspot/share/oops/constantPool.cpp
```diff
@@ -1276,6 +1276,33 @@ void ConstantPool::unreference_symbols() {
...
+  jbyte t2 = cp2->tag_at(index2).value();
+  if (t2 != JVM_CONSTANT_Class && t2 != JVM_CONSTANT_UnresolvedClass) {
+    return false;  // wrong entry type; not our special case
+  }
+
+  if (t1 == t2) {
+    return false;  // not a mismatch; not our special case
+  }
+
+  char *s1 = klass_name_at(index1)->as_C_string();
```
dholmes-ora, line 1298: `as_C_string` needs an active `ResourceMark` - not sure where it is coming from even in the existing code. There should at least be a comment that an active ResourceMark is needed.

src/hotspot/share/oops/constantPool.cpp
```diff
@@ -1276,6 +1276,33 @@ void ConstantPool::unreference_symbols() {
   }
 }
 
+// Returns true if the current mismatch is due to a resolved/unresolved
+// class pair. Otherwise, returns false.
+bool ConstantPool::is_unresolved_class_mismatch(int index1,
```
dholmes-ora, line 1281: Has this been moved verbatim from jvmtiRedefineClasses.cpp?

There are a couple of style nits with that existing code that don't fit this file:
- parameters should line up ie. const under int
- no comment on the closing brace of the method body.

dholmes-ora (commented):
Sorry I'm having a lot of trouble trying to understand the fix in the context of the problem description as outlined in the bug report. The issue pertained only to the treatment of Throwable due to it being pre-resolved by the verifier, but your fix is looking at Field and MethodRefs ??

There are also remaining comments about resolved and unresolved class entries deliberately not being considered the same.

jpbempel:
> Sorry I'm having a lot of trouble trying to understand the fix in the context of the problem description as outlined in the bug report. The issue pertained only to the treatment of Throwable due to it being pre-resolved by the verifier, but your fix is looking at Field and MethodRefs ??

For the Klass in itself, there is this method `is_unresolved_class_mismatch` that compares it correctly if entry differs by resolved and unresolved state. Except that for MethodRef and FieldRef, the Klass here is also compared but strictly without taking into account the already resolved state (in this case of the Throwable). So that's why I am adding the call to `is_unresolved_class_mismatch` for those cases.

> There are also remaining comments about resolved and unresolved class entries deliberately not being considered the same.

Sorry I don't understand what you mean, can you elaborate please?

src/hotspot/share/oops/constantPool.cpp
jpbempel: > Has this been moved verbatim from jvmtiRedefineClasses.cpp?

yes

> There are a couple of style nits with that existing code that don't fit this file:
> 
> * parameters should line up ie. const under int
> * no comment on the closing brace of the method body.

Will adjust accordingly

src/hotspot/share/oops/constantPool.cpp
```diff
@@ -1276,6 +1276,33 @@ void ConstantPool::unreference_symbols() {
...
+
+  if (t1 == t2) {
+    return false;  // not a mismatch; not our special case
+  }
+
+  char *s1 = klass_name_at(index1)->as_C_string();
+  char *s2 = cp2->klass_name_at(index2)->as_C_string();
+  if (strcmp(s1, s2) != 0) {
+    return false;  // strings don't match; not our special case
+  }
```
coleenp, line 1302: klass_name_at() returns a Symbol* that's in the SymbolTable so this never needed to do a strcmp.  It can test for equality.

coleenp (changes requested):
I wonder if compare_entry_to can just handle this case directly and not have this function.

src/hotspot/share/oops/constantPool.cpp
```diff
@@ -1331,6 +1358,7 @@ bool ConstantPool::compare_entry_to(int index1, const constantPoolHandle& cp2,
     int recur1 = uncached_klass_ref_index_at(index1);
     int recur2 = cp2->uncached_klass_ref_index_at(index2);
     bool match = compare_entry_to(recur1, cp2, recur2);
+    match |= is_unresolved_class_mismatch(recur1, cp2, recur2);
```
dholmes-ora, line 1361: This is changing the definition of a "match" such that in all cases an unresolved class entry is considered equal to a resolved class entry - but only for these Field and MethodRefs. I don't see how this connects back to the original problem. Nor do I see why this is alway a correct thing to do.

dholmes-ora:
> Sorry I don't understand what you mean, can you elaborate please?

In `ConstantPool::compare_entry_to` it states:
```
  if (t1 != t2) {
    // Not the same entry type so there is nothing else to check. Note
    // that this style of checking will consider resolved/unresolved
    // class pairs as different.
    // From the ConstantPool* API point of view, this is correct
    // behavior. See VM_RedefineClasses::merge_constant_pools() to see how this
    // plays out in the context of ConstantPool* merging.
    return false;
  }
```
and in ` VM_RedefineClasses::merge_constant_pools()` it says
```
 // Pass 0:
    // The old_cp is copied to *merge_cp_p; this means that any code
    // using old_cp does not have to change. This work looks like a
    // perfect fit for ConstantPool*::copy_cp_to(), but we need to
    // handle one special case:
    // - revert JVM_CONSTANT_Class to JVM_CONSTANT_UnresolvedClass
// This will make verification happy.
```
these comments both pertain to the problem at hand:  the merge reverts the resolved class entry to be unresolved; unresolved entries are not considered matches for resolved ones, hence we grow the constant pool.

Your fix is addressing the second part of the issue by forcing a match, but as per my other comment, it is not clear to me how the placement of that fix addresses the issue originally reported with the catch block, nor is it obvious to me that we should always consider resolved and unresolved class entries to be equal.

coleenp:
ConstantPool merging is still really complicated and JVM_CONSTANT_Class/UnresolvedClass (and UnresolvedClassInError) may have changed to not track with what RedefineClasses does.  If I'm reading this correctly, we are always calling find_matching_entry() and compare_entries_to() between the merged_cp and scratch_cp.  Since we unresolve classes in merge_cp, and scratch_cp hasn't resolved anything yet, how are we getting this mismatch?  Can you post a stack trace for us?

src/hotspot/share/oops/constantPool.cpp
jpbempel: I agree that it may be not the right place to call is_unresolved_class_mismatch method because we capture all cases calling compare_entry_to. For a JVM_CONSTANT_Class, compare_entry_to is recovered by the call to is_unresolved_class_mismatch in VM_RedefineClasses::merge_constant_pools. The idea is to do the same for other reference to the class in FieldRefs and (Interface)MethodRefs because otherwise the entry does not match with pre-resolved Throwable class.

jpbempel:
> ConstantPool merging is still really complicated and JVM_CONSTANT_Class/UnresolvedClass (and UnresolvedClassInError) may have changed to not track with what RedefineClasses does. If I'm reading this correctly, we are always calling find_matching_entry() and compare_entries_to() between the merged_cp and scratch_cp. Since we unresolve classes in merge_cp, and scratch_cp hasn't resolved anything yet, how are we getting this mismatch? Can you post a stack trace for us?

I have try to explain in the comment to David above. Tell me if there is something that is still not clear.

coleenp:
Hi, I now see why there's a resolved class in the scratch constant pool (ie the constant pool for the new class bytes).  It looks like javac has preresolved it for us, which you said but I didn't know why.  The merged constant pool always has an unresolved class copy of that class entry, because we create the merged constant pool with all unresolved classes.

I now believe that ConstantPool::compare_entry_to should always compare class and resolved class as equivalent.  Since the klass_name_at() function is the same call for both unresolved and resolved class, you could change the tag at the beginning of compare_entry_to() to JVM_CONSTANT_UnresolvedClass if it's resolved, like we do for error classes.  Then remove the is_unresolved_class_mismatch function entirely (the comment at that call might be useful to explain why you're changing the tag in compare_entry_to).  This code is only used for RedefineClasses, but even if it wasn't this comparison is the right thing to do.  For special measure, you could #if INCLUDE_JVMTI around this and the operand functions that call it or file an RFE to do so for further cleanup (compare_entry_to, find_matching_entry, find_matching_operand, compare_operand_to).

dholmes-ora:
>  It looks like javac has preresolved it for us, which you said but I didn't know why.

The preresolving is done by the verifier. https://bugs.openjdk.org/browse/JDK-8308762?focusedCommentId=14584707&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14584707

coleenp:
Also there is a nice test harness for class redefinition in the test/hotspot/jtreg/serviceability/jvmti/RedefineClasses tests that you might be able to use to add a test for this.

jpbempel:
Thanks @coleenp for the hints about fixing this issue.
Please review again the new changes that include:
 - jtreg test that reproduces the issue
 - revert the unresolved class state if the current entry is resolved
 - remove `is_unresolved_class_mismatch` method as useless now.

Thanks
(sorry to the force push, old habit)

src/hotspot/share/oops/constantPool.cpp
```diff
@@ -1286,6 +1286,13 @@ bool ConstantPool::compare_entry_to(int index1, const constantPoolHandle& cp2,
   jbyte t1 = tag_at(index1).non_error_value();
   jbyte t2 = cp2->tag_at(index2).non_error_value();
 
+  if (t1 == JVM_CONSTANT_Class) {
+    // Some classes are pre-resolved (like Throwable) which may lead to
+    // consider it as a different entry. we then revert them back temporarily
+    // to ensure proper comparison.
+    t1 = JVM_CONSTANT_UnresolvedClass;
+  }
+
```
sspitsyn, line 1295: All consequences of this change are not clear to me yet.
The lines 1307-1314 become not needed anymore.
Also, should the same be done for t2 as well?

coleenp: t2 could be a resolved class, and can be compared with unresolved class because the function klass_name_at() works for both.  It might be a good idea to change both of them though, although not necessary imo.
You're right 1307-1314 are not reached anymore.  Neither is the ClassIndex case but not to remove as part of this change.

coleenp (approved):
Sorry I didn't see this update.  The change looks good to me.  You could do the cleanup that Serguei suggests with t2 and removing JVM_CONSTANT_Class case.  I thought maybe they should be left in case we want to generalize this compare_entry_to function someday, that's why I didn't suggest removing it.

test/hotspot/jtreg/serviceability/jvmti/RedefineClasses/RedefineLeakThrowable.java
```diff
@@ -0,0 +1,127 @@
+/*
+ * Copyright (c) 2016, 2023, Oracle and/or its affiliates. All rights reserved.
```
coleenp, line 2: Please remove the 2016 copyright on the new test.

test/hotspot/jtreg/serviceability/jvmti/RedefineClasses/RedefineLeakThrowable.java
```diff
@@ -0,0 +1,127 @@
...
+public class RedefineLeakThrowable {
+    static class Tester {
+        void test() {
+            try {
+                int i = 42;
+            } catch (Throwable t) {
+                t.printStackTrace();
+            }
+        }
+    }
```
coleenp, line 58: If you look in that directory, there's an even simpler way to redefine classes without the Transformer and agent code.  It looks like this:  RedefineClassHelper.redefineClass(RedefineRunningMethods_B.class, evenNewerB);

coleenp (changes requested):
I have a suggestion for your test and copyright fix.

src/hotspot/share/prims/jvmtiRedefineClasses.cpp
```diff
@@ -1700,14 +1671,6 @@ bool VM_RedefineClasses::merge_constant_pools(const constantPoolHandle& old_cp,
...
         // found a match at the same index so nothing more to do
         continue;
-      } else if (is_unresolved_class_mismatch(scratch_cp, scratch_i,
-                                              *merge_cp_p, scratch_i)) {
-        // The mismatch in compare_entry_to() above is because of a
-        // resolved versus unresolved class entry at the same index
-        // with the same string value. Since Pass 0 reverted any
-        // class entries to unresolved class entries in *merge_cp_p,
-        // we go with the unresolved class entry.
-        continue;
```
coleenp, line 1710: I'm sorry for the piecemeal review.   There's another comment that mentions this comment that should be removed.  It starts with this:

-      // The find_matching_entry() call above could fail to find a match

jpbempel:
@coleenp I have made the changes: cleanup code and rewrite the unit test

jpbempel:
Hey @coleenp could you review my last changes?

jpbempel:
Please @coleenp review my last changes

test/hotspot/jtreg/serviceability/jvmti/RedefineClasses/RedefineLeakThrowable.java
```diff
@@ -0,0 +1,67 @@
...
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/*
+ * @test
+ # @bug 8308762
+ * @library /test/lib
+ * @summary Test that redefinition of class containing Throwable refs does not leak constant pool
```
dholmes-ora, line 28: Exactly how is this test tracking whether there is a leak or not? Is it simply setting metaspace size small enough that the 500 iterations would exhaust metaspace if there were a leak?

dholmes-ora (commented):
Based on the previous discussion and comments this simple change seems quite reasonable to me.

One query on the test though.

Thanks

test/hotspot/jtreg/serviceability/jvmti/RedefineClasses/RedefineLeakThrowable.java
jpbempel: > Is it simply setting metaspace size small enough that the 500 iterations would exhaust metaspace if there were a leak?

yes exactly. with a leak  you end up with `OutOfMemoryError: metaspace` running the test

coleenp (approved):
I'm sorry for the delay in reviewing this.  This looks great!

dholmes-ora (approved):
Looks good. Thanks.

jpbempel:
Thanks!
/integrate

PR #10085, opened 2022-08-30
JDK-8293114: GC should trim the native heap

tstuefe:
(*Updated 2023-07-05 to reflect the current state of the patch*)

This RFE adds the option to auto-trim the Glibc heap as part of the GC cycle. If the VM process suffered high temporary malloc spikes (regardless of whether from JVM- or user code), this could recover significant amounts of memory.

We discussed this a year ago [1], but the item got pushed to the bottom of my work pile, therefore, it took longer than I thought.

### Motivation

The Glibc is reluctant to return memory to the OS, more so than other allocators. Temporary malloc spikes often carry over as permanent RSS increase. Note that C-heap retention is difficult to observe. Since it is freed memory, it won't appear in NMT; it is just a part of RSS.

This is, effectively, caching, and a performance tradeoff by the glibc. It makes a lot of sense with applications that cause high traffic on the C-heap (the typical native application). The JVM, however, clusters allocations and for a lot of use cases rolls its own memory management via mmap. And app's malloc load can fluctuate wildly, with temporary spikes and long idle periods.

To help, Glibc exports an API to trim the C-heap: `malloc_trim(3)`. With JDK 18 [2], SAP contributed a new jcmd command to *manually* trim the C-heap on Linux. This RFE adds a complementary way to trim automatically.

#### Is this even a problem?

Yes. 

The JVM clusters most native memory allocations and satisfies them with mmap. But there are enough C-heap allocations left to cause malloc spikes that are subject of memory retention. Note that one example are hotspot arenas themselves.

But many cases of high memory retention in Glibc I have seen in third-party JNI code. Libraries allocate large buffers via malloc as temporary buffers. In fact, since we introduced the jcmd "System.trim_native_heap", some of our customers started to call this command periodically in scripts to counter these issues.

### How trimming works

Trimming is done via `malloc_trim(2)`. `malloc_trim` will iterate over all arenas and trim each one subsequently. While doing that, it will lock the arena, which may cause some (but not all) subsequent actions on the same arenas to block. glibc also trims automatically on free, but that is very limited (see https://github.com/openjdk/jdk/pull/10085#issuecomment-1619638641 for details).

`malloc_trim` offers almost no way to control its behavior; in particular, no way to limit its runtime. Its runtime will depend on the size of the reclaimed memory. Not reclaiming anything is very fast (sub-ms). Reclaiming very large memory sections (many GB) may take considerably longer.

### When to trim?

[...]

PR #14781, opened 2023-07-06
JDK-8293114: JVM should trim the native heap

tstuefe:
This is a continuation of https://github.com/openjdk/jdk/pull/10085. I closed https://github.com/openjdk/jdk/pull/10085 because it had accumulated too much comment history and got confusing. For a history of this issue, see previous discussions [1] and the comment section of 10085.
 
---------------

This RFE adds the option to trim the Glibc heap periodically. This can recover a significant memory footprint if the VM process suffers from high-but-rare malloc spikes. It does not matter who causes the spikes: the JDK or customer code running in the JVM process.

### Background:

The Glibc is reluctant to return memory to the OS. Temporary malloc spikes often carry over as permanent RSS increase. Note that C-heap retention is difficult to observe. Since it is freed memory, it won't appear in NMT; it is just a part of RSS.

This is, effectively, caching - a performance tradeoff by the glibc. It makes a lot of sense with applications that cause high traffic on the C-heap. The JVM, however, clusters allocations and often rolls its own memory management based on virtual memory for many of its use cases.

To manually trim the C-heap, Glibc exposes `malloc_trim(3)`. With JDK 18 [2], we added a new jcmd command to *manually* trim the C-heap on Linux (`jcmd System.trim_native_heap`). We then observed customers running this command periodically to slim down process sizes of container-bound jvms. That is cumbersome, and the JVM can do this a lot better - among other things because it knows best when *not* to trim.

#### GLIBC internals

The following information I took from the glibc source code and experimenting.

##### Why do we need to trim manually? Does the Glibc not trim on free?

Upon `free()`, glibc may return memory to the OS if:
- the returned block was mmap'ed
- the returned block was not added to tcache or to fastbins
- the returned block, possibly merged with its two immediate neighbors, had they been free, is larger than FASTBIN_CONSOLIDATION_THRESHOLD (64K) - in that case:
  a) for the main arena, glibc attempts to lower the brk()
  b) for mmap-ed heaps, glibc attempts to completely unmap or shrink the heap.
In both cases, (a) and (b), only the top portion of the heap is reclaimed. "Holes" in the middle of other in-use chunks are not reclaimed.

So: glibc *may* automatically reclaim memory. In normal configurations, with typical C-heap allocation granularity, it is unlikely.

To increase the chance of auto-reclamation happening, one can do one or more things:
- a) increase allocation sizes
- b) limit mallocs to very few threads, ideally just one 
- c) set MALLOC_ARENA_MAX=1 
- d) set the `glibc.malloc.trim_threshold` to a very low value, e.g., 1

But:

- (a) is not possible for foreign code; not that even within hotspot, where we cluster allocations using hotspot arenas, the typical arena chunk size is too small to be auto-reclaimed
- (b) may just not be feasible
- (c) is *terrible* for performance since many C-Heap operations will compete over the lock of that single arena
- (d) works if either (b) or (c) are in place and if the released block happens to border the top of the arena. And it also costs performance.

The JVM only has limited influence on (a), none on (b), (c) is a really bad idea, and hence (d) often does little. That mirrors my practical experiences.

##### How does `malloc_trim()` differ from trimming on free() ?

`malloc_trim()`, will look for holes that are larger than a page; so it limits itself not to just reclaiming memory at the top of the arena. It will then `madvise(MADV_DONTNEED)` those holes. It does that for every arena.

##### What are the cons of calling `malloc_trim()`?

`malloc_trim()` cannot be interrupted. Once it runs, it runs. The runtime of `malloc_trim()` is not predictable. If there is nothing to reclaim, it is very fast (sub-ms). If there is a lot to reclaim (e.g. >32GB), I saw times of up to 800ms.

Moreover, `malloc_trim`, while trimming each arena, locks the arena. That may lock out concurrent C-heap operations in the thread that uses this arena. Note, however, that this is rare since many operations will be satisfied from the tcache and therefore don't lock.

##### What about the `pad` parameter for `malloc_trim()`

I found it has very little effect. It only affects how many bytes are preserved at the top of the main arena. It does not affect other arenas, nor does it affect how much space malloc_trim reclaims by releasing "holes", which is the main part of memory release.

### The Patch

Patch adds new options (experimental):

```
-XX:+GCTrimNativeHeap
-XX:GCTrimNativeHeapInterval=<seconds> (defaults to 60)
```

`GCTrimNativeHeap` is off by default. If enabled, it will cause the VM to trim the native heap periodically. The period is defined by `GCTrimNativeHeapInterval`.

Periodic trimming is done in its own thread. We cannot burden the ServiceThread, since the runtime of trims is unpredictable. 

The patch also adds a way to suspend trimming temporarily; if suspended, no trims will start, but ongoing trims will still finish.

The patch uses this mechanism to suspend trimming during GC STW phases and whenever we are about to do bulk C-heap operations (e.g. deleting deflated monitors).

### Examples:

This is an artificial test that causes two high malloc spikes with long idle periods. 

(yellow) NMT shows two spikes for malloc'ed memory;

(red) RSS of the baseline JVM shows that we reach a maximum and then never recover. This is the glibc retaining the free'd memory.

(blue) RSS of the patched JVM shows that we recover RSS in steps by doing periodic C-heap trimming.

alloc-test

(See here for parameters: [run script](http://cr.openjdk.java.net/~stuefe/other/autotrim/run-all.sh) )

### Tests

Tested older Glibc (2.31), and newer Glibc (2.35) (`mallinfo()` vs` mallinfo2()`), on Linux x64.

Older versions of this patch were routinely tested at SAP for almost half a year.

- [1] https://mail.openjdk.org/pipermail/hotspot-dev/2021-August/054323.html
- [2] https://bugs.openjdk.org/browse/JDK-8269345

---------
### Progress

### Issue
 * [JDK-8293114](https://bugs.openjdk.org/browse/JDK-8293114): JVM should trim the native heap (**Enhancement** - P4)

### Reviewers
 * [Robbin Ehn](https://openjdk.org/census#rehn) (@robehn - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14781/head:pull/14781` \
`$ git checkout pull/14781`

`$ git checkout pull/14781` \
`$ git pull https://git.openjdk.org/jdk.git pull/14781/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14781`

`$ git pr show -t 14781`

</details>

<a href="https://git.openjdk.org/jdk/pull/14781.diff">https://git.openjdk.org/jdk/pull/14781.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14781#issuecomment-1623399540)


jdksjolen:
>And app's malloc load can fluctuate wildly, with temporary spikes and long idle periods.

Are you talking about allocations into native memory that a Java application does on its own accord and not as a consequence of the JVM doing its own allocs? For compiling, for example.

tstuefe:
> > And app's malloc load can fluctuate wildly, with temporary spikes and long idle periods.
> 
> Are you talking about allocations into native memory that a Java application does on its own accord and not as a consequence of the JVM doing its own allocs? For compiling, for example.

This does not matter. The resulting malloc load is the sum of whatever we do and whatever the native code does.

tstuefe:
Ping @shipilev and @robehn

shipilev:
I don't think the comments from my yesterday's review were addressed :) There are some old comments (marked with `(Outdated)`, helpfully), but some are new. Please take a look at them?

tstuefe:
> I don't think the comments from my yesterday's review were addressed :) There are some old comments (marked with `(Outdated)`, helpfully), but some are new. Please take a look at them?

Of course, sorry. Seems I missed most of them.

About Pause in all VM ops, an alternative would be to just check in the trimmer if we are at safepoint, and if yes treat it as pause. I'll see if that's easier (I'm worried about pulling a mutex or atomic increasing the pauser variable in every VM op we run).

shipilev:
> About Pause in all VM ops, an alternative would be to just check in the trimmer if we are at safepoint, and if yes treat it as pause. I'll see if that's easier (I'm worried about pulling a mutex or atomic increasing the pauser variable in every VM op we run).

Oh yes, I like it. We can just check `SafepointSynchronize::is_at_safepoint()` and `SafepointSynchronize::is_synchronizing()`. Would be even better, because we could stop trimming when there is safepoint sync pending. Makes the whole thing much less intrusive.

src/hotspot/share/memory/arena.cpp
```diff
@@ -91,9 +91,26 @@ class ChunkPool {
...
+
+  static bool needs_cleaning() {
     for (int i = 0; i < _num_pools; i++) {
-      _pools[i].prune();
+      if (!_pools[i].empty()) {
+        return false;
+      }
+    }
+    return true;
+  }
```
jdksjolen, line 105: Something seems wrong here? Having only empty pools means that `::prune()` is a no-op.

tstuefe: Thanks for catching this.

tstuefe: Somewhat scary that we don't have a test for that.

tstuefe:
@jdksjolen @shipilev New version:
- I removed all manual pause calls from all GC code and replaced those with just not trimming when at or near a safe point. This is less invasive since I expect the typical trim interval to be much larger than the interval we do our VM operations in.
- The pauses in runtime code remain
- I restored the original arena coding, I just added the pause. Though this coding could greatly benefit from cleanups, I want to keep this patch focused and easy to downport.
- Since we no longer have close ties to GC coding, the tests don't need to run per gc, so I dumbed down the tests.

@shipilev : I kept the SuspendMark inside TrimNative, because I like it that way. Otherwise, I would name it something like TrimNativeSuspendMark, so nothing gained but another symbol at global scope.

src/hotspot/share/runtime/trimNative.cpp
```diff
@@ -0,0 +1,194 @@
...
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questioSns.
```
lkorinth, line 22: This copyright notice differs in "questioSns" (extra 'S') and lack of  "DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER. ", might create problems for Oracle scripts

lkorinth:
The description says `-XX:GCTrimNativeHeapInterval=<seconds> (defaults to 60)`, but the code says milliseconds.

src/hotspot/share/runtime/trimNative.cpp
```diff
@@ -0,0 +1,246 @@
...
+#include "runtime/os.inline.hpp"
+#include "runtime/safepoint.hpp"
+#include "runtime/trimNative.hpp"
+#include "utilities/debug.hpp"
+#include "utilities/globalDefinitions.hpp"
+#include "utilities/ticks.hpp"
+
+class NativeTrimmerThread : public NamedThread {
+
+  Monitor* const _lock;
```
lkorinth, line 42: Maybe inline this instead of having it a pointer? Even if the thread and lock is not destroyed until VM shutdown, I always feel the need to match new in constructors with delete in destructors.

src/hotspot/share/runtime/trimNative.cpp
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  bool at_or_nearing_safepoint() const {
+    return
+        SafepointSynchronize::is_at_safepoint() ||
+        SafepointSynchronize::is_synchronizing();
+  }
+  static constexpr int safepoint_poll_ms = 250;
+
+  static int64_t now() { return os::javaTimeMillis(); }
```
lkorinth, line 78: I think it would be better to not use CLOCK_REALTIME in case of clock changes by NTP etc.

test/hotspot/jtreg/runtime/os/TestTrimNative.java
```diff
@@ -0,0 +1,279 @@
...
+
+        if (args.length == 0) {
+            throw new RuntimeException("Argument error");
+        }
+
+        if (args[0].equals("RUN")) {
+
+            System.out.println("Will spike now...");
+            for (int i = 0; i < numAllocations; i++) {
+                ptrs[i] = Unsafe.getUnsafe().allocateMemory(szAllocations);
```
shipilev, line 247: Pull `Unsafe.getUnsafe()` into a local or a field?

test/hotspot/jtreg/runtime/os/TestTrimNative.java
```diff
@@ -0,0 +1,279 @@
...
+            System.out.println("Sleeping...");
+            Thread.sleep(3000);
+            System.out.println("Done.");
+
+            return;
+
+        } else if (args[0].equals("test")) {
+            runTest(Arrays.copyOfRange(args, 1, args.length));
+        } else if (args[0].equals("testOffOnNonCompliantPlatforms")) {
+            testOffOnNonCompliantPlatforms();
```
shipilev, line 270: Inline these `test*`? This would give some symmetry against `runTest`.

test/hotspot/jtreg/runtime/os/TestTrimNative.java
```diff
@@ -0,0 +1,279 @@
...
+        checkExpectedLogMessages(output, false, 0);
+    }
+
+    public static void main(String[] args) throws Exception {
+
+        if (args.length == 0) {
+            throw new RuntimeException("Argument error");
+        }
+
+        if (args[0].equals("RUN")) {
```
shipilev, line 243: The usual trick is to pull this "internal" main into a class, and then reference it. Like:

```
public Test {
 
  public void test() {
    runWith("...", "Test$Workload");
  }  

  public class Workload {
    public static void main(String... args) {} 
  }
}
```

src/hotspot/share/runtime/trimNative.hpp
```diff
@@ -0,0 +1,65 @@
...
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ *
+ */
+
+#ifndef SHARE_GC_SHARED_TRIMNATIVE_HPP
```
shipilev, line 27: This is now `SHARE_RUNTIME_TRIMNATIVE_HPP`.

shipilev (commented):
Another read. I think we would need to tighten up style a bit too: the newline spacing style is different across the change.

src/hotspot/share/runtime/trimNative.cpp
```diff
@@ -0,0 +1,246 @@
...
+      if (sc.after != SIZE_MAX) {
+        const size_t delta = sc.after < sc.before ? (sc.before - sc.after) : (sc.after - sc.before);
+        const char sign = sc.after < sc.before ? '-' : '+';
+        log_info(trim)("Trim native heap: RSS+Swap: " PROPERFMT "->" PROPERFMT " (%c" PROPERFMT "), %1.3fms",
+                           PROPERFMTARGS(sc.before), PROPERFMTARGS(sc.after), sign, PROPERFMTARGS(delta),
+                           trim_time.seconds() * 1000);
+        log_debug(trim)("Total trims: " UINT64_FORMAT ".", _num_trims_performed);
+        return true;
+      } else {
+        log_info(trim)("Trim native heap (no details)");
```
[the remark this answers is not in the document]
robehn: I would still like to know the trim_time.

robehn (approved):
Thanks, looks good to me.

dholmes-ora (commented):
I had an initial look at this. Seems okay in principle. The naming/terminology needs some updating IMO: "trimNative" doesn't convey enough information, please use "trimNativeHeap". "trim" for logging tag is also too non-descript.

As this is initially experimental I'm not overly concerned about the impact, though I do cringe at yet-another-VM-thread.

FYI I will be away until next Thursday, but no need to wait for me for further comments.

src/hotspot/share/runtime/trimNative.cpp
tstuefe: Done. Thanks for catching that.

src/hotspot/share/runtime/trimNative.cpp
tstuefe: I rather avoid that. There is a timing factor here: JVM issues ThreadNative::cleanup, which stops the thread, but the thread is not guaranteed to stop before VM ends (may be in a middle of a trim operation). I don't want to have to think about this.

Ultimately, I don't see the point of the cleanup. I prefer a fast shutdown. Which is in line with a lot of code in the hotspot (e.g. we don't clean global mutexes either).

But I'll add an assert to make sure only one NativeTrimmerThread is ever created. So we know its a singleton.

src/hotspot/share/runtime/trimNative.cpp
tstuefe: Okay, changed to os::elapsedTime().

src/hotspot/share/runtime/trimNative.cpp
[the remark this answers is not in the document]
tstuefe: Done

tstuefe:
> I had an initial look at this. Seems okay in principle. The naming/terminology needs some updating IMO: "trimNative" doesn't convey enough information, please use "trimNativeHeap". "trim" for logging tag is also too non-descript.
> 
> As this is initially experimental I'm not overly concerned about the impact, though I do cringe at yet-another-VM-thread.
> 
> FYI I will be away until next Thursday, but no need to wait for me for further comments.

Thank you @dholmes-ora. 

I dislike the introduction of another thread too, but the benefits are undeniably there, and we need  to keep the invocation of malloc_trim away from other code paths. That said, maybe a future glibc will offer us a ways to influence trim to make its runtime predictable. Which would remove the need of an own thread.

src/hotspot/share/runtime/trimNative.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+    }
+  }
+
+  // Execute the native trim, log results.
+  bool execute_trim_and_log(int64_t tnow) const {
+    assert(os::can_trim_native_heap(), "Unexpected");
+    os::size_change_t sc;
+    Ticks start = Ticks::now();
+    log_debug(trim)("Trim native heap started...");
```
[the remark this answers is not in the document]
tstuefe: I'll revert to using elapsedTime for all timing needs here.

TraceTime seems too much hassle. I want the time combined with other output in one line, and I don't want to provide a TraceTimerLogPrintFunc.

tstuefe:
Thanks @lkorinth @shipilev @robehn for the reviews.

Next version:
- renamed TrimNative namespace to NativeHeapTrimmer, the log tag to "trimnh", files, include guards etc
- Reworked the trimmer thread to:
   - uniformly use elapsedTime
   - Use atomics for the trim count
   - get rid of run_inner
   - made suspend count 16bit
   - tightened code around the trim loop
- added minimal gtest
- reworked jtreg test to use whitebox (which required a new WB method for pre-touching memory) and reshaped the test according to Alekseys suggestions.
- and lots of other smaller stuff.

Just went through the remarks again and I hope I got everything.

ashu-mehra:
> (just noticed the patch adds +666 lines, bad sign, I should add another line somewhere).

It also deletes 2 lines so that makes it 664 :wink:

tstuefe:
> > (just noticed the patch adds +666 lines, bad sign, I should add another line somewhere).
> 
> It also deletes 2 lines so that makes it 664 wink

We are good then :-)

src/hotspot/share/runtime/trimNativeHeap.cpp
```diff
@@ -0,0 +1,219 @@
...
+        if (_stop) return;
+
+        do { // handle spurious wakeups
+          if (is_suspended()) {
+            ml.wait(0); // infinite
+          } else if (next_trim_time > tnow) {
+            const int64_t wait_ms = MAX2(1.0, to_ms(next_trim_time - tnow));
+            ml.wait(wait_ms);
+          } else if (at_or_nearing_safepoint()) {
+            ml.wait(safepoint_poll_ms);
```
shipilev, line 106: OK, so here is a little problem. Suppose I want to run trims very often, like every 10ms. This loop would stall for 250ms when safepoint is detected, which throws off this guarantee. Can we instead go and sleep for `TrimNativeHeapInterval`? AFAICs, this plays nicely with heuristic guidance (short intervals -> more interference), and it would best-effort stall for twice the interval when safepoint interjects.

shipilev (commented):
This is a nice progress! Another read follow.

Bikeshedding: I think we also need to decide how we call this thing (and related symbols): "native heap trim" or "trim native heap". AFAICT, from there, the log tag name follows, the options name follow, the VM symbol names follow.

src/hotspot/share/runtime/trimNativeHeap.cpp
tstuefe: But then we have a problem for larger trim intervals. Loosing one or multiple trim attempts because a safepoint happened to happen hurts if the interval is e.g. 5 minutes.

We could either wait for `MIN2(TrimNativeHeapInterval, safepoint_poll_ms)`.

Or, at the cost of one Mutex grab per safepoint, I could do a `notify_all()` at the end of a safepoint.

tstuefe:
New Version:

Mostly all feedback for @shipilev. Other than that:
- slightly changed the way we log when trimming (one line only)
- only calculated RSS reduction if we actually log, saves an access to procfs, which may matter for small intervals
- slight test tweaks

Re-executed the tests for release, fastdebug, fastdebug x86.

src/hotspot/share/runtime/trimNativeHeap.cpp
tstuefe: It was arbitrarily chosen to have a higher chance of "slipping" in between safepoints for larger trim intervals, but not be too small to save CPU. Admittedly I thought about this less time that this sentence takes writing.

src/hotspot/share/runtime/trimNativeHeap.cpp
```diff
@@ -0,0 +1,229 @@
...
+    LogTarget(Info, trimnative) lt;
+    const bool logging_enabled = lt.is_enabled();
+
+    // We only collect size change information if we are logging; save the access to procfs otherwise.
+    if (os::trim_native_heap(logging_enabled ? &sc : nullptr)) {
+      _num_trims_performed++;
+      if (logging_enabled) {
+        double t2 = now();
+        if (sc.after != SIZE_MAX) {
+          const size_t delta = sc.after < sc.before ? (sc.before - sc.after) : (sc.after - sc.before);
```
ashu-mehra, line 139: @tstuefe under what situations can `sc.after` be more than `sc.before` after trimming? Is it to handle the case where memory allocations happened in-between the malloc_trim() and the calls to get process memory?

tstuefe: Yes. The numbers we print out are RSS; there is no way to get the exact figure for how much memory has been released by the glibc (not even itself knows). RSS is influenced by many factors, and its a rough outline only.

src/hotspot/share/runtime/trimNativeHeap.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+        do { // handle spurious wakeups
+          if (is_suspended()) {
+            ml.wait(0); // infinite
+          } else if (next_trim_time > tnow) {
+            const int64_t wait_ms = MAX2(1.0, to_ms(next_trim_time - tnow));
+            ml.wait(wait_ms);
+          } else if (at_or_nearing_safepoint()) {
+            const int64_t wait_ms = MIN2((int64_t)TrimNativeHeapInterval, safepoint_poll_ms);
+            ml.wait(safepoint_poll_ms);
```
shipilev, line 110: `MIN2<int64_t>(..., ...)` might work better without a cast?
Also, let's actually use `wait_ms` here :)

src/hotspot/share/runtime/trimNativeHeap.cpp
```diff
@@ -0,0 +1,229 @@
...
+          }
+
+          if (_stop) return;
+
+          tnow = now();
+
+        } while (at_or_nearing_safepoint() || is_suspended() || next_trim_time > tnow);
+      } // Lock scope
+
+      // 2 - Trim outside of lock protection.
```
shipilev, line 120: There is no `1 -` to match this `2 -` to.

src/hotspot/share/runtime/trimNativeHeap.cpp
```diff
@@ -0,0 +1,229 @@
...
+          } else if (at_or_nearing_safepoint()) {
+            const int64_t wait_ms = MIN2((int64_t)TrimNativeHeapInterval, safepoint_poll_ms);
+            ml.wait(safepoint_poll_ms);
+          }
+
+          if (_stop) return;
+
+          tnow = now();
+
+        } while (at_or_nearing_safepoint() || is_suspended() || next_trim_time > tnow);
```
shipilev, line 117: Maybe invert this to pre-condition while?

shipilev (commented):
Hopefully a final read. I think there are minor things left, see comments. Alternatively, apply this patch over your current PR, which contains fixes for my comments, and then some polishing: [trimnative-shipilev-1.patch](https://github.com/openjdk/jdk/files/12025768/trimnative-shipilev-1.patch)

shipilev:
...also, Windows builds are failing.

shipilev:
One more thing, as I play with it: the GC logging does not have a comma before timestamp, see:

```
[1.210s][info][trimnative] Trim native heap (1): RSS+Swap: 1192M->1191M (-1552K), 0.353ms
[1.528s][info][gc        ] GC(0) Pause Young (Normal) (G1 Evacuation Pause) 91M->78M(1024M) 73.040ms
```

src/hotspot/share/runtime/trimNativeHeap.cpp
```diff
@@ -0,0 +1,229 @@
...
+#include "utilities/globalDefinitions.hpp"
+
+class NativeTrimmerThread : public NamedThread {
+
+  Monitor* const _lock;
+  bool _stop;
+  uint16_t _suspend_count;
+
+  // Statistics
+  unsigned _num_trims_performed;
```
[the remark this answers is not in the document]
tstuefe: No, since `_num_trims_performed` is the number of trims performed during the lifetime of the JVM. It should probably bumped to 64-bit, now that we have millisecond intervals.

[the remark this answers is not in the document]
tstuefe: Ah, I see you meant uint64_t, at least your patch says so.

tstuefe:
Thanks @shipilev!

We are closing in. 

Applied your patch, fixed Windows.

tstuefe:
Oh wait, I forgot your "RSS+Swap" trim message suggestion. One sec...

shipilev:
Yes, the only thing left is to bikeshed the logging statements a little.

tstuefe:
> Yes, the only thing left is to bikeshed the logging statements a little.

Okay done.

shipilev:
Want to replace "Native heap trimmer" with "Periodic native heap trimmer" too? Would be clear that we are suspending only the periodic one. The DCmd command would still be accepted and acted upon. Thinking about it, maybe we should do a follow-up PR and just forward that request to this thread? If so, we don't need to rename it to "Periodic".

shipilev:
Realized that in production, we would like to see why trimmer might be late.
I think this would look even better: [trimnative-shipilev-2.patch](https://github.com/openjdk/jdk/files/12043977/trimnative-shipilev-2.patch)

test/hotspot/gtest/runtime/test_trim_native.cpp
```diff
@@ -0,0 +1,48 @@
...
+    return;
+  }
+
+  // Try recursive pausing
+  {
+    NativeHeapTrimmer::SuspendMark sm1("Test1");
+    {
+      NativeHeapTrimmer::SuspendMark sm2("Test2");
+      {
+        NativeHeapTrimmer::SuspendMark sm3("Test3");
```
dholmes-ora, line 44: What is this actually testing? Everything could be a no-op.

dholmes-ora (commented):
How have you determined all of the places that need `NativeHeapTrimmer::SuspendMark`?

Couple of minor comments below but otherwise this seems good. I'll take it for a spin through our CI.

Thanks

tstuefe:
> Realized that in production, we would like to see why trimmer might be late. I think this would look even better: [trimnative-shipilev-2.patch](https://github.com/openjdk/jdk/files/12043977/trimnative-shipilev-2.patch)

I thought about this too, but you don't really want to know if it was suspended for every wait interval, but for every trim interval. In other words, you want to know how many trims had been moved up because a safepoint had been happening, and how many trims had been skipped due to pause. Getting these infos is harder than just increasing a counter. Is it worth the added complexity?

tstuefe:
> Want to replace "Native heap trimmer" with "Periodic native heap trimmer" too? Would be clear that we are suspending only the periodic one. The DCmd command would still be accepted and acted upon. Thinking about it, maybe we should do a follow-up PR and just forward that request to this thread? If so, we don't need to rename it to "Periodic".

I had initially considered it and then avoided since it would add too much unnecessary complexity. E.g., you then would have to start the native trimmer thread even if periodic trimming is off in the off-chance that someone may want to issue a trim from outside. Or, you'd need to start the trimmer thread delayed and on-demand if someone issues a dcmd; then you need to deal with the fact that the trimmer thread may or may not be there. 

Another issue then would be whether trimming would expedite the current trim, aka restart the interval timer, or be additive to the periodic trim. 

My first solution did all that, and it was a lot more complex. I like the simplicity of this patch. Customers issuing trims concurrently via dcmd will hopefully be a very rare occurrence - all those folks that used to issue the dcmd scripted every second hopefully will just switch to the periodic trimming.

Idk. Its all solvable, its just code I guess. Maybe for a later RFE? Integrating the dcmd with periodic trimming may have one pro, that is preventing customers from shooting themselves in the foot who issue the dcmd via script and then forgot they did that :-P

tstuefe:
Hi David,

> How have you determined all of the places that need `NativeHeapTrimmer::SuspendMark`?
> 

We now avoid trimming if the VM is in or nearing a SafePoint, which takes care of the majority of phases where we don't want to trim. The rest of the suspensions was just based on my gut feeling: mallocs are usually spread out more, but some releases happen in bulk, and those are easily identifiable and have the SuspendMark.

When thinking this up, I considered a lot of alternatives, e.g., hooking into os::malloc/free and setting a dead mans timer each time to give the process a "cool down" after invocations - all in the assumption that mallocs are clustered. But that was too expensive - we did a lot of work to make the normal malloc paths with NMT=off very cheap - and also ineffective since we don't see malloc use from outside hotspot (JDK or third-party), and those are the main targets. 

To really catch every malloc, I even hooked into glibc malloc hooks, but that was horribly ugly and they removed the malloc hooks in recent glibc versions.

Bottom line, I think the current solution is a good compromise of complexity vs usefulness.

> Couple of minor comments below but otherwise this seems good. I'll take it for a spin through our CI.
> 

Thank you!

src/hotspot/share/runtime/trimNativeHeap.cpp
```diff
@@ -0,0 +1,234 @@
...
+    return _stop;
+  }
+
+  bool at_or_nearing_safepoint() const {
+    return SafepointSynchronize::is_at_safepoint() ||
+           SafepointSynchronize::is_synchronizing();
+  }
+
+  // in seconds
+  static double now() { return os::elapsedTime(); }
```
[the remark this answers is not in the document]
tstuefe: I rather keep it since it keeps the caller sites a bit better readable.

test/hotspot/gtest/runtime/test_trim_native.cpp
tstuefe: Whether it blows up or not with internal asserts that test suspend count overflow.

But you are right; this raises another issue. One is that this is never tested since we need a gtest variant running with native trimming enabled. Another is that we need a printout of the trim state, at least for VMError::report, and that could be used here too to check the expected suspend count.

test/hotspot/jtreg/runtime/os/TestTrimNative.java
```diff
@@ -0,0 +1,308 @@
...
+            } break;
+
+            case "testOffOnNonCompliantPlatforms": {
+                OutputAnalyzer output = runTestWithOptions(
+                        new String[] { "-XX:+UnlockExperimentalVMOptions", "-XX:TrimNativeHeapInterval=1" },
+                        new String[] { "-version" }
+                );
+                checkExpectedLogMessages(output, false, 0);
+                parseOutputAndLookForNegativeTrim(output, 0, 0);
+                output.shouldContain("Native trim not supported on this platform");
```
dholmes-ora, line 285: This needs updating - the current log output is:
```
Native heap trim is not supported on this platform
```

dholmes-ora (changes requested):
Test needs a fix for non-Linux.

shipilev:
> Idk. Its all solvable, its just code I guess. Maybe for a later RFE? Integrating the dcmd with periodic trimming may have one pro, that is preventing customers from shooting themselves in the foot who issue the dcmd via script and then forgot they did that :-P

Yes, I am not saying we should do it in this PR. Just thinking ahead on interaction with manual trim.

shipilev:
> > Realized that in production, we would like to see why trimmer might be late. I think this would look even better: [trimnative-shipilev-2.patch](https://github.com/openjdk/jdk/files/12043977/trimnative-shipilev-2.patch)
> 
> I thought about this too, but you don't really want to know if it was suspended for every wait interval, but for every trim interval. In other words, you want to know how many trims had been moved up because a safepoint had been happening, and how many trims had been skipped due to pause. 

Well, yes. Isn't that what my patch did?

tstuefe:
> > > Realized that in production, we would like to see why trimmer might be late. I think this would look even better: [trimnative-shipilev-2.patch](https://github.com/openjdk/jdk/files/12043977/trimnative-shipilev-2.patch)
> > 
> > 
> > I thought about this too, but you don't really want to know if it was suspended for every wait interval, but for every trim interval. In other words, you want to know how many trims had been moved up because a safepoint had been happening, and how many trims had been skipped due to pause.
> 
> Well, yes. Isn't that what my patch did?

I thought you wanted to collect overarching stats about how many trims got delayed. I see now you wanted to do something much simpler. Okay.

tstuefe:
Hi @dholmes-ora, @shipilev,

New version:

- added the log Aleksey suggested. Tested manually with +SafepointALot, works fine. Briefly considered making a patch out of it but did not since such a test is not reliable enough unless one runs a longer time, which I wanted to avoid.
- Discussion with David made me realize a status print function would be nice. Added said function. It is now called for hs_err file generation as well as in dcmd VM.info.
- I then used that function to beef up the gtest. I also now call the gtest (only the relevant os.trim... subsection) as a separate jtreg-controlled test.
- fixed the unsupported-platforms-case test

src/hotspot/share/runtime/trimNativeHeap.cpp
```diff
@@ -0,0 +1,271 @@
...
+    assert(NativeHeapTrimmer::enabled(), "Only call if enabled");
+    uint16_t n = 0;
+    {
+      MonitorLocker ml(_lock, Mutex::_no_safepoint_check_flag);
+      n = dec_suspend_count();
+      if (n == 0) {
+        ml.notify_all(); // pause end
+      }
+    }
+    log_debug(trimnative)("Trim resumed after %s (%u suspend requests)", reason, n);
```
shipilev, line 204: Can you do it like I did in https://github.com/openjdk/jdk/files/12043977/trimnative-shipilev-2.patch ?

We don't say "resumed" if it really want not resumed due to non-zero suspend count.

src/hotspot/share/runtime/trimNativeHeap.cpp
```diff
@@ -0,0 +1,271 @@
...
+            ml.wait(wait_ms);
+          }
+
+          if (_stop) return;
+
+          tnow = now();
+        }
+      }
+
+      log_trace(trimnative)("Times %u suspended, %u timed, %u safepoint",
```
shipilev, line 133: `Times: `, I think.

shipilev (approved):
I am okay with this, with only minor nits left.

tstuefe:
Thanks a lot @shipilev!

Fixed the last nits, and fixed an issue where the "this platform does not support..." text was not displayed with warning level and hence not visible without Xlog. Plus, test for that.

shipilev (approved):
Still okay with it.

tstuefe:
@dholmes-ora Are you okay with this final version? Did your CI report any problems?

dholmes-ora (approved):
Nothing further from me.

Thanks

tstuefe:
I'll wait for any last remarks until tomorrow morning (CET). Barring any objections, I'll push.

robehn (approved):

tstuefe:
Many thanks @robehn, @shipilev and @dholmes-ora !

Feels good to have this finally in.

/integrate

dholmes-ora:
@tstuefe unfortunately the test is failing intermittently in our CI. I will file a bug and assign it to you. Likely we will need a quick ProblemListing though. Thanks

PR #14782, opened 2023-07-06
8311514: Incorrect regex in TestMetaSpaceLog.java

quadhier:
The regex in jtreg test `TestMetaSpaceLog.java` is intended to capture the metaspace size before and after the reclamation. But it captures the wrong value. which might cause this test to always pass. The results of the capture can be displayed using a visualization tool.

Before this patch:
屏幕快照 2023-07-06 17 18 15

After this patch:
屏幕快照 2023-07-06 17 18 25

---------
### Progress

### Issue
 * [JDK-8311514](https://bugs.openjdk.org/browse/JDK-8311514): Incorrect regex in TestMetaSpaceLog.java (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14782/head:pull/14782` \
`$ git checkout pull/14782`

`$ git checkout pull/14782` \
`$ git pull https://git.openjdk.org/jdk.git pull/14782/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14782`

`$ git pr show -t 14782`

</details>

<a href="https://git.openjdk.org/jdk/pull/14782.diff">https://git.openjdk.org/jdk/pull/14782.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14782#issuecomment-1623331969)

### DIFF
--- a/test/hotspot/jtreg/gc/logging/TestMetaSpaceLog.java
+++ b/test/hotspot/jtreg/gc/logging/TestMetaSpaceLog.java
@@ -58,7 +58,10 @@ public class TestMetaSpaceLog {
     // Do this once here.
     // Scan for Metaspace update notices as part of the GC log, e.g. in this form:
     // [gc,metaspace   ] GC(0) Metaspace: 11895K(14208K)->11895K(14208K) NonClass: 10552K(12544K)->10552K(12544K) Class: 1343K(1664K)->1343K(1664K)
-    metaSpaceRegexp = Pattern.compile(".*Metaspace: ([0-9]+)K\\([0-9]+K\\)->([0-9]+)K\\([0-9]+K\\).*");
+    // This regex has to be up-to-date with the format used in hotspot to print metaspace change.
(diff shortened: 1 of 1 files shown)


tstuefe:
Good catch. But is this guaranteed to always be in K?

tstuefe:
> > Good catch. But is this guaranteed to always be in K?
> 
> Hi Thomas! That's a good question and I believe the answer is yes.
> 

Okay, thanks for checking. The test will fail should the JVM ever change that format, right?

tstuefe (changes requested):
Can you please use `\d+` for numbers?
Also, maybe define `([0-9]+)K` as substring once and assemble the regex from that?

tstuefe (approved):
Good. Test errors on x86 unrelated.

tstuefe:
@quadhier Technically this would need two reviewers, since its a hotspot-related test. But this is trivial enough to make this a Trivial change, which just needs one reviewer.

/sponsor

tstuefe:
> Okay! Thanks for your explanations, Thomas. I'll pay attention to that.

No problem, and thanks for the fix.

PR #14785, opened 2023-07-06
8310190: C2 SuperWord: AlignVector is broken, generates misaligned packs

eme64:
I want to push this in JDK23.
After this fix here, I'm doing [this refactoring](https://github.com/openjdk/jdk/pull/16620).

To calm your nerves: most of the changes are in auto-generated tests, and tests in general.

**Context**

`-XX:+AlignVector` ensures that SuperWord only creates LoadVector and StoreVector that can be memory aligned. This is achieved by iterating in the pre-loop until we reach the alignment boundary, then we can start the main loop properly aligned. However, this is not possible in all cases, sometimes some memory accesses cannot be guaranteed to be aligned, and we need to reject vectorization (at least partially, for some of the packs).

Alignment is split into two tasks:
 - Alignment Correctness Checks: only relevant if `-XX:+AlignVector`. Need to reject vectorization if alignment is not possible. We must check if the address of the vector load/store is aligned with (divisible by) `ObjectAlignmentInBytes`.
 - Alignment by adjusting pre-loop limit: alignment is desirable even if `-XX:-AlignVector`. We would like to align the vectors with their vector width.

**Problem**

I have recently found a bug with our AlignVector [JDK-8310190](https://bugs.openjdk.org/browse/JDK-8310190).
In that bug, we perform a misaligned memory vector access, which results in a `SIGBUS` on an ARM32 machine.
Thanks @fg1417 for confirming this!
Hence, we need to fix the alignment correctness checks.

While working on this task, I also found some bugs in the "alignment by adjusting pre-loop limit": there were cases where it did not align the vectors correctly.

**Problem Details**

Reproducer:

```
    static void test(short[] a, short[] b, short mask) {
        for (int i = 0; i < RANGE; i+=8) {
            // Problematic for AlignVector
            b[i+0] = (short)(a[i+0] & mask); // best_memref, align 0

            b[i+3] = (short)(a[i+3] & mask); // pack at offset 6 bytes
            b[i+4] = (short)(a[i+4] & mask);
            b[i+5] = (short)(a[i+5] & mask);
            b[i+6] = (short)(a[i+6] & mask);
        }
    }
```

During `SuperWord::find_adjacent_refs` we used to check if the references are expected to be aligned. For that, we look at each "group" of references (eg all `LoadS`) and take the reference with the lowest offset. For that chosen reference, we check if it is alignable. If yes, we accept all references of that group, if no we reject all.

This is problematic as shown in this example. We have references at index offset `0, 3, 4, 5, 6`, and byte offset `0, 6, 8, 10, 12`. While the reference at position `0` is alignable, the pack that is eventually generated has offsets `6, 8, 10, 12`. Offset `6` means the pack cannot be aligned modulo `ObjectAlignmentInBytes = 8` nor the vector width `vw = 4*2 = 8`.

The problem is thus fundamental: we cannot just check the reference with the lowest offset, because this does in general not imply the alignability of the generated packs.

**Solution**

The new approach is to allow `SuperWord::find_adjacent_refs` to generate pair packs without regard for alignment. Instead, we should do the alignability check in the "filter stage". At that point, all packs are created, and no new ones will be added or modified - at most they would be removed. Thus, I now implemented `SuperWord::filter_packs_for_alignment`. It does all the checks in one place, rather than the multiple locations that were previously used for alignability checks. This is also a very useful refactoring, as it detangles the pack generation from the alignability check. SuperWord is more modular after this fix.

**Introducing -XX:+VerifyAlignVector**

Machines that actually require AlignVector are rare. Hence, I want to add a runtime verification that checks alignment on other machines / platforms. This will give me the confidence to fix the bug.

`-XX:+VerifyAlignVector` works when `-XX:+AlignVector` is enabled (for now just implemented on `x64`). It adds an extra `VerifyAlignmentNode` which takes the address of the LoadVector and StoreVector, performs the alignment check at runtime, and then passes the address on if the alignment is ok. If the address is misaligned, then we call `stop("verify_alignment found a misaligned vector memory access")`. We only want to do this check for vectors that are generated by SuperWord, hence we add a new flag to LoadVector and StoreVector (`must_verify_alignment`).

**Criterion for Alignment**

Previously, we have generally been operating with the understanding that vectors need to be vector width `vw` aligned. However, it turns out that we cannot get arrays aligned more than `ObjectAlignmentInBytes = 8` in the worst case. As far as I know, on the machines that have hard alignment requirements, the alignment needs to either be 4 bytes or 8 bytes. So that is sufficient. Thus, from now on we should operate under the assumption that vectors need to be aligned with the alignment width `aw = min(ObjectAlignmentInBytes, vw)`.

This alignment width `aw` is the basis both for the verification with `-XX:+VerifyAlignVector` and the re-implementation of the alignment checks in `SuperWord::filter_packs_for_alignment`.

**Solution Details**

`SuperWord::find_adjacent_refs` is now vastly simplified. Previously, it had to ensure that all references are alignable with the `best_align_to_mem_ref`. Now, we can just pick the first reference. In case of `-XX:-AlignVector` that gives us a simple best effort alignment - this has the same effect as previously `best_align_to_mem_ref` did. Since we have now no alignment checks in `SuperWord::find_adjacent_refs`, we can remove all related memops/pack removal and finding a new `best_align_to_mem_ref`.

The new logic is in `AlignmentSolver` which finds an `AlignmentSolution` for a memops pack. To find such a solution, we analyze the address of a reference in the following form:

https://github.com/openjdk/jdk/blob/4acae448a21c2c4d803c1b790511eb2be247d818/src/hotspot/share/opto/vectorization.cpp#L723-L778

I have tried ensure that we vectorize the same cases as before with the previous logic.
For example `scale` and `stride` cannot be zero (memref has no dependency on `iv`) and has to be power of 2.
Also: If variable `init` or `invar` are present, then there are further restrictions, for example all memop packs must have the same `invar`.

We try to determine if alignment is possible with the following modulo-equations:
https://github.com/openjdk/jdk/blob/4acae448a21c2c4d803c1b790511eb2be247d818/src/hotspot/share/opto/vectorization.cpp#L793-L1094

`SuperWord::filter_packs_for_alignment` iterates over all memops packs, finds an `AlignmentSolution` for each with the modulo-equations from above. It removes all packs that do not have a compatible solution, and filters the solution down to one that works for all remaining memop packs.

The solutions can be described as follows:
https://github.com/openjdk/jdk/blob/4acae448a21c2c4d803c1b790511eb2be247d818/src/hotspot/share/opto/vectorization.hpp#L266-L295

Especially interesting is checking if two constrained solutions are compatible:
https://github.com/openjdk/jdk/blob/4acae448a21c2c4d803c1b790511eb2be247d818/src/hotspot/share/opto/vectorization.hpp#L395-L472

**Alignment by adjusting Pre-Loop Limit**

This is done by `SuperWord::adjust_pre_loop_limit_to_align_main_loop_vectors` (previously called `align_initial_loop_index`).

There was at least one bug with the old logic. For example, it assumed that `scale == elt_size`. In most cases this holds, the scale (how many bytes we jump from iteration to iteration) is the same as the element size: most of the time we just have array accesses like `a[i]`. But that is not always the case:

https://github.com/openjdk/jdk/blob/3b8150750e050f08557116858670488285ef496a/test/hotspot/jtreg/compiler/loopopts/superword/TestAlignVector.java#L1432-L1443

Hence, I derived the equations anew, and implemented the new logic. I also aimed at improving the comments, such that they contain the whole derivation/proof.

https://github.com/openjdk/jdk/blob/4acae448a21c2c4d803c1b790511eb2be247d818/src/hotspot/share/opto/superword.cpp#L3468-L3609

**Other Details**

I made `VectorizeDebugOption` a debug print only flag now. Before this fix, it also had the same effect as `VectorizeOption` (which ensures that only nodes from the same original pre-unrolling node are packed, preventing hand-unrolled code to be vectorized but enabling some edge cases to be vectorized that would not otherwise vectorize).

I added `is_trace_align_vector` with bit `128`, since `64` was recently used for `is_trace_loop_reverse`, removed with [JDK-8309204](https://bugs.openjdk.org/browse/JDK-8309204).

I plan to refactor `VectorizeDebugOption` soon, as it now has a few subflags / bits that are not used. I may also refactor how `TraceSuperWord` works in general. Filed [JDK-8317572](https://bugs.openjdk.org/browse/JDK-8317572).

**Added AlignVector to IR Framework whitelist**

Since this is a hardware setting, we should whitelist it so that the IR tests can be verified to be correct on machines that do not have AlignVector set by default. I wanted to do this for a while anyway [JDK-8309662](https://bugs.openjdk.org/browse/JDK-8309662).

**New Tests**

A few tests had to be adapted for the more fine-grained logic of `AlignVector` (that it now allows alignment modulo `aw = min(ObjectAlignmentInBytes, vw)`, and not just only `vw`).

A few tests had to adapt the IR rules for `+-AlignVector`. Though this already would have been necessary, as some rare platforms already have by default `-XX:-AlignVector`. Adding `AlignVector` to the IR framework whitelist just allows us to catch those issues on other machines now.

I added some concrete regression / test cases in `TestAlignVector.java`. But I also added a "fuzzer" style test with `TestAlignVectorFuzzer.java`: it basically creates loops with "fake" compile time constants that can be randomly chosen. This way, we can fuzz loops with all sorts of constant and variable `init, limit, stride, scale, offset` and even hand-unrolling factors. This has helped me to find some bugs in my own implementation.

I also discovered some other bugs:
 - [JDK-8313717](https://bugs.openjdk.org/browse/JDK-8313717)
 - [JDK-8316594](https://bugs.openjdk.org/browse/JDK-8316594)

**Testing**

Tier1-6 and stress testing. With `-XX:+VerifyAlignVector -XX:+AlignVector` and without.
Ran `TestAlignVectorFuzzer.java` more than 500 times.

Running performance test. Once with, once without `-XX:+AlignVector`. No significant change detected.

---------
### Progress

### Issue
 * [JDK-8310190](https://bugs.openjdk.org/browse/JDK-8310190): C2 SuperWord: AlignVector is broken, generates misaligned packs (**Bug** - P4)

### Reviewers
 * [Vladimir Kozlov](https://openjdk.org/census#kvn) (@vnkozlov - **Reviewer**) ⚠️ Review applies to [aef48ab4](https://git.openjdk.org/jdk/pull/14785/files/aef48ab490e974be488ae71903106d28571fc97b)
 * [Christian Hagedorn](https://openjdk.org/census#chagedorn) (@chhagedorn - **Reviewer**) ⚠️ Review applies to [aef48ab4](https://git.openjdk.org/jdk/pull/14785/files/aef48ab490e974be488ae71903106d28571fc97b)

### Contributors
 * Christian Hagedorn `<chagedorn@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14785/head:pull/14785` \
`$ git checkout pull/14785`

`$ git checkout pull/14785` \
`$ git pull https://git.openjdk.org/jdk.git pull/14785/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14785`

`$ git pr show -t 14785`

</details>

<a href="https://git.openjdk.org/jdk/pull/14785.diff">https://git.openjdk.org/jdk/pull/14785.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14785#issuecomment-1750128055)


eme64:
@tstuefe @reinrich Would you mind testing this for PPC?

reinrich:
> @tstuefe @reinrich Would you mind testing this for PPC?
> 
> And if you have a chance, the run this test repeatedly: `test/hotspot/jtreg/compiler/loopopts/superword/TestAlignVectorFuzzer.java`

I've added this PR to our nightly testing.

src/hotspot/share/opto/compile.cpp
```diff
@@ -3666,6 +3666,33 @@ void Compile::final_graph_reshaping_main_switch(Node* n, Final_Reshape_Counts& f
...
+      if (must_verify_alignment) {
+        jlong memory_size = n->is_LoadVector() ? n->as_LoadVector()->memory_size() :
+                                                 n->as_StoreVector()->memory_size();
+        // The memory access should be aligned to the vector length in bytes.
+        // However, the underlying array is possibly less well aligned, but at least
+        // to ObjectAlignmentInBytes. Hence, even if multiple arrays are accessed in
+        // a loop we can expect at least the following alignment:
+        jlong alignment = MIN2(memory_size, (jlong)ObjectAlignmentInBytes);
+        assert(2 <= alignment && alignment <= 64, "alignment must be in range");
+        assert(is_power_of_2(alignment), "alignment must be power of 2");
```
vnkozlov, line 3684: Do we have somewhere check that vector size is always power of 2 or multiply of object alignment (for case when it is bigger than object alignment)?

src/hotspot/share/opto/superword.cpp
```diff
@@ -2790,6 +2919,17 @@ bool SuperWord::output() {
         return false; // bailout
       }
 
+#ifdef ASSERT
+      // Mark Load/Store Vector for alignment verification
+      if (AlignVector && VerifyAlignVector && cl->is_main_loop()) {
```
vnkozlov, line 2924: `AlignVector &&` no need if you do check when VerifyAlignVector is set.

src/hotspot/share/opto/superword.cpp
```diff
@@ -2790,6 +2919,17 @@ bool SuperWord::output() {
...
+        if (vn->Opcode() == Op_LoadVector) {
+          vn->as_LoadVector()->set_must_verify_alignment();
+        } else if (vn->Opcode() == Op_StoreVector) {
```
vnkozlov, line 2927: Do you mark all Load/StoreVector nodes for verification? Then why do you even need to set such field?
I you verify only few you need comment explaining which one you are verifying.

zifeihan:
Hi, JVM option -XX:+UseRVV was turned on in the QEMU riscv64 environment, and the Tier1-3 test results were OK.

reinrich:
> @tstuefe @reinrich Would you mind testing this for PPC?
> 
> And if you have a chance, the run this test repeatedly: `test/hotspot/jtreg/compiler/loopopts/superword/TestAlignVectorFuzzer.java`

Testing on PPC is done. No issues found.

eme64:
Thanks @reinrich @fg1417 @zifeihan for the testing!
Thanks @vnkozlov for the initial review. I will address them later.

Sadly, I just got a performance regression on `SPECjmh-JVM2008-LU.small-G1` and `SPECjmh-JVM2008-LU.small-ParGC`. It turns out that variable `init` values have to be accepted with `AlignVector`, at least in some cases. I reduced it down to this:

```
./java -XX:CompileCommand=compileonly,Test::test0 -XX:+TraceNewVectors -XX:+TraceSuperWord -XX:+TraceLoopOpts -XX:+AlignVector -XX:+Verbose -XX:CompileCommand=VectorizeDebug,Test::test*,128 Test.java
```

```
public class Test {
    static int RANGE = 1024*64;
​
    public static void main(String[] strArr) {
        double a[] = new double[RANGE];
        for (int i = 0; i < RANGE; i++) {
            test0(a, i);
        }
    }
​
    static void test0(double[] a, int start) {
        for (int i = start; i < RANGE; i++) {
            a[i]++;
        }
    }
}
```

Such an inner loop is part of the LU-decomposition for matrices. This lead to `30-50%` performance loss with my patch, as the loop is not vectorized anymore.

I'll try to find a more general solution that allows for variable `init` and maybe even `invar`.

eme64:
I'm now trying to take a new approach, that allows `invar` and `init` values more generally.
For this, I will try to detect the constant factor of `invar` as far as possible, such that `invar = Cinvar * Vinvar` where `Cinvar` is the largest possible extractable constant, and `Vinvar` is some variable expression.

Then, we can continue as follows:
```
adr = base + offset + invar + scale * iv
      = base + offset + Cinvar * Vinvar + scale * init + scale * pre_iter * pre_stride + scale * j * main_stride
      = base + C0 + C1 * v + C2 * pre_iter + C3 * j
```
Where
```
C2 = scale * pre_stride
C3 = scale * main_stride
```
If we have both a variable `init` and `invar`:
```
C0 = offset
C1 = min(Cinvar, scale)
```
If we have a constant `init`, and `invar`:
```
C0 = offset + scale * init
C1 = Cinvar
```
If we have variable `init` but no `invar`:
```
C0 = offset
C1 = scale
```
If we have constant `init` but no `invar`:
```
C0 = offset + scale * init
C1 = 0
```

src/hotspot/share/opto/classes.hpp
```diff
@@ -449,6 +449,7 @@ macro(StoreVectorScatter)
 macro(StoreVectorScatterMasked)
 macro(LoadVectorMasked)
 macro(StoreVectorMasked)
+macro(VerifyAlignment)
```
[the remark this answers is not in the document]
eme64: donw

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8971,6 +8971,25 @@ instruct vmasked_store_evex(memory mem, vec src, kReg mask) %{
 %}
 
 #ifdef _LP64
+instruct verify_alignment(rRegP r, rRegL mask, rRegP tmp) %{
```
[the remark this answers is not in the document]
eme64: done

src/hotspot/share/opto/compile.cpp
```diff
@@ -3666,6 +3666,33 @@ void Compile::final_graph_reshaping_main_switch(Node* n, Final_Reshape_Counts& f
 
   case Op_LoadVector:
   case Op_StoreVector:
+#ifdef ASSERT
+#ifdef AMD64
```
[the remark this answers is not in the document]
eme64: done

src/hotspot/share/opto/compile.cpp
```diff
@@ -3666,6 +3666,33 @@ void Compile::final_graph_reshaping_main_switch(Node* n, Final_Reshape_Counts& f
...
+    // Add VerifyAlignment node between adr and load / store.
+    if (AlignVector && VerifyAlignVector) {
```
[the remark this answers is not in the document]
eme64: `CompilerConfig::check_args_consistency()` seems to be too late. The reason is that `AlignVector` can still be changed by `VM_Version::get_processor_features` because of `UseUnalignedLoadStores`.
@vnkozlov do you have any other suggestion where to do this - ie when all the flags are final?

src/hotspot/share/opto/compile.cpp
```diff
@@ -3666,6 +3666,33 @@ void Compile::final_graph_reshaping_main_switch(Node* n, Final_Reshape_Counts& f
 
   case Op_LoadVector:
   case Op_StoreVector:
```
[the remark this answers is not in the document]
vnkozlov: ok

vnkozlov (commented):
Looks better.

src/hotspot/share/opto/compile.cpp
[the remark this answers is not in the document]
eme64: @vnkozlov ok, this seems to work. I now disable `VerifyAlignVector` with a warning.
`Java HotSpot(TM) 64-Bit Server VM warning: VerifyAlignVector disabled because AlignVector not enabled.`

src/hotspot/cpu/x86/x86.ad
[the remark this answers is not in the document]
eme64: @vnkozlov it seems that if I wrap it into `#ifdef ASSERT`, then it is no longer available for `Matcher::has_match_rule(Op_VerifyAlignment)`, both for slowdebug and debug. That's quite confusing. Is it somehow not correctly parsed by whatever processes the AD files?

src/hotspot/share/opto/compile.cpp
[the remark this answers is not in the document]
eme64: Done

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8971,6 +8971,25 @@ instruct vmasked_store_evex(memory mem, vec src, kReg mask) %{
...
+  effect(TEMP tmp);
+  format %{ "verify_alignment $r $mask \t! verify alignment" %}
+  ins_encode %{
+    Register Rr = $r$$Register;
+    Register Rmask = $mask$$Register;
+    Register Rtmp = $tmp$$Register;
+    Label Lskip;
+    __ movq(Rtmp, Rr);    // copy r to tmp
+    __ andq(Rtmp, Rmask); // mask off tmp
+    __ cmpq(Rtmp, 0x0);   // check if zero
```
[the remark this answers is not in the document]
eme64: Thanks for the idea, done :)

src/hotspot/share/opto/compile.cpp
eme64: Yes, at least we do those checks in SuperWord. We only ever create power of 2 vector sizes. And `ObjectAlignmentInBytes` is verified to also always be power of 2.
These asserts here are just there to catch any issues if this assumption is ever broken in the future.

src/hotspot/share/opto/superword.cpp
eme64: Done

src/hotspot/share/opto/superword.cpp
eme64: I will add a comment. We only mark the nodes created by SuperWord, because AlignVector does only apply to SuperWord, and not for example to VectorAPI nodes.

src/hotspot/cpu/x86/x86.ad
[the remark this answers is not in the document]
TobiHartmann: Parsing of preprocessor statements in adlc is quite limited. Maybe that's the issue:
https://github.com/openjdk/jdk/blob/master/src/hotspot/share/adlc/adlparse.cpp#L5152-L5169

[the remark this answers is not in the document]
eme64: @vnkozlov @TobiHartmann As we just found out: adlc is never passed the `ASSERT` flag in its input.
It seems we would have to set it in `make/hotspot/gensrc/GensrcAdlc.gmk` just like `ADLCFLAGS += -DLINUX=1 -D_GNU_SOURCE=1`.

eme64:
@fg1417 @reinrich @zifeihan

I am getting more confident with the updated version. I reran testing, and the first performance testing results look promising.

Would you mind re-running testing on your platforms?
(I ran on x64 and aarch64 without SVE)

@fg1417 I don't know aarch64 very well. Would you want to create a matching-rule for aarch64 for `VerifyVectorAlignment`, similar to what I did in `x86.ad`? It is not necessary, but could be nice.

zifeihan:
> @fg1417 @reinrich @zifeihan
> 
> I am getting more confident with the updated version. I reran testing, and the first performance testing results look promising (i.e. no regression).
> 
> Would you mind re-running testing on your platforms? (I ran on x64 and aarch64 without SVE)
> 
> @fg1417 I don't know aarch64 very well. Would you want to create a matching-rule for aarch64 for `VerifyVectorAlignment`, similar to what I did in `x86.ad`? It is not necessary, but could be nice.

Hi, I have performed tier1-3 test on linux-riscv64 using QEMU with -XX:+UseRVV. Still good.

fg1417:
> Would you mind re-running testing on your platforms? (I ran on x64 and aarch64 without SVE)

Hi @eme64 , I've submitted testing on aarch64 (sve) and arm32 separately. Let's wait for results.

> @fg1417 I don't know aarch64 very well. Would you want to create a matching-rule for aarch64 for `VerifyVectorAlignment`, similar to what I did in `x86.ad`? It is not necessary, but could be nice.

Of course, I will create the matching rule for aarch64. BTW, what kind of testing have you done for `VerifyVectorAlignment` to see if it works as expected? Then I can verify it as well on aarch64.

eme64:
@fg1417 thanks for the help!
I ran tier1-6 with `-XX:+AlignVector -XX:+IgnoreUnrecognizedVMOptions -XX:+VerifyAlignVector`.
Though `VerifyAlignVector` only has an effect if there is a matching rule in the `ad` files.
And I ran quite a few repetitions of `compiler/loopopts/superword/TestAlignVectorFuzzer.java`.

reinrich:
I'm testing on PPC. I've put `compiler/loopopts/superword/TestAlignVectorFuzzer.java` into a loop this morning. Other test results will be available tomorrow.

fg1417:
> @fg1417 thanks for the help! I ran tier1-6 with `-XX:+AlignVector -XX:+IgnoreUnrecognizedVMOptions -XX:+VerifyAlignVector`. Though `VerifyAlignVector` only has an effect if there is a matching rule in the `ad` files. And I ran quite a few repetitions of `compiler/loopopts/superword/TestAlignVectorFuzzer.java`.

Hi @eme64, we have [ `tst`](https://developer.arm.com/documentation/ddi0602/2023-09/Base-Instructions/TST--immediate---Test-bits--immediate---an-alias-of-ANDS--immediate--?lang=en) on aarch64.
```
instruct verify_vector_alignment(iRegP r, immL_positive_bitmaskI mask, rFlagsReg cr) %{
  match(Set r (VerifyVectorAlignment r mask));
  effect(KILL cr);
  format %{ "verify_vector_alignment $r $mask \t! verify alignment" %}
  ins_encode %{
    Label Lskip;
    // check if masked bits of r are zero
    __ tst($r$$Register, $mask$$constant);
    __ br(Assembler::EQ, Lskip);
    __ stop("verify_vector_alignment found a misaligned vector memory access");
    __ bind(Lskip);
  %}
  ins_pipe( pipe_slow );
%}
```

I tested tier1-tier3 with `-XX:+AlignVector -XX:+IgnoreUnrecognizedVMOptions -XX:+VerifyAlignVector` on aarch64(neon and 128-bit sve) platforms, and several repetitions of `compiler/loopopts/superword/TestAlignVectorFuzzer.java`. No new failures found. Hope it works for you. Thanks.

eme64:
@fg1417 @zifeihan @reinrich  thanks for the testing!

@fg1417 thanks for the snippet! Where do I add it? I don't know how the `ad` and `m4` work. Is there some script that converts them?

reinrich:
Testing on PPC didn't show failures.

fg1417:
> @fg1417 thanks for the snippet! Where do I add it? I don't know how the `ad` and `m4` work. Is there some script that converts them?

Hi @eme64 , you can add it [here](https://github.com/openjdk/jdk/blob/d9a89c59daa40fdc8da620940d5c518a9f18bc7b/src/hotspot/cpu/aarch64/aarch64.ad#L16394). Generally, we add **vector** operations to `src/hotspot/cpu/aarch64/aarch64_vector_ad.m4`, and redirect the output of the command: `m4 src/hotspot/cpu/aarch64/aarch64_vector_ad.m4` to `src/hotspot/cpu/aarch64/aarch64_vector.ad`. And other instructions are often added to `src/hotspot/cpu/aarch64/aarch64.ad`. Thanks!

eme64:
@bulasevich @snazarkin Can you please run this on arm32 and tell me if I broke anything? @fg1417 already kindly ran it, and found some failures. I don't have access to such a machine. This fix here is also for arm32 (my test case SIGBUs-es on arm32 with master). @fg1417 thought it might well be a backend issue. It would be good to confirm that, or else I have to figure out what I'm doing wrong in this patch.

It seems all other platforms are passing.

bulasevich:
Hi, on ARM32 jdk-tier1 and hotspot jtreg tests shows no regressing with this change.

eme64:
@bulasevich thanks for running tier1! I fear that tier1 is not sufficient to test the auto-vectorizer / SuperWord. Would you mind rerunning tier1-3? And explicitly run the test `compiler/loopopts/superword/TestAlignVectorFuzzer.java` many times. @fg1417 hit some issues with that test.

test/hotspot/jtreg/compiler/loopopts/superword/TestMulAddS2I.java
```diff
@@ -75,15 +74,16 @@ public static void compare(int[] out) {
...
         counts = {IRNode.MUL_ADD_S2I, "> 0", IRNode.MUL_ADD_VS2VI, "> 0"})
-    @IR(applyIfCPUFeature = {"sse2", "true"}, applyIf = {"UseUnalignedLoadStores", "false"},
-        failOn = {IRNode.MUL_ADD_VS2VI}, // Can only pack LoadS if UseUnalignedLoadStores is true (default if sse4.2)
-        counts = {IRNode.MUL_ADD_S2I, "> 0"})
-    @IR(applyIfCPUFeature = {"asimd", "true"}, applyIf = {"MaxVectorSize", "16"}, // AD file requires vector_length = 16
-            counts = {IRNode.MUL_ADD_S2I, "> 0", IRNode.MUL_ADD_VS2VI, "> 0"})
-    @IR(applyIfCPUFeature = {"avx512_vnni", "true"}, applyIf = {"UseUnalignedLoadStores", "true"},
-            counts = {IRNode.MUL_ADD_S2I, "> 0", IRNode.MUL_ADD_VS2VI_VNNI, "> 0"})
+    @IR(applyIfCPUFeature = {"asimd", "true"},
+        applyIfPlatform = {"64-bit", "true"},
```
fg1417, line 81: Maybe we can drop the vm-bit check. The condition `(applyIfCPUFeature = {"asimd", "true"}` should hold only on `aarch64` platform.

src/hotspot/share/opto/superword.cpp
```diff
@@ -3547,161 +3818,231 @@ LoadNode::ControlDependency SuperWord::control_dependency(Node_List* p) {
...
+  //
+  // In most cases, scale is the element size (elt_size), for example:
+  //
+  //   for (i = 0; i < a.length; i++) { a[i] = ...; }
+  //
+  // It is thus reasonable to assume that both abs(scale) and abs(stride) are
+  // strictly positive powers of 2. Further, they can be assumed to be non-zero,
+  // otherwise the address does not depend on iv, and the alignment cannot be
+  // affected by adjusting the pre-loop limit.
+  // Further, if abs(scale) >= aw, then N has no effect on alignment, and we are not
```
fg1417, line 3874: Add one blank line after line 3873？

src/hotspot/share/opto/superword.cpp
```diff
@@ -3547,161 +3818,231 @@ LoadNode::ControlDependency SuperWord::control_dependency(Node_List* p) {
...
-
-  int vw = vector_width_in_bytes(align_to_ref);
+  //   pm_e = -+ E / abs(scale)
+  //        = pm_E / abs(scale)
+  //
+  //   pm_E = -+ (base + offset + invar)
+  //        = -+ offset -+ base -+ invar
+
+  // We chose an aw that is the maximal possible vector width for the type of
+  // align_to_ref.
```
fg1417, line 3916: I have a question here: Could we always get benefit from aligning address to maximal possible vector width for small-size types, as vector width becomes large? E.g., for `byte` type on `512-bit` platform, will the pre-loop limit become very large and cost much more to execute the whole loop?

src/hotspot/share/opto/superword.cpp
```diff
@@ -1802,20 +1590,494 @@ void SuperWord::combine_packs() {
...
+    return AlignmentSolution("EQ(2*) not satisfied (cannot align across main-loop iterations)");
+  }
+
+  // In what follows, me must ensure that the C_pre term can align the C_const, C_init and C_invar terms,
+  // by adjusting the pre-loop limit (pre_iter). We must check
+  //
+  //   C_init  % abs(C_pre) = 0                                          (3a*)
+  //   C_invar % abs(C_pre) = 0                                          (3b*)
+  //
+  // to ensure that the variable term for init and invar can be aligned with the C_pre term.
```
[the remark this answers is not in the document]
eme64: Your comment prompted me to rewrite the proof a bit all the way down. I think things are now a bit more explicit. Let me know what you think!

src/hotspot/share/opto/superword.cpp
eme64: I agree, this is a concern. But this was already like that before my fix:
`int vw = vector_width_in_bytes(p.mem());`

If we want to relax this, I suggest we do that in an a future RFE.
Some thoughts:
- I did not want to change this now, since it may affect performance, and this is a bug fix here.
- We may relax the `aw` to be the maximal maximal vector size used in the vectorization. Or we can make it even smaller, i.e. `MIN2(max_vw, ObjectAlignmentInBytes);', which would then be analogue to `AlignmentSolution SuperWord::pack_alignment_solution`.
- We may even want to completely remove pre-loop adjustment if alignment is neither a strict requirement, and actually on average a performance cost that outweighs the performance gains of alignment. It is for example questionable to align one of the mem_refs if there are multiple, and hence we cannot guarantee alignment of all anyway.

eme64:
@fg1417 I rewrote the proof a bit, after some of your suggestions. I wanted to make the decomposition `pre_iter = pre_iter_C_const + pre_iter_C_invar + pre_iter_C_init` more explicit and bring it up earlier.

src/hotspot/share/opto/superword.cpp
```diff
@@ -1802,20 +1590,541 @@ void SuperWord::combine_packs() {
...
+  // to show that the C_const term can be aligned with pre_iter_C_const.
+  //
+  // We can assume that abs(C_pre) is a power of 2.
+  //
+  // If abs(C_pre) >= aw, then:
+  //
+  //   for any pre_iter >= 0: C_pre * pre_iter = 0 (mod aw)
+  //   for any pre_iter_C_const >= 0: C_pre * pre_iter_C_const = 0 (mod aw)
+  //
+  // which implies that C_iter (and pre_iter_C_const) have no effect on the alignment of
```
chhagedorn, line 1894: What is `C_iter`?

src/hotspot/share/opto/superword.cpp
```diff
@@ -1802,20 +1590,541 @@ void SuperWord::combine_packs() {
...
+    tty->print_cr(" (alignment across iterations)");
+  }
+#endif
+
+  if (C_main_mod_aw != 0) {
+    return AlignmentSolution("EQ(2*) not satisfied (cannot align across main-loop iterations)");
+  }
+
+  // In what follows, we need to show that the C_const, init and invar terms can be aligned by
+  // adjusting the pre-loop limit (pre-iter). We decompose pre_iter:
```
chhagedorn, line 1820: (pre-iter) -> (pre_iter)?

chhagedorn (commented):
Impressive work! I'm still working my way through the proofs but here are some first comments.

src/hotspot/share/opto/superword.cpp
```diff
@@ -1802,20 +1590,494 @@ void SuperWord::combine_packs() {
...
+  // Trivial constants
+  int C_const_init  = 0;
+  int C_init  = 0;
+  int C_invar = 0;
+
+  if (init_node->is_ConI()) {
+    C_const_init = init_node->as_ConI()->get_int();
+  } else {
+    C_init = scale;
+  }
```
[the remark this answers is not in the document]
eme64: I should also add some more comments here

src/hotspot/share/opto/superword.hpp
```diff
@@ -225,6 +225,158 @@ class OrderedPair {
...
+//      and the same scale.
+//
+// A solution can be:
+//   1. Invalid with a failure reason.
+//   2. Trivial (any pre-loop limit guarantees alignment).
+//   3. Constrained (r, q, mem_ref, aw, scale, invar)
+//        Where scale is 0 if no scale dependency,
+//        and invar is nullptr if no invar dependency.
+class AlignmentSolution {
+private:
```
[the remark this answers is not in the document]
eme64: I would prefer to keep it. I think it is nicer to read, I make the intention explicit. Let me know if you disagree.

src/hotspot/share/opto/superword.cpp
```diff
@@ -1802,20 +1590,494 @@ void SuperWord::combine_packs() {
...
+  // All vector loads and stores need to be memory aligned. The alignment width (aw) in
+  // principle is the vector width (vw). But when vw > ObjectAlignmentInBytes this is
+  // too strict, since any memory object is only guaranteed to be ObjectAlignmentInBytes
+  // aligned. For example, the relative offset between two arrays is only guaranteed to
+  // be divisible by ObjectAlignmentInBytes.
+  uint pack_size    = p->size();
+  MemNode* mem_ref  = p->at(0)->as_Mem();
+  int element_size  = mem_ref->memory_size();
+  int vw            = pack_size * element_size;         // vector_width
+  int aw            = MIN2(vw, ObjectAlignmentInBytes); // alignment_width
```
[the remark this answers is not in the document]
eme64: I replaced `vw` with `vector_width`, there were relatively few occurances. But if I replace `aw` with `alighment_width` then everything becomes much more "wordy", all lines become much longer. I would like to keep `aw`, I think it is much more readable then the long form. What do you think?

src/hotspot/share/opto/superword.cpp
```diff
@@ -1802,20 +1590,494 @@ void SuperWord::combine_packs() {
...
+                  pre_stride, main_stride);
+
+    // adr = base + offset + invar + scale * iv
+    tty->print("  adr = base");
+    print_icon_or_idx(base);
+    tty->print(" + offset(%d) + invar", offset);
+    print_icon_or_idx(invar);
+    tty->print_cr(" + scale(%d) * iv", scale);
+  }
+#endif
```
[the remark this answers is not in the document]
eme64: @chhagedorn What if I make it a `AlignmentSolver` that then returns an `AlignmentSolution`?

src/hotspot/share/opto/superword.cpp
```diff
@@ -1682,244 +1682,280 @@ AlignmentSolution SuperWord::pack_alignment_solution(Node_List* p) {
...
+  // In the following, we restate the simple form of the address expression, by first
+  // expanding the iv varialbe. In a second step, we reshape the expression again, and
+  // state it as a linear expression, consisting of 6 terms.
+  //
+  //          Simple form           Expansion of iv variable                  Reshaped with constants   Comments for terms
+  //          -----------           ------------------------                  -----------------------   ------------------
+  //   adr =  base               =  base                                   =  base                      (base mod aw = 0)
+  //        + offset              + offset                                  + C_const                   (sum of constant terms)
+  //        + invar               + invar_factor * var_invar                + C_invar * var_invar       (term for variable init)
+  //                          /   + scale * init                            + C_init  * var_init        (term for invariant)
```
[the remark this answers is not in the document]
eme64: I really don't know anything better than `var_invar`. The idea is that there is both a constant `C` and variable `var` term for both the `init` and `invar` terms. I even explain that below, that there is a such a factorization into constant and variable.

src/hotspot/share/opto/superword.cpp
eme64: typo. `pre_iter`

src/hotspot/share/opto/superword.cpp
```diff
@@ -3547,161 +3782,231 @@ LoadNode::ControlDependency SuperWord::control_dependency(Node_List* p) {
...
-  //     e == offset [ +/- invar ]
+  // We want to align the address of align_to_ref to some alignment width (aw, power of 2):
+  //
+  //   adr = base + offset + invar + scale * iv
+  //
+  // The limit of the pre-loop needs to be adjusted.
+  //
+  //   lim0:     current pre-loop limit
+  //   lim:      new pre-loop limit
+  //   N:        difference between lim and lim0
```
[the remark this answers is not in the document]
eme64: done

eme64:
@chhagedorn @fg1417 thanks for the first passes of reviews!

I think I want to follow Christian's idea, and refactor `SuperWord::pack_alignment_solution` into a separate class.
I will make it a bit more general if I can, so that future vectorizers can use it as well.
I'll call it `AlignmentSolver`, which returns a `AlignmentSolution`.
Wrapping it like this allows me to split the code more into sub-functions, without having to pass massive amounts of parameters.
Hopefully that will make the code more readable.

src/hotspot/share/opto/superword.cpp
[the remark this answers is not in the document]
chhagedorn: That's even better! I like that idea.

src/hotspot/share/opto/superword.cpp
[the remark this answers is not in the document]
chhagedorn: I agree with you - I also cannot think of something better. Let's keep it like that.

src/hotspot/share/opto/superword.cpp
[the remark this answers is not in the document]
chhagedorn: Great, thanks! :-)

src/hotspot/share/opto/vectorization.cpp
```diff
@@ -685,5 +702,426 @@ void VPointer::Tracer::offset_plus_k_11(Node* n) {
...
+  //     dependency on scale. We could also add a dependency for init, but since
+  //     it is the same for all mem_refs in the loop this is unnecessary. If init
+  //     is constant, then we could add a dependency that there is no variable init.
+  //     But since init is the same for all mem_refs, this is unecessary.
+  //   - If an invariant is present (i.e. C_invar = abs(invar_factor)), then we know
+  //     from (5b*), that Y = abs(invar_factor) / (scale * pre_stride). The solution
+  //     depends on Y * var_invar = abs(invar_factor) * var_invar / (scale * pre_stride),
+  //     hence we have to add a dependency for invar, and scale (pre_stride is the
+  //     same for all mem_refs in the loop). If there is no invariant, then we add
+  //     a dependency that there is no invariant.
```
chhagedorn, line 940: I found it a little difficult to understand this part. I had to write it down to verify that we indeed are not missing a dependency. Maybe we can be more explicit here. How about something like that?

```
 // Hence, the solution for pre_iter depends on:
  //   - Always: pre_r and pre_q
  //   - If an init is present, we have a dependency on it. But since init is fixed and given by the loop for all mem_refs
  //     of all packs, we can drop it.
  //   - If init is a constant, we do not have another dependency.
  //   - If init is a non-constant, we have a dependency on scale since C_init = scale. We need to keep since it is not 
  //     given by the loop and could vary from pack to pack. We do not have to add another dependency since (5a*):
  //
  //       C_init % abs(C_pre) = 0   <=>   C_init = C_pre * X
  //
  //     can only be satisfied if X is constant:
  //
  //       C_init  = C_pre * X
  //       X = C_init / C_pre
  //       X = scale / (scale * pre_stride)
  //       X = 1 / pre_stride
  //
  //   - If an invariant is present, we have a dependency on it which we need to keep since it is not given by the loop
  //     and could vary from pack to pack. We additionally have to add another dependency on scale since (5b*):
  //
  //       C_invar % abs(C_pre) = 0  <=>   C_invar = C_pre * Y
  //
  //     can only be satisfied if we have the following Y:
  //
  //       C_invar = C_pre * Y
  //       Y = (C_invar / C_pre)
  //       Y = abs(invar_factor) / (scale * pre_stride)
  //
  //     A dependency for pre_stride is not required as it is fixed and given by the loop for all mem_refs of all packs
  //     (similar to init).
```

chhagedorn: Another thought: Should we assert that if init is a non-constant that `abs(stride) = 1` if there is a solution?

src/hotspot/share/opto/vectorization.hpp
```diff
@@ -259,4 +263,322 @@ class VectorElementSizeStats {
...
+//
+// Essentially, we get a periodic solution for the C_const term, which
+// is shifted by the init and the invar terms. For two solutions to be
+// compatible, they must have the same periodic part (or subset), which
+// we can easily check with p an q. Compatible solutions must also have
+// the same shifting, which we can check with the scale and invar dependency.
+//
+class AlignmentSolutionEmpty;
+class AlignmentSolutionTrivial;
+class AlignmentSolutionConstrained;
```
chhagedorn, line 293: I usually prefer to have the other way round to better distinguish the different classes in the code. But that's just personal taste. I leave it up to you to decide which one you like better.
```suggestion
class EmptyAlignmentSolution;
class TrivialAlignmentSolution;
class ConstrainedAlignmentSolution;
```

src/hotspot/share/opto/vectorization.cpp
```diff
@@ -685,5 +702,426 @@ void VPointer::Tracer::offset_plus_k_11(Node* n) {
...
+    const int eq10_val = AlignmentSolution::mod(C_const + C_pre * pre_r, _aw);
+
+    DEBUG_ONLY( trace_find_pre_r(C_const, C_pre, pre_q, pre_r, eq10_val); )
+
+    if (eq10_val == 0) {
+      assert((C_init == 0) == _init_node->is_ConI(), "init consistent");
+      assert((C_invar == 0) == (_invar == nullptr), "invar consistent");
+
+      const Node* invar_dependency = _invar;
+      const int scale_dependency  = (_invar != nullptr || !_init_node->is_ConI()) ? _scale : 0;
```
chhagedorn, line 957:
```suggestion
      const int scale_dependency   = (_invar != nullptr || !_init_node->is_ConI()) ? _scale : 0;
```

src/hotspot/share/opto/superword.cpp
```diff
@@ -3553,164 +3462,286 @@ LoadNode::ControlDependency SuperWord::control_dependency(Node_List* p) {
...
   } else {
-    lim = new AddINode(lim0, N);
+    new_limit = new AddINode(old_limit, adjust_pre_iter);
   }
-  _igvn.register_new_node_with_optimizer(lim);
-  _phase->set_ctrl(lim, pre_ctrl);
+  _igvn.register_new_node_with_optimizer(new_limit);
+  _phase->set_ctrl(new_limit, pre_ctrl);
+
+  // 6. Make sure not to exceed the original limit with the new limit
```
chhagedorn, line 3741:
```suggestion
  // 5: Make sure not to exceed the original limit with the new limit
```

chhagedorn (commented):
Thanks for addressing my other comments! I really like the new structure of having an `AlignmentSolution` interface with different alignment solution classes. I have some more comments but mostly fine tuning - it's already in a quite good shape and the comments really add a lot of benefit to understand the idea behind the code. We get there :-)

src/hotspot/share/opto/vectorization.cpp
```diff
@@ -685,5 +702,426 @@ void VPointer::Tracer::offset_plus_k_11(Node* n) {
...
+  //   = 0                                                                       (3)
+  //
+  // We strengthen the constraints by splitting the equation into 3 equations, where the C_const,
+  // init, and invar term are aligned individually:
+  //
+  //   (C_init  * var_init  + C_pre * pre_iter_C_init ) % aw = 0                 (4a)
+  //   (C_invar * var_invar + C_pre * pre_iter_C_invar) % aw = 0                 (4b)
+  //   (C_const             + C_pre * pre_iter_C_const) % aw = 0                 (4c)
+  //
+  // We can only guarantee solutions to (4a) and (4b) if:
```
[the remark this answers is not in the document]
eme64: Ok, I added some extra comments here.

src/hotspot/share/opto/vectorization.cpp
eme64: I am refactoring this, to make the proof more tight and even correct a mistake.

src/hotspot/share/opto/vectorization.hpp
```diff
@@ -259,4 +263,345 @@ class VectorElementSizeStats {
...
+  {
+    assert(_mem_ref != nullptr &&
+           (_mem_ref->is_Load() || _mem_ref->is_Store()),
+           "only load or store vectors allowed");
+  }
+
+  AlignmentSolution solve() const;
+
+private:
+#ifdef ASSERT
```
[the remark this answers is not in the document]
eme64: Actually, I want to go towards DEBUG/ASSERT only printing, and make the trace flags debug-only (they currently only have effects in debug anyway).
I will refactor this in https://github.com/openjdk/jdk/pull/16620, and for now I will only add ASSERT guarded tracing.

eme64:
@chhagedorn I refactored the solution quite a bit, and even fixed a bug with `init` (there should never be a constraint because of it). I state the solution more precise now, and the constraints follow more directly out of it.

src/hotspot/share/opto/vectorization.cpp
```diff
@@ -685,5 +702,559 @@ void VPointer::Tracer::offset_plus_k_11(Node* n) {
...
+
+  if (C_main_mod_aw != 0) {
+    return new EmptyAlignmentSolution("EQ(2) not satisfied (cannot align across main-loop iterations)");
+  }
+
+  // In what follows, we need to show that the C_const, init and invar terms can be aligned by
+  // adjusting the pre-loop iteration count (pre_iter), which is controlled by the pre-loop
+  // limit. We decompose pre_iter:
+  //
+  //   pre_iter = pre_iter_C_const + pre_iter_C_invar + pre_iter_C_init
```
eme64, line 814: integer or fractional?

src/hotspot/share/opto/vectorization.cpp
```diff
@@ -685,5 +702,606 @@ void VPointer::Tracer::offset_plus_k_11(Node* n) {
...
+  //      and variable term. If there is no invariant, then "C_invar" is zero.
+  //
+  //        invar = C_invar * var_invar                                                                          (FAC_INVAR)
+  //
+  //   4) The "C_init * var_init" is the factorization of "scale * init" into a
+  //      constant and a variable term. If "init" is constant, then "C_init" is
+  //      zero, and "C_const" accounts for "init" instead.
+  //
+  //        scale * init = C_init * var_init + scale * C_const_init                                              (FAC_INIT)
+  //        C_init       = (init is constant) ? 0    : (scale * init / var_init)
```
chhagedorn, line 769:
```suggestion
  //        C_init       = (init is constant) ? 0    : scale
```

src/hotspot/share/opto/vectorization.cpp
```diff
@@ -685,5 +702,606 @@ void VPointer::Tracer::offset_plus_k_11(Node* n) {
...
+  //                    = my2 * q - sign(C_pre) * C_invar * var_invar / abs(C_pre)
+  //                    = my2 * q - sign(C_pre) * invar               / abs(C_pre)
+  //                    = my2 * q - invar / C_pre
+  //                    = my2 * q - invar / (scale * pre_stride)                                    (11b, with invar)
+  //
+  // If there is no invariant (i.e. C_invar = 0 ==> Y = 0):
+  //
+  //   pre_iter_C_invar = my2 * q                                                                   (11b, no invar)
+  //
+  // If init is variable (i.e. C_init = scale * init / var_init):
```
chhagedorn, line 1044:
```suggestion
  // If init is variable (i.e. C_init = scale, init = var_init):
```

src/hotspot/share/opto/vectorization.cpp
```diff
@@ -685,5 +702,606 @@ void VPointer::Tracer::offset_plus_k_11(Node* n) {
...
+  //
+  //   pre_iter_C_init  = mz2 * q - sign(C_pre) * Z       * var_init
+  //                    = mz2 * q - sign(C_pre) * C_init  * var_init  / abs(C_pre)
+  //                    = mz2 * q - sign(C_pre) * scale * init        / abs(C_pre)
```
chhagedorn, line 1048:
```suggestion
  //                    = mz2 * q - sign(C_pre) * scale   * init      / abs(C_pre)
```

src/hotspot/share/opto/vectorization.cpp
```diff
@@ -685,5 +702,606 @@ void VPointer::Tracer::offset_plus_k_11(Node* n) {
...
+  //            =   mx2 * q  - C_const / (scale * pre_stride)
+  //              + my2 * q [- invar / (scale * pre_stride) ]
+  //              + mz2 * q [- init / pre_stride            ]
+  //
+  //            =   m * q                                 (periodic part)
+  //              - C_const / (scale * pre_stride)        (align constant term)
+  //             [- invar / (scale * pre_stride)   ]      (align invariant term, if present)
+  //             [- init / pre_stride              ]      (align variable init term, if present)    (12)
+  //
+  // We can still simply simplifiy this solution, with:
```
chhagedorn, line 1070:
```suggestion
  // We can further simply this solution by introducing integer 0 <= r < q:
```

src/hotspot/share/opto/vectorization.cpp
```diff
@@ -685,5 +702,606 @@ void VPointer::Tracer::offset_plus_k_11(Node* n) {
...
+  //                                        [- invar / (scale * pre_stride) ]             (if invar present)
+  //                                        [- init / pre_stride            ]             (if init variable)
+  //                                  )
+  //   ) % aw =
+  //
+  //   -> assuming invar = 0 if it is not present
+  //   -> merge the two init terms (variable or constant)
+  //   -> apply (8): q = aw / (abs(C_pre)) = aw / abs(scale * pre_stride)
+  //   -> and hence: (scale * pre_stride * q) % aw = 0
+  //   -> all terms are cancled out
```
chhagedorn, line 1133:
```suggestion
  //   -> all terms are canceled out
```

src/hotspot/share/opto/vectorization.hpp
```diff
@@ -259,4 +263,405 @@ class VectorElementSizeStats {
...
+                                      const int C_init,
+                                      const int C_pre,
+                                      const int C_main,
+                                      const int C_main_mod_aw) const;
+  void trace_constrained_solution(const int C_const,
+                                  const int C_invar,
+                                  const int C_init,
+                                  const int C_pre,
+                                  const int q,
+                                  const int r) const;
```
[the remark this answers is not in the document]
eme64: I would rather be explicit with everything.

src/hotspot/share/opto/vectorization.hpp
```diff
@@ -259,4 +263,405 @@ class VectorElementSizeStats {
...
+    assert(s1->_q <= s2->_q, "s1 is a smaller modulo space than s2");
+
+    // Is S2 subset of (or equal to) S1?
+    //
+    // for any m2, there are integers a, b, m1: m2 * q2     + r2          =
+    //                                          m2 * a * q1 + b * q1 + r1 =
+    //                                          (m2 * a + b) * q1 + r1
+    //
+    // Since q1 and q2 are both powers of 2, and q1 <= q2, we know there
+    // is an integer a: a * q1 = q1. Thus, it remains to check if there
```
chhagedorn, line 459:
```suggestion
    // is an integer a: a * q1 = q2. Thus, it remains to check if there
```

src/hotspot/share/opto/vectorization.hpp
```diff
@@ -259,4 +263,405 @@ class VectorElementSizeStats {
...
+    tty->print("m * q(%d) + r(%d)", _q, _r);
+    if (_invar != nullptr) {
+      tty->print(" - invar[%d] / (scale(%d) * pre_stride)", _invar->_idx, _scale);
+    }
+    tty->print_cr(" [- init / pre_stride], mem_ref[%d]", mem_ref()->_idx);
+  };
+};
+
+// When strict alignment is required (e.g. -XX:+AlignVector), then we must ensure
+// that all vector memory accesses can be aligned. We acheive this alignment by
```
chhagedorn, line 484:
```suggestion
// that all vector memory accesses can be aligned. We achieve this alignment by
```

src/hotspot/share/opto/vectorization.cpp
```diff
@@ -685,5 +702,606 @@ void VPointer::Tracer::offset_plus_k_11(Node* n) {
...
+  //   -> Adjustment has no effect, C_const not aligned   -> empty
+  //
+  //   abs(C_pre) <  aw  AND  C_const % abs(C_pre) == 0
+  //   -> alignment has effect
+  //   -> C_const can be aligned with C_pre
+  //   -> Not trivial, because abs(C_pre) < aw            -> constrained
+  //
+  //   abs(C_pre) <  aw  AND  C_const % abs(C_pre) != 0
+  //   -> alignment has effect
+  //   -> But C_const cannot be aligned with C_pre        -> empty
```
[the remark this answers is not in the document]
eme64: I took your part as inspiration, and wrote this part with more detail

src/hotspot/share/opto/vectorization.cpp
```diff
@@ -890,20 +895,63 @@ AlignmentSolution* AlignmentSolver::solve() const {
...
+  //
+  //        For any pre_iter_C_const: (C_pre * pre_iter_C_const) % aw = 0
+  //
+  //        (C_const + C_pre * pre_iter_C_const) % aw = 0
+  //         C_const                             % aw = 0
+  //
+  //      Hence, we can only satisfy (4a) if C_Const is aw aligned:
+  //
+  //      C_const % aw == 0:
+  //      -> (4a) has a trivial solution since we can choose any value for pre_iter_C_Const.
```
chhagedorn, line 908:
```suggestion
  //      -> (4a) has a trivial solution since we can choose any value for pre_iter_C_const.
```

src/hotspot/share/opto/vectorization.cpp
```diff
@@ -890,20 +895,63 @@ AlignmentSolution* AlignmentSolver::solve() const {
...
+  //
+  //      C_const % aw != 0:
+  //      -> (4a) has an empty solution since no pre_iter_C_Const can achieve aw alignment.
```
chhagedorn, line 911:
```suggestion
  //      -> (4a) has an empty solution since no pre_iter_C_const can achieve aw alignment.
```

src/hotspot/share/opto/vectorization.cpp
```diff
@@ -890,20 +895,63 @@ AlignmentSolution* AlignmentSolver::solve() const {
...
+  //
+  //   abs(C_pre) < aw:
+  //   -> Since both abs(C_pre) and aw are powers of two, we know:
+  //
+  //        There exists integer x > 1: aw = abs(C_pre) * x
+  //
+  //      C_const % abs(C_pre) == 0:
+  //      -> Exists integer z: C_const = C_pre * z
```
chhagedorn, line 919: For consistency to the line above:
```suggestion
  //      -> There exists integer z: C_const = C_pre * z
```

eme64:
@chhagedorn thank you very much for the detailed review!
I think now we are at a point where other reviewers can jump in ;)
@fg1417 @vnkozlov @TobiHartmann

chhagedorn:
Thanks a lot Emanuel for all the discussions and for addressing all my comments online and offline :-)

It looks very good now and it's easy to follow the logic. The proofs are great and really helpful to better understand the (rather simple in the end) code for proofing and calculating the alignment solutions. Thanks for putting the extra effort in here.

I will have another complete look at the entire PR in the new year. But I think it looks good!

chhagedorn (approved):

eme64:
Thank you very much @chhagedorn for the very extensive review process here :blush: 
/contributor add chhagedorn

src/hotspot/share/opto/superword.cpp
```diff
@@ -725,68 +666,17 @@ void SuperWord::find_adjacent_refs() {
...
       }
     }
+  } // while (memops.size() != 0)
 
-  } // while (memops.size() != 0
-  set_align_to_ref(best_align_to_mem_ref);
+  assert(_packset.is_empty() || align_to_mem_ref != nullptr,
+         "packset empty or we find the alignment reference");
 
   if (TraceSuperWord) {
```
vnkozlov, line 674: Missing #ifndef PRODUCT

src/hotspot/share/opto/superword.cpp
```diff
@@ -1802,20 +1599,123 @@ void SuperWord::combine_packs() {
...
-  for (int i = _packset.length() - 1; i >= 0; i--) {
-    Node_List* p1 = _packset.at(i);
-    if (p1 == nullptr) {
-      _packset.remove_at(i);
-    }
-  }
+  // Remove all nullptr from packset
+  compress_packset();
 
   if (TraceSuperWord) {
```
vnkozlov, line 1605: Missing #ifndef PRODUCT

src/hotspot/share/opto/compile.cpp
```diff
@@ -3698,6 +3698,31 @@ void Compile::final_graph_reshaping_main_switch(Node* n, Final_Reshape_Counts& f
...
+      bool must_verify_alignment = n->is_LoadVector() ? n->as_LoadVector()->must_verify_alignment() :
+                                                        n->as_StoreVector()->must_verify_alignment();
+      if (must_verify_alignment) {
+        jlong vector_width = n->is_LoadVector() ? n->as_LoadVector()->memory_size() :
+                                                  n->as_StoreVector()->memory_size();
+        // The memory access should be aligned to the vector width in bytes.
+        // However, the underlying array is possibly less well aligned, but at least
+        // to ObjectAlignmentInBytes. Hence, even if multiple arrays are accessed in
+        // a loop we can expect at least the following alignment:
+        jlong guaranteed_alignment = MIN2(vector_width, (jlong)ObjectAlignmentInBytes);
```
vnkozlov, line 3713: This is more relaxed check than the actual alignment required. As I understand it is because it checks only base address of array and not actually memory address to which vector instruction is accessed (which is (base,index,offset)).
It is useful but does not guarantee correct alignment of vector access instructions.

Consider using `lea` instruction on x86 to load memory address into register and check it.

vnkozlov (commented):
Few comments.

src/hotspot/share/opto/compile.cpp
vnkozlov: May be hack only `loadV` and `storeV` instructions in .ad file to use `lea` and do the check.

eme64: I don't understand this comment.
The `LoadVector` and `StoreVector` have both a `MemNode::Address` input, which I think it the memory address.
The address itself usually consists of `AddP` nodes, which do the (base, index, offset) computation. These nodes later can be folded into the load/store itself, or be computed with a `lea`.
I simply take the address value, check it for alignment and pass it on to the load/store.

Take this example:
```
public class Test {
    static int RANGE = 1024*64;

    public static void main(String[] strArr) {
        int a[] = new int[RANGE];
        test0(a);
    }

    static void test0(int[] a) {
        for (int i = 0; i < RANGE; i++) {
            a[i]++;
        }
    }
}
```

`./java -XX:CompileCommand=compileonly,Test::test* -XX:+TraceSuperWord -Xcomp -XX:+PrintIdeal -XX:+AlignVector  -XX:+VerifyAlignVector -XX:CompileCommand=print,Test::test* Test.java`

This looks like the main loop:
```
 ;; B22: #	out( B22 B23 ) <- in( B21 B22 ) Loop( B22-B22 inner post of N743) Freq: 4.49988
  0x00007f83c8bb2f68:   lea    0x10(%rbp,%rbx,4),%r10       ;*iaload {reexecute=0 rethrow=0 return_oop=0}
                                                            ; - Test::test0@12 (line 11)
  0x00007f83c8bb2f6d:   mov    %r10,%r8
  0x00007f83c8bb2f70:   test   $0x7,%r8b
  0x00007f83c8bb2f74:   je     0x00007f83c8bb2f8a
  0x00007f83c8bb2f76:   movabs $0x7f83d77e2fc8,%rdi         ;   {external_word}
  0x00007f83c8bb2f80:   and    $0xfffffffffffffff0,%rsp
  0x00007f83c8bb2f84:   callq  0x00007f83d71a0162           ;   {runtime_call MacroAssembler::debug64(char*, long, long*)}
  0x00007f83c8bb2f89:   hlt    
  0x00007f83c8bb2f8a:   test   $0x7,%r10b
  0x00007f83c8bb2f8e:   je     0x00007f83c8bb2fa4
  0x00007f83c8bb2f90:   movabs $0x7f83d77e2fc8,%rdi         ;   {external_word}
  0x00007f83c8bb2f9a:   and    $0xfffffffffffffff0,%rsp
  0x00007f83c8bb2f9e:   callq  0x00007f83d71a0162           ;   {runtime_call MacroAssembler::debug64(char*, long, long*)}
  0x00007f83c8bb2fa3:   hlt    
  0x00007f83c8bb2fa4:   vpaddd (%r10),%zmm5,%zmm0
  0x00007f83c8bb2faa:   vmovdqu32 %zmm0,(%r8)               ;*iastore {reexecute=0 rethrow=0 return_oop=0}
                                                            ; - Test::test0@15 (line 11)
  0x00007f83c8bb2fb0:   add    $0x10,%ebx                   ;*iinc {reexecute=0 rethrow=0 return_oop=0}
                                                            ; - Test::test0@16 (line 10)
  0x00007f83c8bb2fb3:   cmp    %r11d,%ebx
  0x00007f83c8bb2fb6:   jl     0x00007f83c8bb2f68           ;*if_icmpge {reexecute=0 rethrow=0 return_oop=0}
                                                            ; - Test::test0@6 (line 10)
```
What I see: `lea` computes address, stores to register `r10`. Move value to `r8`, do alignment check `test   $0x7,%r8b`, which checks for 8 byte alignment. We do the same check again with `r10b`, since we use the same address for load and store. And then we directly load/store with those register values:
```
vpaddd (%r10),%zmm5,%zmm0
vmovdqu32 %zmm0,(%r8)
```

src/hotspot/share/opto/compile.cpp
```diff
@@ -1056,7 +1056,7 @@ void Compile::Init(bool aliasing) {
   set_has_monitors(false);
 
   if (AllowVectorizeOnDemand) {
-    if (has_method() && (_directive->VectorizeOption || _directive->VectorizeDebugOption)) {
+    if (has_method() && _directive->VectorizeOption) {
```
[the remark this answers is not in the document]
eme64: This is my justification in the PR description:

> Other Details
> 
> I made VectorizeDebugOption a debug print only flag now. Before this fix, it also had the same effect as VectorizeOption (which ensures that only nodes from the same original pre-unrolling node are packed, preventing hand-unrolled code to be vectorized but enabling some edge cases to be vectorized that would not otherwise vectorize).
> 
> I added is_trace_align_vector with bit 128, since 64 was recently used for is_trace_loop_reverse, removed with [JDK-8309204](https://bugs.openjdk.org/browse/JDK-8309204).
> 
> I plan to refactor VectorizeDebugOption soon, as it now has a few subflags / bits that are not used. I may also refactor how TraceSuperWord works in general. Filed [JDK-8317572](https://bugs.openjdk.org/browse/JDK-8317572).

src/hotspot/share/opto/compile.cpp
eme64: And without `-XX:-VerifyAlignVector`

```
 ;; B20: #	out( B20 B21 ) <- in( B19 B20 ) Loop( B20-B20 inner post of N743) Freq: 4.49988
  0x00007ff22cbb2924:   vpaddd 0x10(%rbx,%r13,4),%zmm0,%zmm1
  0x00007ff22cbb292f:   vmovdqu32 %zmm1,0x10(%rbx,%r13,4)   ;*iastore {reexecute=0 rethrow=0 return_oop=0}
                                                            ; - Test::test0@15 (line 11)
  0x00007ff22cbb293a:   add    $0x10,%r13d                  ;*iinc {reexecute=0 rethrow=0 return_oop=0}
                                                            ; - Test::test0@16 (line 10)
  0x00007ff22cbb293e:   cmp    %r10d,%r13d
  0x00007ff22cbb2941:   jl     0x00007ff22cbb2924           ;*if_icmpge {reexecute=0 rethrow=0 return_oop=0}
                                                            ; - Test::test0@6 (line 10)
```

src/hotspot/share/opto/compile.cpp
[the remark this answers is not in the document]
eme64: If you really want, then I can not touch `VectorizeDebugOption` at all, i.e. not activate `is_trace_align_vector` with that flag, but instead simply use `TraceSuperWord` (that might be a little verbose though).

I already have the CSR for [JDK-8315361](https://bugs.openjdk.org/browse/JDK-8315361), so that I can remove  `VectorizeDebugOption`. This has 2 effects:
1. remove the product effect of `VectorizeDebugOption`, which is the same effect as enabling `VectorizeOption`.
2. introduce a more general auto-vectorization tracing flag that allows more fine-grained control for debug printing.

My idea here was to simply add the alignment tracing to `VectorizeDebugOption`. But currently one cannot enable that tracing without having the side-effects that also `VectorizeOption` has. Hence, I already now remove that product-side effect.

@vnkozlov what do you think?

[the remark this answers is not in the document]
vnkozlov: I missed that in your long description ;^)
I agree with your suggestion. The option was indeed strange: mixing prints with affects on code.

src/hotspot/share/opto/compile.cpp
vnkozlov: Can you show assembler code for simple load and store instructions (move data from one array to another)?
My concern is that LoadV and StoreV are defined only with `memory` input:
```
instruct loadV(vec dst, memory mem) %{
  match(Set dst (LoadVector mem));
```
I would assume it will be embedded memory only. But C2 may be smart enough to generate `lea` if it sees not AddP node.

vnkozlov: Also why your assembler example have tested alignment twice for the same address? May be because the same array's element for load and store?:
```
  0x00007f83c8bb2f6d:   mov    %r10,%r8
  0x00007f83c8bb2f70:   test   $0x7,%r8b
  0x00007f83c8bb2f74:   je     0x00007f83c8bb2f8a
...
  0x00007f83c8bb2f8a:   test   $0x7,%r10b
  0x00007f83c8bb2f8e:   je     0x00007f83c8bb2fa4
```
No need to optimize I think since it is only for debugging.

src/hotspot/share/opto/compile.cpp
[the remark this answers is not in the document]
eme64: Ok, great, I will leave it then :+1:

src/hotspot/share/opto/compile.cpp
eme64: > My concern is that LoadV and StoreV are defined only with memory input

```
// Indirect Memory Operand
operand indirect(any_RegP reg)
%{
  constraint(ALLOC_IN_RC(ptr_reg));
  match(reg);

  format %{ "[$reg]" %}
  interface(MEMORY_INTER) %{
    base($reg);
    index(0x4);
    scale(0x0);
    disp(0x0);
  %}
%}
```

```
opclass memory(indirect, indOffset8, indOffset32, indIndexOffset, indIndex,
               indIndexScale, indPosIndexScale, indIndexScaleOffset, indPosIndexOffset, indPosIndexScaleOffset,
               indCompressedOopOffset,
               indirectNarrow, indOffset8Narrow, indOffset32Narrow,
               indIndexOffsetNarrow, indIndexNarrow, indIndexScaleNarrow,
               indIndexScaleOffsetNarrow, indPosIndexOffsetNarrow, indPosIndexScaleOffsetNarrow);
```
It seems that `memory` summarizes many different patterns. One of them is the `indirect` one, which simply loads the address from a register. In our case this address was computed by a `lea`, then used in the alignment verification, and then passed on as `memory` to load / store.

vnkozlov:
Okay, so C2 is smart enough to use `lea` when needed. Good.
No more questions.

vnkozlov (approved):
Looks good.

src/hotspot/share/opto/vectorization.cpp
```diff
@@ -685,5 +702,655 @@ void VPointer::Tracer::offset_plus_k_11(Node* n) {
...
+  //          (C_pre * z + C_pre * pre_iter_C_const) % aw               = 0
+  //          ==>
+  //          (C_pre * z + C_pre * pre_iter_C_const) % (abs(C_pre) * x) = 0
+  //          ==>
+  //          (        z +         pre_iter_C_const) %               x  = 0
+  //          ==>
+  //          for any m: pre_iter_C_const = m * x - z
+  //
+  //        Hence, pre_iter_C_const has a non-trivial (because x > 1) periodic (periodicity x)
+  //        solution, i.e it has a constrained solution.
```
chhagedorn, line 934:
```suggestion
  //        solution, i.e. it has a constrained solution.
```

src/hotspot/share/opto/vectorization.cpp
```diff
@@ -685,5 +702,655 @@ void VPointer::Tracer::offset_plus_k_11(Node* n) {
...
+  //
+  //           C_const                             %  abs(C_pre)      != 0
+  //          ==>
+  //          (C_const + C_pre * pre_iter_C_const) %  abs(C_pre)      != 0
+  //          ==>
+  //          (C_const + C_pre * pre_iter_C_const) % (abs(C_pre) * x) != 0
+  //          ==>
+  //          (C_const + C_pre * pre_iter_C_const) % aw               != 0
+  //
+  //        This is in constradiction with (4a), and therefore there cannot be any solution,
```
chhagedorn, line 947:
```suggestion
  //        This is in contradiction with (4a), and therefore there cannot be any solution,
```

src/hotspot/share/opto/vectorization.cpp
```diff
@@ -685,5 +702,655 @@ void VPointer::Tracer::offset_plus_k_11(Node* n) {
...
+  //
+  //   C_const = X * abs(C_pre)   ==>   X = C_const / abs(C_pre)                 (6a)
+  //   C_invar = Y * abs(C_pre)   ==>   Y = C_invar / abs(C_pre)                 (6b)
+  //   C_init  = Z * abs(C_pre)   ==>   Z = C_init  / abs(C_pre)                 (6c)
+  //
+  // Further, we define:
+  //
+  //   sign(C_pre) = C_pre / abs(C_pre) = (C_pre > 0) ? 1 : -1,                  (7)
+  //
+  // We know that abs(C_pre) as well as aw are a powers of 2, and since (5) we can define integer q:
```
chhagedorn, line 1038:
```suggestion
  // We know that abs(C_pre) as well as aw are powers of 2, and since (5) we can define integer q:
```

chhagedorn (approved):
Some last minor comments. Otherwise, looks good!

src/hotspot/share/opto/superword.cpp
```diff
@@ -1802,18 +1601,123 @@ void SuperWord::combine_packs() {
...
   }
+#endif
+}
+
+// Find the set of alignment solutions for load/store pack.
+const AlignmentSolution* SuperWord::pack_alignment_solution(Node_List* pack) {
+  assert(pack != nullptr && (pack->at(0)->is_Load() || pack->at(0)->is_Store()), "only load/store packs");
+
+  const MemNode* mem_ref = pack->at(0)->as_Mem();
+  VPointer mem_ref_p(mem_ref, phase(), lpt(), nullptr, false);
```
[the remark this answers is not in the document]
eme64: I keep it without your suggestion.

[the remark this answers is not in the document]
eme64: The idea is that it is the pointer of the mem_ref

eme64:
Thanks @chhagedorn for all the help to get this over the line :)
Thanks @vnkozlov for the review!

/integrate

PR #14789, opened 2023-07-06
8066869: Add Closeable::closeUnchecked that is the equivalent of close but throws UncheckedIOException

bplb:
Add a default method `java.io.Closeable::closeUnchecked` which is equivalent to `Closeable::close` except that it instead throws `java.io.UncheckedIOException`.

---------
### Progress
- [ ] Change requires a CSR request matching fixVersion 22 to be approved (needs to be created)

### Issue
 * [JDK-8066869](https://bugs.openjdk.org/browse/JDK-8066869): Add Closeable::closeUnchecked that is the equivalent of close but throws UncheckedIOException (**Sub-task** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14789/head:pull/14789` \
`$ git checkout pull/14789`

`$ git checkout pull/14789` \
`$ git pull https://git.openjdk.org/jdk.git pull/14789/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14789`

`$ git pr show -t 14789`

</details>

<a href="https://git.openjdk.org/jdk/pull/14789.diff">https://git.openjdk.org/jdk/pull/14789.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14789#issuecomment-1624255925)

### DIFF
--- a/src/java.base/share/classes/java/io/Closeable.java
+++ b/src/java.base/share/classes/java/io/Closeable.java
@@ -43,7 +43,7 @@ public interface Closeable extends AutoCloseable {
      * with it. If the stream is already closed then invoking this
      * method has no effect.
      *
-     * <p> As noted in {@linkplain AutoCloseable#close()}, cases where the
+     * <p> As noted in {@link AutoCloseable#close()}, cases where the
      * close may fail require careful attention. It is strongly advised
      * to relinquish the underlying resources and to internally
      * <em>mark</em> the {@code Closeable} as closed, prior to throwing
@@ -67,11 +67,11 @@ public interface Closeable extends AutoCloseable {
      *     throw new UncheckedIOException(cause);
      * }
      * }
-     * Therefore if an {@linkplain UncheckedIOException} is thrown then its
-     * {@linkplain UncheckedIOException#getCause() getCause} method will return the
-     * {@linkplain IOException} thrown by {@link #close() close}.
+     * Therefore if an {@link UncheckedIOException} is thrown then its
+     * {@link UncheckedIOException#getCause() getCause} method will return the
+     * {@link IOException} thrown by {@link #close() close}.
      *
-     * @throws UncheckedIOException If invoking {@code close} throws
+     * @throws UncheckedIOException if invoking {@code close} throws
      * an {@code IOException}.
      *
      * @since 22
--- a/src/java.base/share/classes/java/io/Closeable.java
+++ b/src/java.base/share/classes/java/io/Closeable.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2003, 2013, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2003, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -26,11 +26,13 @@
 package java.io;
 
 import java.io.IOException;
+import java.io.UncheckedIOException;
 
 /**
  * A {@code Closeable} is a source or destination of data that can be closed.
- * The close method is invoked to release resources that the object is
- * holding (such as open files).
+ * The {@code close} method is invoked to release resources that the object is
+ * holding (such as open files).  The {@code closeUnchecked} method
+ * closes the stream and may throw only an unchecked exception.
  *
  * @since 1.5
  */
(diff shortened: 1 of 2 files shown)


bplb:
A CSR will follow if there is agreement on the concept.

src/java.base/share/classes/java/io/Closeable.java
```diff
@@ -50,4 +52,35 @@ public interface Closeable extends AutoCloseable {
      * @throws IOException if an I/O error occurs
      */
     public void close() throws IOException;
+
+    /**
+     * Closes this stream and releases any system resources associated
+     * with it. If the stream is already closed then invoking this
+     * method has no effect.
+     *
```
RogerRiggs, line 60: Can you add an example showing the recommended pattern of use?

bplb: > Can you add an example showing the recommended pattern of use?

One pattern is for example when reading from a stream and closing it is highly unlikely to generate an exception, then instead of the try block in
```
InputStream in = ...;
while(someCondition) {
    int b = in.read();
    // process b
}
try {
    in.close();
} catch (IOException ignored) {
}
```
one could simply write
```
in.closeUnchecked();
```
and avoid the boilerplate imposed by the checked exception. This sort of code can be found in many places in the tests in `<OpenJDK>/test/jdk`.

stuart-marks: OK, but note that using `closeUnchecked` in this way changes the behavior. Instead of ignoring an `IOException` it will get rethrown wrapped in an `UncheckedIOException`. This is probably reasonable for test code, where it's often acceptable to let exceptions propagate instead of handling them.

Are there other places in the JDK (in library code or in the tests) where `closeUnchecked` can and should be used? If so, maybe they should be updated as well.

bplb: > This is probably reasonable for test code, where it's often acceptable to let exceptions propagate instead of handling them.

Understood.

> Are there other places in the JDK (in library code or in the tests) where `closeUnchecked` can and should be used? If so, maybe they should be updated as well.

Unless it's a matter of further background investigation, I think actually making any such changes would be better handled as a separate issue.

stuart-marks: > Unless it's a matter of further background investigation, I think actually making any such changes would be better handled as a separate issue.

OK. It doesn't have to be part of this change, but it would good if some example use cases could be identified. All too often I come up with great ideas for APIs, and then when I look around for example use sites, I can't find any ... and maybe this is telling me the API idea isn't as great as I thought it was.

bplb: > and then when I look around for example use sites, I can't find any ... and maybe this is telling me the API idea isn't as great as I thought it was.

I hear you.

src/java.base/share/classes/java/io/Closeable.java
```diff
@@ -50,4 +52,35 @@ public interface Closeable extends AutoCloseable {
...
+     * {@snippet lang=java :
+     * try {
+     *     close();
+     * } catch (IOException cause) {
+     *     throw new UncheckedIOException(cause);
+     * }
+     * }
+     * Therefore if an {@linkplain UncheckedIOException} is thrown then its
+     * {@linkplain UncheckedIOException#getCause() getCause} method will return the
+     * {@linkplain IOException} thrown by {@link #close() close}.
```
liach, line 72: These links should all be `@link` as they should be rendered like code like the `{@code IOException}` below.

bplb: I agree. So changed in c89945b81a9d8312087593ed5e6ccba06a413ce0.

bplb:
> I'm not a big fan of hiding exceptions, why not wrapping the IOException in an UncheckedIOException ?
> 
> So closeUnchecked() means close and throw an unchecked exception if an exception occurs.

I am confused by this comment. That is what `closeUnchecked()` is specified to do.

Glavo:
Can it be a static method instead of a default method? `Closeable` is a widely used class, so adding new default methods risks conflicting with existing methods.

jddarcy:
Hmm.

If the goal was to preserve the ability to use the try-with-resources statement, then a utility method that wrapped a java.io.Closeable with a java.io.UncheckedCloseable would be needed, where UncheckedCloseable had a close method that was declared to not throw any exceptions and did the IOException wrapping.

AlanBateman:
> Add a default method `java.io.Closeable::closeUnchecked` which is equivalent to `Closeable::close` except that it instead throws `java.io.UncheckedIOException`.

I'm not sure about the usefulness of this as it means using try-finally instead of try-with-resources. To continue to use try-with-resources would need a new interface + some kind of adaptor. Also probably need to think about the code in the block. Calls to methods such as read/write will throw IOException so code will have to deal with them anyway - maybe you could a migrate a few examples using some of the common closeables to see if this is the case.

Glavo:
I don't like this PR. The problems it can solve are very limited, and the way to solve the problem is not an excellent solution worth promoting.

After the introduction of lambda in Java 8, the cumbersomeness brought by checked exceptions became more prominent. I think we need a more general construct to solve this, rather than just adding some helper methods.

bplb:
This work would be better performed in conjunction with language feature support and try-with-resources. Withdrawing for now.

PR #14791, opened 2023-07-06
8155902: DataOuputStream should clarify that it might write primitive types as multiple byte groups

bplb:
Add a disclaimer to `java.io.DataOutputStream` to the effect that it makes no guarantee as to how the underlying output stream actually writes the bytes provided to it.

---------
### Progress
- [x] Change requires CSR request [JDK-8311882](https://bugs.openjdk.org/browse/JDK-8311882) to be approved

### Issues
 * [JDK-8155902](https://bugs.openjdk.org/browse/JDK-8155902): DataOuputStream should clarify that it might write primitive types as multiple byte groups (**Bug** - P4)
 * [JDK-8311882](https://bugs.openjdk.org/browse/JDK-8311882): DataOuputStream should clarify that it might write primitive types as multiple byte groups (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14791/head:pull/14791` \
`$ git checkout pull/14791`

`$ git checkout pull/14791` \
`$ git pull https://git.openjdk.org/jdk.git pull/14791/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14791`

`$ git pr show -t 14791`

</details>

<a href="https://git.openjdk.org/jdk/pull/14791.diff">https://git.openjdk.org/jdk/pull/14791.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14791#issuecomment-1624351015)

### DIFF
--- a/src/java.base/share/classes/java/io/DataOutputStream.java
+++ b/src/java.base/share/classes/java/io/DataOutputStream.java
@@ -37,6 +37,11 @@ import jdk.internal.util.ByteArray;
  * thread then access to the data output stream should be controlled
  * by appropriate synchronization.
  *
+ * @apiNote
+ * The write methods of this class merely delegate to the methods of
+ * the underlying output stream, hence there is no guarantee as to
+ * the details of how the bytes are written.
+ *
  * @see     java.io.DataInputStream
  * @since   1.0
  */
--- a/src/java.base/share/classes/java/io/DataOutputStream.java
+++ b/src/java.base/share/classes/java/io/DataOutputStream.java
@@ -32,9 +32,8 @@ import jdk.internal.util.ByteArray;
  * types to an output stream in a portable way. An application can
  * then use a data input stream to read the data back in. A data output
  * stream wraps another output stream and delegates writing bytes to the
- * write methods of that output stream. Hence writing data consisting of
- * more than a single byte may cause several writes to the underlying
- * output stream.
+ * write methods of that output stream. Writing data consisting of more than
+ * a single byte may cause several writes to the underlying output stream.
  * <p>
  * A DataOutputStream is not safe for use by multiple concurrent
  * threads. If a DataOutputStream is to be used by more than one
--- a/src/java.base/share/classes/java/io/DataOutputStream.java
+++ b/src/java.base/share/classes/java/io/DataOutputStream.java
@@ -30,17 +30,16 @@ import jdk.internal.util.ByteArray;
 /**
  * A data output stream lets an application write primitive Java data
  * types to an output stream in a portable way. An application can
- * then use a data input stream to read the data back in.
+ * then use a data input stream to read the data back in. A data output
+ * stream wraps another output stream and delegates writing bytes to the
+ * write methods of that output stream. Hence writing data consisting of
+ * more than a single byte may cause several writes to the underlying
+ * output stream.
  * <p>
  * A DataOutputStream is not safe for use by multiple concurrent
  * threads. If a DataOutputStream is to be used by more than one
  * thread then access to the data output stream should be controlled
  * by appropriate synchronization.
- *
- * @apiNote
- * The write methods of this class delegate to the methods of
- * the underlying output stream, hence there is no guarantee as to
- * the details of how the bytes are written.
  *
  * @see     java.io.DataInputStream
  * @since   1.0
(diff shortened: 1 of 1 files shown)


bplb:
It might be equally reasonable simply to resolve the issue as "Not an Issue." A CSR will be created if that is not the case.

djelinski:
This issue is reported against JDK 8 and 9. It was fixed in 16 with 17f04fc9. I think you can close it as duplicate.

src/java.base/share/classes/java/io/DataOutputStream.java
```diff
@@ -37,6 +37,11 @@
  * thread then access to the data output stream should be controlled
  * by appropriate synchronization.
  *
+ * @apiNote
+ * The write methods of this class merely delegate to the methods of
+ * the underlying output stream, hence there is no guarantee as to
+ * the details of how the bytes are written.
+ *
```
RogerRiggs, line 44: I'd drop "merely".

I would not want to give any assurances about whether the bytes of an encoded primitive were written individually or as a group.  
For the primitive types the value is encoded and written as one or more bytes to the underlying stream.

bplb:
> This issue is reported against JDK 8 and 9. It was fixed in 16 with [17f04fc](https://github.com/openjdk/jdk/commit/17f04fc9e755a83727ac9d13a39b487f99311efd). I think you can close it as duplicate.

Thanks. I saw that change go by but did not connect the dots. Was it actually verified that things are not split up into two or more packets? It seems that `write(2)` could in fact return a value less than the number of bytes requested.

src/java.base/share/classes/java/io/DataOutputStream.java
bplb: > I would not want to give any assurances about whether the bytes of an encoded primitive were written individually or as a group.

Would it be helpful to add further verbiage?
```
For example, the sequence of bytes comprising the primitive data type may be written as more than one packet of bytes.
```

djelinski:
> Was it actually verified that things are not split up into two or more packets? It seems that `write(2)` could in fact return a value less than the number of bytes requested.

Yes, the test case included in JBS sends all 4 bytes in a single packet now.
We can't guarantee that every writeInt will always send all bytes in the same packet, but at least we won't send single-byte packets.

src/java.base/share/classes/java/io/DataOutputStream.java
RogerRiggs: "packet" is network centric terminology.
Perhaps something like:  "may be written individually or in any grouping of bytes"

bplb: Commit 43c392e8c797f5a5c29a67e87f350e8dc9c5b190 changed it to
```
The write methods of this class delegate to the methods
of the underlying output stream, hence there is no 
guarantee as to the details of how the bytes are written.
```

AlanBateman:
In passing, we should probably look into the inconsistency in the synchronization. The write methods are synchronized, the writeXXX methods and other methods are not. The class specified that it is not safe for use by concurrent threads without appropriate synchronization.

src/java.base/share/classes/java/io/DataOutputStream.java
```diff
@@ -37,6 +37,11 @@
  * thread then access to the data output stream should be controlled
  * by appropriate synchronization.
  *
+ * @apiNote
+ * The write methods of this class delegate to the methods of
+ * the underlying output stream, hence there is no guarantee as to
+ * the details of how the bytes are written.
+ *
```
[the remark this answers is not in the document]
bplb: Changed accordingly in f89aa63ac2b03e66ae3f7238e4decbb0d28794de.

src/java.base/share/classes/java/io/DataOutputStream.java
```diff
@@ -30,7 +30,11 @@
 /**
  * A data output stream lets an application write primitive Java data
  * types to an output stream in a portable way. An application can
- * then use a data input stream to read the data back in.
+ * then use a data input stream to read the data back in. A data output
+ * stream wraps another output stream and delegates writing bytes to the
+ * write methods of that output stream. Hence writing data consisting of
+ * more than a single byte may cause several writes to the underlying
+ * output stream.
```
AlanBateman, line 37: This looks much better than the apiNote. One suggestion is to drop "Hence", I don't think it is needed here.

djelinski (approved):
LGTM.

RogerRiggs (approved):

PR #14799, opened 2023-07-07
8311557: [JVMCI] deadlock with JVMTI thread suspension

tkrodriguez:
Java based JVMCI compiler threads are more like normal Java threads so they aren't `hidden_from_external_view` like the native compilers.  This can leak to deadlocks if you use JVMTI to suspend all threads since this will block the compiler queue and can block execution if background compilation is disabled.  It's reasonable to treat libgraal threads like native threads in this regard.  Making jargraal threads hidden too would interfere with using profiling and debugging tool on them so I've left that alone but it might be worth changing the JVMTI suspend and resume functions to explicitly skip compiler threads as well.

### DIFF
--- a/src/hotspot/share/compiler/compilerThread.cpp
+++ b/src/hotspot/share/compiler/compilerThread.cpp
@@ -61,3 +61,8 @@ void CompilerThread::thread_entry(JavaThread* thread, TRAPS) {
 bool CompilerThread::can_call_java() const {
   return _compiler != nullptr && _compiler->is_jvmci();
 }
+
+// Hide native compiler threads from external view.
+bool CompilerThread::is_hidden_from_external_view() const {
+  return _compiler == nullptr || !_compiler->is_jvmci() || UseJVMCINativeLibrary;
+}
--- a/src/hotspot/share/compiler/compilerThread.hpp
+++ b/src/hotspot/share/compiler/compilerThread.hpp
@@ -72,8 +72,7 @@ class CompilerThread : public JavaThread {
 
   virtual bool can_call_java() const;
 
-  // Hide native compiler threads from external view.
-  bool is_hidden_from_external_view() const      { return !can_call_java(); }
+  virtual bool is_hidden_from_external_view() const;
 
   void set_compiler(AbstractCompiler* c)         { _compiler = c; }
   AbstractCompiler* compiler() const             { return _compiler; }


TobiHartmann (approved):
That looks reasonable to me.

> Making jargraal threads hidden too would interfere with using profiling and debugging tool on them so I've left that alone but it might be worth changing the JVMTI suspend and resume functions to explicitly skip compiler threads as well.

Please file a follow-up issue for that.

tkrodriguez:
Thanks for the review.  Part of the problem is that it's not clear what the relationship between things like `can_call_java` and `hidden_from_external_view` is and whether making some of these methods safer for JVMCI threads will confuse users of JVMTI.  There will likely be some follow up issues related to this as we navigate these distinctions.

tkrodriguez:
I did some local testing and unsurprisingly this changes make it impossible to attach a debugger to these threads.  In rare cases during development it might be necessary to this since there are still some helper Java calls with native JVMCI.  So I've added a new flag LibJVMCICompilerThreadHidden to control this behaviour.  I also refactored the code a bit because I figured I was missing some JVMCI_ONLY macros in the shared code.  Could I get a rereview?

src/hotspot/share/compiler/abstractCompiler.hpp
```diff
@@ -151,6 +151,8 @@ class AbstractCompiler : public CHeapObj<mtCompiler> {
   bool is_jvmci() const                  { return _type == compiler_jvmci; }
   CompilerType type() const              { return _type; }
 
+  virtual bool is_hidden_from_external_view() const { return false; }
```
dougxc, line 154: It would be nice if this had a comment explaining what "hidden from external view" implies. But I see that `Thread::is_hidden_from_external_view` has no documentation either so I guess there's no much that can be explained here if the broader concept is somewhat undefined.

src/hotspot/share/jvmci/jvmciCompiler.hpp
```diff
@@ -103,6 +103,9 @@ class JVMCICompiler : public AbstractCompiler {
   bool is_c1   ()                                { return false; }
   bool is_c2   ()                                { return false; }
 
+  virtual bool is_hidden_from_external_view() const { return UseJVMCINativeLibrary && LibJVMCICompilerThreadHidden; }
+
```
dougxc, line 107: Remove extra blank line.

src/hotspot/share/jvmci/jvmci_globals.hpp
```diff
@@ -157,6 +157,11 @@ class fileStream;
           "error data to this file"                                         \
           "[default: ./" LIBJVMCI_ERR_FILE "] (%p replaced with pid)")      \
                                                                             \
+  product(bool, LibJVMCICompilerThreadHidden, true, DIAGNOSTIC,             \
+          "If true then native JVMCI compiler threads are hidden from "     \
+          "JVMTI and FlightRecorder.  This must be set to false if you"     \
```
dougxc, line 162: `you"` -> `you "`
`threads"` -> `threads."`

As far as I can help, this help text is never printed (is it?) but it may as well be properly formatted.

dougxc (approved):

src/hotspot/share/jvmci/jvmci_globals.hpp
tkrodriguez: Fixed.  It's not in the product but it does function a documentation for the options.

src/hotspot/share/jvmci/jvmciCompiler.hpp
tkrodriguez: Ok

src/hotspot/share/compiler/abstractCompiler.hpp
tkrodriguez: I added a comment in CompilerThread.

TobiHartmann (approved):
Still looks good to me.

dougxc (approved):

dougxc (approved):

tkrodriguez:
Thanks!

/integrate

src/hotspot/share/jvmci/jvmci_globals.hpp
```diff
@@ -157,6 +157,11 @@ class fileStream;
           "error data to this file"                                         \
           "[default: ./" LIBJVMCI_ERR_FILE "] (%p replaced with pid)")      \
                                                                             \
+  product(bool, LibJVMCICompilerThreadHidden, true, EXPERIMENTAL,           \
```
dholmes-ora, line 160: This sounds more like a diagnostic flag than experimental.

tkrodriguez: It kind of is but because of the way the way [JVMCIGlobals::enable_jvmci_product_mode ](https://github.com/tkrodriguez/jdk/blob/tkr-hidden-fix/src/hotspot/share/jvmci/jvmci_globals.cpp#L169) works it seems easier to just make it product like all the other special JVMCI flags.

PR #14800, opened 2023-07-07
8311548: AArch64: [ZGC] Many tests fail with "assert(allocates2(pc)) failed: not in CodeBuffer memory" on some CPUs

shqking:
### Problem:

We got many ZGC related JTreg test failures on ThunderX2 CPU. The failure can be easily reproduced by the following command as well.

  `java -XX:+UseZGC -XX:+ZGenerational --version`

Here shows the snippet of error message:

```
 A fatal error has been detected by the Java Runtime Environment:

 Internal Error (~/jdk_build/jdk_src/src/hotspot/share/asm/codeBuffer.hpp:200), pid=108369, tid=108373
 assert(allocates2(pc)) failed: not in CodeBuffer memory: 0x0000ffff9c921100 <= 0x0000ffff9c934c04 <= 0x0000ffff9c934c00

 JRE version: (22.0) (fastdebug build )
 Java VM: OpenJDK 64-Bit Server VM (fastdebug 22-internal-git-0916e6a60, mixed mode, sharing, compressed class ptrs, z gc, linux-aarch64)
 Problematic frame:
 V [libjvm.so+0x45413c] Instruction_aarch64::~Instruction_aarch64()+0xbc
```

### Root cause:

From the backtrace (See the Description session in the JBS [1]), we can see that it's an assembler failure and the failure occurred when the VM generated "final stubs". The root cause is that the buffer size of "final stubs" is too small for ThunderX2 CPU with ZGC on.

The reason that this failure only occurred on ThunderX2 CPU but not on CPUs like Neoverse N1/N2, is that 1) VM flag "AvoidUnalignedAccesses" is enabled by default on CPUs like ThunderX2 (See the code [2]), and 2) more instructions would be generated especially with ZGC (E.g., see the code [3]).

Hence, the failure would also occur on CPUs like Neoverse N1/N2 if we pass VM option "-XX:+AvoidUnalignedAccesses", e.g.,

`java -XX:+UseZGC -XX:+ZGenerational -XX:+AvoidUnalignedAccesses --version`

### Fix:

Increasing the buffer size for "final blobs", i.e. variable `_final_stubs_code_size`, would fix the failure.

We manually computed the code size for "final stubs" on ThunderX2 CPU and the size is roughly "105568" bytes. In this patch, we increase the buffer size from "60000" to "100000" for ZGC_ONLY().

### Test:

Tier1~3 passed on Linux/ThunderX2 and Linux/Neoverse N1.

[1] https://bugs.openjdk.org/browse/JDK-8311548
[2] https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp#L154-L177
[3] https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp#L919-L1109

### DIFF
--- a/src/hotspot/cpu/aarch64/stubRoutines_aarch64.hpp
+++ b/src/hotspot/cpu/aarch64/stubRoutines_aarch64.hpp
@@ -39,7 +39,7 @@ enum platform_dependent_constants {
   _initial_stubs_code_size      = 10000,
   _continuation_stubs_code_size =  2000,
   _compiler_stubs_code_size     = 30000 ZGC_ONLY(+10000),
-  _final_stubs_code_size        = 20000 ZGC_ONLY(+60000)
+  _final_stubs_code_size        = 20000 ZGC_ONLY(+100000)
 };
 
 class aarch64 {


xmas92 (approved):

RealFYang (approved):
LGTM. I could also reproduce this issue on HiSilicon TSV110.

Stub sizes after this change for reference:
```
$ java -XX:+UseZGC -XX:+ZGenerational -Xlog:stubs -version
[0.013s][info][stubs] StubRoutines (initial stubs)       [0x0000ffff8f920080, 0x0000ffff8f922a10] used: 5408, free: 5232
[0.212s][info][stubs] StubRoutines (continuation stubs)  [0x0000ffff8f923500, 0x0000ffff8f923f50] used: 648, free: 1992
[0.246s][info][stubs] StubRoutines (final stubs)         [0x0000ffff8f96b480, 0x0000ffff8f988bc0] used: 103664, free: 16976
[0.270s][info][stubs] StubRoutines (compiler stubs)      [0x0000ffff8fa1c580, 0x0000ffff8fa27ac0] used: 24552, free: 21848
```

kimbarrett (approved):
Looks good.

tschatzl (approved):

shqking:
Thanks for your reviews. The GHA tests are green. Let me integrate it.
/integrate

shqking:
As described in [JBS-8311548](https://bugs.openjdk.org/browse/JDK-8311548?focusedCommentId=14594147&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14594147), this bug was introduced in jdk21 and also affected jdk21.
Besides, it's a P2 bug. I'd like to backport this patch to jdk21.

/backport jdk21

PR #14803, opened 2023-07-07
8311648: Refactor the Arena/Chunk/ChunkPool interface

jdksjolen:
Hi,

This PR refactors the Arena code without introducing any changes in behavior. I believe that this leads to a much clearer API and that this will in turn simplify the maintenance of this code. `Chunk` is now much more "dumb", mainly acting as a way of interrogating the chunk of data that it is holding. The `ChunkPool` gains a more prominent role, being entirely responsible for the memory allocation strategy used. `Arena` is basically unchanged, as it orchestrates the usage of these two APIs. The `chop` and `next_chop` methods are re-defined to be static inside of `Chunk`, having an object `delete` itself  is very surprising, so I'm happy to get rid of that.

I hope that you agree with me on these changes!

Cheers,
Johan

---------
### Progress

### Issue
 * [JDK-8311648](https://bugs.openjdk.org/browse/JDK-8311648): Refactor the Arena/Chunk/ChunkPool interface (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14803/head:pull/14803` \
`$ git checkout pull/14803`

`$ git checkout pull/14803` \
`$ git pull https://git.openjdk.org/jdk.git pull/14803/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14803`

`$ git pr show -t 14803`

</details>

<a href="https://git.openjdk.org/jdk/pull/14803.diff">https://git.openjdk.org/jdk/pull/14803.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14803#issuecomment-1625339753)


src/hotspot/share/memory/arena.cpp
```diff
@@ -42,93 +42,81 @@ STATIC_ASSERT(is_aligned((int)Chunk::medium_size, ARENA_AMALLOC_ALIGNMENT));
...
-    }
-    _first = nullptr;
-  }
-
-  static void clean() {
-    for (int i = 0; i < _num_pools; i++) {
-      _pools[i].prune();
-    }
-  }
+  void prune();
```
tstuefe, line 60: Comment just repeats the name. I'd either omit it or describe what the function is doing.

src/hotspot/share/memory/arena.cpp
```diff
@@ -42,93 +42,81 @@ STATIC_ASSERT(is_aligned((int)Chunk::medium_size, ARENA_AMALLOC_ALIGNMENT));
...
-  Chunk*       _first;        // first cached Chunk; its first word points to next chunk
-  const size_t _size;         // (inner payload) size of the chunks this pool serves
-
   // Our four static pools
   static const int _num_pools = 4;
   static ChunkPool _pools[_num_pools];
 
- public:
-  ChunkPool(size_t size) : _first(nullptr), _size(size) {}
+  Chunk*       _first;        // first cached Chunk; its first word points to next chunk
```
tstuefe, line 52: I'd probably leave out the comment since its pretty obvious how this works.

tstuefe (commented):
Hi Johan,

cursory review.

Could you please give this a massage and minimize deltas a bit? Would make review and backporting easier. Seems several things moved without much reason to move. 

Thanks!

jdksjolen:
> Hi Johan,
> 
> cursory review.
> 
> Could you please give this a massage and minimize deltas a bit? Would make review and backporting easier. Seems several things moved without much reason to move.
> 
> Thanks!

Hi,

The PR is much smaller now!

tstuefe (approved):
Looks good.

src/hotspot/share/memory/arena.cpp
```diff
@@ -212,19 +181,37 @@ void Chunk::start_chunk_pool_cleaner_task() {
...
+  _first = _chunk = ChunkPool::allocate_chunk(AllocFailStrategy::EXIT_OOM, init_size);
   _hwm = _chunk->bottom();      // Save the cached hwm, max
   _max = _chunk->top();
   MemTracker::record_new_arena(flag);
   set_size_in_bytes(init_size);
 }
 
 Arena::Arena(MEMFLAGS flag) : _flags(flag), _size_in_bytes(0) {
-  _first = _chunk = new (AllocFailStrategy::EXIT_OOM, Chunk::init_size) Chunk(Chunk::init_size);
+  _first = _chunk = ChunkPool::allocate_chunk(AllocFailStrategy::EXIT_OOM, Chunk::init_size);
```
coleenp, line 214: I have one request - can you make the AllocFailStrategy the last parameter?

coleenp (approved):
Looks good but I had one request.

src/hotspot/share/memory/arena.cpp
jdksjolen: Fixed!

coleenp (approved):
Looks good, thanks!

tstuefe:
Still good. Make sure x86 works.

jdksjolen:
Hm, apparently this PR breaks something with current master. Very strange, I'll have to look over it.

jdksjolen:
Hi,

This PR introduces a bug in commit [Remove operator new/delete](https://github.com/openjdk/jdk/pull/14803/commits/3b2fc13382c75ce348453ff92ded4ad7549131e6) wherein a `ChunkPool` creates a circular linked list of chunks. No clue how that happens so far. Here's a gdb printout showing the phenomenom (breakpoint in `return_to_pool`.

```
(gdb) 
Continuing.
_first: (nil), chunk: 0x7ffff0171300, chunk->_next: (nil)
(gdb) 
Continuing.
_first: (nil), chunk: 0x7ffff0171300, chunk->_next: (nil)
(gdb) 
Continuing.
_first: 0x7ffff0171300, chunk: 0x7ffff002e680, chunk->_next: (nil)
(gdb) 
Continuing.
_first: 0x7ffff0171300, chunk: 0x7ffff002e680, chunk->_next: 0x7ffff0171300
(gdb) 
Continuing.
_first: 0x7ffff002e680, chunk: 0x7ffff0171300, chunk->_next: (nil)
(gdb) c
Continuing.
_first: 0x7ffff002e680, chunk: 0x7ffff0171300, chunk->_next: 0x7ffff002e680
(gdb) 
Continuing.
_first: 0x7ffff0171300, chunk: 0x7ffff002e680, chunk->_next: 0x7ffff0171300
(gdb) 
Continuing.
_first: 0x7ffff002e680, chunk: 0x7ffff0171300, chunk->_next: 0x7ffff002e680
(gdb) 
Continuing.
_first: 0x7ffff0171300, chunk: 0x7ffff002e680, chunk->_next: 0x7ffff0171300
(gdb) 
Continuing.
_first: 0x7ffff002e680, chunk: 0x7ffff0171300, chunk->_next: 0x7ffff002e680
```

jdksjolen:
Hi,

@tstuefe , @coleenp, I found the bug. The issue was that `Chunk::operator new` would implicitly call the constructor of `Chunk` and thus overwrite previous content of chunks taken from the pool. I forgot to call the constructor for chunks taken from the pool. I've rewritten `allocate_chunk` so that there are two allocation branches that both end with the constructor being called via placement new.

I also merged in the new `clean` method. We'll see what CI says now.

tstuefe:
Hi Johan,

> Hi,
> 
> @tstuefe , @coleenp, I found the bug. The issue was that `Chunk::operator new` would implicitly call the constructor of `Chunk` and thus overwrite previous content of chunks taken from the pool. I forgot to call the constructor for chunks taken from the pool. I've rewritten `allocate_chunk` so that there are two allocation branches that both end with the constructor being called via placement new.
> 

I don't get this :-)

What new call are you talking about? And in the current version, why should a constructor call be necessary if all we do is handing an existing - already initialized - chunk from and to the pool? Yes, it still carries _next from the pool, but that is soon to be overwritten when the Arena adds that chunk to its list (although setting _next to NULL in debug deliberately can't hurt).

src/hotspot/share/memory/arena.cpp
```diff
@@ -43,41 +44,32 @@ STATIC_ASSERT(is_aligned((int)Chunk::medium_size, ARENA_AMALLOC_ALIGNMENT));
...
-  void free(Chunk* chunk) {
+  void return_to_pool(Chunk* chunk) {
     assert(chunk->length() == _size, "wrong pool for this chunk");
     ThreadCritical tc;
     chunk->set_next(_first);
     _first = chunk;
   }
 
-  // Prune the pool
   void prune() {
```
tstuefe, line 73: Admittedly the old comment was useless, but could you add a comment explaining what happens?

Alternatively, rename the function. E.g. free_all_chunks or similar. It is not really pruning anything since it completely empties out the list (pruning suggests there is something pruned left afterwards).

tstuefe:
> Hi Johan,
> 
> > Hi,
> > @tstuefe , @coleenp, I found the bug. The issue was that `Chunk::operator new` would implicitly call the constructor of `Chunk` and thus overwrite previous content of chunks taken from the pool. I forgot to call the constructor for chunks taken from the pool. I've rewritten `allocate_chunk` so that there are two allocation branches that both end with the constructor being called via placement new.
> 
> I don't get this :-)
> 
> What new call are you talking about? And in the current version, why should a constructor call be necessary if all we do is handing an existing - already initialized - chunk from and to the pool? Yes, it still carries _next from the pool, but that is soon to be overwritten when the Arena adds that chunk to its list (although setting _next to NULL in debug deliberately can't hurt).

Ah, I think I get it. I believe the real error is that when adding a chunk to the end of the Arena linked list, old code did not NULL-terminate the list since it relied on the Chunk being freshly initialized. IMHO, a slightly more robust and standard solution would be to explicitly NULL-terminate the list on add.

src/hotspot/share/memory/arena.cpp
```diff
@@ -280,15 +273,19 @@ void* Arena::grow(size_t x, AllocFailType alloc_failmode) {
     return nullptr;
   }
 
-  Chunk *k = _chunk;            // Get filled-up chunk address
-  _chunk = new (alloc_failmode, len) Chunk(len);
+  Chunk* k = _chunk;            // Get filled-up chunk address
+  _chunk = ChunkPool::allocate_chunk(len, alloc_failmode);
```
tstuefe, line 277: Odd that we would fatal out if init chunk cannot be allocated, and here we leave the OOM behavior to the caller. Is this ever called with anything else than exit_oom?

tstuefe (commented):
Small nits remain. Any chance you can come up with some gtests for the pool? Up to you; it would mean restructuring the code and moving the pool to a header, probably.

src/hotspot/share/memory/arena.cpp
jdksjolen: Looking at the callers: No, we don't, but I think changing this API would go under a different PR.

jdksjolen:
>Ah, I think I get it. I believe the real error is that when adding a chunk to the end of the Arena linked list, old code did not NULL-terminate the list since it relied on the Chunk being freshly initialized. IMHO, a slightly more robust and standard solution would be to explicitly NULL-terminate the list on add.

I think mentioning that `allocate_chunk` returns an initialized and null-terminated Chunk of requested size is sufficient. Explicitly null-terminating each time a `Chunk` is 'allocated' would require remembering that `Chunk`s may not be fresh.

tstuefe:
> > Ah, I think I get it. I believe the real error is that when adding a chunk to the end of the Arena linked list, old code did not NULL-terminate the list since it relied on the Chunk being freshly initialized. IMHO, a slightly more robust and standard solution would be to explicitly NULL-terminate the list on add.
> 
> I think mentioning that `allocate_chunk` returns an initialized and null-terminated Chunk of requested size is sufficient. Explicitly null-terminating each time a `Chunk` is 'allocated' would require remembering that `Chunk`s may not be fresh.

Okay. Having a linked list op not terminating the end of the list goes against all my instincts, but that is fine :-)

tstuefe (approved):
Looked through the changes again, looks good.

src/hotspot/share/memory/arena.cpp
```diff
@@ -129,28 +129,31 @@ Chunk* ChunkPool::allocate_chunk(size_t length, AllocFailType alloc_failmode) {
...
+    // Either the pool was empty, or this is a non-standard length. Allocate a new Chunk from C-heap.
+    size_t bytes = ARENA_ALIGN(sizeof(Chunk)) + length;
+    void* p = os::malloc(bytes, mtChunk, CALLER_PC);
+    if (p == nullptr && alloc_failmode == AllocFailStrategy::EXIT_OOM) {
+      vm_exit_out_of_memory(bytes, OOM_MALLOC_ERROR, "Chunk::new");
+    }
+    chunk = (Chunk*)p;
   }
-  Chunk* chunk = ::new(p) Chunk(length);
+  ::new(chunk) Chunk(length);
```
coleenp, line 153: You need to do this because if you find and take a chunk out of the pool, it's not properly initialized?

coleenp: Okay, I read the code before the comments in the PR, so that answers my question.

coleenp (approved):
Looks good relying on prior review and current review of fixed allocate_chunk.

jdksjolen:
Thanks for the patience and thorough reviews!

/integrate

PR #14805, opened 2023-07-07
8262742: (fs) Add Path::resolve with varargs string

bplb:
Add to `java.nio.file.Path` methods which allow resolution of multiple descendants.

---------
### Progress
- [x] Change requires CSR request [JDK-8312123](https://bugs.openjdk.org/browse/JDK-8312123) to be approved

### Issues
 * [JDK-8262742](https://bugs.openjdk.org/browse/JDK-8262742): (fs) Add Path::resolve with varargs string (**Enhancement** - P4)
 * [JDK-8312123](https://bugs.openjdk.org/browse/JDK-8312123): (fs) Add Path::resolve with varargs string (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14805/head:pull/14805` \
`$ git checkout pull/14805`

`$ git checkout pull/14805` \
`$ git pull https://git.openjdk.org/jdk.git pull/14805/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14805`

`$ git pr show -t 14805`

</details>

<a href="https://git.openjdk.org/jdk/pull/14805.diff">https://git.openjdk.org/jdk/pull/14805.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14805#issuecomment-1625900982)

### DIFF
--- a/src/java.base/share/classes/java/nio/file/Path.java
+++ b/src/java.base/share/classes/java/nio/file/Path.java
@@ -522,9 +522,9 @@ public interface Path
      *
      * <p> If {@code more} does not specify any {@code Path}s,
      * then the result is {@link #resolve(Path) resolve(first)}. If
-     * {@code more} specifies one or more {@code Path}s, then each non-empty
-     * {@code Path}, including {@code first}, is resolved iteratively against
-     * this {@code Path}.
+     * {@code more} specifies one or more {@code Path}s, then {@code first}
+     * is resolved against this path and then the additional paths in
+     * {@code more} are iteratively resolved.
      *
      * If {@code first} and {@code more} specify only empty paths, then this
      * method returns this path. If any of the paths in {@code first} and
@@ -551,7 +551,7 @@ public interface Path
      *          the first path to resolve against this path
      *
      * @param   more
-     *          additional paths to be iteratively resolved against this path
+     *          additional paths to be iteratively resolved
      *
      * @return  the resulting path
      *
@@ -574,10 +574,11 @@ public interface Path
      * <p> If {@code more} does not
      * specify any path strings, then the result is {@link #resolve(String)
      * resolve(first)}. If {@code more} specifies one or more path strings,
-     * then each non-empty path string, including {@code first}, is iteratively
-     * converted to a {@code Path} and resolved against this {@code Path}.
-     * Path strings which convert to empty or absolute paths are handled as
-     * specified by {@link #resolve(Path,Path...)}.
+     * then {@code first} is converted to a {@code Path}, resolved against
+     * this path, and then the additional path strings in {@code more} are
+     * iteratively converted to {@code Path}s and resolved. Path strings which
+     * convert to empty or absolute paths are handled as specified by
+     * {@link #resolve(Path,Path...)}.
      *
      * <p> The result of this method is the same as would be obtained by
      * resolving the path converted from the first path string against
@@ -600,7 +601,7 @@ public interface Path
      *
      * @param   more
      *          additional path strings to be iteratively converted to
-     *          paths and resolved against this path
+     *          paths and resolved
      *
      * @return  the resulting path
      *
--- a/src/java.base/share/classes/java/nio/file/Path.java
+++ b/src/java.base/share/classes/java/nio/file/Path.java
@@ -521,7 +521,7 @@ public interface Path
      *
      * <p> The result of this method is the same as would be obtained
      * by resolving the first parameter path against this path, then
-     * resovling the second parameter path against the derived path,
+     * resolving the second parameter path against the derived path,
      * and so on until all parameters have been processed. Each single
      * resolution in the chain behaves as specified by {@linkplain
      * #resolve(Path) resolve}.
--- a/src/java.base/share/classes/java/nio/file/Path.java
+++ b/src/java.base/share/classes/java/nio/file/Path.java
@@ -517,7 +517,18 @@ public interface Path
     }
 
     /**
-     * Resolves a path, or a sequence of paths against this path.
+     * Resolves one or more {@code Path}s iteratively against this
+     * {@code Path}. If {@code more} does not specify any {@code Path}s,
+     * then the result is {@link #resolve(Path) resolve(first)}. If
+     * {@code more} specifies one or more {@code Path}s, then each non-empty
+     * {@code Path}, including {@code first}, is resolved iteratively against
+     * this {@code Path}.
+     *
+     * If {@code first} and {@code more} specify only empty paths, then this
+     * method returns this path. If any of the paths in {@code first} and
+     * {@code more} is absolute, then the result will be as if the paths
+     * after the last absolute path in the sequence were resolved iteratively
+     * against that last absolute path.
      *
      * <p> The result of this method is the same as would be obtained
      * by resolving the first parameter path against this path, and then
@@ -536,7 +547,7 @@ public interface Path
      *
      * @param first the first path to resolve against this path
      *
-     * @param more additional paths to be joined to form the path
+     * @param more additional paths to be iteratively resolved against this path
      *
      * @return the resulting path
      *
@@ -553,8 +564,14 @@ public interface Path
     }
 
     /**
-     * Resolves the path or paths converted from a path string, or a
-     * sequence of path strings, against this path.
+     * Resolves one or more {@code Path}s converted from the supplied path
+     * srings iteratively against this {@code Path}. If {@code more} does not
+     * specify any path strings, then the result is {@link #resolve(String)
+     * resolve(first)}. If {@code more} specifies one or more path strings,
+     * then each non-empty path string, including {@code first}, is iteratively
+     * converted to a {@code Path} and resolved against this {@code Path}.
+     * Path strings which convert to empty or absolute paths are handled as
+     * specified by {@link #resolve(Path,Path...)}.
      *
      * <p> The result of this method is the same as would be obtained by
      * resolving the path converted from the first path string against
@@ -571,9 +588,11 @@ public interface Path
      * }
      * }
      *
-     * @param first the first path string to resolve against this path
+     * @param first the first path string to convert to a path and
+     *        resolve against this path
      *
-     * @param more additional path strings to be joined to form the path
+     * @param more additional path strings to be iteratively converted to
+     *        paths and resolved against this path
      *
      * @return the resulting path
      *
--- a/src/java.base/share/classes/java/nio/file/Path.java
+++ b/src/java.base/share/classes/java/nio/file/Path.java
@@ -517,13 +517,12 @@ public interface Path
     }
 
     /**
-     * Resolves a path or a sequence of paths against this path.
+     * Resolves a path, or a sequence of paths against this path.
      *
      * <p> The result of this method is the same as would be obtained
-     * by resolving the first parameter path against this path, then
-     * resolving the second parameter path against the derived path,
-     * and so on until all parameters have been processed. Each single
-     * resolution in the chain behaves as specified by {@linkplain
+     * by resolving the first parameter path against this path, and then
+     * repeating the same procedure for each path in the sequence. Each
+     * single resolution in the chain behaves as specified by {@link
      * #resolve(Path) resolve}.
      *
      * @implSpec
@@ -554,17 +553,14 @@ public interface Path
     }
 
     /**
-     * Converts a path string or a sequence of path strings into one or more
-     * {@code Path}s and resolves them against this {@code Path} in exactly
-     * the same manner specified by the {@link #resolve(Path,Path...) resolve}
-     * method.
+     * Resolves the path or paths converted from a path string, or a
+     * sequence of path strings, against this path.
      *
-     * <p> The result of this method is the same as would be obtained
-     * by resolving the first parameter path string against this path, then
-     * resolving the second parameter path string against the derived path,
-     * and so on until all parameters have been processed. Each single
-     * resolution in the chain behaves as specified by {@linkplain
-     * #resolve(String) resolve}.
+     * <p> The result of this method is the same as would be obtained by
+     * resolving the path converted from the first path string against
+     * this path, and then repeating the same procedure for each path
+     * string in the sequence. Each single resolution in the chain
+     * behaves as specified by {@link #resolve(String) resolve}.
      *
      * @implSpec
      * The default implementation is equivalent for this path to:
@@ -582,7 +578,7 @@ public interface Path
      * @return the resulting path
      *
      * @throws  InvalidPathException
-     *          if the path string cannot be converted to a Path.
+     *          if a path string cannot be converted to a Path.
      *
      * @see #resolve(String)
      *
--- a/src/java.base/share/classes/java/nio/file/Path.java
+++ b/src/java.base/share/classes/java/nio/file/Path.java
@@ -517,8 +517,10 @@ public interface Path
     }
 
     /**
-     * Resolves one or more {@code Path}s iteratively against this
-     * {@code Path}. If {@code more} does not specify any {@code Path}s,
+     * Resolves a path against this path, and then iteratively resolves any
+     * additional paths.
+     *
+     * <p> If {@code more} does not specify any {@code Path}s,
      * then the result is {@link #resolve(Path) resolve(first)}. If
      * {@code more} specifies one or more {@code Path}s, then each non-empty
      * {@code Path}, including {@code first}, is resolved iteratively against
@@ -545,11 +547,13 @@ public interface Path
      * }
      * }
      *
-     * @param first the first path to resolve against this path
+     * @param   first
+     *          the first path to resolve against this path
      *
-     * @param more additional paths to be iteratively resolved against this path
+     * @param   more
+     *          additional paths to be iteratively resolved against this path
      *
-     * @return the resulting path
+     * @return  the resulting path
      *
      * @see #resolve(Path)
      *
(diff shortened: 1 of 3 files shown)


bplb:
Default methods are added to the `Path` interface and a Unix version is added for the principal `resolve` method using `Path` parameters. The Unix implementation is about 20% faster than the default implementation.

src/java.base/share/classes/java/nio/file/Path.java
```diff
@@ -516,6 +516,89 @@ default Path resolve(String other) {
...
+     * @throws  InvalidPathException
+     *          if the path string cannot be converted to a Path.
+     *
+     * @see #resolve(Path)
+     *
+     * @since 22
+     */
+    default Path resolve(Path first, Path... more) {
+        Path result = resolve(first);
+        for (Path p : more) {
```
jaikiran, line 553: Hello Brian, do you think there should be a `@throws NullPointerException` in the javadoc, if any of these `Path` values is `null`?

src/java.base/share/classes/java/nio/file/Path.java
```diff
@@ -516,6 +516,89 @@ default Path resolve(String other) {
         return resolve(getFileSystem().getPath(other));
     }
 
+    /**
+     * Resolves a path or a sequence of paths against this path.
+     *
+     * <p> The result of this method is the same as would be obtained
+     * by resolving the first parameter path against this path, then
+     * resovling the second parameter path against the derived path,
```
jaikiran, line 524: Typo - should have been "resolving" instead of "resovling"

src/java.base/share/classes/java/nio/file/Path.java
```diff
@@ -516,6 +516,89 @@ default Path resolve(String other) {
...
+     * }
+     *
+     * @param first the first path to resolve against this path
+     *
+     * @param more additional paths to be joined to form the path
+     *
+     * @return the resulting path
+     *
+     * @throws  InvalidPathException
+     *          if the path string cannot be converted to a Path.
```
jaikiran, line 545: Is this `@throws InvalidPathException` applicable for this method, since this doesn't take any string arguments?

src/java.base/share/classes/java/nio/file/Path.java
```diff
@@ -516,6 +516,89 @@ default Path resolve(String other) {
...
+
+    /**
+     * Converts a path string or a sequence of path strings into one or more
+     * {@code Path}s and resolves them against this {@code Path} in exactly
+     * the same manner specified by the {@link #resolve(Path,Path...) resolve}
+     * method.
+     *
+     * <p> The result of this method is the same as would be obtained
+     * by resolving the first parameter path string against this path, then
+     * resovling the second parameter path string against the derived path,
```
jaikiran, line 567: Typo - should have been "resolving"

src/java.base/share/classes/java/nio/file/Path.java
```diff
@@ -516,6 +516,89 @@ default Path resolve(String other) {
...
+     * @return the resulting path
+     *
+     * @throws  InvalidPathException
+     *          if the path string cannot be converted to a Path.
+     *
+     * @see #resolve(Path)
+     *
+     * @since 22
+     */
+    default Path resolve(Path first, Path... more) {
```
ExE-Boss, line 551: It might be better to have a `(Path...)Path` overload, so that when you have an array of `Path`s or `String`s, you don’t have to manually extract the first element and copy the array using:
```java
path.resolve(paths[0], Arrays.copyOfRange(paths, 1, paths.length))
```

jaikiran: Hello @ExE-Boss 

> It might be better to have a (Path...)Path

Do you mean introducing:
```
default Path resolve(Path... more) {
...
}
```
If so, then it would allow calls like:

```
Path p = ...
p.resolve();
```
(i.e. calling resolve without any parameter)
I think that would be odd and wrong for the API to allow that.

src/java.base/share/classes/java/nio/file/Path.java
jaikiran: You're right indeed. The `package-info.java` does say:

```
   Unless otherwise noted, passing a {@code null} argument to a
 * constructor or method in any class or interface in this package
 * will cause a {@link java.lang.NullPointerException
 * NullPointerException} to be thrown.
```
Before commenting, I had checked if the class level documentation had this detail and didn't see it, but I had forgotten to check package level documentation.

src/java.base/share/classes/java/nio/file/Path.java
bplb: Fix in b55ade04905de70264a682404996f2f982964b01. Thanks.

src/java.base/share/classes/java/nio/file/Path.java
bplb: > I think that would be odd and wrong for the API to allow that.

Agreed.

src/java.base/share/classes/java/nio/file/Path.java
bplb: Fixed in dd8c1413fc42b6b526d77c7131a10969353e0d08.

src/java.base/share/classes/java/nio/file/Path.java
```diff
@@ -516,6 +516,86 @@ default Path resolve(String other) {
...
+     *
+     * @param first the first path string to resolve against this path
+     *
+     * @param more additional path strings to be joined to form the path
+     *
+     * @return the resulting path
+     *
+     * @throws  InvalidPathException
+     *          if the path string cannot be converted to a Path.
+     *
```
RogerRiggs, line 586: When an InvalidPathException is thrown, what information is available to identify the illegal string?

"if the path" -> "if a path" or "if any of the paths".

src/java.base/share/classes/java/nio/file/Path.java
```diff
@@ -516,6 +516,86 @@ default Path resolve(String other) {
         return resolve(getFileSystem().getPath(other));
     }
 
+    /**
+     * Resolves a path or a sequence of paths against this path.
```
RogerRiggs, line 520: "or"  -> "and then"

src/java.base/share/classes/java/nio/file/Path.java
```diff
@@ -516,6 +516,86 @@ default Path resolve(String other) {
...
+     * and so on until all parameters have been processed. Each single
+     * resolution in the chain behaves as specified by {@linkplain
+     * #resolve(String) resolve}.
+     *
+     * @implSpec
+     * The default implementation is equivalent for this path to:
+     * {@snippet lang=java :
+     * Path result = resolve(first);
+     * for (String s : more) {
+     *     result = result.resolve(s);
```
RogerRiggs, line 574: The code here doesn't match the description, converts to Paths and then resolves.
Though it might not be detectable.

test/jdk/java/nio/file/Path/PathOps.java
```diff
@@ -1669,6 +1706,26 @@ static void doUnixTests() {
...
+            .resolve("", "", "")
+            .resolve("/bar", "foo", "/bar", "")
+            .resolve("foo/bar/gus", "foo", "bar", "gus")
+            .resolve("baz", "", "", "baz");
+        test("/")
+            .resolve("/foo", "", "", "foo", "")
+            .resolve("/foo", "foo", "")
+            .resolve("/foo", "", "foo")
+            .resolve("/bar", "foo", "", "/bar");
+
```
RogerRiggs, line 1728: Should there be test cases that include multiple directories in a path:  "abc/def/xyz"?
Also cases with "..", or is that only relevant if canonicalizing.

src/java.base/share/classes/java/nio/file/Path.java
bplb: The message of `InvalidPathException` is most often created in `UnixPath`, `WindowsPath`, and `WindowsPathParser` and indicates the problem, e.g., `nul` character, unmappable character, etc.

Exception doc updated in 1b0c5228e05cc4a685f18b949be99f73f5e1b208.

src/java.base/share/classes/java/nio/file/Path.java
bplb: Like in `Path.of`, the first parameter may be considered element zero of the sequence which was the intent here.

src/java.base/share/classes/java/nio/file/Path.java
bplb: Description updated in 1b0c5228e05cc4a685f18b949be99f73f5e1b208.

test/jdk/java/nio/file/Path/PathOps.java
bplb: Multiple directories would be good but I don't know about `..`.

src/java.base/share/classes/java/nio/file/Path.java
```diff
@@ -516,6 +516,82 @@ default Path resolve(String other) {
         return resolve(getFileSystem().getPath(other));
     }
 
+    /**
+     * Resolves a path, or a sequence of paths against this path.
```
AlanBateman, line 520: I think we'll need to come up with a better first line because the method doesn't resolve the sequence of paths against this path, it instead resolves the first path against this path, then iteratively resolves.

I think the second paragraph will need to include something about empty elements, as we have in the existing resolve spec.

src/java.base/share/classes/java/nio/file/Path.java
```diff
@@ -516,6 +516,82 @@ default Path resolve(String other) {
...
+     * {@snippet lang=java :
+     * Path result = resolve(first);
+     * for (Path p : more) {
+     *     result = result.resolve(p);
+     * }
+     * }
+     *
+     * @param first the first path to resolve against this path
+     *
+     * @param more additional paths to be joined to form the path
```
AlanBateman, line 539: "additional paths to be joined to form the path" may be copied from elsewhere :-)

bplb: Blindly, even.

src/java.base/share/classes/java/nio/file/Path.java
```diff
@@ -516,6 +516,101 @@ default Path resolve(String other) {
         return resolve(getFileSystem().getPath(other));
     }
 
+    /**
+     * Resolves one or more {@code Path}s iteratively against this
+     * {@code Path}. If {@code more} does not specify any {@code Path}s,
```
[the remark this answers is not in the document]
bplb: Addressed in f78894c86eeeb4c94ed417eae43793c9dd1c80ee.

src/java.base/share/classes/java/nio/file/Path.java
```diff
@@ -516,6 +516,101 @@ default Path resolve(String other) {
...
+     * for (Path p : more) {
+     *     result = result.resolve(p);
+     * }
+     * }
+     *
+     * @param first the first path to resolve against this path
+     *
+     * @param more additional paths to be iteratively resolved against this path
+     *
+     * @return the resulting path
```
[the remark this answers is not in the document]
bplb: Addressed in f78894c86eeeb4c94ed417eae43793c9dd1c80ee.

src/java.base/share/classes/java/nio/file/Path.java
```diff
@@ -516,6 +516,101 @@ default Path resolve(String other) {
...
+        Path result = resolve(first);
+        for (Path p : more) {
+            result = result.resolve(p);
+        }
+        return result;
+    }
+
+    /**
+     * Resolves one or more {@code Path}s converted from the supplied path
+     * srings iteratively against this {@code Path}. If {@code more} does not
```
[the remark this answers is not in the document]
bplb: Addressed in f78894c86eeeb4c94ed417eae43793c9dd1c80ee.

src/java.base/unix/classes/sun/nio/fs/UnixPath.java
```diff
@@ -395,6 +395,84 @@ UnixPath resolve(byte[] other) {
...
+               }
+           }
+       }
+
+       // If the base is not being superseded by a descendant which is an
+       // absolute path, then if at least one descendant is non-empty and
+       // the base consists only of a '/', then decrement the length to
+       // account for an extra '/' added in the length computation.
+       if (start == 0 && length > base.length &&
+           base.length == 1 && base[0] == '/')
```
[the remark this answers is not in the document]
bplb: Addressed in f78894c86eeeb4c94ed417eae43793c9dd1c80ee.

src/java.base/unix/classes/sun/nio/fs/UnixPath.java
```diff
@@ -395,6 +395,84 @@ UnixPath resolve(byte[] other) {
...
+       // descendant is that last one which is an absolute path
+       int start = 0;
+       int length = base.length;
+
+       // Locate the last descendant which is an absolute path and calculate
+       // the total number of bytes in the resolved path
+       final int count = descendants.length;
+       if (count > 0) {
+           for (int i = 0; i < count; i++) {
+               byte[] b = descendants[i];
```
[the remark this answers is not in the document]
bplb: Addressed in f78894c86eeeb4c94ed417eae43793c9dd1c80ee.

src/java.base/unix/classes/sun/nio/fs/UnixPath.java
```diff
@@ -395,6 +395,84 @@ UnixPath resolve(byte[] other) {
         return resolve(new UnixPath(getFileSystem(), other));
     }
 
+   private static final byte[] resolve(byte[] base, byte[]... descendants) {
+       // 'start' is either zero, indicating the base, or indicates which
+       // descendant is that last one which is an absolute path
+       int start = 0;
+       int length = base.length;
```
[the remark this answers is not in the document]
bplb: Addressed in f78894c86eeeb4c94ed417eae43793c9dd1c80ee.

src/java.base/share/classes/java/nio/file/Path.java
```diff
@@ -516,6 +516,110 @@ default Path resolve(String other) {
...
 
+    /**
+     * Resolves a path against this path, and then iteratively resolves any
+     * additional paths.
+     *
+     * <p> If {@code more} does not specify any {@code Path}s,
+     * then the result is {@link #resolve(Path) resolve(first)}. If
+     * {@code more} specifies one or more {@code Path}s, then each non-empty
+     * {@code Path}, including {@code first}, is resolved iteratively against
+     * this {@code Path}.
```
AlanBateman, line 527: The second paragraph has the same issue as before, it says "resolved iteratively against this Path" but that isn't correct. Same issue in the description of the "more" method.  The first paragraph is correct, and I think we'll need to re-phrase the second paragraph to accurate specify how the iterative resolution works.

bplb: 0fa126fd2d3861c56b85093e8a2483afa1ad9ef6 attempts to improve the aforementioned verbiage.

test/jdk/java/nio/file/Path/PathOps.java
bplb: > Should there be test cases that include multiple directories in a path: "abc/def/xyz"?

Added some multiple directory cases in 505d72b48c5d34bbf4474c6fed43e2f269e8e135.

> Also cases with ".."

This is not relevant. For example, using the existing `resolve(Path)`:

```
jshell> java.nio.file.Path.of("/tmp").resolve("..").resolve("..")
$1 ==> /tmp/../..
```

src/java.base/share/classes/java/nio/file/Path.java
```diff
@@ -522,9 +522,9 @@ default Path resolve(String other) {
      *
      * <p> If {@code more} does not specify any {@code Path}s,
      * then the result is {@link #resolve(Path) resolve(first)}. If
-     * {@code more} specifies one or more {@code Path}s, then each non-empty
-     * {@code Path}, including {@code first}, is resolved iteratively against
-     * this {@code Path}.
+     * {@code more} specifies one or more {@code Path}s, then {@code first}
+     * is resolved against this path and then the additional paths in
+     * {@code more} are iteratively resolved.
```
ExE-Boss, line 527: Does the following better describe what this method does?
```suggestion
     * {@code more} are iteratively resolved against the previous path.
```

src/java.base/share/classes/java/nio/file/Path.java
AlanBateman: > 0fa126f  attempts to improve the aforementioned verbiage.

The wording in the latest (db4d9bd6) looks good.

test/jdk/java/nio/file/Path/PathOps.java
```diff
@@ -182,6 +183,19 @@ PathOps resolve(String other, String expected) {
...
+    // Note: "expected" is first parameter here
+    PathOps resolve(String expected, String first, String... more) {
+        out.format("test resolve %s varargs\n", path());
+        checkPath();
+        Path[] others = new Path[more.length];
+        int i = 0;
+        for (String s : more) {
+            others[i++] = Path.of(s);
+        }
+        check(path.resolve(Path.of(first), others), expected);
```
RogerRiggs, line 195: This looks like it tests only the new method with Path elements.
Are there tests for the new default method that accepts String.

test/jdk/java/nio/file/Path/PathOps.java
bplb: Addressed in c2a8faaa1b81dbb40e1a5989a02c9e5935417358.

test/jdk/java/nio/file/Path/PathOps.java
bplb: Addressed in c2a8faaa1b81dbb40e1a5989a02c9e5935417358. Thanks for pointing it out.

AlanBateman (approved):

PR #14807, opened 2023-07-07
8311666: Disabled tests in test/jdk/sun/java2d/marlin

cushon:


prrace (approved):
Hmm. Puzzling.
These tests seem to have been like this since (at least) JDK8u days.
I don't understand how they managed to be updated several times over the years without anyone addressing this. I hope I'm not missing something but they have no comment to explain it.
And they do pass for me on current JDK.
But best to push this on Monday not Friday in case there is a surprise.

cushon:
Thanks for taking a look!

> Hmm. Puzzling. These tests seem to have been like this since (at least) JDK8u days. I don't understand how they managed to be updated several times over the years without anyone addressing this. I hope I'm not missing something but they have no comment to explain it. And they do pass for me on current JDK. But best to push this on Monday not Friday in case there is a surprise.

I think that matches my understanding, that it was an oversight when they were originally contributed, and then it went unnoticed and unfixed since then.

Submitting earlier in the week sounds good to me.

bourgesl:
The test dates are from 2010 for pisces, moved into marlin tests once pisces was removed in 2017.
LGTM

PR #14814, opened 2023-07-10
8305636: Expand and clean up predicate classes and move them into separate files

chhagedorn:
This is the third clean-up PR towards fixing issues with Assertion Predicates ([JDK-8288981](https://bugs.openjdk.org/browse/JDK-8288981)). This patch does not change anything in the way the old Assertion Predicates work.

After collecting and moving the predicate code in the last clean-up PR https://github.com/openjdk/jdk/pull/14017 to the classes `Predicates/ParsePredicates`, I'm now completely moving the code to separate `predicates.cpp/hpp` files. By doing so, I also updated the predicate description and updated some namings. Since this description is also moved to the new files, I've committed the description update separately to better reflect these changes.

Changes include:
- Moved `Predicates/ParsePredicates` classes to new files `predicates.cpp/hpp`.
- Turning the `Predicates` utility class into a real class to represent all predicates:
  - Contains three `PredicateBlock` fields for each Predicate Block (see description of `Predicate Block`).
    - The `PredicateBlock` class offers methods to query the presence of predicates and to access them (e.g. get the Parse Predicate projection).
  - In the process, the `ParsePredicates` could be removed as the Parse Predicates are now covered by the `PredicateBlock` class.
- New `AssertionPredicatesWithHalt` class to skip over Assertion Predicates (will be further cleaned up later with the complete fix in JDK-8288981).
- Updated predicate description and moved to `predicates.hpp`.
  - While testing a prototype fix of JDK-8288981, I've came to the conclusion that we should not move all Assertion Predicates to a separate block below the Parse and Hoisted Predicates, because it prevented further application of Loop Predication due to pins of data nodes to these Assertion Predicates while the Hoisted Predicates needed them above the Assertion Predicates (i.e. dominance problems leading to bad graph assertions). I've removed that part of the description that gave a heads-up about that change.
- Small clean-ups such as variable renaming or code move.

Not included:
- Refactoring predicate traversal to clone/copy/initialize predicates for loop unswitching, pre/main/post, loop peeling etc. (this is only done in the actual fix in JDK-8288981 which requires some updates anyways - so this refactoring is not done here (yet)).

Testing: Tier1-7, hs-precheckin-comp, hs-comp-stress

Thanks,
Christian

---------
### Progress

### Issue
 * [JDK-8305636](https://bugs.openjdk.org/browse/JDK-8305636): Expand and clean up predicate classes and move them into separate files (**Sub-task** - P3)

### Reviewers
 * [Roland Westrelin](https://openjdk.org/census#roland) (@rwestrel - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14814/head:pull/14814` \
`$ git checkout pull/14814`

`$ git checkout pull/14814` \
`$ git pull https://git.openjdk.org/jdk.git pull/14814/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14814`

`$ git pr show -t 14814`

</details>

<a href="https://git.openjdk.org/jdk/pull/14814.diff">https://git.openjdk.org/jdk/pull/14814.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14814#issuecomment-1629190338)


src/hotspot/share/opto/predicates.hpp
```diff
@@ -0,0 +1,361 @@
...
+ *       Predicate           the case when adding the stride to the induction variable would cause an overflow which
+ *                           will not satisfy the loop limit exit condition. This overflow is unexpected for further
+ *                           counted loop optimizations and could lead to wrong results. Therefore, when this predicate
+ *                           fails at runtime, we must trap and recompile the method without turning the loop into a
+ *                           counted loop to avoid these overflow problems.
+ *                           The predicate does not replace an actual check inside the loop. This predicate can only
+ *                           be added once above the Loop Limit Check Parse Predicate for a loop.
+ * - Assertion Predicate: An always true predicate which will never fail (its range is already covered by an earlier
+ *                        Hoisted Predicate or the main-loop entry guard) but is required in order to fold away a dead
+ *                        sub loop inside which some data could be proven to be dead (by the type system) and replaced
```
TobiHartmann, line 72: This needs re-phrasing: "a dead sub loop inside which some data could be"

src/hotspot/share/opto/predicates.hpp
```diff
@@ -0,0 +1,361 @@
...
+ *                      Predicate) or a Loop Limit Check Predicate. These predicates will be checked at runtime while the
+ *                      Parse and Assertion Predicates are always removed before code generation (except for Initialized
+ *                      Assertion Predicates which are kept in debug builds while being removed in product builds).
+ *     - Hoisted Predicate: Either a Loop Predicate or a Profiled Loop Predicate that is created during Loop Predication
+ *                          to hoist a check out of a loop.
+ *         - Hoisted Invariant Check Predicate: A Hoisted Predicate to hoist a loop-invariant check out of a loop.
+ *         - Hoisted Range Check Predicate:     A Hoisted Predicate to hoist a range check of the form
+ *                                              "a[i*scale + offset]", where scale and offset are loop-invariant, out of
+ *                                              a counted loop. Each Hoisted Range Check Predicate is accompanied by
+ *                                              additional Assertion Predicates.
```
TobiHartmann, line 52: As we discussed offline, I would suggest to remove this part.

src/hotspot/share/opto/predicates.hpp
```diff
@@ -0,0 +1,361 @@
...
+ *                                                   unrolling the main-loop, we create two Initialized Assertion
+ *                                                   Predicates from the Template Assertion Predicates by replacing the
+ *                                                   OpaqueLoop* nodes by actual values for the unrolled loop.
+ *                                                   The Initialized Assertion Predicates are always true: They are true
+ *                                                   when entering the main-loop (because we adjusted the pre-loop exit
+ *                                                   condition), when executing the last iteration of the main-loop
+ *                                                   (because we adjusted the main-loop exit condition), and during all
+ *                                                   other iterations of the main-loop in-between by implication.
+ *                                                   Note that Range Check Elimination could remove additional range
+ *                                                   checks which we were not possible to remove with Loop Predication
```
TobiHartmann, line 136:
```suggestion
 *                                                   checks which were not possible to remove with Loop Predication
```

src/hotspot/share/opto/predicates.hpp
```diff
@@ -0,0 +1,361 @@
...
+
+  static Node* find_entry(Node* start_proj);
+  static bool has_opaque4(const Node* predicate_proj);
+  static bool has_halt(const Node* success_proj);
+  static bool is_assertion_predicate_success_proj(const Node* predicate_proj);
+
+ public:
+  AssertionPredicatesWithHalt(Node* assertion_predicate_proj) : _entry(find_entry(assertion_predicate_proj)) {}
+
+  // Returns the control input node into the first assertion predicate If. If there are no assertion predicates, it.
```
TobiHartmann, line 211:
```suggestion
  // Returns the control input node into the first assertion predicate If. If there are no assertion predicates, it
```

TobiHartmann (approved):
Nice refactoring. Looks good to me otherwise!

chhagedorn:
Thanks a lot Tobias for your careful review! I've pushed the discussed changes.

src/hotspot/share/opto/loopPredicate.cpp
```diff
@@ -321,15 +157,8 @@ IfProjNode* PhaseIdealLoop::create_new_if_for_predicate(IfProjNode* cont_proj, N
       fatal("no other If variant here");
   }
   register_control(new_iff, lp, entry);
-  IfProjNode* if_cont;
```
rwestrel, line 324: Why is it ok to remove that code?

rwestrel (approved):
Otherwise, looks good to me.

src/hotspot/share/opto/loopPredicate.cpp
chhagedorn: Thanks Roland for your review!

When I [originally introduced](https://github.com/openjdk/jdk/commit/a51dd58f388bb04851961dc5265291f01be422b8#diff-57d27180e9c833e88433025fd97a49467aed4a5d6e1867f86622dcffa47e951aL147-L149) the parameter `if_cont_is_true_proj`, I've copied all range checks to the unswitched loop (including hoisted check predicates for range checks). Later, this was [corrected](https://github.com/openjdk/jdk/commit/3fe5886b#diff-57d27180e9c833e88433025fd97a49467aed4a5d6e1867f86622dcffa47e951aL252) to only clone the template assertion predicates which always have a true projection as success projection. The only other case where we use `create_new_if_for_predicate` is with parse predicates which also only  have a true projection as success projection. I therefore cleaned this up here.

chhagedorn (commented):
Thanks Roland for your review!

src/hotspot/share/opto/loopPredicate.cpp
rwestrel: Thanks for the explanation.

chhagedorn:
Additional testing after merging with master looked good.

Thanks @TobiHartmann and @rwestrel again for your reviews!

PR #14817, opened 2023-07-10
8310551: vmTestbase/nsk/jdb/interrupt/interrupt001/interrupt001.java timed out due to missing prompt

plummercj:
After [JDK-8308232](https://bugs.openjdk.org/browse/JDK-8308232), both the test and the debuggee shared the same waittime of 5 minutes. The test would wait up to 5 minutes for the expected prompt, but the debuggee would also in some cases wait 5 minutes before generating the prompt, which sometimes was just a bit too late. Before [JDK-8308232](https://bugs.openjdk.org/browse/JDK-8308232), the debuggee would wait at most 2 minutes before generating the prompt, so it was always generated in time.

The main issue is that after the while loop checks that there are still uninterrupted threads remaining, the last of the threads is interrupted before the wait() call is made. This means wait() won't return until it times out, and by then it is too late, and the test side has already timed out waiting for the prompt.

The fix is to widen the synchronized block so the count of not interrupted threads remains correct until `wait()` is entered. The other choice was to refetch the count after entering the synchronized block, which is what I did for the initial fix, but widening seems a better choice so no refetch is needed.

For details see https://bugs.openjdk.org/browse/JDK-8310551?focusedCommentId=14594838&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14594838

Tested with 300 runs each on linux-x64 and linux-aarch64 with -Xcomp options that reproduced the issue, and then again with no options.

---------
### Progress

### Issue
 * [JDK-8310551](https://bugs.openjdk.org/browse/JDK-8310551): vmTestbase/nsk/jdb/interrupt/interrupt001/interrupt001.java timed out due to missing prompt (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14817/head:pull/14817` \
`$ git checkout pull/14817`

`$ git checkout pull/14817` \
`$ git pull https://git.openjdk.org/jdk.git pull/14817/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14817`

`$ git pr show -t 14817`

</details>

<a href="https://git.openjdk.org/jdk/pull/14817.diff">https://git.openjdk.org/jdk/pull/14817.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14817#issuecomment-1629829134)

### DIFF
--- a/test/hotspot/jtreg/vmTestbase/nsk/jdb/interrupt/interrupt001/interrupt001a.java
+++ b/test/hotspot/jtreg/vmTestbase/nsk/jdb/interrupt/interrupt001/interrupt001a.java
@@ -56,8 +56,8 @@ public class interrupt001a {
                         lock.wait();
                     }
                 } catch (InterruptedException e) {
-                    notInterrupted.decrementAndGet();
                     synchronized (waitnotify) {
+                        notInterrupted--;
                         waitnotify.notify();
                     }
                 }
@@ -83,7 +83,7 @@ public class interrupt001a {
     private JdbArgumentHandler argumentHandler;
     private Log log;
 
-    public static final AtomicInteger notInterrupted = new AtomicInteger(numThreads);
+    public static volatile int notInterrupted = numThreads;
 
     public int runIt(String args[], PrintStream out) {
 
@@ -123,7 +123,7 @@ public class interrupt001a {
         long waitTime = argumentHandler.getWaitTime() * 60 * 1000;
         long startTime = System.currentTimeMillis();
         synchronized (waitnotify) {
-            while (notInterrupted.get() > 0 && System.currentTimeMillis() - startTime <= waitTime) {
+            while (notInterrupted > 0 && System.currentTimeMillis() - startTime <= waitTime) {
                 try {
                     waitnotify.wait(waitTime);
                 } catch (InterruptedException e) {
--- a/test/hotspot/jtreg/vmTestbase/nsk/jdb/interrupt/interrupt001/interrupt001a.java
+++ b/test/hotspot/jtreg/vmTestbase/nsk/jdb/interrupt/interrupt001/interrupt001a.java
@@ -120,13 +120,17 @@ public class interrupt001a {
             breakHere();  // a break to get thread ids and then to interrupt MyThreads.
         } while (!allWorkersAreWaiting);
 
-        long waitTime = argumentHandler.getWaitTime() * 60 * 1000; // max time in loop
+        long waitTime = argumentHandler.getWaitTime() * 60 * 1000;
         long startTime = System.currentTimeMillis();
         while (notInterrupted.get() > 0 && System.currentTimeMillis() - startTime <= waitTime) {
             synchronized (waitnotify) {
                 try {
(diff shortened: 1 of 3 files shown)


kevinjwalls (approved):
Yes that look like a more normal sync/wait/notify block.

sspitsyn (approved):
Looks good.
Thanks,
Serguei

test/hotspot/jtreg/vmTestbase/nsk/jdb/interrupt/interrupt001/interrupt001a.java
```diff
@@ -83,7 +83,7 @@ static void breakHere () {}
     private JdbArgumentHandler argumentHandler;
     private Log log;
 
-    public static final AtomicInteger notInterrupted = new AtomicInteger(numThreads);
+    public static volatile int notInterrupted = numThreads;
```
dholmes-ora, line 86: Nit: If always accessed under lock this does not need to be volatile.

plummercj: It's also read outside the lock by the debugger side of the test using JDI -> JDWP -> JVMTI. My guess is there is so much other synchronization going on that it probably doesn't need volatile, but technically I think it should have it.

test/hotspot/jtreg/vmTestbase/nsk/jdb/interrupt/interrupt001/interrupt001a.java
```diff
@@ -122,8 +122,8 @@ public int runIt(String args[], PrintStream out) {
 
         long waitTime = argumentHandler.getWaitTime() * 60 * 1000;
         long startTime = System.currentTimeMillis();
-        while (notInterrupted.get() > 0 && System.currentTimeMillis() - startTime <= waitTime) {
-            synchronized (waitnotify) {
+        synchronized (waitnotify) {
+            while (notInterrupted > 0 && System.currentTimeMillis() - startTime <= waitTime) {
```
viktorklang-ora, line 126: @plummercj I'd use a monotonic clock like `System.nanoTime()` for things like this since the system clock can be changed to anything at any time.

plummercj: > I'd use a monotonic clock like `System.nanoTime()` for things like this since the system clock can be changed to anything at any time.

We have 100's of uses of `System.currentTimeMillis()` in our tests. I think changing them should be handled by a separate effort.

plummercj:
@kevinjwalls and @sspitsyn can you please review the latest changes. Thanks!

sspitsyn (approved):
The latest simplification looks good.
Thanks,
Serguei

kevinjwalls (approved):

plummercj:
Thanks Serguei and Kevin!

/integrate

PR #14820, opened 2023-07-11
8309603: Deallocate hashtables in ClassLoaderData::unload

chadrako:
[JDK-8309603](https://bugs.openjdk.org/browse/JDK-8309603)

Memory is released in `ClassLoaderdata::~ClassLoaderData` for the dictionary and module hashtables but they could be deleted earlier in `ClassLoaderdata::unload`. Since we're unloading the class loader is no longer being used for lookup.

Additional testing:
- [x] Linux x86_64 fastdebug `tier2`
- [x] Linux x86_64 release `tier2`
- [x] Linux x86_64 fastdebug `gtest:all`
- [x] Linux x86_64 release `gtest:all`
- [x] Linux x86_64 fastdebug `test/hotspot/jtreg/runtime`
- [x] Linux x86_64 release `test/hotspot/jtreg/runtime`

---------
### Progress

### Issue
 * [JDK-8309603](https://bugs.openjdk.org/browse/JDK-8309603): Deallocate hashtables in ClassLoaderData::unload (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14820/head:pull/14820` \
`$ git checkout pull/14820`

`$ git checkout pull/14820` \
`$ git pull https://git.openjdk.org/jdk.git pull/14820/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14820`

`$ git pr show -t 14820`

</details>

<a href="https://git.openjdk.org/jdk/pull/14820.diff">https://git.openjdk.org/jdk/pull/14820.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14820#issuecomment-1631455423)

### DIFF
--- a/src/hotspot/share/classfile/classLoaderData.cpp
+++ b/src/hotspot/share/classfile/classLoaderData.cpp
@@ -624,6 +624,13 @@ void ClassLoaderData::unload() {
   // Clean up global class iterator for compiler
   ClassLoaderDataGraph::adjust_saved_class(this);
 
+  // Release C heap allocated hashtable for all the packages.
+  if (_packages != nullptr) {
+    // Destroy the table itself
+    delete _packages;
+    _packages = nullptr;
+  }
+
   // Release C heap allocated hashtable for all the modules.
   if (_modules != nullptr) {
     // Destroy the table itself
@@ -745,13 +752,6 @@ ClassLoaderData::~ClassLoaderData() {
 
   // Release the WeakHandle
   _holder.release(Universe::vm_weak());
-
-  // Release C heap allocated hashtable for all the packages.
-  if (_packages != nullptr) {
-    // Destroy the table itself
-    delete _packages;
-    _packages = nullptr;
-  }
 
   // release the metaspace
   ClassLoaderMetaspace *m = _metaspace;
--- a/src/hotspot/share/classfile/classLoaderData.cpp
+++ b/src/hotspot/share/classfile/classLoaderData.cpp
@@ -624,6 +624,13 @@ void ClassLoaderData::unload() {
   // Clean up global class iterator for compiler
   ClassLoaderDataGraph::adjust_saved_class(this);
 
+  // Release C heap allocated hashtable for all the packages.
+  if (_packages != nullptr) {
+    // Destroy the table itself
+    delete _packages;
+    _packages = nullptr;
+  }
+
   // Release C heap allocated hashtable for all the modules.
   if (_modules != nullptr) {
     // Destroy the table itself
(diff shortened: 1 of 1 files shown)


src/hotspot/share/classfile/classLoaderData.cpp
```diff
@@ -734,20 +748,6 @@ ClassLoaderData::~ClassLoaderData() {
     _packages = nullptr;
   }
```
coleenp, line 749: Can you do packages too?

src/hotspot/share/classfile/classLoaderData.cpp
```diff
@@ -734,20 +748,6 @@ ClassLoaderData::~ClassLoaderData() {
...
-  }
-
-  // Release C heap allocated hashtable for the dictionary
-  if (_dictionary != nullptr) {
-    // Destroy the table itself
-    delete _dictionary;
-    _dictionary = nullptr;
-  }
-
   if (_unnamed_module != nullptr) {
```
coleenp, line 751: And maybe unnamed module?

coleenp (changes requested):

dholmes-ora:
When do we delete the ClassLoaderData? Does moving this to unload time help with a specific problem? It seems odd to split the deallocation operations between unload and the destructor - why does unload not just call the destructor?

coleenp:
Unload cannot just call the destructor because it's not safe to release the metaspace chunk yet, until the next handshake operation.  Since ZGenerational may take a while to request a handshake, this might be able to release some memory beforehand.  @fisk asked for this.
Did releasing the package table not work?

chadrako:
Started failing a handful of tier1/tier2 tests after moving packages and unnamed module to unload. Running a couple more tests to see what the issue is

chadrako:
> Started failing a handful of tier1/tier2 tests after moving packages and unnamed module to unload. Running a couple more tests to see what the issue is

Failed tests passed when ran again. They do not appear to be related

coleenp (approved):
This looks good.  While it's nicer to remove these in the destructor, this releases memory more eagerly for class unloading without waiting for a safepoint.

fisk:
The design for concurrent class unloading was to essentially do this:
1. Unlink unloading data
2. Handshake all threads potentially accessing unloading data
3. Delete unloading data
The reason for this order, is that it provides safe memory reclamation. We want to do this consistently, so we don't get races where the code works 99.999999% of the time, except when you get a nasty race that crashes the JVM.
So unfortunately, this patch goes in the wrong direction, I think. I don't think we should do this.

PR #14821, opened 2023-07-11
6211126: ICC_ColorSpace.toCIEXYZ(float[]): NPE is not specified

mrserb:
The specification of some methods has been updated to include current behaviour on null.

---------
### Progress
- [x] Change requires CSR request [JDK-8311850](https://bugs.openjdk.org/browse/JDK-8311850) to be approved

### Issues
 * [JDK-6211126](https://bugs.openjdk.org/browse/JDK-6211126): ICC_ColorSpace.toCIEXYZ(float[]): NPE is not specified (**Bug** - P4)
 * [JDK-6211139](https://bugs.openjdk.org/browse/JDK-6211139): ICC_ColorSpace.toRGB(float[]): NPE is not specified (**Bug** - P4)
 * [JDK-8311850](https://bugs.openjdk.org/browse/JDK-8311850): ICC_ColorSpace.toCIEXYZ(float[]): NPE is not specified (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14821/head:pull/14821` \
`$ git checkout pull/14821`

`$ git checkout pull/14821` \
`$ git pull https://git.openjdk.org/jdk.git pull/14821/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14821`

`$ git pr show -t 14821`

</details>

<a href="https://git.openjdk.org/jdk/pull/14821.diff">https://git.openjdk.org/jdk/pull/14821.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14821#issuecomment-1630110022)

### DIFF
--- a/src/java.desktop/share/classes/java/awt/color/CMMException.java
+++ b/src/java.desktop/share/classes/java/awt/color/CMMException.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1997, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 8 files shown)


prrace (approved):
I think this is OK. The two things that popped into my head were
(1) Should the  abstract methods in ColorSpace really dictate what the concrete methods should do, but
  (a) We already do some of that
  (b) it is hard to imagine a scenario in which the args aren't needed so NPE would only be averted if some implementation decided to handle it in an unlikely way
(2) should the concrete methods explicitly check for null and immediately throw NPE but (b) above should mean it isn't needed, and the AIOBE cases seem like they might be the ones that need it more but don't seem to have an explicit check.

PR #14822, opened 2023-07-11
8311847: Fix -Wconversion for assembler.hpp emit_int8,16 callers

coleenp:
Please review changes to fix -Wconversion warnings that come from assembler_<cpu>.cpp by adding narrow_casts to the emit_int8,16,24, and 32 functions.  And some other fixups with checked_cast.

Ran tier1 on Oracle platforms, and tier1-4 on linux-x64-debug, linux-aarch64-debug, windows-x64-debug.

---------
### Progress

### Issue
 * [JDK-8311847](https://bugs.openjdk.org/browse/JDK-8311847): Fix -Wconversion for assembler.hpp emit_int8,16 callers (**Enhancement** - P4)

### Reviewers
 * [Dean Long](https://openjdk.org/census#dlong) (@dean-long - **Reviewer**) ⚠️ Review applies to [71c8fdec](https://git.openjdk.org/jdk/pull/14822/files/71c8fdeca5e449145806dbda6a4d58ab2ce62465)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14822/head:pull/14822` \
`$ git checkout pull/14822`

`$ git checkout pull/14822` \
`$ git pull https://git.openjdk.org/jdk.git pull/14822/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14822`

`$ git pr show -t 14822`

</details>

<a href="https://git.openjdk.org/jdk/pull/14822.diff">https://git.openjdk.org/jdk/pull/14822.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14822#issuecomment-1629966377)

### DIFF
--- a/src/hotspot/cpu/aarch64/assembler_aarch64.hpp
+++ b/src/hotspot/cpu/aarch64/assembler_aarch64.hpp
@@ -221,7 +221,7 @@ public:
     mask <<= lsb;
     unsigned target = *(unsigned *)a;
     target &= ~mask;
-    target |= checked_cast<unsigned>(val);
+    target |= (unsigned)val;
     *(unsigned *)a = target;
   }
 
@@ -236,7 +236,7 @@ public:
     mask <<= lsb;
     unsigned target = *(unsigned *)a;
     target &= ~mask;
-    target |= checked_cast<unsigned>(uval);
+    target |= (unsigned)uval;
     *(unsigned *)a = target;
   }
 
@@ -265,7 +265,7 @@ public:
     uint64_t uval = val;
     unsigned mask = checked_cast<unsigned>(right_n_bits(nbits));
     uval &= mask;
-    f(checked_cast<unsigned>(uval), lsb + nbits - 1, lsb);
+    f((unsigned)uval, lsb + nbits - 1, lsb);
   }
 
   void rf(Register r, int lsb) {
--- a/src/hotspot/cpu/aarch64/assembler_aarch64.hpp
+++ b/src/hotspot/cpu/aarch64/assembler_aarch64.hpp
@@ -221,7 +221,7 @@ public:
     mask <<= lsb;
     unsigned target = *(unsigned *)a;
     target &= ~mask;
-    target |= val;
+    target |= checked_cast<unsigned>(val);
     *(unsigned *)a = target;
   }
 
@@ -229,14 +229,14 @@ public:
     int nbits = msb - lsb + 1;
     int64_t chk = val >> (nbits - 1);
     guarantee (chk == -1 || chk == 0, "Field too big for insn at " INTPTR_FORMAT, p2i(a));
-    unsigned uval = val;
+    uint64_t uval = val;
     unsigned mask = checked_cast<unsigned>(right_n_bits(nbits));
     uval &= mask;
     uval <<= lsb;
     mask <<= lsb;
     unsigned target = *(unsigned *)a;
     target &= ~mask;
-    target |= uval;
+    target |= checked_cast<unsigned>(uval);
     *(unsigned *)a = target;
   }
 
@@ -262,10 +262,10 @@ public:
     int nbits = msb - lsb + 1;
     int64_t chk = val >> (nbits - 1);
     guarantee (chk == -1 || chk == 0, "Field too big for insn");
-    unsigned uval = val;
+    uint64_t uval = val;
     unsigned mask = checked_cast<unsigned>(right_n_bits(nbits));
     uval &= mask;
-    f(uval, lsb + nbits - 1, lsb);
+    f(checked_cast<unsigned>(uval), lsb + nbits - 1, lsb);
   }
 
   void rf(Register r, int lsb) {
@@ -553,7 +553,7 @@ class Address {
           i->sf(offset(), 20, 12);
         } else {
           i->f(0b01, 25, 24);
-          i->f(offset() >> size, 21, 10);
+          i->f(checked_cast<unsigned>(offset() >> size), 21, 10);
         }
       }
       break;
@@ -653,7 +653,7 @@ class Address {
   static bool offset_ok_for_sve_immed(int64_t offset, int shift, int vl /* sve vector length */) {
     if (offset % vl == 0) {
       // Convert address offset into sve imm offset (MUL VL).
-      int sve_offset = offset / vl;
+      int64_t sve_offset = offset / vl;
       if (((-(1 << (shift - 1))) <= sve_offset) && (sve_offset < (1 << (shift - 1)))) {
         // sve_offset can be encoded
         return true;
@@ -2412,7 +2412,7 @@ public:
       ld_st(Vt, T, a.base(), op1, op2);
       break;
     case Address::post:
-      ld_st(Vt, T, a.base(), a.offset(), op1, op2, regs);
+      ld_st(Vt, T, a.base(), checked_cast<int>(a.offset()), op1, op2, regs);
       break;
     case Address::post_reg:
       ld_st(Vt, T, a.base(), a.index(), op1, op2);
@@ -3522,7 +3522,7 @@ private:
               int op1, int type, int imm_op2, int scalar_op2) {
     switch (a.getMode()) {
     case Address::base_plus_offset:
-      sve_ld_st1(Zt, a.base(), a.offset(), Pg, T, op1, type, imm_op2);
+      sve_ld_st1(Zt, a.base(), checked_cast<int>(a.offset()), Pg, T, op1, type, imm_op2);
       break;
     case Address::base_plus_offset_reg:
       sve_ld_st1(Zt, a.base(), a.index(), Pg, T, op1, type, scalar_op2);
--- a/src/hotspot/share/asm/assembler.hpp
+++ b/src/hotspot/share/asm/assembler.hpp
@@ -295,15 +295,14 @@ class AbstractAssembler : public ResourceObj  {
   // that the range is correct.
   template <typename T>
   constexpr T narrow_cast(int x) const {
-   if (x < 0) {
-    using stype = std::make_signed_t<T>;
-    assert(x >= std::numeric_limits<stype>::min(), "too negative"); // >= -128 for 8 bits
-    return static_cast<T>(x);  // cut off sign bits
-   } else {
-     return checked_cast<T>(x);
-   }
+    if (x < 0) {
+      using stype = std::make_signed_t<T>;
+      assert(x >= std::numeric_limits<stype>::min(), "too negative"); // >= -128 for 8 bits
+      return static_cast<T>(x);  // cut off sign bits
+    } else {
+      return checked_cast<T>(x);
+    }
   }
-
 
  public:
 
--- a/src/hotspot/share/asm/assembler.hpp
+++ b/src/hotspot/share/asm/assembler.hpp
@@ -291,10 +291,13 @@ class AbstractAssembler : public ResourceObj  {
 #endif
 
   // sign-extended tolerant cast needed by callers of emit_int8 and emit_int16
+  // Some callers pass signed types that need to fit into the unsigned type so check
+  // that the range is correct.
   template <typename T>
   constexpr T narrow_cast(int x) const {
    if (x < 0) {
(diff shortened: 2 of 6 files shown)


src/hotspot/cpu/aarch64/assembler_aarch64.hpp
```diff
@@ -221,22 +221,22 @@ class Instruction_aarch64 {
     mask <<= lsb;
     unsigned target = *(unsigned *)a;
     target &= ~mask;
-    target |= val;
+    target |= checked_cast<unsigned>(val);
```
dean-long, line 224: Any value that doesn't fit in 32 bits is going to fail, so it's tempting to force the callers to pass 32-bit types, but that's a bigger change.  How about something like this:

static ALWAYSINLINE void patch(address a, int msb, int lsb, uint32_t val) {
/* original code, no additional checked_cast needed */
}

static ALWAYSINLINE void patch(address a, int msb, int lsb, uint64_t val) {
  patch(a, msb, lsb, checked_cast<unsigned>(val));
}

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
```diff
@@ -221,22 +221,22 @@ class Instruction_aarch64 {
...
-    unsigned uval = val;
+    uint64_t uval = val;
     unsigned mask = checked_cast<unsigned>(right_n_bits(nbits));
     uval &= mask;
     uval <<= lsb;
     mask <<= lsb;
     unsigned target = *(unsigned *)a;
     target &= ~mask;
-    target |= uval;
+    target |= checked_cast<unsigned>(uval);
```
dean-long, line 239: Same suggestion as patch() above.

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
```diff
@@ -262,10 +262,10 @@ class Instruction_aarch64 {
     int nbits = msb - lsb + 1;
     int64_t chk = val >> (nbits - 1);
     guarantee (chk == -1 || chk == 0, "Field too big for insn");
-    unsigned uval = val;
+    uint64_t uval = val;
```
dean-long, line 265:
```suggestion
    unsigned uval = checked_cast<unsigned>((uint64_t)val);
```

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
```diff
@@ -262,10 +262,10 @@ class Instruction_aarch64 {
...
     unsigned mask = checked_cast<unsigned>(right_n_bits(nbits));
     uval &= mask;
-    f(uval, lsb + nbits - 1, lsb);
+    f(checked_cast<unsigned>(uval), lsb + nbits - 1, lsb);
```
dean-long, line 268:
```suggestion
    f(uval, lsb + nbits - 1, lsb);
```

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
```diff
@@ -553,7 +553,7 @@ class Address {
           i->sf(offset(), 20, 12);
         } else {
           i->f(0b01, 25, 24);
-          i->f(offset() >> size, 21, 10);
+          i->f(checked_cast<unsigned>(offset() >> size), 21, 10);
```
dean-long, line 556: I remember there being issues with checked_cast and sign extension.  When going from int64_t to unsigned and back, I think we need to do int64_t --> int32_t --> unsigned, and not int64_t --> uint64_t --> unsigned.  Is that what checked_cast will do?  To be safe, or at least make it easier to understand, shound't we use checked_cast only to change the size or sign, but not both?  So going to int64_t to unsigned would require two checked_casts.

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
dean-long: See my comment below about not trusting checked_cast to do the right thing for int64_t --> unsigned.

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
dean-long: uint64_t uval64 = val;
    unsigned uval = checked_cast<unsigned>(uval64);
This won't work for negative values, right?

src/hotspot/share/asm/assembler.hpp
```diff
@@ -298,15 +310,22 @@ class AbstractAssembler : public ResourceObj  {
...
+  void emit_int24(      int x1,     int x2,     int x3)             { code_section()->emit_int24(narrow_cast<uint8_t>(x1),
+                                                                                                 narrow_cast<uint8_t>(x2),
+                                                                                                 narrow_cast<uint8_t>(x3)); }
 
   void emit_int32(  uint32_t x)                                     { code_section()->emit_int32(x); }
-  void emit_int32(  uint8_t x1, uint8_t x2, uint8_t x3, uint8_t x4) { code_section()->emit_int32(x1, x2, x3, x4); }
+  void emit_int32(      int x1,     int x2,     int x3,     int x4) { code_section()->emit_int32(narrow_cast<uint8_t>(x1),
+                                                                                                 narrow_cast<uint8_t>(x2),
+                                                                                                 narrow_cast<uint8_t>(x3),
+                                                                                                 narrow_cast<uint8_t>(x4)); }
```
dean-long, line 328: I'd rather add new alternative entry points for "int", so the existing callers using uint8_t don't need to perform unnecessary widening and narrowing.
Instead of narrow_cast<uint8_t>(x), how about (uint8_t)checked_cast<int8_t>(x)?

src/hotspot/share/asm/assembler.hpp
```diff
@@ -290,6 +290,18 @@ class AbstractAssembler : public ResourceObj  {
...
+   if (x < 0) {
+    assert(x > -(std::numeric_limits<T>::max() - 1), "too negative");
+    return static_cast<T>(x);  // cut off sign bits
+   } else {
+     return checked_cast<T>(x);
+   }
+  }
+
+
  public:
```
dean-long, line 305: I don't think we need this.  See below.

src/hotspot/share/asm/assembler.hpp
dean-long: How many callers are passing in negative values and actually need these convenience functions?

src/hotspot/share/asm/assembler.hpp
dean-long: Nevermind, I tried my alternative idea below and it didn't work.  For these particular cases where we only care about going to uint8_t, we could check is8bit().  Another trick I've seen is checking if (val >> width) is 0 or -1.

src/hotspot/share/asm/assembler.hpp
```diff
@@ -290,6 +290,18 @@ class AbstractAssembler : public ResourceObj  {
   };
 #endif
 
+  // sign-extended tolerant cast needed by callers of emit_int8 and emit_int16
+  template <typename T>
+  constexpr T narrow_cast(int x) const {
+   if (x < 0) {
+    assert(x > -(std::numeric_limits<T>::max() - 1), "too negative");
```
dean-long, line 297:
```suggestion
    assert(x >= -std::numeric_limits<T>::max(), "too negative"); // >= -256 for 8 bits
```

coleenp: Thank you - I meant to check these values.

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
coleenp: val comes in signed, so we want to just chop off the sign.  checked_cast<unsigned>(signed val) will assert.  checked_cast<> doesn't do sign conversion.  We don't have a cast that does sign conversion.

coleenp: Not sure I trust this either.  I'm going to write a gtest for this.

src/hotspot/share/asm/assembler.hpp
coleenp: There is an is8bit() test that precedes these casts.

src/hotspot/share/asm/assembler.hpp
coleenp: > How many callers?

Actually quite a lot for the emit_int16/24/32 ones.  They pass an int imm8 value and have some (value | encode) parameter where encode is an int passed in.

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
theRealAph: We already have a havrd guarantee that this signed value fits in the signed field. There's no need for any more asserting for that: just chop it off.

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
theRealAph: That's right.

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
coleenp:
```
src/hotspot/cpu/aarch64/nativeInst_aarch64.hpp:614:35: error: call to 'patch' is ambiguous
[2023-07-11T14:00:01,568Z]   void set_kind(int order_kind) { Instruction_aarch64::patch(addr_at(0), 11, 8, order_kind); }
[2023-07-11T14:00:01,568Z]                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~
```
The compiler thinks its ambiguous.

dean-long: OK, nevermind.

src/hotspot/share/asm/assembler.hpp
```diff
@@ -290,6 +290,21 @@ class AbstractAssembler : public ResourceObj  {
...
+  // that the range is correct.
+  template <typename T>
+  constexpr T narrow_cast(int x) const {
+   if (x < 0) {
+    using stype = std::make_signed_t<T>;
+    assert(x >= std::numeric_limits<stype>::min(), "too negative"); // >= -128 for 8 bits
+    return static_cast<T>(x);  // cut off sign bits
+   } else {
+     return checked_cast<T>(x);
+   }
```
theRealAph, line 304: Indentation looks wrong?

dean-long (approved):
This seems more than adequate for removing -Wconversion warnings.

theRealAph (approved):
Looks good, thanks. After all that back-and-forth I think you've come up with a clean and maintainable solution to a tricky problem.

coleenp:
Thank you for your time and attention, @aph and @dean-long.

jaikiran:
Hello Coleen, looks like the github actions job failure in this PR for windows-aarch64 are genuine - I see it failing in some other recent unrelated PRs containing this commit, like here https://github.com/openjdk/jdk/pull/15012

```
=== Output from failing command(s) repeated here ===
* For target hotspot_variant-server_libjvm_gtest_objs_BUILD_GTEST_LIBJVM_pch.obj:
BUILD_GTEST_LIBJVM_pch.cpp
d:\a\jdk\jdk\src\hotspot\cpu\aarch64\assembler_aarch64.hpp(657): error C2220: the following warning is treated as an error
d:\a\jdk\jdk\src\hotspot\cpu\aarch64\assembler_aarch64.hpp(657): warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?)
   ... (rest of output omitted)

* All command lines available in /d/a/jdk/jdk/build/windows-aarch64/make-support/failure-logs.
=== End of repeated output ===

No indication of failed target found.
HELP: Try searching the build log for '] Error'.
```

coleenp:
@jaikiran I don't think we have that platform.  Do you? Can you fix it?

jaikiran:
Hello Coleen, I didn't have access to this platform other than Github actions. I see that you fixed this issue in https://bugs.openjdk.org/browse/JDK-8312979, so thank you for that.

PR #14825, opened 2023-07-11
8280482: Window transparency bug on Linux

TejeshR13:
The bug mentions about transparency issue on Linux which actually got resolved with [JDK-8006421](https://bugs.openjdk.org/browse/JDK-8006421) fix. Now there is another problem related to the bug, which is screen selection going wrong during checking for new screen when the test is run. The problem is that (As seen in the pictures attached in bug) the transparency is lost for windows which are in screen 0 (default screen) too, which is not supposed to happen where windows on screen 1 should have lost there transparency. The main reason being the call to `toGlobal()` when window bounds are passed to `checkIfOnNewScreen` method. I didn't get the actual reason for `toGlobal` being used here (actually not required to check monitor number), but it actually doubled the X position of the window [here]( https://github.com/openjdk/jdk/blob/4b1403d06b99b91ddd89ad6e54669b0595f1f8e5/src/java.desktop/unix/classes/sun/awt/X11/XBaseWindow.java#L776). Removing `toGlobal` solve the issue and neither didn't cause any regression (existing test + the reason `toGlobal()` was added [JDK-8143295](https://bugs.openjdk.org/browse/JDK-8143295)).

The automated test fails if GC is changed for windows within screen 0 (default screen). CI testing is green.

---------
### Progress

### Issue
 * [JDK-8280482](https://bugs.openjdk.org/browse/JDK-8280482): Window transparency bug on Linux (**Bug** - P4)

### Reviewers
 * [Damon Nguyen](https://openjdk.org/census#dnguyen) (@DamonGuy - Committer)
 * [Alexander Zvegintsev](https://openjdk.org/census#azvegint) (@azvegint - **Reviewer**) ⚠️ Review applies to [4e47192a](https://git.openjdk.org/jdk/pull/14825/files/4e47192aca5a6fa4c25060a64f5c0d9dcf24703d)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14825/head:pull/14825` \
`$ git checkout pull/14825`

`$ git checkout pull/14825` \
`$ git pull https://git.openjdk.org/jdk.git pull/14825/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14825`

`$ git pr show -t 14825`

</details>

<a href="https://git.openjdk.org/jdk/pull/14825.diff">https://git.openjdk.org/jdk/pull/14825.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14825#issuecomment-1630610559)

### DIFF
--- a/src/java.desktop/unix/classes/sun/awt/X11/XWindowPeer.java
+++ b/src/java.desktop/unix/classes/sun/awt/X11/XWindowPeer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2002, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2002, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
--- a/src/java.desktop/unix/classes/sun/awt/X11/XWindowPeer.java
+++ b/src/java.desktop/unix/classes/sun/awt/X11/XWindowPeer.java
@@ -804,10 +804,10 @@ class XWindowPeer extends XPanelPeer implements WindowPeer,
         if (insLog.isLoggable(PlatformLogger.Level.FINE)) {
             insLog.fine(xe.toString());
         }
-        checkIfOnNewScreen(toGlobal(new Rectangle(scaleDown(xe.get_x()),
+        checkIfOnNewScreen(new Rectangle(scaleDown(xe.get_x()),
                 scaleDown(xe.get_y()),
                 scaleDown(xe.get_width()),
-                scaleDown(xe.get_height()))));
+                scaleDown(xe.get_height())));
 
         Rectangle oldBounds = getBounds();
 
--- a/test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
+++ b/test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
@@ -53,6 +53,8 @@ public class MultiScreenCheckScreenIDTest extends MouseAdapter {
     private static GraphicsDevice[] screens;
     static List<Window> windowList = new ArrayList<>();
     static Robot robot;
+    static JWindow window;
+
 
     public static void main(final String[] args) throws Exception {
         try {
@@ -89,7 +91,7 @@ public class MultiScreenCheckScreenIDTest extends MouseAdapter {
                 try {
                     SwingUtilities.invokeAndWait(() -> {
                         try {
-                            createWindow(r);
+                            window = createWindow(r);
                         } catch (Exception e) {
                             throw new RuntimeException(e);
                         }
@@ -99,35 +101,34 @@ public class MultiScreenCheckScreenIDTest extends MouseAdapter {
                 }
                 robot.delay(50);
                 robot.waitForIdle();
-                if (windowList.get(windowList.size() - 1).getBounds().intersects
-                        (screen.getDefaultConfiguration().getBounds())) {
-                    if (!(windowList.get(windowList.size() - 1).
-                            getGraphicsConfiguration().getBounds().
-                            intersects(screen.getDefaultConfiguration().
-                                    getBounds()))) {
+                if (window.getBounds().intersects(screenBounds)) {
+                    if (!(window.getGraphicsConfiguration().getBounds().
+                            intersects(screenBounds))) {
                         throw new RuntimeException("Graphics configuration " +
                                 "changed for screen :" + screenNumber);
                     }
                 }
+                windowList.add(window);
             }
             screenNumber++;
         }
     }
 
-    private void createWindow(Rectangle bounds) {
+    private JWindow createWindow(Rectangle bounds) {
         JWindow window = new JWindow();
         window.setBounds(bounds);
         window.setBackground(BACKGROUND);
         window.setAlwaysOnTop(true);
         window.addMouseListener(this);
         window.setVisible(true);
-        windowList.add(window);
+        return window;
     }
 
     @Override
     public void mouseClicked(MouseEvent e) {
         ((Window) e.getSource()).dispose();
     }
+
     private static List<Rectangle> gridOfRectangles(Rectangle r, int cols, int rows) {
         List<Rectangle> l = new ArrayList<>();
         for (int row = 0; row < rows; row++) {
--- a/test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
+++ b/test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
@@ -76,8 +76,9 @@ public class MultiScreenCheckScreenIDTest extends MouseAdapter {
                 .getLocalGraphicsEnvironment()
                 .getScreenDevices();
 
-        if(screens.length < 2) {
-            System.out.println("Testing aborted. Required min of 2 screens. Found : " + screens.length );
+        if (screens.length < 2) {
+            System.out.println("Testing aborted. Required min of 2 screens. " +
+                    "Found : " + screens.length );
             return;
         }
 
@@ -112,17 +113,21 @@ public class MultiScreenCheckScreenIDTest extends MouseAdapter {
         window.setVisible(true);
         windowList.add(window);
 
-        window.addPropertyChangeListener("graphicsConfiguration", new PropertyChangeListener() {
+        window.addPropertyChangeListener("graphicsConfiguration",
+                new PropertyChangeListener() {
             @Override
             public void propertyChange(PropertyChangeEvent propertyChangeEvent) {
                 double windowXPos = window.getBounds().getX();
-                double screen1Width = screens[0].getDefaultConfiguration().getBounds().getWidth();
-                System.out.println("window x Position : " + windowXPos + ", screen[0] width : " + screen1Width);
+                double screen1Width = screens[0].getDefaultConfiguration()
+                        .getBounds().getWidth();
+                System.out.println("window x Position : " + windowXPos + "," +
+                        " screen[0] width : " + screen1Width);
 
                 //Check if GC is changed for windows positioned in screen 0.
 
                 if (windowXPos < screen1Width) {
-                    throw new RuntimeException("Graphics configuration changed for screen 1");
+                    throw new RuntimeException("Graphics configuration changed" +
+                            " for screen 1");
                 }
             }
         });
--- a/test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
+++ b/test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
@@ -48,6 +48,7 @@ import java.util.List;
 
 public class MultiScreenCheckScreenIDTest {
     private static final int COLS = 12;
+    private static final int ROWS = 8;
     private static final Color BACKGROUND = new Color(0, 0, 255, 64);
     private static GraphicsDevice[] screens;
     static List<Window> windowList = new ArrayList<>();
@@ -86,7 +87,7 @@ public class MultiScreenCheckScreenIDTest {
         for (GraphicsDevice screen : screens) {
             Rectangle screenBounds = screen.getDefaultConfiguration().getBounds();
 
-            for (Rectangle r : gridOfRectangles(screenBounds, COLS)) {
+            for (Rectangle r : gridOfRectangles(screenBounds, COLS, ROWS)) {
                 try {
                     SwingUtilities.invokeAndWait(() -> {
                         try {
@@ -132,12 +133,16 @@ public class MultiScreenCheckScreenIDTest {
         windowList.add(window);
     }
 
-    private static List<Rectangle> gridOfRectangles(Rectangle r, int cols) {
+    private static List<Rectangle> gridOfRectangles(Rectangle r, int cols, int rows) {
         List<Rectangle> l = new ArrayList<>();
-        for (int col = 0; col < cols; col++) {
-            int x1 = r.x + (int) Math.round(r.width * (double) col / cols);
-            int x2 = r.x + (int) Math.round(r.width * (double) (col + 1) / cols);
-            l.add(new Rectangle(x1, 0, x2 - x1, x2-x1));
+        for (int row = 0; row < rows; row++) {
+            int y1 = r.y + (int) Math.round(r.height * (double) row / rows);
+            int y2 = r.y + (int) Math.round(r.height * (double) (row + 1) / rows);
+            for (int col = 0; col < cols; col++) {
+                int x1 = r.x + (int) Math.round(r.width * (double) col / cols);
+                int x2 = r.x + (int) Math.round(r.width * (double) (col + 1) / cols);
+                l.add(new Rectangle(x1, y1, x2 - x1, y2 - y1));
+            }
         }
         return l;
     }
--- a/test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
+++ b/test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
@@ -63,6 +63,9 @@ public class MultiScreenCheckScreenIDTest extends MouseAdapter {
             for (Window win : windowList) {
                 win.dispose();
             }
+            if (window != null) {
+                window.dispose();
+            }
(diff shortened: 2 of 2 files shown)


test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
```diff
@@ -0,0 +1,149 @@
...
+        new MultiScreenCheckScreenIDTest().createWindowGrid();
+    }
+
+    private void createWindowGrid() {
+        screens = GraphicsEnvironment
+                .getLocalGraphicsEnvironment()
+                .getScreenDevices();
+
+        if(screens.length < 2) {
+            System.out.println("Testing aborted. Required min of 2 screens. Found : " + screens.length );
```
DamonGuy, line 80: There are a few lines like this that exceed the line character limit.

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
```diff
@@ -0,0 +1,149 @@
...
+    private static void createGUI() {
+        new MultiScreenCheckScreenIDTest().createWindowGrid();
+    }
+
+    private void createWindowGrid() {
+        screens = GraphicsEnvironment
+                .getLocalGraphicsEnvironment()
+                .getScreenDevices();
+
+        if(screens.length < 2) {
```
DamonGuy, line 79:
```suggestion
        if (screens.length < 2) {
```

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
```diff
@@ -0,0 +1,149 @@
...
+        window.addMouseListener(this);
+        window.setVisible(true);
+        windowList.add(window);
+
+        window.addPropertyChangeListener("graphicsConfiguration", new PropertyChangeListener() {
+            @Override
+            public void propertyChange(PropertyChangeEvent propertyChangeEvent) {
+                double windowXPos = window.getBounds().getX();
+                double screen1Width = screens[0].getDefaultConfiguration().getBounds().getWidth();
+                System.out.println("window x Position : " + windowXPos + ", screen[0] width : " + screen1Width);
```
DamonGuy, line 120: Line character limit

DamonGuy (commented):
Suggested some minor changes. I ran the changes and test in CI as well, and it passed with a higher repetition test. Test looks fine to me on Linux.

src/java.desktop/unix/classes/sun/awt/X11/XWindowPeer.java
```diff
@@ -804,10 +804,10 @@ public void handleConfigureNotifyEvent(XEvent xev) {
         if (insLog.isLoggable(PlatformLogger.Level.FINE)) {
             insLog.fine(xe.toString());
         }
-        checkIfOnNewScreen(toGlobal(new Rectangle(scaleDown(xe.get_x()),
+        checkIfOnNewScreen(new Rectangle(scaleDown(xe.get_x()),
```
azvegint, line 807: After the fix I am unable to create a translucent window on a second screen. (you can reproduce it with your test)

TejeshR13: Yeah, with [JDK-8006421](https://bugs.openjdk.org/browse/JDK-8006421) fix the transparency issue is resolved. So this bug focus on the correcting the opaque window seen in default (Starting from half of default screen). As highlighted in the below image.
image

azvegint: I see, just the TejeshR13:branch_8280482(from this PR) branch does not contain JDK-8006421 fix. That's why I observed this behavior.

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
```diff
@@ -0,0 +1,154 @@
...
+            public void propertyChange(PropertyChangeEvent propertyChangeEvent) {
+                double windowXPos = window.getBounds().getX();
+                double screen1Width = screens[0].getDefaultConfiguration()
+                        .getBounds().getWidth();
+                System.out.println("window x Position : " + windowXPos + "," +
+                        " screen[0] width : " + screen1Width);
+
+                //Check if GC is changed for windows positioned in screen 0.
+
+                if (windowXPos < screen1Width) {
```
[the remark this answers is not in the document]
TejeshR13: > _I suggest to check if window's GraphicsConfiguration is associated with correct screen device
(e.g. check if the window bounds within the window.getGraphicsConfiguration().getDevice().getBounds())._
I didn't get this suggestion. Does it means to compare GC or the position of every `window` with corresponding `window.getGraphicsConfiguration().getBounds()`.......? `window.getGraphicsConfiguration().getBounds()` shows same for all the windows, either in primary/secondary monitor.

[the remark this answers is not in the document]
azvegint: > shows same for all the windows, either in primary/secondary monitor.

For me it is not true:

[GCBounds.java.txt](https://github.com/openjdk/jdk/files/12113425/GCBounds.java.txt)

> Screen bounds java.awt.Rectangle[x=0,y=0,width=2560,height=1440]
Frame gc bounds: java.awt.Rectangle[x=0,y=0,width=2560,height=1440]
Screen bounds java.awt.Rectangle[x=2560,y=0,width=1920,height=1200]
Frame gc bounds: java.awt.Rectangle[x=2560,y=0,width=1920,height=1200]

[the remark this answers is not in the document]
TejeshR13: Looks like I was checking for `window.getGraphicsConfiguration().getBounds()` right after making it visible, waiting for the events to finish up gives right values. I have updated the test and validation is moved outside the GC change event. Tested multiple times and the test looks fine.

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
```diff
@@ -0,0 +1,154 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.util.ArrayList;
+import java.util.List;
+
+/*
+ * @test
+ * @bug 8280482
+ * @key headful
+ * @requires (os.family == "linux")
+ * @library /java/awt/regtesthelpers
```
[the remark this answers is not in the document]
TejeshR13: Removed.

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
```diff
@@ -0,0 +1,154 @@
...
+    private static final int ROWS = 8;
+    private static final Color BACKGROUND = new Color(0, 0, 255, 64);
+    private GraphicsDevice[] screens;
+    static List<Window> windowList = new ArrayList<>();
+
+    public static void main(final String[] args) throws Exception {
+        try {
+            createGUI();
+        } finally {
+            for (Window win : windowList) {
```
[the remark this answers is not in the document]
TejeshR13: Since GC change event is removed now and also added some delay.

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
```diff
@@ -0,0 +1,154 @@
...
+                try {
+                    SwingUtilities.invokeAndWait(() -> {
+                        try {
+                            createWindow(r);
+                        } catch (Exception e) {
+                            throw new RuntimeException(e);
+                        }
+                    });
+
+                    Thread.sleep(100);
```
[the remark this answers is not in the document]
TejeshR13: Removed. Also No need to test for full screen I guess, just the top row checking will be fine I guess. Updated the test to check only for top row.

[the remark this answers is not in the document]
azvegint: > Also No need to test for full screen I guess, just the top row checking will be fine I guess. 
This area is very dependent from screen arrangement:

image

image


image

image

image
image

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
```diff
@@ -0,0 +1,154 @@
...
+import java.beans.PropertyChangeListener;
+import java.lang.reflect.InvocationTargetException;
+import java.util.ArrayList;
+import java.util.List;
+
+/*
+ * @test
+ * @bug 8280482
+ * @key headful
+ * @requires (os.family == "linux")
```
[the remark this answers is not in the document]
azvegint: So we should check this, just in case other systems have similar issues.
(and remove the requires tag if there are no issues).

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
```diff
@@ -0,0 +1,144 @@
...
+                        (screens[0].getDefaultConfiguration().getBounds())) {
+                    if (!(windowList.get(windowList.size() - 1).
+                            getGraphicsConfiguration().getBounds().
+                            intersects(screens[0].getDefaultConfiguration().
+                                    getBounds()))) {
+                        throw new RuntimeException("Graphics configuration " +
+                                "changed for screen 0");
+                    }
+                } else if (windowList.get(windowList.size() - 1).getBounds().
+                        intersects(screens[1].getDefaultConfiguration().getBounds())) {
```
azvegint, line 113: What if user has more than 2 screens?
It looks like the 3rd and subsequent screens are not tested.

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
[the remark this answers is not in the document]
TejeshR13: Ok, didn't try with these settings. Will check for full screen then.

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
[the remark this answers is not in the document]
TejeshR13: I verified with Windows and Mac, not issue found. Removing the linux specific tag and other unused tag.

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
TejeshR13: I have made the validation as per number of screens now. I guess this is fine to check multiple screens.

TejeshR13:
@azvegint @DamonGuy Please let me know if there are any more review suggestions.

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
azvegint: I don't see why we should be limited to only two when more may be available.
What if it helps to find some issue?
Anyway, the test place windows on all screens, but checking only on two of them.

TejeshR13: I have updated the test to available screens (Though not able to test since I can extend to two screens now), I guess it should work, it's generic now.

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
```diff
@@ -0,0 +1,144 @@
...
+                            throw new RuntimeException(e);
+                        }
+                    });
+                } catch (InterruptedException | InvocationTargetException e) {
+                    e.printStackTrace();
+                }
+                robot.delay(50);
+                robot.waitForIdle();
+                if (windowList.get(windowList.size() - 1).getBounds().intersects
+                        (screen.getDefaultConfiguration().getBounds())) {
```
azvegint, line 103: Looks like you can reuse `screenBounds` variable for lines 103 and 106.

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
```diff
@@ -0,0 +1,144 @@
...
+                        } catch (Exception e) {
+                            throw new RuntimeException(e);
+                        }
+                    });
+                } catch (InterruptedException | InvocationTargetException e) {
+                    e.printStackTrace();
+                }
+                robot.delay(50);
+                robot.waitForIdle();
+                if (windowList.get(windowList.size() - 1).getBounds().intersects
```
azvegint, line 102: Maybe it's better to make `createWindow(r)` return a window and cache it instead of several querying of windowList?

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
```diff
@@ -0,0 +1,144 @@
...
+        window.setAlwaysOnTop(true);
+        window.addMouseListener(this);
+        window.setVisible(true);
+        windowList.add(window);
+    }
+
+    @Override
+    public void mouseClicked(MouseEvent e) {
+        ((Window) e.getSource()).dispose();
+    }
```
azvegint, line 130: New line missing.

azvegint (approved):

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
azvegint: FYI, you can arrange up to 8 screens using Virtual Box(but I don't think it's necessary in this case)

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
```diff
@@ -0,0 +1,145 @@
...
+                robot.delay(50);
+                robot.waitForIdle();
+                if (window.getBounds().intersects(screenBounds)) {
+                    if (!(window.getGraphicsConfiguration().getBounds().
+                            intersects(screenBounds))) {
+                        throw new RuntimeException("Graphics configuration " +
+                                "changed for screen :" + screenNumber);
+                    }
+                }
+                windowList.add(window);
```
azvegint, line 111: This window will not be disposed in case if the exception is thrown.

DamonGuy (approved):
LGTM after the changes on my VM

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
```diff
@@ -0,0 +1,148 @@
...
+    private void createWindowGrid() throws AWTException {
+        screens = GraphicsEnvironment
+                .getLocalGraphicsEnvironment()
+                .getScreenDevices();
+
+        if (screens.length < 2) {
+            System.out.println("Testing aborted. Required min of 2 screens. " +
+                    "Available : " + screens.length);
+            return;
+        }
```
aivanov-jdk, line 86: In such cases, it's better to use `jtreg.SkippedException` to indicate clearly that the test is not applicable in the current environment.

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
```diff
@@ -0,0 +1,148 @@
...
+        window.setAlwaysOnTop(true);
+        window.addMouseListener(this);
+        window.setVisible(true);
+        return window;
+    }
+
+    @Override
+    public void mouseClicked(MouseEvent e) {
+        ((Window) e.getSource()).dispose();
+    }
```
aivanov-jdk, line 133: In the automatic test, it's not necessary at all — you don't expect any input.

aivanov-jdk:
It has fixed the problem, I verified it using the attached reproducer and the automatic test case.

test/jdk/java/awt/Multiscreen/MultiScreenCheckScreenIDTest.java
azvegint: Sometimes it's easier to run the test directly, without the jtreg.
Dispose is useful in this case to avoid locking the user's screen.

aivanov-jdk: It's *always* easier to run the test directly without jtreg.

It was in the original test case, you can dispose of a window by clicking. I see no real problem preserving it. :)

PR #14831, opened 2023-07-11
8311864: Add ArraysSupport.hashCode(int[] a, fromIndex, length, initialValue)

pavelrappo:
This PR adds an internal method to calculate hash code from the provided integer array, offset and length into that array, and the initial hash code value.

Current options for calculating a hash code for int[] are inflexible. It's either ArraysSupport.vectorizedHashCode with an offset, length and initial value, or Arrays.hashCode with just an array.

For an arbitrary int[], unconditional vectorization might be unwarranted or puzzling. Unfortunately, it's the only hash code method that operates on an array subrange or accepts the initial hash code value.

A more convenient method could be added and then used, for example, here:

* https://github.com/openjdk/jdk/blob/0ef03f122866f010ebf50683097e9b92e41cdaad/src/java.base/share/classes/java/util/concurrent/CopyOnWriteArrayList.java#L1076-L1083

* https://github.com/openjdk/jdk/blob/0ef03f122866f010ebf50683097e9b92e41cdaad/src/java.base/share/classes/java/util/ArrayList.java#L669-L680

* https://github.com/openjdk/jdk/blob/0ef03f122866f010ebf50683097e9b92e41cdaad/src/java.base/share/classes/sun/security/pkcs10/PKCS10.java#L356-L362

This PR adds such a method and provides tests for it. Additionally, this PR adds tests for `null` passed to `java.util.Arrays.hashCode` overloads, behaviour which was specified but untested.

---------
### Progress

### Issue
 * [JDK-8311864](https://bugs.openjdk.org/browse/JDK-8311864): Add ArraysSupport.hashCode(int[] a, fromIndex, length, initialValue) (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14831/head:pull/14831` \
`$ git checkout pull/14831`

`$ git checkout pull/14831` \
`$ git pull https://git.openjdk.org/jdk.git pull/14831/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14831`

`$ git pr show -t 14831`

</details>

<a href="https://git.openjdk.org/jdk/pull/14831.diff">https://git.openjdk.org/jdk/pull/14831.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14831#issuecomment-1631152685)


pavelrappo:
If people agree with this PR in principle, I'll include variants for byte[] and Object[] in this PR. So it could be immediately used in already identified sites in JDK, once integrated.

merykitty:
> For an arbitrary int[], unconditional vectorization might be unwarranted or puzzling.

Could you clarify this statement? Thanks a lot.

merykitty:
@pavelrappo Actually I look at the implementation of `ArraySupport.vectorizedHashCode` and it does branching and only uses vector instructions if the length is sufficiently large. Please do benchmarking if you are uncertain with performance characteristics of functions. Thanks.

pavelrappo:
> @pavelrappo Actually I look at the implementation of `ArraySupport.vectorizedHashCode` and it does branching and only uses vector instructions if the length is sufficiently large. Please do benchmarking if you are uncertain with performance characteristics of functions. Thanks.

Are you sure you are talking about `vectorizedHashCode` rather than `vectorizedMismatch`? While the latter indeed branches on array length, the former consists of simple for-loops, which might not be the actual thing anyway because the whole method is `@IntrinsicCandidate`.

Regardless of any performance characteristics, the call to something named `vectorizedHashCode` might needlessly draw attention to something that does not deserve it.
 
My mental model might be wrong, but let me give you a stream analogy. Would you not pause and think if you unexpectedly saw this?

    ...
    stream().
    ...
    .parallel()
    ...

Glavo:
> > @pavelrappo Actually I look at the implementation of `ArraySupport.vectorizedHashCode` and it does branching and only uses vector instructions if the length is sufficiently large. Please do benchmarking if you are uncertain with performance characteristics of functions. Thanks.
> 
> Are you sure you are talking about `vectorizedHashCode` and not `vectorizedMismatch`? While the latter indeed branches on array length, the former consists of simple for-loops, which might not be the actual thing anyway because the whole method is `@IntrinsicCandidate`.
> 
> Regardless of any performance characteristics, the call to something named `vectorizedHashCode` might needlessly draw attention to something that does not deserve it.
> 
> My mental model might be wrong, but let me give you a stream analogy. Would you not pause and think if you unexpectedly saw this?
> 
> ```
> ...
> .stream()
> ...
> .parallel()
> ...
> ```

I don't think it's a good analogy. In my understanding, it is an obvious fact that `vectorizedHashCode` will not always vectorize, as vectorized code must always handle suffixes that cannot be vectorized.

I don't think judging outside `vectorizedHashCode` will make the code clear, unless `vectorizedHashCode` only handles the part that can be vectorized, and the processing of suffixes is also outside `vectorizedHashCode`.

RogerRiggs (commented):
These look like useful additions, plus versions for byte[] and Object[].

pavelrappo:
Found one more use case for hashCode on an array subrange:

https://github.com/openjdk/jdk/blob/0ef03f122866f010ebf50683097e9b92e41cdaad/src/jdk.zipfs/share/classes/jdk/nio/zipfs/ZipFileSystem.java#L3381-L3389

Sadly, ArraysSupport cannot be used there because it's not in the same module.

cl4es:
I think we could consider dropping `vectorized` from `vectorizedHashCode`: there's nothing in the library-side implementation for this that is explicitly setting up for vectorization of each case (unlike the `mismatch/vectorizedMismatch` pair). Handling of short arrays is dealt with in the intrinsic, and care was taken to make sure that the implementation is as fast or faster for small as well as large inputs.

It might make sense to move the tiny-input case switch to `ArraysSupport` though, and have it delegate to the IntrinsicCandidate method (the point of those is to avoid calling into the intrinsic routine for tiny inputs).

pavelrappo:
> I think we could consider dropping `vectorized` from `vectorizedHashCode`: there's nothing in the library-side implementation for this that is explicitly setting up for vectorization of each case (unlike the `mismatch/vectorizedMismatch` pair).

Sounds good; I specifically dislike "vectorized" being in the name of the method that is called by clients who don't seem to require that behaviour. It's misleading.

Method naming aside, `ArraysSupport.vectorizedHashCode` has one additional wrinkle. One of its arguments, `int basicType`, accepts untyped constants whose names might be so counterintuitive that a reviewer might insist on adding an inline comment to indicate that a particular combination of arguments is not a typo, but an intended case of _unsigned bytes_, for example:

https://github.com/openjdk/jdk/blob/2bdfa836adbeba3319bee4ee61017907d6d84d58/src/java.base/unix/classes/sun/nio/fs/UnixPath.java#L801-L811

Could we hide that `int basicType` behind overloads that have static safety better than `Object array` and are appropriately named? For example:

    public static int hashCode(byte[] array, int fromIndex, int length, int initialValue);
    public static int unsignedHashCode(byte[] array, int fromIndex, int length, int initialValue);

Related, I assume this is a typo

    jdk.internal.util.ArraysSupport#signedHashCode

and it should've been

    jdk.internal.util.ArraysSupport#unsignedHashCode
                                    ^^^
?

pavelrappo:
Perhaps surprisingly, we don't need int[], what JDK seems to use is these:

 * byte[]
 * unsigned byte[]
 * Object[]

cl4es:
Yes, it should have been `unsignedHashCode`. 

Adding a layer of `ArraysSupport.hashCode` methods with no `basicType` parameter exposed, which in turn calls a renamed `vectorizedHashCode` (`@IntrinsicCandidate private static int internalHashCode`?) seem like a reasonable cleanup on the face of it. Assuming we can retain performance characteristics. One complication is that we'd also need to expose a `utf16HashCode` that takes a `byte[]` and interprets it as holding `char` values.

cl4es:
> Perhaps surprisingly, we don't need int[]; what JDK seems to use is these:
> 
> * byte[]
> * unsigned byte[]
> * Object[]

All `vectorizedHashCode` variants are exposed (and testable) via the public `java.util.Arrays.hashCode(byte/char/short/int[])` methods.

pavelrappo:
> > Perhaps surprisingly, we don't need int[]; what JDK seems to use is these:
> > 
> > * byte[]
> > * unsigned byte[]
> > * Object[]
> 
> All `vectorizedHashCode` variants are exposed (and testable) via the public `java.util.Arrays.hashCode(byte/char/short/int[])` methods.

There seems to be misunderstanding. Here's the intent of my earlier comment: it's only those three variants above that I found to be used in JDK where a subrange and, to a lesser extent, a different initial value is desired. This is ironic since this PR is titled as "Add ArraysSupport.hashCode(**int**[] a, fromIndex, length, initialValue)". I guess it's because I filed the PR before I gathered enough stats.

One lighter-weight alternative to providing more methods would be to rename `basicType` constants to be more self-descriptive. Thoughts?

pavelrappo:
@cl4es, why are these constants from https://docs.oracle.com/javase/specs/jvms/se9/html/jvms-6.html#jvms-6.5.newarray relevant to our vectorizedHashCode method?

pavelrappo:
Even with better named constants the Object type is missing. Something needs to be done about it.

pavelrappo:
To Skara bots: keep this PR alive.

pavelrappo:
To Skara bots: keep this PR alive.

pavelrappo:
To Skara bots: keep this PR alive.

pavelrappo:
To Skara bots: keep this PR alive.

PR #14834, opened 2023-07-11
JDK-8311892: TrustManagerFactory loading an invalid keystore yield vague exception

candrews:
When loading the default JVM trust store, if the JVM trust store contains an invalid certificate, the exception contains insufficient information to determine which certificate is invalid, making it very difficult to fix the problem.

To reproduce the issue:
1. Modify the default JVM trust store to contain invalid information. A very easy way to do this on openjdk / red hat systems is to edit /etc/pki/ca-trust/extracted/java/cacerts and add garbage text to the file.
2. Run this code:
```
TrustManagerFactory = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
// initializing the trust store with a null KeyStore will load the default JVM trust store
tmf.init((KeyStore) null);
```

This stack trace results:
```
Caused by: java.security.KeyStoreException: problem accessing trust store
	at java.base/sun.security.ssl.TrustManagerFactoryImpl.engineInit(TrustManagerFactoryImpl.java:73)
	at java.base/javax.net.ssl.TrustManagerFactory.init(TrustManagerFactory.java:282)
	... 81 common frames omitted
Caused by: java.io.IOException: toDerInputStream rejects tag type 97
	at java.base/sun.security.util.DerValue.toDerInputStream(DerValue.java:1155)
	at java.base/sun.security.pkcs12.PKCS12KeyStore.engineLoad(PKCS12KeyStore.java:2013)
	at java.base/sun.security.util.KeyStoreDelegator.engineLoad(KeyStoreDelegator.java:221)
	at java.base/java.security.KeyStore.load(KeyStore.java:1473)
	at java.base/sun.security.ssl.TrustStoreManager$TrustAnchorManager.loadKeyStore(TrustStoreManager.java:390)
	at java.base/sun.security.ssl.TrustStoreManager$TrustAnchorManager.getTrustedCerts(TrustStoreManager.java:336)
	at java.base/sun.security.ssl.TrustStoreManager.getTrustedCerts(TrustStoreManager.java:57)
	at java.base/sun.security.ssl.TrustManagerFactoryImpl.engineInit(TrustManagerFactoryImpl.java:49)
	... 83 common frames omitted
```

Throwing an exception with a more detailed error message facilitates debugging and ultimately fixing such problems.

---------
### Progress

### Issue
 * [JDK-8311892](https://bugs.openjdk.org/browse/JDK-8311892): TrustManagerFactory loading an invalid keystore yield vague exception (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk.git pull/14834/head:pull/14834` \
`$ git checkout pull/14834`

`$ git checkout pull/14834` \
`$ git pull https://git.openjdk.org/jdk.git pull/14834/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 14834`

`$ git pr show -t 14834`

</details>

<a href="https://git.openjdk.org/jdk/pull/14834.diff">https://git.openjdk.org/jdk/pull/14834.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk/pull/14834#issuecomment-1631410809)

### DIFF
--- a/src/java.base/share/classes/sun/security/ssl/TrustStoreManager.java
+++ b/src/java.base/share/classes/sun/security/ssl/TrustStoreManager.java
@@ -386,7 +386,11 @@ final class TrustStoreManager {
             if (!"NONE".equals(descriptor.storeName)) {
                 try (@SuppressWarnings("removal") FileInputStream fis = AccessController.doPrivileged(
                         new OpenFileInputStreamAction(descriptor.storeFile))) {
-                    ks.load(fis, password);
+                    try {
+                        ks.load(fis, password);
+                    } catch (Exception e) {
+                        throw new KeyStoreException("Failed to load key store: " + descriptor.storeName, e);
+                    }
                 } catch (FileNotFoundException fnfe) {
                     // No file available, no KeyStore available.
                     if (SSLLogger.isOn && SSLLogger.isOn("trustmanager")) {


seanjmullan:
Can you provide an example of the stack trace you now get with your proposed fix?

candrews:
```
Caused by: java.security.KeyStoreException: problem accessing trust store
	at java.base/sun.security.ssl.TrustManagerFactoryImpl.engineInit(TrustManagerFactoryImpl.java:73)
	at java.base/javax.net.ssl.TrustManagerFactory.init(TrustManagerFactory.java:282)
	... 73 common frames omitted
Caused by: java.security.KeyStoreException: Failed to load key store: /usr/lib/jvm/java-17-openjdk-17.0.7.0.7-5.fc38.x86_64/lib/security/cacerts
	at java.base/sun.security.ssl.TrustStoreManager$TrustAnchorManager.loadKeyStore(TrustStoreManager.java:390)
	at java.base/sun.security.ssl.TrustStoreManager$TrustAnchorManager.getTrustedCerts(TrustStoreManager.java:336)
	at java.base/sun.security.ssl.TrustStoreManager.getTrustedCerts(TrustStoreManager.java:57)
	at java.base/sun.security.ssl.TrustManagerFactoryImpl.engineInit(TrustManagerFactoryImpl.java:49)
	... 79 common frames omitted
Caused by: java.io.IOException: toDerInputStream rejects tag type 97
	at java.base/sun.security.util.DerValue.toDerInputStream(DerValue.java:1155)
	at java.base/sun.security.pkcs12.PKCS12KeyStore.engine(PKCS12KeyStore.java:2013)
	at java.base/sun.security.util.KeyStoreDelegator.engineLoad(KeyStoreDelegator.java:221)
	at java.base/java.security.KeyStore.load(KeyStore.java:1473)
	... 83 common frames omitted
```

The `KeyStoreException` inside `KeyStoreException` could be eliminated by adding a `catch` clause at https://github.com/openjdk/jdk/blob/257bc1745cf275d691db1801f8dd270b9ff1b324/src/java.base/share/classes/sun/security/ssl/TrustManagerFactoryImpl.java#L67 like this:
```java
            } catch (KeyStoreException ke) {
                throw ke;
```

Please let me know if you'd like me to include that change.

Thanks!

seanjmullan:
Making this change would potentially reveal sensitive information about the file system (i.e. the pathname of the keystore) in an Exception message. This would go against the recommendations in [guideline 2.1 of the Java Secure Coding Guidelines](https://www.oracle.com/java/technologies/javase/seccodeguide.html). Thus, I do not accept this change as proposed.

You can already obtain similar helpful information by enabling debugging by specifying `-Djavax.net.debug=ssl:trustmanager` on the command-line.

candrews:
Excellent point, thank you for your thoughtful review and for linking the documentation :)

It would be really nice to give the user some information to help them along without requiring a JVM argument, though... What if we just included the file name (but not the path) in the exception message, something like:
```java
throw new KeyStoreException("Failed to load key store with file name: " + descriptor.storeFile.getName(), e);
```
Would that be acceptable? Or is there something else we could do to provide a little more helpful of a message?

seanjmullan:
> Excellent point, thank you for your thoughtful review and for linking the documentation :)
> 
> It would be really nice to give the user some information to help them along without requiring a JVM argument, though... What if we just included the file name (but not the path) in the exception message, something like:
> 
> ```java
> throw new KeyStoreException("Failed to load key store with file name: " + descriptor.storeFile.getName(), e);
> ```
> 
> Would that be acceptable? Or is there something else we could do to provide a little more helpful of a message?

It would be more acceptable, but it also may be more confusing if for example "cacerts" was emitted, and there were more than one truststore with a filename of "cacerts" in the system.

There are really only a couple of different ways a truststore can be set. See https://docs.oracle.com/en/java/javase/20/security/java-secure-socket-extension-jsse-reference-guide.html#GUID-32CF3420-56E8-4BC5-8D3B-1F6B4692A290 for more info.

I think in most cases, you should be able to figure out which truststore has the problem. Can you explain or give a scenario where that would be difficult to determine? However, what might be acceptable to me is an exception that gives a hint as to what truststore is being used, w/o giving away the full path. For example, one of the following, depending on how it was configured:

 "Problem accessing truststore specified in javax.net.ssl.trustStore system property"
 "Problem accessing truststore with file name: <java-home>/lib/security/jssecacerts"
 "Problem accessing truststore with file name: <java-home>/lib/security/cacerts"
 
This change would probably require enhancing TrustStoreManager and TrustStoreDescriptor to cache additional info about what type of truststore was used. But it is not really clear to me that the effort is justified.

If you are able to re-run the application with debugging, you can obtain this information. Also, with JFR you can get a dump of the system properties used by an application, which also would be helpful if it is setting the  javax.net.ssl.trustStore system property.

ecki:
Btw: the initial issue „which certificate“ would suggest the byte position beeing more important than the file name - especially if there can only be so many ways to specify one. Maybe instead of throwing the details just error log the path?

seanjmullan:
> Btw: the initial issue „which certificate“ would suggest the byte position beeing more important than the file name - especially if there can only be so many ways to specify one. Maybe instead of throwing the details just error log the path?

You can already get the truststore path name if you turn on debugging by specifying -Djavax.net.debug=ssl:trustmanager on the command-line.

